PR #2247, opened 2026-07-21
fix(desktop): Windows goose discovery, install PATH, and hidden consoles

tzarebczan:
## Summary

Fixes Windows onboarding/runtime install friction reported in #2239 and #2238.

### #2239 — Goose shows Installed but Next stays blocked
- Goose's official Windows installer writes to `%USERPROFILE%\goose` and does not put it on PATH.
- Buzz never searched that directory, so rediscovery stayed `NotInstalled` after a successful install.
- Onboarding showed **Installed** from local `installSuccess` while Next requires `availability === available`.
- **Fix:** probe `%USERPROFILE%\goose`, post-install resolve verification, only show Installed after discovery confirms (Checking while waiting).

### #2238 — Codex ACP install fails on Windows
- Install shells replaced PATH with only the Buzz managed npm prefix when `login_shell_path()` is `None` (always on Windows).
- System Node/npm (nvm4w, etc.) disappeared → `npm: command not found`.
- **Fix:** fall back to the process PATH on Windows install shells (managed bins still prepended).

### Console focus-steal
- Discovery probes, git, shell tool, auth/version probes lacked `CREATE_NO_WINDOW`.
- **Fix:** hide those consoles on Windows.

## Test plan
- [x] Unit: `windows_well_known_dirs` include goose path
- [x] Unit: install shell PATH falls back to process PATH without login shell
- [ ] Manual: Windows onboarding → Install Goose → Installed + Next enabled (existing `%USERPROFILE%\goose` should be found without reinstall)
- [ ] Manual: Install Codex adapter without manual `codex-acp` install (system npm present)
- [ ] Manual: setup/rediscovery no longer steals focus with Git Bash windows

Closes #2239
Closes #2238

PR #2416, opened 2026-07-22
fix(buzz-acp): suppress agent console window on Windows (CREATE_NO_WINDOW)

philpof102-svg:
## Summary

On Windows, an external ACP agent runtime (e.g. the Claude Code adapter) pops a fresh, lingering console window for every agent subprocess. This makes `buzz-acp` on Windows visually broken when the agent isn't the bundled `buzz-agent`.

## Root cause

The desktop already spawns the harness with `CREATE_NO_WINDOW` — `desktop/src-tauri/src/managed_agents/runtime.rs` sets it, with a comment noting *"Without this a bare terminal pops for buzz-acp.exe and lingers"*. So `buzz-acp` runs with **no console of its own**.

But `AcpClient::spawn` (`crates/buzz-acp/src/acp.rs`) launches the agent subprocess **without** that flag (only `#[cfg(unix)] process_group(0)` is set). Because the parent has no console to share, Windows allocates a **new visible console** for each console-subsystem agent child — the `claude-agent-acp` / `goose` cmd shim, `node`, `claude.exe`. External adapters (Claude Code runtime) hit this; the GUI-subsystem bundled `buzz-agent` does not, which is why it only shows up with external runtimes.

## Fix

Mirror the desktop's suppression at the harness spawn: set `CREATE_NO_WINDOW` on Windows in `AcpClient::spawn`, alongside the existing `#[cfg(unix)] process_group(0)`. `tokio::process::Command::creation_flags` is inherent on Windows. stdio pipes are unaffected.

```
#[cfg(windows)]
{
    const CREATE_NO_WINDOW: u32 = 0x0800_0000;
    cmd.creation_flags(CREATE_NO_WINDOW);
}
```

## Repro

Windows, Buzz 0.4.22, deploy an agent using the **Claude Code** runtime → a console window pops (per agent child) and lingers. With `buzz-agent` (GUI subsystem) it does not.

## Testing

Change is a 6-line `#[cfg(windows)]` addition mirroring code already present in the desktop crate. Full disclosure: I could not run a local Windows build to exercise it end-to-end (my machine has no MSVC linker installed), so I'm relying on the review + CI here rather than claiming a local run. Happy to adjust placement/style.

Issue #2261, opened 2026-07-21
[Bug] Windows: Codex model discovery times out / empty models (cold start >10s)


tzarebczan:
## Summary

On Windows, Codex model discovery via `buzz-acp models` (spawn adapter → ACP `initialize` → `session/new`) often takes **~15–25s** cold. Shipped `MODELS_TIMEOUT` is **10s**, so discovery fails or returns empty models even though the adapter eventually succeeds.

This breaks / degrades onboarding and agent config model pickers for Codex after a successful install.

## Environment

- OS: Windows 11
- Runtime: `@agentclientprotocol/codex-acp` (Node shim under nvm, e.g. `~\.nvm\versions\node\...\bin\codex-acp.cmd`)
- Observed: full probe ~**20s** with absolute path under a longer timeout; **times out at 10s** with product timeout

## Reproduction

1. Install Codex / `codex-acp` so discovery finds the runtime.
2. Open first-run onboarding (or agent defaults / model picker that calls `discover_agent_models` / `buzz-acp models`).
3. Observe empty models, silent failure, or timeout-shaped errors while Codex is still cold-starting.

Direct probe (illustrative):

```text
buzz-acp models --json --agent <path-to-codex-acp.cmd>
# cold: often ~20s on Windows; product budget is 10s
```

## Expected

- Discovery waits long enough for a cold Codex ACP boot, **or**
- UI shows clear **loading** for a slow probe, then a **clear timeout error** with an easy **Retry**
- Prefer caching models after first success so every UI open is not a full cold spawn

## Actual

- 10s timeout kills the probe mid-boot
- Models list empty / no useful error in some onboarding paths
- Re-probe always cold-starts Node + Codex App Server (no warm cache)

## Root cause (split)

| Layer | Responsibility |
|--------|----------------|
| **Upstream** `@agentclientprotocol/codex-acp` + Codex App Server | Real cold-start cost (stdio adapter boots Codex app server) |
| **Buzz** `MODELS_TIMEOUT` (10s) + desktop discovery UX | Timeout too short for Windows cold start; weak loading/error/retry |

Git Bash is **not** on the models spawn path (`Command::new` / absolute `.cmd` → node). Related but separate: Windows install/PATH ([#2238](https://github.com/block/buzz/issues/2238)).

## Suggested Buzz fixes

1. Raise `MODELS_TIMEOUT` (e.g. 30–45s) or make it adaptive for slower adapters
[...]

PR #2533, opened 2026-07-23
fix(desktop): preserve Windows PATH for ACP probes

aleixrodriala:
## Summary

- preserve the native process `PATH` when login-shell PATH discovery is unavailable
- include the same existing well-known Windows Node/npm directories used by the install-shell fix
- add regression tests for process-PATH fallback and login-shell precedence

This is stacked on #2247 and should be reviewed as the single additional commit `06e8a5e9`. Once #2247 lands, the effective diff should collapse to `desktop/src-tauri/src/managed_agents/runtime/path.rs`.

## Root cause

#2247 fixes PATH composition for the ACP **install subprocess**, but runtime discovery and launched agents use `managed_agents/runtime/path.rs::build_augmented_path()` instead. On Windows, `login_shell_path()` is intentionally `None`, managed Node is unsupported, and the existing builder did not preserve the native process PATH or add well-known Node directories.

As a result, a correctly installed npm shim such as `%APPDATA%\Buzz\node-tools\codex-acp.cmd` starts but cannot resolve `node.exe`. Buzz then classifies the adapter as outdated and retries installation, leaving onboarding at **Installation failed**.

## User impact

The affected Windows v0.4.23 machine had Node and Codex installed. Installing `@agentclientprotocol/codex-acp` 1.1.7 into Buzz's private prefix was not sufficient: the live onboarding UI still failed. Reproducing Buzz's exact augmented PATH produced `"node" is not recognized as an internal or external command`.

A temporary local Node delegate placed in Buzz's managed prefix made the exact probe succeed and clicking Retry changed the live UI to **Ready**. That delegate is only a local workaround and is not part of this PR; this source change fixes the underlying PATH construction.

## Validation

- affected Windows machine: failed probe reproduced with Buzz's current PATH; equivalent fixed PATH returned `@agentclientprotocol/codex-acp 1.1.7`; onboarding confirmed **Ready**
- focused `runtime/path.rs` tests: 7 passed, 0 failed
- `rustfmt --edition 2021 desktop/src-tauri/src/managed_agents/runtime/path.rs`
- `git diff --check`

The full Tauri crate test was not completed in this WSL environment because its Linux build requires the full WebKitGTK development stack.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -20,6 +20,19 @@ use crate::usage::{TurnUsage, UsageTracker};
 /// Lines exceeding this limit are rejected to prevent OOM from rogue agents.
 const MAX_LINE_SIZE: usize = 10_000_000; // 10 MB
 
+/// Suppress a console window for agent adapter children on Windows (no-op elsewhere).
+fn hide_console(cmd: &mut tokio::process::Command) {
+    #[cfg(windows)]
+    {
+        const CREATE_NO_WINDOW: u32 = 0x0800_0000;
+        cmd.creation_flags(CREATE_NO_WINDOW);
+    }
+    #[cfg(not(windows))]
+    {
+        let _ = cmd;
+    }
+}
+
 /// An MCP server configuration passed to `session/new`.
 ///
 /// Corresponds to the `McpServerStdio` variant in the ACP schema.
@@ -466,14 +479,8 @@ impl AcpClient {
         #[cfg(unix)]
         cmd.process_group(0);
 
-        // Desktop launches buzz-acp with CREATE_NO_WINDOW. Without the same flag
-        // here, each adapter (cmd.exe → node, buzz-agent, …) allocates its own
-        // console — one window per agent (#2292).
-        #[cfg(windows)]
-        {
-            const CREATE_NO_WINDOW: u32 = 0x0800_0000;
-            cmd.creation_flags(CREATE_NO_WINDOW);
-        }
+        // Desktop launches buzz-acp windowless; adapters must not allocate consoles (#2292).
+        hide_console(&mut cmd);
 
         let mut child = cmd.spawn()?;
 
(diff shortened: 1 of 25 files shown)


wpfleger96:
🤖 hey @aleixrodriala — thank you for this contribution. It covers a lot of ground: four crates, the Git Bash probe, nvm layouts, console-window suppression. Some of what it fixes is now on main and some of it we're still missing, so let me be specific about why I'm closing it.

The PATH problem it set out to fix is now handled on main by three merged commits: #2563 (`cca16635d`), #2753 (`3bd3a014c`), and #2750 (`f3981dbfe`), the last covering native PowerShell install spawning. Beyond that, the PR is 25 files across four crates with several unrelated concerns in one diff, and it's conflicting against main. Rebasing it is more work than re-landing the pieces that are still missing, and each of those would be easier to review on its own.

Two things in here that I checked and are genuinely still gaps, if you want to pick either up as a focused PR:

1. **nvm-windows discovery** — the one I'd most like to see. `find_nvm_default_bin` (`managed_agents/discovery.rs`) only understands POSIX nvm layout (`~/.nvm/versions/node`); `NVM_SYMLINK` and `NVM_HOME` appear nowhere in the tree, and the Windows arm of `common_binary_paths` probes only `%APPDATA%\npm` and `%LOCALAPPDATA%` — not `%ProgramFiles%\nodejs`. #2261 calls this out directly ("nvm4w ≠ `.nvm`"). Your `windows_paths.rs` was on exactly the right track.
2. **Git Bash probe ordering** — main has the WindowsApps alias filtering already (`is_windows_apps_alias` in `git_bash.rs`), but your version reorders the probe sequence to prefer the registry lookup over PATH-resolved bash. If that ordering fixed a real failure you hit, it's worth its own PR with the symptom described — I'd take it.

For the record, the `CREATE_NO_WINDOW` additions here are already covered on main: both spawn sites in `buzz-acp`/`buzz-agent`, `project_git_exec.rs`, and `taskkill_tree` all set the flag. The console-flash issue is tracked separately in #2685 with #2493 and #2416 open against it, so that work has a home.

PR #2547, opened 2026-07-23
fix(cli): resolve @mentions wrapped in markdown emphasis and skip code regions

### LINKED ISSUE #2526: [Bug] Bold-wrapped @mentions (**@Name**) send zero p tags from buzz-cli — no notification, exit 0, no warning
**Describe the bug**

An `@mention` written with markdown emphasis — `**@Name**`, `*@Name*`, `_@Name_` — is silently dropped by `buzz messages send`. The message posts normally, the CLI exits 0 with no warning, and the event carries **zero `p` tags**, so nobody is notified and no buzz-acp agent wakes.

This affects the CLI/SDK path only. **The Desktop composer is not affected** — its TypeScript parser already handles emphasis (see "Why Desktop is fine" below). The impact lands on agents, which post exclusively through `buzz messages send`, and on anyone scripting the CLI.

**To Reproduce**

```bash
buzz messages send --channel <uuid> --content '**@Alice** please take a look'
# exit 0, message appears in the channel, zero p tags on the event

buzz messages send --channel <uuid> --content '@Alice please take a look'
# p tag for Alice present
```

**Expected behavior**

`**@Alice**` should resolve exactly like `@Alice`. Failing that, the CLI should warn when the content contains `@` tokens but zero mentions resolved, rather than exiting 0 silently.

**Observed in production**

Three consecutive messages from one agent, same channel, same command, same intended recipients. The only difference is the `**`:

| event | mention as written | `p` tags on the event |
|---|---|---|
| `d2c755cc` | `… directly. @Atlas that assumption …` | `a8a3fd71`, `3f770d65` |
| `9a0c6e2c` | `**@Atlas** your fix is aimed …` / `**@Derek** so item 1 …` | **none at all** |
| `f9402f75` | `**@Atlas** …` plus one bare `@Atlas` inside a table | `a8a3fd71` only — the bare one |

Pulled from the relay with `buzz messages get --channel <uuid>`; `9a0c6e2c` has an empty `p` set. Two agents in that thread went unnotified and the handoff stalled until a human noticed.

**Root cause**

`crates/buzz-sdk/src/mentions.rs:123` — `extract_at_mentions_with_known` only accepts an `@` at start-of-string or preceded by ASCII whitespace:

```rust
let preceded = i == 0 || content.as_bytes()[i - 1].is_ascii_whitespace();
if !preceded { continue; }
```

In `**@Atlas` the preceding byte is `*`, so the mention is skipped. `extract_at_names` carries the identical rule at `mentions.rs:75`.

The call site is `resolve_content_mentions` at `crates/buzz-cli/src/commands/messages.rs:191`. It returns whatever the parser found and never reports an empty result — hence exit 0, no warning.

The whitespace rule is deliberate: it stops `user@host` in an email address from becoming a mention (documented at `mentions.rs:57-59`). Markdown emphasis is collateral damage.

**Why Desktop is fine — and why that is the actual defect**

Desktop has a *second, independent* implementation of the same rule, and it was fixed there over three months ago. `desktop/src/features/messages/lib/hasMention.ts:143-150`:

```ts
const pattern = new RegExp(
  `(^|\\s|\\(|[*_]{1,3}|\\|\\|)(@${escaped})(?=\\|\\||[\\s,;.!?:)\\]}*_]|$)`,
  "i",
);
```

The leading group explicitly admits `*`/`_` ×1–3, `(`, and `||`; the trailing lookahead admits `*` and `_` as well. Added in `0f93ef44` ("fix: mentions survive copy/paste from chat into composer", #328, 2026-04-15) — TipTap's Bold extension wraps pasted mentions in `**`, which broke them in exactly this way. **That commit touched only `desktop/`. The Rust parser never got the equivalent.**

So the defect is parser divergence: two implementations of "is this an @mention" giving different answers, on the two surfaces that have to agree.

**Two further divergences in the same pair — worth fixing together**

1. **Trailing emphasis breaks multi-word display names.** Rust's `is_word_boundary` (`mentions.rs:154-157`) accepts whitespace and `,;.!?:)]}` — not `*`, `_`, or `|`. For a member named `Will Pfleger`, `@Will Pfleger**` fails the known-name match, falls through to the single-word tokenizer, and yields `will`, which matches nobody. The TS lookahead accepts `*_` and handles this case.

2. **The Rust path tags mentions inside code blocks; TS does not.** `strip_code_regions` is applied only to NIP-27 URI extraction (`crates/buzz-cli/src/commands/messages.rs:534`), while `@name` extraction at line 191 receives the raw content. TS masks fenced blocks, indented blocks and backtick spans before matching (`maskMarkdownCode` in `hasMention.ts`). So an `@name` appearing inside a code sample sent from the CLI fires a real notification — a false positive mirroring the false negative above.

**Suggested direction** (a direction, not a prescription)

Bring `mentions.rs` to behavioural parity with `hasMention.ts` rather than patching the one symptom:

- accept `*`, `_`, `(`, `||` as leading delimiters at `mentions.rs:75` and `mentions.rs:123`;
- widen `is_word_boundary` (`mentions.rs:154`) to match, so trailing emphasis stops breaking multi-word names;
- apply `strip_code_regions` before `@name` extraction, not only before URI extraction.

**Do not port the TypeScript regex into Rust.** That would create a second hand-maintained delimiter list that drifts from the first exactly as the first drifted from the parser — which is this bug. Prefer widening the two existing Rust predicates against one shared "is this a delimiter" definition, and let the tests pin the behaviour. (Direction refined by Prometheus in triage; I originally suggested mirroring the regex and that was the worse call.)

Add Rust unit tests for the emphasis / spoiler / code cases. There are **51 tests in `crates/buzz-sdk/src/mentions.rs` (module at line 390+) and 18 more in `crates/buzz-cli/src/commands/messages.rs` (line 877+) — and not one of the 69 contains an emphasis character next to an `@`** (`grep -n '\*\*@\|\*@\|_@' crates/buzz-sdk/src/mentions.rs` → no matches). That gap is why the divergence survived a release.

Longer term the two parsers should be one shared implementation, or the divergence recurs.

A cheap independent mitigation, valuable on its own even if the parser is fixed: have `buzz messages send` warn on stderr when the content contains `@` tokens but zero mentions resolved. Silent zero-recipient sends are the part that costs real time.

**Related**

- #2442 — handoffs between agents don't always work. Different root cause (the agent posts nothing at all), but this is a second, independent way an agent handoff fails with no signal to anyone.
- #1743 / #1741 / #1742 — @mentions silently fail when the target agent is offline. Same failure class: the mention does not deliver and the sender gets no feedback.
- #2319 — mentions should use NIP-27. **Checked: it does not supersede this. See below — do not re-ask.**

**Does #2319 / NIP-27 make this moot? No.** Checked at `origin/main` = `acfbb1bb`:

- #2503 ("fix(desktop): community create handoff, NIP-27 mentions, workflows list") is the open PR carrying the NIP-27 work. All 15 changed files are under `desktop/` — `nip27Mentions.ts`, `rewriteMentionsToNip27.ts`, `collectMentionPubkeys.ts`, `hasMention.ts`, `MessageComposer.tsx`, `useMentionSendFlow.ts` and friends. **Zero files under `crates/`.** Same shape as `0f93ef44`: the TS side moves, the Rust side does not.
- The SDK can *read* NIP-27 — `extract_nostr_uris` at `mentions.rs:353` — but there is **no write path**: nothing in `crates/` rewrites an `@name` into a `nostr:npub1…` ref.

So after #2503 lands, `buzz messages send` still emits bare `@name` text and still resolves it through `extract_at_mentions_with_known`. The text parser stays load-bearing on the agent path specifically, which is the exact path this bug breaks. Waiting for the NIP-27 work does not save the work.

(Verified by Prometheus and independently re-checked during triage.)

**Environment**

- `block/buzz` @ `acfbb1bb` (`main`). Originally traced at `f8fd055b`; `crates/buzz-sdk/src/mentions.rs` and `crates/buzz-cli/src/commands/messages.rs` are byte-identical between the two, so every line number above is current.
- Linux; `buzz-cli` as invoked by `buzz-acp` managed agents

---

Reported by Prometheus in Buzz `#ditto`, 2026-07-23. Triaged by Muse.

bosoud:
Fixes #2526

## Problem

`buzz messages send` silently dropped any `@mention` wrapped in markdown emphasis — `**@Name**`, `*@Name*`, `_@Name_` — or preceded by `(` / `||`. The message posted with **zero `p` tags**, exit 0, no warning, so no one was notified and no buzz-acp agent woke. Desktop's TypeScript parser (`hasMention.ts`) was fixed for exactly this in `0f93ef44` (#328); the Rust parser never got the equivalent, and none of the 69 existing mention tests contained an emphasis character next to an `@`.

## Changes

Following the direction in the issue — widen the existing Rust predicates against one shared delimiter definition rather than porting the TS regex:

**`crates/buzz-sdk/src/mentions.rs`**
- New `is_mention_lead(prev, prev2)` — the single definition of "what may precede an `@`": start-of-string, ASCII whitespace, `(`, emphasis `*`/`_`, or spoiler `||`. Both `extract_at_names` and `extract_at_mentions_with_known` now use it, so the two extraction paths can no longer drift from each other.
- `is_word_boundary` widened to accept trailing `*`, `_`, and `||` (single `|` intentionally stays a non-boundary, matching TS), so `**@Will Pfleger**` matches the known multi-word name instead of falling through to a useless `will` token.
- Doc comments on both predicates point at `hasMention.ts` as the counterpart that must stay in agreement.

**`crates/buzz-cli/src/commands/messages.rs`**
- `resolve_content_mentions` now applies `strip_code_regions` before `@name` extraction (previously only NIP-27 URI extraction was stripped). An `@name` inside a fenced block or backtick span no longer fires a notification — the false-positive mirror of the reported false negative.
- `buzz messages send` prints a warning to stderr when the (code-stripped) content contains `@mention` candidates but zero mentions resolved — the independent mitigation suggested in the issue, so a zero-recipient send is no longer silent. Email addresses and `@ ` never count as candidates, so ordinary prose does not warn.

## Tests

- 10 new unit tests in `buzz-sdk` covering bold/italic/spoiler/paren leads, trailing emphasis on multi-word names, the single-`|` non-delimiter, emails inside emphasis, code-region stripping, and the deliberate `_@carol_` fallback-tokenizer trade-off (the `_` glues to the name in the *fallback* path only; the known-names path resolves it correctly — pinned in a test with a comment).
- 5 new tests in `buzz-cli` covering the send-pipeline regression case (`**@Atlas**`), code-region skipping, and the `mentions_went_unresolved` warning predicate.
- `cargo test -p buzz-sdk -p buzz-cli`, `cargo clippy --all-targets`, and `cargo fmt` all clean.

https://claude.ai/code/session_01CqMxL1inNtU7icywZxbrGY

### DIFF
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -9,8 +9,8 @@ use crate::validate::{
     validate_content_size, validate_hex64, validate_uuid, MAX_DIFF_BYTES,
 };
 use buzz_sdk::mentions::{
-    extract_at_mentions_with_known, extract_nostr_uris, merge_mentions, strip_code_regions,
-    MENTION_CAP,
+    extract_at_mentions_with_known, extract_at_names, extract_nostr_uris, merge_mentions,
+    strip_code_regions, MENTION_CAP,
 };
 
 /// Extract the thread root event ID from a Nostr tag array.
@@ -121,16 +121,22 @@ async fn resolve_channel_id(client: &BuzzClient, event_id: &str) -> Result<Uuid,
 
 /// Resolve `@name` mentions in `content` against this channel's members.
 ///
-/// Queries kind 39002 (channel members) then kind 0 (profiles), parses
-/// display names once, and feeds them to [`extract_at_mentions_with_known`]
-/// for multi-word matching. On any I/O or parse failure, returns an empty
-/// vec — auto-tagging is best-effort and must never block a send.
+/// Strips code regions first — an `@name` inside a fenced block or backtick
+/// span must not notify anyone. Queries kind 39002 (channel members) then
+/// kind 0 (profiles), parses display names once, and feeds them to
+/// [`extract_at_mentions_with_known`] for multi-word matching. On any I/O or
+/// parse failure, returns an empty vec — auto-tagging is best-effort and
+/// must never block a send.
 async fn resolve_content_mentions(
     client: &BuzzClient,
     channel_id: &str,
     content: &str,
 ) -> Vec<String> {
     if !content.contains('@') {
+        return vec![];
+    }
+    let scannable = strip_code_regions(content);
+    if !scannable.contains('@') {
         return vec![];
     }
 
@@ -188,7 +194,7 @@ async fn resolve_content_mentions(
 
     // 4. Two-pass extraction: known multi-word names first, single-word fallback.
     let known_refs: Vec<&str> = display_names.iter().map(|s| s.as_str()).collect();
-    let names = extract_at_mentions_with_known(content, &known_refs);
+    let names = extract_at_mentions_with_known(&scannable, &known_refs);
 
     // 5. Look up matched names → pubkeys via the map we already built.
     names
@@ -196,6 +202,17 @@ async fn resolve_content_mentions(
         .flat_map(|n| name_to_pubkeys.get(n).into_iter().flatten())
         .cloned()
         .collect()
+}
+
+/// True when code-stripped content carries `@mention` candidates but zero
+/// pubkeys were resolved — the silent-failure case of issue #2526, where a
+/// send exits 0 and nobody is notified.
+///
+/// `scannable` must already have code regions stripped, so `@names` inside
+/// code samples don't trigger a spurious warning. Email addresses never
+/// count as candidates (see [`extract_at_names`]).
+fn mentions_went_unresolved(scannable: &str, resolved: &[String]) -> bool {
+    resolved.is_empty() && !extract_at_names(scannable).is_empty()
 }
 
 /// Fetch raw events for `filter` via the relay's `/query` endpoint.
@@ -534,6 +551,13 @@ pub async fn cmd_send_message(
     let stripped = strip_code_regions(&p.content);
     let uri_pubkeys = extract_nostr_uris(&stripped);
     merge_mentions(&mut auto_resolved, &uri_pubkeys, MENTION_CAP);
+
+    if mentions_went_unresolved(&stripped, &auto_resolved) {
+        eprintln!(
+            "warning: content contains @mentions but none resolved to a channel member; \
+             no one will be notified"
+        );
+    }
 
     let mention_refs: Vec<&str> = auto_resolved.iter().map(|s| s.as_str()).collect();
 
@@ -876,9 +900,12 @@ pub async fn dispatch(
 
 #[cfg(test)]
 mod tests {
-    use super::{find_root_from_tags, match_profiles_by_name, parse_member_pubkeys};
+    use super::{
+        find_root_from_tags, match_profiles_by_name, mentions_went_unresolved, parse_member_pubkeys,
+    };
     use buzz_sdk::mentions::{
-        extract_at_mentions_with_known, extract_at_names, match_names_to_profiles, MentionProfile,
+        extract_at_mentions_with_known, extract_at_names, match_names_to_profiles,
+        strip_code_regions, MentionProfile,
     };
     use serde_json::json;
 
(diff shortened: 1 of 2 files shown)


SeanGearin:
Nice — this matches the direction from the triage notes, and the zero-resolved-mentions warning is a good addition. Two things worth a look:

1. **Table pipes.** Keeping a single `|` a non-boundary means a mention flush against markdown table pipes (`|@Atlas|`) still resolves nothing, while `f9402f75` in the issue shows table mentions are expected to tag (that one only worked because of surrounding spaces). Accepting a single `|` on both sides is a strict superset of `||` and keeps the delimiter predicate single-character. Repro: `extract_at_mentions_with_known("|@Atlas|status|", &["Atlas"])` — currently `[]`, expected `["atlas"]`.

2. **There's a third parser this doesn't reach.** `resolve_mention_pubkeys` in `crates/buzz-relay/src/workflow_sink.rs` is an independent @-mention parser — its doc comment says it "defines the matching contract" for workflow-emitted kind:9 `p` tags, and its left-boundary check (`workflow_sink.rs:102`) still accepts only start-of-string / whitespace / `(`. Those `p` tags are what ACP agent wake is gated on (`event_mentions_agent`, `buzz-acp/src/lib.rs:2656`), so a workflow-generated `**@Robby**` emits zero `p` tags — the exact stalled-handoff failure mode from this issue, on a surface where the text is machine-generated. Nothing this PR needs to absorb, but the issue's "longer term the parsers should be one implementation" note actually spans three parsers, not two — worth a follow-up either way.

bosoud:
Thanks for the review @SeanGearin!

1. **Table pipes** — done in 0a39d45d. A single `|` is now accepted as both lead and boundary; your repro (`|@Atlas|status|` → `["atlas"]`) is pinned as a test, along with `@Will Pfleger|x` resolving the full known name. It also simplified both predicates back to single-character checks (the two-char `||` special case is gone). Documented in the doc comments as a deliberate superset of the TS regex's `||`-only rule so a future parity sweep doesn't "fix" it backwards.

2. **Third parser** — confirmed: `is_left_boundary` in `workflow_sink.rs` accepts only start / whitespace / `(` (its right side is already emphasis-tolerant, though a trailing `_` extends the name there). Filed #2686 to track it, with a note that since `buzz-relay` already depends on `buzz-sdk`, the cheapest convergence is for `workflow_sink` to call the now-shared `is_mention_lead` — three hand-synced parsers down to two. Happy to pick that up as a follow-up PR once this one settles.

bosoud:
@SeanGearin gentle ping when you have a moment — both review points are addressed (table pipes in 0a39d45d with your repro pinned as a test; #2686 filed for the third parser). No rush on my end, but Chessing234 has offered to take the buzz-sdk convergence PR for workflow_sink on top of this one (see #2691), so this is now the head of a small train. 🚂

SeanGearin (approved):
Both points landed exactly right. The single-`|` superset is the correct call - collapsing both predicates back to single-character checks is simpler than what I asked for, and pinning the `|@Atlas|status|` repro plus documenting the deliberate TS divergence means the parity sweep can't walk it back. Thanks for filing #2686 with the convergence path scoped; the shared `is_mention_lead` route is the right shape and I'm glad Chessing234 is picking it up. Approving - this unblocks the train. 🚂

PR #2575, opened 2026-07-23
fix(relay): send 1012 restart close to all clients on graceful drain

tlongwell-block:
## Problem

When a relay pod is rolled/restarted, connected clients are never told. The SIGTERM path flips readiness to 503 and gracefully drains the axum listeners — but **upgraded WebSocket connections outlive the listener drain**. Every live client rides the dying pod until the forced `process::exit(1)` ~35s later and learns about the restart from a raw TCP reset. On an abrupt kill it's worse: the desktop stall watchdog needs up to 60s of inbound silence before it declares the socket dead.

This is the relay half of the "desktop must aggressively reconnect when its pod dies" work (client half: #2564).

## Solution

At drain start, fan a Close frame — code **1012 Service Restart**, reason `"relay restarting"` — out to every live connection via its control channel, then cancel its token. This reuses the exact "queue frame on ctrl, then cancel" idiom `disconnect_pubkey` already relies on: the send loop drains queued control frames ahead of its cancel branch, so the close reaches the client before the socket drops. Best-effort by design — a wedged control buffer still gets the close via cancel, just without the frame.

Effects:
- Clients get an immediate, well-attributed reconnect signal ~5s after SIGTERM instead of 35s (graceful) / 60s (stall-watchdog) of silence.
- The pod actually drains its WS connections instead of always hitting the 30s force-exit.
- 1012 gives the client a clean hook for a fast-reconnect class (server restart ≠ policy close ≠ network flap) — follow-up on the client side.

## Validation

- New unit tests: `drain_all_sends_restart_close_and_cancels_every_conn` (cross-community, frame content), `drain_all_full_control_buffer_still_cancels` (best-effort path).
- Full `cargo test -p buzz-relay` at HEAD: 737 passed, 1 failed — the failure is `api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo`, which fails identically on clean `origin/main` in the same environment (pre-existing flake, verified via stash/run/pop).
- `cargo clippy -p buzz-relay --all-targets` clean, `cargo fmt` clean.
- **Live proof**: spawned the relay, attached a raw WS client, sent SIGTERM. Client received `CLOSE code=1012 reason='relay restarting'` at t=5.00s in **14/14 runs** (previously: nothing until process exit).

## Follow-up (tracked in the reconnect thread)

- Real-process kill/restart e2e proving the desktop client reconnects + resubscribes against an actual relay — the trust gate for this and #2564.
- Client-side fast-reconnect policy keyed on 1012.

### DIFF
--- a/crates/buzz-relay/src/main.rs
+++ b/crates/buzz-relay/src/main.rs
@@ -1127,6 +1127,7 @@ async fn serve(
 
     let (shutdown_tx, _) = tokio::sync::watch::channel(false);
     let shutdown_flag = Arc::clone(&state.shutting_down);
+    let drain_conn_manager = Arc::clone(&state.conn_manager);
     let tx = shutdown_tx.clone();
     tokio::spawn(async move {
         shutdown_signal().await;
@@ -1136,6 +1137,16 @@ async fn serve(
         tokio::time::sleep(std::time::Duration::from_secs(5)).await;
         info!("Starting graceful drain (30s timeout)");
         let _ = tx.send(true);
+        // Tell every connected client to reconnect NOW. Without this, upgraded
+        // WebSocket connections outlive the listener drain: clients ride the
+        // dying pod until the forced exit below and only learn about the
+        // restart from a TCP reset. The 1012 close frame turns a 35s silent
+        // death into an immediate, well-attributed reconnect.
+        let closed = drain_conn_manager.drain_all();
+        info!(
+            connections = closed,
+            "Sent restart close frame to all live WebSocket connections"
+        );
         // Hard timeout: force exit if connections don't drain within 30s.
         tokio::time::sleep(std::time::Duration::from_secs(30)).await;
         tracing::error!("Drain timeout exceeded — forcing exit");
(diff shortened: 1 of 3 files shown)


tlongwell-block:
Pushed `e30dc8f99` closing the shutdown-boundary race found in review (thanks @wren):

**Race:** `handle_connection` awaits the community-active DB check before registering, so an upgrade accepted just before SIGTERM could register *after* `drain_all()`'s one-shot snapshot — missing both the 1012 and the cancel, pinning graceful drain to the 30s force-exit. Separately, `nip11_or_ws_handler` never consulted `shutting_down`, so direct/in-flight upgrades could still enter during the 5s pre-drain grace.

**Fix:**
- Sticky `draining` flag in `ConnectionManager`: `drain_all` does store-then-iterate; `register` does insert-then-check and self-signals (queue 1012 on ctrl, then cancel). Either the iteration sees the entry or the check sees the flag — no interleaving survives.
- Router rejects WS upgrades with 503 once `shutting_down` flips; NIP-11/SPA paths unaffected. Clients treat it as a dial failure and retry onto a healthy pod.

**Verification at `e30dc8f99`:**
- New deterministic test `register_after_drain_self_signals_restart_close_and_cancel` — registers after `drain_all`, asserts the late registration still gets the 1012 + cancellation.
- Full `-p buzz-relay --lib`: 738 passed, 1 failed = `api::mesh_demo::...round_trips_echo`, pre-existing flake (red on clean base too). Clippy + fmt clean.
- Live SIGTERM run: pre-established socket gets `1012 "relay restarting"` at t=5.00s; new upgrade attempts at t=1s and t=3s inside the grace window are rejected with HTTP 503.

PR #2577, opened 2026-07-23
fix(acp): surface agent authentication failures

purybr365:
## Summary

- classify adapter-reported authentication failures centrally in `AcpError`
- surface terminal credential failures immediately through the existing threaded failure-notice path
- avoid silently retrying the same credential failure while preserving retry behavior for unrelated adapter errors

## Problem

An ACP adapter can initialize successfully and later return a generic JSON-RPC error such as `-32000: Authentication required` while processing a prompt. The harness previously treated that as transient, requeued it with exponential backoff, and produced no channel-visible explanation while the process remained alive.

ACP does not currently define one universal authentication error code, so the shared ACP error layer recognizes Buzz's typed `-32001` code plus common adapter-agnostic authentication messages. Queue orchestration only consumes that classification and reuses the existing signed, threaded failure-notice transport.

## Validation

- `CARGO_INCREMENTAL=0 just ci`
- `cargo test -p buzz-acp` (584 unit tests, 9 integration tests, doc tests)
- `git diff --check`

## Checklist

- [x] Changes are focused and use the existing ACP error and failure-notice architecture
- [x] Authentication and negative retry-policy cases are covered by tests
- [x] Full repository CI gate passes
- [x] No public API or endpoint documentation changes are required

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -108,6 +108,27 @@ pub enum AcpError {
     AgentError { code: i64, message: String },
 }
 
+impl AcpError {
+    /// Whether this adapter-reported error requires the user to authenticate
+    /// before another prompt can succeed.
+    ///
+    /// ACP adapters do not currently share a dedicated authentication error
+    /// code: Buzz's native agent uses `-32001`, while other adapters commonly
+    /// return the generic JSON-RPC server-error code (`-32000`) with an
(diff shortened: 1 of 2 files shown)


purybr365:
Closing as superseded by #2751, which was merged in commit 0a9c26ee8 and implements the same authentication-error handling: classify auth failures, avoid futile retries, surface a user-facing re-authentication notice, and cover the behavior with regression tests. Thanks!

PR #2578, opened 2026-07-23
Improve huddle audio failure handling

damienrj:
## Summary

- Preserve plain-string errors returned by Tauri and show an actionable message when huddle audio is unavailable in the relay deployment.
- Use one error formatter across channel, timeline, wave, and profile huddle actions while preserving other relay and device errors.
- Complete the huddle lifecycle when audio setup fails after publishing a start event, preventing peers and reloaded clients from reconstructing a phantom active huddle.
- Cover unavailable-audio formatting and START → rollback-END reconstruction with regression tests.

## Behavior

The Tauri huddle commands reject with a plain string. Several desktop toast call sites only preserved JavaScript `Error` objects, so the relay message was discarded and replaced with “Failed to join huddle.”

The desktop now recognizes `huddle_audio_unavailable` and the current relay message, then shows:

> Huddle audio isn’t available on this server. Ask an administrator to turn it on.

Other relay and device messages remain intact, including microphone errors.

`start_huddle` also publishes `KIND_HUDDLE_STARTED` before audio setup. If setup fails, rollback now publishes `KIND_HUDDLE_ENDED` through the normal end-and-archive path before resetting local state. This makes the failed start observable to lifecycle reconstruction and prevents stale join affordances.

## Checks

- `cd desktop && pnpm test` — 3,405 passed
- `cd desktop/src-tauri && cargo test` — 1,560 passed, 13 ignored; 3 diagnostic tests passed
- `cd desktop && pnpm exec playwright test tests/e2e/channels.spec.ts --project=smoke --grep 'huddle rollback end event'` — passed
- `just desktop-tauri-check`
- `cd desktop && pnpm check`
- Pre-commit and pre-push hooks

### DIFF
--- a/desktop/src-tauri/src/huddle/mod.rs
+++ b/desktop/src-tauri/src/huddle/mod.rs
@@ -289,14 +289,9 @@ pub async fn start_huddle(
             // Audio relay failure is fatal — no point in a huddle without audio.
             if let Err(e) = post_connect_setup(&state, &ephemeral_channel_id).await {
                 // Rollback: audio relay failed after state was committed.
-                // Archive the ephemeral channel and reset state.
-                if let Ok(archive_builder) = events::build_archive(ephemeral_uuid) {
-                    if let Err(ae) = submit_event(archive_builder, &state).await {
-                        eprintln!(
-                            "buzz-desktop: rollback archive of {ephemeral_channel_id} failed: {ae}"
-                        );
-                    }
-                }
+                // Publish the terminal lifecycle event before archiving so
+                // other clients do not reconstruct a phantom active huddle.
+                emit_end_and_archive(&parent_channel_id, &ephemeral_channel_id, &state).await;
                 if let Ok(mut hs) = state.huddle_state.lock() {
                     hs.reset_preserving_generation();
                 }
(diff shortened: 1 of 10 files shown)


wpfleger96:
I think failed starts can leave a phantom active huddle for other or reloaded clients. `desktop/src-tauri/src/huddle/mod.rs` publishes `KIND_HUDDLE_STARTED` before audio setup, but the unavailable-audio rollback archives the ephemeral channel without publishing `KIND_HUDDLE_ENDED`. `HuddleIndicator` reconstructs state solely from lifecycle events, so peers—or the starter after remount—can keep seeing and trying to join a huddle that never became usable.

This looks adjacent and pre-existing rather than introduced by the formatter, but this PR's target failure mode exercises it directly. Could we publish an ended event during rollback, or otherwise make rollback observable to lifecycle reconstruction, and cover start → rollback reconstruction?

damienrj:
> I think failed starts can leave a phantom active huddle for other or reloaded clients. `desktop/src-tauri/src/huddle/mod.rs` publishes `KIND_HUDDLE_STARTED` before audio setup, but the unavailable-audio rollback archives the ephemeral channel without publishing `KIND_HUDDLE_ENDED`. `HuddleIndicator` reconstructs state solely from lifecycle events, so peers—or the starter after remount—can keep seeing and trying to join a huddle that never became usable.
> 
> This looks adjacent and pre-existing rather than introduced by the formatter, but this PR's target failure mode exercises it directly. Could we publish an ended event during rollback, or otherwise make rollback observable to lifecycle reconstruction, and cover start → rollback reconstruction?

I went ahead now and improve huddle audio failure handling.

wpfleger96 (approved):

PR #2582, opened 2026-07-23
Lock internal local agents to approved relays

brow:
### Summary

Internal (managed) builds now lock `BackendKind::Local` agents to a baked-in relay allowlist. OSS builds and provider-backed agents keep unrestricted relay configuration.

- [Bake a newline-delimited local-agent relay-origin allowlist into managed builds](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/build.rs#L18-L35) through `BUZZ_BUILD_LOCAL_AGENT_RELAY_ALLOWLIST` at compile time.
- [Require exact canonical WebSocket origin matches for local agents](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/relay_policy.rs#L10-L23), with [exact equality instead of wildcard or hostname-suffix trust](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/relay_policy.rs#L61-L80).
- [Fail closed in internal builds](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/relay_policy.rs#L26-L46) when the allowlist is missing, empty, or malformed, or when the effective relay is not allowed. [OSS builds and provider-backed agents bypass this policy](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/relay_policy.rs#L61-L72).
- [Enforce the policy at every path that can put a local agent on a relay](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/relay_policy.rs#L83-L108):
  - [saving a new or changed legacy relay pin](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/storage.rs#L301-L351)
  - [creating an agent, including an agent with no explicit relay pin](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/agents.rs#L590-L640)
  - [updating an agent's relay or name](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/agent_models.rs#L798-L920)
  - [adding an agent to a channel](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/channels.rs#L783-L795) or [changing its channel role](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/channels.rs#L836-L860)
  - [starting a huddle with agents](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/huddle/mod.rs#L172-L200) or [adding an agent to an active huddle](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/huddle/mod.rs#L938-L946)
  - [probing relay access](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/runtime_commands.rs#L390-L405) and [starting or restarting a runtime pair](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/runtime_commands.rs#L239-L267)
  - [reconciling an agent profile after start or restore](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/agents_profile.rs#L88-L110)
  - [publishing linked agent profiles after a persona name or avatar change](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/personas/mod.rs#L197-L218)
  - [importing one agent from a backup](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/personas/snapshot/import.rs#L310-L348) or [importing a team from a backup](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/team_snapshot.rs#L499-L533)
  - [the final child-process spawn boundary](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/runtime.rs#L1642-L1653)

  The spawn check is the final backstop. The earlier checks are still needed so users get a clear error at the action they took, before data is changed or relay work begins.
- [Importing an agent or team from a backup file creates fresh local agent identities, so the import checks the workspace relay before creating them](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/personas/snapshot/import.rs#L291-L348). The import then [keeps using that checked relay for the new agent's profile and memory](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/personas/snapshot/import.rs#L534-L545), and [does the same for every member of an imported team](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/commands/team_snapshot.rs#L782-L790). This prevents a race where the relay setting changes mid-import.
- [A corrupted agent store is now preserved and reported as an error](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/storage.rs#L185-L193), and [saving no longer treats a failed read as an empty store](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/src/managed_agents/storage.rs#L322-L351). Previously, a later save could silently replace the corrupted store from an empty starting point.
- [Show a warning and disable local-agent attachment when the active community's relay is blocked](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src/features/channels/ui/AddChannelBotDialog.tsx#L199-L204), with [the relay-policy query keyed to the active community](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src/features/agents/hooks.ts#L919-L930).

This PR is stacked on [#2537](https://github.com/block/buzz/pull/2537). The two build variables serve different purposes: [`BUZZ_BUILD_INTERNAL` marks the distribution as an internal build](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/build.rs#L22-L26), while [`BUZZ_BUILD_LOCAL_AGENT_RELAY_ALLOWLIST` supplies the relay origins approved for that deployment](https://github.com/block/buzz/blob/eb7354dd80e0a3205a1c52b4fcbfe48e6ddd6654/desktop/src-tauri/build.rs#L28-L35).

### Packaging contract

Internal packaging must set both:

```sh
BUZZ_BUILD_INTERNAL=1
BUZZ_BUILD_LOCAL_AGENT_RELAY_ALLOWLIST=wss://buzz.block.builderlab.xyz
```

Future internal or staging origins must be explicit newline-delimited entries in [`buzz-releases`](https://github.com/squareup/buzz-releases/pull/75) (Block-internal packaging configuration).

### Screenshots

| Before: base behavior | After: managed-build warning |
|---|---|
| Before: Add agents dialog without relay warning | After: Add agents dialog with blocked-relay warning |

Both screenshots are cropped to the same 576 x 398 dialog canvas.

### Validation

At [source commit `931865f0478b25e3566637dc824f7cc3a3e387a1`](https://github.com/block/buzz/commit/931865f0478b25e3566637dc824f7cc3a3e387a1), whose code was reapplied as `eb7354dd`, the full desktop Rust suite passed with 1,582 tests and 13 ignored tests, plus 3 diagnostic integration tests. TypeScript typechecking passed, all 3,402 JavaScript unit tests passed, and 2 focused browser tests passed for the blocked internal-build case and unaffected OSS case. Formatting, static checks, file checks, and commit hooks also passed.

A separate internal-build test run confirmed the relay-policy behavior. Six existing tests from #2537 still encode OSS access expectations and fail when internal mode intentionally forces owner-only access.

### DIFF
--- a/desktop/src-tauri/src/commands/channels.rs
+++ b/desktop/src-tauri/src/commands/channels.rs
@@ -1,4 +1,4 @@
-use tauri::State;
+use tauri::{AppHandle, State};
 
 use crate::{
     app_state::AppState,
@@ -785,9 +785,20 @@ pub async fn add_channel_members(
     channel_id: String,
     pubkeys: Vec<String>,
     role: Option<String>,
+    app: AppHandle,
     state: State<'_, AppState>,
 ) -> Result<serde_json::Value, String> {
     let uuid = parse_channel_uuid(&channel_id)?;
+    let relay_url = crate::relay::relay_ws_url_with_override(&state);
+    let local_agents = crate::managed_agents::load_managed_agents(&app)?;
+    for pubkey in &pubkeys {
+        if let Some(record) = local_agents.iter().find(|record| {
+            record.pubkey.eq_ignore_ascii_case(pubkey)
+                && record.backend == crate::managed_agents::BackendKind::Local
+        }) {
+            crate::managed_agents::validate_local_agent_relay(&record.backend, &relay_url)?;
+        }
+    }
     let role_str = match role.as_deref() {
         Some("admin") => Some("admin"),
         Some("bot") => Some("bot"),
--- a/desktop/src-tauri/src/commands/channels.rs
+++ b/desktop/src-tauri/src/commands/channels.rs
@@ -791,14 +791,7 @@ pub async fn add_channel_members(
     let uuid = parse_channel_uuid(&channel_id)?;
     let relay_url = crate::relay::relay_ws_url_with_override(&state);
     let local_agents = crate::managed_agents::load_managed_agents(&app)?;
-    for pubkey in &pubkeys {
-        if let Some(record) = local_agents.iter().find(|record| {
-            record.pubkey.eq_ignore_ascii_case(pubkey)
-                && record.backend == crate::managed_agents::BackendKind::Local
-        }) {
-            crate::managed_agents::validate_local_agent_relay(&record.backend, &relay_url)?;
-        }
-    }
+    crate::managed_agents::validate_local_agent_members(&local_agents, &pubkeys, &relay_url)?;
     let role_str = match role.as_deref() {
         Some("admin") => Some("admin"),
         Some("bot") => Some("bot"),
--- a/desktop/src-tauri/src/commands/channels.rs
+++ b/desktop/src-tauri/src/commands/channels.rs
@@ -790,8 +790,9 @@ pub async fn add_channel_members(
 ) -> Result<serde_json::Value, String> {
     let uuid = parse_channel_uuid(&channel_id)?;
     let relay_url = crate::relay::relay_ws_url_with_override(&state);
-    let local_agents = crate::managed_agents::load_managed_agents(&app)?;
-    crate::managed_agents::validate_local_agent_members(&local_agents, &pubkeys, &relay_url)?;
+    crate::managed_agents::validate_local_agent_members_from_store(&pubkeys, &relay_url, || {
+        crate::managed_agents::load_managed_agents(&app)
+    })?;
     let role_str = match role.as_deref() {
         Some("admin") => Some("admin"),
         Some("bot") => Some("bot"),
--- a/desktop/src-tauri/src/commands/channels.rs
+++ b/desktop/src-tauri/src/commands/channels.rs
@@ -838,6 +838,7 @@ pub async fn change_channel_member_role(
     channel_id: String,
     pubkey: String,
     role: String,
+    app: AppHandle,
     state: State<'_, AppState>,
 ) -> Result<(), String> {
     let uuid = parse_channel_uuid(&channel_id)?;
@@ -848,6 +849,12 @@ pub async fn change_channel_member_role(
         "owner" => return Err("cannot assign owner role — use transfer ownership".into()),
         other => return Err(format!("invalid role: {other}")),
     };
+    let relay_url = crate::relay::relay_ws_url_with_override(&state);
+    crate::managed_agents::validate_local_agent_members_from_store(
+        std::slice::from_ref(&pubkey),
+        &relay_url,
+        || crate::managed_agents::load_managed_agents(&app),
+    )?;
     let builder = events::build_add_member(uuid, &pubkey, Some(role_str))?;
     submit_event(builder, &state).await?;
     Ok(())
--- a/desktop/src-tauri/src/managed_agents/storage.rs
+++ b/desktop/src-tauri/src/managed_agents/storage.rs
@@ -199,6 +199,9 @@ fn load_agent_store(app: &AppHandle) -> Result<Vec<ManagedAgentRecord>, String> 
 pub fn load_managed_agents(app: &AppHandle) -> Result<Vec<ManagedAgentRecord>, String> {
     let mut records = load_agent_store(app)?;
     records.retain(|record| !record.pubkey.is_empty());
+    records
+        .iter()
+        .try_for_each(super::validate_managed_agent_relay_pin)?;
     hydrate_keys(&mut records);
     Ok(records)
 }
@@ -302,6 +305,7 @@ pub fn save_managed_agents(app: &AppHandle, records: &[ManagedAgentRecord]) -> R
     let mut sorted = records.to_vec();
     for record in &mut sorted {
         super::normalize_managed_agent_access(record);
+        super::validate_managed_agent_relay_pin(record)?;
     }
     // A caller-supplied key-less record would collide with the definition
     // half re-read below; instances always carry a pubkey.
--- a/desktop/src-tauri/src/managed_agents/storage.rs
+++ b/desktop/src-tauri/src/managed_agents/storage.rs
@@ -193,15 +193,18 @@ fn load_agent_store(app: &AppHandle) -> Result<Vec<ManagedAgentRecord>, String> 
     })
 }
 
+fn retain_managed_agent_instances(records: &mut Vec<ManagedAgentRecord>) {
+    records.retain(|record| !record.pubkey.is_empty());
+}
+
 /// Load the keyed agent *instances*. Key-less definitions (former personas,
 /// folded into the same store) are filtered out so every pre-fold call site
-/// keeps seeing exactly the records it always did.
+/// keeps seeing exactly the records it always did. Relay-policy violations are
+/// enforced at save/attach/start boundaries, not here, so a bad legacy pin can
+/// still be listed, fixed, or deleted.
 pub fn load_managed_agents(app: &AppHandle) -> Result<Vec<ManagedAgentRecord>, String> {
     let mut records = load_agent_store(app)?;
-    records.retain(|record| !record.pubkey.is_empty());
-    records
-        .iter()
-        .try_for_each(super::validate_managed_agent_relay_pin)?;
+    retain_managed_agent_instances(&mut records);
     hydrate_keys(&mut records);
     Ok(records)
 }
@@ -295,17 +298,49 @@ fn hydrate_keys_with(store: &impl KeyStore, records: &mut [ManagedAgentRecord]) 
     }
 }
 
+fn validate_changed_relay_pins_with<F>(
+    records: &[ManagedAgentRecord],
+    previous: &[ManagedAgentRecord],
+    validate: F,
+) -> Result<(), String>
+where
+    F: Fn(&ManagedAgentRecord) -> Result<(), String>,
+{
+    for record in records {
+        let unchanged = previous.iter().any(|old| {
+            old.pubkey == record.pubkey
+                && old.backend == record.backend
+                && old.relay_url == record.relay_url
+        });
+        if !unchanged {
+            validate(record)?;
+        }
+    }
+    Ok(())
+}
+
 /// Save the keyed agent *instances*, preserving the key-less definitions that
 /// share the unified store: callers pass exactly the records they loaded via
 /// [`load_managed_agents`], and this re-reads the definition half from disk
 /// before the wholesale rewrite so a definition is never dropped by an
 /// instance-side save (and vice versa via [`save_agent_definitions`]).
 pub fn save_managed_agents(app: &AppHandle, records: &[ManagedAgentRecord]) -> Result<(), String> {
-    let definitions = load_agent_definitions(app).unwrap_or_default();
+    let stored = load_agent_store(app).unwrap_or_default();
+    let definitions = stored
+        .iter()
+        .filter(|record| record.pubkey.is_empty())
+        .cloned()
+        .collect();
+    let previous_instances: Vec<_> = stored
+        .into_iter()
+        .filter(|record| !record.pubkey.is_empty())
+        .collect();
+    validate_changed_relay_pins_with(records, &previous_instances, |record| {
+        super::validate_managed_agent_relay_pin(record)
+    })?;
     let mut sorted = records.to_vec();
     for record in &mut sorted {
         super::normalize_managed_agent_access(record);
-        super::validate_managed_agent_relay_pin(record)?;
     }
     // A caller-supplied key-less record would collide with the definition
     // half re-read below; instances always carry a pubkey.
@@ -802,7 +837,8 @@ mod tests {
 
     use super::{
         agent_keyring_name, hydrate_keys_with, migrate_inline_key, persist_agent_keys_with,
-        KeyMigration, KeyStore, KeyringProbe, ManagedAgentRecord,
+        retain_managed_agent_instances, validate_changed_relay_pins_with, KeyMigration, KeyStore,
+        KeyringProbe, ManagedAgentRecord,
     };
 
     /// In-memory [`KeyStore`] for testing the migrate decision without the OS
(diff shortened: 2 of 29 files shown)


desktop/src-tauri/src/commands/channels.rs
```diff
@@ -785,9 +785,20 @@ pub async fn add_channel_members(
...
 ) -> Result<serde_json::Value, String> {
     let uuid = parse_channel_uuid(&channel_id)?;
+    let relay_url = crate::relay::relay_ws_url_with_override(&state);
+    let local_agents = crate::managed_agents::load_managed_agents(&app)?;
+    for pubkey in &pubkeys {
+        if let Some(record) = local_agents.iter().find(|record| {
+            record.pubkey.eq_ignore_ascii_case(pubkey)
+                && record.backend == crate::managed_agents::BackendKind::Local
+        }) {
+            crate::managed_agents::validate_local_agent_relay(&record.backend, &relay_url)?;
```
brow: 🤖 This only guards `add_channel_members`, but huddle enrollment emits membership events directly. `start_huddle` calls `events::build_add_member` for every invited pubkey, and `add_agent_to_huddle` does the same for both ephemeral and parent channels. The latter is reachable from the Add Agent dialog, which lists running managed agents. On a disallowed workspace relay, either path can therefore attach a local managed agent without reaching this validation. Please put the same effective-relay policy below every huddle membership path and add direct coverage for both initial and incremental enrollment.

desktop/src-tauri/src/managed_agents/storage.rs
```diff
@@ -199,6 +199,9 @@ fn load_agent_store(app: &AppHandle) -> Result<Vec<ManagedAgentRecord>, String>
 pub fn load_managed_agents(app: &AppHandle) -> Result<Vec<ManagedAgentRecord>, String> {
     let mut records = load_agent_store(app)?;
     records.retain(|record| !record.pubkey.is_empty());
+    records
+        .iter()
+        .try_for_each(super::validate_managed_agent_relay_pin)?;
```
brow: 🤖 Rejecting a legacy disallowed pin while loading makes one stale record brick the entire managed-agent store. `load_managed_agents` is also required by listing, update/delete remediation, snapshots/import, membership, restore, and shutdown cleanup, so the user cannot clear or remove that record and unrelated agents cannot be managed. Since spawning already validates the effective workspace relay and legacy record pins are not authoritative at runtime, please keep reads/remediation available by clearing or quarantining a bad legacy pin (or otherwise returning the store safely), while retaining fail-closed enforcement at save/start/spawn boundaries. Add a regression test showing a bad legacy pin remains removable and does not block shutdown or unrelated records.

desktop/src-tauri/src/commands/channels.rs
brow: 🤖 Fixed in [1a328cf](https://github.com/block/buzz/pull/2582/changes/1a328cfa8bb3a4c30d6092e13834342099d65d28): both initial and incremental huddle enrollment now preflight matching local agents against the effective workspace relay before any huddle state mutation or membership event, with direct member-policy regression coverage.

desktop/src-tauri/src/managed_agents/storage.rs
brow: 🤖 Fixed in [1a328cf](https://github.com/block/buzz/pull/2582/changes/1a328cfa8bb3a4c30d6092e13834342099d65d28): stale disallowed pins remain loadable and unchanged pins permit remediation/unrelated saves, while new or changed relay/backend pins are revalidated and runtime start/attach guards remain enforced, with direct regressions.

desktop/src-tauri/src/managed_agents/storage.rs
```diff
@@ -292,13 +298,46 @@ fn hydrate_keys_with(store: &impl KeyStore, records: &mut [ManagedAgentRecord])
...
+}
+
 /// Save the keyed agent *instances*, preserving the key-less definitions that
 /// share the unified store: callers pass exactly the records they loaded via
 /// [`load_managed_agents`], and this re-reads the definition half from disk
 /// before the wholesale rewrite so a definition is never dropped by an
 /// instance-side save (and vice versa via [`save_agent_definitions`]).
 pub fn save_managed_agents(app: &AppHandle, records: &[ManagedAgentRecord]) -> Result<(), String> {
-    let definitions = load_agent_definitions(app).unwrap_or_default();
+    let stored = load_agent_store(app).unwrap_or_default();
```
brow: 🤖 This turns a parse/read failure into an empty previous store, then proceeds to rewrite `managed-agents.json`. That can silently drop all key-less definitions and bypass the documented `.invalid` fail-loud preservation contract. It also changes policy behavior: an unchanged legacy bad pin is treated as new because `previous_instances` is empty, so remediation-safe saves fail again. Please propagate `load_agent_store(app)?` here and keep only policy violations non-fatal at load time.

desktop/src-tauri/src/commands/channels.rs
```diff
@@ -785,9 +785,13 @@ pub async fn add_channel_members(
     channel_id: String,
     pubkeys: Vec<String>,
     role: Option<String>,
+    app: AppHandle,
     state: State<'_, AppState>,
 ) -> Result<serde_json::Value, String> {
     let uuid = parse_channel_uuid(&channel_id)?;
+    let relay_url = crate::relay::relay_ws_url_with_override(&state);
+    let local_agents = crate::managed_agents::load_managed_agents(&app)?;
```
brow: 🤖 This now loads the managed-agent store before every membership addition in OSS builds too, even though the relay policy is intentionally a no-op there. A malformed or unreadable agent store therefore prevents adding any human or unknown pubkey, which is a new unrelated OSS regression. The huddle paths have the same coupling. Please bypass the store lookup when `!internal_build()` (or otherwise preserve the pre-change OSS behavior) while keeping internal builds fail closed.

desktop/src/features/agents/hooks.ts
```diff
@@ -910,10 +911,26 @@ export function useRuntimeFileConfigQuery(
...
+] as const;
 export const agentAccessOwnerOnlyQueryKey = [
   "agent-access-owner-only",
 ] as const;
 
+export function useLocalAgentRelayAllowedQuery(options?: {
+  enabled?: boolean;
+}) {
+  return useQuery({
+    queryKey: localAgentRelayAllowedQueryKey,
```
brow: 🤖 This verdict depends on the active workspace relay, but the static query key survives community switches in the app-level `QueryClientProvider` and `resetCommunityState()` does not invalidate it. With the 30-second stale window, switching allowed → blocked can leave this dialog enabled until backend submission fails, and blocked → allowed can leave it disabled. Please scope the key to the active community/relay or explicitly invalidate it on community change, with a switch regression test.

desktop/src-tauri/src/managed_agents/storage.rs
brow: 🤖 Fixed in [62c9f36](https://github.com/block/buzz/pull/2582/changes/62c9f36578f67ed5d8dab00a944c829385565524): `save_managed_agents` now propagates the raw-store read/parse error and never rewrites from an empty fallback; regression tests cover fail-loud baseline loading and preservation of definitions alongside instances.

desktop/src-tauri/src/commands/channels.rs
brow: 🤖 Fixed in [62c9f36](https://github.com/block/buzz/pull/2582/changes/62c9f36578f67ed5d8dab00a944c829385565524): membership and huddle guards now lazy-load the agent store only in internal builds, so OSS human/unknown enrollment is independent of store health while internal enforcement still fails closed, with direct tests for both branches.

desktop/src/features/agents/hooks.ts
brow: 🤖 Fixed in [62c9f36](https://github.com/block/buzz/pull/2582/changes/62c9f36578f67ed5d8dab00a944c829385565524): the local-agent relay-policy query key now includes the active community ID, with a regression test proving communities cannot share a cached verdict.

brow (commented):
🤖 Changes are still required because agent-keyed relay I/O remains unguarded outside the newly patched create/update paths:

- `confirm_agent_snapshot_import` mints and persists a `BackendKind::Local` record with an empty pin, then sends its profile and any memory events to the effective workspace relay. `confirm_team_snapshot_import` does the same for every member. On an internal build using a public relay, these commands create local identities and attach them without policy validation. Validate before key mint/store mutation and before profile or engram I/O, preserving OSS behavior, with individual and team import regressions.
- `reconcile_agent_profile` resolves a local record's effective relay, then queries and may publish with the agent key without policy validation. This background path runs after start and restore. Carry the backend into `ProfileReconcileData` and validate before either query or write, with legacy empty-pin coverage.
- `update_persona` publishes linked local-agent profile changes to each effective relay without policy validation. Validate each affected record before mutation/save or enqueueing profile sync, while keeping provider and OSS records unrestricted.

These paths are outside the PR diff, so GitHub cannot anchor inline comments to them. They are at `desktop/src-tauri/src/commands/personas/snapshot/import.rs:508-570`, `desktop/src-tauri/src/commands/team_snapshot.rs:756-826`, `desktop/src-tauri/src/commands/agents_profile.rs:79-160`, and `desktop/src-tauri/src/commands/personas/mod.rs:257-305` at head `93ea37c2`.

brow (commented):
🤖 Re-reviewed the full diff at `adbc1112dae464f3ba687ea524b242cbebf4ce48`. No blocking findings. The round-5 changes close snapshot import, profile reconciliation, and linked-persona profile-sync relay paths before key mint, persistence, query, or publish while preserving OSS/provider short-circuits. Prior membership, storage-remediation, create/update, start/probe/spawn, UI cache, and normalization fixes remain intact.

wpfleger96 (changes requested):
🤖 The security design here is strong and I'd be happy to see it land — requesting changes only for two mechanical items below.

What I like, having walked all the enforcement seams: the allowlist is baked at build time with a base64 round-trip so cargo's line-oriented output can't truncate it (consistent with the existing `BUZZ_BUILD_AGENT_ENV` pattern); internal builds **fail closed** on a missing, empty, or malformed allowlist; matching is exact canonical origin via `normalize_relay_url` (case, default ports, and paths all normalize — verified against the tests) with no wildcard or hostname-suffix trust; and OSS builds plus provider-backed agents are cleanly out of scope. Enforcement covers every path I could find that puts a local agent on a relay: create, update (pin change and rename), the final `spawn_agent_child` chokepoint (which catches anything upstream missed), runtime pairs, the relay probe, profile reconcile, save-time validation of *changed* pins only (so a legacy bad pin stays listable/fixable/deletable — good remediation story), channel member add and role change, huddle start/add, and both snapshot import flows, plus the UI preflight that disables submit with a banner. The tests are discriminating — call-counters proving OSS never touches the agent store, fail-loud on a broken store, and preflight-before-mutation ordering. CI is fully green at head.

**Blocking:**

- **The PR is currently conflicting with its base branch.** `quietly/internal-owner-only` merged main forward after this branch diverged, and four files conflict (`desktop/scripts/check-file-sizes.mjs`, `desktop/src-tauri/build.rs`, `desktop/src/testing/e2eBridge.ts`, `desktop/tests/helpers/bridge.ts`). They all look like the usual high-churn seams, so the resolution should be mechanical — but it needs a merge-forward/rebase before this can land.
- **The title needs a conventional-commit prefix** — we squash-merge, so the PR title becomes the commit subject. Something like `feat(desktop): lock internal local agents to approved relays`.

**Minor, non-blocking:**

- `validated_snapshot_import_relay_with` (personas/snapshot/import.rs) and `validated_team_snapshot_import_relay_with` (team_snapshot.rs) are identical helpers in two files — could be one shared function.
- `local_agent_relay_allowed` can never return `Ok(false)` — a blocked relay is an `Err`, and the frontend only checks `isError` and shows a hardcoded banner. Returning `Ok(false)` for the policy-blocked case (or surfacing the error text in the banner) would make the command's contract less surprising to the next caller.

Happy to approve once the base conflict is resolved and the title is fixed.

dophsquare:
Nudge on behalf of @doph — high-priority merge list. Two things blocking:

1. **Stacked + conflicting** — this PR targets base `quietly/internal-owner-only` (PR #2537), not `main`, and currently shows `CONFLICTING/DIRTY`. It can't cleanly rebase until its parent (#2537) lands; once #2537 merges, re-target/rebase this onto `main`.
2. **Conventional-commit prefix** — the title `Lock internal local agents to approved relays` needs a CC prefix (e.g. `fix(...)` or `feat(...)`) to land cleanly.
3. Plus an outstanding `CHANGES_REQUESTED` review.

@brow flagging so the stack + prefix can be sorted. 🐝

PR #2584, opened 2026-07-23
fix(desktop): center date divider rule

kursmark-sq:
## Summary

- Gives the date pill a tokenized `h-7` height and vertically centers its label.
- Centers the `h-px` separator with Tailwind spacing and transform utilities (`top-3.5` and `-translate-y-1/2`) instead of an arbitrary CSS calculation.
- Adds an E2E geometry assertion that accounts for the computed CSS translate while preserving the existing sticky-pill/scrolling-rule behavior.

## Root cause

The separator previously used `top-4`, which placed the rule below the rendered midpoint of the date pill. The initial exact-centering fix relied on an arbitrary `calc(...)` value; the final implementation expresses the same alignment with Tailwind utilities.

## User impact

Date-divider pills now sit vertically centered on their separator lines in message timelines, using design-system spacing tokens.

## Test plan

- `bin/pnpm --dir desktop check`
- `bin/pnpm --dir desktop test` (3,885 passed)
- `bin/pnpm --dir desktop typecheck`
- `bin/pnpm --dir desktop build:e2e`
- `bin/pnpm --dir desktop exec playwright test tests/e2e/channels.spec.ts --grep "channel date divider"`
- Full pre-push hooks passed: desktop checks/tests, mobile tests, Rust tests, and Desktop Tauri checks.


kursmark-sq:
🤖 ## Before / after

| Before | After |
| --- | --- |
| Date divider before | Date divider after |

PR #2593, opened 2026-07-23
fix(desktop): flatten private/DM timelines without stripping reply tags

StephGlansberg:
## What Problem This Solves
Private rooms and DMs (including AEON Aspect offices like `#aspect-nexus`) show reply-tagged messages as “1 reply” thread summaries instead of a normal flat chat feed, even when those replies are the primary conversation.

## Why This Change Was Made
Presentation-only flatten for `visibility === "private"` and DM channels: include reply-tagged events on the main timeline with no summary badge, while keeping NIP-10 reply tags on the wire so ACP turn receipts still correlate.

## User Impact
Private/DM timelines read as ordinary chat. Public/broadcast channels are unchanged. Receipt/ACP behavior is unchanged.

## Evidence
- Unit tests: `flattenChannelTimeline` + `threadPanel` flatten cases (27 passing in prior verification).
- Pre-push hooks green on this branch (desktop-check/test, rust-tests, mobile-test, desktop-tauri-test).


StephGlansberg:
@codex review

StephGlansberg:
@codex review

StephGlansberg:
@codex review

Issue #2369, opened 2026-07-22
test(buzz-db): replica_fence unit test fails on Linux due to sub-microsecond Utc::now() precision


leesalminen:
## Summary

`replica_fence::tests::fence_starts_closed_and_opens_on_advance` fails deterministically on Linux (glibc), on a clean `main` checkout with no local changes:

```
assertion `left == right` failed
  left: Some(2026-07-22T14:04:14.529519Z)
 right: Some(2026-07-22T14:04:14.529519155Z)
```

## Cause

The fence stores/returns the verified-through timestamp at **microsecond** precision, but the test compares it against the original `Utc::now()` value. On Linux, `Utc::now()` carries **nanosecond** precision, so `fence.verified_through() == Some(ts)` fails whenever `now()` lands on a non-zero sub-microsecond component (i.e. almost always). On macOS `now()` is microsecond-precision, which is why this passes in local dev and (presumably) CI runners there.

## Repro

```bash
# Linux, clean main (7e34bee)
cargo test -p buzz-db --lib fence_starts_closed
# fails 3/3 runs
```

## Suggested fix

Either truncate the expectation in the test (`ts.trunc_subsecs(6)` / compare through the same storage round-trip), or store nanoseconds in the fence. Test-side truncation is the minimal change if microsecond storage is intentional (matches Postgres `timestamptz` precision).

dalton-turner:
https://github.com/block/buzz/pull/2384

PR #2599, opened 2026-07-23
feat(relay): make per-owner community limit configurable via BUZZ_MAX_COMMUNITIES_PER_OWNER

### LINKED ISSUE #2600: Per-owner community limit is hardcoded (MAX_COMMUNITIES_PER_OWNER = 3), blocking multi-tenant operators
## Problem

`MAX_COMMUNITIES_PER_OWNER` is a hardcoded const (`crates/buzz-db/src/relay_members.rs`), enforced in community provisioning (`create_community_with_owner`) and ownership transfer. On a self-hosted multi-tenant deployment (one relay serving many communities via host-based tenancy), the operator's identity hits the cap of 3 almost immediately.

The failure mode is also confusing: `POST /operator/communities` returns a 409 whose body is `limit_reached: owner already owns the maximum number of communities`, but since `community already exists` uses the same status code, provisioning frontends tend to collapse both into "that subdomain is already taken" — which sends the operator chasing a name conflict that doesn't exist (mine did exactly that).

## Reproduction

1. Provision 3 communities owned by the same pubkey via `POST /operator/communities`.
2. Provision a 4th with a fresh, unused host.
3. `GET /operator/communities/availability` reports the host as available, but the create call 409s with `limit_reached`.

## Proposal

Make the limit deployment-configurable via an env var (`BUZZ_MAX_COMMUNITIES_PER_OWNER`), defaulting to the current 3 so existing deployments are unaffected. Enforcement should stay in the authoritative relay-layer checks — this is purely about making the ceiling operator-tunable.

PR: #2599

leesalminen:
Closes #2600

## Summary

Self-hosted multi-tenant deployments (one relay serving many communities via host-based tenancy) routinely need more than three communities owned by the same operator identity. `MAX_COMMUNITIES_PER_OWNER` is currently a hardcoded const, and hitting it surfaces as a `limit_reached` 409 from `POST /operator/communities` — which provisioning UIs tend to mislabel (mine reported it as "subdomain already taken").

This makes the limit configurable per deployment:

- New env var `BUZZ_MAX_COMMUNITIES_PER_OWNER` — read once per process, must parse as a positive integer; missing/invalid/non-positive values fall back to the existing default of **3**, so current deployments are unaffected.
- Enforcement locations are unchanged and stay in the authoritative relay-layer checks: community provisioning (`create_community_with_owner`) and ownership transfer (inside the advisory-lock transaction).
- Parse/fallback rules are extracted into a pure helper (`effective_owner_limit`) with unit tests, keeping the cached getter trivial.

## Test plan

- `cargo test -p buzz-db --lib` — new `owner_limit_*` tests cover default, invalid, non-positive, and positive-override cases. (Pre-existing unrelated failure on clean main: `replica_fence::tests::fence_starts_closed_and_opens_on_advance`, tracked in #2369.)
- `cargo clippy -p buzz-db --all-targets` and `cargo fmt` clean.
- Deployed on my multi-tenant relay (vibecode.casa) with `BUZZ_MAX_COMMUNITIES_PER_OWNER=100`: provisioning a 4th community for the same owner succeeds; without the var the stock limit of 3 still applies.

### DIFF
--- a/crates/buzz-db/src/lib.rs
+++ b/crates/buzz-db/src/lib.rs
@@ -898,7 +898,7 @@ impl Db {
             .fetch_one(&mut *tx)
             .await?;
 
-            if owned_count >= relay_members::MAX_COMMUNITIES_PER_OWNER {
+            if owned_count >= relay_members::max_communities_per_owner() {
                 tx.rollback().await?;
                 return Ok(CreateCommunityWithOwnerResult::LimitReached);
             }
--- a/crates/buzz-db/src/relay_members.rs
+++ b/crates/buzz-db/src/relay_members.rs
@@ -373,10 +373,36 @@ pub enum TransferResult {
     LimitReached,
 }
 
-/// Maximum number of communities a single pubkey can own. Enforced at the
-/// relay layer — the authoritative layer — so that concurrent transfers or
+/// Default maximum number of communities a single pubkey can own. Enforced at
+/// the relay layer — the authoritative layer — so that concurrent transfers or
 /// transfer-vs-create races cannot both pass a preflight count.
 pub const MAX_COMMUNITIES_PER_OWNER: i64 = 3;
+
+/// Effective per-owner community limit for this deployment.
+///
+/// Reads `BUZZ_MAX_COMMUNITIES_PER_OWNER` once (cached for the process
+/// lifetime); a missing, unparsable, or non-positive value falls back to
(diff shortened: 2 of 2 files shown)


tlongwell-block (approved):

tlongwell-block:
Thanks for the contribution, @leesalminen 👋

PR #2447, opened 2026-07-23
feat: adjust order of agent status and text box

alexhancock:
I noticed a UI issue where the text entry box doesn't have consistent padding vertically and horizontally because the agent status bar (which is only active sometimes) sits below it. This makes the text box further away from the bottom than it is from the sides when the agent isn't doing anything and looked a bit off to me.

This change flips it so the agent status bar is _above_ the text entry box, and the text entry box can sit in a more natural spot. It also subtly moves the messages up when the agent status bar is there so that they don't run under the status bar.

**Before**

Screenshot 2026-07-22 at 8 43 25 PM

**After**

Screenshot 2026-07-22 at 8 49 38 PM

**Agent Active State**

Screenshot 2026-07-22 at 8 49 48 PM

PR #2602, opened 2026-07-23
fix: push text box down closer to the bottom

alexhancock:
The text entry box currently sits a bit far away from the bottom. This makes the agent status reporting smaller so it can sit closer to the bottom and almost match the horizontal padding.

Replaces https://github.com/block/buzz/pull/2447

**Before**

Screenshot 2026-07-23 at 3 55 31 PM

**After**

Screenshot 2026-07-23 at 3 54 30 PM


alexhancock:
This lead to discussion/experimentation with @morgmart and she has https://github.com/block/buzz/pull/3151 up now. Closing this out

PR #2149, opened 2026-07-19
fix(desktop): prefer live agent mentions

baxen:
## Why
Agent mention typeahead could surface old p-tag/channel-member identities and other users' concrete agents even when those identities were absent from the current user's managed Agents list. This produced multiple stale results for a persona such as Bumble.

Add Agents is related but not identical: it lists active persona templates, then creates or reuses a concrete managed identity after selection. Mention typeahead must also retain people and persona/team additions, so only concrete candidates identified as agents are restricted.

## What
- Treat the current user's managed Agents list as the authority for concrete agent identities in mention typeahead
- Keep managed agents regardless of running, stopped, deployed, or not-deployed state
- Exclude old p-tag/channel-member agents and other users' agent identities when their pubkeys are not in that list
- Leave people and non-concrete persona/team additions unchanged
- Preserve the 50-result cap and add focused regression coverage

## Risk Assessment
Medium-low — scoped to desktop mention autocomplete candidate selection. The filter applies only to concrete identity candidates marked as agents; people, persona templates, and teams retain their existing behavior.

## References
- Focused autocomplete test: 18 passed
- Desktop TypeScript typecheck passed
- Desktop test suite: 3,188 passed
- Desktop lint/file-size/policy checks passed
- Full pre-push rerun reached unrelated Codex adapter Rust test failures (1,482 passed, 5 failed); the branch was pushed with the already-completed desktop validation

Generated with Fizz

PR #2605, opened 2026-07-23
fix(desktop): shared relay agents are un-mentionable/un-addable — let directory eligibility run

### LINKED ISSUE #2603: Desktop: shared relay agents owned by other members are un-mentionable and un-addable (managed-list gate precedes directory eligibility)
## Problem

The desktop drops any agent-flagged identity that is not in the **local** managed-agents list before the relay-directory eligibility logic ever runs:

- `useMentions.ts` calls `isAgentIdentityInManagedList(candidate, managedAgentPubkeys)` as a pre-emptive gate in `addCandidate`, ahead of `shouldHideAgentFromMentions`. The directory-based path (`mentionableAgentPubkeys` built from kind:10100 `RelayAgent`s with `respond_to: anyone`/allowlist + shared channels) is therefore unreachable for agents the viewer doesn't locally manage — including the "Invocable => always show" and member/Option-B rules that `shouldHideAgentFromMentions` explicitly implements.
- `MembersSidebar.tsx` applies the same managed-list-only gate to add-member search candidates, so a shared agent can't be added to a channel by anyone but the desktop that owns it.

Net effect: a relay-hosted shared agent (kind:0 with NIP-OA auth tag + kind:10100 profile, `respond_to: anyone`, member of the channel) responds fine when p-tagged, but no other member can @-complete it from the composer or add it to channels — it's structurally invisible as an agent to everyone except its owning device. (Found running a headless shared agent on a self-hosted multi-tenant relay; the only workaround is stripping the agent's auth tag/10100 so it masquerades as a human member.)

## Expected

`shouldHideAgentFromMentions` is already the correct, complete policy (invocable → show; non-member non-invocable → hide; member + directory-excluded → hide; member + unknown → show). It should be the single authority in the composer, and the add-member gate should accept relay-directory agents that would respond to the current user.

leesalminen:
Closes #2603

## Summary

A shared relay agent (kind:0 + NIP-OA auth tag, kind:10100 directory profile with `respond_to: anyone`, channel member) responds fine when p-tagged, but nobody except its owning device can @-complete it or add it to channels: `addCandidate` in `useMentions.ts` drops every agent-flagged identity that isn't in the **local** managed-agents list *before* `shouldHideAgentFromMentions` runs, making the directory-eligibility path (`mentionableAgentPubkeys`, and the "Invocable => always show" / member Option-B rules) unreachable for foreign agents. `MembersSidebar` applies the same managed-list-only gate to add-member search.

## Changes

- **`useMentions.ts`**: remove the pre-emptive `isAgentIdentityInManagedList` gate. `shouldHideAgentFromMentions` is the single authority — it already shows invocable agents (local managed ∪ relay-directory-eligible), hides non-member non-invocable ones, and applies the member/directory rules. Local managed agents are unaffected (they're seeded into `mentionableAgentPubkeys`), and other members' `owner-only` agents stay hidden (non-member + non-invocable, or member + directory-excluded).
- **`MembersSidebar.tsx`**: add-member candidates now pass when the agent is locally managed **or** relay-directory invocable by the current user (new `relayAgentIsInvocableByUser`: `respond_to: anyone`, or an allowlist containing them). Channel overlap is deliberately not required here — adding the agent to the channel is what creates the overlap.
- **`agentAutocompleteEligibility.ts`**: new `relayAgentIsInvocableByUser` helper; `relayAgentIsSharedWithUser` delegates its allowlist branch to it (behavior unchanged); doc comment on `isAgentIdentityInManagedList` clarifying the caller-supplied allow-set contract.

## Test plan

- `node --import ./test-loader.mjs --experimental-strip-types --test` on `agentAutocompleteEligibility.test.mjs`: 22/22 (4 new: invocable-by-user modes, and a regression test that a directory-eligible member agent is shown without a local managed entry).
- `biome check` and `tsc --noEmit` clean on touched files.
- Manually verified against a self-hosted multi-tenant relay running a headless shared agent (buzz-acp + kind:10100, `respond_to: anyone`): with this fix the agent appears in mention autocomplete and add-member search on a desktop that does not manage it.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -6,6 +6,7 @@ import {
   getMentionableAgentPubkeys,
   getSharedChannelIds,
   isAgentIdentityInManagedList,
+  relayAgentIsInvocableByUser,
   relayAgentIsSharedWithUser,
   shouldHideAgentFromMentions,
 } from "./agentAutocompleteEligibility.ts";
@@ -305,4 +306,65 @@ test("coalesceAgentAutocompleteCandidates: leaves non-agents alone", () => {
   const second = makeAgent({ pubkey: PUB_B, isAgent: false });
 
   assert.deepEqual(coalesce([first, second]), [first, second]);
+});
+
+test("relayAgentIsInvocableByUser: anyone responds regardless of channel overlap", () => {
+  assert.equal(
+    relayAgentIsInvocableByUser(
+      { respondTo: "anyone", respondToAllowlist: [] },
+      CURRENT_PUBKEY,
+    ),
+    true,
+  );
+});
+
+test("relayAgentIsInvocableByUser: allowlist honors membership", () => {
+  const agent = { respondTo: "allowlist", respondToAllowlist: [CURRENT_PUBKEY] };
+  assert.equal(relayAgentIsInvocableByUser(agent, CURRENT_PUBKEY), true);
+  assert.equal(relayAgentIsInvocableByUser(agent, OTHER_OWNER_PUBKEY), false);
+});
+
+test("relayAgentIsInvocableByUser: owner-only and unknown modes are not invocable", () => {
+  assert.equal(
+    relayAgentIsInvocableByUser(
+      { respondTo: "owner-only", respondToAllowlist: [] },
+      CURRENT_PUBKEY,
+    ),
+    false,
+  );
+  assert.equal(
+    relayAgentIsInvocableByUser({ respondTo: null, respondToAllowlist: [] }, CURRENT_PUBKEY),
+    false,
+  );
+});
+
+test("shouldHideAgentFromMentions: relay-shared agent shows without local managed entry", () => {
+  // Regression: a shared agent owned by another member (directory-eligible,
+  // respond_to=anyone, shared channel) must be mentionable even though it is
+  // not in this device's managed-agents list.
+  const sharedChannelIds = new Set(["chan-1"]);
+  const relayAgents = [
+    {
+      pubkey: PUB_D,
+      respondTo: "anyone",
+      respondToAllowlist: [],
+      channelIds: ["chan-1"],
+    },
+  ];
+  const mentionable = getMentionableAgentPubkeys({
+    currentPubkey: CURRENT_PUBKEY,
+    managedAgentPubkeys: [],
+    relayAgents,
+    sharedChannelIds,
+  });
+  assert.equal(
+    shouldHideAgentFromMentions({
+      isAgent: true,
+      isMember: true,
+      pubkey: PUB_D,
+      mentionableAgentPubkeys: mentionable,
+      directoryAgentPubkeys: new Set([PUB_D]),
+    }),
+    false,
+  );
 });
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
@@ -9,9 +9,12 @@ export function getSharedChannelIds(channels: readonly Channel[] | undefined) {
   );
 }
 
-export function relayAgentIsSharedWithUser(
-  agent: Pick<RelayAgent, "channelIds" | "respondTo" | "respondToAllowlist">,
-  sharedChannelIds: ReadonlySet<string>,
+/// Whether the relay directory says this agent would respond to the current
+/// user at all, ignoring channel overlap. Used where the user is about to
+/// CREATE the overlap (e.g. adding the agent to a channel), so requiring an
+/// existing shared channel would be circular.
+export function relayAgentIsInvocableByUser(
+  agent: Pick<RelayAgent, "respondTo" | "respondToAllowlist">,
   currentPubkey?: string | null,
 ) {
   const normalizedCurrentPubkey = currentPubkey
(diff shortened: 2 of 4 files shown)


jonathanleiva15:
Confirming the root cause from another angle in case it's useful for the review.

**Repro (mobile↔desktop parity):** an externally-attested agent — one that joined a community via a NIP-OA owner `auth` tag (`ViaOwner` relay membership) and is a channel member — is mentionable in the **mobile** client but not in **desktop**, on the same account/relay. Mobile's `buildMentionCandidates` (`mobile/lib/features/channels/mentions/mention_candidates.dart`) adds all channel members unconditionally and only applies the invocability filter to *non-members*, so membership is sufficient there. On desktop, `isAgentIdentityInManagedList` (added in #2149) runs inside `addCandidate` *before* `shouldHideAgentFromMentions` and drops any `isAgent && !managedAgentPubkeys.has(pk)`, so a member agent I don't locally manage never reaches the directory gate. Removing that pre-gate (this PR) restores the mobile semantics 👍

One sanity check for the review: #2149's stated intent was narrower than "managed-only" — de-duping *owner-less* stale p-tag copies of a persona I already manage (e.g. Bumble). Removing the gate entirely also lets owner-less **member** duplicates back in. If preserving that dedupe matters, a more surgical variant exempts only *member + attested* agents:

```ts
if (candidate.isAgent !== true) return true;
if (managedAgentPubkeys.has(normalizePubkey(candidate.pubkey))) return true;
return candidate.isMember === true && Boolean(candidate.ownerPubkey);
```

`ownerPubkey` is only set for a verified auth tag (`profile_valid_oa_owner_pubkey` in `nostr_convert.rs`), so it discriminates a legit external agent from a stale owner-less copy — and `shouldHideAgentFromMentions` still hides the "Fizz" (directory-excluded) case either way. Both approaches fix the reported symptom; just flagging the trade-off in case the owner-less dedupe was intentional. Happy to send a follow-up with tests if the surgical variant is preferred.

Bartok9 (commented):
## Collaborative review (Bartok9)

Thanks @leesalminen — this is the cleanest desktop-only fix of the #2987 swarm I’ve looked at.

### What’s strong
1. **Right authority**: dropping the managed-list pre-gate in `useMentions` so `shouldHideAgentFromMentions` actually runs for external agents matches the diagnostic (#2987 report point 1).
2. **Add-member circularity fixed deliberately**: `relayAgentIsInvocableByUser` (anyone/allowlist, **no** channel-overlap requirement) is the right split from `relayAgentIsSharedWithUser` for “invite path.”
3. **Tests of the regression**: `shouldHideAgentFromMentions: relay-shared agent shows without local managed entry` + invocable helper cases — this is what reviewers need when five PRs claim the same one-line gate.
4. **Doc comment on `isAgentIdentityInManagedList`**: renaming the mental model to *caller-supplied allow-set* is worth more than another prose PR body.

### Suggestions (non-blocking unless maintainers want polish first)
1. **Owner self-mention edge** (#3448’s `isOwnAgent` / `respond_to: owner-only`): with only `relayAgentIsInvocableByUser`, owner-only self agents on a remote host may still not mention/add themselves. Worth either:
   - a short “Out of scope / parented in #2987 footing / follow-up” note, or
   - a small additive own-pubkey escape aligned with 3448 if maintainers want one carrier for both.
2. **Name drift**: helper is `isAgentIdentityInManagedList` but set is now “addable/mentionable.” A rename is optional and debatable (API churn); a one-line doc is already good.
3. **Call-site parity**: only sidebar uses the expanded allow-set assembly; mentions relies on hide helper. That’s fine if `getMentionableAgentPubkeys` already unions managed ∪ invocable relay — a one-sentence call-path claim in the PR body would help maintainers grepping later (Brad-class: prove path, don’t just assert).

### Swarm coordination (not a code nit)
Several open PRs touch the same gate (`#3110`, `#3292`, `#2752`, `#3108`, `#3448`, …). Prefer **this + or #3448** as the merge vehicle so the thrash collapses. Happy to re-test a tip commit after rebase onto current `main` if useful.

**Verdict:** Approve-shape from a third-party collab angle — small, tested, correct layer. Not a maintainer formal approve.

Cyberworkshop:
Heads up for whoever picks this as the merge vehicle: fixing the picker-eligibility gate alone won't fully close out #2950 for existing agents. Found and verified a second, separate bug while reproducing this end-to-end on my own relay — Edit Agent's "Who can talk to this agent" toggle doesn't reliably persist to the on-disk `managed-agents.json` `respond_to` field the harness actually reads (stuck at `owner-only` despite the UI showing "Anyone," even across an explicit restart). Full repro with timestamped before/after is on [#2950's latest comment](https://github.com/block/buzz/issues/2950#issuecomment-5141665839) — once this merges, worth checking that agents whose access was changed post-creation actually respond, not just that they're visible in the picker.

bharani-gif (approved):

Bartok9:
Thanks @bharani-gif for the approve — appreciated.

@Cyberworkshop good catch on the separate track. Agree this PR should stay scoped to **picker/directory eligibility** (managed-list pre-gate + invocable add-member path). The Edit Agent “Who can talk to this agent” → on-disk `managed-agents.json` `respond_to` stuck at `owner-only` while the UI shows “Anyone” is a **distinct persistence/harness bug** and won’t be closed by merging #2605 alone.

For whoever drives merge of this vehicle:
1. Land #2605 (or the chosen #2987-family carrier) for mention/add visibility.
2. Treat the #2950 repro you linked ([issue comment](https://github.com/block/buzz/issues/2950#issuecomment-5141665839)) as a **follow-up** — verify post-create access toggles actually rewrite harness `respond_to`, not just the UI label.
3. After merge, smoke both: shared agent appears in autocomplete **and** a toggled-to-Anyone agent actually answers non-owners.

Happy to re-test a tip commit after rebase onto current `main` if useful. Not claiming maintainer merge authority — just collab coordination so #2950 doesn’t get marked done on visibility alone.

PR #2606, opened 2026-07-23
fix(profile): read-merge-write kind:0 so republish stops deleting unmodeled fields

### LINKED ISSUE #2534: kind:0 profile republish drops every field outside a 5-key allowlist — silently deletes bot: true, nip05 and website
Publishing a kind:0 through Buzz replaces the whole event from a 5-key allowlist, so **every field outside that allowlist is deleted** — including `bot`, `nip05`, `website`, `banner` and `lud16`. kind:0 is replaceable, so a partial build is not a partial update; it is a deletion.

The field most likely to disappear is `bot: true`, the NIP-24 flag that discloses an account is automated. It vanishes with no error and no warning, and nothing in the UI or CLI output indicates anything was lost.

## Why this one matters more than a normal field-drop

- **It silently un-discloses a bot.** An agent account that has quietly lost `bot: true` looks like a human account to every client that renders that flag. That is a correctness problem for anyone downstream who filters, labels, or rate-limits automated accounts, and it happens without the operator doing anything wrong.
- **`nip05` loss breaks verification, and the failure mode is worse than absence.** Several clients render an unverifiable or missing NIP-05 as a warning badge rather than as blank.
- **It is silent, so it is not self-correcting.** The operator's next look at the profile shows a name, a picture and a bio — everything they normally check. Nothing signals the loss, so it persists until someone diffs the raw event.
- **The most likely trigger is a no-op.** Setting a field to the value it already had is enough (repro below). So is a rename, an avatar change, or an app restart on the Desktop path.
- **Externally-published profiles are the ones at risk.** Any identity whose complete kind:0 was published by something other than Buzz — a fleet script, `nak`, a custom publisher — has fields Buzz does not model, and Buzz will strip them the first time it touches that profile.

## Repro (CLI, ~10 seconds)

Publish a complete kind:0 for an identity — including `bot: true` — by any means other than Buzz. Then set one field to the value it already has:

```
$ buzz users set-profile --nip05 dennis@nave.pub
```

Read the profile back:

```
  name       *** GONE ***
  website    *** GONE ***
  bot        *** GONE ***
```

`display_name`, `about`, `picture` and `nip05` survive — the allowlist exactly. No error, no warning, exit 0.

## Root cause

`build_profile` constructs the kind:0 content from exactly five optional keys and nothing else:

- `crates/buzz-sdk/src/builders.rs:537`
- `desktop/src-tauri/src/events.rs:466` (duplicated)

```rust
pub fn build_profile(
    display_name: Option<&str>,
    name: Option<&str>,
    picture: Option<&str>,
    about: Option<&str>,
    nip05: Option<&str>,
) -> Result<EventBuilder, SdkError> {
```

There is no passthrough for keys the builder does not know about, and because kind:0 is replaceable, whatever is not built is destroyed.

The relay is not the constraint — it accepts a 7-field kind:0 (`bot`, `website` included) without complaint. This is entirely client-side.

### Three call paths, worst last

**1 · CLI — `cmd_set_profile` (`crates/buzz-cli/src/commands/users.rs:150`).** Read-merge-write, but only over the five keys. Two extra wrinkles:

- `name` is hardcoded to `None` at `users.rs:201` (*"`name` field (username) — not exposed by CLI"*), so `name` is dropped on **every** CLI write, even though `build_profile` accepts it.
- The `merged_name` fallback chain is `display_name` → `current.display_name` → `current.name`, and the result is passed as the *display_name* argument. An identity with a `name` but no `display_name` therefore has its short handle silently promoted into `display_name`.

**2 · Desktop profile command (`desktop/src-tauri/src/commands/profile.rs:76`).** Passes all five, so it is the allowlist and nothing worse.

**3 · Desktop managed-agent sync — the automatic one.** `build_profile_event` (`desktop/src-tauri/src/relay.rs:377`) calls:

```rust
crate::events::build_profile(Some(display_name), None, avatar_url, None, None)?
```

Two fields. `about`, `nip05` and everything unmodelled are destroyed, and this path is not user-invoked. `sync_managed_agent_profile` is reached from at least seven places — agent save (`commands/agents.rs:894`), model change (`commands/agent_models.rs:949`), persona edit (`commands/personas/mod.rs:295`), persona snapshot import (`personas/snapshot/import.rs:509`), team snapshot import (`commands/team_snapshot.rs:761`), profile reconciliation (`commands/agents_profile.rs:153`), and managed-agent restore (`managed_agents/restore.rs:355`).

`reconcile_agent_profile` does query the existing profile first — but `profile_needs_sync` (`commands/agents_profile.rs:167`) compares only `display_name` and `picture`. If either diverges from the local record, it republishes the two-field event. So the read happens and the merge does not.

The practical consequence: an agent whose complete profile was published externally, and whose stored `avatar_url` differs at all from what is on the relay, gets its `about`, `nip05`, `bot` and `website` erased on the next reconcile — with no user action at all. (Derived from the code path above; I have reproduced the CLI path directly and not yet observed the reconcile path in the wild.)

## Suggested fix

1. **Give `build_profile` a passthrough.** Take the current content object and merge the known keys into it, rather than constructing a fresh map. Any field Buzz does not model then survives contact with Buzz — which also future-proofs this against the next NIP that adds a kind:0 key.
2. **Model `bot`, `website`, `banner` and `lud16` explicitly**, and expose `--bot` / `--website` on the CLI. `bot` in particular deserves to be first-class rather than merely preserved.
3. **Expose `--name`** and stop passing `None` at `users.rs:201`; `build_profile` already accepts it. This is the cheapest half of the fix.
4. **Make the Desktop managed-agent sync merge rather than replace.** It already fetches the existing profile in the reconcile path; carrying the untouched keys forward is a small change with the largest blast radius reduction.
5. Deduplicate the two `build_profile` copies, or the fix has to land twice.

Possibly related: #1822 (agent rename — relay profile sync failure silently swallowed). Different failure, same "profile write goes wrong and says nothing" shape.

---

*Filed by My Dude (`npub1p28qwgxra3fvd07euf296csv7k8zarfhzfzz2hhuu0wa27aq5vkq05erd2`), an agent, on behalf of the account owner. The repro was run by another agent on the same nest. Source references are against `origin/main` at `06e3d82`.*

SeanGearin:
Fixes #2534.

## Problem

kind:0 is replaceable — relays keep only the newest event per author, so the
content object Buzz publishes *is* the entire profile. But every Buzz publish
path rebuilt that object from a 5-key allowlist (`display_name`, `name`,
`picture`, `about`, `nip05`). Anything outside the allowlist — `bot` (NIP-24),
`nip05`-adjacent verification state, `website`, `banner`, `lud16`, or any key a
future NIP adds — was silently deleted the first time Buzz touched a profile
that had been published completely by anything else (`nak`, a fleet script, a
custom publisher).

The worst path was the Desktop managed-agent sync: it published a **two**-field
profile (`display_name`, `picture`) and is reached automatically from at least
seven places (agent save, model change, persona edit, two snapshot imports,
profile reconciliation, managed-agent restore). An agent whose stored
`avatar_url` diverged at all from the relay got its `about`, `nip05`, `bot` and
`website` erased with no user action — silently un-disclosing a bot account.

The CLI additionally hardcoded `name: None` on every write (deleting the `name`
key even though the builder accepts it) and promoted `name` into `display_name`
via its fallback chain.

## Fix

Read-merge-write, enforced by the builder signature:

- **`buzz-sdk`**: new pure helper `merge_profile_content(current_content, …)`
  parses the currently-published content and overlays only the caller-supplied
  fields; every other key carries forward unchanged in value (non-string and
  nested values included; key order may be normalized on re-serialization). `build_profile` now takes `current_content: Option<&str>`
  as its first parameter and delegates to the helper. The signature change is
  deliberate: it makes it impossible to call the builder without deciding what
  happens to the current profile — the destructive call shape no longer
  compiles, for this and every future call site.
- **`buzz-cli`** `users set-profile`: passes the fetched current profile
  through the merge. This also fixes the two wrinkles called out in #2534:
  `name` is no longer deleted on every write (the merge carries it forward),
  and the `display_name → current.name` promotion chain is gone.
- **Desktop `update_profile`**: passes the prior event's raw content through
  the merge instead of hand-copying the five known keys.
- **Desktop managed-agent sync** (`sync_managed_agent_profile`): now fetches
  the agent's current kind:0 before publishing, authenticated as the agent
  itself (mirroring the auth the write uses), and merges
  `display_name`/`picture` over it. If the read fails, the sync aborts instead
  of publishing a blind two-field profile — a write without the current content
  is a deletion, not a sync. All seven call sites are covered by fixing the one
  function they share; none of their signatures change. The read and the write
  are each preceded by their own `wait_for_rate_limit()`, preserving the
  relay-admission contract that every HTTP send is individually gated.
- **Deduplication** (#2534 suggestion 5): the merge rules now live in exactly
  one place — `buzz_sdk::merge_profile_content`, pure JSON with no nostr types
  — and the desktop's nostr-0.37 `build_profile` delegates to it. The two
  `build_profile` copies can no longer drift, while each crate keeps its own
  `EventBuilder` for its pinned nostr version.

Unparseable or non-object current content is treated as absent (matching how
every existing read path handles it), so a corrupt relay profile can't wedge
profile updates.

`profile_needs_sync` intentionally still compares only `display_name`/`picture`
— it only decides *whether* to write; the write itself is now non-destructive.

## Tests

`buzz-sdk` (new):
- `profile_merge_preserves_unknown_keys` — the #2534 repro: a no-op `nip05`
  update over a 10-field profile; asserts `bot: true`, `website`, `banner`,
  `lud16`, a nested extension object, and all five modeled keys survive, and
  that the field count is unchanged (nothing gained or lost).
- `profile_merge_overwrites_only_supplied_keys` — supplied fields win;
  unsupplied known fields carry forward.
- `profile_merge_treats_invalid_current_as_absent` — garbage/non-object/empty
  current content behaves like a fresh profile, no panic.

Desktop (`relay.rs`, new):
- `profile_event_preserves_unmodelled_fields` — the managed-agent rename/avatar
  sync over a profile carrying `about`, `nip05`, `bot`, `website`; asserts all
  survive and supplied fields overwrite.

Existing profile/auth-tag tests updated for the new parameter; all pass.

Evidence (scoped per build discipline; `just ci` runs the same fmt/clippy/test
gates workspace-wide; run on the branch rebased onto current main):

```
cargo fmt -p buzz-sdk -p buzz-cli -p countdown-bot -- --check    # clean
cargo clippy -p buzz-sdk --all-targets -- -D warnings            # clean
cargo clippy -p buzz-cli --all-targets -- -D warnings            # clean
cargo clippy -p countdown-bot --all-targets -- -D warnings       # clean
cargo test -p buzz-sdk    # 236 passed; 0 failed (incl. 3 new)
cargo test -p buzz-cli    # 250 passed; 0 failed
# desktop (own workspace):
cargo fmt --manifest-path desktop/src-tauri/Cargo.toml -- --check   # clean
cargo clippy --manifest-path desktop/src-tauri/Cargo.toml \
  --all-targets -- -D warnings                                      # clean
cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib
#   1562 passed; 0 failed; 13 ignored (incl. new
#   relay::tests::profile_event_preserves_unmodelled_fields)
```

## Out of scope (follow-ups)

- Modeling `bot`/`website`/`banner`/`lud16` as first-class builder fields and
  exposing `--bot`/`--website`/`--name` on the CLI (#2534 suggestions 2–3).
  Preservation is the data-loss fix; new flags are a feature and deserve their
  own review.
- A way to intentionally *clear* a field. Before this change `None` meant
  "delete by omission" on every unspecified key; now it means "preserve". An
  explicit clear (e.g. `Some("")` semantics or a dedicated flag) is a product
  decision.


dophsquare:

Verified the data-loss path: kind:0 is replaceable, so the content object Buzz publishes *is* the entire profile, but the old code rebuilt it from a 5-key allowlist (`display_name`/`name`/`picture`/`about`/`nip05`). Any field authored elsewhere — `bot` (NIP-24), `website`, `banner`, `lud16`, or any future NIP key — got silently dropped the first time Buzz touched the profile.

`merge_profile_content` in the SDK is the right fix: parse current content into a map, overwrite only the `Some` fields, carry everything else forward, and treat unparseable/non-object content as absent. The CLI `cmd_set_profile` now does a true read-merge-write by passing `current_json` through instead of hand-merging five keys. Cleaner and correct.

Before merge, please confirm the **other publish paths** (desktop republish, any fleet/persona republish) also route through `merge_profile_content` — the bug was "every publish path rebuilt from the allowlist," so fixing only cli+sdk leaves desktop able to clobber. LGTM on the code shown.

SeanGearin:
Thanks for the thorough triage. Confirmed on the other publish paths: the desktop human-profile republish routes through `buzz_sdk_pkg::merge_profile_content` in this PR (events.rs), so it read-merge-writes the same way cli+sdk do — no allowlist rebuild left on the user-profile path. The managed-agent/persona sync (`sync_managed_agent_profile`) publishes Buzz-owned agent metadata rather than user-authored profiles, so the unmodeled-field clobber doesn't apply there; happy to harden it too as a follow-up if you'd like symmetry. Nothing else rebuilds kind:0 from the allowlist.

wpfleger96 (changes requested):
🤖 The design here is right — putting the merge helper in `buzz-sdk` and changing `build_profile` to a 6-param signature so every caller has to decide what happens to the current content is exactly the shape I'd want, and the regression tests are good. But there's one blocking issue:

**Missed call site — this doesn't compile.** `build_deferred_profile_event` at `desktop/src-tauri/src/commands/profile.rs:152` (the deferred-avatar save path) still calls `events::build_profile(display_name, name, Some(avatar_url), about, nip05)` with the old 5-arg shape. CI confirms it: `error[E0061]: this function takes 6 arguments but 5 arguments were supplied` in both the Windows Rust and macOS Desktop Build jobs.

The bigger problem is that this path is itself the data-loss bug the PR fixes — it hand-copies the 4 known fields out of the current profile, so unknown kind:0 keys still get dropped on any avatar-deferred save. The fix should be small (~5 lines): route `prior_event.content` through the merge the same way the other call sites do.

Two smaller notes, non-blocking:

- The merge semantics can never *delete* a key — once a field exists in someone's kind:0 there's no way to clear it through this path. Probably fine for now, but worth a comment or a follow-up issue so it's a known limitation rather than a surprise.
- `sync_managed_agent_profile` now fail-closes when the read errors. The doc comment covers it and I think it's the right call, just flagging that it's a behavior change.

Happy to re-review once the deferred-avatar call site goes through the merge and CI is green.

SeanGearin:
Good catch, and you were right that it was more than a signature slip — the deferred-avatar path was doing exactly the thing this PR set out to stop. It hand-copied `display_name`/`name`/`about`/`nip05` out of the current profile and rebuilt kind:0 from those four keys, so `bot`, `website`, `banner`, `lud16` and anything else were dropped on every deferred save. Fixing the arg count alone would have compiled and still lost data.

It now does what the other call sites do: hands `prior_event.content` to the merge and passes `None` for the fields it does not set, so they carry forward. That also made the `current: &Value` parameter redundant, so it is gone and the caller updated — the caller still parses `current` for the pre-save avatar-changed guard, which is unrelated.

One more that CI would have caught next: `deferred_profile_event_is_strictly_newer_than_prior_head` was still calling the old 3-arg shape, so it was a second stale call site in the same file. Updated, and it now also asserts the published `display_name` survives rather than being re-supplied by the caller.

Added `deferred_profile_save_preserves_unmodeled_profile_fields`: publishes a profile carrying the four modeled fields plus `bot`, `website`, `banner` and `lud16`, defers an avatar save, and asserts every one of them survives alongside the new `picture`. That pins the regression rather than trusting the call site to stay correct.

I also swept every `build_profile` / `merge_profile_content` call site in the repo — `buzz-cli` users, the countdown-bot example, `relay.rs` and `update_profile` all pass the 6-arg shape correctly. `profile.rs` was the only miss.

On your two non-blocking notes:

**Deletion.** Documented on `merge_profile_content` as a known limitation: the merge can add or overwrite a key but never remove one, so once a field exists in an author's kind:0 there is no way to clear it through this path — a caller needing deletion has to build the content object itself. Worth being explicit that this is a deliberate trade: every caller passes `None` for fields it does not model, so treating `None` as "delete" would make each partial update erase everything outside its own parameter list. Happy to file a follow-up issue if you would rather it be tracked than commented.

**`sync_managed_agent_profile` fail-closing.** Intentional, and thanks for flagging it as a behavior change. If the read fails we cannot know what the published content holds, and the alternative is publishing a rebuilt profile from an unknown baseline — which is the data loss, just triggered by a transient error instead of a missed call site. Refusing to publish leaves the existing profile intact until the next sync.

No Rust toolchain on the machine I am working from, so I have not compiled this locally — CI is the check. Ready for another look when you are.

SeanGearin:
Quick note that isn't visible from the PR page: the run on the new head is parked at `action_required` rather than executing, so nothing has actually compiled since the push — the only checks that ran are DCO, Semgrep and zizmor. Whenever you get a moment to approve the workflow run, that'll give the re-review something real to look at.

SeanGearin:
The deferred-avatar call site now goes through the merge, so `build_deferred_profile_event` hands `prior_event.content` to `build_profile` and passes `None` for everything except the picture. That drops the four hand-copied keys entirely, which means the unmodelled-field loss you pointed at is gone from that path too, not just the compile error. There is a regression test alongside it that seeds `bot`, `website`, `banner` and `lud16` and asserts they survive a deferred save.

While confirming that, I found the run was red for a second reason you did not mention and I had not seen: the size gate. `relay.rs` was 1099 checker-lines against its 1077 override, because this PR adds ~85 lines of read-merge-write to the agent profile sync. Flagging it explicitly since it would have burned another approval to discover.

I did not bump the override. `check-file-sizes.mjs` says not to, and the `relay.rs` entry already carried its own note - "Queued to split test helpers to relay/tests.rs" - so I did that instead. The test module moves to `relay_tests.rs` behind `#[cfg(test)] #[path = "relay_tests.rs"] mod tests;`, matching what `event_sync.rs` and `migration.rs` already do. The moved body is byte-identical apart from one level of de-indentation and all 57 test items come with it, so there is no logic change to review - only the file boundary moved.

That puts `relay.rs` at 698, under the default 1000, so the override entry is removed rather than lowered. One fewer file on that list.

Worth knowing regardless of this PR: `events.rs` currently sits at 999 against the default 1000, so the next edit to it trips the same gate.

The CI and Docker runs are still sitting at `action_required` on this head, so none of the above has actually executed yet.

dophsquare:

PR #2468, opened 2026-07-23
feat(agents): add Hermes ACP and external-agent integration

NYTEMODEONLY:
## Summary

- add Hermes Agent to the Desktop runtime catalog and document local/VPS deployment
- prefer the official `hermes acp` entrypoint, retain direct `hermes-acp` fallback, and verify optional ACP dependencies with the matching `--check` command
- recognize exact `python` and `python3` interpreter names for marker-owned managed-process cleanup without admitting versioned prefixes or unrelated Python jobs
- publish complete kind `10100` directory profiles from externally hosted `buzz-acp` agents and refresh them after membership changes
- derive external-agent ownership only from a verified NIP-OA auth tag
- show eligible externally hosted agents under **External agents** without local runtime controls
- use live kind `20001` presence for external cards and ignore policy-only kind `10100` records as directory entries
- let verified owners customize an external agent's Buzz display name and avatar without changing its runtime-owned state
- resolve that owner presentation consistently across cards, profiles, messages, mentions, DMs, and sidebars
- ingest encrypted kind `24200` ACP activity from verified owner-declared relay agents into the same Desktop Activity surface used by managed agents
- document the complete external-agent deployment, presentation, observer, and troubleshooting path

## Root cause

Externally hosted agents could have valid identity, ownership, and channel membership while remaining absent from Desktop mention surfaces. Their replaceable kind `10100` record was incomplete, policy changes could clobber it, and Desktop did not carry verified owner identity into eligibility checks.

After directory discovery was fixed, three additional gaps remained: owner-selected external-agent presentation did not win in profile-backed message/sidebar surfaces, observer ingestion only considered locally managed agents, and the Agents card incorrectly treated persistent directory metadata as live presence. Policy-only kind `10100` records could also appear as raw-key cards even though they were not complete agent declarations.

The new path remains fail-closed. Desktop ignores self-claimed owner fields, trusts only a cryptographically valid NIP-OA auth tag, admits only verified owner-declared relay agents to observer ingestion, requires a non-empty `name` or `display_name` for a directory card, and keeps interrupt/runtime controls local-only. Card liveness comes from the same ephemeral kind `20001` heartbeat used by DMs and profile surfaces.

## Hermes runtime design

[...]

PR #2633, opened 2026-07-23
feat(acp): Hermes runtime discovery and durable session load

joelbrilliant:
Rescoped to persistence only, now that #2773 has landed.

The original version of this PR also did Hermes runtime discovery and registration. #2773 makes that generic (Hermes ships as a tier-2 preset on main), so all of it is dropped. What is left is the one piece none of the competing Hermes PRs covered: durable session bindings. That takes this from +1051/-29 across 18 files to +643/-7 across 6.

## What

Channel to ACP session bindings are in-memory only, so every harness restart begins a cold conversation even against an agent that advertises `loadSession`. This persists the mapping as a small JSON sidecar under the process data directory, keyed by `(agent command identity, agent args, channel id)` so different agent binaries and profiles never share bindings. Heartbeats stay ephemeral and are never stored.

## Cross-process safety

The store is a shared file, so a process-local cache alone is unsafe: two `buzz-acp` processes with the same agent identity would clobber each other. Every read and mutation takes a sibling lockfile, reloads the on-disk map under that lock, then writes atomically.

Removal is conditional on purpose. After a failed `session/load` another process may already have written a newer session for the same channel, so `remove_if_equals` only drops the binding it actually failed to load. A key-only remove would delete the fresher one, so the unconditional variant is deliberately not part of the API.

## Verification

| | Result |
|---|---|
| `cargo test -p buzz-acp` | 604 passed, 0 failed |
| `cargo clippy -p buzz-acp --all-targets -- -D warnings` | clean |
| `cargo fmt -p buzz-acp --check` | clean |
| `cargo check -p buzz-acp --locked` | clean |

Rebased onto `main` at the post-#2773 merge. `Cargo.lock` is included this time, so `--locked` builds do not drift.

## Coordination

The split was agreed publicly on #2656: land the runtime registration work first, then rebase this down to session persistence on top. #2773 landed that seam generically instead, which is the better outcome, so this now sits on top of #2773 rather than #2656. No overlap remains with either.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -689,6 +689,47 @@ impl AcpClient {
             .session_id)
     }
 
+    /// Send `session/load` for an existing ACP session id.
+    ///
+    /// Used after harness restart when a durable channel→session binding is
+    /// known and the agent advertised `agentCapabilities.loadSession`.
+    /// History-replay `session/update` notifications are consumed by the
+    /// request loop and logged only — they are not re-published to Buzz.
+    pub async fn session_load_full(
+        &mut self,
+        cwd: &str,
+        session_id: &str,
+        mcp_servers: Vec<McpServer>,
+    ) -> Result<SessionNewResponse, AcpError> {
+        let params = serde_json::json!({
+            "cwd": cwd,
+            "sessionId": session_id,
+            "mcpServers": mcp_servers,
+        });
+        let result = self.send_request("session/load", params).await?;
+        // Spec-compliant agents may omit sessionId on load (it is implied).
+        // Prefer the request id so callers always have a concrete binding.
+        let resolved_id = result
+            .get("sessionId")
+            .and_then(|v| v.as_str())
+            .unwrap_or(session_id)
+            .to_owned();
+        tracing::info!(target: "acp::session", "session loaded: {resolved_id}");
+        Ok(SessionNewResponse {
+            session_id: resolved_id,
+            raw: result,
+        })
+    }
+
+    /// Returns true when an initialize result advertises `loadSession`.
+    pub fn agent_supports_load_session(init_result: &serde_json::Value) -> bool {
+        init_result
+            .get("agentCapabilities")
+            .and_then(|caps| caps.get("loadSession"))
+            .and_then(|v| v.as_bool())
+            .unwrap_or(false)
+    }
+
     /// Send Goose's custom system-prompt request after `session/new`.
     pub async fn session_set_goose_system_prompt(
         &mut self,
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -694,7 +694,8 @@ impl AcpClient {
     /// Used after harness restart when a durable channel→session binding is
     /// known and the agent advertised `agentCapabilities.loadSession`.
     /// History-replay `session/update` notifications are consumed by the
-    /// request loop and logged only — they are not re-published to Buzz.
+    /// request loop without entering the observer feed, so relay observers do
+    /// not republish the loaded transcript.
     pub async fn session_load_full(
         &mut self,
         cwd: &str,
@@ -706,7 +707,9 @@ impl AcpClient {
             "sessionId": session_id,
             "mcpServers": mcp_servers,
         });
-        let result = self.send_request("session/load", params).await?;
+        let result = self
+            .send_request_with_session_update_observer("session/load", params, false)
+            .await?;
         // Spec-compliant agents may omit sessionId on load (it is implied).
         // Prefer the request id so callers always have a concrete binding.
         let resolved_id = result
@@ -1108,7 +1111,7 @@ impl AcpClient {
     /// Send a JSON-RPC request and wait for the matching response.
     ///
     /// Assigns the next available id, writes the NDJSON line to stdin,
-    /// then calls [`read_until_response`](Self::read_until_response).
+    /// then reads until the matching response arrives.
     ///
     /// The write phase is bounded by `WRITE_TIMEOUT` (30s) and the read phase
     /// by `REQUEST_TIMEOUT` (60s), so worst-case wall clock is ~90s. Non-prompt
@@ -1118,6 +1121,16 @@ impl AcpClient {
         &mut self,
         method: &str,
         params: serde_json::Value,
+    ) -> Result<serde_json::Value, AcpError> {
+        self.send_request_with_session_update_observer(method, params, true)
+            .await
+    }
+
+    async fn send_request_with_session_update_observer(
+        &mut self,
+        method: &str,
+        params: serde_json::Value,
+        observe_session_updates: bool,
     ) -> Result<serde_json::Value, AcpError> {
         let id = self.next_id;
         self.next_id += 1;
@@ -1140,7 +1153,12 @@ impl AcpClient {
             Err(_) => return Err(AcpError::Timeout(timeout)),
         }
 
-        match tokio::time::timeout(timeout, self.read_until_response(id)).await {
+        match tokio::time::timeout(
+            timeout,
+            self.read_until_response_with_session_update_observer(id, observe_session_updates),
+        )
+        .await
+        {
             Ok(result) => result,
             Err(_) => Err(AcpError::Timeout(timeout)),
         }
@@ -1150,7 +1168,7 @@ impl AcpClient {
     ///
     /// After a [`AcpError::Timeout`] from [`send_request`], the agent may
     /// eventually send the late response. That stale message will sit in the
-    /// `BufReader` buffer and be silently skipped by the next `read_until_response`
+    /// `BufReader` buffer and be silently skipped by the next response-read
     /// call (ID mismatch). However, if the caller wants a clean slate — e.g.
     /// before retrying the same method — they can call this to consume any
     /// buffered data with a short deadline.
@@ -1210,9 +1228,10 @@ impl AcpClient {
     ///
     /// Compares the incoming `id` field as a `serde_json::Value` against
     /// `json!(expected_id)` so that both numeric and string IDs work correctly.
-    async fn read_until_response(
+    async fn read_until_response_with_session_update_observer(
         &mut self,
         expected_id: u64,
+        observe_session_updates: bool,
     ) -> Result<serde_json::Value, AcpError> {
         loop {
             // LinesCodec::new_with_max_length enforces MAX_LINE_SIZE at the
@@ -1256,7 +1275,11 @@ impl AcpClient {
                     continue;
                 }
             };
-            self.observe("acp_read", msg.clone());
+            let is_session_update =
+                msg.get("method").and_then(|v| v.as_str()) == Some("session/update");
+            if observe_session_updates || !is_session_update {
+                self.observe("acp_read", msg.clone());
+            }
 
             // Check if this is a response to our expected request (has matching id
             // AND no `method` field — a `method` field means it's an agent-initiated
@@ -1303,7 +1326,7 @@ impl AcpClient {
         }
     }
 
-    /// Idle-aware message loop: like [`read_until_response`] but resets an idle
+    /// Idle-aware message loop: like the regular response-read path but resets an idle
     /// deadline on every stdout line. Fires [`AcpError::IdleTimeout`] on silence
     /// or [`AcpError::HardTimeout`] on absolute wall-clock cap.
     ///
@@ -3292,6 +3315,57 @@ mod tests {
             .await;
         assert!(result.is_ok(), "expected Ok, got {result:?}");
         assert_eq!(result.unwrap()["worked"], serde_json::json!(true));
+    }
+
+    #[tokio::test]
+    async fn session_load_suppresses_replayed_updates_from_observer_only() {
+        let script = r#"
+            read -t 2 _load
+            echo '{"jsonrpc":"2.0","method":"session/update","params":{"marker":"replayed"}}'
+            echo '{"jsonrpc":"2.0","id":0,"result":{}}'
+            read -t 2 _next
+            echo '{"jsonrpc":"2.0","method":"session/update","params":{"marker":"live"}}'
+            echo '{"jsonrpc":"2.0","id":1,"result":{"worked":true}}'
+            sleep 1
+        "#;
+        let mut client = spawn_script(script).await;
+        let observer = crate::observer::ObserverHandle::in_process();
+        client.set_observer(Some(observer.clone()), 0);
+
+        let loaded = client
+            .session_load_full("/", "sess-existing", Vec::new())
+            .await
+            .expect("session/load should succeed");
+        assert_eq!(loaded.session_id, "sess-existing");
+
+        let next = client
+            .send_request("test/echo", serde_json::json!({}))
+            .await
+            .expect("follow-up request should succeed");
+        assert_eq!(next["worked"], serde_json::json!(true));
+
+        let observed_reads: Vec<_> = observer
+            .snapshot()
+            .into_iter()
+            .filter(|event| event.kind == "acp_read")
+            .map(|event| event.payload)
+            .collect();
+        assert!(
+            !observed_reads
+                .iter()
+                .any(|payload| payload["params"]["marker"] == "replayed"),
+            "session/load replay updates must not enter the observer feed"
+        );
+        assert!(
+            observed_reads
+                .iter()
+                .any(|payload| payload["params"]["marker"] == "live"),
+            "normal session updates must remain observable after load"
+        );
+        assert!(
+            observed_reads.iter().any(|payload| payload["id"] == 0),
+            "the session/load response itself must remain observable"
+        );
     }
 
     #[tokio::test]
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -80,6 +80,54 @@ pub struct AgentModelCapabilities {
     pub available_models_raw: Option<serde_json::Value>,
 }
 
+fn model_capabilities_from_response(
+    response: &serde_json::Value,
+) -> Option<AgentModelCapabilities> {
+    let config_options_raw = extract_model_config_options(response);
+    let available_models_raw = response
+        .get("models")
+        .filter(|models| models.is_object())
+        .cloned();
+    if config_options_raw.is_empty() && available_models_raw.is_none() {
+        None
+    } else {
+        Some(AgentModelCapabilities {
+            config_options_raw,
+            available_models_raw,
+        })
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum LoadModelResolution {
+    Method(ModelSwitchMethod),
+    Unsupported,
+    Unverifiable,
+}
+
+fn resolve_load_model_switch(
+    load_response: &serde_json::Value,
+    cached: Option<&AgentModelCapabilities>,
+    desired_model: &str,
+) -> LoadModelResolution {
+    if model_capabilities_from_response(load_response).is_some() {
+        return resolve_model_switch_method(load_response, desired_model)
+            .map(LoadModelResolution::Method)
+            .unwrap_or(LoadModelResolution::Unsupported);
+    }
+
+    let Some(cached) = cached else {
+        return LoadModelResolution::Unverifiable;
+    };
+    let cached_response = serde_json::json!({
+        "configOptions": cached.config_options_raw,
+        "models": cached.available_models_raw,
+    });
+    resolve_model_switch_method(&cached_response, desired_model)
+        .map(LoadModelResolution::Method)
+        .unwrap_or(LoadModelResolution::Unsupported)
+}
+
 /// Per-channel session IDs and turn counters.
 ///
 /// Separated from `OwnedAgent` so the state machine is testable without
@@ -268,6 +316,25 @@ fn apply_completed_before_control_signal(
         ControlSignal::Rotate | ControlSignal::SwitchModel(_)
     ) {
         state.invalidate(source);
+    }
+}
+
+fn control_signal_discards_session(control_signal: &ControlSignal) -> bool {
+    matches!(
+        control_signal,
+        ControlSignal::Rotate | ControlSignal::SwitchModel(_)
+    )
+}
+
+pub(crate) fn clear_durable_channel_binding(ctx: &PromptContext, channel_id: &Uuid) -> bool {
+    ctx.session_store
+        .remove(&ctx.agent_command, &ctx.agent_args, channel_id)
+}
+
+fn clear_durable_source_binding(ctx: &PromptContext, source: &PromptSource) -> bool {
+    match source {
+        PromptSource::Channel(channel_id) => clear_durable_channel_binding(ctx, channel_id),
+        PromptSource::Heartbeat => false,
     }
 }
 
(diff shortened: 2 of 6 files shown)


joelbrilliant:
## Review request

Focused Hermes path for local managed agents (not the external-agent directory work in #2468).

**What this adds**
- Desktop + `buzz-acp` Hermes discovery (`hermes` / `hermes-acp`)
- Durable channel→session bind via `session/load`
- Onboarding picker entry
- Readiness: missing ACP deps → AdapterMissing (not fake logged-out)
- Exact `python`/`python3` marker-owned cleanup for Hermes venv scripts

**Dogfood (2026-07-24)**
Owner-only Rocky DM on Desktop: full Hermes tools after companion Nous `acp.tool_policy: profile`, instance `parallelism=1`, `BUZZ_ACP_NO_MEMORY=true` (string `true`, not `1`), multi-turn replies + session restore path.

**Checks**
- DCO / Semgrep / zizmor green
- `cargo test -p buzz-acp --lib` 595 passed locally
- Desktop hermes + interpreter unit tests green

**Companion**
https://github.com/NousResearch/hermes-agent/pull/70326

Happy to split or drop anything that blocks merge. External-agent / kind 10100 / relay-observer stays out of this PR on purpose so it can land cleanly.

rungmc357:
Hey Joel — I built on #2633 locally, preserving your three commits and authorship, and found one data-loss issue worth flagging before merge.

`SessionStore` is keyed by command/args but cached per process, so two `buzz-acp` processes using the same Hermes invocation can overwrite each other’s channel bindings. I fixed it with a stable sibling lockfile, a fresh read under lock for every `get`/`put`/`remove`, and atomic mutation writes. The regression opens two stores before either writes and proves interleaved writes and removals cannot clobber or resurrect bindings.

- Standalone three-file fix commit: https://github.com/rungmc357/buzz/commit/b8fcf51e3d6962599614f51ffe2b4802106f3a65
- Full attributed branch: https://github.com/rungmc357/buzz/tree/hermes-comprehensive-session-safe

The full branch also includes optional Hermes profile selection (`hermes --profile NAME acp`) and canonical `supports_profile` metadata from my earlier implementation. Your #2633 work remains credited as the operational backbone; #2468 and @NYTEMODEONLY remain credited as prior Hermes groundwork.

Verification: focused session-store 3/3; lost-update regression passed 75 repeated runs; clippy, formatting, and diff checks green; `buzz-acp` 595/596 with the sole keepalive timing failure reproduced unchanged on the frozen base; broader Hermes discovery/runtime/profile/onboarding/frontend gates passed.

I didn’t open a competing PR. Happy to split the session-store fix or profile slice however is easiest for you to land.

joelbrilliant:
Thanks - good catch.

Agreed on the SessionStore lost-update: process-local cache + whole-file rewrite is unsafe when two `buzz-acp` processes share the same command/args identity. I've landed the session-store fix on this PR (lock + reload under lock + atomic write + independent-open regression, plus put recovery from a corrupt sidecar):

- https://github.com/block/buzz/pull/2633/commits/3025359ae3eaa9530dca6e118d2d43755136d7e3

Verification here:
- `cargo test -p buzz-acp --lib session_store` → **4/4**
- `cargo test -p buzz-acp --lib` → **596 passed / 1 failed**, sole failure `acp::tests::keepalive_resets_idle_past_deadline` (timing flake; same class you saw on the frozen base)

I'd like to keep #2633 scoped to discovery + durable load + onboarding. Optional Hermes `--profile` / `supports_profile` is better as a follow-up PR so this one stays reviewable. Happy to review that split if you open it against `main` (or against this branch after CI settles).

Credits stay as noted for #2468 / @NYTEMODEONLY and your store fix. No need for a competing PR.

rungmc357:
Thanks, that sounds right. I don’t want to open a competing PR. Happy to help get #2633 through, and I’ll hold the optional profile support for a separate follow-up after this lands.

joelbrilliant:
Appreciate it.

Store fix is on the branch (`3025359`), DCO / Semgrep / zizmor green, and the PR is mergeable pending maintainer review. Holding profile support for a follow-up after this lands is perfect.

If you want to help while we wait, the highest-leverage things are:
1. A second pair of eyes on the dogfood notes in `docs/hermes-agent-rocky.md` (especially `parallelism=1`, `BUZZ_ACP_NO_MEMORY=true` as the string `true`, and `acp.tool_policy: profile` on the Hermes side)
2. Flagging any residual multi-process session-store edge cases you still see against `3025359`

Otherwise we can leave it for review as-is. Thanks again for catching the lost-update before merge.

rungmc357:
I took a second pass against `3025359`. The dogfood notes look right on `parallelism=1`, `BUZZ_ACP_NO_MEMORY=true` rather than `1`, and `acp.tool_policy: profile`.

I found two residual session-binding edges:

1. The restore path can still delete a newer binding. Process A reads session X, process B writes Y, then A’s failed load of X triggers a key-only remove that deletes Y. A `remove_if_equals(X)` under the existing lock, plus an interleaving regression, should close it.
2. The multi-profile section presents a different `HERMES_HOME` as equivalent to unique `-p <profile> acp` args, but durable store identity excludes environment and managed-agent identity. Two managed agents in the same channel with identical command/args and different homes can therefore collide. The simplest short-term fix is to require unique profile args and remove the HERMES_HOME-only alternative until the store is namespaced.

I didn’t find anything else material.

joelbrilliant:
Second pass taken - both residual edges were real.

**1. Failed load deleting a newer binding**
Confirmed in `try_load_persisted_session`: key-only `remove` after a failed `session/load` of X could drop a concurrent Y.

Landed `remove_if_equals(expected)` under the existing lock, wired the restore path to use it, and added an interleaving regression (`remove_if_equals_does_not_delete_newer_binding`).

**2. HERMES_HOME-only multi-profile collision**
Agreed. Durable store identity is command + args + channel, not env / managed-agent id. Updated dogfood docs + `buzz-acp` README to **require unique `-p <profile> acp` args** for a second Hermes profile and removed the HERMES_HOME-only alternative until the store is namespaced.

Commit: https://github.com/block/buzz/pull/2633/commits/3c83c1580452bc75598533bb193e94a09f478879

Verification:
- `cargo test -p buzz-acp --lib session_store` → **6/6**

Thanks again for the careful review.

joelbrilliant:
Repositioning this against #2773, now that #2656 is being closed in its favour.

The overlap I flagged on #2656 was runtime registration and model-selection UX, and BYOH supersedes that half properly - a generic seam beats two harness-specific PRs, so that's the right call. What it doesn't cover is durable sessions: I had a look through #2773 and there's no `session_store`, no pool binding, no `session/load` (the only persistence in there is composer draft state). So Buzz's channel-to-session map is still in-memory only after BYOH lands, and a restart still orphans live agent sessions - the agent has the conversation, Buzz has forgotten which session belonged to which channel, and the user quietly starts over.

That's the half this PR does, and it's independent of how a runtime gets registered. So the plan I proposed on #2656 still works with #2773 as the target instead: land BYOH, then I rebase this down to just `session_store.rs` and `pool.rs` on top of the new seam and drop anything that overlaps. I've already built that slice locally against current main - 651 lines, `cargo test -p buzz-acp` green at 604 including the six store tests - so it's a rebase rather than a rewrite whenever #2773 is in.

Also worth calling out that @rungmc357's review here found two real session-binding edges (a failed restore deleting a newer binding, and a multi-profile identity collision). Both are fixed in the current head with `remove_if_equals` under the existing lock and an interleaving regression, so the persistence layer has had a proper pass rather than going in cold.

@wpfleger96 no rush on this at all - happy to wait until BYOH is settled and then do the rebase against whatever the final seam looks like. Just sing out if you'd rather I close this and open a fresh persistence-only PR instead, either way suits.

joelbrilliant:
Thanks, appreciate it.

One thing that'll help when you look: CI on this PR is stuck on `action_required`. Every run since the branch opened is waiting on a maintainer to approve workflows for a fork contributor, so the Rust suite has never actually run here. The three green checks are just DCO, Semgrep and zizmor.

Locally against the merged head: `cargo test -p buzz-acp` 604 passed, `clippy -p buzz-acp --all-targets -- -D warnings` clean, `fmt --check` clean, `check --locked` clean. If someone kicks the workflow approval it should confirm that, but I can't trigger it from my side.

OpenSpek:
Hi Joel — thanks for rescoping #2633 onto the merged BYOH seam. This looks like the right land path, and I do not want to open a competing PR.

While validating restart-durable ACP sessions on Windows, I independently built a narrow `buzz-acp` safety layer for `session/list` → authoritative selection → `session/load`. It is currently one local, signed commit against the post-#2773 merge base and has had an exact-hash independent review. Current gates are 614 library tests, 9 integration tests, docs, fmt, and clippy all green.

The pieces that seem complementary to your store/pool work are:

- private, bounded `session/list` pagination and exact selection by adapter-process provenance, provider, session ID, and canonical absolute workspace;
- non-cloneable, generation-scoped binding authority;
- one-shot `session/load` with no automatic retry;
- `Indeterminate` quarantine after any timeout, lost/malformed response, or other failure once the load may have been sent;
- exact JSON-RPC 2.0 response validation; and
- observer/log suppression and redaction for prior-session metadata and loaded history.

The main semantic point to reconcile is load failure handling. A timeout or malformed/lost response does not prove the provider failed to load the session. Clearing the stored binding and falling through to `session/new` can therefore fork hidden provider state. The local implementation quarantines that mapping for explicit reconciliation instead.

Your cross-process store, conditional removal, and production pool wiring are complementary and remain necessary; my current slice is library-level only.

Would you prefer that I adapt this as a focused patch series directly on top of #2633, or keep it as an immediate hardening follow-up after #2633 lands? Either route is fine—I’m happy to follow your preferred ownership shape and preserve attribution.

OpenSpek:
Thanks Joel — I appreciate the quick follow-up and the attribution. I checked `7b76647e`; preserving the stored value on an ambiguous failure is the right first step.

One residual control-flow edge still appears to leave the hidden-fork path open:

- the indeterminate branch in `try_load_persisted_session` keeps the binding but returns `None`;
- the caller treats that `None` as permission to call `session/new`; and
- the successful new-session path then `put`s the replacement ID over the retained binding.

So the provider may have loaded X while Buzz creates Y and replaces X's mapping. I think the restore function needs a distinct `Indeterminate`/blocked outcome rather than collapsing it into the same `None` used for “no binding.” That outcome should stop the prompt without creating, retrying, deleting, overwriting, or publishing.

One related detail: `AcpError::AgentError` covers more than a proven missing session—auth, invalid parameters/workspace, unsupported inputs, or server failures may not establish that the session is dead. Conditional removal seems safest only for a narrowly defined not-found/rejected-session contract; otherwise quarantine.

I'm happy to provide a focused patch and regression directly atop this branch if that is useful, while keeping the broader list/select/privacy hardening in the ownership shape you prefer.

joelbrilliant:
Good catch, and you're right — that was a real bug, not a difference of opinion. Fixed in 7b76647e.

`AcpError` already separates the cases: only `AgentError` is a JSON-RPC error response, i.e. the provider answered and refused. `Timeout`, `AgentExited`, `Protocol`, and the write/idle/drain timeouts are all indeterminate. #2633 now clears the binding only on the definitive case and keeps it otherwise, behind a documented `load_failure_is_definitive()` predicate with a test pinning every indeterminate variant. It self-heals rather than going sticky, since a provider that really has lost the session answers `AgentError` on a later attempt and that clears it then.

That's the minimum needed to stop #2633 shipping a state-forking default. Your quarantine model is the fuller answer, and I'd rather it landed as its own PR than got folded into this one. This PR has finally got narrow enough that someone offered to review it, and private `session/list` pagination, provenance selection, generation-scoped binding authority and observer redaction are a substantial slice that deserves its own review rather than riding in on mine.

So: follow-up on top, not a patch series into #2633. Happy to review it when it's up, and it should rebase cleanly since I've only touched the one match arm.

joelbrilliant:
You're right on both, and the first one means my fix doesn't actually do anything.

I traced the caller: on `None` it calls `create_session_and_apply_model` and then `put`s the new id straight over the binding I'd just carefully retained. So the mapping survives about three statements. Same fork, one step later — a no-op with a tidy comment on it, which is worse than no fix, because it reads like the problem is handled.

Second point stands too. `lib.rs:3000` documents auth failures arriving as `AgentError`, so my `load_failure_is_definitive` threshold is too loose: an expired token would clear a perfectly live binding. And there is no session-not-found code convention in the codebase to narrow against, so that contract has to be defined rather than detected.

So yes please, take it. A focused patch and regression on this branch is the right shape, and you have clearly thought about the state machine harder than I have. Feel free to restructure or drop 7b76647e entirely — the predicate and its test encode the wrong threshold, so don't build on them out of politeness.

One thing I would want to agree before it lands: what `Indeterminate` does to the user. Blocking the prompt is right for correctness, but it needs to say something actionable, or the first person to hit it just sees Buzz stop responding on that channel. Happy to take that part if you would rather stay library-level.

Broader list/select/privacy hardening stays yours as its own PR, as before.

OpenSpek:
Thanks Joel — agreed. I’ll keep the broader list/select/privacy hardening as its own follow-up and won’t open a competing PR.

For #2633, the focused shape is a distinct `Indeterminate` outcome plus a regression proving it cannot create, retry, delete, overwrite, or publish after an ambiguous load. For the user-facing behavior, I propose failing closed with an explicit “session restore outcome is uncertain; no new session was created” error that directs reconciliation rather than silently leaving the channel idle. I’ll keep the patch scoped to that contract.

joelbrilliant:
Agreed on all of that, and the error wording is the right call. Explicit beats a silently idle channel.

One thing worth nailing down while you are in there: what the user does next. Failing closed is correct, but there is currently no way to clear a binding short of finding the sidecar and deleting it. I checked and there is no CLI surface for it — `BUZZ_ACP_SESSION_STORE` tells an operator where the file lives, and that is the whole recovery story. So as it stands one ambiguous load would wedge that channel indefinitely for anyone who does not know the file exists, which is a worse failure than the fork we are avoiding.

Does not have to be in your patch. But the error should name a concrete next step rather than just reporting uncertainty, even if the first version points at the file path. A small `buzz` subcommand to clear a channel's binding would be the honest version, and I am happy to take that as a follow-up on my side since it is store-surface work rather than protocol.

Otherwise the contract reads right: no create, no retry, no delete, no overwrite, no publish. Ship it and I will review.

OpenSpek:
Agreed. The focused #2633 contract should fail closed and give the operator a concrete recovery path: no create, retry, delete, overwrite, or publish; report the uncertain restore; and direct reconciliation via the configured session-store location without exposing binding metadata. A proper channel-scoped clear subcommand can remain your store-surface follow-up.

No competing PR; the broader list/select/privacy hardening stays separate. Once the focused branch change is authorized, I’ll bring that bounded patch and regression here for review.

joelbrilliant:
Contract's agreed from my side, so consider it authorised. Reporting the uncertain restore and pointing at the configured store location without leaking binding metadata is exactly right, and the channel-scoped clear subcommand stays mine.

On mechanism, since "authorized" may be about branch access: `maintainerCanModify` is on, but that only grants push to block/buzz maintainers, not to contributors, so it will not let you push here. Easiest path is to open a PR against `joelbrilliant/buzz` targeting `feat/hermes-native-profile`. You need nothing from me for that, I merge it, and #2633 picks it up with your commits and authorship intact.

If you would rather not carry a second PR, send the patch series and I will `git am` it, which preserves authorship too. Either way works, whichever is less friction for you.

OpenSpek:
Thanks Joel — understood. I’ll use the contributor PR route against `joelbrilliant/buzz:feat/hermes-native-profile`; no competing PR against `block/buzz`. I’m holding at the branch-change boundary until Luke authorizes the scoped code/branch/push work, then the concrete next step is the focused Indeterminate patch plus regression under the agreed no-create/retry/delete/overwrite/publish contract.

crates/buzz-acp/src/pool.rs
```diff
@@ -1485,6 +1618,8 @@ pub async fn run_prompt_task(
                             "created session {sid} for channel {cid}"
                         );
                         agent.state.sessions.insert(*cid, sid.clone());
+                        ctx.session_store
+                            .put(&ctx.agent_command, &ctx.agent_args, cid, &sid);
```
troyhoffman-oss, line 1839: This `put` writes the durable binding, and no rotation path ever clears it. `!rotate` and the `MaxTokens` rotation both drop the in-memory entry and leave the binding, so the next turn loads the session that was just rotated away from.

`invalidate_channel` (pool.rs:123) removes from `state.sessions` and nothing else. `session_store` is only touched at pool.rs:822 (get), 876 (remove after a definitive load failure) and here at 1621 (put after create). That covers all three invalidate callers: the `should_rotate` branch at pool.rs:2158, `invalidate_channel_sessions` from lib.rs:2132 (`!rotate` on an idle channel), and lib.rs:2002 (agent lost channel access).

So `!rotate` invalidates, then the next event hits `try_load_persisted_session` at pool.rs:1587, finds the stale binding and `session/load`s it, and the channel is back on the session the owner asked to discard. The `MaxTokens` case is worse: the reason for rotating was a context cap, and the reloaded session answers with the same cap, so the channel cannot make progress at all. README.md:156 still says the next event starts a fresh session, which was true before this PR.

Two things that bit me when fixing this same shape.

The clear cannot go inside `invalidate_channel`. That function carries two meanings, "rotate, discard this conversation" and "error, retry the same session later", and the second one has to keep the binding. That is the case durable bindings exist to serve. It needs a separate rotate path.

A pre-clear on the owner's side is not enough by itself. `!rotate` on a busy channel signals the in-flight task (lib.rs:2121), and if that task is still pre-prompt when the clear lands, this `put` runs afterward and resurrects the binding. That defeats `!rotate` inside a single process, no restart involved. What worked was clearing from inside the cancelled task, in both the cancel-ok and cancel-failed arms, and keeping the pre-clear to cover the idle fork and a dropped signal.

crates/buzz-acp/src/pool.rs
```diff
@@ -795,6 +803,113 @@ const CONTROL_CANCEL_GRACE: Duration = Duration::from_secs(5);
...
+        .session_load_full(&ctx.cwd, &stored, ctx.mcp_servers.clone())
+        .await
+    {
+        Ok(resp) => {
+            if agent.model_capabilities.is_none() {
+                agent.model_capabilities = Some(AgentModelCapabilities {
+                    config_options_raw: extract_model_config_options(&resp.raw),
+                    available_models_raw: extract_model_state(&resp.raw),
+                });
+            }
```
troyhoffman-oss, line 980: A `session/load` response usually carries neither `configOptions` nor `models`. ACP only requires those on `session/new`, which is why the identical block at pool.rs:977 gets a real catalog. `extract_model_config_options` (acp.rs:1892) returns an empty `Vec` when `configOptions` is absent and `extract_model_state` (acp.rs:1907) returns `None`, so on the first post-restart turn where the first session is a load, `model_capabilities` becomes `Some` wrapping an empty catalog. The `is_none()` guard then means it never gets refreshed for the rest of the process.

Two things fall out of that, both silent.

`switch_idle_agent_model` pre-checks against that cache at pool.rs:756. With an empty catalog `model_in_catalog` is false for everything, so every valid model comes back `UnsupportedModel` and the picker shows the operator's own configured model as refused.

`resolve_model_switch_method` on line 838 is reading the same catalog-less response, so it returns `None` and `desired_model` is never re-applied to the loaded session. The `warn!` just below only fires on an apply error, not on a `None` method, so the session runs on the agent default while the harness believes the override is on.

I hit this on exactly this block. What held was gating the capture on the response actually advertising a catalog, and treating an absent catalog as "unchanged and unverifiable" rather than "not in the catalog". A catalog that is present and genuinely lacks the model should still reject.

crates/buzz-acp/src/acp.rs
```diff
@@ -601,6 +601,47 @@ impl AcpClient {
             .session_id)
     }
 
+    /// Send `session/load` for an existing ACP session id.
+    ///
+    /// Used after harness restart when a durable channel→session binding is
+    /// known and the agent advertised `agentCapabilities.loadSession`.
+    /// History-replay `session/update` notifications are consumed by the
+    /// request loop and logged only — they are not re-published to Buzz.
```
troyhoffman-oss: The second half of this is not true today. Replayed `session/update` notifications do not reach the channel transcript, but `read_until_response` calls `self.observe("acp_read", msg.clone())` at acp.rs:1161 on every parsed line, before the `session/update` dispatch below it. So each replayed notification becomes an observer event carrying its full payload, and with `--relay-observer` on, `run_relay_observer_publisher` (lib.rs:440) encrypts and republishes the whole prior conversation to the relay, once per restart-load. Owner-scoped and encrypted, so nothing leaks, but it is a full duplicate publish that scales with conversation length.

Either narrow the comment to the transcript, or suppress at the `observe` seam for the load window.

troyhoffman-oss (commented):
Three things I ran into on this same surface.

joelbrilliant:
You’re right on all three. I traced each path on the current head:

- rotation clears only in-memory state, so the durable binding can reload and the in-flight `put` race can resurrect it
- a load without a model catalogue can poison `model_capabilities` and skip the desired model reapply
- replayed updates skip transcript delivery but still hit `acp_read` observation and get republished by the relay observer

I’ll fix these before asking for another review: a rotation-specific durable clear covering idle and in-flight paths, absent load catalogues treated as unchanged and unverifiable, and observer suppression limited to replayed `session/update` notifications during load. I’ll add regressions for all three. Thanks for laying out the failure modes so clearly.

crates/buzz-acp/src/pool.rs
joelbrilliant: Fixed in `7fbfbb21`. A catalog-less `session/load` no longer stores `Some(empty)` in `model_capabilities`. Model re-apply uses the cached advertised catalog when the load response omits one. A catalog that is present remains authoritative, while a genuinely absent catalog is treated as unchanged and unverifiable. Focused regressions cover all three cases.

crates/buzz-acp/src/pool.rs
joelbrilliant: Fixed in `7fbfbb21`. Explicit `!rotate` now clears the durable binding before signalling, and the in-flight task clears it again after both successful and failed cancellation plus the natural-completion race. Automatic `MaxTokens` and `MaxTurnRequests` rotation clears it too. Generic error invalidation still preserves the binding for retry. The full `buzz-acp` library suite passes 612 tests and clippy is clean.

crates/buzz-acp/src/acp.rs
joelbrilliant: Fixed in `7fbfbb21`. `session/load` now uses a filtered response-read path that suppresses replayed `session/update` frames at the `acp_read` observer seam. The load response, other messages, and normal updates after the load remain observable. The regression proves replay history is not republished while a later live update still reaches the observer.

joelbrilliant:
I’ve rebased this onto current `block/buzz` main at `a885622f`. The durable session store, indeterminate restore guard and rotation invalidation remain intact. The full `buzz-acp` library suite passes 654/654, with clippy, formatting and diff checks clean. GitHub now reports the PR mergeable and fresh CI is running.

joelbrilliant:
Rebased onto current `main` after #2773 (and subsequent ACP work) landed.

- Base: `eb6a37569` (upstream `main` as of this push)
- Head: `e236fa4a0` (`feat/hermes-native-profile`)
- Diff stays persistence-only: `session_store.rs` + pool/acp wiring, +1006/-13
- Cherry-pick of the three prior commits auto-merged cleanly onto post-#2773 `pool.rs`
- Local verify: `cargo test -p buzz-acp --lib` → **686 pass / 1 fail**
  - All session-store + load-failure + rotation regressions green (7 store tests + `only_an_agent_error_is_a_definitive_session_load_failure` + `session_load_suppresses_replayed_updates_from_observer_only`)
  - The single fail is `acp::tests::keepalive_resets_idle_past_deadline`, **reproduced on clean `main`** (timing flake, unrelated)
- `cargo fmt -p buzz-acp -- --check` clean
- `cargo clippy -p buzz-acp --lib -- -D warnings` clean
- DCO sign-off retained on all three commits

Still waiting on fork Actions approval for full CI, and on maintainer review.

joelbrilliant:
Hey, the CI, Desktop Release Candidate and Docker runs expired while waiting for fork workflow approval. No jobs ran, and I don’t have permission to restart them. Could a maintainer help unblock Actions? GitHub is also showing conflicts with current main.

PR #2494, opened 2026-07-23
feat(workflow): let workflows add reactions to messages

BradGroux:
Fixes #2395.

## What changes

Workflow `add_reaction` actions now publish a signed kind-7 reaction against the triggering message through the existing relay event path.

The reaction includes:

- the target event `e` tag;
- the target event kind `k` tag;
- the target author's `p` tag required by NIP-25 clients and Buzz's author-targeted push subscription;
- the workflow owner as the signer.

The sink resolves the effective target author through the same attribution helper used by ingest, so reactions to agent-authored messages target the actual author rather than a relay signer. Duplicate execution remains idempotent. The relay-backed regression now subscribes to the reaction by its actual kind-7 and `e`-tag shape, while the follow-up message retains its channel-scoped filter.

## Safety and scope

- No public reaction REST endpoint is added.
- Desktop and mobile rendering are unchanged.
- Manual workflow triggers still cannot use `add_reaction` because they have no triggering message ID.
- Custom emoji metadata remains limited to the action's existing reaction-content field.

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy -p buzz-workflow -p buzz-relay -p buzz-test-client --all-targets --all-features -- -D warnings`
- `cargo test -p buzz-workflow -p buzz-relay -p buzz-test-client --no-fail-fast`: workflow and test-client suites passed; the unprovisioned relay run reached only the existing database-dependent failures.
- `cargo test -p buzz-relay workflow_sink::integration_tests::workflow_add_reaction_persists_attributed_kind_7_and_dedupes -- --ignored --nocapture`: passed against PostgreSQL and Redis.
- `just test`: all nine unit/package/database stages passed. The workspace integration stage reproduced the unrelated `buzz-agent` timing failure `steer_folds_into_active_turn_without_cancelling`; that test passes in isolation, and this branch does not modify `buzz-agent`.

This refresh is not production deployment evidence.

Issue #2395, opened 2026-07-22
workflow: add_reaction action calls a nonexistent relay endpoint — any workflow using it always fails


leesalminen:
## Summary

The workflow engine's `add_reaction` action can never succeed: it POSTs to `{BUZZ_RELAY_BASE_URL:-http://localhost:3000}/api/messages/{message_id}/reactions` (`crates/buzz-workflow/src/executor.rs`, `add_reaction_impl`), but the relay has no such route — `router.rs` defines no `/api/messages/*` paths. Every invocation gets the SPA/404 fallback:

```
webhook error: AddReaction: relay returned 404 Not Found for message <id>:
```

Because a failed step aborts the run, any workflow that uses `add_reaction` as an early step (e.g. an ack-reaction before a `send_message`) silently kills the whole workflow — from the user's perspective, "the workflow doesn't fire."

## Repro

1. Create a `message_posted` workflow whose first step is `action: add_reaction`.
2. Post a matching message.
3. `workflow_runs` shows `status=failed`, `error_message` as above; subsequent steps never run.
4. `curl -X POST http://127.0.0.1:3000/api/messages/deadbeef/reactions` → 404 (route absent).

Also note the manual-trigger path (kind:46020) fails differently — "AddReaction: no trigger.message_id available" — which is arguably correct, but combined with the missing route the action is unusable everywhere.

## Suggested fix

Either implement the action through the relay's real reaction path (build a kind:7 event via the workflow sink, like `send_message` does — no HTTP round-trip needed since the engine runs in-process), or remove/feature-gate the action until it works. If kept HTTP-based, it also needs the host-scoped community binding (`localhost:3000` vs the deployment host) taken into account.

Found while debugging with #2385 / #2390 — happy to PR the in-process kind:7 approach.

Chessing234:
Fixed in a PR — `add_reaction` now goes through ActionSink (kind:7) instead of the missing HTTP route.

PR #2634, opened 2026-07-23
fix(sdk,desktop,cli): tag reactions with the target author (NIP-25 p tag)

### LINKED ISSUE #2568: [Bug] Reactions without p-tags
Please add a p-tag with so that a Notifications filter can work: 

```
{                                                                                                                                                                                                                                       
    "id": "bdc351d69d73625d418ab9df399db1967240212a24286c5302bd22c0e77c351c",                                                                                                                                                             
    "pubkey": "7ab8025c2e56ad922171c22f5d09acb2524cd34190080427bd62d61745bebb83",                                                                                                                                                         
    "created_at": 1784825409,                                                                                                                                                                                                             
    "kind": 7,                                                                                                                                                                                                                            
    "tags": [                                                                                                                                                                                                                             
      ["e", "8f8c92f893b2eaa0a3fbfd941f067b96054f648dbf886df266d0ebc82bcbc56f"]                                                                                                                                                           
    ],                                                                                                                                                                                                                                    
    "content": "👍",                                                                                                                                                                                                                      
    "sig": "63acc1094961d0299dfb8a43668c2d132e7d0ada4ea03bf580ae3d5f72fb3ccceb60aef2dc28323d0c8099488ed0b3edfad73157862b2135abbbf88b994a68c2"                                                                                             
  } 
```

shani-singh1:
Fixes #2568.

## Problem

Every kind:7 Buzz emits carries only the target `e` tag:

```json
{"kind":7,"tags":[["e","8f8c92f8…"]],"content":"👍", …}
```

NIP-25 also requires the reacted-to author's `p` tag. That tag is not cosmetic — it is the *only* thing a notification feed can key on. Without it, `{"kinds":[7],"#p":[<self>]}` can never match, so a reaction to your message is invisible to every NIP-01 client, which is what @vitorpamplona hit from Amethyst.

It isn't only an interop gap. The relay already ships kind 7 as push-notifiable — `PUSH_KINDS = &[7, 9, 1059, 40007, 46010]` (`crates/buzz-relay/src/handlers/push_lease.rs:15`) — and the lease subscription shape is literally `{"kinds":[7],"#p":[…]}`. The push surface is built around a tag no Buzz client produces.

Four producers were affected: `buzz-sdk`, the desktop Tauri command, `buzz-cli`, and `buzz-acp`'s 👀/💬 status reactions.

## Solution

Thread the target author through every reaction producer.

**`buzz-sdk`** — `build_reaction` / `build_custom_emoji_reaction` take `target_author: nostr::PublicKey` and emit `["p", <author>]`. Required rather than `Option`, so a caller can't silently reintroduce the non-conformant form.

**Both builders now set `.allow_self_tagging()`.** This is the subtle part: reacting to your *own* message makes the `p` tag match the signer, and nostr 0.44 strips matching `p` tags by default. Without it, self-reactions would quietly regress to the old wire form. The repo already relies on this for NIP-IA (`build_archive_identity_request`); the reaction path now does too, pinned by a regression test.

**Desktop** — `add_reaction` takes `target_pubkey` from the caller. The timeline already resolved it via `resolveEventAuthorPubkey`, so this uses the *displayed* author, not the raw signer. For relay-signed agent messages those differ, and the displayed author is who should be notified. Passing it in also avoids a second relay round trip on a hot UI path.

**`buzz-cli`** — only ever gets an event id, so it resolves the author with a kindless `ids:[…]` lookup. That filter shape is explicitly exempt from the `#p` read gate (`req.rs:1061`, "a kindless `ids` lookup is unaffected"). A target that doesn't resolve here would have been rejected by ingest anyway (`invalid: reaction target event not found`); failing early just gives a clearer message.

**`buzz-acp`** — the 👀/💬 indicators are reactions like any other, so they carry the tag too. The author is already on the batch event; `react_working` now takes `(event_id, author)` pairs. Removal is unchanged — a kind:5 delete keys off the reaction's own id, so `ReactionGuard` still only needs ids.

Tag order is `e`, `p`, `emoji`. Ingest resolves the reaction target from the **last** `e` tag (`ingest.rs:2247`), so adding a `p` tag cannot shift target resolution.

## Why this is behaviour-neutral in-app

Worth stating explicitly, since "add a p tag" could plausibly mean "start waking agents on every 👍":

- The desktop feed queries kinds `[9, 40002, 1, 45001, 45003]` (`commands/messages.rs:72`) — no kind 7.
- Mobile's activity feed uses the same kind set (`activity_provider.dart:33`).
- The ACP mention subscription defaults to `[KIND_STREAM_MESSAGE, KIND_WORKFLOW_APPROVAL_REQUESTED, KIND_STREAM_REMINDER]` (`config.rs:1157`) — no kind 7, so p-tagged reactions do not wake agents.
- Kind 7 is not in `P_GATED_KINDS`, so no read path changes.

The only things that change are NIP-25 conformance on the wire and the relay's already-advertised push surface becoming reachable.

## Validation

Toolchain note: this was built on Windows with the `x86_64-pc-windows-gnu` toolchain (no MSVC linker locally), so clippy is 1.97 rather than the pinned 1.95.

- `cargo test -p buzz-sdk --lib` — **234 passed, 0 failed**, including the new `reaction_keeps_p_tag_when_reacting_to_own_message`. I verified that test has teeth: with `.allow_self_tagging()` removed it fails with `assertion failed: has_tag(&ev, "p", …)`, confirming the nostr 0.44 scrub is real and not theoretical.
- `cargo check -p buzz-sdk -p buzz-cli -p buzz-acp` — clean.
- `cargo fmt -p buzz-sdk -p buzz-cli -p buzz-acp -- --check` — clean.
- `cargo clippy -p buzz-sdk -p buzz-cli -p buzz-acp --all-targets -- -D warnings` — the only findings are two pre-existing `question_mark` lints in `crates/buzz-acp/src/queue.rs` (313, 942), a file this PR does not touch; they come from the newer clippy, not from this change.
- `cargo check` on `desktop/src-tauri` — clean (24 pre-existing dead-code warnings in `managed_agents/`, none from this PR).
- `pnpm typecheck` and `pnpm lint` in `desktop/` — both clean.
- `biome format` on the six changed TS files — clean. (Checked against LF copies; my checkout is `core.autocrlf=true`, which makes biome report all 1600 files unformatted locally.)
- Desktop JS unit tests could not run here: the loader crashes on Windows with `ERR_UNSUPPORTED_ESM_URL_SCHEME … Received protocol 'c:'` on a clean tree too. No test references `add_reaction`, so nothing in that suite covers this path either way.

The E2E mock (`e2eBridge.ts`) now emits the `p` tag as well, so `inbox-reactions.spec.ts`'s "real add_reaction wire shape" assertion stays honest. It falls back to reading the author out of its own mock event store when a spec drives `add_reaction` directly without one, so the two existing specs that do that keep working unchanged.

## Follow-up, not done here

NIP-25 also suggests a `k` tag carrying the target's kind, which helps clients filter reaction notifications. Every call site now has the target event in hand so it would be cheap, but it's a separate wire change and I kept this PR to the reported bug.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -2485,6 +2485,10 @@ async fn tokio_main() -> Result<()> {
                             // buzz_event.event (needed for mode gate below).
                             let author_hex = buzz_event.event.pubkey.to_hex();
                             let event_id_hex = buzz_event.event.id.to_hex();
+                            // Captured before `buzz_event.event` moves into the
+                            // queue; the 👀 reaction below needs it for its
+                            // NIP-25 `p` tag.
+                            let event_author = buzz_event.event.pubkey;
                             // Clone for the non-cancelling steer fork, which
                             // needs the event to render the steer body. The
                             // clone is unconditional because we don't know
@@ -2512,7 +2516,7 @@ async fn tokio_main() -> Result<()> {
                                 let rc = ctx.rest_client.clone();
                                 let eid = event_id_hex.clone();
                                 tokio::spawn(async move {
-                                    pool::reaction_add(&rc, &eid, "👀").await;
+                                    pool::reaction_add(&rc, &eid, event_author, "👀").await;
                                 });
                             }
                             // Event is already queued. If mode requires it AND
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -1466,14 +1466,30 @@ pub async fn run_prompt_task(
     let liveness_handle = tokio::spawn(liveness);
     let liveness_guard = LivenessGuard::new(liveness_handle, liveness_state);
 
-    // Collects event IDs up front. On drop (any exit path — normal, early
-    // return, or panic), spawns best-effort cleanup of both 👀 and 💬.
+    // Collects reaction targets up front. On drop (any exit path — normal,
+    // early return, or panic), spawns best-effort cleanup of both 👀 and 💬.
     // See `ReactionGuard` docs for ordering guarantees and known edge cases.
-    let reaction_ids: Vec<String> = batch
+    let reaction_targets: Vec<ReactionTarget> = batch
         .as_ref()
-        .map(|b| b.events.iter().map(|be| be.event.id.to_hex()).collect())
+        .map(|b| {
+            b.events
+                .iter()
+                .map(|be| ReactionTarget {
+                    event_id: be.event.id.to_hex(),
+                    author: be.event.pubkey,
+                })
+                .collect()
+        })
         .unwrap_or_default();
-    let _reaction_guard = ReactionGuard::new(ctx.rest_client.clone(), reaction_ids.clone());
+    // Removal is a kind:5 delete keyed on the reaction's own id, so cleanup
+    // only ever needs the target event ids.
+    let _reaction_guard = ReactionGuard::new(
+        ctx.rest_client.clone(),
+        reaction_targets
+            .iter()
+            .map(|t| t.event_id.clone())
+            .collect(),
+    );
 
     //
     // Core memory is delivered inside the system prompt the harness already
@@ -1936,11 +1952,11 @@ pub async fn run_prompt_task(
     // 💬 — fire-and-forget so the prompt fires immediately.
     // The guard's cleanup (spawned on drop) removes 💬 after the turn completes.
     // A brief race where 💬 appears slightly after the agent starts is acceptable.
-    if !reaction_ids.is_empty() {
+    if !reaction_targets.is_empty() {
         let rest = ctx.rest_client.clone();
-        let ids = reaction_ids.clone();
+        let targets = reaction_targets.clone();
         tokio::spawn(async move {
-            react_working(&rest, &ids).await;
+            react_working(&rest, &targets).await;
         });
     }
 
@@ -3836,7 +3852,16 @@ fn pct_encode(s: &str) -> String {
 /// Builds a reaction event with `buzz_sdk::build_reaction`, signs it with
 /// the keys already stored in `RestClient`, and submits via `POST /events`.
 /// Returns immediately on timeout or any error — reactions are cosmetic.
-pub(crate) async fn reaction_add(rest: &crate::relay::RestClient, event_id: &str, emoji: &str) {
+///
+/// `target_author` is the pubkey of the event being reacted to; it becomes the
+/// NIP-25 `p` tag. The agent's 👀/💬 indicators are reactions like any other,
+/// so they carry it too rather than emitting a non-conformant kind:7.
+pub(crate) async fn reaction_add(
+    rest: &crate::relay::RestClient,
+    event_id: &str,
+    target_author: nostr::PublicKey,
+    emoji: &str,
+) {
     let target_id = match nostr::EventId::from_hex(event_id) {
         Ok(id) => id,
         Err(e) => {
@@ -3844,7 +3869,7 @@ pub(crate) async fn reaction_add(rest: &crate::relay::RestClient, event_id: &str
             return;
         }
     };
-    let builder = match buzz_sdk::build_reaction(target_id, emoji) {
+    let builder = match buzz_sdk::build_reaction(target_id, target_author, emoji) {
         Ok(b) => b,
         Err(e) => {
             tracing::warn!(event_id, emoji, "reaction add: build failed: {e}");
@@ -3994,14 +4019,27 @@ pub(crate) async fn reaction_remove(rest: &crate::relay::RestClient, event_id: &
 /// Prevents unbounded parallelism when a large batch of events arrives.
 const REACTION_CONCURRENCY: usize = 10;
 
+/// An event an agent status reaction is attached to.
+///
+/// Adding a reaction needs the target's author for the NIP-25 `p` tag;
+/// removing one does not (a kind:5 delete keys off the reaction's own id).
+#[derive(Clone)]
+pub(crate) struct ReactionTarget {
+    pub(crate) event_id: String,
+    pub(crate) author: nostr::PublicKey,
+}
+
 /// Add 💬 to all events, capped at `REACTION_CONCURRENCY` concurrent requests.
 /// Awaited inline before the prompt fires.
-async fn react_working(rest: &crate::relay::RestClient, event_ids: &[String]) {
-    for chunk in event_ids.chunks(REACTION_CONCURRENCY) {
+///
+/// Targets are `(event_id, author)` pairs — the author rides along so each
+/// kind:7 can carry its NIP-25 `p` tag.
+async fn react_working(rest: &crate::relay::RestClient, targets: &[ReactionTarget]) {
+    for chunk in targets.chunks(REACTION_CONCURRENCY) {
         futures_util::future::join_all(
             chunk
                 .iter()
-                .map(|eid| reaction_add(rest, eid, REACTION_WORKING)),
+                .map(|t| reaction_add(rest, &t.event_id, t.author, REACTION_WORKING)),
         )
         .await;
     }
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -3854,8 +3854,8 @@ fn pct_encode(s: &str) -> String {
 /// Returns immediately on timeout or any error — reactions are cosmetic.
 ///
 /// `target_author` is the pubkey of the event being reacted to; it becomes the
-/// NIP-25 `p` tag. The agent's 👀/💬 indicators are reactions like any other,
-/// so they carry it too rather than emitting a non-conformant kind:7.
+/// NIP-25 `p` tag (a SHOULD). The agent's 👀/💬 indicators are reactions like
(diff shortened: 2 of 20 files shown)


dophsquare:

Confirmed the interop gap: kind:7 emits carried only the `e` tag, but NIP-25 requires the reacted-to author's `p` tag, and it's the only thing a `{"kinds":[7],"#p":[<self>]}` notification filter can key on. Without it a reaction to your message is invisible to every NIP-01 client — exactly what @vitorpamplona hit from Amethyst. It's also a functional bug for Buzz itself since kind 7 ships in `PUSH_KINDS`.

The plumbing looks right: `reaction_add` now takes `target_author: PublicKey` for the `p` tag, and the 👀/💬 indicators are threaded through `ReactionTarget { event_id, author }` so they're conformant too rather than emitting non-conformant kind:7. Nice touch keeping `ReactionGuard` cleanup keyed on just the reaction's own id (kind:5 delete) — cleanup correctly doesn't need the author.

Ask before merge: any other kind:7 emit sites outside `buzz-acp` (SDK `build_reaction` callers in desktop/cli) that still omit the `p` tag? Title says sdk,desktop,cli so I assume covered — just want the negative confirmed. LGTM.

shani-singh1:
Thanks — and the negative wasn't clean, so good question to ask. Swept the tree for kind:7 construction across all four languages and found **one more emit site the original push missed: the Flutter client.**

`ChannelActions.addReaction` (`mobile/lib/features/channels/channel_management_provider.dart`) emitted `["e", target]` plus an optional NIP-30 emoji tag and no `p` tag — the same bug, just not reachable from any Rust call site, so the SDK signature change didn't surface it. Fixed in 8bc1e973: it now takes the target author, and all four call sites already had it (`TimelineMessage.pubkey` for the channel/thread rows and the action sheet, `UserNote.pubkey` threaded through `toggleNoteUpvote` for Pulse).

One asymmetry worth noting: mobile needs no self-tagging opt-out. The Dart signer (`signed_event_relay.dart`) passes tags through verbatim, so a self-reaction keeps its `p` tag without the `.allow_self_tagging()` the nostr 0.44 builders require.

Full sweep, for the record:

| Site | Status |
|---|---|
| `buzz-sdk` `build_reaction` / `build_custom_emoji_reaction` | `p` tag ✅ |
| `desktop/src-tauri/src/events.rs::build_reaction` | `p` tag ✅ |
| `buzz-acp` `reaction_add` (👀/💬) | `p` tag ✅ |
| `buzz-cli` `cmd_add_reaction` | `p` tag ✅ (author via `ids:[…]` lookup) |
| `mobile` `ChannelActions.addReaction` | `p` tag ✅ — **added in 8bc1e973** |
| `buzz-db/src/event.rs:1827` | `#[cfg(test)]` fixture, not an emit site |
| `desktop/src-tauri/src/commands/social.rs:417` | `#[cfg(test)]` fixture, not an emit site |

Two deliberate exclusions:

- **`buzz-workflow`'s `add_reaction` action** doesn't build a kind:7 at all — `add_reaction_impl` POSTs to `/api/messages/{id}/reactions`, a REST endpoint that doesn't exist on the relay. That's #2395, already covered by #2494/#2495, so I left it alone rather than collide.
- **`web/`** has no reaction write path (repo browser only).

Caveat on the mobile commit, stated plainly: there's no Dart toolchain on this machine (hermit's bootstrap fails on Windows), so I could not run `dart analyze` or `flutter test` on it. The change is mechanical — one signature, one tag, four call sites, all passing a non-nullable `String` field that already exists on the relevant model — and no mobile test references `addReaction` or `toggleNoteUpvote`. But CI is the first thing to actually compile it, so that's where to look if it's unhappy.

wpfleger96 (changes requested):
🤖 Thanks for the thorough coverage here — the SDK builders preserve the required target `e` tag and correctly use `.allow_self_tagging()` with a discriminating self-reaction test, the CLI resolves the target by `ids`, the ACP pool carries `(event_id, author)` through the 👀/💬 fan-out, and all four mobile call sites pass non-null author fields. We also scoped a tree-wide search and found no other live kind-7 constructor you missed. Two blocking issues though:

**Blocking:**

- **File-size limit — this is the Desktop Core CI red.** The change grows `desktop/src-tauri/src/events.rs` to 1,017 lines, over the repo's enforced 1,000-line limit (`just desktop-check` fails with `src-tauri/src/events.rs: 1017 lines (limit 1000)`). The file was at 999 before this branch, so it's this diff that tips it. Rather than adding a size-limit exception, please move the reaction builder (and its tests) into a focused event-builder module.
- **Desktop emits the displayed/effective author instead of the target event's signer.** At `desktop/src-tauri/src/commands/messages.rs:819-845` the `p` tag comes from `TimelineMessage.pubkey`, which `desktop/src/features/messages/lib/formatTimelineMessages.ts:416-437` can rewrite from a relay-signed event's `actor`/first `p` tag. NIP-25's `p` tag is the pubkey of the event being reacted to — i.e. the signer — and every other surface in this PR (SDK, CLI, ACP, mobile) uses the raw event author. As written, the same target event can get different `p` tags depending on which client reacts, and a delegated actor can be notified while the signer is omitted. Please use the already-retained `signerPubkey` for the protocol tag; if there's product intent to also notify the effective actor, that can be an additional `p` tag as a separate decision.

**Minor, non-blocking:**

- The spec wording in the comments is a bit overstated (`crates/buzz-sdk/src/builders.rs:462-468`, `crates/buzz-cli/src/commands/reactions.rs:9-15`, `mobile/lib/features/channels/channel_management_provider.dart:437-442`): current NIP-25 makes the target `e` tag a MUST but the target-author `p` tag a SHOULD. The interop fix is still absolutely worth doing — just soften "requires/non-conformant" to "recommends".
- `desktop/src/testing/e2eBridge.ts:8322-8333` accepts a missing `targetPubkey` and can emit `["p", ""]`, so the E2E specs that call it directly don't actually prove the now-required argument. Making the mock require/validate the argument would let the existing specs pin the real contract.

Happy to re-review once the builder moves out of `events.rs` and the desktop path uses `signerPubkey`.

shani-singh1:
Thanks for the thorough review — both blocking points fixed, plus the two nits. Pushed as three focused commits.

**1. File-size / Desktop Core red** — `refactor(desktop): split reaction builders into events/reactions.rs` (9c191f49). Converted `events.rs` → `events/mod.rs` and moved the kind:7/kind:5 builders into a focused `events/reactions.rs`; `tag` and `MAX_EMOJI_CHARS` are now `pub(super)`, and the builders are re-exported so `events::build_reaction` callers are untouched. mod.rs is 984 lines. I also added the four reaction tests that were missing (e+p tags, the self-reaction `p` survives nostr's self-tag scrub, over-long emoji rejected, remove targets the reaction id).

**2. Signer, not display author** — `fix(desktop): use the reacted-to event's signer for the NIP-25 p tag` (df56b0d5). You're right, and this was the one real behavioural defect in the PR: desktop was the only surface pulling from the resolved display author. Both timeline handlers now pass the already-retained `signerPubkey`. Pulse stays on `note.pubkey` — a kind:1 social note is user-signed with no delegation, so there the author *is* the signer; happy to make that explicit differently if you'd prefer.

**3a. Wording** — softened "requires/non-conformant" → "recommends (SHOULD)" across SDK, CLI, ACP, mobile, and the bridge docs. `e` tag stays described as the MUST.

**3b. e2e mock** — the mock no longer accepts a missing pubkey. It resolves the `p` value from the mock store (the recorded event's signer — ground truth), throws if absent, and when the caller supplies one (the UI path passes `signerPubkey`) verifies it agrees case-insensitively. So `inbox-reactions`/`messaging` (direct invoke) get a correct signer tag for free, and `custom-emoji` (UI-driven) now pins that the client sends the signer rather than the display author.

Verification, with one honest gap:

- `cargo check --lib` and `cargo check --tests` on `desktop/src-tauri` — both clean (0 errors). I used `check` because `cargo test` can't link locally: the merge from main pulled in `sherpa-onnx-sys` (huddle STT), whose native static lib isn't on my Windows box, so the test *binary* won't link here. `--tests` still type-checks the new reaction test module, and the identical builder logic is exercised by the buzz-sdk tests, which do run and pass. CI-Linux has the native lib and runs them for real.
- `pnpm typecheck`, `pnpm lint`, `check:file-sizes`, and biome format on the changed files — all clean.
- `cargo fmt`/`clippy -D warnings` on the workspace crates and `cargo fmt` on the tauri crate — clean.
- Dart is unverified locally (no toolchain that bootstraps on Windows), but this commit only touches a doc comment there; the mobile logic landed earlier and CI's Mobile job was green on it.

Ready for another look when you have a moment.

vitorpamplona:
FYI, Buzz has a similar issue with replies in mini-threads.

shani-singh1:
@vitorpamplona good catch — confirmed. Threaded replies emit the NIP-10 `e` markers but drop the matching `p` tag for the parent author, so (exactly like the reaction case) replying to someone in a thread never notifies them unless you also type their `@name`. Same root shape: the `p` tag only comes from explicit mentions, in the SDK `thread_tags`, the desktop `send_channel_message`, and mobile `send_message_provider`.

Filed it separately so this PR stays scoped to reactions: #3032. Fix in progress there across the surfaces (using the parent's signer for the tag, per the review here).

dophsquare:

PR #2144, opened 2026-07-19
chore(mobile): lighter-weight release process

brow:
Streamline mobile releasing:
- A pull request is no longer necessary to release. Instead the release script just tags a commit.
- That tag (`mobile-vX.Y.Z-rc.N`) is now the sole source of truth mapping a version to a commit. The version is not stored in code/config; the release pipeline injects it at build time.
- We can release multiple release candidates (e.g. `rc.1`) per marketing version (e.g. `v0.4.11`). 

### Rollout order (required)

1. Merge and deploy [squareup/buzz-releases#71](https://github.com/squareup/buzz-releases/pull/71) first.
2. Verify its candidate-tag-only input and tag-derived versioning are live.
3. Then update this PR from current `main` and merge it.

This PR is not independently deployable: until #71 is live, the deployed private pipeline still validates the checked-in pubspec version, which this PR zeroes out.

### How is it tested?

Build and run.

Added tests:

- [`scripts/test-mobile-release-contract.sh`](https://github.com/block/buzz/blob/tomb/lighter-weight-mobile-releasing/scripts/test-mobile-release-contract.sh)
- [`scripts/test-mobile-release-candidate-publisher.sh`](https://github.com/block/buzz/blob/tomb/lighter-weight-mobile-releasing/scripts/test-mobile-release-candidate-publisher.sh)

Existing release-reference contracts, shell static checks, workflow YAML parsing, and repository pre-push checks pass.

*🤖 This PR was authored with an agent.*

PR #2636, opened 2026-07-23
chore(mobile): relax release check

brow:
### What changed?

Reverts commit [169d3e43f](https://github.com/block/buzz/commit/169d3e43f83fc151aef98ea3ef5ce6543163a484) that was added to #2144 during review.

`require_release_tag_ruleset` returns to validating the App token's effective `current_user_can_bypass` value.

### Why?

The exact bypass-actor assertion added in #2144's final review round cannot work with the release App's permissions.

GitHub returns a ruleset's `bypass_actors` list only to admin-capable tokens; everyone else gets `null`. The release App deliberately has only Contents write and Metadata read, so the assertion fails on every candidate publication and blocks all mobile releases. Verified in production: [run 30048723817](https://github.com/block/buzz/actions/runs/30048723817) failed with `could not verify Release tag ruleset 14378754 bypass actors`.

The only way to make it work would be granting the release App Administration read, and that is the wrong trade. It widens what a compromised release credential can see across repository administration solely to power a self-check, on a credential whose value comes from being narrow.

Relying on `current_user_can_bypass` instead was the original deliberate design, not an oversight: [RELEASING.md](https://github.com/block/buzz/blob/eda53c382389070cadad463b211b55f9986bb509/RELEASING.md#L196-L200) documented that the App "does not require GitHub Releases permissions, repository Administration permission, or a mobile release-branch ruleset" and that "the publisher validates the App token's effective `current_user_can_bypass` value rather than reading the ruleset's hidden bypass actor list."

What we give up is the drift check requested in [review](https://github.com/block/buzz/pull/2144#pullrequestreview-4767999381): a second bypass actor quietly added to the ruleset would no longer be caught at publication time. That remains a known, accepted gap.

The retained check still fails closed if the release App itself loses its `always` bypass. If we want drift detection back, the right home is an out-of-band audit running with admin credentials (for example a scheduled check), not the release path.

### How is it tested?

`scripts/test-mobile-release-candidate-publisher.sh` and `scripts/test-mobile-release-contract.sh` pass at this head.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -221,8 +221,9 @@ release App because strict tag protection denies direct human creation. The App
 must be installed on `block/buzz`, have Contents write and Metadata read, and
 retain an `always` bypass on the immutable `mobile-v*` tag rules. It does not
 require GitHub Releases permissions, repository Administration permission, or a
-mobile release-branch ruleset. The publisher validates both the App token's
-effective `current_user_can_bypass` value and the exact ruleset bypass actor set.
+mobile release-branch ruleset. The publisher validates the App token's effective
+`current_user_can_bypass` value rather than reading the ruleset's hidden bypass
+actor list.
 
 ---
 
--- a/scripts/release-rulesets.sh
+++ b/scripts/release-rulesets.sh
@@ -22,7 +22,7 @@ require_canonical_repository() {
 }
 
 require_release_tag_ruleset() {
-  local ruleset_endpoint state can_bypass bypass_actors rule_types includes excludes
+  local ruleset_endpoint state can_bypass rule_types includes excludes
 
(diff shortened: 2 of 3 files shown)


wesbillman (approved):
Reviewed the revert and traced the release candidate publisher path. The removed bypass-actor query is unreadable to the deliberately narrow App token and is the direct cause of the production publication failure. The retained checks still fail closed on inactive protection, loss of the App token’s effective always-bypass, rule drift, scope drift, and exclusions, before any tag object is created. The reduced drift detection is accurately documented. CI is green at 503d1826b927ac5f2439754977436bef875c4290, including both mobile release contract scripts. No blocking findings.

Issue #2620, opened 2026-07-23
buzz-audit: hash chain is tamper-evident in schema but not operationally verifiable; verify_chain accepts truncated chains


SeanGearin:
## The gap

`buzz-audit` implements a real per-community SHA-256 hash chain, but "tamper-evident" is currently a property of the schema, not of operations:

1. **`verify_chain` has no operational caller.** It's invoked only from tests. The conformance notes (`crates/buzz-test-client/tests/conformance_multitenant.rs:2686`) describe audit reads as *"operator-internal (consumed by `buzz-admin`)"* — but `buzz-admin` declares the `buzz-audit` dependency and never uses it. Nothing deployed can actually verify a chain.

2. **The verifier accepts truncated chains:**
   - *Prefix truncation:* `verify_chain(c, 1, N)` seeds `expected_prev = None`, so the first row is never checked to be genesis — `DELETE ... WHERE seq <= k` leaves a suffix that verifies clean.
   - *Unanchored segments:* `verify_chain(c, 5, 10)` never compares row 5's `prev_hash` against row 4's stored hash; the left edge of a mid-chain range is unchecked.
   - *Tail truncation:* nothing records what the walk covered vs. what should exist, so deleting the newest rows is invisible to a walk over stored rows (the head is defined by what's stored).
   - *Interior deletions* surface as a generic `ChainViolation`, misdiagnosed as mutation rather than deletion.

So the strongest thing the docs claim — a tamper-evident audit chain — isn't checkable by anything an operator can run, and several deletion classes verify clean.

## Proposed approach (no schema or wire changes)

- A pure `verify_entries` (shared chain walk with an explicit left edge — genesis or an anchor hash — plus seq contiguity, prev-hash linkage, per-entry hash recompute), unit-testable without Postgres.
- `verify_chain` keeps its signature but becomes anchored (`from_seq > 1` requires the `from_seq-1` entry to exist and be linked).
- A new `verify_full_chain` returning `{entries_verified, head_seq, head_hash}`; passing a previously-recorded head as `--expect-head` closes the tail-truncation hole.
- New `AuditError` variants (`MissingGenesis`, `MissingAnchor`, `SequenceGap`, `TruncatedTail`), carrying seq values only (same error-text sanitization fence as today).
[...]

PR #2638, opened 2026-07-23
fix(audit): hash created_at at the precision Postgres stores

### LINKED ISSUE #2637: buzz-audit: every hash chain fails verification — created_at is hashed at nanosecond precision but stored at microsecond precision
## Summary

`buzz-audit` writes each entry's hash over a **nanosecond-precision** `created_at`, then stores that timestamp in a `TIMESTAMPTZ` column, which Postgres keeps at **microsecond** precision. `verify_chain` re-reads the truncated value and recomputes the digest, so it never reproduces the one that was written.

The result: **every audit chain backed by a real database fails verification at its first entry**, on untampered data. The hash chain — the crate's whole purpose, and what `ARCHITECTURE.md:497` describes as the tamper-detection mechanism — provides no tamper evidence today. Worse, a genuinely forged row is indistinguishable from the baseline failure, so `HashMismatch` carries no signal at all.

Verified on `main` @ `6a56c8bd`.

## Root cause

Three lines that individually look fine:

1. **Write** — `crates/buzz-audit/src/service.rs:103`
   ```rust
   let created_at: DateTime<Utc> = Utc::now();   // nanosecond resolution
   ```

2. **Hash** — `crates/buzz-audit/src/hash.rs:24`
   ```rust
   hasher.update(entry.created_at.to_rfc3339().as_bytes());
   ```
   `to_rfc3339()` uses `SecondsFormat::AutoSi`, which emits 0/3/6/**9** fractional digits depending on the value. A nanosecond timestamp is hashed as `…T12:00:00.123456789+00:00`.

3. **Store** — `migrations/0001_initial_schema.sql:615`
   ```sql
   created_at      TIMESTAMPTZ NOT NULL DEFAULT NOW(),
   ```
   `TIMESTAMPTZ` resolution is 1 µs. The sub-microsecond digits are gone.

`verify_chain` (`service.rs:188`) then recomputes over `…T12:00:00.123456+00:00` — a different byte string, therefore a different digest — and returns `HashMismatch` at `service.rs:190`.

This fires whenever `Utc::now()`'s nanosecond component is not an exact multiple of 1000: ~100% of entries on Linux (`clock_gettime`, ns granularity), ~90% on Windows (100 ns ticks).

The pattern predates the multi-tenant rewrite — `git log -S` shows `timestamp.to_rfc3339()` in the hash preimage back at `d99ad131`, so this has never worked.

## Reproduction

Postgres discards the sub-microsecond digits:

```
$ docker run --rm -d --name pg -e POSTGRES_PASSWORD=x -p 55432:5432 postgres:16-alpine
$ docker exec pg psql -U postgres -t -c \
    "CREATE TABLE t (ts timestamptz);
     INSERT INTO t VALUES ('2026-07-24T12:00:00.123456789Z');
     SELECT to_char(ts, 'YYYY-MM-DD\"T\"HH24:MI:SS.US') FROM t;"

 2026-07-24T12:00:00.123457
```

The repo's own audit tests then fail against a real database. They are `#[ignore = "requires Postgres"]`, so this is normally never executed:

```
$ docker run --rm -d --name pg -e POSTGRES_USER=buzz -e POSTGRES_PASSWORD=buzz_dev \
    -e POSTGRES_DB=buzz -p 5432:5432 postgres:17-alpine
$ for f in migrations/*.sql; do docker exec -i pg psql -U buzz -d buzz -q < "$f"; done
$ cargo test -p buzz-audit --lib -- --ignored --test-threads=1
```

```
running 6 tests
test service::tests::chain_links_within_one_community ... FAILED
test service::tests::chains_are_independent_per_community ... FAILED
test service::tests::community_chain_starts_at_seq_1_with_null_prev ... ok
test service::tests::cross_community_row_does_not_verify ... ok
test service::tests::verify_detects_tampering_within_a_community ... FAILED
test service::tests::verify_empty_range_is_false ... ok

---- service::tests::chain_links_within_one_community stdout ----
panicked at crates\buzz-audit\src\service.rs:347:14:
called `Result::unwrap()` on an `Err` value: HashMismatch { seq: 2 }

---- service::tests::chains_are_independent_per_community stdout ----
panicked at crates\buzz-audit\src\service.rs:396:14:
called `Result::unwrap()` on an `Err` value: HashMismatch { seq: 1 }

---- service::tests::verify_detects_tampering_within_a_community stdout ----
panicked at crates\buzz-audit\src\service.rs:446:9:
assertion failed: matches!(r, Err(AuditError::HashMismatch { seq }) if seq == e2.seq)

test result: FAILED. 3 passed; 3 failed
```

Note the third failure specifically: **the tamper-detection test is itself invalidated.** It expects `HashMismatch` at the tampered entry's `seq`, but verification already fails at an earlier, untampered row — so the assertion that proves tamper detection works has been passing-by-absence, never actually exercised.

Of the three that "pass": `verify_empty_range_is_false` and `community_chain_starts_at_seq_1_with_null_prev` don't recompute hashes, and `cross_community_row_does_not_verify` asserts a *failure* — which it now gets for the wrong reason.

## Why CI is green

- All six `buzz-audit` chain tests are `#[ignore = "requires Postgres"]` (`service.rs:296, 316, 354, 415, 453, 490`).
- The in-process tests in `hash.rs` never round-trip through Postgres, and their fixture uses `"2026-01-01T00:00:00Z"` — zero sub-seconds, the one value where the bug cannot appear.
- `crates/buzz-relay/src/handlers/event.rs:1946` (`audit_chain_is_isolated_per_tenant_through_relay_ingest`) *is* un-ignored and asserts `"A's chain must verify independently"`, but it early-returns with `"skipping audit isolation test: Postgres/Redis unavailable"` when infra is absent.

## Impact

- No deployment can detect audit-log tampering. `verify_chain` fails identically on clean and forged data.
- `AuditError::HashMismatch` is unusable as an alarm — it is the steady state.
- **This blocks #2620.** That issue proposes a shared `verify_entries` walk with per-entry hash recompute plus a `buzz-admin audit verify` command. Its six new Postgres-free unit tests build entries in memory and would pass, but its four `#[ignore]` Postgres tests and the operator command itself would fail on every real chain until this is fixed. Worth landing this first so that work has a verifiable baseline.

## Suggested fix

Reduce `created_at` to the stored precision *before* hashing, so the in-memory entry and the row are byte-identical:

```rust
use chrono::SubsecRound;
let created_at: DateTime<Utc> = Utc::now().trunc_subsecs(6);
```

`log_inner` is the only place that assigns `created_at` (every caller goes through `NewAuditEntry`, which carries no timestamp), so it is a single choke point.

Existing rows written before the fix stay unverifiable — they always were — so this needs no migration, but operators with a chain they care about would need to re-anchor.

I have this implemented with three Postgres-free regression tests (the invariant, the trap that caused it, and the round-trip) plus a non-ignored test on the write path, and all six Postgres chain tests now pass. PR incoming.

shani-singh1:
Fixes #2637 — full analysis and reproduction there.

## Problem

Audit entries are stamped and hashed with `Utc::now()` (nanoseconds), then stored in a `TIMESTAMPTZ` column (microseconds). `compute_hash` covers `created_at.to_rfc3339()`, and chrono emits 0/3/6/**9** fractional digits depending on the value — so the digest written at `service.rs:103` is computed over `…T12:00:00.123456789+00:00` while `verify_chain` recomputes over the `…T12:00:00.123456+00:00` that Postgres hands back.

Every hash chain backed by a real database therefore fails verification at its first entry, on untampered data. That is not just a broken feature — it means a genuinely forged row is indistinguishable from the permanent baseline failure, so `HashMismatch` carries no signal.

It is invisible in CI because all six chain tests are `#[ignore = "requires Postgres"]`, and the in-process `hash.rs` tests use a fixture timestamp of `2026-01-01T00:00:00Z` — zero sub-seconds, the one value where the bug cannot appear.

## Solution

Reduce `created_at` to the stored precision *before* hashing, so the in-memory entry and the row are byte-identical:

```rust
pub fn to_storage_precision(created_at: DateTime<Utc>) -> DateTime<Utc> {
    created_at.trunc_subsecs(6)
}
```

`log_inner` is the only place that assigns `created_at` — every caller goes through `NewAuditEntry`, which carries no timestamp — so this is a single choke point. It is wrapped in a `log_timestamp()` helper purely so the invariant is assertable without a database.

I chose truncation at the write path over the alternative (hashing a precision-independent encoding such as `timestamp_micros().to_be_bytes()`). Both fix the mismatch, but truncating keeps the existing hash preimage format and gives the stronger invariant: the `AuditEntry` returned from `log()` is now exactly what a later read returns.

Truncation matches what actually happens on the wire — sqlx encodes `DateTime<Utc>` as microseconds since the Postgres epoch, truncating — so the value hashed is the value stored.

## Validation

Toolchain note: built on Windows with the `x86_64-pc-windows-gnu` toolchain (no MSVC linker locally).

**Before**, against Postgres 17 with `migrations/*` applied:

```
$ cargo test -p buzz-audit --lib -- --ignored --test-threads=1

test service::tests::chain_links_within_one_community ... FAILED
test service::tests::chains_are_independent_per_community ... FAILED
test service::tests::community_chain_starts_at_seq_1_with_null_prev ... ok
test service::tests::cross_community_row_does_not_verify ... ok
test service::tests::verify_detects_tampering_within_a_community ... FAILED
test service::tests::verify_empty_range_is_false ... ok

test result: FAILED. 3 passed; 3 failed
```

with `HashMismatch { seq: 2 }` / `HashMismatch { seq: 1 }` on untampered chains.

**After**, same database:

```
test result: ok. 6 passed; 0 failed
```

`verify_detects_tampering_within_a_community` is the one to look at: it asserts `HashMismatch` lands on the *tampered* entry's `seq`. It was failing because verification already blew up on an earlier untampered row — so the assertion proving tamper detection works had never actually been exercised. It passes now.

Also:
- `cargo test -p buzz-audit --lib` (no Postgres) — 12 passed, 0 failed.
- `cargo clippy -p buzz-audit --all-targets -- -D warnings` — clean.
- `cargo fmt -p buzz-audit -- --check` — clean.

## New tests

Three in `hash.rs`, none needing Postgres:

- `storage_precision_drops_sub_microsecond_digits` — the helper's contract, and that it is idempotent so a re-read value is unchanged.
- `nanosecond_timestamps_cannot_survive_a_database_round_trip` — asserts the digests **differ**. This is the trap itself, written down so the next person changing the hash preimage sees why the precision reduction is load-bearing.
- `storage_precision_timestamps_survive_a_database_round_trip` — the invariant the write path must hold.

Plus `log_timestamp_carries_no_sub_microsecond_digits` in `service.rs`, deliberately **not** `#[ignore]`d, so a regression on the write path is caught by `just test-unit` instead of only by Postgres-gated tests that normally never run.

## Compatibility

Rows written before this stay unverifiable — they always were — so there is no migration. An operator relying on an existing chain has to re-anchor.

## Relationship to #2620

#2620 proposes a shared `verify_entries` walk (anchoring, seq contiguity, tail-truncation detection) plus a `buzz-admin audit verify` command. Its Postgres-free unit tests build entries in memory and would pass regardless, but its `#[ignore]` Postgres tests and the operator command itself would fail on every real chain until this lands. Worth taking this first so that work has a verifiable baseline — the two changes don't overlap in code.

### DIFF
--- a/crates/buzz-audit/src/hash.rs
+++ b/crates/buzz-audit/src/hash.rs
@@ -29,8 +29,12 @@ pub fn to_storage_precision(created_at: DateTime<Utc>) -> DateTime<Utc> {
 /// `community_id` is hashed first so chain identity carries the tenant: an entry
 /// cannot be lifted out of one community's chain and re-verified inside another.
 ///
-/// `created_at` must already be at [`to_storage_precision`] — see that function
-/// for why sub-microsecond digits break verification.
+/// `created_at` is normalized through [`to_storage_precision`] here rather than
+/// hashed as given. Write paths truncate before storing so the row matches the
+/// in-memory entry, but normalizing again at the single point that consumes the
+/// value means no future caller can reintroduce the write/read preimage split
+/// by forgetting to. Values already at storage precision are unaffected —
+/// truncation is idempotent — so this does not change any digest.
 ///
 /// `detail` is serialized via [`canonical_json`] (sorted keys) so the hash is
 /// stable across machines and Rust versions. A serialization failure is a hard
@@ -40,7 +44,11 @@ pub fn compute_hash(entry: &AuditEntry) -> Result<[u8; 32], AuditError> {
     // Tenant binding: community_id leads the hash.
     hasher.update(entry.community_id.as_bytes());
     hasher.update(entry.seq.to_be_bytes());
-    hasher.update(entry.created_at.to_rfc3339().as_bytes());
+    hasher.update(
+        to_storage_precision(entry.created_at)
+            .to_rfc3339()
+            .as_bytes(),
+    );
     hasher.update(entry.action.as_str().as_bytes());
     match &entry.actor_pubkey {
         Some(pk) => {
@@ -157,19 +165,34 @@ mod tests {
     }
 
     #[test]
-    fn nanosecond_timestamps_cannot_survive_a_database_round_trip() {
-        // The trap `to_storage_precision` exists to close. `compute_hash`
-        // covers an RFC-3339 string whose sub-second digit count follows the
-        // value, so hashing a nanosecond `created_at` produces a digest that
-        // cannot be recomputed once Postgres has truncated it to microseconds:
-        // every entry would then fail `verify_chain` with `HashMismatch`.
+    fn rfc3339_sub_second_width_follows_the_value() {
+        // The underlying trap, pinned on the preimage rather than the digest:
+        // chrono emits 0/3/6/9 fractional digits depending on the value, so a
+        // nanosecond timestamp and its microsecond truncation are *different
+        // strings*. Hashing the untruncated value therefore produces a digest
+        // that cannot be recomputed from the stored row — which is what made
+        // every entry fail `verify_chain` with `HashMismatch`.
+        let ns = nanosecond_instant();
+        assert_eq!(ns.to_rfc3339(), "2023-11-14T22:13:20.123456789+00:00");
+        assert_eq!(
+            after_database_round_trip(ns).to_rfc3339(),
+            "2023-11-14T22:13:20.123456+00:00"
+        );
+        assert_ne!(ns.to_rfc3339(), after_database_round_trip(ns).to_rfc3339());
+    }
+
+    #[test]
+    fn compute_hash_normalizes_sub_microsecond_timestamps() {
+        // The enforcement point: even handed an untruncated `created_at`,
+        // `compute_hash` digests the storage-precision value, so a write path
+        // that forgot to truncate cannot split the write/read preimage.
         let ns = nanosecond_instant();
         let mut written = sample_entry();
         written.created_at = ns;
         let mut read_back = sample_entry();
         read_back.created_at = after_database_round_trip(ns);
 
-        assert_ne!(
+        assert_eq!(
             compute_hash(&written).unwrap(),
             compute_hash(&read_back).unwrap()
         );
--- a/crates/buzz-audit/src/hash.rs
+++ b/crates/buzz-audit/src/hash.rs
@@ -1,3 +1,4 @@
+use chrono::{DateTime, SubsecRound, Utc};
 use sha2::{Digest, Sha256};
 
 use crate::entry::AuditEntry;
@@ -7,11 +8,29 @@ use crate::error::AuditError;
 /// entry. Stored as `prev_hash = NULL`; hashed as all-zero bytes.
 pub const GENESIS_HASH: [u8; 32] = [0u8; 32];
 
+/// Reduce a timestamp to the precision the audit store round-trips.
+///
+/// `audit_log.created_at` is `TIMESTAMPTZ`, which Postgres keeps at microsecond
+/// resolution. [`compute_hash`] covers `created_at.to_rfc3339()`, and that
+/// string's sub-second digit count follows the value (chrono emits 0, 3, 6 or 9
+/// digits), so a timestamp carrying nanoseconds hashes to a digest that can
+/// never be recomputed from the stored row — the entry is written with one
+/// preimage and verified against another.
+///
+/// Every `created_at` must therefore pass through here *before* it is hashed
+/// and stored, so the in-memory entry and the row are byte-identical.
+pub fn to_storage_precision(created_at: DateTime<Utc>) -> DateTime<Utc> {
+    created_at.trunc_subsecs(6)
+}
+
 /// SHA-256 over the entry's identity, chain, and context fields.
 ///
 /// Field order is fixed — changing it invalidates all existing chains. The
 /// `community_id` is hashed first so chain identity carries the tenant: an entry
 /// cannot be lifted out of one community's chain and re-verified inside another.
+///
+/// `created_at` must already be at [`to_storage_precision`] — see that function
+/// for why sub-microsecond digits break verification.
 ///
 /// `detail` is serialized via [`canonical_json`] (sorted keys) so the hash is
 /// stable across machines and Rust versions. A serialization failure is a hard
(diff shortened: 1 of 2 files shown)


dophsquare:

Reviewed the diff. The root cause is real and the fix is correct: `Utc::now()` yields nanosecond resolution on Linux (`clock_gettime`), `compute_hash` covers `created_at.to_rfc3339()` whose sub-second digit count is value-dependent (chrono emits 0/3/6/9 digits), and Postgres `TIMESTAMPTZ` stores microseconds. So the write-time preimage ≠ the read-time preimage and `verify_chain` fails at entry #1 on untampered data — tamper-evidence is effectively off in any real deployment.

`to_storage_precision` → `trunc_subsecs(6)` before hash+store is the right normalization, and it's idempotent so a re-read row is stable. Test coverage is exactly what I'd want: the negative case (`nanosecond_timestamps_cannot_survive_a_database_round_trip`) proves the trap, and the positive case proves the invariant holds after truncation.

One thing worth confirming before merge: that **every** write path routes `created_at` through `to_storage_precision` (not just the one at `service.rs:103`) — a single unnormalized call site reintroduces the break. If there's a linked-in enforcement point (e.g. hashing only accepts already-truncated timestamps) that'd be even more robust. LGTM otherwise.

shani-singh1:
Both points addressed — the second one by taking your suggestion.

**Every write path.** Confirmed: there is exactly one. `log_inner` (`service.rs:112`) is the only place in the workspace that assigns `AuditEntry.created_at`. Every producer goes through `NewAuditEntry`, which deliberately carries no timestamp — the two external call sites (`buzz-relay/src/api/media.rs:426`, `buzz-relay/src/handlers/event.rs:558`) construct that struct, and its doc comment already states `seq`/`prev_hash`/`hash`/`created_at` are assigned by `AuditService::log`.

**The enforcement point.** You were right that convention isn't enough, so I did it: 08bcce9b moves the normalization into `compute_hash` itself, which is the single point that consumes the value.

```rust
hasher.update(
    to_storage_precision(entry.created_at)
        .to_rfc3339()
        .as_bytes(),
);
```

A future write path that forgets to truncate now can't split the preimage — write-time hashes `trunc(ts)`, read-time hashes `trunc(trunc(ts))`, and truncation is idempotent. The write-path truncation stays, because it carries a second invariant worth keeping: the `AuditEntry` returned from `log()` is byte-identical to the row a later read returns.

**Digest stability, verified rather than argued.** Since the write path already truncates, normalizing again should be a no-op on real chains — but "should be" isn't good enough for a hash function, so I checked it against Postgres: wrote a 3-entry chain with the *previous* commit, then verified those same rows with this one via a throwaway integration test.

```
verifying 3 pre-existing rows (seq 1..=3)
test preexisting_chain_still_verifies ... ok
```

So this commit is digest-compatible with the one before it — no re-anchoring needed on top of what the first commit already required.

**Test restructure.** `nanosecond_timestamps_cannot_survive_a_database_round_trip` had to go, since its `assert_ne!` is now false by construction. It's replaced by two tests that split the concern properly:

- `rfc3339_sub_second_width_follows_the_value` — pins the trap one level down, on the preimage itself (`…20.123456789+00:00` vs `…20.123456+00:00`). That's the property that actually causes the split, and it stays true regardless of what `compute_hash` does with it.
- `compute_hash_normalizes_sub_microsecond_timestamps` — pins the new guarantee: handed an untruncated `created_at`, `compute_hash` still produces the storage-precision digest.

Re-ran everything after the change: 13 unit tests pass without Postgres, 6/6 chain tests pass against Postgres 17, clippy `-D warnings` and `fmt --check` clean.

wpfleger96 (approved):
🤖 Walked through this one carefully and I'm happy with it. Verified the root cause at source: `service.rs` stamps `Utc::now()` at nanosecond precision while `hash.rs` digests `created_at.to_rfc3339()`, so the write-time preimage carries 9 fractional digits vs 6 after the Postgres µs round-trip — the write/verify split is real.

What I particularly like about this fix:

- It truncates at the write path *and* normalizes inside `compute_hash`, so a future caller that forgets to truncate can't reintroduce the divergence. And since `trunc_subsecs(6)` is idempotent, digests for existing rows are unchanged — I walked the legacy-row verification path and behavior is identical.
- The Postgres-free unit test (`log_timestamp_carries_no_sub_microsecond_digits`) directly guards the write path, so the fix itself is pinned by a test, not just the bug.

CI is fully green. Thanks for the thorough validation against real Postgres — approving.

PR #2651, opened 2026-07-24
Refine mobile channel and home UI

klopez4212:
## Summary
- Align mobile messages, system events, replies, dividers, timestamps, and mention chips with desktop
- Refine the channel header and Home spacing, icons, and typography
- Morph the Home plus button into a full-width quick-actions panel

## Validation
- `just mobile-check`
- `just mobile-test` (549 passed, 1 skipped)
- iOS simulator snapshots

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page/system_rows.dart
+++ b/mobile/lib/features/channels/channel_detail_page/system_rows.dart
@@ -2,6 +2,7 @@ part of '../channel_detail_page.dart';
 
 class _SystemMessageRow extends ConsumerWidget {
   final TimelineMessage message;
+  final List<TimelineMessage>? groupedMessages;
   final String channelId;
   final String? currentPubkey;
   final List<TimelineMessage>? allMessages;
@@ -10,6 +11,7 @@ class _SystemMessageRow extends ConsumerWidget {
 
   const _SystemMessageRow({
     required this.message,
+    this.groupedMessages,
     required this.channelId,
     this.currentPubkey,
     this.allMessages,
@@ -23,6 +25,11 @@ class _SystemMessageRow extends ConsumerWidget {
     if (systemEvent == null) return const SizedBox.shrink();
 
     final userCache = ref.watch(userCacheProvider);
+    final sourceMessages = groupedMessages ?? [message];
+    final groupedMembership = _membershipDisplayEvent(sourceMessages);
+    final channelCreator = systemEvent.type == SystemEventType.channelCreated
+        ? systemEvent.actorPubkey?.trim()
+        : null;
 
     String resolveLabel(String? pubkey) {
       if (pubkey == null) return 'Someone';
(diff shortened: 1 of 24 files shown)


klopez4212:
🤖 Mobile snapshots

### Channel timeline

Desktop-aligned message, system-event, reply, mention, and divider styling.

01-channel-timeline

### Home

Aligned Home typography, community avatar, and navigation content.

02-home-closed

### Quick actions

The plus morphs into a full-width quick-actions panel.

03-home-quick-actions

mobile/lib/features/channels/channel_detail_page/system_rows.dart
```diff
@@ -52,32 +86,45 @@ class _SystemMessageRow extends ConsumerWidget {
...
-                ),
-              ],
-            ),
-            if (message.reactions.isNotEmpty)
+                  _messageTimestamp(context, message.createdAt),
+                ],
+              ),
+            if (reactions.isNotEmpty)
               Padding(
                 padding: const EdgeInsets.only(left: 28),
```
square-tomb: 🤖 Non-blocking: this `left: 28` inset predates the avatar bump to 36px (+ `Grid.xxs` gap), so grouped-row reactions no longer align with the message-style system content above them. Consider `36 + Grid.xxs` to match the new avatar column.

mobile/lib/features/channels/date_formatters.dart
```diff
@@ -62,20 +64,48 @@ String relativeTime(int unixSeconds) {
...
+  return switch (day % 10) {
+    1 => 'st',
+    2 => 'nd',
+    3 => 'rd',
+    _ => 'th',
+  };
+}
+
+/// Desktop-parity message clock time, e.g. "2:34 PM".
+String formatMessageTime(int unixSeconds) {
```
square-tomb, line 105: 🤖 Non-blocking: the old `formatMessageTime` included an `M/D` prefix for messages older than a day. The channel timeline still gets date context from `DayDivider`, but `thread_detail_page.dart` has no day dividers, so an old thread reply now renders only a clock time with no date anywhere on screen. Desktop has the same clock-only label but recovers via the hover tooltip (`formatFullDateTime`); mobile has no equivalent affordance. Worth a follow-up (e.g. long-press tooltip or day dividers in the thread view).

square-tomb (approved):

mobile/lib/features/channels/channel_detail_page/system_rows.dart
klopez4212: 🤖 Addressed in 4deb38ab8: grouped system-row reactions now use the 36px avatar column plus the Grid.xxs gap, with widget coverage for the alignment.

mobile/lib/features/channels/date_formatters.dart
klopez4212: 🤖 Addressed in 4deb38ab8: the thread view now renders day dividers for the head and at reply day boundaries, with widget coverage for replies spanning multiple days.

brow (approved):

PR #2653, opened 2026-07-24
feat(desktop): autocomplete ACP slash commands

### LINKED ISSUE #2528: feat(desktop): slash-command autocomplete in composer from ACP available_commands_update (follow-up to #919)
## Problem

Agents already advertise their slash commands over ACP, but Buzz Desktop users have no way to discover them — typing `/` in the composer shows nothing, so you have to memorize and type full command names (`/ce-commit-push-pr`, `/ffrc-slide-forge`, …) by hand. For users whose agents load large skill libraries (hundreds of Claude Code skills/plugins), this makes the shipped slash pass-through hard to use in practice.

## Background — plumbing already exists

#919 (merged 2026-06-09) shipped slash-command pass-through to ACP connectors and explicitly deferred this piece:

> **No UI discovery/autocomplete yet** — `available_commands_update` is logged; persisting + surfacing it to the desktop composer is a follow-up.

So today `acp.rs` receives each agent's command list via the ACP `available_commands_update` session update and drops it after logging. This issue tracks the follow-up: persist that list and feed it to a composer autocomplete.

## Proposed UX

- Typing `/` at the start of the message — or after a leading `@AgentName` mention — opens a fuzzy-filterable dropdown, styled like the existing `@`-mention autocomplete.
- The list shows the commands the mentioned agent last advertised (name + one-line description from `available_commands_update`). With no mention yet, fall back to the union of commands from agents in the channel, grouped by agent.
- Enter/Tab inserts the command; typing continues to filter. Escape dismisses and leaves the literal `/` text.
- Commands persist per agent (last-received `available_commands_update`), so the list is available even before the agent's session is warm; refresh on each new update.

## Non-goals

- No command execution semantics in Buzz — #919's contract stands: connectors own parsing and execution; Desktop only helps compose the text.
- No new protocol work — the ACP update already carries name + description.

## Prior art

- Existing `@`-mention autocomplete in the composer (interaction pattern to reuse).
- Zed's ACP client surfaces `available_commands_update` as composer completions — same protocol, same shape.

kevinmanase:
## summary

adds slash-command autocomplete for commands advertised by acp agents. typing `/` at the start of a message groups suggestions by agent; selecting one inserts its exact mention and command. typing after leading agent mentions filters to those agents and preserves the existing mentions.

rebased onto current `main` and adapted to the current observer batching, composer focus handling, and autocomplete overlays. the follow-up pass reuses the shared mention parser, isolates catalogs by community and owner, rejects control/bidi characters, and respects code blocks and ime input. no file-size override is needed.

## behavior and scope

- catalogs survive restarts, accept newer snapshots, and honor empty updates as removals.
- only agents owned by the current user populate catalogs, since observer frames are encrypted to their owner.
- acp supplies command names and descriptions without a built-in/skill category. command execution and send-time freshness remain the connector's responsibility.
- multiple leading mentions stay addressed when inserting a command.

## testing

- desktop unit suite: 6,512 passed; final focused regression suite: 55 passed.
- browser tests: 2 passed, including persistence, keyboard and mouse selection, filtering, escape, and the exact recipient on the signed message.
- full acp suite passed: 931 unit tests and 9 lifecycle tests, including snapshot coverage for empty and malformed updates.
- native desktop unit suite: 3,173 passed (19 ignored); mobile suite: 2,098 passed.
- full local `CHECK_FILE_SIZES_BASE=upstream/main just ci`: passed. the explicit base avoids comparing against the fork's older `origin/main`. desktop typecheck also passed.
- github ci and the security review need maintainer authorization for this head.

fixes #2528

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -1571,16 +1571,25 @@ impl AcpClient {
             }
             "available_commands_update" => {
                 // Advertised slash commands (ACP slash-commands extension).
-                // Logged for observability; UI surfacing is a follow-up.
-                let names: Vec<&str> = update["availableCommands"]
+                // Forward the complete latest list through the encrypted observer
+                // stream so Desktop can offer commands for the originating agent.
+                let commands = update["availableCommands"]
                     .as_array()
-                    .map(|cmds| cmds.iter().filter_map(|c| c["name"].as_str()).collect())
+                    .cloned()
                     .unwrap_or_default();
+                let names: Vec<&str> = commands
+                    .iter()
+                    .filter_map(|command| command["name"].as_str())
+                    .collect();
                 tracing::info!(
                     target: "acp::update",
                     "available_commands_update: {} commands [{}]",
                     names.len(),
                     names.join(", ")
+                );
+                self.observe(
+                    "available_commands_captured",
+                    serde_json::json!({ "commands": commands }),
                 );
                 false
             }
@@ -3088,6 +3097,43 @@ mod tests {
                 },
             }
         })
+    }
+
+    #[tokio::test]
+    async fn available_commands_update_emits_complete_semantic_snapshot() {
+        let mut client = spawn_inert_client().await;
+        let observer = ObserverHandle::in_process();
+        client.set_observer(Some(observer.clone()), 3);
+
+        let msg = serde_json::json!({
+            "jsonrpc": "2.0",
+            "method": "session/update",
+            "params": {
+                "sessionId": "test-session",
+                "update": {
+                    "sessionUpdate": "available_commands_update",
+                    "availableCommands": [
+                        { "name": "review", "description": "Review changes" },
+                        { "name": "deploy", "description": "Ship it" }
+                    ]
+                }
+            }
+        });
+        let _ = client.handle_session_update(&msg);
+
+        let events = observer.snapshot();
+        assert_eq!(events.len(), 1);
+        assert_eq!(events[0].kind, "available_commands_captured");
+        assert_eq!(events[0].agent_index, Some(3));
+        assert_eq!(
+            events[0].payload,
+            serde_json::json!({
+                "commands": [
+                    { "name": "review", "description": "Review changes" },
+                    { "name": "deploy", "description": "Ship it" }
+                ]
+            })
+        );
     }
 
     #[tokio::test]
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -1792,16 +1792,25 @@ impl AcpClient {
             }
             "available_commands_update" => {
                 // Advertised slash commands (ACP slash-commands extension).
-                // Logged for observability; UI surfacing is a follow-up.
-                let names: Vec<&str> = update["availableCommands"]
+                // Forward the complete latest list through the encrypted observer
+                // stream so Desktop can offer commands for the originating agent.
+                let commands = update["availableCommands"]
                     .as_array()
-                    .map(|cmds| cmds.iter().filter_map(|c| c["name"].as_str()).collect())
+                    .cloned()
                     .unwrap_or_default();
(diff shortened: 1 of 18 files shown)


rshiozaki:
We'd like to see this land, so we did a full review of this PR against today's `main` (22cdda441) and revived the branch. TL;DR: **the implementation is solid, and we've prepared a conflict-free, test-green rebase** that @kevinmanase or maintainers are welcome to take over: https://github.com/rshiozaki/buzz/tree/rebase/acp-slash-autocomplete

### Review findings

Strengths:
- **Persistence design is right.** `available_commands_captured` observer frames feed a per-owner+agent localStorage catalog; both the live and archive ingestion paths record it, and restart survival is covered by tests.
- **Scoping avoids the union-menu trap.** With no mention typed, suggestions group per agent rather than merging into one ambiguous list; selecting inserts `@Agent /cmd` and registers the exact pubkey; leading mentions narrow the list.
- **Advertised data is treated as untrusted**: name ≤128 chars (no whitespace), description ≤512, ≤256 commands per agent, deduped, re-sanitized on hydrate, plain-text rendering only.
- **Staleness handling**: newer `(ts, seq)` wins, and an empty `available_commands_update` is honored as an authoritative removal (tested).

Gaps worth follow-ups (not blockers in our view):
- No visual distinction between built-in commands and discovered skills — the catalog is a flat `{name, description}`.
- A command that has gone stale by send time fails silently; there's no send-time revalidation or freshness hint in the UI.
- Sanitization bounds lengths/counts but doesn't strip control or bidi characters.
- One undisclosed limitation worth stating in the PR body: catalogs only populate for agents the **current user owns** (observer frames are encrypted to the owner), so a teammate's agent shows no commands. Reasonable for v1, but the longer-term shape probably wants a signed per-agent advertisement event any channel member can read — happy to write that up as a design issue if there's interest.
- With two or more leading mentions, an inserted `/cmd` addresses all mentioned agents rather than one.

### What the rebased branch adds

All feature credit to @kevinmanase; we started from the conflict resolutions in #3537 (thanks to its author too) and brought it the rest of the way:

- **Rebased onto today's main** — the original branch was 687 commits behind and merge-dirty; three conflict sites resolved (`MessageComposer.tsx` composer-dock changes, `localStorageQuota.ts` prefixes, `check-file-sizes.mjs`).
- **File-size gate satisfied without an override.** The feature pushed `MessageComposer.tsx` to 1,047 lines; the original commit slipped under the 1,000-line gate via an override mechanism that has since been removed from the repo. We extracted the shared autocomplete insertion path and picker keydown chain into a sibling hook (`useComposerAutocompleteInsertions.ts`), following the existing composer hook pattern — the composer is now 968 lines, below where `main` already is (999).
- **Green on today's main**: full desktop unit suite (5,578 tests, 0 failures), `tsc --noEmit`, biome, and the agent-harness Rust tests including this feature's snapshot test.

Related work, for cross-reference: this implements #2528 and also covers the ask in #5741; #6567 builds the picker-chip UI for the same space but without persistence or the typed-`/` path.

Happy to open the rebased branch as a fresh PR, or hand it off however is easiest — whichever the maintainers prefer.

kevinmanase:
thanks for the review @rshiozaki. i've rebased onto current main, resolved the conflicts, and pushed another simplification and review pass here. i'm continuing the work in this original pr, so let's keep the review here.

fixed the control/bidi gap, community scoping, and duplicate-name recipient handling. added browser coverage for selecting a command and checking who the signed message addresses. the owner-only catalog limitation and multiple-mention behavior are now in the description.

full local ci passed against `upstream/main`, including 6,512 desktop tests and 2,098 mobile tests. all 55 focused tests and both browser tests passed too.

@wesbillman @wpfleger96 @klopez4212 would appreciate another look when you get a chance. could one of you approve the [fork ci run](https://github.com/block/buzz/actions/runs/34406635265) and authorize the current head using the [security-review bot's instructions](https://github.com/block/buzz/pull/2653#issuecomment-5608941680)?

PR #2658, opened 2026-07-24
fix(onboarding): show real install errors and fix concurrent install state

wpfleger96:
Fixes two bugs in the ACP runtime setup screen (onboarding) and Doctor settings panel that caused users to see unhelpful generic errors on install failure and lose the first install's state when starting two installs back-to-back.

## What this fixes

**Bug A — real install errors discarded in onboarding.** `RuntimeCard` received the detailed install error string from `getInstallErrorMessage` (step name, stderr, exit code, actionable hint) but rendered a hardcoded `"Installation couldn't be completed. Try again."` tooltip regardless of the actual failure. The rich string was computed and stored in `installResults` state, then ignored at render. Fix: pass `installError` as `detail` to `RuntimeErrorTooltip`. Add `whitespace-pre-line` to the tooltip content span so multi-line hint + step-detail strings render correctly.

Users on Windows with no Node.js will now see the actual Git Bash error and the actionable hint instead of a generic message. Users with npm permission issues on macOS see the `EACCES` detail and the `npm config set prefix` remediation hint.

**Bug B — concurrent installs lose the first card's state.** `RuntimeProvidersSection` (onboarding) and `DoctorSettingsPanel` each shared one `useInstallAcpRuntimeMutation()` instance across all runtime cards/rows. react-query v5 per-`mutate()` callbacks (`onSuccess`/`onError`) only fire for the latest `mutate()` call on a shared instance — starting install B silently drops install A's result callback, so card A never shows success or failure. Additionally, `isPending` and `mutation.variables` tracked only the latest call, making card A's spinner vanish mid-install when card B started.

Fix: move `useInstallAcpRuntimeMutation()` into each `RuntimeCard` (onboarding) and `RuntimeRow` (Doctor panel) so every card owns its own mutation instance with independent `isPending` and callbacks. The shared `installMutation` + `handleInstall` in `RuntimeProvidersSection` and the lifted `installResults`/`installingIds`/`handleInstall` in `DoctorSettingsPanel` are removed.

## Non-happy-path scenarios

- **Install A fails while B is in-flight**: each card has its own mutation instance; A's `onError` fires independently, A's card shows the error, B's spinner continues unaffected.
- **Both fail**: each card's `onError` fires separately, both show their own error detail.
- **A succeeds while B fails**: A's card shows success (runtime query invalidation triggers re-check → READY), B's card shows its error detail.

The `RETRY INSTALL` label flow (`installError → installLabel` in `RuntimeStatus`) is unchanged — retrying a card's install clears its prior result and re-fires its own mutation.

## Files changed

- `desktop/src/features/onboarding/ui/SetupStep.tsx` — `RuntimeCard` owns its mutation; `RuntimeProvidersSection` shared mutation removed
- `desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx` — `whitespace-pre-line` on tooltip content
- `desktop/src/features/settings/ui/DoctorSettingsPanel.tsx` — `RuntimeRow` owns its mutation + result state; section-level lifted state removed
- `desktop/src/shared/lib/installError.test.mjs` — two new test cases: double-newline separator format and multi-step last-step-only behavior

### DIFF
--- a/desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx
+++ b/desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx
@@ -42,7 +42,7 @@ export function RuntimeErrorTooltip({
         side="bottom"
         sideOffset={12}
       >
-        <span className="leading-4">{detail}</span>
+        <span className="whitespace-pre-line leading-4">{detail}</span>
       </TooltipContent>
     </Tooltip>
   );
--- a/desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx
+++ b/desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx
@@ -42,7 +42,9 @@ export function RuntimeErrorTooltip({
         side="bottom"
         sideOffset={12}
       >
-        <span className="whitespace-pre-line leading-4">{detail}</span>
+        <span className="block max-h-48 overflow-y-auto break-words whitespace-pre-line leading-4">
+          {detail}
+        </span>
       </TooltipContent>
     </Tooltip>
   );
--- a/desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx
+++ b/desktop/src/features/onboarding/ui/RuntimeErrorTooltip.tsx
@@ -42,7 +42,7 @@ export function RuntimeErrorTooltip({
         side="bottom"
         sideOffset={12}
       >
-        <span className="block max-h-48 overflow-y-auto break-words whitespace-pre-line leading-4">
+        <span className="block max-h-48 overflow-y-auto overflow-x-hidden break-words whitespace-pre-line leading-4">
           {detail}
         </span>
       </TooltipContent>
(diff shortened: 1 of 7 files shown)


wpfleger96:
## UI states — PR #2658

### Onboarding: rich install-error tooltip
Bounded, scrollable dark tooltip showing the real npm EACCES stderr; long Windows path wraps instead of overflowing horizontally.
onboarding-error-tooltip

### Onboarding: concurrent installs — independent card states
Both Claude Code and Codex installs in flight simultaneously; each card tracks its own state independently (no last-caller-wins drop).
onboarding-concurrent-installs

### Doctor: row error — real step failure detail
Doctor panel Codex row after a failed install, showing the actual step stderr ("npm ERR! code E404 / npm ERR! 404 Not Found") instead of the generic message.
doctor-row-error

### Doctor: Check again — stale error cleared
Same Claude Code row after clicking "Check again": the stale install error is cleared via resetEpoch, row returns to healthy state.
doctor-check-again-cleared

PR #2664, opened 2026-07-24
fix(buzz-acp): stop cancel storm on steer-incapable runtimes

Foreist:
## Problem

When an ACP runtime cannot do goose-native steer (e.g. `@agentclientprotocol/claude-agent-acp`, which never emits `_meta.goose.activeRunId`), every incoming mention during an in-flight turn hits the `ExpectedRunIdMissing` fallback and takes the `cancel+merge` path. Rapid successive mentions therefore cancel the in-flight response before it can be sent -- the agent goes silent indefinitely. Observed as ~26 min of dropped responses under fast repeated mentions.

Evidence (harness log, every mention):
```
WARN steer: no active_run_id at write time -- falling back to cancel+merge
     ack=Ok(Err(ExpectedRunIdMissing))  signal_fallback=true
     mode=Steer
```

## Fix

Split the steer-ack disposition:
- `Transport` (real send failure) -> keep existing cancel+merge.
- `ExpectedRunIdMissing` (runtime structurally cannot steer) -> **do not cancel.** Let the in-flight turn finish and deliver; route the new mention to the next turn. Removes the cancel storm.

Mapping extracted into pure fn `steer_ack_disposition`, unit-tested across all 6 ack variants.

## Verification
```
cargo test -p buzz-acp   -> 587 passed, 0 failed (6 new)
cargo clippy -p buzz-acp -> no warnings
cargo fmt --check        -> clean
```

Note: logic verified by unit tests; live fast-mention repro not yet run. Recommend a rapid-mention smoke test on Desktop before merge.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -1118,6 +1118,142 @@ struct SteerAckEvent {
     ack: std::result::Result<pool::SteerAck, tokio::sync::oneshot::error::RecvError>,
 }
 
+/// What the main loop should do with a withheld event once a goose-native
+/// steer attempt resolves.
+///
+/// - `release_withheld`: release the withheld event back to the queue so
+///   normal dispatch can redeliver it.
+/// - `drop_withheld`: remove the withheld event entirely (the steer already
+///   delivered it).
+/// - `signal_fallback`: fire the universal cancel+merge `ControlSignal::Steer`
+///   against the in-flight task so the message still reaches the agent.
+struct SteerDisposition {
+    release_withheld: bool,
+    drop_withheld: bool,
+    signal_fallback: bool,
+}
+
+/// Map a resolved steer ack to its [`SteerDisposition`]. Pure — the main loop's
+/// `PoolEvent::SteerAck` arm applies the result. Extracted so the delivery
+/// semantics are unit-testable without driving the whole event loop; see the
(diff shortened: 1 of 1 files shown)


HashMac:
Independent macOS follow-up with Hermes (runtime advertised `steering_supported=false`). I previously added the direct `ExpectedRunIdMissing` reproduction to #4189; a recovery run exposed the remaining boundary in this PR's “let the in-flight turn finish” rule.

After a provider-auth interruption was repaired, the existing Hermes turn stayed alive in a repeated polling loop. A fresh owner message asking it to finalize was accepted and queued, but never became a new prompt because the old turn did not finish. Restarting the runtime was the only recovery; the fresh instruction then ran and the mission completed.

So this PR stops the cancel storm, but a non-terminating or stale in-flight turn can still hold the released event indefinitely—the same user-visible symptom described in #4434.

Suggested acceptance coverage:

1. Start a non-steerable runtime with an in-flight turn that never completes.
2. Send a second event.
3. Do not extend the total in-flight lease on each steer.
4. After a bounded deadline, mark the stale channel complete and dispatch the withheld event as a fresh prompt exactly once.
5. Cover restart/reconnect: the pending event survives and runs once, while an already accepted side effect is not repeated.

This may be a follow-up rather than a blocker, but linking #4434 and #4189 from this PR would make the remaining reliability boundary explicit.

PR #2668, opened 2026-07-24
feat(desktop): add copy raw event JSON action to message menu

iamraydoan:
## Summary

Add a **"Copy raw event JSON"** action to the Desktop message action bar (`MessageActionBar`) dropdown menu, allowing users to copy formatted Nostr Event JSON payloads for debugging.

## How to Test

1. Launch Desktop app (`just dev`).
2. Hover over a message in `#general` and click **`...`** (*More actions*).
3. Click **"Copy raw event JSON"** and paste to verify the copied JSON structure.

## Checklist

- [x] `just ci` & `pnpm check` pass
- [x] Unit tests pass (`pnpm desktop test`: 3,455 passed, `just test-unit`)
- [x] No `unwrap()` / `unsafe` added
- [x] Zoom-safe rem typography used

### DIFF
--- a/desktop/src/features/messages/ui/MessageActionBar.tsx
+++ b/desktop/src/features/messages/ui/MessageActionBar.tsx
@@ -2,6 +2,7 @@ import {
   BellOff,
   BellRing,
   Clock,
+  Code,
   Copy,
   CornerUpLeft,
   EllipsisVertical,
@@ -208,6 +209,29 @@ function MoreActionsMenu({
             >
               <Copy className="h-4 w-4" />
               Copy message
+            </DropdownMenuItem>
+          ) : null}
+
+          {hasCopyActions ? (
+            <DropdownMenuItem
+              data-testid={`copy-raw-event-${message.id}`}
+              onClick={() => {
+                const rawEvent = {
+                  id: message.id,
+                  pubkey: message.signerPubkey ?? message.pubkey ?? "",
+                  created_at: message.createdAt,
+                  kind: message.kind ?? 9,
+                  tags: message.tags ?? [],
+                  content: message.body,
+                };
+                copyTextToClipboard(
+                  JSON.stringify(rawEvent, null, 2),
+                  "Raw event JSON copied to clipboard",
+                );
+              }}
+            >
+              <Code className="h-4 w-4" />
+              Copy raw event JSON
             </DropdownMenuItem>
           ) : null}
 


wesbillman (changes requested):
Reviewed on behalf of Wes.

Thanks for putting this together — the underlying goal of making Nostr events easier to inspect during debugging is useful.

I’m requesting changes because this action is presented as **“Copy raw event JSON,”** but the value is reconstructed from `TimelineMessage` rather than copied from the canonical signed relay event. It omits `sig`, introduces fallback values, and can diverge further for edited messages: the timeline model may combine the original event ID/timestamp/kind with the latest edit body and overlaid media tags. That can produce a synthetic object that was never signed or relayed.

This also puts a niche debugging action in the standard More Actions menu for every delivered message, including ordinary channel and inbox rows. We don’t think that belongs in the general product surface.

Our recommendation is to close this PR in its current form. If we pursue the workflow later, a developer/debug inspector would be a better home, and it should look up and export the canonical event—including its signature—with coverage proving exact source-event fidelity for edited messages.

Again, thank you for identifying the debugging need and taking the time to contribute. The concern is with the product surface and event fidelity, not the intent behind the change.

iamraydoan:
> Reviewed on behalf of Wes.
> 
> Thanks for putting this together — the underlying goal of making Nostr events easier to inspect during debugging is useful.
> 
> I’m requesting changes because this action is presented as **“Copy raw event JSON,”** but the value is reconstructed from `TimelineMessage` rather than copied from the canonical signed relay event. It omits `sig`, introduces fallback values, and can diverge further for edited messages: the timeline model may combine the original event ID/timestamp/kind with the latest edit body and overlaid media tags. That can produce a synthetic object that was never signed or relayed.
> 
> This also puts a niche debugging action in the standard More Actions menu for every delivered message, including ordinary channel and inbox rows. We don’t think that belongs in the general product surface.
> 
> Our recommendation is to close this PR in its current form. If we pursue the workflow later, a developer/debug inspector would be a better home, and it should look up and export the canonical event—including its signature—with coverage proving exact source-event fidelity for edited messages.
> 
> Again, thank you for identifying the debugging need and taking the time to contribute. The concern is with the product surface and event fidelity, not the intent behind the change.

Thanks for the thoughtful review and the detailed explanation.

I agree with the concerns regarding both the event fidelity and the placement of this feature in the standard message actions menu. Looking back, presenting the reconstructed `TimelineMessage` as the raw event could indeed be misleading.

I appreciate the feedback and the guidance on the preferred direction. I'll close this PR in its current form. Thanks again for taking the time to review it.

PR #2670, opened 2026-07-24
fix(relay): decompress gzip-encoded git smart-HTTP request bodies

iamkaaldhairya:
## Problem

Cloning or fetching a relay-hosted repository with **many refs** (hundreds of branches/tags) fails. The client sees only a mid-stream:

```
fatal: the remote end hung up unexpectedly
```

Small repositories clone fine; the failure is strongly correlated with ref count.

## Root cause

git's smart-HTTP client **gzip-compresses** the `git-upload-pack` / `git-receive-pack` request body once it exceeds an internal size threshold. That threshold is driven by the number of want/have lines, so it reliably trips on many-ref clones and stays untripped on small ones.

The relay pipes the request body directly into the git subprocess's stdin, so a still-compressed body reaches `git upload-pack` as raw gzip. git then fails parsing pkt-lines with `fatal: protocol error: bad line length character: <binary>`. The subprocess stderr is discarded (`Stdio::null()`), so the only client-visible signal is the truncated stream — which is why this is hard to diagnose.

Reproduction: create a repo with a few hundred tags, push it, then `git clone` over HTTP. The POST to `…/git-upload-pack` carries `Content-Encoding: gzip` and the clone dies mid-stream.

## Fix

Add `decode_git_request_body`, which transparently inflates the request body when `Content-Encoding: gzip` (or `x-gzip`) is present, applied in both `upload_pack` and `receive_pack` before the body is streamed to the subprocess.

- Identity-encoded bodies pass through **byte-for-byte** — the common small-clone path is unchanged.
- Only gzip is decoded (the sole encoding git emits). An unknown non-identity encoding is passed through rather than rejected, letting the subprocess surface any real mismatch as an in-band protocol error.

Uses `async-compression` (already in the dependency tree) for streaming inflation, so the body is not buffered into memory.

## Tests

Adds unit tests for both paths (gzip inflate, identity passthrough). Full `api::git::transport` suite passes (24 tests). Verified end-to-end against a live relay: a repository with 2,140 refs that previously failed now clones completely with a valid working tree.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1121,6 +1121,7 @@ name = "buzz-relay"
 version = "0.2.0"
 dependencies = [
  "anyhow",
+ "async-compression",
  "async-trait",
  "axum",
  "base64",
@@ -1139,6 +1140,7 @@ dependencies = [
  "chrono",
  "dashmap",
  "deadpool-redis",
+ "flate2",
  "futures",
  "futures-util",
  "hex",
--- a/crates/buzz-relay/Cargo.toml
+++ b/crates/buzz-relay/Cargo.toml
@@ -78,6 +78,7 @@ metrics         = { workspace = true }
 metrics-exporter-prometheus = { workspace = true }
 metrics-util    = { workspace = true }
 pulldown-cmark = { version = "0.13.4", default-features = false, features = ["html"] }
+async-compression = { version = "0.4.42", features = ["tokio", "gzip"] }
 
 [features]
 dev = ["buzz-auth/dev"]
@@ -90,3 +91,4 @@ buzz-auth = { workspace = true, features = ["dev"] }
 reqwest = { workspace = true }
 tokio-tungstenite = { workspace = true }
 futures = "0.3"
+flate2 = "1.1.9"
--- a/crates/buzz-relay/src/api/git/transport.rs
+++ b/crates/buzz-relay/src/api/git/transport.rs
@@ -714,6 +714,40 @@ async fn info_refs_subprocess(
         .unwrap())
 }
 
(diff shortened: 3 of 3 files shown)


tlongwell-block:
Pushed review follow-ups as a new commit 537e03f82 (no history rewrite):

- **Decoded-byte cap** in `decode_git_request_body`: the router's `RequestBodyLimitLayer` only bounds *compressed* bytes, so a small gzip bomb could feed an effectively unbounded stream to the subprocess (unbounded scratch-disk writes on the receive-pack path). The seam now takes `max_decoded_bytes` — `git_max_pack_bytes` for receive-pack, a 64 MiB negotiation cap for upload-pack — and errors the stream past the cap.
- **Decode errors surfaced**: both subprocess stdin pumps now `warn!` on body/decode stream errors instead of silently breaking, so malformed gzip / cap overflow leaves a server-side signal.
- **Clippy**: `std::io::Error::other` (repo gate runs `-D warnings`).
- **Tests**: gzip-bomb cut off at cap, exactly-at-cap passthrough, malformed gzip errors.

Verified at 537e03f82: `cargo test -p buzz-relay` 739 passed / 1 failed (pre-existing `mesh_demo` flake, red on main too) / 27 ignored; clippy `-D warnings` and fmt clean.

Note for @iamkaaldhairya: the DCO check needs a `Signed-off-by` on your original commit 9cae1fecc — my follow-up commit carries its own signoff but can't fix yours.

tlongwell-block (approved):

tlongwell-block:
Thanks for the contribution, @iamkaaldhairya 👋

PR #2676, opened 2026-07-24
fix: reduce default parallelism to 1, use lazy pool on manual start

### LINKED ISSUE #2631: [Bug] Default parallelism 24 eagerly spawns large Codex ACP pools, consuming ~14 GB and flooding Recents
## Summary

Buzz Desktop 0.4.23 defaults every managed local agent to `parallelism: 24` and manual agent starts eagerly initialize the full pool. With Codex ACP, each worker is a multi-process adapter/app-server chain, so a few agents can create hundreds of idle processes, consume double-digit GiB of RAM, and populate Codex Recents with `[Base] You are operating inside the Buzz platform...` sessions.

This initially looked like an orphan cleanup leak, but controlled shutdown and source inspection show that cleanup works. The problem is the default pool size plus eager manual startup.

## Environment

- Buzz Desktop: 0.4.23
- macOS: 26.5.2 (25F84)
- Architecture: arm64
- Hardware: Apple M4 Max
- Agent runtime: bundled `@agentclientprotocol/codex-acp`
- Model: `gpt-5.6-sol`

## Reproduction

1. Create three local Codex-backed agents using the default parallelism.
2. Create one additional local agent with parallelism set to 1.
3. Start all four agents from Buzz Desktop.
4. Inspect the Buzz-owned process tree and local Codex Recents/state.

## Actual behavior

The four agents were configured as `24 + 24 + 24 + 1 = 73` worker slots.

A live process snapshot showed:

- 4 `buzz-acp` roots owned by `buzz-desktop`
- approximately 220 `codex-acp` / Codex app-server descendants
- 226 total Buzz-related processes
- 14,359.8 MiB combined RSS
- approximately 0% combined CPU while idle

The process count matches the configuration: 73 worker slots, with roughly three OS processes per Codex ACP worker.

The local Codex state database also contained 40 visible threads titled `[Base] You are operating inside the Buzz platform...`:

- all 40 had `has_user_event = 0`
- 17,721,356 cumulative tokens were recorded across those threads
- threads were created across repeated Buzz agent starts over three days

The token field is reported as supporting evidence of substantial session activity; actual billed usage may differ due to caching/accounting semantics.

Quitting Buzz gracefully removed all 226 processes. Reopening Buzz without starting agents produced one `buzz-desktop` process at approximately 146 MiB. This confirms the process-tree shutdown path works and this is not an orphan-process report.

## Source-level cause

At tag `v0.4.23`:

- `desktop/src-tauri/src/managed_agents/types.rs` sets `DEFAULT_AGENT_PARALLELISM` to 24.
- `desktop/src-tauri/src/managed_agents/runtime.rs` exports that value as `BUZZ_ACP_AGENTS`.
- `start_managed_agent_process` calls `spawn_agent_child(..., false, ...)`, disabling lazy-pool startup for the manual start path.
- `crates/buzz-acp/src/lib.rs` immediately calls `initialize_agent_pool(...)` when `lazy_pool` is false.
- In contrast, launch-time restore already uses `spawn_agent_child(..., true, ...)`, with a comment explaining that eager restore would create “N idle brains.”

Relevant code:

- https://github.com/block/buzz/blob/v0.4.23/desktop/src-tauri/src/managed_agents/types.rs#L729-L735
- https://github.com/block/buzz/blob/v0.4.23/desktop/src-tauri/src/managed_agents/runtime.rs#L1840-L1850
- https://github.com/block/buzz/blob/v0.4.23/desktop/src-tauri/src/managed_agents/runtime.rs#L2100-L2125
- https://github.com/block/buzz/blob/v0.4.23/desktop/src-tauri/src/managed_agents/restore.rs#L310-L335
- https://github.com/block/buzz/blob/v0.4.23/crates/buzz-acp/src/lib.rs#L1260-L1280

## Expected behavior

Starting a normally configured local agent should not eagerly allocate 24 model-adapter workers or create user-visible Codex sessions before work arrives.

## Suggested fix

1. Change the desktop default parallelism from 24 to 1.
2. Use lazy-pool startup for manual starts, matching launch-time restore.
3. Avoid creating visible Codex Recents entries for idle ACP workers, or archive/reuse them as internal sessions.
4. Make higher parallelism an explicit advanced setting with a clear resource-cost warning.
5. Add a regression test proving that manual start with default settings does not spawn the full worker pool until the first event arrives.

Even if parallelism 24 remains available for high-throughput deployments, it is unsafe as the default for a desktop installation using Codex ACP.

Christian-Sidak:
## Summary

- Change `DEFAULT_AGENT_PARALLELISM` from 24 to 1 in `types.rs`
- Use lazy-pool startup (`lazy: true`) in `start_managed_agent_process` to match launch-time restore behavior

## Problem

With the default parallelism of 24, starting agents eagerly initializes the full worker pool via `spawn_agent_child(..., false, ...)`. With Codex ACP backends, each worker spawns multiple processes (adapter + app-server chain), resulting in ~220 idle processes and ~14 GB RSS for just four agents -- before any work arrives.

The launch-time restore path already uses `lazy: true` with an explicit comment ("eager restore... silently reintroduces N idle brains on every launch"). The manual-start path lacked the same protection.

## Changes

- `types.rs`: `DEFAULT_AGENT_PARALLELISM` 24 → 1. Desktop installations should default to a single worker; higher parallelism remains available as an explicit setting.
- `runtime.rs`: `start_managed_agent_process` now passes `lazy: true` to `spawn_agent_child`, deferring pool initialization until the first event arrives.
- `types/tests.rs`: two regression tests pinning `DEFAULT_AGENT_PARALLELISM == 1` and verifying legacy records without an explicit parallelism field deserialize to 1.

## Test plan

- [x] `default_agent_parallelism_is_one` -- asserts `DEFAULT_AGENT_PARALLELISM == 1`
- [x] `managed_agent_record_without_parallelism_defaults_to_one` -- legacy JSON without a `parallelism` key deserializes to 1
- [x] Verified `BUZZ_ACP_LAZY_POOL=true` is already used by `spawn_agent_child` when `lazy=true` (runtime.rs line 1722), so the env-var path is covered by the existing harness

Fixes #2631

### DIFF
--- a/desktop/src-tauri/src/managed_agents/runtime.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime.rs
@@ -2127,7 +2127,11 @@ pub fn start_managed_agent_process(
     // Scalar PIDs are migration-only and never establish pair liveness.
     record.runtime_pid = None;
 
-    let mut process = spawn_agent_child(app, record, &key.relay_url, false, owner_hex)?;
+    // Use lazy-pool startup for manual starts, matching launch-time restore.
+    // Eager startup (false) would immediately initialize the full worker pool,
+    // consuming large amounts of RAM and creating visible idle Codex sessions
+    // before any work arrives. See restore.rs F1 comment for the same rationale.
+    let mut process = spawn_agent_child(app, record, &key.relay_url, true, owner_hex)?;
     let now = now_iso();
     let receipt = super::ManagedAgentRuntimeReceipt {
         key: key.clone(),
--- a/desktop/src-tauri/src/managed_agents/types.rs
+++ b/desktop/src-tauri/src/managed_agents/types.rs
@@ -728,7 +728,7 @@ pub struct UpdateTeamRequest {
 pub const DEFAULT_ACP_COMMAND: &str = "buzz-acp";
 /// ~5 min (320s) — matches the CLI harness default (BUZZ_ACP_IDLE_TIMEOUT).
 pub const DEFAULT_AGENT_TURN_TIMEOUT_SECONDS: u64 = 320;
-pub const DEFAULT_AGENT_PARALLELISM: u32 = 24;
+pub const DEFAULT_AGENT_PARALLELISM: u32 = 1;
 
 fn default_agent_parallelism() -> u32 {
     DEFAULT_AGENT_PARALLELISM
--- a/desktop/src-tauri/src/managed_agents/types/tests.rs
+++ b/desktop/src-tauri/src/managed_agents/types/tests.rs
@@ -649,3 +649,47 @@ fn mint_rejects_out_of_range_input_parallelism() {
         "input-branch error must not blame the definition: {err}"
     );
 }
+
+/// Regression test for issue #2631: the default parallelism must be 1 so that
+/// desktop agents do not eagerly spawn large ACP pools on startup, consuming
+/// double-digit GiB of RAM and creating idle Codex sessions.
+#[test]
+fn default_agent_parallelism_is_one() {
+    assert_eq!(
+        super::DEFAULT_AGENT_PARALLELISM,
+        1,
+        "default parallelism must be 1 to avoid eagerly spawning large ACP pools; \
+         see issue #2631"
+    );
+}
+
+/// Records created before explicit parallelism was stored must deserialize to
+/// the safe default of 1, not the old default of 24.
+#[test]
+fn managed_agent_record_without_parallelism_defaults_to_one() {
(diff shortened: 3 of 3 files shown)


Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Friendly bump -- let me know if anything needs changing.

Christian-Sidak:
Closing due to inactivity. Happy to reopen if there's interest.

PR #2678, opened 2026-07-24
fix(buzz-audit): truncate created_at to microseconds before hashing

### LINKED ISSUE #2637: buzz-audit: every hash chain fails verification — created_at is hashed at nanosecond precision but stored at microsecond precision

Christian-Sidak:
## Summary

- `Utc::now()` returns nanosecond precision but Postgres `TIMESTAMPTZ` stores only microseconds; `to_rfc3339()` emits 9 vs 6 fractional digits, producing a different byte string and thus a different SHA-256 digest
- `verify_chain` recomputes the hash from the stored (truncated) row and compares it to the written hash -- they never matched, so every chain failed verification on untampered data
- This made `AuditError::HashMismatch` the steady state, rendering tamper detection useless

Fix: call `.trunc_subsecs(6)` in `log_inner` before computing the hash so the value hashed at write time is identical to what Postgres returns on read. One-line change at the single choke point (`NewAuditEntry` carries no timestamp; all callers go through `log_inner`).

Adds three Postgres-free regression tests to `hash.rs`:
- `nanosecond_and_microsecond_timestamps_hash_differently` -- documents the root cause
- `hash_survives_microsecond_precision_round_trip` -- proves the fix works
- `untruncated_nanosecond_timestamp_fails_round_trip` -- regression guard

## Test plan

- [x] `cargo fmt -p buzz-audit` passes (no formatting changes)
- [x] Three new regression tests in `hash.rs` cover the invariant, the trap, and the round-trip
- [x] All six `#[ignore = "requires Postgres"]` chain tests in `service.rs` are expected to pass once run against a real database (per the issue, they fail on unpatched main and should pass with this fix)

Fixes #2637

### DIFF
--- a/crates/buzz-audit/src/hash.rs
+++ b/crates/buzz-audit/src/hash.rs
@@ -174,4 +174,95 @@ mod tests {
         let b = serde_json::json!({"a": 2, "m": 3, "z": 1});
         assert_eq!(canonical_json(&a).unwrap(), canonical_json(&b).unwrap());
     }
+
+    // Regression tests for #2637: created_at must be truncated to microsecond
+    // precision before hashing so the digest survives the round-trip through a
+    // TIMESTAMPTZ column (Postgres discards sub-microsecond digits on write).
+
+    /// The trap: `to_rfc3339()` emits 9 fractional digits for nanosecond
+    /// timestamps and 6 for microsecond ones -- different byte strings, different
+    /// hashes. This test documents the invariant that the compute_hash function
+    /// depends on.
+    #[test]
+    fn nanosecond_and_microsecond_timestamps_hash_differently() {
+        use chrono::SubsecRound as _;
+
+        // A timestamp with sub-microsecond digits.
+        let ns_ts: DateTime<Utc> =
+            chrono::DateTime::parse_from_rfc3339("2026-07-24T12:00:00.123456789Z")
+                .unwrap()
+                .with_timezone(&Utc);
+        // The same timestamp truncated to µs -- what Postgres returns after storing it.
+        let us_ts = ns_ts.trunc_subsecs(6);
+
+        assert_ne!(ns_ts, us_ts, "test setup: timestamps must differ");
+
+        let mut ns_entry = sample_entry();
+        ns_entry.created_at = ns_ts;
+        let mut us_entry = sample_entry();
+        us_entry.created_at = us_ts;
+
+        assert_ne!(
+            compute_hash(&ns_entry).unwrap(),
+            compute_hash(&us_entry).unwrap(),
+            "nanosecond and microsecond timestamps must produce different hashes"
+        );
+    }
+
+    /// The fix: hashing a µs-truncated timestamp produces a digest that can be
+    /// reproduced after a simulated Postgres round-trip (which also truncates to
+    /// µs). Before the fix, the hash was computed over the nanosecond value and
+    /// could never be reproduced from the stored row.
+    #[test]
+    fn hash_survives_microsecond_precision_round_trip() {
+        use chrono::SubsecRound as _;
+
+        // A timestamp with sub-microsecond digits, as returned by Utc::now().
+        let raw: DateTime<Utc> =
+            chrono::DateTime::parse_from_rfc3339("2026-07-24T12:00:00.123456789Z")
+                .unwrap()
+                .with_timezone(&Utc);
+
+        // Truncate before hashing (the fix in log_inner).
+        let stored = raw.trunc_subsecs(6);
+        let mut entry = sample_entry();
+        entry.created_at = stored;
+        let hash_at_write = compute_hash(&entry).unwrap();
+
+        // Simulate what verify_chain gets back from Postgres: the µs-truncated value.
+        let reread_from_db = stored; // TIMESTAMPTZ returns the same µs value
+        entry.created_at = reread_from_db;
+        let hash_at_verify = compute_hash(&entry).unwrap();
+
+        assert_eq!(
+            hash_at_write, hash_at_verify,
+            "hash computed before write must equal hash recomputed from stored row"
+        );
+    }
+
+    /// Guard against regressions that re-introduce nanosecond hashing: a timestamp
+    /// with non-zero sub-microsecond digits must round-trip only when truncated.
+    #[test]
+    fn untruncated_nanosecond_timestamp_fails_round_trip() {
+        use chrono::SubsecRound as _;
+
+        let raw: DateTime<Utc> =
+            chrono::DateTime::parse_from_rfc3339("2026-07-24T12:00:00.123456789Z")
+                .unwrap()
+                .with_timezone(&Utc);
+
+        // Hash WITHOUT truncation (the old, broken behaviour).
+        let mut entry = sample_entry();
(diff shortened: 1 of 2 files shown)


wpfleger96:
🤖 hey — thanks for tracking this down, the diagnosis is exactly right (ns-precision `Utc::now()` at write time vs µs after the Postgres round-trip, so the verify-time preimage differs). We ended up going with #2638 for the same root cause since it also normalizes inside `compute_hash` (so future call sites can't reintroduce the split) and carries a unit test that guards the fix itself rather than just pinning the bug. I think this one is superseded by that, but appreciate the report and the repro tests — they made verifying the root cause easy.

Christian-Sidak:
Makes sense, normalizing inside compute_hash is the more durable fix and the better test to carry. Closing in favor of #2638. Thanks for the quick review!

PR #2448, opened 2026-07-23
fix(desktop): shared-compute usability — share toggle, usage indicator, model resync

micspiral:
image

Shared-compute usability fixes, discovered while getting a local agent to reliably run and reply on Buzz shared compute. Three related but distinct changes:

### 1. Share toggle reflects serve mode, not slot occupancy
- Serve (sharing) and consume (client) share one runtime slot, and both report `running` — so consuming a peer's compute wrongly lit the Share toggle.
- Toggle is now mode-aware: `serve` = sharing (on), `client` = consuming (off, disabled with an explanatory line).
- Backend guard: `mesh_stop_node` refuses to tear down a client node, so a stale UI can't kill a consume session.
- A **failed** serve node still occupies the slot, so it reads as sharing and stays turn-off-able (no dead-end).
- Pure predicate + serde-contract + lifecycle tests; e2e regression proven red→green.

### 2. Surface who is using shared compute on the Share card
- Read-only `mesh_serving_usage` command projects the node's own `routing_metrics` (inflight, tokens served, tok/s, and the local-vs-remote/endpoint split that distinguishes *your* agent from a peer consuming your compute).
- Small indicator under the Share status line: idle / serving your agent / in use by another member.
- No new trust surface — just exposes what the runtime already computes. Stubbed for non-mesh builds. Unit + e2e coverage.

### 3. Resync the Share model field when the served model changes
- The card mirrored the node's `modelId` into the field only when the field was empty, so switching models (node restarts on the new one) left a stale model shown.
- Now syncs whenever the served model differs from the field. Safe from clobbering a mid-edit because the field is disabled while the slot is occupied.

### Testing
- Rust: predicate/serde/lifecycle unit tests (all green); clippy + fmt clean.
- Frontend: predicate + serving-usage unit tests, e2e regression specs.
- Manual: full serve lifecycle (on → serve/route agent → clean stop → cache-backed restart) on the meshllm community.

---

> **Note:** This PR originally also included a buzz-acp content-delivery fallback
> (posting a plain-text reply when a weak model answers in prose and never calls
> the send tool). That change has been split out into its own PR — #2681 — so it
> can be reviewed independently of these shared-compute usability fixes, and has
> been reverted out of this branch.

### Dependency bump
Pins the six `mesh-llm` crates (`sdk`, `host-runtime`, `client`, `node`,
`system`, `events`) from tag `v0.73.1` → rev `c441ea7` (v0.73.1 + 1 commit) —
[...]

PR #2681, opened 2026-07-24
fix(acp): deliver plain-text replies when the model skips the send tool

micspiral:
## What

Adds a **content-delivery fallback** to `buzz-acp`: when a managed agent answers a
conversational prompt in plain assistant text and never calls the publish tool
(`buzz messages send`), the harness posts that streamed content as a threaded
kind-9 reply — instead of silently dropping it.

Capable hosted models reliably publish via the send tool, but weaker local /
shared-compute models (e.g. a small Gemma/Qwen) often just answer in prose and
never call the tool, so the reply is lost. This closes that gap.

## How

Delivery is confirmed from the send tool call's **terminal outcome**, not its
input text — intent is not delivery:

- A `tool_call` whose input matches the publish signature (normalized so both
  shell strings *and* argv forms like `subprocess.run(['buzz','messages','send',…])`
  match) only registers a **candidate** `toolCallId`.
- The candidate is resolved on its terminal `tool_call_update`:
  - `failed` / `cancelled` → **not** delivered (fallback stays armed).
  - `completed` → inspect `rawOutput.isError`, the CLI's `{"accepted":…}`
    response envelope, and any reported `exit_code`. A bare `completed` with no
    inspectable output keeps the status-quo direction (suppress).
- On a normal `EndTurn` with buffered content but no confirmed publish, post the
  content as a threaded reply (`buzz_sdk::build_message`, signed with the agent
  keys, best-effort — mirrors `post_failure_notice`). Bare acknowledgements the
  base prompt forbids publishing are skipped.

Fires only on `EndTurn` (not `MaxTokens`/`MaxTurnRequests`, which are
truncated/aborted turns) and only when a triggering channel event was present
(not heartbeats).

## Testing

- Unit: send-tool detection incl. read-vs-send discrimination and argv-style
  publishes; the outcome classifier matrix
  (pending/failed/cancelled/completed × envelope/isError/exit_code); bare-ack
  filtering.
- Handler-level lifecycle tests through `handle_session_update` +
  `take_undelivered_turn_message`: success, failure, cancel, failed-then-retry,
  and argv-style publish.
- Original fallback was live-validated against a small Gemma on a real
  community: fires on prose answers, stays dormant when the model calls the
  send tool.

## Notes

Split out of #2448 so this reply-delivery fix stands on its own, independent of
that PR's shared-compute usability changes. `+710 / crates/buzz-acp` only — no
desktop / mesh / toggle changes.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -200,6 +200,17 @@ pub struct AcpClient {
     /// deltas. Both goose and buzz-agent emit this notification; goose gates
     /// on client capability advertisement, buzz-agent emits unconditionally.
     goose_usage: UsageTracker,
+    /// Accumulated `agent_message_chunk` text for the current turn. Used by the
+    /// content-delivery fallback: weak local models (e.g. via Buzz shared
+    /// compute) often answer a conversational prompt in plain assistant
+    /// `content` instead of calling `buzz messages send`, which would otherwise
+    /// be silently dropped (buzz-agent's output is its tool calls; streamed
+    /// text is observability-only). Reset at the start of every turn.
+    turn_message_text: String,
+    /// Whether a `buzz messages send` (or forum-post/comment) publish tool call
+    /// was observed this turn. When true the fallback does NOT fire — the agent
+    /// delivered its own reply. Reset at the start of every turn.
+    turn_sent_message: bool,
 }
 
 /// Recursively merge `overlay` into `base`, with `overlay` winning on scalar/shape
@@ -496,6 +507,8 @@ impl AcpClient {
             active_run_id: None,
             steer_rx: None,
             goose_usage: UsageTracker::default(),
+            turn_message_text: String::new(),
+            turn_sent_message: false,
         })
     }
 
@@ -689,6 +702,10 @@ impl AcpClient {
         // misattributed to this turn.
         self.goose_usage.begin_turn(session_id);
 
+        // Reset the content-delivery fallback trackers for this turn.
+        self.turn_message_text.clear();
+        self.turn_sent_message = false;
+
         self.last_prompt_id = Some(self.next_id);
         let id = self.next_id;
         self.next_id += 1;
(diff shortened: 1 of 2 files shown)


erauner12:
Additional live validation: I tested this PR unchanged with OpenClaw 2026.6.1 acting as a plain-text ACP server behind `buzz-acp`.

Path: Buzz mention → `buzz-acp` → `kubectl exec -i` → `openclaw acp` → existing OpenClaw Gateway session.

OpenClaw emitted standard `agent_message_chunk` text and had no Buzz CLI/MCP tools. The harness logged `content-delivery fallback: posted undelivered agent content as channel reply`, and the relay contained exactly one correctly threaded kind:9 reply with the requested text. The full `buzz-acp` test suite also passed locally (600 unit tests + 9 lifecycle tests).

This confirms the fallback covers an ACP bridge where giving the downstream agent the Buzz signing key or a general-purpose shell/MCP solely for reply delivery would be undesirable.

micspiral:
I am going to close this as we have had a few runs at other ways of this - but if this comes back, we will bring it back

cragnet:
Another independent recurrence of exactly this failure mode, on top of @erauner12's report above.

**Setup:** `buzz-acp` → `BUZZ_ACP_AGENT_COMMAND=openclaw`, `openclaw acp` as the ACP server — same shape as the report above (generic ACP bridge, no Buzz-specific tools/MCP exposed to the downstream agent, local/cloud-routed model). This is on a self-hosted community/relay, not the hosted one. Symptom: agent reacts to @mentions, generates real conversational text, but never calls `buzz messages send` — reply silently dropped, only recoverable by giving the downstream agent the signing key directly (undesirable, as noted above).

We've hit this on both x86_64 and arm64 hosts in the same fleet, so it isn't architecture-specific — and had to apply this patch ourselves to fix it.

**What we did:** cherry-picked this commit + `977457865` (the outcome-based delivery-confirmation follow-up) cleanly onto current `main` (`eed74bde`, no conflicts outside two adjacent struct-field additions). Full `buzz-acp` suite passes (838 unit + 9 lifecycle tests). Deployed to a live community — confirmed the fallback fires correctly and produces properly NIP-10-threaded replies when the model skips the tool, and stays dormant when it doesn't.

Given the maintainer's own bar for reviving this was "if this comes back" — two independent deployments hitting the identical silent-drop bug seems like exactly that.

PR #2683, opened 2026-07-24
fix(desktop): refine floating content surface

villemki:
## Summary

- reduce the floating desktop content surface right and bottom inset from 8px to 4px
- introduce an 8px optical radius so the inner surface better follows the native macOS window contour
- apply the same treatment to Settings and cover the geometry in the existing screenshot test

## Why

The previous 16px surface radius and 8px inset made the inner page corner feel disconnected from the surrounding macOS window. The smaller inset and dedicated optical radius create a tighter, more coherent frame while retaining a visible strip of the Buzz gradient.

## Testing

- `just ci`
- `just desktop-check`
- `just desktop-test`
- `pnpm -C desktop build:e2e`
- `pnpm -C desktop exec playwright test tests/e2e/buzz-theme-screenshots.spec.ts --project=smoke --grep "settings content uses the same inset surface as the main app"`
- repository pre-push suite (desktop, Rust, Tauri, and mobile)

## Manual verification

- verified the main channel surface and Settings surface in the macOS desktop app
- checked the bottom-right corner at Retina scale

### DIFF
--- a/desktop/src/app/BuzzThemeSurfaces.tsx
+++ b/desktop/src/app/BuzzThemeSurfaces.tsx
@@ -22,7 +22,7 @@ export function GradientLayer() {
 export function ContentSurface({ children }: { children: ReactNode }) {
   return (
     <div
-      className="relative z-10 mb-2 ml-px mr-2 mt-px flex min-h-0 flex-1 flex-col overflow-hidden rounded-2xl bg-background shadow-content-edge"
+      className="relative z-10 mb-1 ml-px mr-1 mt-px flex min-h-0 flex-1 flex-col overflow-hidden rounded-content-surface bg-background shadow-content-edge"
       data-buzz-content-surface
     >
       {children}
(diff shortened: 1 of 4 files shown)


wesbillman:
> **Agent disclosure:** Carl is posting this review feedback on Wes’s behalf.

From my agent, Carl.

This visual refinement looks coherent in isolation, but the branch needs a small refresh against current `main` before merge.

Current `main` now has `desktop/src/features/channels/ui/FocusThreadDrawer.tsx`, whose left corners deliberately use `rounded-l-2xl` (16px). The adjacent comment explicitly says that radius matches the app content surface. This PR changes the enclosing content surface from 16px to the new 8px `content-surface` radius, but because the drawer landed after this branch's base, it is not updated here. Merging as-is would leave focused-thread mode with mismatched corner geometry.

Please:

1. Refresh this branch against current `main`.
2. Update the focused-thread drawer's left corners to use the same new 8px content-surface radius instead of `rounded-l-2xl`.
3. Update the adjacent stale `rounded-2xl` comment.
4. If inexpensive, add a focused-thread geometry assertion so the shared-radius contract cannot drift again.

The rest of the change looks good: the main and Settings surfaces move together, the new token is explicit, the Settings geometry coverage is targeted, and completed desktop checks are green. This is a small integration fix, not a request to redesign the patch.

Issue #2685, opened 2026-07-24
Windows: terminal windows flash constantly — discovery/auth probes and login-shell resolution spawn consoles without CREATE_NO_WINDOW


Acenta-Radovljica:
## Environment

- Buzz Desktop **v0.4.23** (installed from the release `Buzz_0.4.23_x64-setup_alpha-unsigned.exe`)
- Windows 11 Home 10.0.26200, Windows Terminal set as the default terminal application
- Agents running on the Claude Code harness (`@agentclientprotocol/claude-agent-acp`), `@agentclientprotocol/codex-acp` 1.1.7 installed, Goose **not** installed

## Symptom

While the app is running (especially when agents wake or the Agents settings view triggers runtime discovery), multiple terminal windows keep popping up, often several at once, stealing focus. With Windows Terminal as the default host each one opens as a full Windows Terminal window titled `C:\Windows\system32\cmd.exe`.

## Evidence — captured process spawns

A process monitor (400 ms polling, parent-chain resolution) captured this burst during one discovery pass. Every chain roots in `buzz-desktop.exe`:

```
[wsl.exe]    chain: wsl.exe < buzz-desktop.exe
  cmdline: "C:\Users\<u>\AppData\Local\Microsoft\WindowsApps\bash.exe" -l -c "command -v -- \"$1\"" _ goose   (x5 in one burst)
[buzz-acp.exe] chain: buzz-acp.exe < buzz-desktop.exe
  cmdline: "C:\Users\<u>\AppData\Local\Buzz\buzz-acp.exe" models --json
[cmd.exe]    chain: cmd.exe < buzz-desktop.exe
  cmdline: cmd.exe /e:ON /v:OFF /d /c ""C:\Users\<u>\...\codex-acp.cmd" --version"      (x2)
[claude.exe] chain: claude.exe < buzz-desktop.exe
  cmdline: "c:\users\<u>\.local\bin\claude.exe" auth status
[cmd.exe]    chain: cmd.exe < buzz-desktop.exe
  cmdline: cmd.exe /e:ON /v:OFF /d /c ""C:\Users\<u>\...\codex.cmd" login status"
```

Each of these is a console-subsystem child spawned from a GUI process, so Windows allocates a visible console for every one of them.

## Root cause

[...]

Issue #2643, opened 2026-07-23
[Bug] Invisible/transparent window on AMD RDNA4: WebKitGTK Skia GPU path fails on non-conformant radv; WEBKIT_SKIA_ENABLE_CPU_RENDERING=1 fixes it


bealmot:
## Describe the bug

On AMD RDNA4 GPUs (Mesa `radv` still marks gfx1200 non-conformant), the Buzz Desktop window is **invisible/transparent**: the window opens, but the webview never paints (or paints one frame and then permanently stops repainting after any move/minimize). Because the Tauri window is configured with `"transparent": true` (`desktop/src-tauri/tauri.conf.json`), the failure presents as a ghost window — desktop wallpaper visible through a Buzz-shaped frame, titlebar rendered normally.

The root cause is **not** the dmabuf renderer issue from #2338 (that workaround does not help here). WebKitGTK 2.52 renders via **Skia**, and on this GPU Skia's GPU path goes through Vulkan/`radv`, which self-reports:

```
WARNING: radv is not a conformant Vulkan implementation, testing use only.
```

Forcing Skia to CPU rasterization fully fixes rendering:

```bash
WEBKIT_SKIA_ENABLE_CPU_RENDERING=1
```

## To Reproduce

1. Machine with an AMD RDNA4 GPU (tested: RX 9060 XT / gfx1200), Mesa 26.1.2, KDE Plasma 6.6.5 (Wayland session)
2. Launch Buzz Desktop (either the 0.4.23 AppImage after the #2335 libsystemd fix, **or** a from-source `just dev` build using system webkit2gtk-4.1 2.52.4 — both reproduce)
3. Window opens; content area is transparent (or paints the page background once, then goes permanently transparent after moving/minimizing the window)
4. Relaunch with `WEBKIT_SKIA_ENABLE_CPU_RENDERING=1` → full UI renders correctly and survives move/minimize

## Expected behavior

The window should render its content, or fall back to a rendering path that works instead of an invisible window.

## Supporting Material

Isolation matrix (all on the same machine/session):

| Configuration | Result |
|---|---|
| Default (XWayland, `GDK_BACKEND=x11` from AppRun hook) | transparent |
| `WEBKIT_DISABLE_DMABUF_RENDERER=1` (the #2338 workaround) | WebKitWebProcess stays alive, still transparent |
| `GDK_BACKEND=wayland` (native Wayland; no `eglMakeCurrent` failure on AMD, unlike tauri#8541 on Intel) | transparent |
[...]

Issue #2548, opened 2026-07-23
[Bug] AppImage crashes when rendering COLRv1 emoji fonts on Fedora


tnull:
### Description

On Fedora with Wayland/Hyprland, the Buzz AppImage opens as a blank, transparent window and then aborts while rendering an emoji.

### Error

> ././/include/c++/12/bits/stl_vector.h:1123: constexpr std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = {anonymous}::colrv1_configure_skpaint(FT_Face, const SkSpan<unsigned int>&, SkColor, const FT_COLR_Paint&, SkPaint*)::<lambda(const FT_ColorStopIterator&, std::vector<float>&, std::vector<SkRGBA4f<kUnpremul_SkAlphaType> >&)>::ColorStop; _Alloc = std::allocator<{anonymous}::colrv1_configure_skpaint(FT_Face, const SkSpan<unsigned int>&, SkColor, const FT_COLR_Paint&, SkPaint*)::<lambda(const FT_ColorStopIterator&, std::vector<float>&, std::vector<SkRGBA4f<kUnpremul_SkAlphaType> >&)>::ColorStop>; reference = {anonymous}::colrv1_configure_skpaint(FT_Face, const SkSpan<unsigned int>&, SkColor, const FT_COLR_Paint&, SkPaint*)::<lambda(const FT_ColorStopIterator&, std::vector<float>&, std::vector<SkRGBA4f<kUnpremul_SkAlphaType> >&)>::ColorStop&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.

This appears to be triggered when WebKitGTK/Skia renders Fedora's COLRv1 Noto Color Emoji font. The same crash has been reported in other WebKitGTK-based AppImages: https://github.com/farion1231/cc-switch/issues/2176

### Reproduction

1. Run the Buzz AppImage on Fedora under Wayland/Hyprland.
2. Open a view containing emoji, or allow the initial UI to render.
3. The window remains blank/transparent and the process aborts with the
     assertion above.

### Workaround

A Fontconfig override that prevents Buzz from using the system color emoji font avoids the crash.

### Suggested fix

Bundle and preload a non-COLRv1 emoji font in the desktop application before React renders, and place it ahead of generic font fallbacks. Native Apple and Segoe emoji fonts could remain preferred on macOS and Windows.

### Environment

- Buzz: v0.4.23 AppImage
[...]

PR #2689, opened 2026-07-24
fix(desktop): Skia CPU, COLRv1 emoji, WindowsApps bash

Chessing234:
## Summary
- Prefer WebKit Skia CPU rendering on Linux so AMD RDNA4 doesn’t stay transparent (#2643)
- Reject Fedora’s COLRv1 Noto Color Emoji in the AppImage fontconfig (#2548)
- Skip WindowsApps `bash.exe` aliases so discovery doesn’t boot WSL (#2685)

## Test plan
- [x] `bash -n desktop/scripts/fix-appimage.sh`
- [x] `just desktop-tauri-fmt-check`
- [x] `just desktop-tauri-clippy`
- [ ] AppImage on RDNA4 / Fedora emoji path if available
- [ ] Windows discovery with Git Bash + WindowsApps alias present

### DIFF
--- a/desktop/scripts/fix-appimage.sh
+++ b/desktop/scripts/fix-appimage.sh
@@ -114,6 +114,36 @@ echo "==> Symlinking system GStreamer plugin directory"
 rm -rf "$LIBDIR/gstreamer-1.0"
 ln -s "/usr/lib/$MULTIARCH/gstreamer-1.0" "$LIBDIR/gstreamer-1.0"
 
+# WebKitGTK 2.52 Skia's GPU path goes through Vulkan/radv. On non-conformant
+# AMD RDNA4 (gfx1200) that path paints nothing — transparent ghost window —
+# while WEBKIT_DISABLE_DMABUF_RENDERER does not help (#2643). Prefer Skia CPU
+# raster for AppImage launches; operators can override with =0.
+echo "==> Preferring WebKitGTK Skia CPU rendering for AppImage launches"
+HOOK_DIR="$WORKDIR/squashfs-root/apprun-hooks"
+mkdir -p "$HOOK_DIR"
+cat > "$HOOK_DIR/99-buzz-webkit-skia-cpu.sh" <<'EOF'
+# Allow operators to override (e.g. WEBKIT_SKIA_ENABLE_CPU_RENDERING=0) if needed.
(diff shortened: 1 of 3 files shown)


wesbillman (changes requested):
Reviewed on behalf of Wes.

Thank you for digging into these platform failures and linking each change to a concrete report. The underlying Linux issues are real, and the Windows diagnosis was also valid when this branch was created.

I’m requesting changes because this three-fix bundle is now partly superseded and the remaining fixes need narrower design and affected-host validation:

1. **WindowsApps / WSL alias:** this is already addressed on current `main` by #2587. That implementation filters the precise `Microsoft/WindowsApps` alias path and includes broader console-spawn handling and cross-platform tests. This commit should be dropped rather than adding a second, broader filter inside generic command resolution.

2. **AMD RDNA4 / Skia:** defaulting `WEBKIT_SKIA_ENABLE_CPU_RENDERING=1` for every Linux launch applies a global rendering/performance tradeoff based on a hardware-specific failure, and it duplicates policy in both `lib::run()` and AppImage hooks. Current `main` now has the dedicated, tested pre-WebKit rendering-policy module from #3271, including environment ownership and a `--safe-rendering` escape hatch. Any RDNA4-specific CPU fallback should be designed there and conditioned on appropriate evidence rather than enabled for all Linux users.

3. **Fedora COLRv1 emoji:** #2548/#2982 remain legitimate open compatibility bugs, but this patch rejects the entire `Noto Color Emoji` family and may degrade users to monochrome glyphs. The affected AppImage/Fedora path is also unchecked in this PR’s test plan. This should be a focused follow-up with a known-good fallback font (or another narrowly targeted mitigation), validation on an affected Fedora host using the produced AppImage, and an artifact-level regression check where feasible.

Because one part is already obsolete and the other two require separate redesign/validation against current architecture, I recommend closing this aggregate PR rather than rebasing it. Focused follow-ups for the still-open Linux bugs would be welcome.

Thanks again for identifying and attempting to fix difficult platform-specific failures. This request is about current-main fit and safely scoping the mitigations, not the intent behind the contribution.

Chessing234:
fair points — windowsapps is already covered on main, and the skia/colrv1 bits need narrower follow-ups instead of this bundle. closing this one; will open focused prs for the remaining linux issues.

Chessing234:
closing per review — superseded windowsapps fix is on main; skia/colrv1 need separate scoped follow-ups.

Issue #2686, opened 2026-07-24
workflow_sink's @-mention parser is a third divergent implementation — workflow-emitted **@Name** emits zero p tags


bosoud:
Split out of #2526 — noted by @SeanGearin while reviewing #2547.

## Problem

#2526 identified two divergent @-mention parsers (Rust `buzz-sdk/src/mentions.rs` and TS `desktop/src/features/messages/lib/hasMention.ts`). There is a **third**: `resolve_mention_pubkeys` in `crates/buzz-relay/src/workflow_sink.rs`, whose doc comment says it "defines the matching contract" for workflow-emitted kind:9 `p` tags.

Its left-boundary check accepts only start-of-string, whitespace, or `(`:

```rust
let is_left_boundary = |i: usize| i == 0 || chars[i - 1].is_whitespace() || chars[i - 1] == '(';
```

So a workflow-generated `**@Robby**` emits zero `p` tags. Those `p` tags gate ACP agent wake (`event_mentions_agent`, `crates/buzz-acp/src/lib.rs`), which reproduces the exact stalled-handoff failure mode from #2526 on a surface where the text is machine-generated — an agent or template emitting emphasis around a mention is not a user typo, it's routine markdown.

The right side of this parser is already emphasis-tolerant (it rejects only name-continuation chars, `alphanumeric | _`), so only the left boundary needs widening. Note the `_` wrinkle: a trailing `_` after a name is a name-continuation char here, so `_@robby_` has a second, smaller divergence too.

## Suggested direction

After #2547 lands, `buzz-sdk::mentions::is_mention_lead` is the shared definition of "what may precede an `@`" (whitespace, `(`, `*`, `_`, `|`, start). `buzz-relay` already depends on `buzz-sdk`, so `workflow_sink.rs` can call it directly instead of keeping a private boundary rule — that would take the parser count needing manual sync from three to two.

Longer term (already noted in #2526): the Rust and TS implementations should be one contract with shared test vectors, or this class of drift recurs.

Chessing234:
Fix in #: widen workflow mention left-boundary for markdown emphasis.

silveyy:
Supporting data point, distinct sub-case (not markdown-boundary related — flagging here rather than as a new issue since it's the same parser and file):

[...]

Issue #2562, opened 2026-07-23
[Bug] Linux: notification sounds never play, WebKitGTK can't load media from Tauri's custom URI scheme (MEDIA_ERR_SRC_NOT_SUPPORTED)


JoeMattie:
**Describe the bug**

On Linux, notification sounds never play. `playNotificationSound()` builds `new Audio("/sounds/<name>.mp3")` (`desktop/src/features/notifications/lib/sound.ts`), which resolves against the page origin. On Linux, Tauri serves `frontendDist` through a custom URI scheme registered with `webkit_web_context_register_uri_scheme` (`tauri://localhost`), and **WebKitGTK's media backend cannot load media from a custom URI scheme**. The element fails with `MEDIA_ERR_SRC_NOT_SUPPORTED` (code 4) at `readyState 0`, and both error paths in `playNotificationSound` swallow it:

```ts
audio.play().catch(() => {
  // Best-effort — user may not have interacted with the page yet.
});
```

so there is no console output, no toast, and no telemetry. The desktop notification itself still appears, which is why this reads as "sound settings do nothing" rather than "notifications are broken". On GNOME the notification daemon plays its own default alert, so users can hear *a* sound while never hearing the sound they picked in Settings.

This is Linux-only. WKWebView and WebView2 both serve media from their custom scheme handlers.

**To Reproduce:**

1. Run Buzz Desktop on Linux (AppImage or a local `pnpm tauri dev` build).
2. Settings > Notifications: pick any sound for @Mentions, confirm alerts are on.
3. Have someone @mention you in a channel.
4. The desktop notification appears. The selected `.mp3` never plays.

Minimal standalone repro, no Buzz needed. Requires `gjs` and `gir1.2-webkit2-4.1` (Debian/Ubuntu: `gjs libwebkit2gtk-4.1-0 gir1.2-webkit2-4.1`). Put any mp3 at `/tmp/flutter.mp3`:

```js
// repro.js (run: gjs repro.js)
imports.gi.versions.Gtk = '3.0';
const { Gtk, WebKit2, GLib, Gio } = imports.gi;
Gtk.init(null);

const HTML = `<!doctype html><html><body><script>
const a = new Audio('/sounds/flutter.mp3');
a.addEventListener('ended', () => document.title = 'RESULT|ENDED');
a.addEventListener('error', () => document.title =
  'RESULT|MEDIA_ERROR|readyState=' + a.readyState + '|code=' + a.error.code);
[...]

PR #2691, opened 2026-07-24
fix: workflow @mentions and Linux notification sounds

Chessing234:
## Summary
- Workflow-emitted `**@Name**` now gets `p` tags so ACP can wake the agent (#2686)
- Play notification sounds from blob URLs so WebKitGTK can hear them (#2562)
- Keep playing sounds even when the OS toast fails to deliver

## Test plan
- [x] `cargo test -p buzz-relay --lib markdown_emphasis`
- [x] `cargo fmt -p buzz-relay -- --check`
- [x] `cargo clippy -p buzz-relay --lib -- -D warnings`
- [x] biome check on touched desktop files
- [ ] Mention sound on Linux AppImage / `tauri dev`


bosoud:
Two notes on the `workflow_sink.rs` part (I filed #2686):

1. **`_@Robby_` still wakes no one, and the new test can't see it.** The left boundary now accepts `_`, but `extends_name` still treats `_` as a name-continuation char, so after matching `Robby` the trailing `_` fails the right-boundary check and the mention is rejected. The new test passes anyway because the same string also contains `*@Robby*`, which matches the same pubkey — and the output is deduplicated, so the assertion can't distinguish "both matched" from "only the `*` one matched". Standalone repro:

   ```rust
   resolve_mention_pubkeys("_@Robby_ take a look", &[m("Robby", &pk('a'))])  // currently []
   ```

   Dropping `_` from `extends_name` fixes it without harming underscore display names — a member literally named `will_smith` still binds fully because longer names match first (worth pinning with its own test).

2. **This adds a fourth hand-maintained copy of the boundary rule.** The same delimiter set now lives in `hasMention.ts`, `buzz-sdk/mentions.rs` (being consolidated in #2547 per review there), and twice-diverged here — the drift class that produced #2526 in the first place. Once #2547 lands, `buzz_sdk::mentions::is_mention_lead` is the shared definition and `buzz-relay` already depends on `buzz-sdk`; I have a ready branch converging `workflow_sink` on it (including the `extends_name` fix + tests) and am happy to PR it on top of either ordering — whichever lands first.

Chessing234:
Dropped `_` from `extends_name` so `_@Robby_` wakes Robby; added a standalone `_@Robby_` case plus a `will_smith` longest-match pin. Happy to take your buzz-sdk converge PR on top once #2547 lands.

Chessing234:

Chessing234:

PR #2693, opened 2026-07-24
fix(desktop): allow channel member agents in mentions

satsdisco:
## Summary

Fixes Desktop mention resolution for remote/shared managed agents that are already channel members but are not owned by the local Desktop.

The verified user-facing failure was: Satsdisco could see a remote managed agent, Fizz, as a channel member, but typing `@Fizz` did not produce the agent `p` tag. Another user's `@Fizz` message did include Fizz's `p` tag, which narrowed the issue to local Desktop mention candidate filtering/resolution rather than relay delivery.

## Root Cause

Desktop mention autocomplete had two related assumptions that broke for remote managed agents:

- Agent identities from search/autocomplete were filtered unless they were in the local managed-agent list.
- Channel-member agents could still be hidden when relay-directory invocability data made them look non-invocable/owner-only.

That is too strict for shared channel usage. Channel membership is the concrete signal that the identity can be mentioned in that channel. Whether the remote agent actually responds is still enforced by the receiving agent/relay policy.

## Changes

- Keeps channel-member bots/agents visible in mention autocomplete even when they are remote managed agents.
- Counts channel-member bots as known agent pubkeys during message mention resolution, so selected mentions emit the correct agent `p` tag.
- Scopes relay-directory mentionability to the active channel so shared relay agents from other channels do not leak into the current channel's autocomplete.
- Adds coverage for active-channel scoping, allowlist channel placement, remote allowed agent identities, and channel-member agents with non-invocable directory entries.

## Validation

Validated at commit `cff618911b0c26d810055cb1b787f3cc487ffed5`:

- `corepack pnpm --dir desktop test` - 3,374 passed, 0 failed
- `corepack pnpm --dir desktop typecheck`
- `corepack pnpm --dir desktop build`

Also validated manually via the replacement DMG: typing `@Fizz` from Satsdisco now works and emits the intended mention behavior.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -77,6 +77,29 @@ test("relayAgentIsSharedWithUser: accepts shared anyone agents and rejects unsha
   );
 });
 
+test("relayAgentIsSharedWithUser: active channel must match the agent channel", () => {
+  const sharedChannelIds = new Set(["general", "ops"]);
+
+  assert.equal(
+    relayAgentIsSharedWithUser(
+      { respondTo: "anyone", respondToAllowlist: [], channelIds: ["general"] },
+      sharedChannelIds,
+      CURRENT_PUBKEY,
+      "general",
+    ),
+    true,
+  );
+  assert.equal(
+    relayAgentIsSharedWithUser(
+      { respondTo: "anyone", respondToAllowlist: [], channelIds: ["ops"] },
+      sharedChannelIds,
+      CURRENT_PUBKEY,
+      "general",
+    ),
+    false,
+  );
+});
+
 test("relayAgentIsSharedWithUser: accepts allowlist agents for the current user", () => {
   const sharedChannelIds = new Set(["general"]);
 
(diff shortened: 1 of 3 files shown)


rmichelena:

rmichelena:
Cross-linking from #5483, where I'd listed this PR alongside several others as "same symptom, eligibility layer". Having read it, that label was too coarse and I've corrected it there — this one is a *tightening* of `relayAgentIsSharedWithUser` (requiring the active channel to be one of the agent's), not another attempt at the same admit path, so it should stand or fall on its own merits.

Two observations, offered as review rather than as a request:

- **It doesn't close #5363.** On our repro the agent arrives with `channel_ids: []` and `respond_to: null`, because `list_relay_agents` builds the directory from `kind:10100` only while membership is published as `kind:39002` and policy as `kind:30177`. An empty `channel_ids` fails the existing overlap check before your stricter one is reached. #5483 fixes that layer.
- **The signature has moved.** `main`'s fourth parameter is now `channelMemberAgentPubkeys` (the live-membership fallback from #4913), not an active channel id, so this will need a rebase before it applies.

If the intent is the invariant "an agent is only mentionable in a channel it's actually in", that's compatible with #5483 — with a real `channel_ids` behind it, it would finally be testing something. Worth rebasing rather than dropping, I think, but that's a maintainer's call.

Issue #2690, opened 2026-07-24
[Bug] macOS auto-update fails signature verification on Intel — both darwin arches share one unversioned Buzz.app.tar.gz in buzz-desktop-latest


bosoud:
**Describe the bug**

Auto-update on an Intel Mac fails with `Update failed: The signature verification failed` (observed updating a Block-signed 0.4.22 → 0.4.24). The cause is in the `buzz-desktop-latest` updater release: **both macOS platform entries in `latest.json` point to the same unversioned `Buzz.app.tar.gz`, but carry different signatures** — the two mac release jobs upload under one shared filename and the last upload wins, so at most one architecture's signature can ever verify.

`latest.json` as of `version: 0.4.24`:

| platform | url | signature |
|---|---|---|
| `darwin-aarch64` | `…/buzz-desktop-latest/Buzz.app.tar.gz` | `…d1VIeGs3a3REd2QyQ1hYeTNqSlhhL3lUK0RnPT0K` |
| `darwin-x86_64` | `…/buzz-desktop-latest/Buzz.app.tar.gz` | `…cXdrd3pNOWxsbmlZaEVOUzJUZUV2dFJTSUNRPT0K` |
| `linux-x86_64` | `…/Buzz_0.4.24_amd64.AppImage` | (versioned, per-file) |
| `windows-x86_64` | `…/Buzz_0.4.24_x64-setup_alpha-unsigned.exe` | (versioned, per-file) |

Same URL, two different signatures. The file currently at that URL is the **arm64** build:

```
$ gh release download buzz-desktop-latest -R block/buzz -p Buzz.app.tar.gz
$ tar -xzf Buzz.app.tar.gz && file Buzz.app/Contents/MacOS/buzz-desktop
Buzz.app/Contents/MacOS/buzz-desktop: Mach-O 64-bit executable arm64
```

So every darwin-x86_64 client fails signature verification — and the failure is actually load-bearing: if the x86_64 signature ever *did* verify, Intel users would be handed an arm64 binary.

Note the asset list also shows every Linux/Windows artifact since 0.3.20 accumulating in `buzz-desktop-latest` with versioned names, while macOS alone uses the overwriting unversioned name — which is presumably how the collision was introduced.

**To Reproduce**

1. Install `Buzz_0.4.22_x64.dmg` (Intel) from the v0.4.22 release.
2. Settings → check for updates → install.
3. `Update failed: The signature verification failed`.

**Expected behavior**

[...]

Issue #2560, opened 2026-07-23
[Linux/AppImage] Mixed GStreamer dependencies cause plugin failures, registry pollution, and unreliable audio on Ubuntu 26.04


JoeMattie:
## Summary

Buzz 0.4.23's amd64 AppImage mixes older bundled libraries and WebKitGTK with Ubuntu 26.04's GStreamer 1.28.2 plugins and core libraries.

This produces three user-visible or system-visible failures:

1. Four GStreamer plugins fail to load because host plugins resolve against incompatible libraries bundled in the AppImage.
2. Buzz writes its random `/tmp/.mount_Buzz_*` plugin paths into the desktop-wide GStreamer registry, so later non-Buzz GStreamer processes try to load plugins from a stale AppImage mount.
3. The WebKit web process continues to emit `gst_caps_append` criticals, and the originally failing Buzz sound remains silent, even when the three known library mismatches are worked around.

At least one Buzz notification sound did play during testing, so this is not a total failure of the host audio sink. Audio behavior is inconsistent, and the remaining silent path appears separate from the four blacklisted optional plugins.

## Environment

- Buzz: `0.4.23`
- Package: `Buzz_0.4.23_amd64.AppImage`
- OS: Ubuntu 26.04 LTS
- Kernel: `7.0.0-27-generic`
- Architecture: `x86_64`
- Host GStreamer: `1.28.2`

## Steps to reproduce

1. On Ubuntu 26.04, start the AppImage from a terminal:

   ```bash
   ./Buzz_0.4.23_amd64.AppImage
   ```

2. Observe the initial GStreamer plugin scan.
3. Trigger media or an in-app sound handled by the WebKit process.
4. Quit Buzz, then run a normal host GStreamer command and inspect `~/.cache/gstreamer-1.0/registry.x86_64.bin`.

## Expected behavior

- The AppImage uses one internally compatible GStreamer/WebKit dependency set.
- No GStreamer plugins are blacklisted because of symbol or symbol-version mismatches.
- Buzz keeps any AppImage-specific GStreamer registry separate from the user's desktop-wide registry.
- In-app sounds play consistently.
- The WebKit process does not emit GStreamer caps criticals.

## Actual behavior

The AppImage plugin scan reports:

```text
Failed to load plugin '.../libgstcurl.so':
/usr/lib/x86_64-linux-gnu/libcurl-gnutls.so.4: undefined symbol:
nghttp2_option_set_no_rfc9113_leading_and_trailing_ws_validation

[...]

PR #2694, opened 2026-07-24
fix: macOS updater archives, AppImage GST registry, mesh max_tokens

Chessing234:
## Summary
- #2690: ship arch-specific macOS updater tarballs so Intel/arm64 stop clobbering each other
- #2560: keep AppImage GStreamer registry under `~/.cache/buzz/`
- #2558: seed relay-mesh env from user overrides before applying mesh defaults (spawn path)

## Test plan
- [x] `cargo test --features mesh-llm --lib preserves_user_max_output`
- [x] `cargo fmt --check`, `bash -n` fix-appimage, YAML parse release.yml
- [ ] release assemble guard (CI)


bokiko:
Thanks for the quick fix. I validated the #2558 part of this PR on an independent two-node shared-compute testbed (SmolLM2-135M served over relay-mesh, consumer agent with a per-agent `BUZZ_AGENT_MAX_OUTPUT_TOKENS=1024`).

Tested at PR head `5c1f22bde279bbdccf86ce55ae84a95bd5e56a91`. What I confirmed:

- The per-agent `BUZZ_AGENT_MAX_OUTPUT_TOKENS=1024` now reaches the spawned `buzz-acp` / `buzz-agent` (verified on their live process env). Previously the harness always ran `4096` — that clobber is gone.
- The prior failure is gone: no more `503 … all 0 tunnel(s)` from the context fit check. Mesh requests succeed, and a direct completion through the same ingress at `max_tokens=1024` returns `200`.
- The added unit test (`apply_relay_mesh_env_preserves_user_max_output_tokens`) passes.

One caveat so I'm not overstating it: the agent did not post a full channel reply in my run. Inference succeeded (no 503), but SmolLM2-135M didn't produce a postable response — that looks like tiny-model/harness-quality behavior, separate from this PR's fix.

Net: the env-precedence change resolves the clobber and unblocks small-context shared-compute agents at the routing/sizing layer. Happy to run further checks on the testbed if useful.

Chessing234:
nice, thanks for validating on the two-node mesh setup

Chessing234:

Chessing234:

PR #2695, opened 2026-07-24
fix(desktop): apply the configured model to Claude Code agents via ANTHROPIC_MODEL

### LINKED ISSUE #2692: [Bug] Claude Code agents ignore the model set in Buzz Desktop — model picker persists but nothing applies it at spawn
**Describe the bug**

**Root cause**

```rust
model_env_var: None,
provider_env_var: None,
provider_locked: true,
```

**Suggested fix**

**Workaround**

**To Reproduce**

**Environment**

bosoud:
Fixes #2692

## Problem

Picking a model for a **Claude Code** agent in the desktop UI persists the selection but never applies it. The `claude` entry in `KNOWN_ACP_RUNTIMES` declares `model_env_var: None` and `supports_acp_model_switching: false`, so `runtime_metadata_env_vars` has no channel to carry the saved model into the spawned session — the agent silently runs on the user's global Claude Code default. Goose (`GOOSE_MODEL`) and Buzz Agent (`BUZZ_AGENT_MODEL`) both have this channel; claude was the only harness with a model picker but no application path behind it.

Observed in the wild: a stock agent set to Sonnet answering as `claude-fable-5`, across restarts, with no error anywhere.

## Changes

- **`discovery.rs`**: the claude runtime declares `model_env_var: Some(ANTHROPIC_MODEL_ENV_KEY)`. The Claude CLI honors `ANTHROPIC_MODEL` as a session model override (verified: `ANTHROPIC_MODEL=claude-sonnet-5 claude --print "<model id?>"` → `claude-sonnet-5`), and the ACP adapter's child `claude` process inherits the spawn env. `provider_locked` stays `true`, so only the model is injected — no provider env is written for this harness.
- **`env_vars.rs`**: new `ANTHROPIC_MODEL_ENV_KEY` const as the single source of truth for the key — referenced by the runtime entry, the `DERIVED_PROVIDER_MODEL_ENV_KEYS` list (so a stale persisted copy can't shadow the structured model field after a UI edit, same treatment as the other derived keys), and the readiness provider-model mapping in `readiness.rs`, which already used the literal `"ANTHROPIC_MODEL"` for the anthropic provider. Three call sites, one definition, no drift.
- **Tests**: pin the claude runtime shape (model env declared, provider locked), the derived-key classification (including case-insensitivity), and the claude-shaped `runtime_metadata_env_vars` injection. The tests intentionally assert the literal string so the const's value itself is pinned.

## Notes for reviewers

- Users who worked around this by setting `ANTHROPIC_MODEL` manually in the agent's env vars: on the next provider/model edit the derived-key rule strips the manual copy, and the structured field takes over — which is the intended convergence.
- Local `cargo test` for `desktop/src-tauri` could not be run on my machine (Intel mac — `sherpa-onnx-sys` has no usable static lib here, unrelated to this change); CI's Desktop Core job is the authoritative check for this PR.

https://claude.ai/code/session_01CqMxL1inNtU7icywZxbrGY

### DIFF
--- a/desktop/src-tauri/src/managed_agents/discovery.rs
+++ b/desktop/src-tauri/src/managed_agents/discovery.rs
@@ -111,7 +111,11 @@ const KNOWN_ACP_RUNTIMES: &[KnownAcpRuntime] = &[
         adapter_install_hint: "Install the Claude Code ACP adapter via npm.",
         skill_dir: Some(".claude/skills"),
         supports_acp_model_switching: false,
-        model_env_var: None,
+        // The Claude CLI honors this env var as a session model override,
+        // and the adapter's child `claude` process inherits the spawn env —
+        // without this, the model picked in the UI is persisted but never
+        // applied (#2692).
+        model_env_var: Some(crate::managed_agents::ANTHROPIC_MODEL_ENV_KEY),
         provider_env_var: None,
         provider_locked: true,
         default_env: &[],
--- a/desktop/src-tauri/src/managed_agents/discovery/tests.rs
+++ b/desktop/src-tauri/src/managed_agents/discovery/tests.rs
@@ -1270,3 +1270,25 @@ fn test_install_shell_from_some_returns_path() {
         "install_shell_from(Some) must return the path as Ok"
     );
 }
+
+/// The claude runtime must declare a model env channel — without one, the
+/// model picked in the UI is persisted but never applied at spawn (#2692).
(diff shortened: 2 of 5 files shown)


bosoud:
Closing in favor of #2701, which now carries the shared-const approach and the runtime-shape regression test. Thanks for folding those in @Chessing234 — one PR with everything beats two with halves. (For anyone landing here from #2692: the fix ships via #2701.)

PR #2696, opened 2026-07-24
fix(desktop): apply Claude models and run Windows installers natively

### LINKED ISSUE #2688: [Bug] Agent Runtimes cant be installed or activated
**Describe the bug**
Whenever I try to install either codex or claude it fails after 3 attempts.

**To Reproduce:**
1. Go to User
2. Click on Settings (ctrl+,)
3. Click on Agents on the left
4. Scroll down to Agent Runtimes
5. Try to flip either Claude or Codex to enabled
6. See error below.

**Expected behavior:**
I expected the harness to install, but see error message below.

**Supporting Material**

Step "cli" failed: install failed after 3 attempts (retried with backoff)
tar (child): Cannot connect to C: resolve failed

gzip: stdin: unexpected end of file
/usr/bin/tar: Child returned status 128
/usr/bin/tar: Error is not recoverable: exiting now
Downloaded Codex package archive did not contain the expected package layout.
At line:820 char:21
+ ... throw "Downloaded Codex package archive did not contain t ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo : OperationStopped: (Downloaded Code...package layout.:String) [], RuntimeException
+ FullyQualifiedErrorId : Downloaded Codex package archive did not contain the expected package layout.

 - OS: Windows
 - 11

**Additional context**
This is no matter if you try to install the harnesses via the install of app or within settings.

### LINKED ISSUE #2692: [Bug] Claude Code agents ignore the model set in Buzz Desktop — model picker persists but nothing applies it at spawn

tagadearpit:
## Summary

- inject `ANTHROPIC_MODEL` from the structured Claude Code model picker at spawn time
- treat `ANTHROPIC_MODEL` as derived configuration so stale custom environment values cannot override later picker changes
- run catalogued Windows PowerShell installers directly instead of routing them through Git Bash, preventing MSYS path conversion from corrupting native installer paths
- add regression coverage for model injection and native PowerShell command construction

## Root cause

Claude Code's runtime catalog had no model environment variable, so the picker value was saved but never reached the spawned Claude process. On Windows, Claude and Codex installer scripts were invoked through Git Bash; MSYS path conversion could turn `C:` paths into invalid arguments for native tools such as `tar`.

Closes #2692
Closes #2688

Issue #2690 is intentionally not duplicated here: upstream PR #2694 already addresses the macOS artifact-name collision.

## Validation

- `git diff --check`
- verified upstream comparison: one commit ahead of `block/buzz:main`, with only the five intended files
- targeted Rust tests could not link in this Linux workspace because `pkg-config` and the required GLib/GObject development libraries are unavailable; CI should run the full platform matrix.

### DIFF
--- a/desktop/src-tauri/src/commands/agent_discovery.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery.rs
@@ -536,21 +536,36 @@ fn persist_last_error_on_install(
     save_managed_agents(app, &records)
 }
 
-/// Build a login-shell `Command` for `command` with hermit env vars stripped,
-/// Buzz-managed npm locations set, and the user's PATH set. This is the
-/// single source of truth for
-/// the shell selection and environment cleanup shared by `run_install_command`
-/// and managed npm install path — keeping them in sync so the hermit-strip list
-/// can't drift between command execution paths.
+/// Build an install `Command` for `command` with hermit env vars stripped,
+/// Buzz-managed npm locations set, and the user's PATH set. This is the single
+/// source of truth for the shell selection and environment cleanup shared by
+/// `run_install_command` and managed npm install paths — keeping them in sync
+/// so the hermit-strip list can't drift between command execution paths.
 ///
-/// On Windows, resolves Git Bash via `resolve_bash_path` (skips `BUZZ_SHELL`
-/// since install commands require bash syntax). Returns `Err` when no shell
-/// can be found.
+/// On Windows, known PowerShell install commands run directly in PowerShell.
+/// Sending them through Git Bash enables MSYS path conversion, which corrupts
+/// paths consumed by native installers (for example, `C:` passed to `tar`).
+/// Other install commands use Git Bash via `resolve_bash_path`.
 fn install_shell_command(command: &str) -> Result<std::process::Command, String> {
-    let shell: std::path::PathBuf = resolve_install_shell()?;
+    #[cfg(windows)]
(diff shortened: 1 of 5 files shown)


wpfleger96:
🤖 hey @tagadearpit — thank you for this contribution. Closing it, with a note on each half, since it's really two changes and they land differently.

The Windows installer half is superseded by #2750 (`f3981dbfe`). Main's `is_powershell_command` does a case-insensitive check on the first token, so it matches any PowerShell invocation, where your `powershell_install_script` matched one exact literal command prefix. Same intent, and main's version handles more shapes of the command.

The Claude half — setting `model_env_var: Some("ANTHROPIC_MODEL")` and adding it to `DERIVED_PROVIDER_MODEL_ENV_KEYS` — is a separate question, and it's still unaddressed on main (`claude` currently has `model_env_var: None`). I don't want to silently drop it, but I also don't want it merged as a side effect of a Windows fix, since it changes model-selection behavior for every Claude user and deserves review on that basis. Could you open it as its own PR with a short note on what it fixes for you? That one I'll get proper eyes on.

PR #2703, opened 2026-07-24
fix: Claude live models, reply delivery prompt, avatar flicker

### LINKED ISSUE #2581: Claude Code harness shows "Could not load live models for this provider"
## Description
When selecting the **Claude Code** ACP harness (`buzz-acp`), the model selector doesn't load live models and instead shows:

> Using built-in model options. Could not load live models for this provider.

## Steps to reproduce
1. Open the harness/agent selector.
2. Choose "Claude Code" as the harness.
3. Observe the model dropdown.

## Expected behavior
Live models for the Claude Code provider load normally, or (if Claude Code manages its own model selection) no error/fallback message is shown.

## Actual behavior
The message "Using built-in model options. Could not load live models for this provider." appears and the harness falls back to a static/built-in model list.

## Environment
- Repo: block/buzz
- Component: buzz-acp (ACP harness — Claude Code)
- Date observed: 2026-07-23

## Additional context
Happy to provide logs or a screenshot if that would help debug.

### LINKED ISSUE #2665: [Bug] User avatar briefly disappears and falls back to monogram circle
*Posted by Stretch, a Buzz agent, on behalf of @dmnyc.*

## Describe the bug

The user's own profile avatar briefly disappeared from the UI and was replaced by a monogram circle (the colored circle with the user's initials). The avatar returned on its own after the user navigated to their settings screen. This is distinct from the agent avatar issue tracked in #2576, which affects custom agent avatars after a restart.

## To reproduce

1. Use Buzz Desktop with a profile avatar set (confirmed displaying correctly).
2. At an unspecified point during normal use (no restart or explicit profile change), observe the avatar in the UI.
3. Avatar disappears and is replaced by the monogram circle fallback.
4. Navigate to the user's settings screen.
5. Avatar returns and displays correctly again.

## Expected behavior

The user's own profile avatar should remain visible at all times once set. It should not fall back to the monogram circle without an explicit profile change, and should not require visiting the settings screen to recover.

## Environment

- **OS:** macOS
- **Buzz version:** current (observed 2026-07-23)
- **Application type:** Desktop app

## Notes

- This affects the *user's* avatar, not agent avatars.
- Recovery occurred by visiting the settings screen — suggests the avatar URL is still known to the client but something in the render/cache path is dropping it transiently.
- Likely related to the same class of avatar-loading bug as #2576 but a separate symptom and possibly a separate root cause.

## Related issues

- #2576 — Custom agent avatars missing on Agents screen after Buzz restart (agents, not user).
- #2366 — Custom agent restarts and loses custom icon across multiple communities.
- #2515 — Agent/profile collisions across communities causing missing avatars.

### LINKED ISSUE #2698: [Bug] buzz-acp reply delivery is implicit — Sonnet-class agents answer in session text and every reply is silently dropped
**Describe the bug**

A buzz-acp managed agent's reply reaches the channel only if the model *spontaneously decides* to run `buzz messages send`. Nothing in the injected prompts states that session text is not delivered, and the harness has no fallback when a turn ends without a send — the reply is silently dropped. Strong models happen to infer the convention; weaker ones don't, so **switching an agent's model can silently break all of its replies** while everything looks healthy (turn runs, Activity panel streams the answer, tokens are billed, chat stays empty, exit clean, nothing logged).

**Observed** (Buzz Desktop 0.4.24, Claude Code harness, macOS):

The same agent, same 12,435-char base prompt (byte-identical), same `[Buzz event: @mention]` turn block — only the model differs (via `ANTHROPIC_MODEL`, per the #2692 workaround):

| Session model | Behavior | Chat |
|---|---|---|
| `claude-fable-5` (3 sessions) | ends turn with `Bash: printf '…' \| buzz messages send …` | reply appears ✅ |
| `claude-sonnet-5` (3 sessions) | zero tool calls — answers in plain session text | reply lost ❌ |

The Sonnet reply text is visible in the desktop Activity panel (streamed ACP output) and in the session transcript, so generation succeeded; delivery never happened. The owner sees "working…" then silence — same symptom family as #2459, but here the CLI write doesn't *fail*, it is never attempted.

**Root cause**

The delivery contract is implicit. The base prompt says the `buzz` CLI is "your primary interface" and discusses threading/mentions, but never states "your session text is NOT delivered — reply via `buzz messages send`". The per-turn block (`queue::format_prompt` / `format_event_block`, `crates/buzz-acp/src/queue.rs`) describes the incoming event and names a reply destination without stating the mechanism. Turn completion (`crates/buzz-acp/src/lib.rs`) does not check whether any send occurred for the triggering channel.

**Suggested direction**

1. **State the contract explicitly** in the turn prompt (one line: session text is not delivered; reply with `buzz messages send --channel <uuid> [--reply-to <event-id>]`). Cheap, model-agnostic, immediately effective — verified locally: adding the equivalent instruction to the agent's system prompt makes Sonnet 5 post correctly.
2. **Add a harness fallback**: when a prompted turn completes and no `buzz messages send` targeting the triggering channel was observed, either auto-post the final ACP text as the reply (best) or at minimum log at WARN and surface a delivery-failure indicator to the owner. This also covers #2459's failed-write case — silent loss is the real bug in both.

**To Reproduce**

1. Managed Claude Code agent; set `ANTHROPIC_MODEL=claude-sonnet-5` in its env vars (needed because of #2692); restart.
2. DM the agent any question.
3. Activity panel streams a correct answer; no message ever arrives in the chat; harness log shows nothing between turn start and the next lifecycle event.

**Related**: #2459 (reply lost when CLI write fails), #2265 (silent failure when default model lacks credits), #2692 (model setting not applied — the fix that exposes this one, since it makes cheaper models reachable).

**Environment**: Buzz Desktop 0.4.24 (Block-signed), `@agentclientprotocol/claude-agent-acp` 0.60.0, macOS Intel, relay `communities.buzz.xyz`.

Chessing234:
## Summary
- Claude Code model discovery falls through to `buzz-acp models` when no Anthropic API key is set
- Turn/base prompts state that session text is not delivered — reply via `buzz messages send`
- Self avatars keep the cached image while the media proxy warms up

## Test plan
- [x] `just ci`
- [ ] Claude Code harness model picker loads live models without `ANTHROPIC_API_KEY`
- [ ] Sonnet-class agent replies via CLI after a mention (not only Activity panel text)
- [ ] Self avatar stays visible across community switch / brief proxy blip

Closes #2581
Closes #2698
Closes #2665

### DIFF
--- a/crates/buzz-acp/src/base_prompt.md
+++ b/crates/buzz-acp/src/base_prompt.md
@@ -62,7 +62,7 @@ All replies and delegations — including task assignments to other agents — g
 ### General
 
 - Respond promptly to @mentions. Be direct — no preamble. Name what you did, what you found, or what you need.
-- **If your turn produced anything worth knowing, you MUST publish it.** Use `buzz messages send`. Your reasoning and tool calls are invisible — a result, an answer, a deliverable, a decision, a blocker, or a question you need answered exists only if you published it. Work or an answer that someone asked you for always counts. Ending that kind of turn without a message is a silent failure.
+- **If your turn produced anything worth knowing, you MUST publish it.** Use `buzz messages send`. Session text is not delivered to the channel — only a successful `buzz messages send` posts a reply. Your reasoning and tool calls are invisible — a result, an answer, a deliverable, a decision, a blocker, or a question you need answered exists only if you published it. Work or an answer that someone asked you for always counts. Ending that kind of turn without a message is a silent failure.
 - **If a human asked you something, you MUST reply to them** — even if the reply is only that you have nothing to add or nothing to do. Never leave a person waiting on you.
 - **Otherwise, publishing is optional and silence is usually correct.** When a message leaves you nothing new to contribute, end the turn without publishing. That is a success, not a failure.
 - **After a context compaction or session restart, resume silently** — rebuild state from your todos, memory, and the thread, and never post a message announcing the compaction, summarizing what was lost, or asking how to proceed.
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -3625,6 +3625,16 @@ mod agent_draft_prompt_tests {
         assert!(prompt.contains("single-quoted shell strings preserve `\\n` literally"));
         assert!(prompt.contains("buzz messages send ... --content -"));
     }
+
+    #[test]
+    fn shared_base_prompt_states_session_text_is_not_delivered() {
+        let prompt = include_str!("base_prompt.md");
+        assert!(
+            prompt.contains("Session text is not delivered to the channel"),
+            "base prompt must state that session text is not a channel reply"
+        );
+        assert!(prompt.contains("buzz messages send"));
+    }
 }
 
 fn default_heartbeat_prompt() -> String {
--- a/crates/buzz-acp/src/queue.rs
+++ b/crates/buzz-acp/src/queue.rs
@@ -1148,9 +1148,10 @@ pub(crate) fn format_event_block(
 /// top level.
 fn append_reply_instruction(s: &mut String, event_id: &str) {
     s.push_str(&format!(
-        "\nIMPORTANT: For ordinary replies in this turn, use `--reply-to {event_id}` \
-         on `buzz messages send` so the conversation stays threaded. \
-         If the human explicitly asks for a channel-root, top-level, \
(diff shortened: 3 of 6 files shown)


Chessing234:

Chessing234:
closing this one too — three unrelated fixes in one branch, stale since july and now conflicting with main. happy to reopen the claude model fallback on its own branch if it's still useful

Issue #2648, opened 2026-07-23
[Bug] Second Desktop install creates duplicate agent identities, inflating channel membership


sj-unit72:
**Describe the bug**

When the same user installs Buzz Desktop on a second machine and joins the same community, the second install **creates its own set of default agents with fresh Nostr keypairs** and adds them all to the shared channel. The agents carry identical display names to the ones already there, so the channel roster shows each agent multiple times.

Our Welcome channel shows **8 members** for what is really **1 human + 3 agents**.

Single self-hosted community, fresh install — so this is not the cross-community collision in #2515.

**To Reproduce:**

1. Install Buzz Desktop on machine A, join a community, let the default agents (Fizz, Honey, Bumble) be created, and add them to a channel.
2. Install Buzz Desktop on machine B and sign in to the **same** community as the same user.
3. Open the same channel and run `buzz channels members --channel <uuid>`.

**Expected behavior:**

An agent should have one identity and one channel membership per community, regardless of how many machines the owner installs on. A second install should register as an additional *runtime* for the existing agents rather than minting parallel same-named identities.

**Supporting Material**

Channel roster — `buzz channels members`, each pubkey resolved with `buzz users get`:

| Display name | Pubkey (first 8) | Role | Has ever posted |
|---|---|---|---|
| Serge | `bab2c2c9` | owner | yes |
| Fizz | `c9afa13c` | **member** | yes |
| Fizz | `614f28ba` | bot | yes |
| Fizz | `4ecf356f` | bot | yes |
| Honey | `c6cd9fae` | bot | yes |
| Honey | `c4125e50` | bot | **no** |
| Bumble | `8f0a2aa1` | bot | yes |
| Bumble | `90401f8c` | bot | **no** |

Six `bot` identities = 2 installs x 3 default agents. The seventh agent identity is an extra `Fizz` carrying role `member` rather than `bot`, which suggests a second creation path that disagrees on the role to assign.

Message counts with first/last post per pubkey, aggregated across all channels:

```
c9afa13c  n=11  22:04:20 - 22:40:19   Fizz    (role: member)
bab2c2c9  n=32  22:08:11 - 23:28:45   Serge   (human)
614f28ba  n= 7  22:52:44 - 23:18:30   Fizz    \
c6cd9fae  n= 1  23:12:28              Honey    >  install A
8f0a2aa1  n= 1  23:12:33              Bumble  /
4ecf356f  n= 5  23:19:50 - 23:27:07   Fizz    \
[...]

PR #2705, opened 2026-07-24
fix: avatar fallback, custom emoji remove, Welcome Team roster

Chessing234:
## Summary
- Keep the cached avatar visible after a live URL blip instead of sticking on the monogram (#2665)
- Match custom emoji reactions by normalized shortcode when removing (CLI + desktop)
- Don’t add Welcome Team bots to a channel when those names are already members (#2648)

## Test plan
- [ ] Trigger a transient avatar load failure with a local data URL present → monogram should not stick
- [ ] `buzz reactions add --emoji party --emoji-url …` then `buzz reactions remove --emoji party`
- [ ] Second Desktop install against a community that already has Fizz/Honey/Bumble → channel roster does not double


Chessing234:

Chessing234:

PR #2706, opened 2026-07-24
feat(desktop): show each agent's role in the @-mention selector

### LINKED ISSUE #2699: feat(desktop): show each agent's role/description in the @-mention selector
## Problem

The `@`-mention selector lists agents by **display name + avatar only** — there's nothing to tell you what each agent is *for*. In a workspace with several agents (especially a themed set that share a visual style), they're near-indistinguishable at the point of use. You have to already know each agent's role from memory, or scroll back to whenever they introduced themselves.

Concrete case: a hive with three bee agents — Fizz (builder), Honey (writer), Bumble (researcher) — renders as three near-identical rows in the selector. A user hovering the composer can't tell who to call for "debug this" vs "summarize this thread" vs "go research X." Their words: *"we have 3 bee agents, but it's so hard to tell which one does what. Whom should I call for when I need to do something?"*

mention selector — three agents, no role labels

## Proposed UX

- In the `@`-mention autocomplete row, show a **one-line role/description** under (or beside) each agent's display name — same pattern the slash-command autocomplete in #2528 proposes for command descriptions.
- Source the line from the agent's public kind:0 profile `about` field. Fall back to nothing (name only, as today) when `about` is empty, so this degrades gracefully.
- Keep it to one line, truncated — the selector is a quick picker, not a profile card.

## Data-source note

- Agents don't reliably publish an `about` today (observed kind:0 profiles carry only `display_name` + `picture`). Surfacing the field is only useful if agent creation seeds a short role string into `about`. Worth pairing this with a default: when an owner drafts an agent, derive a one-line `about` from its purpose/system-prompt.
- Related: #2534 — kind:0 profile republish drops every field outside a 5-key allowlist. If `about` isn't in that allowlist, it'll get silently stripped on republish and this feature will show blanks. That allowlist needs `about` preserved for this to hold.

## Non-goals

- Full agent profile cards / hover popovers (a bigger surface; this is just the inline picker line).
- Slash-command discovery — that's #2528.

## Workaround today

Agents can self-describe in chat, but that's not discoverable at mention time. A cheat-sheet has to be re-asked for.

---
Reported by a Buzz user (Glomopay hive) via their in-workspace researcher agent. Screenshot is the user's own composer.

webdevtodayjason:
Closes #2699.

## Problem

The `@`-mention selector lists agents by display name + avatar only. In a workspace with several agents that share a visual style (the issue's example: three bee agents — Fizz, Honey, Bumble), the rows are near-indistinguishable at the point of use: nothing says who builds, who writes, who researches.

## Approach

Surface each agent's public kind-0 `about` as a one-line role under its display name in the selector. It renders inside the meta line that already carries the bot icon and the "managed by …" suffix, so row geometry is unchanged — and an empty `about` renders exactly today's row (bot icon + literal "agent"), the graceful degradation the issue asked for. Humans never get a role line, even when their profile has an `about`.

Data path (all client-side, no relay changes):

- The desktop's Tauri profile-summary conversions (`users_batch_from_events`, `user_search_result_from_event`) already parsed kind-0 content but dropped `about`. They now carry it, and the TS layer maps it through `UserProfileSummary` / `UserSearchResult` (the field was previously `Omit`-ed in `tauriProfiles.ts`).
- `useMentions` already batch-resolves owner profiles for the "managed by" label. Agent candidates whose `about` isn't already known get the same treatment via `useUsersBatchQuery` — backed by its per-pubkey delta cache — so the role line resolves even for agents who haven't authored anything in the loaded timeline.
- Suggestion mapping prefers an `about` resolved at candidate-build time (user-search results carry it inline), else falls back to the profile lookup — mirroring the existing `avatarUrl` fallback. A multi-line `about` collapses to a single line; the full text lives in the row's `title` tooltip.
- The E2E mock bridge summaries now include `about` to stay faithful to the Rust response shape.

**Known limitation** (flagged in the issue): agents don't reliably publish an `about` today, and the #2534 republish allowlist can strip the field. Affected rows simply degrade to the current name-only presentation — when those land, this UI lights up with no further changes.

## Tests

- **Rust**: `users_batch_carries_about_when_present`, `user_search_result_carries_about_when_present` (present → carried, absent → `None`).
- **JS unit**: 6 new tests on the suggestion mapping — candidate-first precedence, profile-lookup fallback, humans never get a description, multi-line collapse, whitespace-only → null.
- **Playwright**: new `mention-descriptions-screenshots.spec.ts` asserts the role lines, the no-`about` fallback row, the human row, and that a long `about` truncates on one line (scrollWidth > clientWidth) with the full text in the tooltip. The existing 50 `mentions.spec.ts` + `team-mentions.spec.ts` smoke tests pass unchanged.
- `just ci` passes locally. (No relay/db/auth changes, so the integration suite is unaffected.)

Screenshots in the comment below.

## Checklist

- [x] `just ci` passes (fmt + clippy + unit tests + mobile)
- [x] New public fields have doc comments
- [x] No new `unsafe` blocks

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -58,6 +58,7 @@ export default defineConfig({
         "**/composer-link-shortcut.spec.ts",
         "**/composer-tooltip-dismiss.spec.ts",
         "**/mentions.spec.ts",
+        "**/mention-descriptions-screenshots.spec.ts",
         "**/team-mentions.spec.ts",
         "**/persistent-agent-audience.spec.ts",
         "**/relay-reconnect.spec.ts",
--- a/desktop/src-tauri/src/models.rs
+++ b/desktop/src-tauri/src/models.rs
@@ -48,6 +48,10 @@ pub struct UserProfileSummaryInfo {
     #[serde(default)]
     pub name: Option<String>,
     pub avatar_url: Option<String>,
+    /// Kind-0 `about` field — one-line role/description surfaced in the
+    /// @-mention selector for agents.
+    #[serde(default)]
+    pub about: Option<String>,
     pub nip05_handle: Option<String>,
     pub owner_pubkey: Option<String>,
     #[serde(default)]
@@ -65,6 +69,10 @@ pub struct UserSearchResultInfo {
     pub pubkey: String,
     pub display_name: Option<String>,
     pub avatar_url: Option<String>,
+    /// Kind-0 `about` field — one-line role/description surfaced in the
+    /// @-mention selector for agents.
+    #[serde(default)]
+    pub about: Option<String>,
     pub nip05_handle: Option<String>,
     pub owner_pubkey: Option<String>,
     #[serde(default)]
--- a/desktop/src-tauri/src/nostr_convert.rs
+++ b/desktop/src-tauri/src/nostr_convert.rs
@@ -341,6 +341,7 @@ pub fn users_batch_from_events(
                 .map(str::to_string),
             name: v.get("name").and_then(Value::as_str).map(str::to_string),
             avatar_url: v.get("picture").and_then(Value::as_str).map(str::to_string),
+            about: v.get("about").and_then(Value::as_str).map(str::to_string),
             nip05_handle: v.get("nip05").and_then(Value::as_str).map(str::to_string),
             is_agent: owner_pubkey.is_some(),
             owner_pubkey,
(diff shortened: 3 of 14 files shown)


webdevtodayjason:
### Agent role lines

Fizz, Honey, and Bumble each show their kind-0 `about` as a one-line role next to the bot icon; Buzzy (no `about`) keeps today's plain "agent" row.

01-agent-role-lines

### Long `about` truncates to one line

A long `about` stays on a single truncated line (full text in the hover tooltip) and the "managed by" label survives at the end of the row.

02-long-about-truncates

Issue #2707, opened 2026-07-24
[Bug] when using auto-tag agent, the cursor often slips back into the agents name, and typing breaks it


kieranx:
When using auto-tagging of agents so they stay tagged in chat - very often, after a few msgs, the cursor slips back into the agents name, and typing then breaks the tag and it stops working. Screenshot attached. MacOS.

<img width="788" height="340" alt="Image" src="https://github.com/user-attachments/assets/7f305c03-8a8e-4430-b9ab-d70b705dbb64" />

Chessing234:
Fixed in https://github.com/block/buzz/pull/2713 — caret can’t sit inside the agent chip anymore.

kieranx:
This is still outstanding (not live in the app yet) despite being fixed weeks ago - any ETA? Major QOL issue

Issue #2604, opened 2026-07-23
[Bug] Crash due to dynamically linked libraries in Appimage


evd0kim:
While using appimage-run on NixOs, the app crashes with

```
Uncompress Buzz_0.4.23_amd64.AppImage of type 02 @ offset 944632
[===========================================================================================================================================|] 3465/3465 100%
Buzz_0.4.23_amd64.AppImage is now installed in /home/i/.cache/appimage-run/679f0bfb1eb5287c15abc0da386d7b877a6f8302f020a80f239598e102564327
buzz-desktop: error while loading shared libraries: libzstd.so.1: cannot open shared object file: No such file or directory
```

*Steps to reproduce the behavior:*

1. Run on NixOs or in Nix shell
2. Launch appimage-run Buzz_0.4.23_amd64.AppImage 
3. App crashes

**Expected behavior:**

Running as other binaries.

**Environment (please complete the following information):**
 - OS: Linux NixOs 
 - 26.05

tw-360vier:
I also tried the appimage on NixOS.
I can get beyond these errors by adding some extra libs for appimage-run:

```nix
  programs.appimage.enable = true;
  programs.appimage.binfmt = true;
  programs.appimage.package = pkgs.appimage-run.override {
    extraPkgs = pkgs: [
      pkgs.zstd
      # libelf.so.1 (needed transitively via appimage-run's Mesa/LLVM GL stack).
      # NB: pkgs.libelf ships only a static libelf.a — the shared lib comes from
      # elfutils, whose `out` output buildFHSEnv pulls in automatically.
      pkgs.elfutils
    ];
  };
```

but then webkit still crashes and i only see a transparent window.

what works for me is re-packaging the debian package:

```nix
# Buzz desktop (https://github.com/block/buzz) — a Tauri/WebKitGTK app.
#
{
  lib,
  pkgs,
  ...
}:
let
  inherit (pkgs.stdenv.hostPlatform) isLinux isAarch64;

  version = "0.4.24";

  # Prebuilt binaries + desktop entry + icons, straight out of the .deb.
  buzz-unwrapped = pkgs.stdenvNoCC.mkDerivation {
    pname = "buzz-unwrapped";
    inherit version;
    src = pkgs.fetchurl {
      url = "https://github.com/block/buzz/releases/download/v${version}/Buzz_${version}_amd64.deb";
[...]

Issue #2515, opened 2026-07-23
[Bug] Agents from different communities collide in channels, causing duplicate @mention targets and missing avatars


jw-gsl:
## Describe the bug

When the same Buzz Desktop user belongs to two communities and has a personal agent with the same display name in each community, agent identity is not consistently scoped to the active community.

The Agents view shows only the agent belonging to the current community, but a channel in that community can contain both communities' agent identities. A single `@agent-name` mention can then produce multiple Nostr `p` tags, notifying both identities. In the observed case, the identity associated with the other community replied, and its profile picture did not render because its image URL referenced the other community's media host.

This leaves the user unable to understand or manage the collision from the current Agents view: only one agent is visible there, while the channel and mention resolver operate on two same-named identities.

Belonging to multiple communities should be supported. The bug is the leakage or retention of an agent identity from one community inside another community's channel and mention-resolution context.

## To Reproduce

The following is the observed sequence. The precise operation that first introduced the foreign agent membership has not yet been isolated:

1. Sign in to Buzz Desktop with a user who belongs to two separate Buzz communities.
2. Have a personal agent in each community with the same display name, for example a renamed default agent.
3. In Community B, open the Agents view and observe that only the Community B agent is shown.
4. Add or re-add the Community B agent to a channel in Community B.
5. Restart Buzz Desktop or switch between the two communities.
6. Return to the Community B channel and type `@<agent-name>`.
7. Send the message.
8. Inspect the channel membership and the sent Nostr event.
9. Observe that:
   - both the Community A and Community B agent identities are channel members;
   - one friendly-name mention produces `p` tags for both same-named identities;
   - the identity associated with Community A can reply in the Community B channel; and
   - its avatar may be blank because its profile image URL points to Community A's media host.

## Expected behavior

- Agents, profiles, memberships, runtimes and media should be scoped to the active community unless an explicit cross-community action is supported.
- A channel in Community B should not silently retain or add an agent identity from Community A.
- Selecting one agent in the mention picker should add exactly one `p` tag for that agent's immutable identity.
- Same-named agents should be disambiguated in the UI using community, owner, status and/or a stable unique handle.
- The current community's Agents view and channel membership controls should expose the same manageable set of agent identities.
- Switching communities or restarting Desktop should reconcile stale or foreign agent memberships rather than preserving an invisible collision.
- Profile pictures should not silently fail because an identity unexpectedly references another community's authenticated media host.

## Supporting Material

[...]

PR #2713, opened 2026-07-24
fix: agent chip caret, mention dedupe, AppImage zstd

Chessing234:
## Summary
- Keep the caret out of auto-tagged agent chips so typing doesn’t break `@Agent` (#2707)
- Resolve one `p` tag per `@Name` when same-named agents collide (#2515)
- Keep bundled `libzstd` in the AppImage for NixOS/appimage-run (#2604)

## Test plan
- [x] `just ci`
- [ ] Persist-tag an agent, click the chip / arrow into the name, type — tag stays intact
- [ ] Two same-named channel agents → one `@Name` → one `p` tag (managed preferred)


Chessing234:

PR #2720, opened 2026-07-24
fix(desktop): edit the selected agent instance

### LINKED ISSUE #2717: [Bug] Desktop: agent settings edits always target one community's instance - per-community agent config (env vars) unreachable from the UI
**Describe the bug**

The data model stores managed agents as one definition record per persona plus one instance record **per community/relay** (`managed-agents.json`: instances carry `relay_url` and their own `env_vars`), so per-community agent configuration is clearly supported by the backend. However, the Settings → Agents UI gives no way to target a specific community's instance:

- Each built-in agent (Fizz/Honey/Bumble) appears **once** in the agent list, even when it has instances in multiple communities.
- The edit dialog does not show which community/relay the instance being edited belongs to.
- Every edit we made (adding an env var to "Fizz") was written to the **same instance** (`wss://instance1.soldo.dev`) regardless of intent — repeated attempts to configure the `wss://instance2.soldo.dev` instance always landed on the other community's instance. There appears to be no path in the UI that edits the second instance at all.

From the user's perspective, agent settings behave as if they were global across communities, and it is impossible to configure the same persona differently per community.

Concrete motivating use case: `CLAUDE_CONFIG_DIR` per community, so agents in a work community bill against a work Claude subscription while agents in a personal community bill against a personal one. The instance-level `env_vars` field is exactly right for this — it's only the UI that can't reach it.

**To Reproduce:**

1. Join two communities (self-hosted relays A and B) with the built-in agents active in both — `managed-agents.json` now contains two instance records per persona, one per `relay_url`.
2. Open Settings → Agents, edit agent "Fizz", add an environment variable under Advanced, save.
3. Inspect `~/.local/share/xyz.block.buzz.app/agents/managed-agents.json`.

Observed: the env var is written to the instance for relay A (`wss://instance1.soldo.dev` in our case) — no matter which community is active in the app. There is no UI affordance to edit the relay-B instance instead, and no indication in the dialog of which instance was edited.

**Expected behavior:**

Either of:
- The agent list/edit dialog shows one entry per community instance (or a community picker inside the dialog), labeled with the community name/relay URL, so each instance's settings (env vars, model, respond-to, …) can be edited independently; or
- at minimum, the edit dialog displays which community's instance is being edited.

**Supporting Material**

State after edits — env var repeatedly landed on the `instance1.soldo.dev` instance while trying to configure the `instance2.soldo.dev` one:

```
$ python3 - <<'EOF'
import json
d = json.load(open('.local/share/xyz.block.buzz.app/agents/managed-agents.json'))
for a in d:
    print(a.get('name'), '| relay:', a.get('relay_url') or '(definition)', '| env_vars:', a.get('env_vars'))
EOF
Bumble | relay: (definition) | env_vars: None
Fizz   | relay: (definition) | env_vars: None
Honey  | relay: (definition) | env_vars: None
Bumble | relay: wss://instance1.soldo.dev  | env_vars: None
Bumble | relay: wss://instance2.soldo.dev | env_vars: None
Fizz   | relay: wss://instance1.soldo.dev  | env_vars: {'CLAUDE_CONFIG_DIR': '/home/user/.claude-work'}   <-- edits always land here
Fizz   | relay: wss://instance2.soldo.dev | env_vars: None                                                <-- intended target
Honey  | relay: wss://instance1.soldo.dev  | env_vars: None
Honey  | relay: wss://instance2.soldo.dev | env_vars: None
```

 - OS: Linux (Ubuntu 24.04 host, AppImage)
 - Version: Desktop v0.4.24 (`Buzz_0.4.24_amd64.AppImage`)
 - Communities: two self-hosted relays (`ghcr.io/block/buzz:main`), agents runtime: Claude Code

**Additional context**

Agent defaults (the global dialog) working app-wide is understood and fine — this report is specifically about per-instance settings being unreachable/mistargeted in the UI while the storage layer (`persona_id` + `relay_url` + per-instance `env_vars`) already models them correctly.

oshtz:
## Summary

- route Edit on an agent profile to the selected managed-agent instance instead of the shared persona
- show each instance's relay URL so same-persona deployments can be distinguished
- add regression coverage proving an environment-variable update targets only the selected pubkey

Fixes #2717

## Root cause

`UserProfilePanel.handleEditAgent` preferred `resolvedPersona` whenever one existed. Because managed agents are persona-backed, Edit always opened the persona-definition dialog and never reached the instance editor.

Persona definitions remain editable through the existing actions on the Agents card.

## Validation

- `pnpm typecheck`
- `pnpm build:e2e`
- focused integration E2E for two same-persona instances on different relays — 1 passed
- focused smoke E2E for persona-linked profile editing — 1 passed

### DIFF
--- a/desktop/src/features/profile/ui/UserProfilePanel.tsx
+++ b/desktop/src/features/profile/ui/UserProfilePanel.tsx
@@ -402,12 +402,14 @@ export function UserProfilePanel({
   });
 
   const handleEditAgent = React.useCallback(() => {
-    if (resolvedPersona) {
-      setPersonaDialogState(editPersonaDialogState(resolvedPersona));
+    if (managedAgent) {
+      setEditAgentOpen(true);
       return;
     }
-    setEditAgentOpen(true);
-  }, [resolvedPersona]);
+    if (resolvedPersona) {
+      setPersonaDialogState(editPersonaDialogState(resolvedPersona));
+    }
+  }, [managedAgent, resolvedPersona]);
 
   const { deleteManagedAgentRecord, deleteManagedAgentsForPersona } =
     useProfileAgentDeletion({
--- a/desktop/src/features/profile/ui/UserProfilePanelTabs.tsx
+++ b/desktop/src/features/profile/ui/UserProfilePanelTabs.tsx
@@ -406,8 +406,13 @@ function ProfileInstancesSection({
                 onClick={() => onOpenInstance(instance.pubkey)}
                 type="button"
               >
-                <span className="min-w-0 flex-1 truncate text-sm font-medium">
-                  {instance.name}
+                <span className="min-w-0 flex-1">
+                  <span className="block truncate text-sm font-medium">
+                    {instance.name}
+                  </span>
+                  <span className="block truncate text-xs text-muted-foreground">
+                    {instance.relayUrl}
+                  </span>
                 </span>
                 <span className="text-xs capitalize text-muted-foreground">
                   {isCurrent ? "Current" : instance.status.replace("_", " ")}
--- a/desktop/src/testing/e2eBridge.ts
+++ b/desktop/src/testing/e2eBridge.ts
@@ -78,6 +78,7 @@ type MockCommandAvailability = {
 export type MockManagedAgentSeed = {
   pubkey: string;
   name: string;
+  relayUrl?: string;
   avatarUrl?: string | null;
   personaId?: string | null;
   /** Harness/runtime id pin; `null` = inherit from persona (native default). */
@@ -2182,7 +2183,7 @@ function buildSeededManagedAgent(seed: MockManagedAgentSeed): MockManagedAgent {
     // Native serde always emits this key (`null` when unpinned) — the bridge
     // must mirror the wire shape, not omit the key.
     runtime: seed.runtime ?? null,
-    relay_url: DEFAULT_RELAY_WS_URL,
+    relay_url: seed.relayUrl ?? DEFAULT_RELAY_WS_URL,
     acp_command: "buzz-acp",
     agent_command: agentCommand,
     agent_args: agentArgs,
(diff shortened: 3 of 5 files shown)


oshtz:
### Before

Instances with the same persona are indistinguishable, and Edit opens the shared persona definition.

before-instances

before-edit-target

### After

Each instance shows its relay, and Edit opens the selected managed-agent instance.

after-instances

after-edit-target

djgeorg3:
Thanks for fixing this route. I independently reproduced the same root cause: for a persona-linked keyed managed agent, the profile Edit action opens the shared persona definition instead of the selected runtime instance. I arrived at the same production change on a local branch, so this is a review rather than a competing PR.

One thing that may block CI: `desktop/tests/e2e/global-agent-config-screenshots.spec.ts` isn't in this diff, but tests `10-edit-codex-custom-model-save-enabled-no-reason` (:819) and `11-edit-runtime-less-provider-required-save-blocked` (:908) both click `user-profile-edit-agent` on a keyed, persona-linked agent and then assert `persona-dialog` is visible. After this routing change they'll get `edit-agent-dialog`. Both pass again if you insert the instance dialog → **Edit avatar** hop before the existing assertions, leaving the provider/Save-gate/screenshot assertions untouched.

A regression case worth pinning: seed the persona definition with parallelism `1` and the selected managed instance with `10`; after profile Edit, expand **Advanced** and assert `#edit-agent-parallelism` reads `10`. That proves the instance record, not the definition, is driving the dialog. An assertion on dialog identity alone can't distinguish them.

Two adjacent paths worth explicit coverage: a persona-only profile still opens the definition editor, and a linked custom definition stays reachable from the instance dialog via **Edit avatar**.

Separately, `UserProfilePanel`'s persona-mode fallback still resolves `managedAgent` with `agents.find(a => a.personaId === persona.id)`. Any entry point that opens persona mode while instances exist will now route Edit to an arbitrary instance, the same shape as the symptom in #2717. Showing `relayUrl` per instance helps a lot here; it may be worth making that fallback explicit rather than first-match.

On my local equivalent of this change, the desktop unit suite was green (3,926 tests) and focused routing E2E passed 8/8. No backend, IPC, persistence, or relay behavior is involved.

PR #2742, opened 2026-07-24
feat(desktop): add local voice dictation to message composer

JoelCarriere:
## Summary

- add a microphone control to channel, DM, and thread message composers
- transcribe microphone audio with Buzz's existing local Parakeet STT model
- insert finalized speech at the current editor selection before sending through the normal message and ACP flow
- keep composer dictation on an isolated native PCM/STT path so microphone audio never enters the huddle relay
- expose accessible starting, recording, and finishing states and prevent sending until the final utterance has flushed
- share the low-level PCM capture plumbing with huddles without changing their push-to-talk behavior

## Why

ACP agents consume the text Buzz sends them, but the desktop composer had no voice-to-text input. Buzz already ships an on-device speech model for huddles, so composer dictation can reuse that local capability without adding a cloud transcription dependency or another API key.

## User impact

Users can dictate a draft, review or edit the resulting text, and then send it to people or connected agents through the existing message path. The first use may briefly wait for the local speech model download, and macOS will request microphone permission.

## Validation

- `pnpm -C desktop test` — 3,487 passed
- `pnpm -C desktop typecheck`
- `pnpm -C desktop build`
- `pnpm -C desktop check`
- focused dictation Rust tests — 2 passed
- pre-push desktop Tauri suite — 1,628 passed, 14 ignored
- Playwright dictation flow covering start, transcript insertion, stop/finalization, and send gating
- rendered desktop composer screenshot

## Manual validation

- [x] Confirmed physical-microphone transcription and the macOS permission prompt in the native Buzz Dev build

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -53,6 +53,7 @@ export default defineConfig({
         "**/file-attachment.spec.ts",
         "**/image-attachment-gallery.spec.ts",
         "**/composer-image-draw.spec.ts",
+        "**/composer-dictation.spec.ts",
         "**/video-attachment.spec.ts",
         "**/spoiler.spec.ts",
         "**/composer-link-shortcut.spec.ts",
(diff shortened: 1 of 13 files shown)


JoelCarriere:
## Screenshot

### Composer microphone control

Voice dictation microphone beside the Send button

PR #2498, opened 2026-07-23
fix: project agent mentions, mesh_demo flake, Builderlab sign-in timeout

Chessing234:
## Summary
- Project issue/PR @mentions publish as kind:9 with channel `h` (and read `buzz-channel`) so buzz-acp receives them (`#2462`)
- Raise mesh_demo Redis lease TTL to 60s so the lease-expiry path under load can't flip owned/forwarded mid-test (related to `#2458`; cancel-safety root cause tracked in `#2745` / CI archive gap in `#3461`)
- Cap Builderlab “Waiting for your browser…” with timeout, retry, and back-to-relay recovery (`#2484`)

## Test plan
- [x] Link a project to a channel (`buzz-channel`), @mention an agent on an issue — agent gets a turn
- [ ] `cargo test -p buzz-relay --lib` (default parallelism) — mesh_demo tests stay green
- [ ] Hosted community sign-in: kill/block Builderlab — UI times out with Retry / Back

Fixes #2462
Fixes #2484

Made with [Cursor](https://cursor.com)

PR #2745, opened 2026-07-24
fix(mesh): frame loss when recv_validated is cancelled mid-validation (#2458)

### LINKED ISSUE #2458: test(relay): mesh_demo echo test flaky under parallel suite — passes standalone, fails full-suite on clean main
`api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo` fails under the full parallel suite but passes standalone, on clean `main` (7e34bee6), reproduced twice each way:

```
# fails
cargo test -p buzz-relay --lib
# test api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo ... FAILED
# test result: FAILED. 719 passed; 1 failed; 27 ignored

# passes
cargo test -p buzz-relay --lib demo_join_forwarded_arm_round_trips_echo
# test result: ok. 1 passed; 746 filtered out
```

Environment: macOS (Darwin 25.3.0, M-series), default test parallelism. Behavior is consistent across runs and across two checkouts (main and a feature branch with no relay changes), so it looks like inter-test resource contention (port/state shared with another test in the binary) rather than machine-specific breakage. Same family as #2369.

Not blocking anything on my side — filing since it surfaces as a false red in any full local `cargo test -p buzz-relay --lib` run.

brocoppler:
## Summary

Root-causes the #2458 mesh_demo flake to a cancel-safety bug in production code and fixes it, plus one piece of endpoint hygiene the investigation surfaced.

**The bug:** `run_demo_echo` races `recv_validated` against its 100 ms drain tick in a `select!`. `recv_validated` reads a frame off the stream, then awaits Redis fence validation. When the tick fires inside that await, the future is dropped **with the frame it already consumed** — the stream then looks idle forever, the forwarding peer waits for an echo that can no longer happen, and the probe times out (the familiar 504). Loss probability per exchange ≈ validation latency ÷ 100 ms, which is why it worsens exactly when Redis is slow — i.e. under full-suite parallelism, matching #2458's "fails under `cargo test`, passes standalone" signature. Reproduced at 40–80 % locally, on an in-memory stream pair and live iroh endpoints alike, which rules the transport out as the cause.

## Changes

- **`tunnel/reliable.rs`** — split `validate_received` out of `recv_validated` so callers can extract frames cancel-safely and validate outside any `select!`/`timeout` region; the cancel-safety contract is documented on both.
- **`mesh_boot.rs`** — `run_demo_echo` now selects only over frame *extraction*; decoding + fence validation run outside the cancellable region. Drain semantics unchanged.
- **`buzz-relay-mesh/peer.rs`** — `IrohRecvHalf::recv_frame` is now resumable: partial-read state lives in the half rather than the future (each underlying `RecvStream::read` await is itself cancel-safe), so racing `recv_frame` cannot desync the stream either.
- **`buzz-relay-mesh/endpoint.rs`** — loopback binds disable the iroh portmapper. A loopback-bound socket can never transmit to a WAN mapping, but the gateway probe (UPnP/PCP/NAT-PMP) advertised one as a `Portmapped` candidate and path attempts to it fail (`sendmsg` → `EADDRNOTAVAIL`) — observed live during the investigation on a NAT-PMP-capable network. Production binds (`0.0.0.0`) are unaffected. Regression test included (trivially green on hosts without a portmap-capable gateway).
- **`api/mesh_demo.rs`** — the forwarded-arm round trip now runs over an in-memory `MeshStream` pair by default: deterministic, fast, and it reproduced the frame loss identically, making it the regression test for the fix. The live-iroh variant is kept as `demo_join_forwarded_arm_round_trips_echo_live_mesh` under `#[ignore]` for explicit transport evidence runs.

## Verification

| build | variant | runs | result |
|---|---|---|---|
| pre-fix | in-memory pair | 10 | **6 fail** (504) |
| pre-fix | live iroh | 5 | **3 fail** (504) |
| post-fix | in-memory pair | 15 | 15 pass |
| post-fix | live iroh (`--ignored`) | 15 | 15 pass |
| post-fix | full `cargo test -p buzz-relay --lib` | 3 | 744 passed ×3 |

`cargo clippy -p buzz-relay -p buzz-relay-mesh --lib --tests`: clean. `buzz-relay-mesh` suite: 33 passed.

## Relation to existing work

- Mitigates/closes the remaining mechanism behind #2458. The lease-TTL mechanism proven earlier in that thread is real but distinct; #2498's TTL raise addresses it for the live variant and composes cleanly with this change (this PR deliberately does not touch the TTL line).
- Supersedes my earlier "≤1 s threshold / pre-existing transport mechanism" interpretation posted on #2458/#2498 — the observations there were real, but the threshold framing was wrong: single runs were sampling a tick-alignment race. Corrections posted on those threads.

Fixes #2458

### DIFF
--- a/crates/buzz-relay-mesh/src/endpoint.rs
+++ b/crates/buzz-relay-mesh/src/endpoint.rs
@@ -28,10 +28,22 @@ impl MeshEndpoint {
         bind_addr: SocketAddr,
     ) -> Result<Self, MeshError> {
         let runtime_id = runtime_id_from_public_key(secret_key.public());
-        let endpoint = Endpoint::builder(iroh::endpoint::presets::Minimal)
+        let mut builder = Endpoint::builder(iroh::endpoint::presets::Minimal)
             .secret_key(secret_key)
             .alpns(vec![ALPN.to_vec()])
-            .relay_mode(RelayMode::Disabled)
+            .relay_mode(RelayMode::Disabled);
+        // A loopback-bound socket can never transmit to a WAN address, but the
+        // portmapper still probes the gateway (UPnP/PCP/NAT-PMP) and, on
+        // networks where a mapping succeeds, advertises the external address
+        // as a `Portmapped` direct-addr candidate. Path selection can then
+        // steer the connection onto that unsendable candidate mid-stream
+        // (`sendmsg` → EADDRNOTAVAIL) and frame delivery silently stops.
+        // Loopback reachability is loopback-only by definition — skip gateway
+        // probing for loopback binds.
+        if bind_addr.ip().is_loopback() {
+            builder = builder.portmapper_config(iroh::endpoint::PortmapperConfig::Disabled);
+        }
+        let endpoint = builder
             .bind_addr(bind_addr)
             .map_err(|err| MeshError::Transport(err.to_string()))?
             .bind()
@@ -146,6 +158,26 @@ mod tests {
                 .await
                 .unwrap();
         (a, b)
+    }
+
+    /// A loopback-bound endpoint must never advertise non-loopback direct
+    /// addrs. With the portmapper enabled, a NAT-PMP/UPnP-capable gateway
+    /// hands out a WAN mapping that is advertised as a `Portmapped`
+    /// candidate; path selection can then steer onto an address the
+    /// loopback-bound socket cannot send to (`sendmsg` → EADDRNOTAVAIL) and
+    /// frame delivery stalls (#2458). Trivially green on hosts without a
+    /// portmap-capable gateway; fails pre-fix on hosts with one.
+    #[tokio::test]
+    async fn loopback_bind_advertises_only_loopback_addrs() {
+        let ep = MeshEndpoint::bind(loopback_any()).await.unwrap();
+        // The portmap probe (when enabled) completes within ~a second of bind.
+        tokio::time::sleep(std::time::Duration::from_millis(1200)).await;
+        for addr in ep.ip_addrs() {
+            assert!(
+                addr.ip().is_loopback(),
+                "non-loopback direct addr advertised from loopback bind: {addr}"
+            );
+        }
     }
 
     async fn connected_pair() -> (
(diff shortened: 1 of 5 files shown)


mo0ooo:
Fresh latest-main validation: **CLEAR**.

I re-applied both PR commits (`e9e4ad31`, `eff239c0`) without modification on
top of `origin/main@c3132c3`. The five-file patch applied cleanly.

Reproduction on unmodified current main, with Redis enabled:

- `demo_join_forwarded_arm_round_trips_echo` repeatedly returned `504` even
  standalone and single-threaded.
- A direct owner receive surfaced the Data frame; the shipped
  `run_demo_echo` path opened the stream but lost the receive when its 100 ms
  drain tick cancelled `recv_validated` during Redis fence validation.

The PR's cancellation contract is also supported by the currently locked
transport source: `noq::RecvStream::read` is documented cancel-safe, while
`read_exact` is explicitly not. Persisting prefix/body bytes in
`IrohRecvHalf` and filling them with `read` is therefore the correct primitive
fix. It also protects `audio/join.rs::serve_control_loop`, which races raw
`recv_frame` against nonterminal roster events.

Post-fix verification on Rust 1.95.0, isolated Postgres + Redis:

- `cargo fmt --all -- --check`: pass
- `cargo test --locked -p buzz-relay-mesh -p buzz-relay`:
  - relay lib: **994 passed, 0 failed, 88 ignored**
  - relay binary: **13 passed**
  - relay-mesh: **33 passed**
  - doc tests: pass
- live iroh + Redis ignored smoke
  `demo_join_forwarded_arm_round_trips_echo_live_mesh`: **5/5 passed**
- `cargo clippy --locked -p buzz-relay-mesh -p buzz-relay --all-targets --
  -D warnings`: pass

The original Broc Oppler author/DCO sign-off and Claude co-author attribution
remain correct. I recommend merging this PR rather than opening a duplicate
refresh PR; GitHub currently reports it MERGEABLE with review as the remaining
gate.

PR #2756, opened 2026-07-24
feat(transport): pluggable Transport trait — relay, remote bridge, and in-memory implementations

bluepinkneon:
## What

Makes the event pipe pluggable. This PR introduces `crates/buzz-transport` — a `Transport` trait whose contract is **a bidirectional stream of signed events, nothing else** — plus three implementations:

- **Nostr relay** — `HarnessRelay` (the existing NIP-42 relay client in `buzz-acp`) now implements `Transport`, making today's relay connection just one implementation of the seam.
- **`RemoteTransport`** — dials a user-supplied bridge endpoint speaking a small documented JSON protocol ([`PROTOCOL.md`](crates/buzz-transport/PROTOCOL.md)), so anyone can bridge Buzz onto their own network (Slack, a private mesh, anything) in any language.
- **`memory::InMemoryHub`** — the reference implementation: a working in-process hub where participants exchange verified, `h`-tag-routed signed events with no I/O. It pins the seam contract (verification at publish, channel scoping, subscription matching, replace-on-resubscribe, no self-echo) and gives consumers a real stream to integration-test against. The scripted `MockTransport` (feature `test-utils`) remains for call-log assertions.

## Why

Today the harness is hard-wired to the Buzz relay socket. Everything a consumer needs already rides that socket as signed events — messages, membership, discovery, ephemeral signals — so the seam is naturally *one* stream. Extracting it as a trait lets people write their own network backend without touching Buzz internals, and without inventing REST side-channels.

## Design decisions

**Crate-owned `SignedEvent`, not `nostr::Event`.** The public API carries a plain-data struct (`id`, `pubkey`, `created_at`, `kind`, `tags`, `content`, `sig`). The nostr library is an internal detail used for verification and conversion; implementors of the trait and of bridges never need it. Inbound events are verified (id hash + Schnorr sig) and silently dropped on failure — a tampered event never reaches the consumer.

**No handshake at the seam.** The `Transport` trait is subscribe/publish/next_event/reconnect — a stream. The `hello`/`hello_ack` exchange exists only inside the remote bridge wire protocol, where it carries the consumer pubkey (for `require_mention`), the bearer token on raw-socket carriers, and a crisp version-mismatch error.

**Bridge protocol v1** (`PROTOCOL.md`): JSON frames, ≤524,288 bytes (the relay's default frame cap, so any event the relay accepts fits), forward-compatible (unknown frame types and unknown fields are ignored; breaking changes bump `version`). `ok`/`eose` are optional — the stream is fire-and-forget like the relay socket it mirrors. Channel discovery and membership ride the stream as events (kinds 39000/39002), not as APIs.

**Carriers.** The bridge URL scheme selects the carrier:

| Scheme | Carrier | Framing |
|---|---|---|
| `wss://` / `ws://` | WebSocket | one JSON object per text message |
| `unix:///path.sock` | Unix domain socket | NDJSON — one LF-terminated frame per line |

WebSocket carriers can additionally be tunneled through a **SOCKS5 proxy** (`socks5://[user:pass@]host:port`, RFC 1928 + RFC 1929 hand-rolled — no new heavy deps). Domain names are passed to the proxy verbatim (`socks5h` behavior), so overlay-only names like `.onion` resolve inside the private network.

**Security policy.** `wss://` always allowed. Plaintext `ws://` only to loopback, or to `.onion` through a **loopback** SOCKS5 proxy (Tor's encryption starts at the proxy, so the proxy must be on-machine), or with explicit `allow_insecure` opt-in — plaintext never silently leaves the machine. `unix://` is local by construction (filesystem permissions are the ACL). Bearer token travels as an `Authorization` header on WebSocket upgrade *and* in the `hello` frame (raw-socket carriers have no headers).

## Wiring

`buzz_acp::transport::connect_transport` selects the implementation from the environment:

| Variable | Values |
|---|---|
| `BUZZ_TRANSPORT` | `nostr` (default) \| `remote` |
| `BUZZ_TRANSPORT_URL` | `wss://…` \| `unix:///…` |
| `BUZZ_TRANSPORT_TOKEN` | optional bearer token |
| `BUZZ_TRANSPORT_SOCKS_PROXY` | `socks5://[user:pass@]host:port` |
| `BUZZ_TRANSPORT_ALLOW_INSECURE` | `true`/`1` |

## Testing

- 26 unit tests (frame codec, SignedEvent round-trip + verification, SOCKS5 URL parsing, TLS policy, publish buffering, `h`-tag routing, dyn-compatibility).
- 8 remote-bridge integration tests against in-process bridges: full round-trip over WebSocket, over a Unix socket, and through a real in-process SOCKS5 proxy (both no-auth and username/password, including wrong-password rejection and a tunnel-count assertion); version-mismatch handshake failure; plaintext-to-remote rejected before dialing (direct and proxied); reconnect re-subscribes and flushes buffered publishes.
- 5 in-memory integration tests driving multiple participants through `Box<dyn Transport>`: fan-out between three participants (with receive-side verification and no self-echo), kind + mention filters, unsubscribe/re-subscribe replacement, tampered/unroutable events rejected at publish, reconnect.
- 1 end-to-end test of the whole seam: two `RemoteTransport` clients speaking protocol v1 over real WebSockets to a bridge whose backing network is the `InMemoryHub` — exactly the shape of a production bridge, with the hub playing the private network. Alice's event reaches Bob (verified on receipt) and vice versa.
- 4 adapter tests in `buzz-acp`.
- `just ci` green.

## Scope / follow-ups

This PR proves the seam: the trait, both implementations, the protocol doc, and the env-driven factory. Converting the ACP harness main event loop to run on `Box<dyn Transport>` (it still uses the concrete `HarnessRelay` today) and refactoring `buzz-cli` onto the trait are deliberate follow-ups to keep this reviewable.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -59,6 +59,7 @@ crates/
   # Tooling + shared
   buzz-cli            # Agent-first CLI
   buzz-sdk            # Typed Nostr event builders
+  buzz-transport      # Pluggable Transport seam — signed-event streams (relay, remote bridge, in-memory)
   buzz-admin          # Operator CLI for relay administration
   buzz-ws-client      # Shared NIP-42 WebSocket client (connect, auth, publish)
   buzz-test-client    # Integration test client and E2E test suite
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -87,6 +87,7 @@ buzz-core    (zero I/O — types, verification, filter matching, kind registry)
          └── buzz-relay       (ties everything together — the server)
 
 buzz-acp            (agent harness — bridges relay @mentions → AI agents via ACP/JSON-RPC)
+buzz-transport      (pluggable Transport seam — signed-event streams; used by buzz-acp)
 buzz-sdk            (typed Nostr event builders — used by buzz-acp and buzz-cli)
 buzz-media          (Blossom/S3 media storage)
 buzz-cli            (agent-first CLI)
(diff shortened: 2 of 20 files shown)


bluepinkneon:
Closing per direction — not moving forward with the pluggable Transport trait for buzz-acp right now. Decision, not a defect in the PR.

PR #2763, opened 2026-07-24
fix(acp): answer unmentioned direct messages

### LINKED ISSUE #2747: buzz-acp: agents ignore DMs unless explicitly @mentioned — mention gate has no DM exemption
## Symptom

In a 1:1 DM channel with a managed agent, the agent does not see or answer plain messages — every single message must explicitly @mention it, even though the DM has exactly two participants. Replying to the agent inside a DM thread behaves the same way: no @, no answer.

Expected: a DM is addressed to the counterparty by definition. Chat platforms don't require @-mentioning your DM partner per message, and the mention requirement is invisible — the agent just looks unresponsive.

## Root cause

The mention gate is a single global flag with no channel-type awareness:

- `crates/buzz-acp/src/config.rs` (`subscription_rules`): `let require_mention = !config.no_mention_filter;` is applied uniformly to **every** target channel — DM channels included. `SubscribeMode::Mentions` is the default the desktop uses for managed agents.
- `crates/buzz-acp/src/filter.rs` (`match_event`): re-checks the `p` tag against the agent pubkey; no mention → dropped.

Notably the harness *does* know the channel is a DM further down the pipeline — `queue.rs` branches on `channel_type == "dm"` for reply anchoring and prompt hints — the DM fact just doesn't participate in the *trigger* decision.

Safety-wise, auto-responding in DMs appears already bounded: DM turns are restricted to the owner and verified siblings (#2591), so exempting DMs from the mention gate would not open agents to arbitrary third-party prompting.

## Propose

Exempt `channel_type == "dm"` channels from `require_mention` (subscription filter and `match_event`), so agents treat any owner message in a 1:1 DM as addressed to them. Group channels keep the existing mention gate unchanged.

Related: #2270 tracks the sibling problem in group-channel threads (agents go deaf in threads they've joined unless re-mentioned). This issue is scoped to DMs only, where the addressed-by-definition argument is strongest.

Observed on a self-hosted relay (deploy/compose derivative, `ghcr.io/block/buzz:main` 2026-07-24, desktop 0.4.24) with desktop-managed agents at default settings.

arisgysel-design:
## Summary
- remove the relay-level `#p` mention filter for confirmed DM channels at startup and on dynamic joins
- bypass the ACP rule-level mention gate only for confirmed DMs
- preserve mention requirements for streams, unresolved channel types, and setup mode

## Safety
DM author authorization remains unchanged: only the owner or verified siblings can trigger the agent. Unknown channel types continue to require a mention for rule matching.

## Tests
- `cargo test -p buzz-acp`
- `just ci`

Closes #2747

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -1313,6 +1313,17 @@ pub fn resolve_dynamic_channel_filter(
     }
 }
 
+/// Allow plain messages to reach the ACP harness in confirmed DM channels.
+///
+/// Mention filtering remains unchanged for streams and channels whose type
+/// could not be resolved. The harness performs the same confirmed-DM check
+/// before matching subscription rules.
+pub(crate) fn exempt_confirmed_dm_from_mentions(filter: &mut ChannelFilter, is_confirmed_dm: bool) {
+    if is_confirmed_dm {
+        filter.require_mention = false;
+    }
+}
+
 fn rule_applies_to_channel(rule: &SubscriptionRule, channel_id: Uuid) -> bool {
     use crate::filter::ChannelScope;
     match &rule.channels {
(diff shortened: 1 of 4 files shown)


svndco:
## Independent review — blocking behavior concern

I checked this out at `022b0bc2ac8e6397fd85848dffe5890c5c129a3f` and verified:

- `cargo test -p buzz-acp`: **595 library + 9 integration tests passed**

The relay-side subscription widening is necessary, but the in-process exemption is currently keyed only to `is_confirmed_dm`. The DM author gate intentionally accepts both the owner and verified same-owner siblings, so an unmentioned sibling-agent event in a DM will now match `require_mention` rules. That can reintroduce agent-to-agent wake loops; #2777 correctly narrows the exemption to `is_dm && author == owner`.

I consider that scope issue blocking. Please compute the mention exemption from both confirmed-DM status **and owner authorship**, while continuing to let the broader subscription deliver events for in-process filtering. Add a regression test showing an unmentioned sibling-agent DM does not match.

arisgysel-design:
Addressed the blocking review in `060c72e6`: the relay subscription still delivers all confirmed-DM events, but the in-process mention exemption is now scoped to `confirmed DM && author == configured owner`. Added regressions proving an unmentioned owner DM matches while an unmentioned sibling-agent DM does not. Full `just ci` passes.

PR #2767, opened 2026-07-24
fix(discovery): inject PATH into Codex adapter planning

cameronhotchkies:
## Why
The Codex adapter install-plan test depended on the process-global login-shell PATH cache, making it flaky under concurrent or loaded CI runs.

## What
- Thread an explicit probe PATH through adapter install planning
- Use a controlled PATH in Codex install-plan tests
- Update the existing desktop file-size allowance for the focused seam

## Risk Assessment
Low — production behavior keeps using the same augmented PATH; only dependency injection and deterministic tests change.

## References
- Failure: https://github.com/block/buzz/actions/runs/30110680608/job/89539202266
- Validation: `just desktop-tauri-test`; `cd desktop && pnpm check`; full pre-push hooks

Generated with Codex

### DIFF
--- a/desktop/scripts/check-file-sizes.mjs
+++ b/desktop/scripts/check-file-sizes.mjs
@@ -500,7 +500,9 @@ const overrides = new Map([
   // Includes unit tests for detection, routing, and -Command body preservation.
   // +16: test_powershell_command_goose_catalog_dequoted proves the \$→$ escape
   // fix for the Goose Windows installer (PR #2680 interaction with #2750).
-  ["src-tauri/src/commands/agent_discovery.rs", 1826],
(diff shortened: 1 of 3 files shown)


wesbillman (approved):
Approved. The failure this fixes is test isolation, not a product race: the valid 1.x fixture inherited process-global augmented/login-shell PATH state, so concurrent tests could make its child probe fail and conservatively classify it as outdated. This change injects the probe PATH at the planning seam; production still computes and passes the same augmented PATH, while tests pass a controlled PATH. I found no missed call sites or behavior change.

Reviewed at dabc2267008a7fb7656d98f8b6769964a3808477. CI is green across Desktop Core, Windows, macOS, smoke, and integration.

Issue #2725, opened 2026-07-24
[Bug] The "Delete agent" button is invisible when matching to system light theme


carimura:
**Describe the bug**
I just installed Buzz and tried to delete an agent but you can't see the button. I switched my system to dark theme and then the button became visible. See the attached screenshot. 

**Supporting Material**
see attached screenshot

<img width="311" height="318" alt="Image" src="https://github.com/user-attachments/assets/80cdf5d4-7568-4ee3-894c-281a43ae4b42" />

**Environment (please complete the following information):**
 - OS: MacOS 26.6 Beta
 - Version 

**Additional context**
I had Fable create a fix and I built from source and everything worked, but didn't want to contribute AI code as I figured y'all could provide a cleaner fix.

https://github.com/carimura/buzz/pull/1

jatinder14:
Taking this. Looking for the Delete agent control styles that disappear under the light theme and will open a PR with a contrast-safe fix.

jatinder14:
Stepping back — #2769 and #2772 already cover the contrast fix. Focusing elsewhere.

PR #2769, opened 2026-07-24
fix(desktop): contrast-guard the destructive accent from theme git colors

ibrahemid:
### Problem

With a light system theme active, the "Delete agent" menu item renders white on white (#2725). The theme engine feeds `--destructive` verbatim from the syntax theme's git-decoration colors: `theme-loader.ts` reads `gitDecoration.deletedResourceForeground` (falling back to diff-gutter background tints), and themes make no readability promise for these values. slack-ochin ships `#FFF`, so destructive text sits at 1.0:1 on the popover surface.

### Fix

`createThemeVars` now resolves the destructive accent through a contrast guard before emitting `--destructive`:

- Composes the surfaces destructive text actually renders on: the base background, the popover surface (`POPOVER_SURFACE_CLASS`: background 80% / muted 20%), and a focused menu row (`focus:bg-muted/50` over the popover).
- Keeps the theme's own accent while it clears a 3:1 floor (WCAG large-text minimum) on the worst of those surfaces, so curated palettes that merely miss AA are not retinted.
- Replaces a genuinely unreadable accent with the existing curated fallback red, pushed toward the readable pole by bounded binary search when the fallback itself misses 4.5:1 on unusual surfaces. Mirrors the `getReviewAccentForeground` pattern already in `ThemeProvider.tsx`.

13 of the 62 bundled themes change, each previously between 1.0:1 and 2.9:1 on its own surfaces: slack-ochin (1.0), min-dark (1.5), monokai (1.6), one-dark-pro (1.8), min-light (2.1), gruvbox-dark-soft (2.2), gruvbox-dark-medium (2.4), tokyo-night (2.5), gruvbox-dark-hard (2.7), snazzy-light (2.7), nord (2.7), everforest-light (2.8), kanagawa-wave (2.9). The other 49 emit byte-identical vars, including buzz, github-*, catppuccin-*, and dracula. The floor is a single constant (`DESTRUCTIVE_KEEP_FLOOR`); raising it to 4.5 would retint 37 themes including the first-party palette, so 3:1 was chosen deliberately. One-line change if a stricter bar is preferred.

The issue links a patch on the reporter's fork that nudges the hue of both status colors against the popover var. This PR instead substitutes the curated fallback and scopes to `--destructive`, so compliant themes keep their exact colors.

Deliberately out of scope: `--status-added` / `--status-deleted` share the root cause (slack-ochin's activity-row deletion counts and DiffViewer markers still receive `#FFF`), but changing them shifts diff colors across themes. Follow-up material.

### Tests

`desktop/src/shared/theme/adaptive-theme.test.mjs` (node:test, runs in `pnpm test` / `just ci`): contrast-ratio poles, all three resolver branches, an exact slack-ochin regression fixture, and a matrix over all 62 bundled themes asserting kept accents stay byte-identical and replaced accents reach 4.5:1 on the recomposed surfaces. With the guard reverted, the slack-ochin fixture and 13 matrix themes fail.

Manual: Settings, pick slack-ochin (or let a light system theme resolve to it), open an agent profile menu. "Delete agent" is readable red instead of invisible.

### Checks

- [x] `just ci` passes (fmt + clippy + unit tests + mobile)
- [x] Integration tests pass (`just test`)
- [x] New public APIs / tools / endpoints are documented
- [x] No new `unsafe` blocks

### DIFF
--- a/desktop/src/shared/theme/adaptive-theme.test.mjs
+++ b/desktop/src/shared/theme/adaptive-theme.test.mjs
@@ -1,0 +1,224 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import {
+  contrastRatio,
+  createThemeVars,
+  hexToHsl,
+  resolveDestructiveAccent,
+} from "./adaptive-theme.ts";
+import {
+  SYNTAX_THEMES,
+  extractThemeInfo,
+  loadThemeData,
+} from "./theme-loader.ts";
+
+// The engine renders destructive text on the plain background, on the popover
+// surface (background 80% / muted 20%) and on a focused dropdown row (popover
+// 50% / muted 50%). Reconstruct those from the emitted vars to check the
+// shipped values rather than a reimplementation of the derivation.
+const MIN_SHIPPED_CONTRAST = 4.4;
+
+function toHex({ r, g, b }) {
+  const clamp = (n) => Math.max(0, Math.min(255, Math.round(n)));
+  return `#${[r, g, b].map((c) => clamp(c).toString(16).padStart(2, "0")).join("")}`;
+}
+
+function hexToRgb(hex) {
+  const long = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+  if (long) {
+    return {
+      r: parseInt(long[1], 16),
+      g: parseInt(long[2], 16),
+      b: parseInt(long[3], 16),
+    };
+  }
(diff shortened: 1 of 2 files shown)


ibrahemid:
Before/after on the issue surface: agent instance settings menu, slack-ochin (light).

### Before

"Delete agent" is present but renders white on the popover, 1.03:1.

01-before-slack-ochin

### After

Same row with the guard, 5.22:1.

02-after-slack-ochin

PR #2771, opened 2026-07-24
fix(desktop): enable webview media capture on Linux

kwsantiago:
Huddle mic capture and avatar recording don't work on Linux. `navigator.mediaDevices` is undefined, so every `getUserMedia` call site fails.

WebKitGTK gates those APIs behind `enable-media-stream`, which is off by default, and denies permission requests that no handler claims. Neither is configured today. macOS and Windows are unaffected, since their system webviews expose the APIs and defer to the OS prompt.

This enables the setting on the main webview and grants audio and video device requests. The frontend never calls `getDisplayMedia`, so screen capture keeps WebKit's default deny.

`webkit2gtk` is pinned to `=2.0.2` to match wry's pin, because `PlatformWebview::inner()` returns wry's type. It was already in the lockfile, so that file changes by one line.

Tested on Ubuntu 24.04 with WebKitGTK 2.52.3. Before: `navigator.mediaDevices` is undefined. After: `getUserMedia({audio: true})` resolves to an active `MediaStream` and `pactl list source-outputs` shows live `WebKitWebProcess` record streams on the capture device. `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test --lib` pass.

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -1082,6 +1082,7 @@ dependencies = [
  "url",
  "user-idle",
  "uuid",
+ "webkit2gtk",
  "window-vibrancy",
  "windows-sys 0.61.2",
  "zeroize",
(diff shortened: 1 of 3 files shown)


wpfleger96:
🤖 hey @kwsantiago, thanks for this — your `is_for_audio_device()` / `is_for_video_device()` check is part of what actually shipped: #3607 (just merged, fixes #3495) adopted that exact device-kind classification in its deny-by-default handler, alongside an origin allowlist. Also appreciate you running the manual Ubuntu 24.04 mic verification over on #3607. Closing this as superseded by the merged fix.

PR #2573, opened 2026-07-23
feat(managed-agents): add Oh My Pi as a native ACP runtime

dkta0:
## Summary
- add Oh My Pi to the ACP runtime catalog with native `omp acp` invocation
- support official macOS/Linux and Windows installers, including the Windows `%LOCALAPPDATA%\omp` path
- gate onboarding on OMP setup completion and expose OMP in the harness picker
- keep OMP first in onboarding so the newly supported native-ACP option is immediately discoverable during rollout; Claude Code and Codex remain available alongside it

### Related issue

None found.

### Testing

- `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml -- --check`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml managed_agents::discovery::tests:: --lib` (78 passed)
- `node --test --experimental-strip-types desktop/src/features/onboarding/ui/onboardingRuntimeSelection.test.mjs` (4 passed)

PR #2773, opened 2026-07-24
feat(acp): bring your own harness (BYOH) — generic ACP runtime seam + settings gallery

wpfleger96:
## What

Implements a "bring your own harness" (BYOH) generic ACP mechanism — replacing per-harness backend code with a data-driven 3-tier system:

- **Tier 1 (compiled-in builtins):** goose, claude, codex, buzz-agent — unchanged behavior
- **Tier 2 (bundled presets):** cursor, omp, grok, opencode, kimi, amp, hermes, openclaw, and any future additions — defined in `PRESET_HARNESSES`, no code duplication, icons stay TerminalSquare/bundled-asset-only
- **Tier 3 (user-defined custom):** JSON definitions saved to `custom_harnesses/` under app data; managed via Settings → Agents UI

### Core data model
- `HarnessDefinition` — id, label, command, args, env, install URL/hint
- `PRESET_HARNESSES` static table — single source of truth for all presets; `preset_harness_ids()` derives reserved IDs (D-11: no hand-maintained copy)
- `source: "builtin" | "preset" | "custom"` tagging on every catalog entry

### Persistence (B-4, B-6)
- `save_custom_harness_to_dir(dir, definition, rename_old_id)` — backup-swap atomic write (backs up target → .bak, commits temp → target, restores .bak on failure, removes .bak on success); safe on Windows where `fs::rename` over an existing file is "access denied"
- `save_and_warm` / `delete_and_warm` — hold `PERSIST_MUTEX` for the write + registry-warm pair, eliminating the lost-update race (B-6) where two concurrent saves could interleave their warm calls and leave a stale registry snapshot
- Validate-before-mutate: both IDs and env validated before any filesystem mutation

### Env validation boundary (B-3)
- `validate_harness_definition_pub` calls `validate_user_env_keys` on definition env at save AND load
- Rejects malformed keys (BUZZ_AUTH_TAG=x forgery shape), reserved keys (BUZZ_MANAGED_AGENT etc.), NUL bytes, oversized values

### TypeScript boundary (B-2 / Thufir CRITICAL)
- `RawAcpRuntimeCatalogEntry` now declares `definition_env?: Record<string,string>` and `source: "builtin" | "preset" | "custom"`
- `fromRawAcpRuntimeCatalogEntry` maps `definition_env → definitionEnv` (camelCase); absent field defaults to `{}`
- Edit form reads `entry.definitionEnv` — env no longer erased on save-then-edit cycle

### Unified descriptor (Phase A / Thufir F4)
- `EffectiveHarnessDescriptor { command, args, env }` in `readiness.rs`
- `resolve_effective_harness_descriptor()` — single resolver used by spawn, spawn_hash, summary, get_agent_models (both saved and unsaved), and readiness
- No competing arg-resolution forms

### Other fixes
- B-5: stop freezing `runtime.defaultArgs` into `record.agent_args` on normal create paths
- B-7: readiness exec-check — `MissingBinary` variant for custom commands not found on PATH
- B-8: onboarding transition — `setTimeout(0)` removed, parent-owned route intent via `navigateAfterComplete` prop
- C-9: collector-discriminating sweep tests with injectable filters
- C-10: `HarnessManagementCard` uses `harnessGalleryLogic` helpers (killed duplicate filter/sort)
- D-11: `BUILTIN_IDS` derived from `PRESET_HARNESSES` (no hand-maintained copy)
- D-12: `mobile/pubspec.lock` churn reverted
- D-13: false ownership fast-path comment fixed
- D-14: URL scheme validation for `installInstructionsUrl`
- D-15: OpenClaw Gateway env-locus README line

### Tests added
**B-4 persistence (6 tests):** `save_to_dir_create_writes_file_and_loads_back`, `save_to_dir_same_id_edit_replaces_content`, `save_to_dir_backup_is_cleaned_up_after_same_id_edit`, `save_to_dir_rename_removes_old_file_and_creates_new`, `save_to_dir_rename_nonexistent_old_id_is_non_fatal`, `save_to_dir_roundtrip_with_env_preserves_values`

**B-3 env validation (6 tests):** `validate_rejects_malformed_key_with_equals_sign`, `validate_rejects_reserved_key_buzz_managed_agent`, `validate_rejects_reserved_key_case_insensitive`, `validate_rejects_nul_byte_in_value`, `validate_rejects_value_over_per_value_size_limit`, `validate_accepts_well_formed_env`

**B-2 API boundary (4 TS tests in tauri.test.mjs):** `fromRawAcpRuntimeCatalogEntry maps definition_env to definitionEnv`, `defaults definitionEnv to {} when absent`, `preserves source preset`, `env round-trips through edit payload shape`

## Preset catalog

| ID | Label | Command |
|----|-------|---------|
| `cursor` | Cursor | `cursor-agent acp` |
| `omp` | Oh My Pi | `omp acp` |
| `grok` | Grok Build | `grok agent --always-approve stdio` |
| `opencode` | OpenCode | `opencode acp` |
| `kimi` | Kimi Code | `kimi acp` |
| `amp` | Amp | `amp-acp` |
| `hermes` | Hermes Agent | `hermes-acp` |
| `openclaw` | OpenClaw | `openclaw acp` |

## Review-fix pass (2026-07-26, Eva)

Fixes from the three-way review (Wren / Dawn / Eva) in the buzz-generic-acp-harnesses thread, pushed as new commits (no rewrite):

1. **installHint edit round-trip** — form seeding extracted to `formValuesFromCatalogEntry` (single source of truth), input rendered, full-definition lossless round-trip regression.
2. **Dangling-delete coherence** — delete allowed; confirm counts referencing agents (direct pin + persona-inherited); summary rows render `harness (deleted): <id>`; spawn errors become actionable sentences (`user_facing_harness_error`); composed delete→summary→start test.
3. **Comma-in-args** — rejected at `validate_harness_definition` (shared by save AND disk load), mirrored inline in the form.
4. **Registry publish race** — collision/dup filtering moved into `load_custom_harnesses` (both loaders inherit shadowing rules); discovery publishes by re-reading the dir under `persist_mutex` (lock scoped to publish only); deterministic interleaving regressions for save-during-discovery and delete-during-discovery.
5. **Mechanical** — discarded `belongs_to_us` sweep arg deleted, `load_global_agent_config` hoisted out of the per-record summary loop, duplicated doc paragraph + stray SAFETY comment removed.
6. **PGID test de-flaked** — leader kept alive through the assertion.

Known follow-up (filed in review, not blocking): file-size split-outs queued in `check-file-sizes.mjs` entries.

## Gate table — head `bf53f1d60`

| Gate | Result |
|------|--------|
| `cargo test --lib` (desktop/src-tauri) | **1701 passed**, 0 failed, 14 ignored |
| desktop JS suite (`pnpm test`) | **3605 passed**, 0 failed |
| `tsc --noEmit` | clean |
| `biome check` + file-size/px/pubkey checks | clean |
| `cargo clippy --lib -- -D warnings` | clean |
| `cargo fmt --check` | clean |

PR head: `bf53f1d60e3cbd07392e1287b83bb37ba90d0d33` — includes merge of origin/main (`c2a4ee711`, conflicts in agent_models composed with #2890's live Databricks discovery)


OpenSpek:
Windows end-to-end proof completed against PR #2773 frozen at `0922869abe9f170879276588f3c4c51c292e230f`.

Proof shape:

- built and launched the disposable Windows desktop directly
- launched installed `hermes-acp.exe` 0.19.0 directly (no wrapper)
- one managed worker
- `openai-codex:gpt-5.6-sol`
- `HERMES_ACP_SKIP_CONFIGURED_MCP=1`
- real channel mention routed through ACP
- Hermes executed the bundled `buzz` CLI reply operation
- Buzz displayed a thread reply attributed to the managed Hermes agent with the exact requested token

Two Windows Hermes terminal issues were exposed by the proof:

1. the Git Bash startup probe could wedge after a timeout while an MSYS descendant retained captured pipe handles — already addressed by NousResearch/hermes-agent#69083;
2. `bash -l` replaced the native host-injected `PATH`, hiding Buzz's bundled CLI from the terminal snapshot — addressed by NousResearch/hermes-agent#71433.

One Buzz-side safety finding: a blank/new managed-agent parallelism value inherited `24` on this machine and launched 24 workers. Explicitly setting `1` fixed containment. Please consider defaulting new managed ACP agents to `1`, or requiring an explicit value/warning before launch.

This validates the generic ACP harness path on Windows; it does not imply a production routing cutover.

tlongwell-block:
2bae233534222b01139e44cb5cc9bc2ce2e778ee634fae11a72659916e97bd4c

bradhallett (commented):
Dropping these as design questions rather than blockers — I know this is mid-flight. Strong execution on the 3-tier model, the bundled-logo drift guard (`presetLogos.test.mjs` catching Rust↔TS drift), and @OpenSpek's real Windows E2E proof with hermes-acp.

**One thing worth a design call: the preset tier can't express a setup gate, and omp is the case that bites.** `PresetHarness` carries only `id / label / command / args / install_instructions_url / install_hint`, and every preset is pinned to `AuthStatus::NotApplicable` — fine for harnesses that work the moment the binary resolves on PATH (cursor, grok, opencode, …), but omp needs `omp setup` (model + auth) post-install. Without a probe, omp shows "ready" the instant the binary is found and the failure surfaces only at ACP `initialize`/`authenticate`. The signal is clean — I verified `omp config get setupVersion` → `1`, exit 0 (omp 17.1.3); `0`/absent = not-set-up — and #2573 wires exactly this as a positive-integer probe on the legacy path. Is setup-gating intentionally out of scope for presets, or worth an optional `auth_probe` field? Relevant to the #2573 overlap either way.

Two smaller notes:
- omp `install_instructions_url` points at `github.com/can1357/oh-my-pi` (source) — `https://omp.sh/docs/quickstart` is the actual install guide. Same one-liner #2573 needs; worth matching whichever lands.
- Seconding @OpenSpek's parallelism-24 safety ask — this PR owns the new-harness creation surface, so it's the natural place to default new managed ACP agents to 1 or require an explicit value/warning before launch.

tlongwell-block:
Hi! Parallelism is set to 24 because it defines how many channels a given bot/identity can work in at once. In some harnesses that's a single agent process with parallel sessions. In other, like Buzz Agent, that's a process per channel. 24 is a good default.

tlongwell-block:
I use my bots in a lot of channels at once.

joelbrilliant:
The BYOH seam is the right shape. Speaking as one of the people who wrote a bespoke Hermes runtime, I'd rather this landed than mine.

Worth flagging the collision in case it helps prioritising. There are four open PRs adding Hermes to Buzz right now (this one, #2633 mine, #2656, #2468) and all four touch `managed_agents/discovery.rs` and its tests. Any one of them merging breaks the other three, which is probably why none have moved.

This PR makes most of that moot. Discovery and registration become generic, and I'm happy to close the discovery half of #2633 as redundant once this is in.

Two things I can contribute that don't overlap with it:

1. A Hermes harness definition for the gallery. You've already got `hermes.png` in here, so it's mostly just the config shape.
2. Durable session load. #2633 is the only one of the four with a `session_store.rs`, so a Hermes session survives a Buzz restart instead of starting cold. Around 650 lines, sits on top of the seam rather than inside it, tests green against current main.

Happy to cut both as small follow-ups once this merges, or push them into this branch if you'd rather have it in one go. Whichever is less work for you.

tlongwell-block:
> * Seconding @OpenSpek's parallelism-24 safety ask — this PR owns the new-harness creation surface, so it's the natural place to default new managed ACP agents to 1 or require an explicit value/warning before launch.

After seeing how heavy some agents are, I've lowered this to 10 by default in #3038

Will follow up on real efficiency gains in #3044

Bartok9:
Quick coordination note while #2656 / the Hermes-specific design work is folded into this BYOH seam:

Happy to defer entirely to #2773 as the land path — generic tier-2 Hermes is the right shape.

**Credit ask (person-level):** when Hermes runtime bits that came from the #2656 line (native Hermes ACP registration shape, config-bridge / readiness / model-selection thoughts, cold-start + `HERMES_ACP_SKIP_CONFIGURED_MCP`, supervisor/`killpg` hardening discussion with @joelbrilliant, etc.) land in the merge to `main`, please keep a person-level nod — `@Bartok9` and/or `Co-authored-by: Bartok9` — not only a PR-number reference to #2656. Same spirit if related follow-ups cherry-pick those ideas.

Glad to keep reviewing this PR for || design nits, tests, and the killpg/signal-forwarding regression class. Thanks for carrying Hermes + OpenClaw in the gallery.

tlongwell-block (approved):

amanning3390:
Thanks for landing the generic BYOH seam. I re-checked the merged head against the Hermes-specific contracts in #2656.

The generic path already covers the important architecture: Hermes is a named tier-2 preset, `hermes-acp` is a supported native Hermes console entrypoint (not a separate adapter), custom harness definitions can carry env, model discovery/selection is generic, and the direct entrypoint avoids #2656's `/bin/sh` supervisor path.

Two focused Hermes preset gaps remain:

1. Set `HERMES_ACP_SKIP_CONFIGURED_MCP=1` in the Hermes preset env. Hermes Agent now documents and implements this host marker so Buzz-owned ACP sessions do not also start the user's globally configured MCP servers.
2. Make the model-probe timeout data-driven and give Hermes 45s. The merged `buzz-acp` path still uses the global 10s `MODELS_TIMEOUT`; Hermes cold starts can legitimately exceed it.

The first should be preset data using the existing env field. The second needs a per-definition timeout field rather than another Hermes-specific match arm. I can send that as a focused follow-up. Please also preserve person-level credit for the Hermes-specific design/testing carried over from #2656, as discussed above.

Bartok9:
Both follow-ups are the right shape and both are preset-data, not new match arms — which is exactly the point of the generic seam.

**1. `HERMES_ACP_SKIP_CONFIGURED_MCP=1` in the Hermes preset env.** This is pure data using the existing env field on the definition — no code path change. It stops Buzz-owned ACP sessions from also spinning up the operator's globally-configured MCP servers, which is the whole reason Hermes documents the marker host-side. Keep the same operator-precedence rule the merged spawn path already uses (`std::env::var(&key).is_err()` before setting) so an operator who exports it wins.

**2. Data-driven probe timeout, 45s for Hermes.** Agreed the merged `buzz-acp` path still uses the global 10s `MODELS_TIMEOUT` at the initialize/probe sites, and Hermes cold starts legitimately exceed that (observed ~13–21s live). The clean fix is a per-definition `model_probe_timeout` field resolved at all probe sites, defaulting to `MODELS_TIMEOUT` for everything else — not a `hermes` match arm. Worth asserting the contract in a test (`hermes > default`, others `== MODELS_TIMEOUT`) rather than freezing `45`, so tuning the budget later doesn't turn it into a change-detector.

Both are small and independent; happy to review them or take one as a focused patch on the merged head, whichever is less work for you. Thanks for carrying the Hermes preset properly on top of the generic machinery.

joelbrilliant:
Small correction to the contract mapping, since it could lead someone to close the wrong PR.

Persistence isn't covered by the merged seam. There's no session store on main, and more to the point `buzz-acp` doesn't call `session/load` at all. Grep `session/load`, `load_session` or `loadSession` under `crates/buzz-acp/src/` on the merged head and it comes back empty. So a harness restart still starts a cold conversation even against an agent that advertises `loadSession`, which Hermes does.

That's the gap #2633 fills, and I've just rescoped it down to exactly that on top of the merged head: 6 files, +643/-7, with everything overlapping discovery or registration dropped. 604 tests green, clippy and fmt clean at CI's settings.

Easy thing to assume given how much #2773 does cover, and the rest of that mapping looks right to me.

Bartok9:
You're right, and thanks for the precise correction — I don't want a bad mapping to get #2633 closed for the wrong reason.

I re-grepped the merged head: `session/load`, `load_session`, and `loadSession` all come back empty under `crates/buzz-acp/src/`, so `buzz-acp` genuinely never calls `session/load`. The generic seam here covers discovery/registration/probe, but not persistence — a harness restart is still a cold conversation even against an agent that advertises `loadSession` (Hermes does). That's a real, distinct gap, not something #2773 subsumes.

Your rescoped #2633 (6 files, +643/-7, everything overlapping discovery/registration dropped, 604 green + clean clippy/fmt) is exactly the right shape for that seam. Happy to review it on the merged head. Appreciate you catching this before the mapping did any damage.

Issue #2748, opened 2026-07-24
buzz-acp: DM replies anchor to the latest message instead of the thread root — every exchange nests one level deeper


webdevtodayjason:
## Symptom

In a 1:1 DM with a managed agent, once a thread exists, every agent reply anchors to the human's **latest comment** rather than the thread root. Each exchange therefore nests one level deeper — the user has to click-expand a sub-thread of a sub-thread of a sub-thread to follow a simple back-and-forth conversation with their own agent.

## Root cause

`crates/buzz-acp/src/queue.rs` — the human-aware reply-anchor logic documents the right intent directly above the code:

> Human-facing turns are anchored so replies stay readable at layer 1:
>   - in a thread → anchor to the thread ROOT (no depth-2 nesting)

…and `resolve_reply_anchor` implements exactly that for group channels. But the DM branch contradicts it:

```rust
let reply_anchor = if is_dm {
    thread_tags
        .root_event_id
        .is_some()
        .then(|| last_event.event.id.to_hex())   // ← anchors to the LATEST message, not the root
} else {
    resolve_reply_anchor(...)                    // ← group channels: anchors to ROOT
};
```

When the triggering DM message is inside a thread, the anchor is `last_event.event.id` — the user's newest comment — so the agent's reply becomes a child of a child, and the nesting compounds every round.

## Propose

In DMs, match the documented layer-1 rule: anchor to `thread_tags.root_event_id` when inside a thread (same as the group-channel human path). Worth discussing whether top-level DM exchanges should thread at all — a 1:1 DM arguably reads best flat, with threading reserved for explicit user-initiated threads — but the root-anchoring fix alone eliminates the compounding sub-chat nesting.

Observed on a self-hosted relay (`ghcr.io/block/buzz:main` 2026-07-24, desktop 0.4.24), desktop-managed agents, default settings. Sibling issues: #2270 (thread deafness without re-mention), and the DM mention-gate issue filed alongside this one.

muqsitnawaz:
Independent reproduction on macOS (Darwin 25.4.0), desktop-managed agent, self-hosted relay,
`main` at 63c62fcf3 — and I arrived at the same line before finding this issue.

**What the user sees:** in a 1:1 DM, a reply typed into the thread composer gets answered *one
level deeper* — the human's in-thread message grows its own `1 reply` sub-thread, and every round
adds another level. Reading a simple back-and-forth means expanding a sub-thread of a sub-thread.

[...]

Issue #2270, opened 2026-07-21
[Bug] buzz-acp agents go deaf in threads they've joined unless re-mentioned.


Ansonhkg:
> Issue generated by Kimi K3

**Describe the bug**

The buzz-acp agent harness only responds to messages that explicitly @mention the agent. Once an agent has joined a thread, follow-up replies in that thread are silently ignored unless the human re-tags the agent on every single message.

The gate exists at two levels:
1. Relay subscription — `SubscribeMode::Mentions` (the default) sets `require_mention: true` on channel filters, so the relay never even delivers untagged thread replies to the harness.
2. Rule matching — `match_event` in the filter module re-checks for the agent's `p` tag; no match → event dropped.

**To Reproduce:**

1. Start a buzz-acp agent with default config (Mentions subscribe mode) joined to a channel.
2. In that channel, send a message @mentioning the agent — it responds in a thread.
3. Reply in that same thread without tagging the agent (e.g. "yes, go ahead").
4. The agent never sees the reply — no reaction, no response. Re-tag the agent in the same thread and it responds immediately.

**Expected behavior:**

The agent should follow threads it is participating in — like Slack/Discord thread notifications — so a human can continue the conversation naturally without re-mentioning on every reply. Agents must never auto-reply to each other without an explicit mention (anti-loop invariant).

**Agreed solution direction (participation-based thread following):**

- Track per-channel "participating thread roots" in the harness — seeded when the agent is mentioned in or replies to a thread (NIP-10 `root_event_id` via the existing thread-tag parser), with TTL/LRU expiry so stale threads don't wake the agent forever.
- Widen the relay subscription in Mentions mode (drop relay-side `require_mention` for chat message kinds) and move gating fully into rule matching: queue if `mentioned(agent)` OR `root_event_id ∈ participating[channel]`.
- Anti-loop invariant: participation-follow applies to human-authored replies only; agent-authored replies still require an explicit mention. Never queue the agent's own events. The existing author gate, per-channel in-flight dedup, and fail-closed filter semantics stay unchanged.
- Config: new per-rule fields `follow_participating_threads` (default true in Mentions mode) and `follow_thread_authors = "humans" | "all"`.
- Scope: `buzz-acp` only — no relay changes, no new event kinds, no protocol changes.

**Supporting Material**

[...]

PR #2777, opened 2026-07-24
fix(acp): answer owner DMs without an explicit @mention

### LINKED ISSUE #2747: buzz-acp: agents ignore DMs unless explicitly @mentioned — mention gate has no DM exemption

webdevtodayjason:
Closes #2747.

## Problem

In a 1:1 DM with a managed agent, the agent ignored every message unless it was explicitly `@`-mentioned — even though a DM has exactly two participants and is addressed to the counterparty by definition. The result: agents look unresponsive to their own owner in DMs.

The mention gate (`require_mention`) was applied uniformly to every channel, DM included, at two levels:
- **Subscription filter** (`config.rs`): the relay-side `#p` filter dropped untagged events before they were ever delivered to the harness.
- **`match_event`** (`filter.rs`): re-checked the agent's `p` tag; no match → dropped.

Dropping only the subscription filter is insufficient — `match_event` would still drop the delivered-but-untagged DM. Both levels have to be exempted together.

## Fix

Exempt the mention gate for **owner-authored DM events**:
- The caller computes `mention_exempt = is_dm && author == owner` and threads it into `match_event`, which skips `require_mention` when set.
- DM channels' subscription filter drops the `#p` requirement so the events are delivered in the first place.
- `setup_mode` opts out (keeps its own mention requirement).

**Why owner-scoped, not a blanket DM exemption:** exempting *all* DM events would let two agents in a DM trigger each other with no mention, re-creating the thread-deafness loop that #2270 is about. Scoping to the owner means the human's DM messages auto-trigger the agent while agent-to-agent DM still requires an explicit mention — preserving the anti-loop invariant. Group channels are unchanged.

**Safety:** DM turns are already restricted to the owner and verified siblings (#2591), so this does not open agents to third-party prompting.

## Tests

- `filter.rs`: an owner-authored DM event with no `p` tag now matches (agent responds); a group-channel event with no `p` tag still does NOT match (gate preserved); a non-owner DM event still requires a mention.
- `config.rs`: the DM subscription filter drops `#p`; non-DM channels keep it.
- Full gate green: `cargo test -p buzz-acp` → 611 passed, `cargo clippy -p buzz-acp --all-targets -- -D warnings` → 0 warnings.

Sibling PR: the DM reply-anchor fix (#2748) is up alongside this. Related open issue: #2270 (thread deafness in group channels), a distinct but adjacent case.

## Checklist
- [x] `cargo test -p buzz-acp` + clippy green
- [x] Bug fix includes regression tests
- [x] No new `unwrap()` in production paths
- [x] No new `unsafe`

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -482,6 +482,24 @@ pub struct ChannelFilter {
     pub require_mention: bool,
 }
 
+impl ChannelFilter {
+    /// Drop the relay-side `#p` mention filter when the channel is a DM.
+    ///
+    /// A 1:1 DM is addressed to the agent by definition, so the relay must
+    /// deliver every message in it — not just `p`-tagged ones — otherwise the
+    /// agent never sees plain (un-@mentioned) DM messages (issue #2747). This
+    /// only widens *delivery*; per-event triggering is still gated by the
+    /// inbound author gate and `match_event`, which restrict DM turns to the
+    /// owner. Group channels keep their mention filter unchanged.
+    #[must_use]
+    pub fn with_dm_exemption(mut self, is_dm: bool) -> Self {
+        if is_dm {
+            self.require_mention = false;
+        }
+        self
+    }
+}
+
 #[derive(Debug)]
 pub struct Config {
     pub keys: Keys,
@@ -1329,6 +1347,33 @@ mod tests {
     use super::*;
     use crate::filter::{ChannelScope, SubscriptionRule};
     use clap::{Parser, ValueEnum};
+
+    #[test]
+    fn test_with_dm_exemption_clears_mention_for_dm() {
+        let filter = ChannelFilter {
+            kinds: Some(vec![9]),
+            require_mention: true,
+        }
+        .with_dm_exemption(true);
+        assert!(
+            !filter.require_mention,
+            "DM must drop the relay-side #p mention filter"
+        );
+        assert_eq!(filter.kinds, Some(vec![9]), "kinds must be preserved");
+    }
+
+    #[test]
+    fn test_with_dm_exemption_preserves_mention_for_non_dm() {
+        let filter = ChannelFilter {
+            kinds: Some(vec![9]),
+            require_mention: true,
+        }
+        .with_dm_exemption(false);
+        assert!(
+            filter.require_mention,
+            "group channels keep the mention filter unchanged"
+        );
+    }
 
     /// Build a minimal Config for testing without CLI parsing.
     fn test_config(mode: SubscribeMode) -> Config {
--- a/crates/buzz-acp/src/filter.rs
+++ b/crates/buzz-acp/src/filter.rs
@@ -351,8 +351,18 @@ const MAX_CONSECUTIVE_TIMEOUTS: u32 = 5;
 /// 2. **kinds** — if non-empty, the event kind must be in the list.
 /// 3. **require_mention** — if `true`, a `p` tag matching `agent_pubkey_hex` must
 ///    exist. Tag kind is checked via `tag.as_slice()` for stable, library-independent
-///    access.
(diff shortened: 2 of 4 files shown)


svndco:
## Independent review — preferred ACP direction, one blocking edge case

I checked this out at `5250b55056bbee5f184e08cb91525eb3ccaebfcd` and verified:

- `cargo test -p buzz-acp`: **602 library + 9 integration tests passed**

This is safer than #2763 because the in-process mention exemption is owner-only, preserving the sibling-agent anti-loop boundary.

There is one blocking inconsistency in the dynamic-channel path: `is_dm_channel()` deliberately returns `true` when metadata resolution fails (fail-closed for the **author gate**), but lines 1987–1988 reuse that value to remove the relay `#p` filter, and line 2197 reuses it to exempt an owner event from `require_mention`. A transient metadata failure for a newly joined non-DM channel therefore makes that channel behave like a DM for mention matching, contradicting the stated “unknown types keep the mention filter” behavior used at startup.

Please split the two verdicts: keep `unknown => DM` for author authorization, but require positive `channel_type == "dm"` metadata for mention-filter bypass. Add a regression test proving unresolved channel type retains `require_mention`. With that adjustment, this is the ACP PR I would keep; #2763 can close as the duplicate.

webdevtodayjason:
Good catch, thanks for taking the time to actually check it out and run the gates. You're right about the dynamic path reusing the fail-closed value. Fixed in b4292e6e (new commit, didn't force push).

I split it into two checks. is_dm_channel() stays exactly as it was, fail-closed, and only gates author auth. Added is_dm_channel_confirmed() which needs channel_type == "dm" to actually resolve, and that's what gates the #p filter removal and the require_mention exemption now. Both come from one resolve_dm_verdict() call per event so they can't drift apart. The dynamic path now matches what startup was already doing with unknown types.

Added the regression test you asked for (unresolved type keeps require_mention and stays fail-closed on auth) plus coverage for the positive DM case and a resolved non-DM.

cargo test -p buzz-acp is at 614 passing, clippy -D warnings and fmt clean.

PR #2788, opened 2026-07-24
fix(desktop): recover from outdated Claude Code auth status probes

fletchertyler914:
## Summary
- Older Claude Code builds (e.g. 2.1.37) lack the `auth` subcommand, so Buzz’s `claude auth status` probe is treated as a free-form prompt, times out, and leaves onboarding **Next** disabled.
- Validate Claude auth probes against JSON `loggedIn` (instead of bare exit status), attach an actionable update diagnostic, and repair outdated CLIs via `claude update` during Install / **UPDATE CLI**.
- Onboarding shows **UPDATE CLI** / **Update required** only when discovery reports `unknown` **with a diagnostic** (not every transient unknown). Doctor surfaces that diagnostic text. Post-install catalog refresh stays pending until discovery catches up, with one diagnostic-gated retry.

## Test plan
- [x] `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib managed_agents::readiness::cli_probe`
- [x] Hard validation: downgrade to Claude Code 2.1.37 → UPDATE CLI → update → READY / Next enabled
- [x] Current Claude Code 2.1.219 → `loggedIn: true` via discovery
- [x] Pre-push hooks (desktop/rust/mobile tests) on push
- [ ] Maintainer: `just ci` / GitHub CI green + review


fletchertyler914:
## Why this change

Buzz onboarding probes Claude Code with `claude auth status`. On older builds (reproduced on **2.1.37**), that subcommand doesn’t exist — Claude treats the args as a free-form prompt, the probe times out, auth stays `unknown`, and **Next** stays disabled even though Claude is installed and the user is fine.

That’s a bad first-run experience: identity works, install looks successful, then the flow silently stalls on “status unavailable.”

## What improved

1. **Correct probe semantics** — Claude auth is judged from JSON `loggedIn`, not bare exit status. Prompt-hijack / missing JSON maps to an actionable “update required” diagnostic instead of a dead-end unknown.
2. **Self-heal on Install** — if Claude is present but too old for `auth status`, install runs `claude update` so users don’t have to leave the app and guess.
3. **Clearer onboarding UX** — when discovery attaches that diagnostic, the card shows **UPDATE CLI** / **Update required** (and **UPDATING** while repair runs). Generic unknown (no diagnostic) still gets **CHECK AGAIN**, so other harnesses aren’t mislabeled as needing an update.
4. **No post-update flicker** — the install mutation stays pending until the runtime catalog refetches (with one diagnostic-gated retry), so READY doesn’t flash back to UPDATE CLI on stale data.
5. **Doctor visibility** — the same diagnostic text is surfaced in Doctor when present.

Net: outdated Claude Code no longer bricks onboarding; the path is diagnose → update → READY, with a small, catalog-driven footprint rather than a harness-id special case in the UI.

PR #2804, opened 2026-07-25
fix(desktop): ship poof sound as mp3 so Linux can decode it

kwsantiago:
`desktop/public/pow/plop.m4a` is AAC. WebKitGTK decodes media through GStreamer, which has no AAC decoder in a default Linux install, so the poof sound never plays and startup logs:

```
Missing decoder: MPEG-4 AAC (audio/mpeg, mpegversion=(int)4, ...)
Error: 12, Your GStreamer installation is missing a plug-in.
```

Every other bundled sound (`desktop/public/sounds/*.mp3`) is already mp3, which decodes out of the box. This converts the one outlier.

Re-encoded from the original asset with:

```
ffmpeg -i plop.m4a -codec:a libmp3lame -b:a 128k plop.mp3
```

17264 -> 2826 bytes, mono 44.1 kHz preserved. The asset is MIT (Emerge Tools); `desktop/public/pow/LICENSE.txt` is unchanged.

## Verification

Ubuntu 24.04, WebKitGTK 2.52.3, without `gstreamer1.0-libav` installed:

- Before: the decoder error appears ~1.3s after startup.
- After: no GStreamer errors over a 14s run reaching the same startup point.

macOS and Windows are unaffected; their system webviews decode both formats.


Chessing234 (commented):
shipping plop as mp3 is the right linux decode fix — m4a was a footgun on gstreamer stacks (same family as #3488). please confirm the asset is re-encoded (not just renamed) and that mac/windows still play it; a one-line note in the pr about encoder settings helps reviewers who can't a/b the binary.

kwsantiago:
Confirmed re-encoded, not renamed. ffprobe on the committed blob:

```
codec_name=mp3
codec_long_name=MP3 (MPEG audio layer 3)
sample_rate=44100
channels=1
bit_rate=128000
duration=0.104490
size=2826
```

It begins with an ID3v2 tag rather than an MP4 `ftyp` box. The original was AAC, 17264 bytes, 0.116100s. The ~12ms duration difference is encoder delay and padding from the transcode.

Encoder settings are in the PR body:

```
ffmpeg -i plop.m4a -codec:a libmp3lame -b:a 128k plop.mp3
```

Mono and 44.1 kHz are carried over from the source.

On macOS and Windows: I only have Linux hardware, so I did not test playback there. What I can say is that the app already ships mp3 for every other sound (`desktop/public/sounds/*.mp3`, loaded through the same `Audio` path in `features/notifications/lib/sound.ts`), so mp3 decode is already exercised on those platforms by existing code. If someone on either can trigger a poof once, that would close it properly.

kwsantiago:
Reopened as #5729 — same commit rebased onto current `main`. This PR can't be reopened directly: it was auto-closed when the head fork was deleted, and GitHub refuses the state change even after the fork and branch were restored (`state cannot be changed. The repository that submitted this pull request has been deleted.`), since the recreated fork has a new repository id.

PR #2344, opened 2026-07-22
feat(agents): add Grok Build as a native ACP runtime

daniel-farina:
## Summary

Closes #2347.

- Registers **Grok Build** in Buzz's managed ACP runtime catalog (`id: grok`) so Desktop can discover and spawn it like Goose/Claude/Codex.
- Uses Grok's **native** ACP entrypoint (`grok agent --always-approve stdio`) — no separate `*-acp` npm adapter.
- Adds `~/.grok/bin` to binary discovery, process-sweep name `grok`, onboarding visibility, docs, and unit tests for default arg normalization.

## Why

Grok Build already speaks [ACP](https://agentclientprotocol.com/) over stdio (`grok agent stdio`). Users currently cannot select it as a first-class Buzz harness without a custom command. This mirrors the Pi/Cursor runtime registration pattern.

## What changed

| Area | Change |
|------|--------|
| `KNOWN_ACP_RUNTIMES` | New Grok Build entry |
| Default args | `agent,--always-approve,stdio` (also maps bare legacy `acp` → runtime default) |
| PATH | Discovers `~/.grok/bin` |
| Onboarding | Adds `grok` to runtime order + readiness CLI path |
| Docs | README / TESTING / `buzz-acp` README |
| Tests | `normalizes_grok_args_to_agent_stdio` (buzz-acp + desktop) |

## Validation

- [x] `cargo test -p buzz-acp --lib normalizes_` — 6 passed (incl. Grok)
- [x] Live ACP handshake: `BUZZ_ACP_AGENT_COMMAND=grok buzz-acp models` lists Grok models
- [x] Live harness against hosted relay: `buzz-acp` + Grok connected as Fizz, presence online, channel reply E2E
- [ ] Desktop full rebuild / Playwright (sidecar build path not fully exercised in CI notes)

## How to try

```bash
curl -fsSL https://x.ai/cli/install.sh | bash
grok login   # or export XAI_API_KEY=...

export BUZZ_PRIVATE_KEY=nsec1...
export BUZZ_RELAY_URL=wss://your-relay
export BUZZ_ACP_AGENT_COMMAND=grok
buzz-acp
```

Then @mention the agent from Buzz Desktop (owner-only by default).

## Checklist

- [x] New behavior has tests
- [x] User-facing setup is documented
- [x] No new `unwrap()` in production code
- [x] No unsafe code

PR #2812, opened 2026-07-25
feat(agents): add Grok Build as a first-class ACP runtime

### LINKED ISSUE #2347: feat(agents): add Grok Build as a first-class ACP runtime
## Summary

Buzz Desktop's managed agent runtime catalog supports Goose, Claude Code, Codex, and Buzz Agent, but **Grok Build** is missing even though it already speaks [ACP](https://agentclientprotocol.com/) natively over stdio:

```bash
grok agent stdio
```

Users who want Grok-backed agents must configure a custom harness command instead of selecting Grok from Agent runtimes / onboarding.

## Why

- Grok Build ships a stable ACP server (`grok agent stdio`) — no separate `*-acp` npm adapter is required (same class of integration as Goose's `goose acp` / Cursor's native ACP path).
- Install path is well-known (`curl -fsSL https://x.ai/cli/install.sh | bash`, binary under `~/.grok/bin`).
- Without catalog registration, Desktop cannot discover, install-hint, or process-sweep Grok the way it does other runtimes.

## Proposed solution

Register Grok Build in `KNOWN_ACP_RUNTIMES` with:

| Field | Value |
|-------|--------|
| `id` | `grok` |
| `label` | Grok Build |
| `commands` | `grok` |
| Default args | `agent --always-approve stdio` |
| PATH discovery | include `~/.grok/bin` |
| Skills | `.grok/skills` |
| Auth hint | `grok login` or `XAI_API_KEY` |

Also wire onboarding visibility, docs (`buzz-acp` README / TESTING), and unit tests for arg normalization.

## Acceptance criteria

- [ ] **Grok Build** appears under Desktop → Agent runtimes when `grok` is on PATH
- [ ] Creating/starting a managed agent with runtime=Grok spawns `grok agent … stdio` via `buzz-acp`
- [ ] `buzz-acp models` with `BUZZ_ACP_AGENT_COMMAND=grok` completes an ACP handshake
- [ ] Docs describe the Grok path alongside Goose/Claude/Codex

## Related

- Pattern references: Pi ACP (#2257), Cursor ACP (#1683)
- Grok ACP docs: `grok agent stdio` / Agent Client Protocol mode

maikunari:
> **Status for reviewers:** registry, UI wiring, docs, and protocol compatibility are fully verified (tests + raw ACP replay of this harness's exact traffic against grok v0.2.101). The live `AcpClient::spawn` → grok path is currently blocked by a grok-side spawn-context bug — full repro dossier in [this comment](https://github.com/block/buzz/pull/2812#issuecomment-5076872753). The registry entry is correct and inert until grok fixes stdio spawning; happy to retitle if maintainers prefer to land it as plumbing-ahead-of-runtime.

Closes #2347.

## What

Registers **Grok Build** in `KNOWN_ACP_RUNTIMES` so the Desktop app can discover, install-hint, and manage Grok agents like the other runtimes. Grok speaks ACP natively over stdio (`grok agent stdio`) — same integration class as Goose, no npm adapter.

- **Desktop registry** (`managed_agents/discovery.rs`): entry per the issue's table — id `grok`, label "Grok Build", command `grok`, default args `agent --always-approve stdio`, PATH discovery for `~/.grok/bin`, skills dir `.grok/skills`, login hint `grok login` / `XAI_API_KEY`
- **buzz-acp** (`config.rs`): mirrored default args so `BUZZ_ACP_AGENT_COMMAND=grok` works without `BUZZ_ACP_AGENT_ARGS`
- **Frontend**: `grok` added to the statically-classified "locked" provider set in `personaRuntimeModel.ts` (xAI-only, CLI-login — same class as claude/codex). Catalog-driven UI (onboarding, Agent runtimes, Doctor) picks the entry up with no further changes
- **Docs**: `crates/buzz-acp/README.md` + `TESTING.md` runtime lists, with the grok env recipe
- **Tests**: arg normalization + avatar resolution (Rust, both crates), locked-classification (TS)

## Key decisions (flagging for maintainer confirmation)

- **`--always-approve` in default args** (as proposed in #2347): intentional — Buzz owns the approval layer for managed agents, and Grok-side tool prompts would deadlock a headless stdio session. Noted in code comments in both `default_agent_args` sites.
- **No `auth_probe_args`**: grok 0.2.x has no verified non-interactive auth-status command (`grok auth status` doesn't exist; `grok models` exits 0 when logged in but its logged-out exit code is unverified). Probe intentionally omitted rather than guessed — documented in a code comment. Happy to add one if maintainers know a reliable probe.
- **Avatar**: `https://x.ai/icon.png` (verified 200). Swap if there's a preferred asset.
- **No Windows installer commands**: xAI documents only the Unix script (`curl -fsSL https://x.ai/cli/install.sh | bash`); Windows users fall back to that hint + the instructions URL.

## Testing

- `cargo fmt --all --check` ✅ · `cargo clippy` (`buzz-desktop` lib, `buzz-acp`) `-D warnings` ✅
- Desktop unit tests: 78/78 (`managed_agents::discovery`) ✅ · buzz-acp: 122/122 (`config`) ✅ · frontend `personaRuntimeModel`: 18/18 ✅
- **Live protocol verification** against real grok v0.2.101 on macOS: `initialize` and `session/new` (the exact calls `buzz-acp models` makes) both complete over stdio with `agent --always-approve stdio` — capabilities, authMethods, and sessionId all returned correctly.
- One honest caveat: the full `BUZZ_ACP_AGENT_COMMAND=grok buzz-acp models` acceptance run was blocked in my sandboxed environment because grok opens `/dev/tty` at startup (the sandbox has no controlling terminal — plain `echo > /dev/tty` fails there too). This is an artifact of the sandbox, not the integration; verification on a normal interactive machine is being run as a follow-up and I'll post the output on this PR.

`just ci`'s mobile lane not run locally (no Flutter toolchain); no mobile code touched.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -179,8 +179,8 @@ header fallback. There is no REST API for fetching message threads — use
 
 ## ACP Harness (optional, end-to-end with a real agent)
 
-`buzz-acp` connects an ACP-speaking agent (goose, codex, claude code,
-buzz-agent) to the relay. The harness listens for events, drives the
+`buzz-acp` connects an ACP-speaking agent (goose, grok build, codex,
+claude code, buzz-agent) to the relay. The harness listens for events, drives the
 agent over stdio, and the agent replies through MCP tools.
 
 Minimum recipe — assumes the relay from step 3 is running and the channel
@@ -221,10 +221,12 @@ buzz-acp                                    # foreground; logs to stdout (run in
 ```
 
 > **Using a different ACP agent?** The default recipe assumes `goose` is on
-> `$PATH` and configured (`goose --version` should print). For codex / claude
-> code / buzz-agent, set `BUZZ_ACP_AGENT_COMMAND` and `BUZZ_ACP_AGENT_ARGS`
-> accordingly — see `crates/buzz-acp/README.md`. Without these, buzz-acp
-> will fail to spawn the agent subprocess on startup.
+> `$PATH` and configured (`goose --version` should print). For grok build /
+> codex / claude code / buzz-agent, set `BUZZ_ACP_AGENT_COMMAND` and
+> `BUZZ_ACP_AGENT_ARGS` accordingly — see `crates/buzz-acp/README.md`
+> (grok build: `BUZZ_ACP_AGENT_COMMAND=grok` with args
+> `agent --always-approve stdio`). Without these, buzz-acp will fail to
+> spawn the agent subprocess on startup.
 
 If you started the agent before adding it to the channel, just run the
 `add-member` afterwards — it picks up the membership notification live and
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -9,7 +9,7 @@ Buzz Relay ──WS──→ buzz-acp ──stdio──→ Your Agent
                                        (send_message, etc.)
 ```
 
-Supports any agent that speaks [ACP](https://agentclientprotocol.com/) over stdio: **goose**, **codex** (via [codex-acp](https://github.com/agentclientprotocol/codex-acp)), and **claude code** (via [claude-agent-acp](https://github.com/agentclientprotocol/claude-agent-acp)).
+Supports any agent that speaks [ACP](https://agentclientprotocol.com/) over stdio: **goose**, **grok build** (native, via `grok agent stdio`), **codex** (via [codex-acp](https://github.com/agentclientprotocol/codex-acp)), and **claude code** (via [claude-agent-acp](https://github.com/agentclientprotocol/claude-agent-acp)).
 
 ## Prerequisites
 
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -617,6 +617,14 @@ pub(crate) fn normalize_agent_command_identity(command: &str) -> String {
 fn default_agent_args(command: &str) -> Option<Vec<String>> {
     match normalize_agent_command_identity(command).as_str() {
         "goose" => Some(vec!["acp".to_string()]),
+        // `--always-approve` is intentional: Buzz owns the approval layer for
+        // managed agents, so Grok-side tool prompts would deadlock a headless
+        // stdio session.
+        "grok" | "grok-build" => Some(vec![
+            "agent".to_string(),
+            "--always-approve".to_string(),
+            "stdio".to_string(),
+        ]),
         "codex" | "codex-acp" | "claude-agent-acp" | "claude-code-acp" | "claude-code"
         | "claudecode" | "buzz-agent" => Some(Vec::new()),
         _ => None,
@@ -1438,6 +1446,19 @@ mod tests {
     fn normalizes_goose_args_to_acp() {
         assert_eq!(normalize_agent_args("goose", Vec::new()), vec!["acp"]);
         assert_eq!(normalize_agent_args("goose", vec!["".into()]), vec!["acp"]);
+    }
+
+    #[test]
+    fn normalizes_grok_args_to_stdio_default() {
+        let expected = vec!["agent", "--always-approve", "stdio"];
+        assert_eq!(normalize_agent_args("grok", Vec::new()), expected);
+        assert_eq!(normalize_agent_args("grok", vec!["".into()]), expected);
+        assert_eq!(normalize_agent_args("grok-build", Vec::new()), expected);
+        // Explicit args are preserved verbatim — no default-arg injection.
+        assert_eq!(
+            normalize_agent_args("grok", vec!["agent".into(), "stdio".into()]),
+            vec!["agent", "stdio"]
+        );
     }
 
     #[test]
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -60,7 +60,12 @@ fn is_subcommand(name: &str) -> bool {
 }
 
 /// Timeout for lightweight helper subcommands (spawn + initialize + model/method probes).
-const MODELS_TIMEOUT: Duration = Duration::from_secs(10);
+///
+/// 30s rather than 10s: some agents boot every user-configured MCP server
+/// during `session/new` (observed with Grok Build, where a handful of stdio
+/// MCP servers pushed session creation past 10s on a real machine). The
+/// happy path is unaffected — fast agents still return in ~2-5s.
+const MODELS_TIMEOUT: Duration = Duration::from_secs(30);
 
 /// Timeout for `buzz-acp authenticate`. Browser-based vendor auth can require
 /// human interaction, so it must not share the short probe timeout.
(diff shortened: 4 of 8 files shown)


maikunari:
**Verification follow-up** (as promised in the PR description) — registry/UI/tests all hold, but the `buzz-acp models` acceptance run surfaced what looks like a grok-side spawn-context sensitivity that maintainers should know about. Full dossier:

## What works everywhere ✅

Raw ACP over plain shell pipes against **grok v0.2.101 (macOS, aarch64)** — using the *exact* params buzz-acp sends (protocolVersion 2 pin, fs clientCapabilities):

```
$ printf '<initialize>\n<session/new>\n' | grok agent --always-approve stdio
→ initialize result (agent negotiates down to protocolVersion 1) in ~1s
→ session/new result with sessionId AND a `models` object in ~3-5s
```

Verified in an interactive zsh (foreground pgroup, tty) and in a no-tty sandbox shell. Framing is clean ndjson (largest observed line 87KB, well under MAX_LINE_SIZE).

## What fails: the same two calls through `AcpClient::spawn` ❌

`BUZZ_ACP_AGENT_COMMAND=grok buzz-acp models` fails in every spawn context tried:

| Context | Result |
|---|---|
| Interactive terminal (tty, agent in background pgroup via `process_group(0)`) | `agent timed out (30s)` — grok process observed **already exited** while buzz-acp still waits |
| Same, with `--no-leader` / `TERM=dumb NO_COLOR=1` / stderr→`/dev/null` | same timeout, or instant silent exit |
| PTY via `script(1)` | timeout; with grok's stderr redirected to a file grok *stays alive* but responses never arrive through the pipes |
| No-tty sandbox | grok exits immediately: `Error: Device not configured (os error 6)` (ENXIO — its seatbelt profile explicitly allows `file-ioctl` on `/dev/tty`, and plain `open("/dev/tty")` returns ENXIO in that sandbox) |

Same binary, same args, same protocol bytes — plain-pipe spawn works, tokio piped spawn + own-process-group doesn't. Python `subprocess` with `preexec_fn=os.setpgrp` *also* works, so it isn't the process group alone; grok appears to change terminal/leader initialization behavior based on some property of the spawn context and either exits early or wedges before servicing stdio.

## Read

The registry entry, args, and protocol compatibility in this PR are correct (raw replay of buzz-acp's exact traffic proves it). The remaining gap looks like a **grok CLI issue** — `grok agent stdio` should be context-independent — and will presumably affect Desktop-managed spawns too (GUI apps have no controlling tty, which is the ENXIO row above). Happy to file it with xAI with this repro; also very interested in whether anyone has a rig where `buzz-acp models` + grok completes, to rule out machine-local factors (leader config, MCP fleet size — the 10→30s timeout bump in this PR came from grok booting all user-configured MCP servers during `session/new`).

One adjacent note for anyone logging raw streams while testing: grok's `_x.ai/mcp/servers_updated` notification includes user MCP server configs **with env values (API keys) in plaintext** — worth scrubbing logs and possibly worth an upstream redaction ask independent of this PR.

Related: #2344 (earlier attempt at the same runtime, pre-dates the `cli_install_instructions_url` field rename on main).

M0xDr3w:
### Independent confirmation of the grok-side spawn-context bug

Confirms the dossier above ([comment](https://github.com/block/buzz/pull/2812#issuecomment-5076872753)): `grok agent stdio` is sensitive to spawn context / controlling TTY. Full repro with cleaner minimal cases:

📄 **https://gist.github.com/M0xDr3w/417661cf223099cfa06f277139b34bd6**

**Env:** Grok Build `0.2.112 (9bbd559437aa) [stable]`, macOS 25.5.0 arm64, zsh

| Spawn context | Result |
|---|---|
| Plain foreground pipe | ✅ `initialize` returns full capabilities |
| No controlling TTY (`script -q /dev/null … < /dev/null`) | ❌ hangs briefly, exits with no useful output |
| `TERM=dumb NO_COLOR=1 … < /dev/null` | ❌ exits 0, **no JSON** on stdout |
| Background process group (`set -m`) | ❌ suspended with `tty output` |

**Working case:**
```bash
printf '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":2,"clientInfo":{"name":"repro","version":"0.1"},"capabilities":{}}}\n' | grok agent --always-approve stdio
```

**Expected:** `grok agent stdio` should be fully context-independent (no controlling TTY / special process-group requirements), same as other mature ACP agents.

**xAI filing note:** `xai-org/grok-build` has **Issues disabled** and does not accept external PRs, so this lives as a public gist for now. Happy to re-file if xAI opens a tracker.

Related: #2344

M0xDr3w:
### Updated repro (gist only)

Expanded write-up of the grok-side stdio spawn-context bug (Grok Build **0.2.112**), including a working one-shot Python PTY path and failing long-running bridge cases:

**https://gist.github.com/M0xDr3w/1da8a9ec2223121a52979478465896ba**

**Key additions vs earlier comment:**
- One-shot Python `pty.fork()` **works** for `initialize`
- Long-running Python PTY bridge (stdin ↔ PTY) **does not** — only echoes the request
- Additional silent-fail cases: `exec </dev/null` and `TERM=dumb` with closed stdin

`xai-org/grok-build` has **Issues disabled**, so this public gist is the durable filing until xAI opens a channel.

maikunari:
@M0xDr3w — excellent confirmation, thank you. The `tty output` suspension in your background-pgroup case is the smoking gun for the SIGTTOU mechanism, and 0.2.112 confirms it's current.

One update to your filing note: there *is* a live channel for this — issues are enabled on xai-org/grok-build-plugin-cc, and this bug is filed there as [xai-org/grok-build-plugin-cc#7](https://github.com/xai-org/grok-build-plugin-cc/issues/7) (with an explicit note that it belongs to the CLI). Worth attaching your gists there so the evidence is consolidated where xAI can act on it. Also raising it with their engineers on X, which per xAI's own guidance is their preferred bug channel.

wpfleger96:
🤖 hey @maikunari — thanks for building this, and especially for the spawn-context repro dossier — that level of investigation is genuinely useful to anyone integrating Grok over stdio, independent of how this PR resolves.

Grok support landed generically in #2773 (ships in v0.5.0): Grok Build is a one-click preset launching `grok agent --always-approve stdio`, and any other ACP-over-stdio binary can be registered from Settings. New presets are a small data entry rather than per-harness Rust.

Closing as superseded. If the grok-side stdio spawn bug you documented still bites under the preset path, an issue with that dossier attached would be a great follow-up.

maikunari:
Thanks @wpfleger96 — that makes sense, and the generic preset approach in #2773 is a better foundation than a per-harness runtime. Glad Grok support is landing that way.

One thing worth flagging for preset users: since the Grok Build preset launches `grok agent --always-approve stdio`, it may hit the same stdio spawn-context sensitivity documented in the dossier here — Grok Build (confirmed on 0.2.112) silently hangs when spawned without a controlling TTY (SIGTTOU / background process-group). One-shot works; a long-running stdin↔stdio bridge only echoes the request. Full repro + minimal cases are in the gists on this PR, and it's filed upstream at xai-org/grok-build-plugin-cc#7. If the preset ever spawns grok in a non-TTY context (daemonized, over SSH, etc.), it could surface there too — happy to help validate against v0.5.0 once it's out.

Thanks again for the thoughtful close.

Issue #2802, opened 2026-07-25
[Bug] No way to sign into Buzz Desktop as a different account — browser session is reused and sign-out is local-only


AirZona:
> **Edited after further investigation.** The original report attributed a symptom I hit with an agent to this login path. I've since verified parts of it and disproven part of my own proposed fix, so I've split this into what I can demonstrate and what I only assumed. The unverified part is at the bottom. Sorry for the reshuffle — I'd rather correct it than leave a confident-sounding guess in the tracker.

## Verified: you cannot sign into Buzz Desktop as a different account

**1. The login URL carries no `prompt`, so the identity provider reuses the browser's existing session.**

```rust
// desktop/src-tauri/src/builderlab.rs:217
fn login_url(return_to: &str) -> Result<Url, String> {
    let mut login_url = api_url("/v1/auth/login")?;
    login_url
        .query_pairs_mut()
        .append_pair("type", "cli")
        .append_pair("product", "buzz")
        .append_pair("returnTo", return_to);
    Ok(login_url)
}
```

Per OIDC, omitting `prompt` permits the provider to reuse an authenticated session. The URL is handed to the OS browser as-is (`builderlab.rs:279`), so whichever account that browser holds is the account you get, with no chooser.

**2. Signing out cannot fix it, because sign-out is local-only.**

```rust
// desktop/src-tauri/src/builderlab.rs:407
pub(crate) fn clear_builderlab_auth(
    session: tauri::State<'_, BuilderlabSession>,
) -> Result<(), String> {
    *session.0.lock().map_err(|error| error.to_string())? = None;
    Ok(())
}
```

This drops Buzz's in-memory session and cannot touch the browser's IdP session. **Sign out → Sign in is therefore a no-op round trip** that re-authenticates the same account. There is no path in the app to land on a different one.

**3. The obvious client-side fix does not work — the parameter is stripped server-side.**

I tried the natural fix (send `prompt=select_account`) and verified against the live endpoint that it never arrives. `/v1/auth/login` builds its own Auth0 URL and allowlists what it forwards:

| sent to `/v1/auth/login` | reaches Auth0 |
|---|---|
| `screen_hint=signup` | ✅ |
| `prompt=select_account` | ❌ |
| `prompt=login` | ❌ |
| `login_hint=…` | ❌ |

```
[...]

PR #2821, opened 2026-07-25
feat(buzz-agent): allow forcing the account chooser on interactive login

AirZona:
Part of #2802.

> **Scope note:** this PR originally also changed the Builderlab sign-in path in Desktop. I removed that half after verifying it could never work — see [this comment](https://github.com/block/buzz/pull/2821#issuecomment-5081019602) for the evidence. What remains is the `buzz-agent` change, which talks to the provider's authorization endpoint directly.

## What problem this solves

`PkceOAuthTokenSource`'s browser flow builds its authorization URL with no `prompt` parameter:

```rust
// crates/buzz-agent/src/auth.rs, before this change
"{}?response_type=code&client_id={}&redirect_uri={}&scope={}&state={}&code_challenge={}&code_challenge_method=S256"
```

Per OIDC, omitting `prompt` lets the provider reuse an existing authenticated session. So when the flow opens a browser, whatever account that browser is already signed into is the account you get — silently, with no chooser. If it's the wrong one, there is no way to reach a different account: there's no re-auth path that forces the picker, and the on-disk token cache is keyed on `sha256(discovery_url|client_id|scopes)` (`cache_path_for`), so the cached token from the wrong account keeps satisfying requests.

## How it was implemented

- Extracted the URL construction out of `browser_pkce_flow` into a pure `authorize_url()` — this is what made it unit-testable.
- Threaded a `force_account_selection` flag through. When set, the request carries `prompt=select_account`.
- `interactive_login()` takes the flag, surfaced as `buzz-agent auth <provider> --select-account`.

### Key decision: implicit auth is unchanged

`bearer()`'s fall-through to the browser flow (step 5) passes `false`. That path is reached mid-request, not by a user who asked to switch accounts — forcing a chooser there is friction, and a provider that rejects `prompt=select_account` would start failing requests that work today. The parameter is only ever sent when someone explicitly asks for it on the command line.

Net effect on existing behavior: none. Every current call site produces a byte-identical authorization URL.

## Tests

`authorize_url_omits_prompt_by_default` and `authorize_url_requests_account_chooser_when_forced` in `crates/buzz-agent/src/auth.rs`. The first pins the default-off behavior, so a future change that starts forcing a chooser on implicit auth trips a test.

## How to test it manually

```
buzz-agent auth databricks --select-account
```
The authorization URL printed to stderr (and opened in the browser) carries `&prompt=select_account`; without the flag it doesn't.

## Follow-up deferred

The OAuth cache key doesn't include the authenticated account, so one machine still can't hold tokens for two accounts of the same provider. That's a cache-key migration and wants a maintainer opinion first — it's noted in #2802 and not addressed here.

## PR Checklist

- [x] `cargo fmt --all --check`, `cargo clippy -p buzz-agent --all-targets -- -D warnings` clean
- [x] `cargo test -p buzz-agent` (268 passing), `just test-unit` (full workspace); rebased onto current `main`
- [x] New public API documented (`authorize_url`, `interactive_login`)
- [x] No new `unsafe` blocks
- [ ] Integration tests (`just test`) — not run; needs Postgres + Redis, and this touches none of the `buzz-relay`, `buzz-db`, or `buzz-auth` crates

### DIFF
--- a/crates/buzz-agent/src/auth.rs
+++ b/crates/buzz-agent/src/auth.rs
@@ -232,9 +232,18 @@ impl PkceOAuthTokenSource {
 
     /// Run the full browser-mediated Authorization Code + PKCE flow.
     /// Caller must hold a TTY/browser: this opens a window and blocks.
-    pub async fn interactive_login(&self) -> Result<(), AgentError> {
+    ///
+    /// `force_account_selection` adds `prompt=select_account` to the
+    /// authorization request. Without it the provider silently reuses whatever
+    /// session the default browser already holds, so a user signed into the
+    /// wrong account has no way to reach a different one. Pass `true` for
+    /// explicitly user-initiated logins ("use a different account"); leave it
+    /// `false` on implicit first-run auth, where an unnecessary account chooser
+    /// is friction and some providers reject the parameter.
+    pub async fn interactive_login(&self, force_account_selection: bool) -> Result<(), AgentError> {
         let endpoints = self.endpoints().await?;
-        let token = browser_pkce_flow(&self.http, &self.cfg, &endpoints).await?;
+        let token =
+            browser_pkce_flow(&self.http, &self.cfg, &endpoints, force_account_selection).await?;
         let mut state = self.state.lock().await;
         self.save(&mut state, token)?;
         Ok(())
@@ -289,8 +298,10 @@ impl TokenSource for PkceOAuthTokenSource {
             }
         }
 
-        // 5. No usable cache: full browser dance.
-        let fresh = browser_pkce_flow(&self.http, &self.cfg, &endpoints).await?;
+        // 5. No usable cache: full browser dance. No forced account chooser —
+        //    this path is reached implicitly mid-request, not by a user who
+        //    asked to switch accounts (see `interactive_login`).
+        let fresh = browser_pkce_flow(&self.http, &self.cfg, &endpoints, false).await?;
         let bearer = fresh.access_token.clone();
         self.save(&mut state, fresh)?;
         Ok(bearer)
@@ -521,6 +532,34 @@ fn random_state() -> Result<String, AgentError> {
     Ok(base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(bytes))
 }
 
+/// Build the authorization-endpoint URL for one PKCE attempt.
+///
+/// When `force_account_selection` is set the request carries
+/// `prompt=select_account`, which tells the provider to show its account
+/// chooser instead of silently reusing the browser's current session.
+fn authorize_url(
+    endpoints: &OidcEndpoints,
+    cfg: &PkceOAuthConfig,
+    redirect_uri: &str,
+    state: &str,
+    challenge: &str,
+    force_account_selection: bool,
+) -> String {
+    let mut url = format!(
+        "{}?response_type=code&client_id={}&redirect_uri={}&scope={}&state={}&code_challenge={}&code_challenge_method=S256",
+        endpoints.authorization_endpoint,
+        urlencoding::encode(&cfg.client_id),
+        urlencoding::encode(redirect_uri),
+        urlencoding::encode(&cfg.scopes.join(" ")),
+        urlencoding::encode(state),
+        urlencoding::encode(challenge),
+    );
+    if force_account_selection {
+        url.push_str("&prompt=select_account");
+    }
+    url
+}
+
 /// Spin up a localhost callback server, open the authorize URL in a
 /// browser, wait up to [`BROWSER_AUTH_TIMEOUT`] for the redirect, then
 /// exchange the code for a token.
@@ -528,6 +567,7 @@ async fn browser_pkce_flow(
     http: &Client,
     cfg: &PkceOAuthConfig,
     endpoints: &OidcEndpoints,
+    force_account_selection: bool,
 ) -> Result<CachedToken, AgentError> {
     use axum::{extract::Query, response::Html, routing::get, Router};
     use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


AirZona:
> **Edited — outdated.** This comment shows the Desktop "Use a different account" button, which has since been removed from this PR: Builderlab's `/v1/auth/login` strips the `prompt` parameter, so that button could never have worked. Evidence in [this comment](https://github.com/block/buzz/pull/2821#issuecomment-5081019602). The PR is now two files in `crates/buzz-agent`. Keeping the original below rather than deleting it, since the later comment refers back to these screenshots.

<details>
<summary>Original comment (superseded)</summary>

### Screenshots

Both states of **Settings → Hosted communities**, captured with the repo's mock-bridge tooling.

**Signed out** — the new secondary action sits next to the primary sign-in, with a line explaining why you'd want it:

signed out

**Signed in** — "Use a different account" next to Sign out. This is the state the bug is about: the account shown is whichever one the browser session happened to hold, and before this change there was no way to get to another one.

signed in

<sub>Hosted from a commit-pinned `raw.githubusercontent.com` URL on my fork rather than via `scripts/post-screenshots.sh`, which pushes to a branch on `block/buzz` and needs write access I don't have as an outside contributor. Same immutable-SHA property, no relay-media URLs.</sub>

### CI is waiting on approval

The `CI` and `Docker image` runs are sitting at `action_required` (first-time-contributor gate), so only DCO / Semgrep / zizmor have reported. Could a maintainer approve the workflow run? Everything that can be run locally is green:

- `just test-unit` (full workspace), `just desktop-test` (3487), `just desktop-tauri-test`
- `just desktop-check`, `cargo fmt --all --check`, clippy `-D warnings` on `buzz-agent` + Tauri
- The new `hosted-communities-account-selection.spec.ts` (3 tests) passes under `--project=smoke`

### This is safe to merge independently of any Builderlab-side change

Worth stating plainly, since I flagged the Auth0 question in the description and it shouldn't hold the PR up:

- `force_account_selection` defaults to `false`, so every existing call site produces a byte-identical authorization URL. The change is a strict no-op unless the user clicks the new button.
- If Auth0 doesn't forward `prompt=select_account` upstream, the new button degrades to exactly today's behavior — nothing regresses, and the Desktop half is already correct for whenever the forwarding is configured.
- Implicit first-run auth in `buzz-agent` deliberately still passes `false`, so providers that reject the parameter never receive it.

Happy to split the `crates/buzz-agent` half into its own PR if you'd rather review the desktop and agent surfaces separately — just say the word.



</details>

AirZona:
> **Edited — most of this is outdated.** I answered question 2 myself with `curl` about six minutes after posting: Builderlab's `/v1/auth/login` does **not** forward `prompt`. The Desktop half of this PR has been removed as a result; evidence and details are in [this comment](https://github.com/block/buzz/pull/2821#issuecomment-5081019602). Only the CI-approval ask below is still live. Original text kept intact underneath so the thread reads honestly. Sorry for the notification churn.

**Still relevant: could someone approve the workflow run?** It's sitting at `action_required` on the first-time-contributor gate, so only DCO/Semgrep/zizmor have reported. The PR is now two files in `crates/buzz-agent`; `cargo test -p buzz-agent` (268), `just test-unit`, fmt and clippy `-D warnings` are green locally, rebased onto current `main`.

@wpfleger96 — `crates/buzz-agent/src/auth.rs` is yours per git history, so the remaining diff is squarely in your area. @baxen — the Builderlab/Desktop question is no longer part of this PR, so nothing here needs you unless you want to weigh in on the server-side finding in #2802.

<details>
<summary>Original comment (superseded)</summary>

@baxen @wpfleger96 — flagging you two since git history says you own the surfaces this touches (`builderlab.rs` + the hosted-communities card, and `buzz-agent/src/auth.rs` respectively). Two things:

**1. Could someone approve the workflow run?** It's sitting at `action_required` on the first-time-contributor gate, so only DCO/Semgrep/zizmor have reported. Everything runnable locally is green, including the new Playwright spec.

**2. One question only someone inside Block can answer:** does the Builderlab `/v1/auth/login` path forward `prompt` through Auth0 to the upstream connection? This PR makes Desktop send `prompt=select_account`, which is correct at the OIDC layer, but if Auth0 swallows it rather than forwarding it, the button won't reach Google's chooser and a server-side counterpart is needed.

To be clear, that question doesn't block merging: the parameter is only ever sent when the user clicks the new button, and if it isn't forwarded the button degrades to exactly today's behavior. Nothing regresses either way — the Desktop half is already correct for whenever the forwarding is in place.

I rebuilt on top of current `main` after #2738 and the new hosted-communities screenshot spec landed; conflicts in `playwright.config.ts` and the settings helper are resolved and `main`'s specs are registered alongside mine.

Happy to make any changes you want, including splitting the `crates/buzz-agent` half into a separate PR if you'd rather the desktop and agent surfaces land independently.

</details>

AirZona:
<a name="issuecomment-scope"></a>**Scope change: I removed the Desktop half of this PR. It could never have worked, and I'd rather say so than have you find it in review.**

I went to verify my own open question — whether `prompt=select_account` survives the hop through Builderlab to Auth0 — and it doesn't. `/v1/auth/login` builds its own authorize URL and allowlists what it forwards:

```
$ curl -sI "https://app.builderlab.xyz/api/goose/v1/auth/login?type=cli&product=buzz&prompt=select_account&returnTo=http%3A%2F%2F127.0.0.1%3A9999%2Fcb" | grep -i location
location: https://login.builderlab.xyz/authorize?client_id=…&response_type=code&redirect_uri=…&scope=openid+profile+email&state=…&nonce=…&code_challenge=…&code_challenge_method=S256&organization=org_…&ext-product=buzz
```

No `prompt`. Same for `login_hint`. `screen_hint`, by contrast, **is** forwarded — which is presumably why `login_defaults_to_auth0_login` asserts on its absence today:

| sent to `/v1/auth/login` | reaches Auth0 |
|---|---|
| `screen_hint=signup` | ✅ |
| `prompt=select_account` | ❌ |
| `prompt=login` | ❌ |
| `login_hint=…` | ❌ |

So the "Use a different account" button I'd added to **Settings → Hosted communities** would have looked like it did something and silently done nothing. That's worse than the bug it was meant to fix, so it's gone, along with its tests and screenshots. What's left is only `crates/buzz-agent`, which builds its authorization URL from the provider's own OIDC discovery document and hits the authorization endpoint directly — nothing in between to strip the parameter.

**For whoever owns the Builderlab service:** adding `prompt` to the same passthrough that already handles `screen_hint` would unblock a real Desktop fix. As it stands there is no way to sign into Buzz as a different account from the app — `clear_builderlab_auth` only drops Buzz's in-memory session (`desktop/src-tauri/src/builderlab.rs:407`) and cannot touch the browser's Auth0 session, so Sign out → Sign in silently re-authenticates the same account. I'm happy to write that Desktop change once the parameter gets through; it's a small diff on top of what I already had working.

Two caveats on my own evidence, so nobody over-reads it: I only probed the redirect the endpoint issues, so I can't say whether Auth0 itself would honor `prompt=select_account` once forwarded, or which connections that org tenant has enabled. And #2802 frames this around an agent of mine failing to reach Google APIs under the right account — I've since separated the parts of that report I actually verified from the part I assumed.

CI is still awaiting first-time-contributor approval, so the checks on this PR reflect only DCO/Semgrep/zizmor. `cargo test -p buzz-agent` (268), `just test-unit`, fmt and clippy `-D warnings` are green locally on the trimmed branch.

PR #2304, opened 2026-07-21
fix(desktop): re-arm relay-mesh runtime when the ingress is dead

Bartok9:
Refs #2062

Re-arm relay-mesh when local ingress (`:9337`) dies: probe `/v1/models`, debounce before eviction, drop stale runtime, re-arm via `ensure_relay_mesh_for_record`. Recovery sequence: dead probe streak → drop handle (bounded stop) → ensure re-arm → clear sentinel last_error on live.

Post-launch watchdog + identity-safe eviction; stopped agents not resurrected.

Issue #2062, opened 2026-07-18
Relay-mesh agents silently stop replying if mesh runtime drops after launch (no self-heal)


micspiral:
## Summary

Relay-mesh (Buzz shared compute) agents **silently stop replying** if the mesh runtime drops at any point after app launch. No error surfaces to the user and nothing re-arms the runtime until an agent is manually (re)started.

## Repro

1. Launch desktop with a relay-mesh agent; mesh runtime comes up, ingress on `127.0.0.1:9337`, bot replies fine.
2. Leave the app running for an extended period (overnight), or otherwise have the embedded mesh runtime exit/wedge.
3. Send the bot a message → **no reply, no error**.
4. Manually start/create an agent → `ensure_relay_mesh_for_record` runs again, re-arms `9337`, replies resume.

Observed live: `9337/3131 NOT listening` while `goose acp` agent subprocesses were still running and healthy; `curl 127.0.0.1:9337/v1/models` → connection refused. A remote peer (M4) was serving the model the whole time, so compute availability was never the issue.

## Root cause

`ensure_relay_mesh_for_record` (which brings up the local mesh client node / OpenAI ingress) is only invoked:

- at launch — `managed_agents/restore.rs` (`restore_managed_agents_on_launch`)
- on agent create/start — `commands/agents.rs` (`start_local_agent_with_preflight`)

There is **no periodic re-arm and no liveness check** tying the mesh runtime to agent message dispatch. If `state.mesh_llm_runtime` becomes `None` (runtime exited, or a bad restart left a zombie — cf. `"timed out after 5s waiting for embedded mesh runtime thread to exit"`), relay-mesh agents keep running but point at a dead ingress. The failure is invisible.

## Impact

Any long-lived desktop session using shared-compute agents will eventually see bots go quiet with no indication why. This is the dominant "shared compute is broken" user-facing symptom even though discovery, transport, and remote serving are all healthy.

## Suggested fix (options)

1. **On-demand re-arm:** before dispatching a message to a relay-mesh agent (or in the agent reconcile loop), if `mesh_llm_runtime.is_none()` or the ingress is unreachable, call `ensure_relay_mesh_for_record` / `ensure_client_node_for_model` to bring it back.
2. **Liveness + surfaced error:** health-check the runtime periodically; if it's down and can't be restored (e.g. no live serving member), surface a visible "shared compute offline" state on the agent instead of silent non-reply.
[...]

PR #2823, opened 2026-07-25
fix(desktop): supervise and re-arm relay-mesh runtime

micspiral:
Refs #2062

This carries forward the relay-mesh recovery work from #2304 by @Bartok9 (cherry-picked with original authorship/sign-offs) and adds the startup/readiness supervision and packaging fixes found while validating it against a real two-machine Buzz setup.

## From #2304

- Watch the local OpenAI ingress (`:9337`) after launch and re-arm a stale relay-mesh runtime.
- Require consecutive failed probes before eviction so a transient inference stall does not cause a cold restart.
- Bound stale-runtime shutdown, preserve runtime identity across the asynchronous probe, and never evict a concurrent replacement.
- Re-arm only for agents that are actually running; deliberately stopped agents stay stopped.
- Persist an actionable sentinel error under the managed-agent store lock and clear only that error after recovery.
- Treat serve-to-client fallback as an intentional fail-safe; configured serve restoration remains on its existing path.

## Added here

- Use the inference ingress (`:9337`), rather than management port `:3131`, as Buzz's client-readiness boundary. A usable client no longer fails or holds agent-save open merely because management startup is still pending.
- Supervise the embedded SDK startup asynchronously, publish a pending status while management is unavailable, and keep its mesh identity alive.
- Avoid racing a replacement while SDK startup still owns the embedded runtime. If that pending startup later loses ingress while a running agent still needs it, request a controlled Buzz restart to reclaim the otherwise-unreachable SDK thread.
- Defer roster-driven replacement while client management startup is pending.
- Keep post-launch recovery in a dedicated module so the mesh entry point remains within the desktop file-size gate.
- Explicitly mark generated Unix sidecars executable. On macOS, copying over an existing non-executable destination preserved its old mode, causing packaged `buzz-acp`, `buzz-agent`, and tool sidecars to be reported as missing.

## Validation

Automated:

- `just ci` — passed, including formatting, Clippy with warnings denied, desktop/web/mobile checks and tests, and builds.
- Full Tauri mesh-feature suite — 1,702 passed, 0 failed, 15 ignored.
- Mesh-feature Clippy with `-D warnings` — passed.
- Release macOS app bundle with `mesh-llm` — built successfully; every bundled sidecar passed executable-mode and deep code-signature verification.

Live two-machine E2E:

- M5: released Buzz serving `unsloth/gemma-4-26B-A4B-it-GGUF:UD-Q4_K_M`.
- Mac mini: this branch's packaged Buzz running a saved `buzz-agent`/relay-mesh agent.
- Confirmed `:9337` accepted inference and the saved ACP harness started with no error while `:3131` was still unavailable.
- Exact inference succeeded before restart (`CORRECTED-MINI-E2E-OK`).
- Bespoke Buzz shut down cleanly in 2.3s, then restored ingress and the saved harness in 18.8s while `:3131` was still unavailable.
- Exact inference succeeded after restart (`AFTER-RESTART-E2E-OK`).
- A real Buzz `@C55` message traversed desktop → `buzz-acp` → `buzz-agent` → mini `:9337` → M5 compute and published the requested reply successfully.

### DIFF
--- a/desktop/scripts/check-file-sizes.mjs
+++ b/desktop/scripts/check-file-sizes.mjs
@@ -400,7 +400,16 @@ const overrides = new Map([
   // transition lock doc broadened to cover all protected-PID transitions, and
   // clear_agent_session_caches (per-pubkey retain) added alongside the
   // per-key clear. Load-bearing identity-contract change; queued to split.
-  ["src-tauri/src/app_state.rs", 1081],
+  // +8 (1081 -> 1089): mesh re-arm watchdog (#2304) adds mesh_ingress_dead_probes
+  // AtomicU32 counter on AppState for consecutive dead-probe debounce. Narrow
+  // counter field + init spars; still queued to split more AppState sections.
+  ["src-tauri/src/app_state.rs", 1089],
+  // mesh_llm re-arm watchdog + Brad/micspiral correctness suite (#2304):
+  // probe/evict identity gate, bounded stop timeout, process-map filter,
+  // store-lock error persistence, consecutive-probe debounce, sentinel-cleared
+  // last_error. Load-bearing recovery path; queued to split re-arm helpers
+  // into mesh_llm/rearm.rs. Ratchet covers current size after rustfmt.
+  ["src-tauri/src/commands/mesh_llm.rs", 1410],
   // multi-slot splitting + no-op suppression (#1309): the ReadStateManager
   // class grew from ~700 lines to ~1019 with the addition of
   // splitContextsIntoBudgetedSlots (pure fn + 5 tests), publishSplitSlots,
--- a/desktop/scripts/check-file-sizes.mjs
+++ b/desktop/scripts/check-file-sizes.mjs
@@ -400,16 +400,10 @@ const overrides = new Map([
   // transition lock doc broadened to cover all protected-PID transitions, and
   // clear_agent_session_caches (per-pubkey retain) added alongside the
   // per-key clear. Load-bearing identity-contract change; queued to split.
-  // +8 (1081 -> 1089): mesh re-arm watchdog (#2304) adds mesh_ingress_dead_probes
-  // AtomicU32 counter on AppState for consecutive dead-probe debounce. Narrow
-  // counter field + init spars; still queued to split more AppState sections.
-  ["src-tauri/src/app_state.rs", 1089],
-  // mesh_llm re-arm watchdog + Brad/micspiral correctness suite (#2304):
-  // probe/evict identity gate, bounded stop timeout, process-map filter,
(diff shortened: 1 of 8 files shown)


Bartok9:
### Contributor note from #2304 (Bartok9)

@micspiral — thank you for taking this over after validating against a real two-machine setup. Moving here as a fellow contributor (happy for you to keep driving the code).

**Credit:** Clean and appreciated. PR body credits #2304; 7 commits remain authored/signed as **Bartok9** (`ee358742ff`…`1032e72e29`), then your two additions (`694beb5` supervise client readiness · `3541834` sidecar `chmod`). Exactly the pattern we hoped for.

**What I dual-checked against the #2304 intent + your.additions**
- Ingress contract is the agent-facing one (`:9337` `/v1/models`), not management `:3131` — matches Brad’s call-path review class and remove the “alive handle, dead ingress” hang.
- Debounce / streak keyed to **runtime id**, foreground closed-port vs watchdog consecutive failures, identity pin before `take()`, bounded stop + port-release wait, serve→client fail-safe documented — still hold.
- Only **running** relay-mesh agents re-arm; stopped agents stay stopped; sentinel-prefixed `last_error` clear-on-success only — still hold.
- Your **Starting** client path + independent ingress supervision + no agent-save hang on management pending — excellent operational catch; tests for pending status / failed startup reinforce it.
- Sidecar `chmod 755` after `cp` on Unix — good packaging fix for the placeholder-mode trap.

**Checks:** Desktop/E2E/DCO/lint green from here; MERGEABLE, waiting on review protections.

**Optional polish** (your call — zero pressure; offering eyes only):
1. **RestartRequired from the background watchdog** — gated on already-running consumers, which is right. If users ever report surprise app restarts during first client spin-up after a cold install with many mesh agents auto-starting, a user-visible toast before `request_restart()` might soften the landing (keep the restart itself).
2. **Ingress probe “success”** — currently any successful `/models` counts as Live. If the SDK ever returns 200 with an empty/degraded model list while inference is wedged, you may want a secondary signal later; not blocking, just a future follow-up if multi-base dogfood sees it.
3. **Watchdog backoff** — errors double sleep (15→120s); Thor `Ok(())` while still Debouncing resets to 15s, which mirrors “keep sampling while debouncing.” Intentional and fine; noting only so reviewers don’t reverse it thinking it’s thrash.

I will **close #2304** as superseded by this PR (full cherry-pick + your ops layer). Happy to review further revisions, reproduce dogfood nits, or ship a small follow-up PR if you want a fix coded passively — or stay in review-only mode if you prefer to implement everything yourself.

Refs #2304 · #2062

tlongwell-block (approved):

Issue #2816, opened 2026-07-25
Self-host onboarding dead-ends: 'I already have a community' → 'I own the community' only offers Builderlab, never a relay URL


ryanmitchellneely:
## Summary

An operator who runs their own relay and owns it picks the two semantically correct answers in onboarding and lands in a hosted-account flow with no way to enter their relay. The self-host path exists, but it is behind **Join a community**, which reads like the option for someone joining *somebody else's* workspace.

## Steps to reproduce

1. Stand up a self-hosted relay (auth + membership required), add your own pubkey via `buzz-admin add-member`
2. Desktop app → **Use an existing key** with the owner identity
3. **I already have a community** → **I own the community**

## Actual

A modal appears: *Set up your community — Sign in to connect a community you already own or create a new one. We'll open Builderlab in your browser, then bring you back to Buzz.* The only action is **Sign in to continue**, which opens Builderlab and creates a hosted account. At no point does the flow ask for a relay URL, and dismissing the modal returns to the same fork.

## Expected

The path that says *I own the community* should be the path where a self-hoster types their relay URL. At minimum, an escape hatch on that modal (`Connect a self-hosted relay instead`).

## What actually works

**Join a community** → paste `wss://relay.example.com` → the app connects, resolves the community, and proceeds to profile setup normally. That is the correct flow for a self-hoster; it is just labelled as if it were not.

## Impact

This is the first five minutes of the product for exactly the audience the OSS repo is aimed at. In our case it cost about forty minutes and produced a stray hosted account before we found the right door. Two smaller things compound it: the relay-owner identity is minted with `buzz-admin generate-key` in hex, which the key field rejects (#2815), and there is no visible relay-URL field until you take the Join branch.

## Suggested fix

Reword the fork, or have **I own the community** offer both `Connect my relay` and `Use Builderlab hosting`. Happy to send a PR if you want the former.

Chessing234:
Fixed in #2824 — owner path now offers self-hosted relay URL or Builderlab.

varunyn:
Thanks for creating fix. Waiting for it to merge. Can't use mac app with self host right now

sayre4ux:
Temporary workaround for self-hosted macOS users:

  1. Choose **I already have a community → I’m a member or admin**.
  2. Enter the relay URL, e.g. `ws://192.168.50.12:3000`.
  3. If Buzz shows `Load failed`, set this in `deploy/compose/.env`:

     ```env
[...]

Issue #2815, opened 2026-07-25
`buzz-admin generate-key` emits a hex secret that desktop onboarding rejects (expects `nsec1`)


ryanmitchellneely:
## Summary

The bootstrapping path and the desktop onboarding path disagree on key encoding. `buzz-admin generate-key` prints a 64-char hex secret (and tells you to put it in `BUZZ_PRIVATE_KEY`), but the desktop app's **Use an existing key** field only accepts bech32 `nsec1...` and reports the hex secret as invalid. Nothing in either surface mentions the conversion, and no converter ships in the tree.

## Steps to reproduce

1. Bootstrap an identity the documented way: `buzz-admin generate-key`
2. Note the printed `Secret key:` (64 hex chars) and the hint `Set BUZZ_PRIVATE_KEY to the secret key.`
3. Launch the desktop app, choose **Use an existing key**, paste that secret

## Actual

The field stays in its pending state (`Waiting for a valid nsec1 key`) and **Next** never enables. The same secret works fine as `BUZZ_PRIVATE_KEY` for `buzz-cli`, so the operator has a key that is simultaneously valid (CLI, env) and invalid (GUI).

## Expected

Any one of:

- desktop accepts 64-char hex and normalizes it internally, or
- `buzz-admin generate-key` also prints the `nsec1` form (ideal — it is the only bootstrap path documented for a self-hosted relay), or
- the field's helper text names the expected encoding and points at a conversion path

## Why it bites self-hosters specifically

On a self-hosted relay, `buzz-admin generate-key` is how you mint the relay owner and every agent identity, and `buzz-admin add-member` takes the hex pubkey — so hex is the native currency of the whole server-side flow. The desktop app is then the one surface that will not take it. Hosted users never see this because their identity is created for them.

Happy to send a PR that adds an `nsec`/`npub` line to `generate-key` output if that is the direction you would want.

Chessing234:
Fixed in #2824 — `generate-key` prints nsec/npub, and desktop import accepts hex.

Chessing234:
opened https://github.com/block/buzz/pull/4504 — desktop accepts hex; generate-key also prints nsec.

Issue #2814, opened 2026-07-25
Fresh `deploy/compose` stack: relay crash-loops on `/data/git/.pack-cache` (root-owned named volume vs non-root uid 1000)


ryanmitchellneely:
## Summary

A first-run of the production Compose bundle in `deploy/compose/` never reaches a healthy relay. The `buzz-git-data` named volume is created root-owned, but the relay container runs as non-root (`buzz:buzz`, uid 1000, per `Dockerfile:153`), so `BUZZ_GIT_PACK_CACHE_PATH` cannot be created and the container restart-loops.

## Environment

- Relay image: `ghcr.io/block/buzz:main` (digest-pinned), Linux x86_64 VPS
- Docker 29.4.1, Compose v5.1.3
- `deploy/compose/compose.yml` + a small local override (port binding + memory limits only; no changes to volumes or user)
- Fresh host, no pre-existing volumes

## Steps to reproduce

1. Copy `deploy/compose/`, fill `.env` from `.env.example` (all `CHANGE_ME` replaced)
2. `docker compose up -d`
3. `docker compose ps` / `docker compose logs relay`

## Actual

Postgres, Redis and MinIO go healthy; the relay never does:

```
ERROR Invalid configuration: invalid config: BUZZ_GIT_PACK_CACHE_PATH=/data/git/.pack-cache could not be created: Permission denied (os error 13)
Error: Configuration error: invalid config: BUZZ_GIT_PACK_CACHE_PATH=/data/git/.pack-cache could not be created: Permission denied (os error 13)
```

repeating on the restart policy indefinitely. `/_readiness` is unreachable for the entire time.

## Expected

Either the relay creates its own cache directory inside the volume it was given, or the bundle prepares ownership so the non-root runtime user can.

## Workaround (confirmed)

```
docker compose down
chown -R 1000:1000 "$(docker volume inspect <project>_buzz-git-data --format '{{.Mountpoint}}')"
docker compose up -d
```

Relay then starts cleanly, passes the A3 git object-store conformance probe, and `/_readiness` returns 200.

## Suggested fix

Any of: an init step in the bundle that chowns the volume (the pattern already used for `minio-init`), a `user:` / entrypoint chown in the relay service, or — cheapest — a documented prerequisite in `deploy/compose/README.md`. Happy to send a PR for whichever you prefer.

## Notes

Not reproducible with the root-level `docker-compose.yml` dev stack, which is presumably why it survives: the failure is specific to the production bundle on a clean host.

[...]

PR #2824, opened 2026-07-25
fix: self-host compose ownership, key import, owner onboarding

Chessing234:
## Summary
- Compose: chown `buzz-git-data` to uid 1000 so the non-root relay can write pack-cache (#2814).
- `buzz-admin generate-key`: print nsec/npub for desktop paste (#2815 admin half).
- Owner onboarding: offer self-hosted relay URL path alongside Builderlab (#2816).

Hex secret import for desktop lives in #4504 (dropped from this tip on rebase to avoid duplicating it).

## Test plan
- [ ] Fresh compose volume: relay starts without pack-cache permission loop
- [ ] `buzz-admin generate-key` prints nsec1… / npub1…
- [ ] Owner onboarding can enter a self-hosted relay URL


BillChirico:
Can we have this merged to fix this issue?

Chessing234:

Chessing234:
hex import is split out as a focused pr: https://github.com/block/buzz/pull/new — wait, will link the number. this selfhost bundle stays conflicting; happy to rebase the compose bits separately if useful.

Chessing234:
hex import is split out as https://github.com/block/buzz/pull/4504 so it can land without the conflicting compose bundle.

Chessing234:
also closed #4244 in favor of #4348; hex import lives in #4504.

Chessing234:
self-host compose ownership + key import path. @MajorTal you've been in the desktop/git weeds lately if this is useful overlap.

Chessing234:
rebased onto main and split hex import into #4504 — still waiting on maintainer review before it can merge.

ryanmitchellneely (approved):
## Review (issue author of #2816)

Thanks @Chessing234 — this unblocks the exact dead-end I hit.

### #2816 owner onboarding — LGTM

`WelcomeSetup.tsx` does the right product split:

- **I own the community** no longer jumps straight to Builderlab
- New **owner** page offers:
  - **Connect a self-hosted relay** → existing member/relay URL flow
  - **Sign in with Builderlab** → hosted path
- `memberBackPage` keeps Back correct from both entry points (member *and* self-host owner)

That matches the expected behavior in #2816 and the provider-neutral framing others confirmed on current `main` (role ≠ hosting).

### Related fixes in this PR — also good

- **`buzz-admin generate-key`**: prints hex **and** npub/nsec so desktop paste works (#2815 half). Clear operator copy.
- **compose `git-data-init`**: chown `buzz-git-data` to uid 1000 before relay start — correct fix for fresh named volumes + non-root `USER buzz` pack-cache loop (#2814).

### Non-blocking notes

- Hex secret import for desktop is correctly deferred to #4504 per PR body — fine; just keep that linked so self-hosters don’t think this PR alone finishes key import.
- No Playwright/onboarding regression in the diff; morthenjensen-wq offered a focused test on #2816 — nice follow-up but not a merge blocker for this UX fix.
- Self-host Mac still needs the known CORS footgun (`BUZZ_CORS_ORIGINS=tauri://localhost`) until that’s documented/defaulted elsewhere — out of scope here.

### Verdict

**Please merge.** This is the first-five-minutes path for the OSS/self-host audience the repo targets. Happy to re-verify on a desktop build once it lands.

— Ryan (reporter of #2816)

Glucksberg:
Confirmed — I just hit this exact onboarding path while setting up Buzz on a second machine with the same identity.

My existing community was already working on Windows, including another member who had successfully joined it. When setting up the Mac with the same nsec, I chose:

"I already have a community" → "I own the community"

That sent me directly into the Builderlab sign-in flow, even though I wasn't trying to create or manage a hosted community.

The workaround was basically to "go through the gap in the fence":

1. Go back
2. Choose "Join a community" instead — even though I'm actually the owner
3. Use the same invite link from my existing community
4. Buzz connects normally to the existing community with the same identity

So #2824 matches the actual user experience very well. The important UX distinction really is ownership vs. hosting: owning a community shouldn't imply that it is Builderlab-hosted.

The proposed split between Connect a self-hosted relay and Sign in with Builderlab would have made the correct path immediately obvious.

PR #2827, opened 2026-07-25
Add Buzz Agent to harness selection page, and OpenAI-compatible base URL field to model settings during onboarding

yashrajd:
This PR bundles two related desktop agent-config fixes:

**Show the bundled Buzz Agent during onboarding**

Adds buzz-agent to ONBOARDING_RUNTIME_ORDER so the bundled harness appears alongside Claude and Codex in the setup list.
Extends sidecar discovery to also look for binaries with the host target-triple suffix (e.g. buzz-agent-aarch64-apple-darwin). This lets local dev builds resolve the bundled sidecar; release DMG builds continue to use the plain buzz-agent name.

_Rationale: The 0.4.25 release build did not list Buzz Agent in onboarding because the allowlist only included Claude and Codex._

**Add an OpenAI-compatible base URL field**
When the Buzz Agent provider is set to openai-compat, the config UI now shows a dedicated OpenAI-compatible base URL input.
The value is stored as the OPENAI_COMPAT_BASE_URL env var, validated as a URL, and hidden from the generic Advanced env editor so it does not appear twice.
Saves are blocked until the base URL is valid when the provider requires it.

_Rationale: The OpenAI-compatible provider also had no structured way to set its base URL, forcing users to edit raw env vars._

### DIFF
--- a/desktop/src-tauri/src/managed_agents/discovery.rs
+++ b/desktop/src-tauri/src/managed_agents/discovery.rs
@@ -398,6 +398,12 @@ fn command_search_dirs() -> Vec<PathBuf> {
             .ok()
             .and_then(|path| path.parent().map(Path::to_path_buf)),
     );
+
+    // Tauri-bundled sidecars are staged here by scripts/bundle-sidecars.sh.
+    // Include the directory so local dev builds can resolve them even though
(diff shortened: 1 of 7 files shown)


yashrajd:
finally-buzz-harness-openai-compat-base-url-works copy

AdnanAbass (approved):

PR #2854, opened 2026-07-25
fix(desktop): keep identity key help dialog readable in dark mode

pfista:
## Problem

With macOS in dark mode (the default for fresh profiles is to follow the
system scheme), the onboarding "What's an identity key?" help dialog renders
its title in near-white on the always-white textured card, making it
unreadable. The body paragraphs stay readable because they use the fixed
olive `--buzz-onboarding-backup-ink`; only the `text-foreground` title (and
the close button's hover color) flip with the theme.

## Cause

The dialog's `DialogContent` carries `buzz-onboarding-neutral-theme` but is
portaled outside the `buzz-startup-shell` subtree, so in dark mode it matches
`.dark .buzz-onboarding-neutral-theme:not(.buzz-startup-shell)`
(`components.css`), which flips `--foreground` to `0 0% 98%`. The textured
powder card (`buzz-card-textured`) has no dark variant — it is baked light —
so the near-white title disappears against it.

## Fix

One attribute: pin the dialog to the light neutral theme with
`data-system-color-scheme="light"`. This is the established pattern for
always-light onboarding dialogs (`HostedCommunityOnboarding.tsx`, the
`CommunityOnboardingFlow.tsx` avatar dialog), and the pinned-light CSS rule
already exists and out-specifies the dark-mode flip. No new CSS.

## Testing

- Added a dark-mode regression test to `tests/e2e/identity-key-help.spec.ts`
  (already registered in the Playwright smoke project): emulates
  `prefers-color-scheme: dark`, opens the dialog, and asserts the title
  resolves to the pinned light-neutral ink `rgb(23, 23, 23)`. Before the fix
  it rendered `rgb(250, 250, 250)`.
- Manual repro: macOS appearance set to Dark → fresh profile → machine
  onboarding → click "What's an identity key?".

Before/after screenshots are in the comment below.

### DIFF
--- a/desktop/src/features/onboarding/ui/IdentityKeyHelpDialog.tsx
+++ b/desktop/src/features/onboarding/ui/IdentityKeyHelpDialog.tsx
@@ -67,6 +67,7 @@ export function IdentityKeyHelpDialog() {
       <DialogContent
         className="buzz-onboarding-neutral-theme max-w-[47.5rem] -translate-y-5"
         closeButtonClassName={ONBOARDING_INK_ICON_CLASS}
+        data-system-color-scheme="light"
         data-testid="identity-key-help-dialog"
         overlayVariant="transparent"
         surface="textured"
--- a/desktop/tests/e2e/identity-key-help.spec.ts
+++ b/desktop/tests/e2e/identity-key-help.spec.ts
@@ -56,3 +56,38 @@ test("identity key help explains the first-run choice", async ({ page }) => {
     "1",
   );
 });
+
+test("identity key help stays readable when the app resolves dark mode", async ({
+  page,
+}) => {
+  await page.emulateMedia({ colorScheme: "dark" });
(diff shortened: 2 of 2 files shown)


pfista:
### Before — dark mode

The dialog title is near-white on the always-light textured card.

01-before-dark

### After — dark mode

The title keeps the pinned light-neutral ink.

02-after-dark

### After — light mode (unchanged)

03-after-light

wesbillman (approved):
Reviewed the theme cascade and regression coverage. This is the right narrowly scoped fix: the textured card is an always-light surface, the attribute is applied to the actual portaled `DialogContent`, and the existing explicit-light selector overrides the ambient dark rule. The test reproduces system dark mode and verifies the computed foreground value that was broken before.

No blocking findings. CI had only DCO/security checks available at review time; desktop checks had not started yet.

pfista:
Thanks @wesbillman good to merge?

PR #2862, opened 2026-07-25
fix(desktop): fetch join policies through native networking

johnmatthewtennant:
## Context

Adding an existing community by relay URL could fail with `Community rejected: Load failed` even when its WebSocket endpoint was reachable. The Add Community flow fetched `/api/join-policy` from the WebView, so a relay without a matching CORS allowance blocked the policy request before the app could join it.

## Summary

This bug fix fetches join policies through Tauri's native networking layer for direct URL joins. Invite-code discovery, policy acceptance, and signed invite claims remain on the WebView path so those operations can migrate together later.

## Changes

- Uses native networking for Add Community and first-community direct URL join-policy requests.
- Validates relay schemes, rejects URLs containing credentials, and refuses redirects.
- Bounds declared and chunked native responses before JSON parsing.
- Preserves existing `404`, non-success status, malformed JSON, and absent-policy behavior.
- Requires every join-policy caller to choose its transport explicitly.

Public relays using Buzz's default permissive CORS configuration are not known to be affected.

### Related issue

Related to #2872.

#### Reviewer-reproducible examples

End-to-end red/green requires a relay with restrictive CORS and a Buzz identity authorized to join it.

##### Red: `main`

From a clean checkout of `main`:

```bash
. ./bin/activate-hermit
just staging
```

In Buzz Desktop:

1. Add another community so the restrictive-CORS relay can be removed.
2. Remove that relay.
3. Open Add Community and enter the relay's WebSocket URL.
4. Select Add Community.

Observed result:

```text
Community rejected: Load failed
```

##### Green: this PR

From a clean checkout of this branch:

```bash
. ./bin/activate-hermit
just staging
```

Repeat the same steps above.

Observed result:

```text
The community rejoins successfully.
```

Supporting checks:

- Six native join-policy tests, including oversized declared and chunked responses.
- Four TypeScript API tests, including the native command contract.
- E2E build and four focused onboarding and sidebar Playwright tests.
- Full `just ci` and pre-push suites.
- Builderbot, Kalvin, and minimize-diff review fanout found no actionable issues after the final rebase.


bp2m6v4sp5-dev (commented):
Submit

desktop/src-tauri/src/commands/join_policy.rs
```diff
@@ -0,0 +1,127 @@
...
+        return Ok(None);
+    }
+    if !response.status().is_success() {
+        return Err(format!("HTTP {}", response.status().as_u16()));
+    }
+
+    let body = response
+        .json::<Value>()
+        .await
+        .map_err(|_| "relay returned malformed join policy".to_string())?;
```
wpfleger96: 🤖 This is the one I'd most want fixed. `response.json()` buffers the entire body before deserializing, and this body comes from an arbitrary relay URL the user typed in — a new native trust boundary, since the old webview `fetch` never handed unbounded attacker bytes to `serde_json` inside the Rust process.

To be fair to the current code: `.timeout()` in reqwest 0.13 is a *total* timeout and it does wrap the response body (`Response::new` passes it into `body::response`, which wraps in `TotalTimeoutBody`), so this isn't an unbounded-forever read. But it's a bandwidth × 15s bound, not a byte bound — a relay on a fast link can push a lot of memory into the app process in 15 seconds, and `serde_json::Value` on top of that isn't cheap either.

The adjacent native fetch already does this right — `link_preview.rs` has `MAX_TITLE_FETCH_BYTES` and streams with `bytes_stream()` accumulating against the cap. Same shape here: cap the read, reject an oversized declared `Content-Length` up front, then `serde_json::from_slice` on the bounded bytes. For the cap value, the relay allows two 256 KiB policy documents (`MAX_POLICY_MARKDOWN_BYTES` in `crates/buzz-relay/src/config.rs`), so ~512 KiB plus JSON overhead is a documented, generous ceiling.

Worth tests for both an oversized `Content-Length` and a chunked/undeclared oversize body, since only the second one survives a header check.

desktop/src/shared/api/invites.ts
```diff
@@ -124,26 +128,38 @@ export function isJoinPolicyDiscoveryCandidate(relayWsUrl: string): boolean {
 /** Fetch relay-hosted policy content for any join surface. */
 export async function getJoinPolicy(
   relayWsUrl: string,
+  transport: "native" | "webview" = "webview",
```
wpfleger96: 🤖 I'd drop the default and make `transport` required. The default here is `"webview"`, which is exactly the path this PR exists to work around — so any future call site that forgets the argument silently gets the CORS-breaking transport, and it'll fail the same way this bug did, on the same class of relay.

All five call sites already pass it explicitly except `InviteRedeemForm.tsx:208`, and making that one spell out `"webview"` is a bonus: the deliberate invite-code exception becomes self-documenting at the call site instead of being an invisible default.

desktop/tests/e2e/sidebar.spec.ts
```diff
@@ -111,24 +111,15 @@ test("add community starts with create and join choices", async ({ page }) => {
...
-        }),
-      });
+  await installMockBridge(page, {
+    applyCommunityDelayMs: 1_000,
+    joinPolicy: {
+      terms_markdown: "# Terms",
+      privacy_markdown: "# Privacy",
+      age_attestation_required: true,
+      version: "policy-v1",
     },
```
wpfleger96, line 121: 🤖 Small test-hygiene thing: converting this spec to `joinPolicy` on the mock bridge is right, but it leaves a few `page.route(".../api/join-policy")` interceptors behind on flows that are now native and never hit them — `sidebar.spec.ts:174-179` (`token.example.com`), and `onboarding.spec.ts:1234-1239` and `1292-1297` (`onboarding.communities.buzz.xyz`). Those three now pass because the mock bridge's `joinPolicy` defaults to `null`, not because the route returned 404, so they read as asserting something they don't.

Worth deleting them (or passing `joinPolicy: null` explicitly) so each spec says what it actually exercises. `onboarding.spec.ts:3291` is still live — that one's the invite-code path.

desktop/src/features/communities/ui/CommunityEditForm.tsx
```diff
@@ -68,7 +68,7 @@ export function CommunityEditForm({
 
     let cancelled = false;
     const timeoutId = window.setTimeout(() => {
-      void getJoinPolicy(normalizedUrl)
+      void getJoinPolicy(normalizedUrl, "native")
```
wpfleger96, line 71: 🤖 Not introduced by this PR, but it makes the description a bit optimistic: the summary says joins "and edits" now go native, and both `CommunityEditForm` call sites are gated on `joinPolicyRequired`, which no caller ever passes — `CommunityChangeOverlay` is the only place that renders this form and it omits the prop, so it defaults to `false`. So these two lines are unreachable today and edits never fetch a policy at all.

Fine to leave the code as-is (it's correct for whenever the prop does get wired up), but I'd soften the "and edits" claim in the description so nobody reads this as an edit-flow fix.

desktop/src/features/onboarding/ui/InviteRedeemForm.tsx
```diff
@@ -118,7 +118,7 @@ export function InviteRedeemForm({
     const code = parsedInvite?.code;
     let cancelled = false;
     const timeoutId = window.setTimeout(() => {
-      void getJoinPolicy(relayWsUrl)
+      void getJoinPolicy(relayWsUrl, code ? "webview" : "native")
```
wpfleger96, line 121: 🤖 Just confirming I read this the way you intended: `code` is always set when `parsedInvite` is non-null (both `ParsedInvite` variants carry a `code`), and `relayWsUrl` only comes from `normalizedRelayUrl` when `parsedInvite` is null — so this ternary is exactly "invite present → webview, plain relay URL → native", matching the submit paths at 159 and 208. No change needed; noting it because it took me a minute to convince myself the background-discovery transport can't diverge from the transport the eventual submit uses.

wpfleger96 (changes requested):
🤖 hey @johnmatthewtennant — I like the direction here, native networking is clearly the right escape hatch for a CORS-restricted relay, and the redirect refusal + credential rejection are nice hardening on top. Two things I'd want changed before this goes in, plus a few smaller notes inline.

Blocking, both left as inline comments:

1. the native command buffers the whole response body from an arbitrary user-entered relay (`response.json()`)
2. `transport` defaults to `"webview"` — the broken path is the default a new caller gets

Non-blocking notes:

`desktop/src/shared/api/invites.test.mjs` still only exercises the webview branch (every `getJoinPolicy` call in it omits the transport arg). Worth one native-transport case that pins the command name `fetch_join_policy`, the `relayUrl` arg shape, and the mapping — the two branches aren't symmetric (webview unwraps `raw.policy`, native returns the already-unwrapped policy), and right now nothing on the TS side locks that down.

The reported bug still reproduces for invite-URL joins against the same relay, since `getJoinPolicy` on the invite path, `acceptJoinPolicy`, and `claimInvite` all stay on webview `fetch`. I get why (the description says as much) — could you file a follow-up issue and link it here so the remaining half doesn't get lost?

Nothing here looks wrong to me on the parts I didn't comment on: `404`/non-success/malformed-JSON semantics match the old flow, and I ran the URL builder against `..` traversal, `%2F`-encoded segments, IPv6 hosts, uppercase schemes, and bare `ws:///foo` — all either rejected or normalized sanely.

wpfleger96 (approved):
🤖 Rechecked the review fixes at exact head `29b65c86e5a3013641be51277d812081a64c7422`. The native response now has declared-length and streamed byte caps with both oversize cases tested; `transport` is required and explicit at every caller; the native TypeScript contract test is in place; stale route mocks are gone; the description is scoped to direct URL joins; and the remaining invite-path CORS work is linked as #2872. I don’t see any remaining blocker.

PR #2863, opened 2026-07-25
fix(mobile): tag DM recipients on send

### LINKED ISSUE #2835: Mobile: DM omits recipient p-tag unless a mention is picked, so agents are never woken
## Summary
A direct message composed on Buzz **mobile** carries **no `p` tag for the DM counterparty** unless the sender explicitly selects the recipient from the mention picker. Buzz **desktop** attaches the counterparty `p` tag automatically on the same DM. Agent harnesses (`buzz-acp`) wake a session only on `p`-tagged events, so a mobile-composed DM without a picked mention is delivered to the relay but never wakes the recipient agent. The message lands silently and gets no reply.

## Impact
Agents are effectively unreachable from mobile for anyone who types a normal message. The failure is silent: the DM appears sent, shows in history, and the recipient agent simply never responds, with no signal to the sender that anything went wrong. The same DM works on desktop, which makes it read as intermittent rather than a clean mobile gap.

## Environment
- Buzz mobile client, first-day availability 2026-07-25.
- Buzz desktop client, same account, same DM channel.
- Recipient is a managed agent running the `buzz-acp` harness, which wakes on `p`-tagged events.

## Reproduction (natural experiment, real events)
All six events below are in one 1:1 DM between a human sender and an agent recipient, same channel, same minute range.

| source | content | recipient `p` tag | agent woke? |
|---|---|---|---|
| desktop, plain | "from desktop..." | present | yes |
| desktop, plain | "Delivery of DM from mobile..." | present | yes |
| mobile, plain text | "can you read this from mobile" | NONE | no |
| mobile, name typed as raw text (not picked) | "@boss write up for buzz team" | NONE | no |
| mobile, name selected from mention picker | "@Boss from mobile" | present | yes |
| mobile, name selected from mention picker | "@Boss write up for buzz team" | present | yes |

Steps:
1. From desktop, open a DM with an agent and send a message. The event carries `p = <agent pubkey>` and the agent replies.
2. From mobile, in the same DM, send a plain message with no mention. `p` is absent, the agent never replies.
3. From mobile, type the agent name as raw text (for example `@boss`) without selecting from the picker. `p` is still absent, the agent never replies.
4. From mobile, type `@` and **select the agent from the mention list**. The event now carries `p = <agent pubkey>` and the agent replies.

## Expected
A DM composed on mobile should attach the counterparty `p` tag automatically, the same way desktop does, so the recipient is notified and (for agents) woken regardless of whether the sender explicitly mentions them. In a 1:1 DM the recipient is unambiguous, so the tag should not depend on the sender remembering to pick a mention.

## Actual
On mobile, the recipient `p` tag is attached only when the sender resolves a mention through the picker. Plain text, and the recipient's name typed as raw text without picking it, both publish with no `p` tag. Desktop attaches the tag with no mention required.

## Workaround (confirmed working)
On mobile, type `@`, select the recipient from the mention list, then send. Typing the name as plain text is not enough: it must be picked so it resolves to a tagged entity.

## Suggested fix
On mobile DM send, populate the `p` tag with the DM counterparty by default, matching desktop behaviour. The recipient in a 1:1 DM is known at send time, so the tag can be added unconditionally rather than relying on an explicit mention.

Baltsat:
## Summary

Mobile DM sends only serialized `p` tags from picker-resolved mentions, so a plain DM could reach the relay without addressing the counterparty. This adds the known DM participants to the existing mention-normalization path.

## Changes

- append DM participant pubkeys after explicit mentions
- reuse the existing case-insensitive dedupe and self-exclusion logic
- pass DM participants from both the channel composer and thread composer
- preserve existing tag order: `h`, thread `e` tags, recipient `p` tags, then media tags
- add focused coverage for plain DM sends, mixed-case dedupe/self exclusion, non-DM behavior, and threaded media ordering

## Verification

- `dart format --output=none --set-exit-if-changed` on all four changed files
- `git diff --check`
- Flutter analysis and tests were not available in the local environment; GitHub Actions is the integration gate

AI assistance: ChatGPT/Codex was used for implementation. The generated patch was independently reviewed, corrected, and reduced to one product commit before publication.

Closes #2835

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page.dart
+++ b/mobile/lib/features/channels/channel_detail_page.dart
@@ -326,6 +326,9 @@ class ChannelDetailPage extends HookConsumerWidget {
                         channelId: channel.id,
                         content: content,
                         mentionPubkeys: mentionPubkeys,
+                        dmParticipantPubkeys: resolvedChannel.isDm
+                            ? resolvedChannel.participantPubkeys
+                            : null,
                         mediaTags: mediaTags,
                       ),
             )
--- a/mobile/lib/features/channels/send_message_provider.dart
+++ b/mobile/lib/features/channels/send_message_provider.dart
@@ -27,19 +27,25 @@ class SendMessage {
   /// If [rootEventId] is null it defaults to [parentEventId] (direct reply to
   /// thread head). Tags are built to match the desktop's `buildReplyTags`
   /// convention with `root` / `reply` markers. Pass [mediaTags] to append
-  /// relay-validated `imeta` tags and NIP-30 `emoji` tags.
+  /// relay-validated `imeta` tags and NIP-30 `emoji` tags. For a DM, pass
+  /// [dmParticipantPubkeys] so every counterparty receives a `p` tag.
   Future<void> call({
     required String channelId,
     required String content,
     String? parentEventId,
     String? rootEventId,
     List<String>? mentionPubkeys,
+    List<String>? dmParticipantPubkeys,
     List<List<String>> mediaTags = const [],
   }) async {
     // Use explicitly passed pubkeys, or resolve @mentions against
     // channel members to avoid matching the wrong user.
     final resolvedMentions =
         mentionPubkeys ?? await _resolveMentions(content, channelId);
+    final allMentionPubkeys = <String>[
+      ...resolvedMentions,
+      ...?dmParticipantPubkeys,
+    ];
     final authorPubkey = _signedEventRelay.pubkey;
 
     // Normalize mentions: lowercase, deduplicate, exclude self (matching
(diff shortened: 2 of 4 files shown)


svndco:
## Independent review

I checked this out at `af2216fbf6e04b7b3b8e182502789cc0b0abe8d9` and ran the full mobile package gates:

- `flutter test`: **589 passed, 1 skipped**
- `just mobile-check`: formatting, analysis, and file-size checks passed

The send-path wiring covers both the channel composer and DM thread composer, and the regression tests correctly cover plain DMs, self exclusion/deduplication, non-DM behavior, and threaded tag ordering.

One consistency fix is worth folding in before merge: `normalizedMentions` deduplicates using `pk.toLowerCase()` but emits the original `pk`, despite the comment saying this matches desktop normalization. Desktop's `normalizeMentionPubkeys` emits the lowercase value, and this path also retains an empty pubkey if supplied. PR #2927 has the tighter loop (`lower.isNotEmpty` and emit `lower`). I recommend adopting that small normalization piece here, retaining this PR's stronger threaded/media-tag tests, and then closing #2927 as the duplicate.

Baltsat:
Closing this in favor of #3258, which now covers the same recipient-tag behavior at the central send path with dedicated wire tests. Keeping both implementations open would duplicate review effort. Thanks for the review and guidance here.

Issue #2853, opened 2026-07-25
feat(search): query operators `from:` / `in:` / `before:` / `after:` (backend already supports them, client never sends them)


oscarjpicazo:
## Problem

Search accepts free text only. There is no way to narrow a query to a person,
a channel, or a date range. The `from:` / `in:` / `before:` / `after:`
operators are muscle memory for anyone coming from Slack, and a Slack-parity
gap in the sense of #2728. Today the only narrowing that exists is
programmatic: the in-channel find bar scopes by the current channel.

## What already exists (the surprising part)

The entire backend path is already built. `SearchQuery` in
`crates/buzz-search/src/query.rs` carries `authors`, `since`, `until`, `kinds`
and a channel scope, and **both** relay search paths already map those fields
from the incoming NIP-01 filter:

- WS NIP-50 path: `crates/buzz-relay/src/handlers/req.rs` (~L599)
- bridge path: `crates/buzz-relay/src/api/bridge.rs` (~L1698)

So a filter like `{"search": "deploy", "authors": [<pubkey>], "since": …,
"until": …, "#h": [<channel>]}` filters correctly on the relay **today**. The
existing per-hit re-authorization is untouched: the relay refetches and
re-authorizes every hit, so operators cannot widen visibility ("search is
never the access boundary").

What's missing is entirely client-side: the desktop sends only
`{search, search_mode, limit, "#h"}` (`build_search_messages_filter` in
`desktop/src-tauri/src/commands/messages.rs`), and nothing anywhere parses
operator syntax.

## Proposed v1 scope

Client-only. No new event kinds, no relay changes, no schema changes.

**Syntax** (parsed out of the search input; remaining text is the FTS query):

| Operator | Maps to | Notes |
|---|---|---|
| `from:@name` | `authors` | resolved via the existing user-search / mention-candidate infra; autocomplete required, so ambiguous names are resolved at compose time |
| `in:#channel` | `"#h"` tag | resolved against the local channel list (same source `useSearchResults.ts` already filters) |
| `after:YYYY-MM-DD` | `since` | start of day, local time |
| `before:YYYY-MM-DD` | `until` | start of day, local time |

**Implementation sketch:**

1. Pure-TS operator parser in `desktop/src/features/search/lib/` (unit-tested;
   invalid operator text simply stays in the FTS query).
2. Autocomplete for `from:` / `in:` in the topbar search, reusing the existing
   user/channel pickers; active operators render as removable filter chips.
[...]

PR #2871, opened 2026-07-25
feat(search): parse from:/in:/after:/before: and pass them in the filter

jatinder14:
## Summary
- Adds a pure-TS operator parser (`from:` / `in:` / `after:` / `before:`) with unit tests. Invalid date tokens stay in the FTS text.
- Extends `search_messages` so the desktop can send `authors` / `since` / `until` (and existing `#h`) on the filter the relay already understands.
- Wires topbar search to strip operators from the prefix query, resolve `in:` against local channels and `from:` against hex pubkeys / known agents, then pass the structured fields through.

This is part 1 of #2853 (parser + command plumbing). Autocomplete chips / richer `from:@name` resolution can follow in a second PR.

## Test plan
- [x] `node --import ./test-loader.mjs --experimental-strip-types --test src/features/search/lib/parseSearchOperators.test.mjs`
- [x] Added `search_messages_filter_emits_operator_fields` unit test (full `buzz-desktop` crate build needs local sidecar binaries in this environment)
- [ ] Manual: topbar `deploy from:<hex> after:2024-01-01` emits authors/since on the bridge filter and returns narrowed hits


wpfleger96 (changes requested):
The required formatting gates fail on this head:

- `pnpm check` reports Biome formatting changes in `desktop/src/features/search/lib/parseSearchOperators.test.mjs` and `desktop/src/features/search/useSearchResults.ts`.
- `just desktop-tauri-fmt-check` reports rustfmt changes in `desktop/src-tauri/src/commands/messages_tests.rs`.

Please run `just desktop-fix` and `just desktop-tauri-fmt`, commit the formatter output, and rerun the gates.

I found no substantive parser, filter-plumbing, or access-control defect. At `e60e1c29f94931c29f109b750d95f3884f33e79b`, `pnpm test` (3,509 tests), `pnpm typecheck`, `pnpm build`, Desktop Tauri `cargo test` (1,630 passed, 14 ignored, plus 3 diagnostic tests), and Desktop Tauri `cargo check` pass; the desktop smoke/integration CI jobs pass as well. Requesting changes solely because the repository-required formatting checks remain red.

desktop/src/features/search/lib/parseSearchOperators.ts
```diff
@@ -0,0 +1,134 @@
...
+  from: string | null;
+  /** Raw `in:` value (channel uuid, #name, …), if present. */
+  in: string | null;
+  /** `after:YYYY-MM-DD` → unix seconds at local start of day. */
+  since: number | null;
+  /** `before:YYYY-MM-DD` → unix seconds at local start of day. */
+  until: number | null;
+};
+
+const OPERATOR_RE = /\b(from|in|after|before):(\S+)/gi;
```
wpfleger96: **Must fix — `\b` lets hyphen/slash-adjacent tokens parse as operators.**

`\b` matches after `-` and `/`, so ordinary text becomes an operator. Ran the parser standalone at this head:

- `built-in:react hooks` -> `text: "built- hooks"`, `in: "react"`
- `sign-in:flow broken` -> `text: "sign- broken"`, `in: "flow"`
- `opt-in:true` -> `text: "opt-"`, `in: "true"`
- `https://x.com/in:foo` -> `text: "https://x.com/"`, `in: "foo"`

So searching `built-in` mangles the query text *and* injects a bogus `in:` scope, which then hits the silent-unfiltered-fallback problem in `useSearchResults`. Anchoring to a token start fixes it:

```ts
const OPERATOR_RE = /(?:^|\s)(from|in|after|before):(\S+)/gi;
```

desktop/src/features/search/lib/parseSearchOperators.ts
```diff
@@ -0,0 +1,134 @@
...
+  /** FTS / prefix query with operators removed. */
+  text: string;
+  /** Raw `from:` value (pubkey hex, npub, @name, …), if present. */
+  from: string | null;
+  /** Raw `in:` value (channel uuid, #name, …), if present. */
+  in: string | null;
+  /** `after:YYYY-MM-DD` → unix seconds at local start of day. */
+  since: number | null;
+  /** `before:YYYY-MM-DD` → unix seconds at local start of day. */
+  until: number | null;
```
wpfleger96, line 33: Nit: `after:` ends up inclusive of its day (`since` = local start of day) while `before:` is exclusive (`until` = local start of day, so messages *on* that date are dropped). Slack does the same thing so it's probably intended — but both doc lines describe the value identically, so the asymmetry is invisible to the next reader. One clause each would fix it.

desktop/src/features/search/useSearchResults.ts
```diff
@@ -34,6 +41,49 @@ function dedupeSearchHits(hits: SearchHit[]) {
...
+  }
+  const handle = normalizeFromHandle(raw).toLowerCase();
+  if (!handle) {
+    return null;
+  }
+  const match = candidates.find((candidate) => {
+    const name = candidate.displayName?.trim().toLowerCase();
+    return name === handle || normalizePubkey(candidate.pubkey) === handle;
+  });
+  return match ? normalizePubkey(match.pubkey) : null;
```
wpfleger96: **Must fix — unresolved operator silently degrades to an unfiltered search.**

Returning `null` here is indistinguishable from "no `from:` operator was given" by the time it reaches line 160 (`authors: resolvedAuthor ? [resolvedAuthor] : undefined`). So `deploy from:@nobody` searches **every** author instead of narrowing — the user asked to restrict and got a wider result set with no signal.

`resolveChannelIdFromOperator` has the identical shape: `deploy in:nosuchchannel` searches every channel. Punctuation makes it easy to trip by accident — I confirmed `in:general,` parses to `in === "general,"`, which matches no channel name and silently un-scopes.

Suggest returning a discriminated result (`{ kind: "none" } | { kind: "resolved", value } | { kind: "unresolved" }`) and, on `unresolved`, either returning zero hits or surfacing it in the dropdown. Never fall through to the broad query.

desktop/src/features/search/useSearchResults.ts
```diff
@@ -55,21 +105,73 @@ export function useSearchResults({
...
+    const candidates: Array<{ pubkey: string; displayName?: string | null }> =
+      [];
+    for (const agent of managedAgentsQuery.data ?? []) {
+      candidates.push({ pubkey: agent.pubkey, displayName: agent.name });
+    }
+    for (const agent of relayAgentsQuery.data ?? []) {
+      candidates.push({ pubkey: agent.pubkey, displayName: agent.name });
+    }
+    return candidates;
+  }, [managedAgentsQuery.data, relayAgentsQuery.data]);
```
wpfleger96: **Should fix — this seed is agents-only, so `from:@name` never resolves a human.**

Only `managedAgentsQuery` and `relayAgentsQuery` feed the candidate list; `userSearchQuery.data` is not included. `from:@Will Pfleger` therefore resolves to `null` and (per the comment on line 84) silently runs unfiltered.

The description scopes richer `from:@name` resolution to a follow-up, which is fair — but as shipped the operator looks like it supports people and quietly doesn't. Either fold `userSearchQuery.data` into the seed, or make unresolved-operator explicit so the failure is visible instead of silent.

Separately: `parsedQuery.from` only ever captures one whitespace-delimited token, so `from:@Will Pfleger` captures `"@Will"` and leaves `Pfleger` in the FTS text (verified). Worth noting in the description as a known limitation of part 1.

desktop/src/features/search/hooks.ts
```diff
@@ -14,16 +17,36 @@ export function useSearchMessagesQuery(
...
+        until: until ?? undefined,
       }),
-    enabled: enabled && trimmedQuery.length >= 2,
+    enabled:
+      enabled &&
+      (trimmedQuery.length >= 2 ||
+        Boolean(authors?.length) ||
+        since != null ||
+        until != null ||
+        Boolean(channelId)),
```
wpfleger96: **Should fix — this broadened predicate is unreachable from both call sites.**

The intent (operators alone can justify a query without 2+ chars of FTS text) never fires: `useSearchResults.ts:157` gates on `ftsQuery.length >= MIN_SEARCH_QUERY_LENGTH` and `useChannelFind.ts:69` gates on `debouncedQuery.length >= MIN_QUERY_LENGTH`, both before this hook is consulted. Those are the only two callers.

Either drop the extra clauses and keep `trimmedQuery.length >= 2`, or make this hook the single owner of the decision and remove the caller-side floors. Holding both means two thresholds that can silently drift.

desktop/src/testing/e2eBridge.ts
```diff
@@ -7943,6 +7943,10 @@ async function handleSearchMessages(
   args: {
     q: string;
     limit?: number;
+    channelId?: string;
+    authors?: string[];
+    since?: number;
+    until?: number;
```
wpfleger96, line 8041: **Should fix — the bridge accepts the new args but no branch uses them.**

The mock branch filters only on `hit.content` / `hit.channel_name`, and the relay branch builds `{ kinds: [9, 40002], search: args.q, limit }` — `channelId`, `authors`, `since`, and `until` are dropped in both. So this type widening is inert and no Playwright spec can assert that an operator actually narrows results.

That matters because the untested layer is exactly where the two must-fix defects live: the parser has unit coverage (syntax) and `build_search_messages_filter` has unit coverage (filter shape), but the resolution-and-wiring middle has none. Applying `authors`/`since`/`until` to `mockHits` plus one spec driving the topbar with `from:`/`after:` would have caught the silent-unfiltered fallback.

wpfleger96 (changes requested):
🤖 Agent-authored review.

## Paul — independent code review: block/buzz#2871

Reviewed at head `e60e1c29f94931c29f109b750d95f3884f33e79b` in an isolated worktree, full files read (not just hunks), traced the whole path: parser -> `useSearchResults` resolution -> `useSearchMessagesQuery` -> `searchMessages` -> `build_search_messages_filter` -> the relay's `handle_bridge_search` / `handle_search_req` pushdown. I also ran the parser standalone over ~20 adversarial inputs to check tokenization empirically rather than reasoning from the regex.

## Verdict
**Request changes** — two substantive defects in operator handling (silent unfiltered fallback, and hyphen/slash-adjacent mis-tokenization) plus the formatting gates that are already red.

## Strengths
- The pushdown is genuinely free on the relay side: `handle_bridge_search` already maps `filter.authors` / `since` / `until` into `buzz_search::SearchQuery`, and `search_hit_accepted` re-authorizes every hit against `accessible_channels` + `reader_authorized_for_event`. Adding these three fields to the desktop filter widens no access boundary, and the comment at `messages.rs:150-152` says exactly that.
- `build_search_messages_filter` omits keys entirely when the operator is absent, and the existing prefix-mode test was extended with negative assertions (`authors`/`since`/`until` all `is_none()`) — that's the right shape for a filter builder.
- `parseLocalDayStart` validates real calendar dates via round-trip (`2024-02-30` correctly rejected, confirmed by running it) and leaves unparseable date tokens in the FTS text instead of dropping them.

## Must fix

**1. An operator that fails to resolve silently degrades to *no filter at all*, widening the result set.** `resolveAuthorFromOperator` returns `null` when the handle doesn't match a candidate, and `resolveChannelIdFromOperator` returns `null` when the channel name doesn't match. Both nulls become `undefined` at `useSearchResults.ts:159-160`, so the query runs **unscoped**. Concretely: `deploy from:@nobody` searches every author, and `deploy in:nosuchchannel` searches every channel — the user asked to narrow and got the opposite, with no indication anything was ignored. This is aggravated by punctuation: `in:general,` yields `in === "general,"` (verified), which never matches a channel name, so a trailing comma silently un-scopes the search.

Fix: distinguish "no operator" from "operator present but unresolved". When an operator was supplied and resolution failed, either return zero results or surface it in the UI — do not fall through to the unfiltered query.

**2. `\b` in `OPERATOR_RE` makes hyphen- and slash-adjacent tokens parse as operators.** `parseSearchOperators.ts:24` uses `/\b(from|in|after|before):(\S+)/gi`, and `\b` matches after `-` and `/`. Measured outputs:

| input | parsed |
|---|---|
| `built-in:react hooks` | `text: "built- hooks"`, `in: "react"` |
| `sign-in:flow broken` | `text: "sign- broken"`, `in: "flow"` |
| `opt-in:true` | `text: "opt-"`, `in: "true"` |
| `https://x.com/in:foo` | `text: "https://x.com/"`, `in: "foo"` |

So searching for `built-in` or `sign-in` — ordinary English in a dev chat — mangles the query text *and* injects a bogus channel scope that then triggers defect 1. Fix: anchor the operator to a token start, e.g. `/(?:^|\s)(from|in|after|before):(\S+)/gi`.

## Should fix

**3. `from:@name` can only ever resolve agents, never people.** `authorCandidateSeed` (`useSearchResults.ts:137-147`) is built exclusively from `useManagedAgentsQuery` + `useRelayAgentsQuery`. `useUserSearchQuery` results are not in the seed, so `from:@Will Pfleger` resolves to `null` -> unfiltered (defect 1 again). Given the description frames richer `from:@name` resolution as follow-up work, that's a defensible scope line — but combined with the silent fallback it currently reads as "`from:` works for humans" when it doesn't. At minimum this needs the resolution-failure signal from defect 1; folding `userSearchQuery.data` into the seed is the small version of the real fix.

**4. The feature has no behavior-level coverage, and the E2E bridge accepts the new args without using them.** `handleSearchMessages` widens its arg type with `channelId`/`authors`/`since`/`until` (`e2eBridge.ts:7946-7949`), but neither branch consumes them: the mock branch filters only on `hit.content` / `hit.channel_name`, and the relay branch builds `{ kinds, search, limit }`. So the type widening is inert and no Playwright spec can assert that `from:`/`in:`/`after:` actually narrow anything. Between the parser `.mjs` test (syntax only) and the Rust test (filter shape only), the resolution-and-wiring layer — where both must-fix defects live — is untested. One spec driving the topbar with an operator would have caught defect 1.

**5. The broadened `enabled` predicate in `useSearchMessagesQuery` is unreachable.** `hooks.ts:43-49` now enables the query when `authors`/`since`/`until`/`channelId` are present even with a short query, but both call sites gate on FTS length first: `useSearchResults.ts:157` requires `ftsQuery.length >= MIN_SEARCH_QUERY_LENGTH`, and `useChannelFind.ts:69` requires `debouncedQuery.length >= MIN_QUERY_LENGTH`. Dead branch — remove it, or move the decision here and drop the caller-side gate. Keeping both invites the two floors drifting apart.

## Consider
- `after:` is inclusive of its day (`since` = local start of day) while `before:` is exclusive (`until` = local start of day, so messages *on* that date are excluded). Slack behaves the same way, so it's likely intended — but the doc comments at `parseSearchOperators.ts:18-21` describe both identically as "unix seconds at local start of day" and don't note the asymmetry. One clause each.
- `hasSearchQuery` (`useSearchResults.ts:121-126`) now becomes true on operator presence alone, which enables `useManagedAgentsQuery` and `useRelayAgentsQuery` for a query like `from:ab` that renders nothing (`userResults` short-circuits on empty `ftsQuery`). `useRelayAgentsQuery`'s own comment calls it "an unfiltered relay query for the whole profile set" — cheap here thanks to the shared cache, but it is work with a guaranteed-empty render.
- `isChannelUuid` matches any RFC-4122-shaped UUID and is trusted as a channel id without membership checking. Harmless — the relay intersects `#h` against `accessible_channels` — but the `in:` short-circuit at `useSearchResults.ts:53-55` means a well-formed UUID for an inaccessible channel returns empty rather than falling back to a name match.

## Gates
Both formatting gates are red at this head and I reproduced the Rust one locally: `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml --all -- --check` reports 2 diffs in `desktop/src-tauri/src/commands/messages_tests.rs` (lines 93 and 109). The `Desktop Core` log shows Biome format errors in `desktop/src/features/search/lib/parseSearchOperators.test.mjs` and `desktop/src/features/search/useSearchResults.ts`. `just desktop-fix` + `just desktop-tauri-fmt` clears both. Everything else in run `30168952860` is green.

desktop/src/features/search/lib/parseSearchOperators.ts
jatinder14: Updated in 31cd10f. Operator matching now starts at a token boundary, so hyphen/slash cases like built-in:react or https://x.com/in:foo stay in normal search text. Added tests for those, and trailing punctuation like in:general, still resolves.

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

desktop/src/features/search/lib/parseSearchOperators.ts
jatinder14: Updated in 31cd10f. Docs now say after: is inclusive (local start of day) and before: is exclusive (local start of day, Slack-style).

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

desktop/src/features/search/useSearchResults.ts
jatinder14: Updated in 31cd10f. from:/in: now use a clear none / resolved / unresolved result. If unresolved, search stays empty instead of quietly widening. Trailing punctuation on values is stripped at parse time.

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

desktop/src/features/search/useSearchResults.ts
jatinder14: Updated in 31cd10f. Author candidates now include a handle lookup for humans as well as the agent results. Multi-word handles are still limited to the first token in this pass, and that is noted in the parser docs.

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

desktop/src/features/search/hooks.ts
jatinder14: Updated in 31cd10f. Removed the extra enable clauses in useSearchMessagesQuery. It now uses one trimmedQuery.length >= 2 gate, same as the call sites.

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

desktop/src/testing/e2eBridge.ts
jatinder14: Updated in 31cd10f. The e2e bridge mock path filters on channelId / authors / since / until, and the relay path forwards those fields in the NIP-50 filter.

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

jatinder14:
Hey @wpfleger96, pushed fixes in 31cd10f for the review notes:

1. Operator regex only matches at token start, so things like built-in:react stay in normal search text
2. Unresolved from:/in: no longer silently widens results
3. from:@name can resolve humans, not only agents
4. e2eBridge now applies the operator fields
5. Docs clarify after/before day boundaries
6. Simplified the hook enable gate
7. Formatted the touched files

Happy to take another look whenever you have time. Thanks!

desktop/src/features/search/useSearchResults.ts
```diff
@@ -55,21 +110,127 @@ export function useSearchResults({
...
+    since: parsedQuery.since,
+    until: parsedQuery.until,
+  });
+
+  const messageResults = React.useMemo(() => {
+    if (hasUnresolvedOperator && !waitingOnFromResolution) {
+      return [];
+    }
+    return dedupeSearchHits(searchQuery.data?.hits ?? []);
+  }, [hasUnresolvedOperator, waitingOnFromResolution, searchQuery.data?.hits]);
```
wpfleger96: **Must fix — during `from:@name` resolution this returns cached *unfiltered* results for the same FTS text.**

The `waitingOnFromResolution` exception is the one path where `hasUnresolvedOperator` is true but the guard falls through to `searchQuery.data`. That data is not empty — it is the previous wide search:

1. User types `deploy` → query key `["search-messages","deploy",12,null,null,null,null]` fetches and caches unfiltered hits (`staleTime: 30_000`).
2. User continues to `deploy from:@will` → `ftsQuery` is still `"deploy"`, so `channelId`/`authors` are `undefined` while unresolved and the key is **byte-identical** to step 1.
3. `authorResolution.status === "unresolved"` and `fromUserSearchQuery.isLoading` → `waitingOnFromResolution` is true → `searchQuery` is disabled, but this `useMemo` skips the `return []` branch and reads `searchQuery.data?.hits`.

A disabled observer still gets cached data for a matching key. Verified against the repo's `@tanstack/react-query` 5.100.14:

```js
const qc = new QueryClient();
const key = ["search-messages", "deploy", 12, null, null, null, null];
qc.setQueryData(key, { hits: [{ id: "wide-1" }, { id: "wide-2" }] });
const obs = new QueryObserver(qc, { queryKey: [...key], enabled: false, queryFn: async () => ({ hits: [] }) });
obs.getCurrentResult();
// => { data: { hits: [ {id:"wide-1"}, {id:"wide-2"} ] }, isLoading: false, fetchStatus: "idle" }
```

So for the duration of the user-search fetch the topbar shows unnarrowed results as if they satisfied `from:@will` — the exact silent-widening the `OperatorResolveResult` union was added to prevent, just moved into the loading window. It resolves itself once the lookup settles (`[]` if unmatched, refetch under a new key if matched), but a few hundred ms of confidently wrong results is worse than a brief empty state.

To be precise about the trigger: it needs a cache entry for the bare FTS text, which means the user searched `deploy` at some point in the last `gcTime` (5 min) and then appended `from:@name` — pause-type-resume, or refining a previous search. It is not every keystroke, since the 300 ms debounce often skips the bare-text state.

Two things to change:

1. Make this guard unconditional: `if (hasUnresolvedOperator) return [];`. Keep `waitingOnFromResolution` for the *empty-state copy* in `TopbarSearch` (show a spinner instead of "no results") rather than as a data-passthrough.
2. Encode the unresolved state in the query key in `hooks.ts` so cache identity can never collide with the plain query. Right now the only thing separating "unresolved operator" from "no operator" in the cache is this one `useMemo` branch, and `searchQuery` is also returned raw from the hook — any future consumer that reads `searchQuery.data` directly reintroduces the bug with no compiler signal.

There are no unit tests for `useSearchResults.ts`, so neither of these has a regression net today.

wpfleger96 (changes requested):
🤖

Re-reviewed at head `31cd10f778a3bfeeac5de25478173f2d6e3bcc48` (was `e60e1c29f94931c29f109b750d95f3884f33e79b`). **All 6 of my earlier comments are addressed** — I re-derived each one by executing code at the new head rather than reading the diff, and resolved those threads. Two things still block: one behavioral defect that the fix left in a narrow window, and the red CI, which has a single root cause and a single fix.

### Verified fixed

| # | Finding | Evidence at `31cd10f77` |
|---|---|---|
| 1 | Unresolved operator silently widens the search | `OperatorResolveResult<T>` (`none` / `resolved` / `unresolved`) now distinguishes absent from unmatched; `useSearchResults` derives `hasUnresolvedOperator`, disables the query and returns `[]`. Correct in the settled state — see the one exception below. |
| 2 | `\b` in `OPERATOR_RE` | Now `/(?:^|\s)(from\|in\|after\|before):(\S+)/gi`. Ran the parser over 27 adversarial inputs: `built-in:react`, `sign-in:flow`, `opt-in:true`, `https://x.com/in:foo`, `a-in:b`, `foo/from:bar`, `--in:x` all stay plain text with `in`/`from` `null`. Legit cases still parse (`deploy in:general`, `IN:General FROM:@bob`, `in:#general!`, `x in:general.`, last-wins on repeats, `email:a@b.com in:general`, `after:2026-02-30` correctly rejected as invalid and left in text). `cleanOperatorValue` handles the `in:general,` case I flagged. |
| 3 | `authorCandidateSeed` agents-only | `useUserSearchQuery(fromHandleForLookup)` added and deduped into the seed on `normalizePubkey`, so `from:@name` can resolve a human. The multi-word case (`from:@Will Pfleger` → `from: "@Will"`, `Pfleger` left in text) is unfixed but now documented as a part-1 limitation in the module doc — fine by me for this PR. |
| 4 | `e2eBridge` args widened but inert | Mock branch now filters on `channelId`, lowercased `authors` set, `since` (`>=`) and `until` (`<`); relay branch forwards `#h`/`authors`/`since`/`until` and widens `kinds` to `[9, 40002, 45001, 45003]` to match `build_search_messages_filter`. |
| 5 | Unreachable `enabled` predicate in `hooks.ts` | Now just `enabled && trimmedQuery.length >= 2`, with a comment putting the floor ownership at the call sites. One threshold, no drift. |
| 6 | `authors`/`since`/`until` pushdown and access control | Unchanged and still fine — `search_hit_accepted` re-authorizes every hit. |

Formatting, which was the other reviewer's only blocker, is clean at this head: `cargo fmt --all -- --check` on `desktop/src-tauri` exits 0, and `biome check .` reports 1631 files with no fixes.

### Still blocking

**1. Must fix — cached *unfiltered* results leak through while `from:@name` is resolving.** Inline on `useSearchResults.ts`. The `waitingOnFromResolution` exception in the `messageResults` guard is the one path where the query is disabled but the memo still reads `searchQuery.data`, and because `channelId`/`authors` are `undefined` while unresolved, the query key for `deploy from:@will` is byte-identical to plain `deploy`. A disabled react-query observer still returns cached data for a matching key — I confirmed that against the repo's `@tanstack/react-query` 5.100.14. Fix is small: make the guard unconditional and encode the unresolved state in the query key so cache identity can't collide.

**2. CI is red on three checks with one cause: the branch merged a broken `main` commit.** Nothing in your search code is at fault.

- `Desktop Core` fails at `pnpm check` → `check-file-sizes`: `src-tauri/src/managed_agents/runtime.rs: 2220 lines (limit 2216)`. That file is untouched by this PR. The merge at `82bdba7b6` pulled in `dd222a509`, which was itself red on `Desktop Core` for the same reason ([job 89771442557](https://github.com/block/buzz/actions/runs/30193727120/job/89771442557)). Since then #2974 split the runtime and `runtime.rs` on `main` is 982 lines with the override removed entirely.
- `Desktop` is just the aggregate gate reporting `Desktop Core finished with: failure`.
- `DCO Check` says `1/3 commits are missing DCO sign-off` — that commit is the merge `82bdba7b6`. Both of your own commits are signed off correctly.

A rebase onto current `main` fixes all three, because it drops the merge commit and the stale `runtime.rs`:

```bash
git fetch upstream main
git rebase --signoff upstream/main
git push -f origin feat/search-query-operators
```

I verified that locally by rebasing `31cd10f77` onto `8c0e8cb16`: no conflicts, `check-file-sizes` exits 0, `cargo fmt --check` exits 0, `biome check .` clean, `tsc --noEmit` clean, and the unit suite is 3527/3528 with the single failure being `styled-qr-code.test.mjs` failing to resolve the `qrcode` package — which reproduces identically on an untouched checkout, so it is a local `node_modules` artifact of my setup, not yours.

### Non-blocking

- **No E2E spec asserts an operator narrows results.** The bridge now consumes the filters, so this is finally testable — one spec seeding two channels and checking `in:<channel>` returns only the in-channel hit would cover the layer where both of the original must-fixes lived. `rg 'in:general|from:@' desktop/tests` returns nothing today.
- **No UI signal for an unresolved operator.** `deploy in:nosuchchannel` renders as an ordinary empty result set, so the user can't tell "nothing matched" from "I typo'd the channel name". `hasUnresolvedOperator` already exists in the hook; surfacing it as "no channel named nosuchchannel" would be a cheap follow-up.
- **`in:...` and `in:!!!` parse to `in: ""`.** `cleanOperatorValue` strips the whole value, and empty string then routes through `resolveChannelIdFromOperator` → `normalizeInChannel("")` → no match → `unresolved`, so behavior is safe (empty results). Worth an early `return { status: "none" }` on empty for clarity, but nothing breaks.

Quality gate: Minimalism 9/10, Elegance 9/10, Correctness 8/10 — held down by the cache-collision window in item 1 above and by `useSearchResults.ts` having no unit tests, so neither that guard nor the resolution logic has a regression net.

desktop/src/features/search/useSearchResults.ts
jatinder14: Updated in 05c5857. While an operator is unresolved (including while from:@name is loading), message results stay empty. The query key also carries unresolvedOperator so it cannot reuse the plain-text search cache.

@wpfleger96 feel free to resolve this when you are happy with it. Thanks!

jatinder14:
@wpfleger96 fixed the cache collision while from:@name is still resolving, and rebased onto current main so CI/DCO are clean. Ready when you are. Thanks!

wpfleger96 (approved):

wpfleger96:
@jatinder14 thank you for this!!

PR #2885, opened 2026-07-25
feat(buzz-acp): add auto permission mode

### LINKED ISSUE #2884: buzz-acp: no way to select Claude Code's `auto` permission mode — adapter advertises it, PermissionMode omits it
## Summary

`buzz-acp` can't be configured into Claude Code's `auto` permission mode, because `PermissionMode` doesn't have an `auto` variant — even though the adapter on the other end advertises it.

`BUZZ_ACP_PERMISSION_MODE=auto` fails clap validation at startup, so the value can never reach the adapter.

## The adapter already offers it

Speaking ACP directly to `claude-agent-acp` (0.61.0), `session/new` returns:

```
MODE currentValue: default
   - auto | Auto
   - default | Manual
   - acceptEdits | Accept Edits
   - plan | Plan Mode
   - dontAsk | Don't Ask
   - bypassPermissions | Bypass Permissions
```

`auto` is first in the list. `session/set_config_option` with `{configId: "mode", value: "auto"}` would be accepted today.

`PermissionMode` (`crates/buzz-acp/src/config.rs:123`) has five variants and omits it:

```rust
Default | AcceptEdits | BypassPermissions | DontAsk | Plan
```

## Why `auto` specifically matters for this harness

`buzz-acp` auto-approves every `session/request_permission` with `allow_once` (`acp.rs:1068`, `acp.rs:1671`). That's the right call — there's no human at a terminal to answer a prompt, and forwarding one would hang the turn.

The consequence is that the modes collapse:

| Mode | Effect under buzz-acp |
| --- | --- |
| `default` | Agent asks, buzz-acp says yes — indistinguishable from bypass |
| `acceptEdits` | Same |
| `bypassPermissions` | Skips the flow (the default) |
| `plan` | No tool execution |
| `dontAsk` | Rejects anything needing permission |

So in practice an operator picks between "no checks at all" and "can't do any work". `auto` is the missing middle: Claude Code describes it as auto-approving tool calls with background safety checks that verify actions align with the request, so a classifier stays in the loop *without* a prompt needing an answer. That's exactly the shape a headless, chat-driven harness wants.

This is a gap rather than a disagreement with the trust model in `VISION_AGENT.md` ("the shell runs at the operator's trust level, like bash itself") — `auto` doesn't restrict what an operator may do, it adds a review step that doesn't require them to be present.

Worth noting the asymmetry it leaves today: Codex-runtime agents run inside Codex's own Seatbelt sandbox (which `config.rs:627-639` has explicit workarounds for), while Claude-runtime agents run with no sandbox and no permission checks.

## Proposed fix

Add the variant and its wire string. PR: skyerus/buzz@feat/acp-auto-permission-mode

```rust
/// Auto-approve tool calls, with a classifier reviewing each one.
#[value(alias = "auto")]
Auto,
// as_wire_str():
Self::Auto => "auto",
```

Tests extended to cover the wire string, `is_default`, and both the kebab-case and camelCase parse paths, matching the existing per-mode coverage. `cargo test -p buzz-acp --lib permission_mode` passes; `cargo fmt --check` clean; no new clippy warnings.

## Known related issue upstream

agentclientprotocol/claude-agent-acp#607 reports that `permissions.defaultMode: "auto"` in `settings.json` has no effect through the adapter. That's a different mechanism from `session/set_config_option`, which is the path `buzz-acp` uses and the one the adapter advertises — but worth knowing about when verifying end-to-end behaviour.

## Environment

- Buzz 0.4.25 (desktop), `claude-agent-acp` 0.61.0, Claude Code 2.1.220, macOS 15.3 arm64

skyerus:
Closes #2884.

## What this solves

`buzz-acp` can't be put into Claude Code's `auto` permission mode. `PermissionMode` has five variants and `auto` isn't one, so `BUZZ_ACP_PERMISSION_MODE=auto` fails clap validation at startup and the value never reaches the adapter — even though the adapter advertises it.

Probing `claude-agent-acp` 0.61.0 over ACP, `session/new` returns `auto` first in the `mode` config option:

```
   - auto | Auto
   - default | Manual
   - acceptEdits | Accept Edits
   - plan | Plan Mode
   - dontAsk | Don't Ask
   - bypassPermissions | Bypass Permissions
```

## Why `auto` is the mode this harness wants

`buzz-acp` auto-approves every `session/request_permission` with `allow_once` (`acp.rs:1068`, `acp.rs:1671`), which is correct for a chat-driven harness — there's no human at a terminal, and forwarding a prompt would hang the turn.

The side effect is that `default` and `acceptEdits` become indistinguishable from `bypassPermissions`, leaving only `plan` and `dontAsk` as modes that constrain anything, and both are too restrictive for real work. `auto` is the missing middle: a classifier reviews each tool call, with no prompt for anyone to answer.

## Implementation

Two lines of behaviour — the enum variant and its wire string:

```rust
/// Auto-approve tool calls, with a classifier reviewing each one.
#[value(alias = "auto")]
Auto,

// as_wire_str()
Self::Auto => "auto",
```

The `#[value(alias = "auto")]` form matches the existing style (`Default` and `Plan` carry the same redundant-but-explicit alias). Doc comments on the enum and on the `--permission-mode` arg updated to mention it.

## Tests

Extended the four existing per-mode tests rather than adding new ones, so `auto` is covered everywhere the other five are:

- `test_permission_mode_wire_strings` — `"auto"`
- `test_permission_mode_is_default` — not the default
- `test_permission_mode_value_enum_kebab_case` — `"auto"` parses
- `test_permission_mode_value_enum_camel_case_aliases` — alias path

```
cargo test -p buzz-acp --lib permission_mode
test result: ok. 7 passed; 0 failed
```

`cargo fmt -p buzz-acp -- --check` clean. `cargo clippy -p buzz-acp --lib --all-features` produces no new warnings (the two it reports are pre-existing, both in `queue.rs`).

## Testing it manually

```bash
BUZZ_ACP_PERMISSION_MODE=auto buzz-acp --relay-url … --private-key …
```

Startup log should read `permission_mode=auto` where it previously rejected the value.

## Deferred

Not addressed here: whether the adapter honours `auto` identically to the interactive CLI. agentclientprotocol/claude-agent-acp#607 reports `permissions.defaultMode: "auto"` in `settings.json` having no effect, though that's a different mechanism from `session/set_config_option` — which is the path this uses and the one the adapter advertises. Worth verifying end-to-end before relying on the classifier as a security control.


dophsquare:
🤖 Review: no line-level code changes requested; the enum, CLI parsing, wire mapping, and the capability gate are consistent. The harness only sends `auto` when the ACP agent advertises that mode.

Action items before merge:
- Resolve the failing `Desktop E2E Integration (1/2)` check, which currently fails during `Start relay`.
- Rerun the full required CI after the failure is addressed or confirmed transient.

Once CI is green, this is ready on code review.

skyerus:
Addressing the CI action item: direct rerun of run 30404015628 was unavailable because this account lacks repository admin rights, so I pushed no-code commit 0cace0905b082e9e30799b9d35f06ca07bd168cb to retrigger the full required workflows. Its tree exactly matches parent 2bb6578ed005785e8068dd0717b0191c5f3e8c25. GitHub has gated both fork runs as action_required; @dophsquare, please approve [CI run 30527035334](https://github.com/block/buzz/actions/runs/30527035334) and [Docker run 30527035484](https://github.com/block/buzz/actions/runs/30527035484). I’ll verify the exact head once they complete.

PR #2888, opened 2026-07-25
fix(deploy): add Tauri webview origins to the compose BUZZ_CORS_ORIGINS default

### LINKED ISSUE #2872: [Bug] deploy/compose: default BUZZ_CORS_ORIGINS omits Tauri webview origins — desktop "Join an existing community" fails with "Load failed"
## Summary

A relay deployed from the `deploy/compose` template blocks the desktop app at CORS preflight. The template ships

```
BUZZ_CORS_ORIGINS=https://buzz.example.com
```

(operators substitute their own domain), but the desktop app's WebView runs from origin `tauri://localhost` (macOS/Linux) or `http://tauri.localhost` (Windows). Since neither is in the allowlist, the WebView's HTTP API calls to the relay are rejected, and "Join an existing community" fails with WebKit's generic `Load failed`.

What makes this confusing to debug: `curl` and NIP-11 checks against the same relay succeed, so the relay looks perfectly healthy from everywhere except the app.

## Environment

- Buzz Desktop v0.4.26 (macOS)
- Relay `ghcr.io/block/buzz@sha256:77e1c6b1…`, self-hosted via `deploy/compose` with Caddy TLS

## Steps to reproduce

1. Fresh `deploy/compose` deployment: copy `.env.example` → `.env`, set `BUZZ_CORS_ORIGINS=https://<your-domain>` per the template.
2. Desktop app → Add community → Join an existing community → enter the relay URL.
3. Join fails with `Load failed`.
4. `curl -H 'Accept: application/nostr+json' https://<your-domain>` returns the NIP-11 document fine.

## Cause

The relay parses `BUZZ_CORS_ORIGINS` as a comma-separated list of exact origins (`crates/buzz-relay/src/config.rs`) and `build_cors_layer` in `crates/buzz-relay/src/router.rs` deliberately does not fall back to permissive CORS when the variable is set. That behavior is sensible — the problem is only that the compose template's default value omits the desktop origins, so every self-hosted deployment configured from it breaks the desktop app.

## Suggested fix

Include the desktop origins in the template default:

```
BUZZ_CORS_ORIGINS=https://buzz.example.com,tauri://localhost,http://tauri.localhost
```

and add a short note to `deploy/compose/README.md` (currently silent on CORS) that the Tauri origins must stay in the allowlist for the desktop app to work against a self-hosted relay.

Verified on the deployment above: adding those two origins and restarting the relay resolves the join failure.

## Related (not duplicates)

- #2617 documents `BUZZ_CORS_ORIGINS` + the Tauri origins, but in the root `.env.example` — the value shipped by `deploy/compose/.env.example` is unchanged.
- #2862 moves the join-policy fetch to native networking, which would bypass CORS for direct relay-URL joins — but other WebView HTTP calls (e.g. invite minting/redemption, moderation) still need these origins allowlisted, so the compose default remains a footgun even after it lands.

nscheifler:
## Summary

The compose template `deploy/compose/.env.example` sets `BUZZ_CORS_ORIGINS` to one public domain. The webview of the desktop app has a different origin. The origin is `tauri://localhost` on macOS and Linux. The origin is `http://tauri.localhost` on Windows.

The relay reads `BUZZ_CORS_ORIGINS` as a comma-separated list of exact origins (`crates/buzz-relay/src/config.rs`; `build_cors_layer` in `crates/buzz-relay/src/router.rs`). When the variable is set, the relay does not permit other origins. This design is correct, but the template value does not include the webview origins. Thus the desktop app cannot use the HTTP API of a relay that an operator deploys from the template.

The first failure that the user sees is in the "Join an existing community" flow. The flow stops and shows the error `Load failed`. At the same time, the relay replies correctly to a NIP-11 request from `curl`. Thus the relay seems healthy, and the CORS cause is hard to find.

This change:

- Adds the two webview origins to the default value of `BUZZ_CORS_ORIGINS` in `deploy/compose/.env.example`.
- Adds a comment to the same file. The comment tells the operator to keep the webview origins when they set their own domain.
- Adds a note with the same requirement to the "Production notes" section of `deploy/compose/README.md`.

### Related issue

Fixes #2872.

The closest existing PRs (not duplicates):

- #2617 adds comments about `BUZZ_CORS_ORIGINS` to the root `.env.example`. It does not change the value in the compose template. The two changes are compatible.
- #2862 moves the join-policy fetch to native networking. The invite and moderation calls stay in the webview. Thus the compose template must continue to supply the webview origins.

### Testing

This is a change to a template and a document only. No code paths change, so there are no new tests.

We did the test on a live deployment: Buzz Desktop v0.4.26 on macOS, relay deployed with `deploy/compose` and Caddy TLS.

1. Deploy the relay with the unmodified template value. The join flow fails with the error `Load failed`.
2. Add `tauri://localhost,http://tauri.localhost` to `BUZZ_CORS_ORIGINS`. Restart the relay.
3. Do the join flow again. The desktop app joins the community correctly.

### DIFF
--- a/deploy/compose/.env.example
+++ b/deploy/compose/.env.example
@@ -10,7 +10,10 @@ BUZZ_DOMAIN=buzz.example.com
 RELAY_URL=wss://buzz.example.com
 BUZZ_MEDIA_BASE_URL=https://buzz.example.com/media
 BUZZ_MEDIA_SERVER_DOMAIN=buzz.example.com
-BUZZ_CORS_ORIGINS=https://buzz.example.com
+# CORS allowlist for the relay HTTP API, comma-separated exact origins. Replace the domain,
+# but keep the tauri entries — the desktop app webview (tauri://localhost on macOS/Linux,
+# http://tauri.localhost on Windows) cannot reach the relay HTTP API without them.
+BUZZ_CORS_ORIGINS=https://buzz.example.com,tauri://localhost,http://tauri.localhost
 
 # Production defaults. Closed relay mode requires RELAY_OWNER_PUBKEY and a stable relay key.
 BUZZ_REQUIRE_AUTH_TOKEN=true
--- a/deploy/compose/README.md
+++ b/deploy/compose/README.md
@@ -28,6 +28,11 @@ keypair.
 - Requires Docker Compose v2.24.4 or newer; the TLS override uses Compose's
   `!reset` tag to remove the direct relay port when Caddy terminates HTTPS.
 - Default `BUZZ_IMAGE` tracks `ghcr.io/block/buzz:main` for early testing. Pin it to `ghcr.io/block/buzz:sha-<7>` or a semver release tag for production once available.
+- `BUZZ_CORS_ORIGINS` must keep the desktop webview origins (`tauri://localhost`
+  for macOS/Linux, `http://tauri.localhost` for Windows) alongside your public
+  domain. The relay treats the list as exact origins with no permissive
+  fallback, so without the tauri entries desktop flows that use the HTTP API
+  (joining a community, invites, moderation) fail with a generic "Load failed".
 - Keep `BUZZ_RELAY_PRIVATE_KEY`, `BUZZ_GIT_HOOK_HMAC_SECRET`, database/Redis,
   and S3 secrets stable across restarts.
 - `RELAY_OWNER_PUBKEY` is intentionally not prefixed with `BUZZ_`; it must be a


quentindecock:
Independently hit this on **Desktop 0.5.0** (report above was 0.4.26) and can confirm the fix. One addition: the impact is broader than "Join an existing community" — it also breaks **invite minting** from the desktop app.

`mintInvite()` in `desktop/src/shared/api/invites.ts` is one of the few paths that calls the relay via the **webview's `fetch()`** rather than the Rust HTTP client, so it's the only place CORS applies. Everything else — `/events`, `/query`, the WebSocket — bypasses it entirely, which makes this fail in a confusing way: chat works perfectly while one button silently doesn't.

Concrete symptom, with `BUZZ_CORS_ORIGINS` set only to the relay's own origin:

```
$ curl -i -X OPTIONS -H 'Origin: tauri://localhost' \
    -H 'Access-Control-Request-Method: POST' \
    -H 'Access-Control-Request-Headers: authorization,content-type' \
    https://<relay>/api/invites
HTTP/2 200
access-control-allow-methods: *
access-control-allow-headers: *
# no access-control-allow-origin
```

Preflight returns 200 but omits `Access-Control-Allow-Origin`, so the webview blocks it, `fetch()` throws before sending, and **the relay logs nothing at all** — I confirmed zero `POST /api/invites` requests with `tower_http=debug`. The UI shows only "Couldn't copy the invite link", because the `catch` wraps both the mint and the clipboard write. Adding the two origins fixed it immediately.

PR #2045, opened 2026-07-17
Inbox refactor

delkc:
## Why

The Inbox mixed overlapping feed categories with personal work queues, so **All** was not actually comprehensive and several filters did not make it clear why an item appeared. Threads and DMs could produce one row per event instead of one row per conversation, drafts were hidden until selected, and reminders appeared through multiple competing presentations.

This refactor makes the **Inbox** a focused, conversation-oriented place to catch up on work relevant to you. It is intentionally not a mirror of every unread event in every channel.

## What changed

- Keep the destination named **Inbox** and use the standard Lucide bell icon.
- Refocus **All** on DMs, mentions, thread replies, needs-action items, replies from agents the user owns or controls, due reminders, and active drafts.
- Exclude generic top-level channel traffic and updates from agents the user does not own or control.
- Group each thread or DM into one row, sorted by latest activity.
- Resume an unread conversation at its oldest unread message while opening the full thread or DM in the detail pane.
- Reuse the existing **New** divider at the unread boundary.
- Make the detail title a direct link to the canonical conversation.
- Give Reminders and Drafts the same list/detail interaction and location metadata as conversation rows.
- Separate Reminders and Drafts from message filters with a subtle divider, without adding another labeled section.
- Put reminder and draft counts beside their corresponding filter labels instead of on the generic filter button.
- Preserve the selected conversation when switching filters if it remains valid; otherwise select a valid replacement without flashing stale detail.
- Use filter-specific empty states and rename the options toggle to **Show unread only**.
- Ship the focused behavior directly. The earlier experiment gate, Custom view, and default-view controls have been removed from this PR to keep the first pass focused.

## Filter model

| Filter | What appears |
| --- | --- |
| **All** | One row per personally relevant conversation, plus due reminders and active drafts. Includes DMs, mentions, thread replies, explicit needs-action items, and replies from agents the current user owns or controls. Excludes generic top-level channel traffic, other agents' updates, and reminders that are not due yet. |
| **Mentions** | Conversations containing a direct mention. Each conversation appears once and opens with full context. |
| **Threads** | Conventional threaded replies, grouped to one row per thread. Broadcast replies are not treated as conventional thread replies. |
| **Needs action** | Feed items explicitly classified as requiring action. |
| **Agents** | Conversations whose representative response was authored by an agent the current user owns or controls, including top-level DM responses. If a human replies afterward, the conversation leaves this filter until an owned agent responds again. |
[...]

PR #2889, opened 2026-07-25
feat(mobile): refactor Activity behavior and ui

tellaho:
**Category:** improvement
**User Impact:** Mobile users can scan Activity as a focused conversation inbox and open the exact unread message or thread represented by each item.

## Context

Mobile's Activity tab had not kept pace with Desktop: it presented isolated event headlines, advertised categories that were often empty, and opened a channel without clearly landing on the selected item.

This PR brings the Mobile surface toward the conversation-oriented direction explored in Clay Delk's Desktop [Inbox refactor PR #2045](https://github.com/block/buzz/pull/2045), while adapting it to Mobile rather than copying the Desktop split-pane implementation. The related product/UX discussion is captured in the originating [Buzz thread](buzz://message?channel=a9bbc0e5-d25d-4740-849c-93c34bb578a4&id=a7d9a4d33dcd8c6bf0dc67d81c328892b9e38dedaa8548920224ef388301b6ab).

## UX decisions in this PR

- **Conversation-oriented, not event-oriented:** related updates collapse into one row per thread/DM conversation, represented by the latest update and ordered by latest activity. Separate top-level conversations in the same channel remain separate rows.
- **Resume at the oldest unread:** tapping a grouped row opens the represented canonical message/thread/DM at its oldest unread item, rather than merely opening the channel at an arbitrary position.
- **Desktop-aligned row hierarchy:** rows lead with a full avatar and sender, followed by contextual location/type metadata, unread dot + time, and a two-line preview. A **New** boundary separates unread and read content.
- **Mobile-native navigation:** Mobile keeps a single-column `Activity → canonical conversation → Back` flow. It does not introduce Desktop's persistent detail pane.
- **Compact filtering:** the old horizontal chip rail becomes a compact filter menu so the source set fits a phone viewport without horizontal scanning. Filters are All, Mentions, Threads, Needs Action, Activity, Agents, Reminders, and Drafts.
- **Focused source semantics:** All covers personally relevant work—DMs, mentions, thread replies, needs-action events, owned-agent activity, due reminders, and active drafts—rather than becoming a generic stream of every channel message. Mobile's standalone Activity source is currently limited to DM traffic because it does not have Desktop's aggregated channel-activity feed.
- **Shared read behavior:** rows project canonical channel/thread/message markers, support unread-only and mark-all-read, and use local overrides only where canonical markers cannot represent an item.
- **Reminders and drafts are real data:** reminders use the same encrypted NIP-ER events as Desktop. Drafts persist device-local composer state, restore on return, survive failed sends, and clear after successful sends.
- **Explain navigation failures:** an unavailable destination produces an explanatory message rather than silently doing nothing or falling back to an unrelated channel position.

## Implementation summary

- Adds a Mobile inbox model for conversation grouping, category priority, contextual labels, sorting, filtering, and oldest-unread targets.
- Expands relay-backed sources for mentions, approvals, owned-agent lifecycle events, and DM traffic.
- Adds fail-closed NIP-ER reminder decryption and device-local compose-draft persistence.
- Redesigns Activity rows, boundaries, filters, unread controls, and empty/loading states.
- Routes rows through Mobile's existing canonical channel/thread screens with precise target IDs.
- Adds model, provider, widget, reminder, read-state, draft-lifecycle, and deep-link coverage.

## Reproduction steps

1. Run Mobile and open **Activity**.
2. Confirm full avatars, sender-first rows, context labels, unread indicators, timestamps, two-line previews, and the compact filter control.
3. Open the filter menu and verify All, Mentions, Threads, Needs Action, Activity, Agents, Reminders, and Drafts.
4. Tap a grouped thread row and confirm the canonical conversation opens at its oldest unread message.
5. Mark rows read/unread, enable unread-only mode, and use mark-all-read; confirm state agrees with the channel/thread destination.
6. Type without sending in a channel or thread, leave, and confirm the draft appears in Activity and restores in the composer.

## Screenshots

| Before — merge-base `dd222a509` | After — PR head `52ad40aee` |
|---|---|
| image | image |

Recaptured on the same authenticated iPhone 17 simulator, account, theme, and Activity view, at this PR's current merge-base (`dd222a509`) and head (`52ad40aee`). Both frames were taken within a few minutes on the same live feed, so the visible conversation set overlaps closely (the recent Ned/Bart/Tommy items appear in both). The compared change is the row *structure*: Before leads with an `@ Mention` headline over a small inline avatar and a horizontal chip rail; After leads with a full avatar, a compact `labelMedium` sender label, contextual "Mentioned in" metadata, and a filter menu. The sender username now renders at the same compact scale the old `@ Mention` label used.

## Verification

- Current rebased head: `5bd87f4f4` on `origin/main` at `dd222a509`; GitHub reports the PR mergeable.
- `flutter analyze` — clean at `5bd87f4f4`.
- Full Mobile suite — 698 passed, 1 skipped, 4 failed; all four failures reproduce identically on clean `origin/main` (`channels_page_test` create-channel sheet and three `compose_bar_test` agent-mention cases).
- The prior PR-specific `home_page_test` failures were fixed by providing the Activity local-state dependency in that harness.
- Independent code and simulator UI review — approved.
- Post-rebase GitHub checks are running.


bhargavms:
the screenshots are broken (could be private, i get a FORBIDDEN error when i try to open it)

mobile/lib/features/activity/activity_provider.dart
```diff
@@ -27,6 +36,14 @@ class ActivityNotifier extends AsyncNotifier<HomeFeedResponse> {
 
     final session = ref.read(relaySessionProvider.notifier);
 
+    // DM channels come from the already-cached channel list when available;
+    // a missing list only skips the DM source for this refresh.
+    final dmChannelIds = [
+      for (final channel
+          in ref.read(channelsProvider).asData?.value ?? const <Channel>[])
```
wesbillman, line 61: **Blocking:** This one-time synchronous read races channel loading. `HomePage` builds the Channels and Activity tabs together, so on a cold start `channelsProvider` can still be loading here; `dmChannelIds` then stays empty, the DM query is skipped, and this provider does not rebuild when channels resolve because it only watches the relay providers. DMs remain absent until a manual refresh or reconnect. Please await/watch the channel result (or explicitly invalidate Activity when it changes) and add a provider test covering loading → data.

mobile/lib/features/activity/inbox_item.dart
```diff
@@ -0,0 +1,261 @@
...
+    InboxFilter.reminders || InboxFilter.drafts => false,
+  };
+}
+
+/// Group raw feed items into conversation rows sorted by latest activity.
+/// Mirrors desktop's `buildInboxItems`.
+List<InboxItem> buildInboxItems(Iterable<FeedItem> feedItems) {
+  final groups = <String, List<FeedItem>>{};
+  for (final item in feedItems) {
+    final key = inboxConversationId(item.tags, item.id);
```
wesbillman: **Blocking:** Ordinary top-level DM messages have no NIP-10 root/reply reference, so this key falls back to each event ID. Multiple messages in the same DM therefore produce separate rows, contrary to the PR's stated one-row-per-DM-conversation behavior. Please use the DM channel identity for non-thread DM events while preserving event-level grouping for ordinary top-level channel posts, and add a test with multiple messages in one DM.

mobile/lib/features/activity/compose_drafts_provider.dart
```diff
@@ -0,0 +1,130 @@
...
+
+String composeDraftKey(String channelId, {String? threadHeadId}) =>
+    threadHeadId == null ? channelId : '$channelId:$threadHeadId';
+
+/// SharedPreferences-backed store of unsent composer drafts, newest first.
+class ComposeDraftsNotifier extends Notifier<List<ComposeDraft>> {
+  @override
+  List<ComposeDraft> build() {
+    final prefs = ref.read(savedPrefsProvider);
+    final raw = prefs.getString(_draftsPrefsKey);
```
wesbillman: **Blocking privacy issue:** Draft plaintext is stored under one app-global preference key, and each draft is keyed only by channel/thread ID. Switching community/account does not clear or rebuild this provider, so the next community can display the previous account's draft text as an unavailable draft; a colliding channel ID could restore it into another composer. Please namespace persisted drafts by community and pubkey, react to identity changes, and add cross-community/account isolation coverage.

wesbillman (changes requested):
Three blocking issues need to be addressed before merge: DM loading must not depend on a one-time pre-channel-load snapshot, top-level messages in one DM must group into one conversation as described, and locally persisted draft text must be isolated by community/account. Please add regression coverage for each scenario.

mobile/lib/features/activity/activity_provider.dart
tellaho: Fixed in 5bebf680c — `ActivityNotifier` now watches the DM-channel id set via `channelsProvider.select(...)`, so the channels-provider loading → data transition (and DM membership changes) trigger a refetch instead of a one-shot read. Regression test in `mobile/test/features/activity/activity_provider_test.dart` reproduces the cold start: first fetch with channels still loading skips DMs; once channels resolve with a DM, the provider rebuilds and queries it.

🤖 Posted by Bart (AI agent) on behalf of tho.

mobile/lib/features/activity/inbox_item.dart
tellaho: Fixed in 3bea6a97d — non-thread DM events now group by `dm:<channelId>` via `inboxConversationId(dmChannelId:)` / `buildInboxItems(isDmChannel:)`, with `inboxItemsProvider` feeding DM channel ids from `channelsProvider`. Thread replies (including inside DMs) still group by thread root, and ordinary top-level channel posts keep per-event grouping. Tests: multi-message DM → one row (grouping + widget level), DM thread replies stay separate, channel posts unaffected (`inbox_item_test.dart`, `activity_page_test.dart`).

🤖 Posted by Bart (AI agent) on behalf of tho.

mobile/lib/features/activity/compose_drafts_provider.dart
tellaho: Fixed in d7ded957d — persisted drafts are now namespaced `compose_drafts_v1:<relayBaseUrl>:<pubkey>`, and the provider watches `relayConfigProvider` + `myPubkeyProvider` so community/account switches rebuild onto that identity's own store. Tests in `compose_drafts_provider_test.dart`: cross-community isolation, cross-account isolation with a colliding channel id, and an in-place community switch that rebuilds to empty then restores the original identity's draft on switch-back.

🤖 Posted by Bart (AI agent) on behalf of tho.

brow (commented):
🤖 I re-reviewed the current head `8332693`. The three issues from the earlier changes-requested review are now addressed with regression coverage:

- cold-start DM loading reacts when the channel list resolves,
- top-level DM messages group by DM channel while thread replies remain grouped by root,
- persisted drafts are namespaced by relay and pubkey.

Two blockers remain:

1. `mobile/lib/features/activity/activity_page.dart` is 1,072 physical lines (1,073 by the repository guard's newline count), above the hard 1,000-line ceiling in `mobile/scripts/check-file-sizes.mjs`. `AGENTS.md` explicitly requires splitting the file rather than adding an override. This is not currently caught by the Mobile CI job, so green checks do not clear it. Extracting the private sub-widgets into sibling part files should be behavior-neutral.

2. Provider-level draft namespacing does not protect a composer that remains mounted during an in-place community or account switch. `ComposeBar` restores and persists through an effect keyed only by `[controller, draftKey]`. The active draft provider rebuilds into identity B's namespace, but the controller can retain identity A's text because the effect does not rerun. The next edit then writes A's text into B's store. Please key or reset the composer integration on the active relay/pubkey identity and add a widget test that switches identity while `ComposeBar` remains mounted.

The second issue is a privacy/correctness blocker independently of the file-size policy. Please re-request review after both are fixed.

tellaho:
🤖 AI-generated update: Tom’s two remaining blockers are exhaustively addressed and pushed at head `6d5754e8e`.

- `6cc7736c7` fixes mounted-composer draft isolation by keying restore/persist lifecycle to relay + account identity and replacing the controller with only the destination identity’s saved draft (or empty). The widget regression covers A→B isolation, B-only persistence, and B→A restoration, and fails without the production fix.
- `6d5754e8e` behavior-neutrally splits `activity_page.dart` into same-library part files. The main file is now 316 lines, the largest part 253, no size override was added, and the file-size guard passes.

Final independent review approved both commits. Push hooks passed, including the full mobile test hook. Please re-review when convenient.

wesbillman (approved):
Re-reviewed at exact head `6fdf32ff9f64ec975dd3865c89db2a1a6592dffe`. The requested changes are addressed:

- Activity now rebuilds when the joined-DM channel set resolves or changes, with cold-start loading → data regression coverage.
- Non-thread DM events group by DM channel while thread replies retain thread-root grouping, with model and widget coverage.
- Draft persistence is namespaced by relay + pubkey, and a still-mounted composer replaces/restores text when identity changes, with provider and widget isolation tests.
- The oversized Activity page was split into sibling part files; all resulting files are below the repository's 1,000-line ceiling without an override.

I found no remaining blocker in the fix patches. `git diff --check` is clean. DCO, Semgrep, zizmor, changed-path detection, and the token guard are green; Mobile and desktop CI are still running at this exact head, so normal branch protection should remain the final merge gate.

mobile/lib/features/activity/inbox_local_state_provider.dart
```diff
@@ -0,0 +1,93 @@
...
+  const InboxLocalState({required this.doneIds, required this.unreadIds});
+
+  const InboxLocalState.empty() : doneIds = const {}, unreadIds = const {};
+}
+
+class InboxLocalStateNotifier extends Notifier<InboxLocalState> {
+  @override
+  InboxLocalState build() {
+    final prefs = ref.read(savedPrefsProvider);
+    final raw = prefs.getString(_localStateKey);
```
brow, line 31: 🤖 Could this local done/unread state be namespaced by relay and pubkey, like compose drafts? The global key and non-reactive provider let one account or community inherit another identity's row overrides during an in-place switch. This leaks only read-state metadata, so I don't think it blocks this PR, but the isolation asymmetry is worth fixing.

brow (approved):

Issue #2899, opened 2026-07-25
buzz-acp: one MCP server per agent forces a choice between the shell and any domain server


ArgussLabs:
## Summary

`buzz-acp` accepts a single MCP binary (`--mcp-command` / `BUZZ_ACP_MCP_COMMAND`), so an agent gets exactly one MCP server. Because a Buzz agent acts in Buzz by shelling out to the `buzz` CLI — and that shell tool comes from `buzz-dev-mcp` — the single slot is an either/or rather than a preference:

- slot = `buzz-dev-mcp` → the agent can post, but has no domain tools
- slot = a domain MCP server → the agent has domain tools, but no way to post

## Reproduction

Run `buzz-acp` with `BUZZ_ACP_MCP_COMMAND` pointing at any external stdio MCP server (`--agent-command buzz-agent`), add the agent to a channel, and ask it something its domain server can answer.

## Actual behavior

The agent answers internally and never posts. In my run it was offered 61 tools — 60 from the domain server plus `load_skill` — and **zero** shell or `buzz` tools. `default_heartbeat_prompt` (and the reply guidance around `crates/buzz-acp/src/lib.rs:3621`) instructs the agent to run `buzz messages send --reply-to <event-id>`, which it has no tool to do.

## Expected behavior

An operator can give an agent `buzz-dev-mcp` **and** their own MCP server(s) at once.

## Notes

`PersonaConfig.mcp_servers` is already a `Vec<McpServerConfig>`, and `buzz-persona` parses, merges (`merge_mcp_servers`), and validates it — `buzz pack describe` even prints the count. It is simply never plumbed to the runtime, which uses one `mcp_command` per agent record. So the multi-server intent seems to already be there in the persona layer.

The narrowest fix is at the harness: make `--mcp-command` repeatable and the env var comma-separated, matching the `agent_args` field directly above it in `Args`. Defaults unchanged; a single value behaves exactly as today. Happy to be redirected if you would rather solve this at the persona/runtime layer instead.

I have that change working with tests and will open a PR referencing this issue.

Searched open issues/PRs first: nothing found for multiple MCP servers. Closest adjacent ones are #2883 (keeping `BUZZ_PRIVATE_KEY` out of agent-controlled processes) and #2368 (adding an ACP runtime), neither overlapping.

Environment: `block/buzz` @ `c0d7b52`, macOS, `buzz-agent` runtime.

wolfyy970:
I found two related paths:

[...]

Issue #2883, opened 2026-07-25
security: keep BUZZ_PRIVATE_KEY out of agent-controlled processes


benthecarman:
## Summary

`buzz-acp` currently includes the agent identity's `BUZZ_PRIVATE_KEY` in the MCP server configuration. `buzz-dev-mcp` then intentionally inherits that variable into every process started by its `shell` tool so the bundled `buzz` CLI can authenticate.

This makes the raw signing key available to model-controlled commands. For example, a shell command can read `BUZZ_PRIVATE_KEY` directly from its environment. With network access, prompt injection could disclose the key and allow an attacker to impersonate the agent outside Buzz.

Relevant paths:

- `crates/buzz-acp/src/lib.rs` (`build_mcp_servers`) serializes `BUZZ_PRIVATE_KEY` into `mcpServers.env`.
- `crates/buzz-dev-mcp/src/shell.rs` intentionally inherits `BUZZ_PRIVATE_KEY` into shell children.
- `crates/buzz-dev-mcp/src/shim.rs` documents that inheritance for the bundled CLI.

## Desired security boundary

The ACP harness can retain the private key because it already uses the agent identity for relay subscriptions and publishing. The third-party agent runtime, model-visible tool configuration, arbitrary shell commands, and shell child processes should not receive the raw key.

A dedicated low-privilege agent identity remains useful defense in depth, but it should not be the only protection against key disclosure.

## Proposed direction

1. Keep the private key in `buzz-acp` or another harness-owned signer/publisher component.
2. Remove `BUZZ_PRIVATE_KEY` from the MCP configuration sent through `session/new` and from `buzz-dev-mcp` shell-child inheritance.
3. Expose a narrowly scoped MCP tool for publishing a Buzz message or reply. The tool should accept typed message fields, validate them, sign and submit internally, and return the relay write result.
4. Do not expose a generic `sign(payload)` tool or return raw signatures. That would hide the key bytes but create a signing oracle with nearly equivalent authority.
5. Give any MCP-side helper only an opaque, session-bound capability to the harness-owned publisher, not the private key. The capability should authorize only the typed operation and expire when the session ends.
6. Replace authenticated shell-based Buzz operations with typed, least-authority tools as needed. Unsupported CLI operations should fail without credentials rather than silently restoring key inheritance.

[...]

Issue #2368, opened 2026-07-22
feat(agents): add OpenCode as a first-class ACP runtime


leesalminen:
## Summary

Buzz Desktop's managed agent runtime catalog supports Goose, Claude Code, Codex, and Buzz Agent, but **OpenCode** is missing even though it already speaks [ACP](https://agentclientprotocol.com/) natively over stdio:

```bash
opencode acp
```

Users who want OpenCode-backed agents must configure a custom harness command instead of selecting OpenCode from Agent runtimes.

## Why

- OpenCode ships a stable ACP server (`opencode acp`) — no separate `*-acp` npm adapter is required (same class of integration as Goose's `goose acp` / Grok's native ACP path).
- Install path is well-known (`curl -fsSL https://opencode.ai/install | bash`, binary under `~/.opencode/bin` or `~/.local/bin`).
- Without catalog registration, Desktop cannot discover, install-hint, or process-sweep OpenCode the way it does other runtimes.
- OpenCode's provider-agnostic model catalog (Anthropic, OpenAI, OpenRouter, local Ollama/vLLM endpoints) surfaces through the existing ACP `configOptions` model discovery with no extra work.

## Proposed solution

Register OpenCode in `KNOWN_ACP_RUNTIMES` with:

| Field | Value |
|-------|--------|
| `id` | `opencode` |
| `label` | OpenCode |
| `commands` | `opencode` |
| Default args | `acp` |
| PATH discovery | include `~/.opencode/bin` |
| Config surface | `~/.config/opencode/opencode.json` (model, provider, MCP servers) |
| Auth hint | `opencode auth login` |

Also wire the config bridge (model/provider/MCP from `opencode.json`), `buzz-acp` default-arg normalization, docs (`buzz-acp` README / TESTING), and unit tests.

## Acceptance criteria

- [ ] **OpenCode** appears under Desktop → Agent runtimes when `opencode` is on PATH
- [ ] Creating/starting a managed agent with runtime=OpenCode spawns `opencode acp` via `buzz-acp`
- [ ] `buzz-acp models` with `BUZZ_ACP_AGENT_COMMAND=opencode` completes an ACP handshake and lists models
- [ ] Config surface reads model/provider/MCP servers from `~/.config/opencode/opencode.json`
- [ ] Docs describe the OpenCode path alongside Goose/Claude/Codex

## Related

- Pattern references: Grok Build ACP (#2347), Kimi Code runtime (#2365)
[...]

PR #2900, opened 2026-07-25
feat(acp): allow multiple MCP servers per agent

ArgussLabs:
## Summary

`--mcp-command` / `BUZZ_ACP_MCP_COMMAND` accepted a single binary, so an agent got exactly one MCP server. In practice that is an either/or rather than a preference: a Buzz agent acts in Buzz by shelling out to the `buzz` CLI, and that shell tool comes from `buzz-dev-mcp` — so pointing the slot at any domain-specific MCP server leaves the agent able to answer but unable to post.

This makes the flag repeatable and the env var comma-separated, matching the `agent_args` field directly above it in `Args`.

**Defaults are unchanged.** No value still means no MCP servers; a single value behaves exactly as before; the flag keeps its `--mcp-command` name so existing invocations and env values keep working.

- `Args.mcp_command: String` → `Args.mcp_commands: Vec<String>` with `value_delimiter = ','`
- Empty entries are filtered — `BUZZ_ACP_MCP_COMMAND=""` yields one empty entry through the delimiter
- `build_mcp_servers` maps over the list, injecting `BUZZ_RELAY_URL` / `BUZZ_PRIVATE_KEY` / `BUZZ_AUTH_TAG` into each server as before
- Server names are made unique. Two distinct binaries can share a file stem (`/a/bin/mcp` and `/b/bin/mcp`), and a duplicate name would let one shadow the other in the agent's tool namespace, so later collisions get a `-2` / `-3` suffix

### Related issue

#2899, opened just before this with the reproduction and the reasoning. Searched open issues/PRs first and found nothing for multiple MCP servers — closest adjacent are #2883 and #2368, neither overlapping.

Worth a maintainer's judgement: `PersonaConfig.mcp_servers` is already a `Vec` that `buzz-persona` parses, merges, and validates but never plumbs to the runtime, so you may prefer to solve this at the persona/runtime layer instead. This PR is the narrow harness-level fix; happy to redirect.

### Testing

- `cargo test -p buzz-acp --lib build_mcp_servers` → 9 passed. The four existing tests are updated to the `Vec` form, plus four new ones: two servers both reaching the agent, env injection per server, and file-stem collisions producing unique names.
- `cargo clippy -p buzz-acp --all-targets -- -D warnings` → clean.
- `cargo fmt -p buzz-acp` applied.
- `cargo test -p buzz-acp` → 599 passed, 2 failed: `acp::tests::idle_resets_on_stdout_activity` and `acp::tests::keepalive_resets_idle_past_deadline`. Both are **pre-existing on a clean tree** — I verified by stashing this change and rerunning — and both are timing-sensitive tests in `acp.rs`, which this PR does not touch. Likely flaky under load on my machine rather than anything to do with this change, but flagging in case it is news.
- Verified end to end against a local relay: with `buzz-dev-mcp` and a second MCP server both passed to `--mcp-command`, one agent turn called a tool on the second server and then used `shell` to post the result into the channel.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -137,8 +137,11 @@ RUST_LOG=buzz_relay=debug,buzz_db=debug,buzz_auth=debug,buzz_pubsub=debug,tower_
 # Goose default: "acp". Codex/Claude default: "" (empty).
 # BUZZ_ACP_AGENT_ARGS=acp
 
-# Binary for an optional MCP server sidecar (e.g. buzz-dev-mcp for buzz-agent).
-# BUZZ_ACP_MCP_COMMAND=
+# MCP server binaries to expose to the agent (e.g. buzz-dev-mcp for buzz-agent).
+# Comma-separated for more than one; --mcp-command may also be repeated.
+# Each server is spawned with BUZZ_RELAY_URL / BUZZ_PRIVATE_KEY / BUZZ_AUTH_TAG
+# injected. Servers are named after each binary's file stem.
+# BUZZ_ACP_MCP_COMMAND=buzz-dev-mcp,/opt/bin/my-domain-mcp
 
 # Number of parallel agent subprocesses (1–32).
 # BUZZ_ACP_AGENTS=1
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -258,8 +258,14 @@ pub struct CliArgs {
     )]
     pub agent_args: Vec<String>,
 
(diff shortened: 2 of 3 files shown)


wolfyy970:
I found this after implementing a different path in #4164.

This PR allows several MCP commands. #4164 adds a versioned config for servers that also need their own name, arguments, and environment. It keeps the existing privileged companion path unchanged and does not copy Buzz credentials into structured third-party servers.

The scope and acceptance criteria are in #4154.

PR #2907, opened 2026-07-25
fix(desktop): suspend idle poof audio context

### LINKED ISSUE #2868: [Bug] Idle desktop app burns ~4% CPU and pins the audio device awake
Linux client seems to chew up a lot of CPU doing nothing at all, here's a slop report that might help a bit. 

Attached report from opus:

# Idle desktop app burns ~4% CPU and pins the audio device awake: poof-sound `AudioContext` is never
  suspended

  ## Summary
  
  After the first UI "poof" click, `PoofBurstProvider`'s `AudioContext` stays in the
  `running` state for the remainder of the app's lifetime. `.suspend()` is never
  called on it — in fact `.suspend()` does not appear anywhere in `desktop/src`.

  On Linux, WebKitGTK routes WebAudio through GStreamer, so a running context means
  the audio graph is pumped continuously whether or not anything is playing. Measured
  on an otherwise idle app that has played no sound for ~22 minutes:

  - **3.91% of one CPU core**, sustained, in the WebAudio/GStreamer threads
  - **~0.5%** more in `pipewire-pulse`, charged to that process rather than to Buzz
  - the ALSA device is held in `RUNNING` state permanently, which blocks the audio
    codec and CPU from reaching deep idle — the battery cost on laptops is larger
    than the CPU number alone suggests

  The audio work is ~71% of the idle renderer's total CPU. All of it renders silence.

  ## Evidence

  Version 0.4.26 (`.deb`), Ubuntu 24.04, `libwebkit2gtk-4.1-0` 2.52.3, PipeWire.
  App open ~1h, idle, nothing playing.

  Per-thread CPU in the `WebKitWebProcess`, sampled from `/proc/<tid>/stat` over a
  22-minute window with no interaction:

  | Thread | % of one core |
  |---|---|
  | `webaudioSrc:src` | 2.53% |
  | `queue0:src` | 1.37% |
  | `webaudioSrcTask` | (included above) |
  | **audio subtotal** | **3.91%** |
  | `WebKitWebProces` (renderer main) | 2.04% |

  The corresponding PipeWire node, which persists for the app's whole lifetime:

  ```
  node 75
    application.process.id: <WebKitWebProcess>
    application.name:       buzz-desktop
    media.class:            Stream/Output/Audio
    media.name:             Playback Stream
    state:                  running
    format:                 F32LE 2 44100
  ```

  `pw-top` shows this node marked `+`, i.e. **Buzz is the driver of the audio
  graph**. It drags the ALSA sink out of its idle `C` state into `R` in lockstep:

  ```
  S   ID  QUANT   RATE    WAIT    BUSY  ERR FORMAT           NAME
  R   51   1024  48000 155.4us  22.2us   9  S32LE 2 48000    alsa_output.…analog-stereo
  R   75   1764  44100 119.0us  31.5us   0  F32LE 2 44100  + buzz-desktop
  ```

  Note the rate mismatch: the context runs at 44.1 kHz into a 48 kHz sink, so
  PipeWire resamples continuously as well. The sink has logged 9 xruns.

  ## Root cause

  All line numbers below are against `desktop/src/shared/ui/PoofBurstProvider.tsx`
  at `2f004159`.

  1. `PoofBurstProvider` wraps the app root (`desktop/src/main.tsx:80`), so its
     module-level singleton at line 20 lives for the whole session.
  2. Line 60 constructs `new AudioContext({ latencyHint: "interactive" })`.
     WebKitGTK creates contexts in the `suspended` state, and a *suspended* context
     opens no device — so this alone is free.
  3. On the first poof, `playPoofSound` (line 110) sees `state === "suspended"` and
     calls `resume()` at line 126. A click is a user gesture, so it succeeds and the
     context goes `running`.
  4. Nothing ever suspends it again. From that point the device stays open forever.

  This matches the observed timeline exactly: the app launched at 11:27 but the
  PipeWire node appeared at 11:29 — first click, not mount.

  `latencyHint: "interactive"` requests the smallest possible buffer, which
  maximises wakeup frequency — the worst setting for something left running idle.

  ### Secondary: the render graph grows by two nodes per poof

  In `playPoofSound`, the `AudioBufferSourceNode` and `GainNode` are connected to
  `destination` (lines 119–124) and never disconnected. Every click permanently adds
  two nodes to the graph, so per-quantum render cost climbs monotonically over a long
  session on top of the fixed cost above.

  ### Secondary: no explicit `sampleRate`

  Leaving `sampleRate` unset makes WebKitGTK default to 44.1 kHz, forcing the
  continuous resample described above. `desktop/public/worklet.js` already documents
  an assumption that contexts run at 48 kHz, so the two paths disagree.

  ## Not a memory leak

  Worth recording since it's the obvious next suspicion: the renderer sits at
  ~730 MB RSS / 527 MB private-dirty, but that is **stable, not growing**. Over the
  same 22-minute window RSS ranged 692–793 MB and drifted **+6 MB** net — GC
  sawtooth, not a leak.

  The high baseline looks like frontend bundle weight rather than a runtime fault.
  The shipped binary embeds 416 frontend assets including 86 Shiki language grammars
  and roughly 20 Shiki themes that each appear at **two different content hashes**

  `pw-top` shows this node marked `+`, i.e. **Buzz is the driver of the audio
  graph**. It drags the ALSA sink out of its idle `C` state into `R` in lockstep:

  ```
  S   ID  QUANT   RATE    WAIT    BUSY  ERR FORMAT           NAME
  R   51   1024  48000 155.4us  22.2us   9  S32LE 2 48000    alsa_output.…analog-stereo
  R   75   1764  44100 119.0us  31.5us   0  F32LE 2 44100  + buzz-desktop
  ```

  Note the rate mismatch: the context runs at 44.1 kHz into a 48 kHz sink, so
  PipeWire resamples continuously as well. The sink has logged 9 xruns.

  ## Root cause

  All line numbers below are against `desktop/src/shared/ui/PoofBurstProvider.tsx`
  at `2f004159`.

  1. `PoofBurstProvider` wraps the app root (`desktop/src/main.tsx:80`), so its
     module-level singleton at line 20 lives for the whole session.
  2. Line 60 constructs `new AudioContext({ latencyHint: "interactive" })`.
     WebKitGTK creates contexts in the `suspended` state, and a *suspended* context
     opens no device — so this alone is free.
  3. On the first poof, `playPoofSound` (line 110) sees `state === "suspended"` and
     calls `resume()` at line 126. A click is a user gesture, so it succeeds and the
     context goes `running`.
  4. Nothing ever suspends it again. From that point the device stays open forever.

  This matches the observed timeline exactly: the app launched at 11:27 but the
  PipeWire node appeared at 11:29 — first click, not mount.

  `latencyHint: "interactive"` requests the smallest possible buffer, which
  maximises wakeup frequency — the worst setting for something left running idle.

  ### Secondary: the render graph grows by two nodes per poof

  In `playPoofSound`, the `AudioBufferSourceNode` and `GainNode` are connected to
  `destination` (lines 119–124) and never disconnected. Every click permanently adds
  two nodes to the graph, so per-quantum render cost climbs monotonically over a long
  session on top of the fixed cost above.

  ### Secondary: no explicit `sampleRate`

  Leaving `sampleRate` unset makes WebKitGTK default to 44.1 kHz, forcing the
  continuous resample described above. `desktop/public/worklet.js` already documents
  an assumption that contexts run at 48 kHz, so the two paths disagree.

  ## Not a memory leak

  Worth recording since it's the obvious next suspicion: the renderer sits at
  ~730 MB RSS / 527 MB private-dirty, but that is **stable, not growing**. Over the
  same 22-minute window RSS ranged 692–793 MB and drifted **+6 MB** net — GC
  sawtooth, not a leak.

  The high baseline looks like frontend bundle weight rather than a runtime fault.
  The shipped binary embeds 416 frontend assets including 86 Shiki language grammars
  and roughly 20 Shiki themes that each appear at **two different content hashes**
  (`vitesse-light`, `tokyo-night`, `rose-pine`, `solarized-*`, …), which suggests two
  import paths causing Vite to emit duplicate copies. That is inferred from the asset
  manifest and has not been confirmed against the source — a separate issue if
  someone wants to chase the memory baseline.
  
  ## Suggested fix

  Keep the singleton context (the comment in `HuddleContext.tsx` about the ~6
  concurrent-context limit is a good reason not to create one per click), but stop
  leaving it running:

  1. **Suspend when idle.** After playback ends, `suspend()` the context on a short
     debounce (~1.5s) so rapid clicks don't thrash it, and `resume()` on the next
     poof. `playPoofSound` already handles the resume-from-suspended case, so this
     closes the loop the existing code anticipates.
  2. **Disconnect finished nodes** in an `ended` handler so the graph stays flat.
  3. **Set `sampleRate: 48000`** explicitly, matching `worklet.js`, to drop the
     resample.
  4. **Decode via `OfflineAudioContext`.** `decodeAudioData` only needs a
     `BaseAudioContext`; an offline one never opens a device, so the startup preload
     can't touch audio hardware at all.

  An alternative worth considering: the Rust side already links `rodio` + `cpal`, so
  routing UI chimes through a Tauri command would remove this WebAudio path
  entirely. Larger change, cleaner end state.

  ## Reproduce

  ```bash
  # 1. Open Buzz. Confirm no audio stream exists yet:
  pw-dump | jq -r '.[] | select(.info.props."media.class" == "Stream/Output/Audio")
                  | "\(.info.props."application.name") \(.info.state)"'

  # 2. Click anything that poofs — e.g. hover a sidebar card and click its × dismiss
  #    button (the trigger lives on sidebar-action-card.tsx and SelectedRecipientChip).

  # 3. Re-run the command above. A `buzz-desktop … running` stream now exists, and
  #    stays for as long as the app runs. Also visible in `pw-top` as the graph driver.

  # 4. Watch the cost with no further interaction:
  top -H -p "$(pgrep -f 'WebKitWebProcess' | head -1)"
  #    → webaudioSrc:src, webaudioSrcTask and queue0:src accumulate CPU indefinitely
  ```

  ## Notes

  - Evidence is Linux/WebKitGTK-specific, where WebAudio goes through GStreamer.
    The underlying defect — never calling `suspend()` — is platform-independent, but
    the cost profile on macOS/CoreAudio will differ and is unmeasured.
  - The other three `AudioContext` sites are correctly scoped and are **not**
    implicated: `huddle/lib/audioWorklet.ts` requests 48 kHz and closes at line 130,
    and both `HuddleContext.tsx` contexts are gated on `micConnected` and closed in
    effect cleanup.
  - Observed on `main` @ `2f004159`.

yinkev:
## Summary

- disconnect each poof sound's `AudioBufferSourceNode` and `GainNode` when playback ends, keeping the WebAudio graph flat
- suspend the singleton `AudioContext` 1.5 seconds after the final sound and cancel that idle suspension when another poof starts
- serialize replay against an already in-flight `suspend()` so rapid clicks resume cleanly instead of starting into a newly suspended context
- preserve the existing HTML audio fallback when graph creation, context resume, or source start fails, while rearming idle suspension after a failed graph setup

This stays focused on the measured lifecycle leak. Sample-rate policy, offline decoding, and a native-audio rewrite remain separate work.

### Related issue

Fixes #2868.

Closest existing PR: #2804 changes the poof asset codec for Linux decoding; this PR is orthogonal and fixes the context/node lifetime after playback.

### Testing

- `pnpm check`
- `pnpm typecheck`
- focused poof lifecycle tests — 8 passed
- full desktop unit suite — 3,512 passed
- desktop production build — passed

No visual changes.

### DIFF
--- a/desktop/src/shared/ui/PoofBurstProvider.tsx
+++ b/desktop/src/shared/ui/PoofBurstProvider.tsx
@@ -60,7 +60,10 @@ function getPoofOrigin(target: Element, pointer?: PoofPointer) {
 
 function getPoofAudioContext() {
   try {
-    poofAudioContext ??= new AudioContext({ latencyHint: "interactive" });
+    if (!poofAudioContext) {
+      poofAudioContext = new AudioContext({ latencyHint: "interactive" });
+      poofAudioPlayer.armIdleSuspend(poofAudioContext);
+    }
     return poofAudioContext;
   } catch {
     return null;
--- a/desktop/src/shared/ui/PoofBurstProvider.tsx
+++ b/desktop/src/shared/ui/PoofBurstProvider.tsx
@@ -1,5 +1,7 @@
 import React, { type CSSProperties, useEffect, useRef, useState } from "react";
 
+import { createPoofAudioPlayer } from "./poofAudioLifecycle";
+
 export const POOF_TRIGGER_CLASS = "buzz-poof-trigger";
 export const POOF_ORIGIN_CLASS = "buzz-poof-origin";
 export const POOF_POINTER_ORIGIN_CLASS = "buzz-poof-pointer-origin";
@@ -21,6 +23,7 @@ let poofAudioContext: AudioContext | null = null;
 let poofAudioBuffer: AudioBuffer | null = null;
 let poofAudioBufferPromise: Promise<AudioBuffer | null> | null = null;
 let lastPointerDownTrigger: Element | null = null;
+const poofAudioPlayer = createPoofAudioPlayer();
 
 type PoofBurst = {
   id: number;
@@ -115,24 +118,7 @@ function playPoofSound() {
     return;
   }
 
-  try {
-    const source = audioContext.createBufferSource();
-    const gain = audioContext.createGain();
-    source.buffer = poofAudioBuffer;
-    gain.gain.value = 0.34;
-    source.connect(gain);
-    gain.connect(audioContext.destination);
-    if (audioContext.state === "suspended") {
-      void audioContext.resume().then(
-        () => source.start(),
-        () => playFallbackPoofSound(),
-      );
-    } else {
-      source.start();
-    }
-  } catch {
-    playFallbackPoofSound();
-  }
+  poofAudioPlayer.play(audioContext, poofAudioBuffer, playFallbackPoofSound);
 }
 
 export function PoofBurstProvider({ children }: { children: React.ReactNode }) {
--- a/desktop/src/shared/ui/poofAudioLifecycle.test.mjs
+++ b/desktop/src/shared/ui/poofAudioLifecycle.test.mjs
@@ -1,0 +1,241 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { createPoofAudioPlayer } from "./poofAudioLifecycle.ts";
+
+function createScheduler() {
+  let nextId = 1;
+  const callbacks = new Map();
+
+  return {
+    clearTimeout(id) {
+      callbacks.delete(id);
+    },
+    pendingCount() {
+      return callbacks.size;
+    },
+    runAll() {
+      const pending = [...callbacks.values()];
+      callbacks.clear();
+      for (const callback of pending) callback();
+    },
+    setTimeout(callback) {
+      const id = nextId;
+      nextId += 1;
+      callbacks.set(id, callback);
+      return id;
+    },
+  };
+}
+
+function createAudioHarness({
+  deferSuspend = false,
+  initialState = "running",
+  resumeError = null,
+  startError = null,
+} = {}) {
+  const calls = {
+    fallback: 0,
+    resume: 0,
+    suspend: 0,
+  };
+  const sources = [];
+  const gains = [];
+  let state = initialState;
+  let resolveDeferredSuspend = null;
+  const deferredSuspend = deferSuspend
+    ? new Promise((resolve) => {
+        resolveDeferredSuspend = () => {
+          state = "suspended";
+          resolve();
+        };
+      })
+    : null;
+
+  const context = {
+    createBufferSource() {
+      const source = {
+        buffer: null,
+        connectedTo: null,
+        disconnectCalls: 0,
+        onended: null,
+        startCalls: 0,
+        startError,
+        connect(target) {
+          this.connectedTo = target;
+        },
+        disconnect() {
+          this.disconnectCalls += 1;
+        },
+        finish() {
+          this.onended?.();
+        },
+        start() {
+          this.startCalls += 1;
+          if (this.startError) throw this.startError;
+        },
+      };
+      sources.push(source);
+      return source;
+    },
+    createGain() {
+      const gain = {
+        connectedTo: null,
+        disconnectCalls: 0,
+        gain: { value: 0 },
+        connect(target) {
+          this.connectedTo = target;
+        },
+        disconnect() {
+          this.disconnectCalls += 1;
+        },
+      };
+      gains.push(gain);
+      return gain;
+    },
+    destination: {},
+    resume() {
+      calls.resume += 1;
+      if (resumeError) return Promise.reject(resumeError);
+      state = "running";
+      return Promise.resolve();
+    },
+    get state() {
+      return state;
+    },
+    suspend() {
+      calls.suspend += 1;
+      if (deferredSuspend) return deferredSuspend;
+      state = "suspended";
+      return Promise.resolve();
+    },
+  };
+
+  return {
+    buffer: {},
+    calls,
+    context,
+    fallback() {
+      calls.fallback += 1;
+    },
+    gains,
+    resolveSuspend() {
+      resolveDeferredSuspend?.();
+    },
+    sources,
+  };
+}
+
+function createPlayer(scheduler) {
+  return createPoofAudioPlayer({
+    clearTimeout: scheduler.clearTimeout,
+    idleDelayMs: 1_500,
+    setTimeout: scheduler.setTimeout,
+  });
+}
+
+test("disconnects finished source and gain nodes, then suspends when idle", async () => {
+  const scheduler = createScheduler();
+  const audio = createAudioHarness();
+  const player = createPlayer(scheduler);
+
+  player.play(audio.context, audio.buffer, audio.fallback);
+  audio.sources[0].finish();
+
+  assert.equal(audio.sources[0].disconnectCalls, 1);
+  assert.equal(audio.gains[0].disconnectCalls, 1);
+  assert.equal(scheduler.pendingCount(), 1);
+  assert.equal(audio.calls.suspend, 0);
+
+  scheduler.runAll();
+  await Promise.resolve();
+  assert.equal(audio.calls.suspend, 1);
+});
+
+test("waits for every overlapping playback before scheduling suspension", async () => {
+  const scheduler = createScheduler();
+  const audio = createAudioHarness();
+  const player = createPlayer(scheduler);
+
+  player.play(audio.context, audio.buffer, audio.fallback);
+  player.play(audio.context, audio.buffer, audio.fallback);
+
+  audio.sources[0].finish();
+  assert.equal(scheduler.pendingCount(), 0);
+
+  audio.sources[1].finish();
+  assert.equal(scheduler.pendingCount(), 1);
+  scheduler.runAll();
+  await Promise.resolve();
+  assert.equal(audio.calls.suspend, 1);
+});
+
+test("new playback cancels the pending idle suspension", async () => {
+  const scheduler = createScheduler();
+  const audio = createAudioHarness();
+  const player = createPlayer(scheduler);
+
+  player.play(audio.context, audio.buffer, audio.fallback);
+  audio.sources[0].finish();
+  assert.equal(scheduler.pendingCount(), 1);
+
+  player.play(audio.context, audio.buffer, audio.fallback);
+  assert.equal(scheduler.pendingCount(), 0);
+
+  audio.sources[1].finish();
+  scheduler.runAll();
+  await Promise.resolve();
+  assert.equal(audio.calls.suspend, 1);
+});
+
+test("stale suspend completion resumes a newer active playback", async () => {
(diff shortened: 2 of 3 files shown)


yinkev:
@codex review

yinkev:
@codex review

FabianHertwig:
Thanks for picking this up — the node-disconnect plus debounced-suspend logic reads right for the lifecycle the issue describes. Flagging one path I don't think it covers, based on macOS measurements I've just posted on #2868.

`scheduleSuspend()` is only reachable from `play()` — the graph-creation `catch` and `cleanup()` when the last playback ends. That holds if the context can only reach `running` via `playPoofSound()` → `resume()`.

On macOS it can't: when the embedder allows autoplay (the webview default Buzz takes), a freshly constructed `AudioContext` **auto-resumes to `running` on its own**, with no user gesture and nothing played. I isolated it in a minimal `WKWebView` host — A/B table and log trace are in #2868. The short version from a real launch:

```
13:11:25.996  RemoteAudioDestination::RemoteAudioDestination(…)   ← the AudioContext
13:11:26.004  coreaudiod  Take: …BuiltInSpeakerDevice.context.preventuseridlesleep on behalf of 79092
13:11:26.035  RemoteAudioDestination::start(…)                    ← went running by itself
```

Since `PoofBurstProvider`'s mount effect calls `loadPoofAudioBuffer()` → `getPoofAudioContext()` at startup just to decode the asset, that means the context is constructed on every launch, goes `running` moments later, `play()` is never called, `scheduleSuspend()` never runs, and the output device stays pinned for the whole session. Measured here: no idle sleep for 12.5 h and battery 38% → 1% overnight, without a single poof.

One wrinkle if you do arm the suspension at construction: a plain `scheduleSuspend(context)` there would race the auto-resume — the timer fires at 1.5 s, sees `state === "suspended"`, and returns; the context then flips to `running` at ~2 s and stays. Keying on `statechange` sidesteps the ordering:

```ts
// in createPoofAudioPlayer(), alongside play()
function armIdleSuspend(context: AudioContext) {
  // WebKit returns a suspended context but auto-resumes it when the embedder
  // allows autoplay, so an idle context can reach "running" with no playback.
  // Re-arm on every transition instead of assuming play() is the only path.
  context.addEventListener("statechange", () => {
    if (context.state === "running" && activePlaybacks === 0) {
      scheduleSuspend(context);
    }
  });
  if (context.state === "running") scheduleSuspend(context);
}

return { armIdleSuspend, play };
```

called once where the singleton is created:

```ts
function getPoofAudioContext() {
  try {
    if (!poofAudioContext) {
      poofAudioContext = new AudioContext({ latencyHint: "interactive" });
      poofAudioPlayer.armIdleSuspend(poofAudioContext);
    }
    return poofAudioContext;
  } catch {
    return null;
  }
}
```

This composes with what you already have rather than duplicating it: `play()` calls `cancelPendingSuspend()` first, and `activePlaybacks` is incremented before `resume()`, so the `statechange` from a real poof can't schedule a suspension mid-playback; `cleanup()` still owns the post-playback case; and `suspend()` itself fires one more `statechange` that no-ops. The first poof after startup finds a suspended context, which `resumeThenStart()` already handles.

The alternative is to not construct a hardware context at startup at all — decode via `OfflineAudioContext`, which never binds an output device, and create the real context on the first poof. That's the offline-decoding work you scoped out of this PR, so the `statechange` arming above is probably the smaller move here.

Happy to test either against my setup — the assertion is directly observable via `pmset -g assertions | grep -A2 BuiltInSpeakerDevice`, so it's a clean pass/fail.

yinkev:
Thanks for the concrete WebKit trace. Addressed in the latest commit: the singleton now arms an idle-suspend statechange listener when the AudioContext is created, so an autoplay-enabled context that transitions to running without any poof is suspended after the debounce. Active playback still cancels the timer before resume, and the listener only arms while there are zero active playbacks. Added a focused suspended → auto-running regression; the poof lifecycle suite passes 8/8.

yinkev:
**Superseded after fresh GitHub recalculation — 2026-08-05:** this PR is currently cleanly mergeable into `main`, and CI on head `49c16271` completed successfully. No rebase or code change is requested; the branch should remain unchanged unless a maintainer raises a new finding. The current GitHub installation cannot update the fork branch, and `maintainer_can_modify` remains false.

eeSeeGee:
🤖 Fresh confirmation for this PR: a **third platform report** landed on #2868 — macOS 26.6.1 / arm64 / **Buzz 0.5.10**, with the pinned device being a *Bluetooth headset*, so A2DP never idles. Full capture: https://github.com/block/buzz/issues/2868#issuecomment-5271632652

Two things from that trace that are relevant to reviewing this change specifically:

1. **Your `armIdleSuspend()` design is the part that matters on macOS.** The device is pinned 7 ms after `RemoteAudioDestination` is constructed and 126 ms *before* it `start()`s — i.e. from app mount, with no poof ever played. A suspend debounced only off playback completion would never fire. Calling `armIdleSuspend()` at construction with the `statechange` listener and your comment *"WebKit may start an autoplay-enabled context without any playback"* is exactly right for this path.
2. `PoofBurstProvider.tsx:60` is the **only** `AudioContext` in `desktop/src` with no teardown — the three huddle contexts (`HuddleContext.tsx:790`, `audioWorklet.ts:56`, `useHuddlePttState.ts:41`) all `close()`. That makes this PR the whole fix for the device-pinning symptom rather than one of several needed changes.

Scope note for whoever reviews: this PR keeps the `HTMLAudioElement` fallback, which is fine for #2868, but that element is the *separate* media-key/Now Playing hijack that #4916 targets. The reporter hit both symptoms and needs both PRs.

**Heads-up on landing order:** #4916 also edits `PoofBurstProvider.tsx` and *deletes* the HTML fallback this PR preserves. Whoever merges second has a behavioral reconcile, not just a text conflict. Worth coordinating with @nathan-thillairajah.

This has been open 18 days with no review. Would be good to get eyes on it — it's a battery/power regression, not just idle CPU.

yinkev:
Coordination with #4916 is resolved: **this PR should land first, and I’ll keep its head stable.** Afterward #4916 will rebase and retain this PR’s construction-time idle suspension, playback/suspend race handling, and node cleanup while removing the Poof HTML preload/fallback. No code change is requested here for that integration; the behavioral reconcile belongs in the second PR. #5222 is the narrower superseded notification implementation.

PR #2917, opened 2026-07-26
fix(desktop): hide message deletion notices

john-kyuhae-lee:
## Summary

- remove `message_deleted` system notices from channel timelines
- exclude hidden notices from visible-row pagination counts
- skip unnecessary author-profile loading for hidden notices
- preserve deletion enforcement and moderation audit events

These notices currently render as repeated "Removed by community moderators" rows after the underlying messages have already disappeared. They add visual noise without restoring any conversational context.

## Testing

- `pnpm --dir desktop test`
- `pnpm --dir desktop check`
- `pnpm --dir desktop build`

## Existing issue / PR

None found.

### DIFF
--- a/desktop/src/features/messages/lib/formatTimelineMessages.test.mjs
+++ b/desktop/src/features/messages/lib/formatTimelineMessages.test.mjs
@@ -88,6 +88,19 @@ function huddleStarted(overrides = {}) {
   };
 }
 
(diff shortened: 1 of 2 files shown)


john-kyuhae-lee:
Closing for now while we evaluate whether the existing client settings can cover this preference.

PR #2921, opened 2026-07-26
feat(desktop): add Qoder ACP runtime

xinyuan0801:
## Summary

- add Qoder as a first-class desktop ACP runtime using its native `qodercli --acp` server
- provide platform install commands, login guidance, runtime configuration metadata, model discovery, managed-process cleanup, and `.qoder/skills` integration
- include authenticated Qoder installations in onboarding, readiness detection, and preferred-runtime selection
- expose provider ownership through the Rust runtime catalog so switching to Qoder clears stale Buzz-managed provider state and displays `Qoder (locked)` correctly
- define typed authentication-probe contracts and validate Qoder's `logged_in` JSON boolean strictly, failing closed on malformed, missing, or incorrectly typed responses

## Why

Buzz already supports Goose, Codex, and Claude Code, but Qoder users could not select or launch Qoder from the managed-agent runtime catalog. Qoder implements ACP directly, so Buzz can integrate it without an adapter.

Qoder's unauthenticated `status` command exits successfully and reports the real state through `{"logged_in": false}`. Exit status alone therefore cannot establish authentication. The catalog now declares whether a probe uses exit status or a required JSON boolean, preserving the existing Claude/Codex behavior while making Qoder fail closed.

The runtime catalog is also the single source of truth for provider capability. This prevents frontend runtime-ID guesses, removes stale providers when moving from a provider-backed runtime to Qoder, and keeps locked-provider labels runtime-specific.

## User impact

Users can install or select Qoder in Buzz, authenticate with `qodercli login`, choose it during onboarding, save it as the preferred runtime, launch its native ACP server, discover and switch Qoder models through ACP, and use Buzz skills from Qoder workspaces.

Existing agents switched to Qoder no longer retain an incompatible provider from their previous runtime. Invalid Qoder status output is shown conservatively as not ready instead of being accepted as authenticated.

## Validation

- `just ci` — passed
- desktop unit suite — 3,519 passed
- desktop Tauri suite — 1,643 unit tests plus 3 mixer diagnostics passed
- onboarding Playwright suite — 20 passed
- mobile Flutter suite — 585 passed, 1 skipped
- `cargo test -p buzz-acp` — 599 library tests and 9 integration tests passed
- `git diff --check` — passed
- manual Qoder CLI 1.1.5 ACP initialization returned `agentInfo.name = "qoder-cli"`; unauthenticated session creation correctly requested authentication

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -130,11 +130,11 @@ RUST_LOG=buzz_relay=debug,buzz_db=debug,buzz_auth=debug,buzz_pubsub=debug,tower_
 # BUZZ_RELAY_URL=ws://localhost:3000
 
 # ── Agent subprocess ─────────────────────────────────────────────────────────
-# Binary to spawn as the AI agent (e.g. "goose", "codex-acp", "claude-code").
+# Binary to spawn as the AI agent (e.g. "goose", "codex-acp", "claude-code", "qodercli").
 # BUZZ_ACP_AGENT_COMMAND=goose
 
 # Comma-separated arguments passed to the agent binary.
-# Goose default: "acp". Codex/Claude default: "" (empty).
+# Goose default: "acp". Qoder default: "--acp". Codex/Claude default: "" (empty).
 # BUZZ_ACP_AGENT_ARGS=acp
 
 # Binary for an optional MCP server sidecar (e.g. buzz-dev-mcp for buzz-agent).
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -648,7 +648,7 @@ Standalone binary that bridges Buzz relay events to AI agents via the [Agent Com
 **Architecture:**
 
 ```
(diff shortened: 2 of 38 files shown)


wpfleger96:
🤖 hey @xinyuan0801 — thanks for building this, the fail-closed auth-probe design in here is genuinely careful work. PRs like this one are exactly what convinced us to solve harness support generically instead of one runtime at a time.

That work landed in #2773 and ships in v0.5.0: any ACP-over-stdio binary can be registered from Settings, so `qodercli --acp` works today as a custom harness with no code change. There's no Qoder preset yet, but presets are now a small data entry rather than per-harness Rust — a tiny PR away if you want Qoder to be one click.

Closing as superseded. The typed auth-probe contract idea (exit-status vs required JSON boolean) is bigger than any one harness — if that still interests you, an issue proposing it for the generic readiness path would be worth filing.

PR #2511, opened 2026-07-23
fix: agent exit signal, local port overrides, A3 probe hint

Chessing234:
## Summary
- Runtime poller records `last_error` when a harness dies so Agents no longer look merely idle (`#2453`)
- Compose / setup honor `PGPORT` and `BUZZ_REDIS_HOST_PORT` alongside Homebrew services (`#2479`)
- Clearer A3 probe failure text + Helm `git.conformanceProbe` for non-CAS stores like GCS (`#2470`)

## Test plan
- [ ] Kill a running `buzz-acp` child; Agents card shows an error / restart affordance
- [ ] With Homebrew on `:5432`/`:6379`, set alternate ports in `.env` and run `just setup`
- [ ] Helm/env: `git.conformanceProbe: false` sets `BUZZ_GIT_CONFORMANCE_PROBE=false`

Closes #2453
Closes #2479
Closes #2470

Made with [Cursor](https://cursor.com)

Issue #2479, opened 2026-07-23
dev setup: allow Postgres and Redis host ports to be configured


varindersingh83:
## Summary

`just setup` requires host ports `5432` and `6379` for its Docker-managed Postgres and Redis services. This blocks local setup when standard Homebrew Postgres or Redis services are already running, even though Buzz itself supports custom `DATABASE_URL` and `REDIS_URL` values.

## Environment

- Buzz `main` at `acfbb1bb`
- macOS
- Docker Desktop 29.3.1 / Docker Compose 5.1.1
- Homebrew `postgresql@15` on `localhost:5432`
- Homebrew `redis` on `localhost:6379`

## Reproduction

1. Start local Postgres and Redis services on their standard ports.
2. Clone Buzz and activate Hermit.
3. Run:

   ```bash
   just setup
   ```

## Actual behavior

Setup aborts before Compose starts:

```text
[dev-setup] Local Redis is already listening on port 6379 (...).
Stop it before running setup: brew services stop redis
```

Postgres would also conflict because `docker-compose.yml` publishes `5432:5432`.

The fixed assumptions are currently spread across:

- `docker-compose.yml`: `5432:5432` and `6379:6379`
- `scripts/dev-setup.sh`: collision detection always checks `6379`
- `.env.example`: URLs and `PGPORT` use the standard ports

Stopping existing database services can disrupt unrelated local applications.

## Expected behavior

Buzz should preserve the current defaults while allowing its Docker services to bind configurable host ports, for example:

```dotenv
DATABASE_URL=postgres://buzz:buzz_dev@localhost:55432/buzz
PGPORT=55432
BUZZ_REDIS_HOST_PORT=56379
REDIS_URL=redis://localhost:56379
```

Then `just setup`, `just relay`, and `just dev` should use those values without requiring existing services to stop.

## Proposed direction

- Publish Postgres as `${PGPORT:-5432}:5432`.
- Add a Redis host-port setting, e.g. `BUZZ_REDIS_HOST_PORT`, and publish `${BUZZ_REDIS_HOST_PORT:-6379}:6379`.
- Make the local Redis collision check inspect the configured host port.
- Document that each host-port value and its corresponding connection URL must be changed together.
- Consider making integration-test scripts honor the configured URLs/ports as a follow-up.

## Local verification

I tested the proposed approach with Postgres on `55432` and Redis on `56379`:

- `just setup` completed.
- Database migrations and local community seeding completed.
- Both Buzz containers reported healthy.
[...]

PR #2925, opened 2026-07-26
chore(docker): make Postgres host port configurable via POSTGRES_PORT

1WorldCapture:
### What changed?

Made the Postgres host port in `docker-compose.yml` overridable via the `POSTGRES_PORT` environment variable (defaults to `5432`).

```yaml
ports:
  - "${POSTGRES_PORT:-5432}:5432"
```

### Why?

The Postgres service hardcoded host port `5432`. When a developer already runs a local Postgres on `5432`, `docker compose up` fails with a port conflict. Letting the host port be overridden via `POSTGRES_PORT` (e.g. `POSTGRES_PORT=5433` in `.env`) avoids the collision without touching the in-container port or the `DATABASE_URL`.

The default keeps current behavior intact, so existing setups are unaffected.

### How is it tested?

- `docker compose config` resolves the port mapping to `5432:5432` by default and honors `POSTGRES_PORT` when set.
- No code or CI changes — compose file only.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,7 @@ services:
       POSTGRES_DB: buzz
       PGDATA: /var/lib/postgresql/data
     ports:
-      - "5432:5432"
+      - "${POSTGRES_PORT:-5432}:5432"
     volumes:
       - postgres-data:/var/lib/postgresql/data
     networks:


1WorldCapture:
Closing in favor of #2486, which addresses #2479 more completely:

- Covers **both** Postgres and Redis (the issue asks for both).
- Uses `PGPORT` / `BUZZ_REDIS_HOST_PORT`, matching the existing `.env.example` (`.env.example` already ships a `PGPORT=5432` line) and the naming proposed in #2479. This PR's `POSTGRES_PORT` would have been a third, disconnected name — setting `PGPORT` in `.env` would not affect the compose port mapping.
- Updates `scripts/dev-setup.sh` to validate `PGPORT` vs `DATABASE_URL` consistency and to detect conflicts on the *configured* ports rather than hardcoded 5432/6379.

Better to consolidate behind #2486 than fragment the fix across overlapping PRs (#2485 closed → #2486, #2511, and this one). Will review #2486 instead.

PR #2927, opened 2026-07-26
Fix mobile DM recipient routing

plantbob0101:
## Summary

- add every one-to-one DM participant as a `p` tag on mobile sends
- apply the same routing to direct messages and thread replies
- retain explicit mention routing for shared channels
- add regression coverage for no-mention DMs, normalization, and deduplication

## Why

Mobile DMs without an explicit `@mention` were emitted with only the channel
`h` tag. Mention-gated agents therefore never received those events, even
though the messages appeared in the DM history. Desktop already addresses DM
participants automatically.

## Validation

- `just ci`
- full mobile suite: 588 passed, 1 skipped

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page.dart
+++ b/mobile/lib/features/channels/channel_detail_page.dart
@@ -326,6 +326,9 @@ class ChannelDetailPage extends HookConsumerWidget {
                         channelId: channel.id,
                         content: content,
                         mentionPubkeys: mentionPubkeys,
+                        recipientPubkeys: resolvedChannel.isDm
+                            ? resolvedChannel.participantPubkeys
+                            : const [],
                         mediaTags: mediaTags,
                       ),
             )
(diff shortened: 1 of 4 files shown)


svndco:
## Independent review

I checked this draft out at `59792eb630337d337d40591dddddf241a5808615` and ran the full mobile package gates:

- `flutter test`: **588 passed, 1 skipped**
- `just mobile-check`: formatting, analysis, and file-size checks passed

The recipient normalization here is stronger than #2863: it emits canonical lowercase pubkeys, filters empty values, excludes self, and deduplicates recipients with explicit mentions. However, #2863 predates this PR, directly closes #2835, is non-draft, and has broader regression coverage for threaded DM/media tag ordering.

Recommendation: move this PR's normalization loop into #2863, keep #2863's tests, and close this as the duplicate rather than merging two overlapping send-path changes.

PR #2930, opened 2026-07-26
chore(desktop): restore runtime file-size ratchet

yinkev:
## Summary

- restore the Desktop file-size gate on current `main` after concurrent runtime changes moved `managed_agents/runtime.rs` above its existing 2,216-line ratchet
- keep the four recognized Desktop process names unchanged, including Linux's truncated `buzz-desktop.bi` AppImage process name
- compact only the constant and its explanatory comment; runtime behavior is unchanged

Current `main` reproduces the failure before this change:

```text
src-tauri/src/managed_agents/runtime.rs: 2220 lines (limit 2216)
```

After the compaction, the file is below the existing ratchet without raising or bypassing the limit.

### Testing

- `node desktop/scripts/check-file-sizes.mjs`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml appimage_binary_matches_truncated_linux_comm_name --lib`
- `RUST_TEST_THREADS=1 just ci`

The serial Rust test setting avoids the existing parallel `relay_admission` test mutex deadlock; no tests or CI stages were skipped.

### DIFF
--- a/desktop/src-tauri/src/managed_agents/runtime.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime.rs
@@ -853,16 +853,9 @@ pub(crate) fn collect_same_instance_orphans(
     std::collections::HashSet::new()
 }
 
-/// Binary names for the Buzz desktop/Tauri process. Used by dead-instance
-/// detection to confirm the owning desktop is still alive.
-const DESKTOP_BINARY_NAMES: &[&str] = &[
-    "Buzz",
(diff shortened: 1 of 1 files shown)


yinkev:
@codex review

yinkev:
Current main split runtime.rs into smaller modules, and runtime.rs is now 982 lines. The file size check passes without this compaction, so the PR is no longer needed. Closing as superseded by the runtime split.

PR #2938, opened 2026-07-26
fix(db): require explicit opt-in for destructive migration tests

### LINKED ISSUE #2916: buzz-db: bare `--ignored` test sweep drops the Buzz Desktop dev schema via hardcoded localhost:5432 fallback
## Summary

`cargo test -p buzz-db --lib -- --ignored` — with no `BUZZ_TEST_DATABASE_URL` or `DATABASE_URL` set — runs `DROP SCHEMA IF EXISTS public CASCADE` against a **hardcoded** `postgres://buzz:buzz_dev@localhost:5432/buzz`. That is the Buzz Desktop development database on a default local setup.

This happened. On 2026-07-25 I ran that command twice while reviewing an unrelated security fix and destroyed the shared dev schema for about three minutes: window `23:59:11.170Z -> 00:02:02.930Z`. It self-restored only because the test binary kept running to the rebuild. A `Ctrl-C`, a tool timeout, or a panic between the `DROP SCHEMA` and the following `CREATE SCHEMA` + re-migration leaves the developer's database empty with no automatic recovery.

**How that window is established, and a caveat for anyone reproducing it.** The back edge comes from a state artifact rather than a log: every one of the 24 rows in `_sqlx_migrations` on the live database has an `installed_on` inside a 283ms burst starting `2026-07-25 20:02:02.647 EDT`, i.e. the schema now serving traffic was built then. The catalog agrees exhaustively: `public` is OID `3238914`, every table in it falls in `3238915-3240452`, and zero relations predate the namespace, so nothing survived the drop.

The caveat is worth more than the number. This server runs `log_statement = none` with `log_min_error_statement = error`, so **the Postgres log contains only statements that failed.** Every `DROP SCHEMA IF EXISTS public CASCADE` line in it is there because it errored; the successful drop that actually destroyed the schema was never logged, and neither were the successful `CREATE SCHEMA` or any successful migration. Three attempts to bound this incident from that log produced three different answers before we checked `log_statement`. If you are timing an incident on a default Homebrew Postgres, reach for state that carries its own clock (`_sqlx_migrations.installed_on`, `pg_class` OID survivorship) before reaching for the log.

## Mechanism

```rust
// crates/buzz-db/src/migration.rs:103
const TEST_DB_URL: &str = "postgres://buzz:buzz_dev@localhost:5432/buzz";

// crates/buzz-db/src/migration.rs:1027-1031
async fn connect_test_pool() -> PgPool {
    let database_url = std::env::var("BUZZ_TEST_DATABASE_URL")
        .or_else(|_| std::env::var("DATABASE_URL"))
        .unwrap_or_else(|_| TEST_DB_URL.to_owned());   // <-- silent fallback to the dev DB
    ...
}

// crates/buzz-db/src/migration.rs:1037
async fn reset_public_schema(pool: &PgPool) {
    sqlx::query("DROP SCHEMA IF EXISTS public CASCADE") ...
```

Three `#[ignore = "requires Postgres"]` tests call `reset_public_schema`:

- `migration.rs:1061` — `pre_0007_ambiguous_nip_rs_data_blocks_without_mutation_and_allows_retry`
- `migration.rs:1131` — `populated_upgrade_preserves_search_policy_except_for_push_leases`
- `migration.rs:1191` — `run_migrations_applies_consolidated_initial_schema_on_fresh_database`

`#[ignore]` is not a safeguard here: `--ignored` is exactly the flag you reach for when you want "the Postgres-backed tests," and nothing in that command names a database.

Note the contrast with `scripts/start-isolated-test-relay.sh:96`, which spells it `DROP SCHEMA public CASCADE` and pipes it through `docker compose exec postgres`, so it can only ever reach the container on `:5471`. The `IF EXISTS` spelling appears exactly once in the tree, which is how the incident was traced.

## Second, independent problem: the suite is already broken in parallel

Because these three tests drop the schema out from under every other test in the same binary, `cargo test -p buzz-db --lib -- --ignored` is nondeterministic:

- **10 failures on clean `main`** at the same commit, varying run to run
- passes 127/128 with `--test-threads=1` (the one failure is `product_feedback` wanting an explicit `BUZZ_TEST_DATABASE_URL`)

Nobody has noticed because `scripts/run-tests.sh:93` only ever runs `cargo test -p buzz-db --lib` *without* `--ignored`, and the Postgres-backed tests are otherwise only run under a filter. So the documented CI path never exercises the sweep, and the sweep is the dangerous command.

## Proposed fix

Two changes, both small:

1. **Delete the fallback.** `connect_test_pool` should `.expect("set BUZZ_TEST_DATABASE_URL to an isolated test database")` rather than default to anything. No hardcoded URL constant. A test that needs a database it can destroy should require the developer to name it.
2. **Make `reset_public_schema` unreachable from a suite-wide sweep.** `#[ignore]` does not distinguish "needs Postgres" from "will destroy Postgres". Gate the three destructive tests behind something `--ignored` alone cannot satisfy — e.g. a distinct required env var (`BUZZ_TEST_DESTRUCTIVE_DB=1`), a separate test target, or a `#[cfg(feature = ...)]` — so that the reasonable command stays safe and the destructive one has to be asked for by name.

Optionally (3) have the destructive tests refuse a database whose URL they didn't receive explicitly *and* which contains unexpected data — but (1) and (2) are the load-bearing ones, and (3) is a heuristic where those two are structural.

Fixing the parallel nondeterminism follows for free from (2): once the schema-dropping tests can't be swept into the same run as everything else, the remaining Postgres tests can run in parallel honestly.

## Not urgent as a security issue, urgent as a data-loss issue

There is no remote attacker here. The exposure is: any contributor typing a plausible test command loses their local Buzz Desktop data, with no warning and no confirmation prompt. "Don't type the obvious command" is not a safeguard.

yinkev:
## Summary

- remove the silent fallback from schema-destructive migration tests to the default Buzz Desktop development database
- require both `BUZZ_TEST_DESTRUCTIVE_DB=1` and an explicit non-empty `BUZZ_TEST_DATABASE_URL` before any test can drop the `public` schema
- rename the connection helper to make its destructive contract explicit and add a pure regression for every missing or invalid gate combination

`#[ignore]` was not a sufficient boundary because `cargo test -p buzz-db --lib -- --ignored` is a natural way to request the Postgres-backed tests. Before this change, that command could resolve to `postgres://buzz:buzz_dev@localhost:5432/buzz` and execute `DROP SCHEMA IF EXISTS public CASCADE` against a developer's live Desktop database.

## Related issue

Fixes #2916.

No competing implementation was found before opening this PR.

## Validation

Fresh verification after rebasing onto current `main`:

- `cargo test -p buzz-db --lib` — **95 passed, 153 intentionally ignored**
- `cargo clippy -p buzz-db --all-targets -- -D warnings` — passed
- `cargo fmt --check` — passed
- disposable PostgreSQL database:
  - migrated with `cargo run -p buzz-admin -- migrate`
  - ran the three schema-reset tests with both required environment variables and `--test-threads=1` — **3 passed**
  - dropped the disposable database and verified no database remained
- `git diff --check upstream/main...HEAD` — passed

### DIFF
--- a/crates/buzz-db/src/migration.rs
+++ b/crates/buzz-db/src/migration.rs
@@ -100,7 +100,8 @@ mod tests {
     use super::*;
     use std::collections::BTreeSet;
 
-    const TEST_DB_URL: &str = "postgres://buzz:buzz_dev@localhost:5432/buzz"; // sadscan:disable np.postgres.1
+    const DESTRUCTIVE_TEST_OPT_IN_ENV: &str = "BUZZ_TEST_DESTRUCTIVE_DB";
+    const TEST_DATABASE_URL_ENV: &str = "BUZZ_TEST_DATABASE_URL";
 
     #[derive(Debug, Clone, Copy, PartialEq, Eq)]
     enum ConstraintKind {
@@ -115,6 +116,45 @@ mod tests {
         kind: ConstraintKind,
         description: String,
         columns: Vec<String>,
+    }
+
+    fn resolve_destructive_test_database_url(
+        opt_in: Option<&str>,
+        database_url: Option<String>,
+    ) -> std::result::Result<String, &'static str> {
+        if opt_in != Some("1") {
+            return Err(
+                "set BUZZ_TEST_DESTRUCTIVE_DB=1 to enable schema-destructive database tests",
+            );
+        }
+
+        database_url
+            .filter(|url| !url.trim().is_empty())
+            .ok_or("set BUZZ_TEST_DATABASE_URL to an isolated disposable Postgres database")
+    }
+
+    fn destructive_test_database_url() -> String {
+        let opt_in = std::env::var(DESTRUCTIVE_TEST_OPT_IN_ENV).ok();
+        let database_url = std::env::var(TEST_DATABASE_URL_ENV).ok();
+
+        resolve_destructive_test_database_url(opt_in.as_deref(), database_url)
+            .unwrap_or_else(|message| panic!("{message}"))
+    }
+
+    #[test]
+    fn destructive_database_tests_require_explicit_opt_in_and_url() {
+        let url = "postgres://buzz:buzz_test@localhost:5432/buzz_test".to_owned();
+
+        assert!(resolve_destructive_test_database_url(None, Some(url.clone())).is_err());
+        assert!(resolve_destructive_test_database_url(Some("0"), Some(url.clone())).is_err());
+        assert!(resolve_destructive_test_database_url(Some("true"), Some(url.clone())).is_err());
+        assert!(resolve_destructive_test_database_url(Some("1"), None).is_err());
+        assert!(resolve_destructive_test_database_url(Some("1"), Some(String::new())).is_err());
+        assert!(resolve_destructive_test_database_url(Some("1"), Some(" \t".to_owned())).is_err());
+        assert_eq!(
+            resolve_destructive_test_database_url(Some("1"), Some(url.clone())),
+            Ok(url)
+        );
     }
 
     /// Concatenated SQL of every embedded migration, in version order.
@@ -1085,14 +1125,12 @@ mod tests {
         );
     }
 
-    async fn connect_test_pool() -> PgPool {
-        let database_url = std::env::var("BUZZ_TEST_DATABASE_URL")
-            .or_else(|_| std::env::var("DATABASE_URL"))
-            .unwrap_or_else(|_| TEST_DB_URL.to_owned());
+    async fn connect_destructive_test_pool() -> PgPool {
+        let database_url = destructive_test_database_url();
 
         PgPool::connect(&database_url)
             .await
-            .expect("connect to test DB")
+            .expect("connect to isolated destructive test DB")
     }
 
     async fn reset_public_schema(pool: &PgPool) {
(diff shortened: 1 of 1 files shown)


yinkev:
@codex review

yinkev:
**Author-status audit — refined 2026-08-05:** this is a one-file mechanical conflict, not a semantic redesign. Current `main` changed the unsafe hard-coded `TEST_DB_URL` line only by appending `// sadscan:disable np.postgres.1`; this branch removes the fallback entirely and therefore also removes the need for that suppression. The remaining five hunks still match current `main` exactly.

Prepared resolution against current main `8342dfcc5890b81a269a8ec3db73a8a56f76ce79`:
- replace the hard-coded URL/suppression with `BUZZ_TEST_DESTRUCTIVE_DB` and `BUZZ_TEST_DATABASE_URL` constants;
- retain the branch’s pure opt-in/url validator and regression;
- retain the destructive-only pool helper and the three explicit ignored-test gates;
- do not preserve the `DATABASE_URL` or development-database fallback.

The six-hunk rebased patch was dry-run against a synthetic fixture built from the exact current-main hunk contexts: all six hunks applied, zero rejects. The PR stays draft because the real `buzz-db` unit/clippy/Postgres gates cannot be rerun until the fork branch is writable. The current GitHub installation still cannot update `yinkev/buzz`.

PR #2536, opened 2026-07-23
feat(desktop): register Cursor as native ACP runtime + onboarding harness

Bartok9:
## Summary

Register **Cursor** as a first-class ACP harness and show it on **Set up your agent harnesses** next to Claude Code and Codex.

Closes #2535. Related: #1683 (prior Cursor PR), #2347 / #2546 (Grok Build companion).

## Why this PR exists

Onboarding currently hard-filters visible harnesses to `claude` + `codex` only (`ONBOARDING_RUNTIME_ORDER`). Cursor users cannot Install/sign-in on that page even though Cursor speaks ACP natively:

```bash
cursor-agent acp
```

## Critical design note (Grok collision)

**Both Cursor and Grok Build install `~/.local/bin/agent`.** This PR deliberately:

- Uses **`cursor-agent`** as the catalog command (not bare `agent`)
- Only treats path-qualified `agent` as Cursor when the binary path is under Cursor’s install tree
- Does **not** process-sweep bare `agent` (would claim Grok workers)

## What changed

- `KNOWN_ACP_RUNTIMES`: `cursor` entry (`cursor-agent`, default args `acp`, install script, `cursor-agent status` auth probe, `.cursor/skills`, `~/.cursor/bin` on PATH)
- Readiness + process-sweep binary names (`cursor-agent` only)
- Onboarding visibility order: `claude`, `codex`, **`cursor`**
- `agentReadiness` treats Cursor like other CLI harnesses
- Unit tests for arg normalization, avatar, path disambiguation vs Grok
- `buzz-acp` README: Cursor path + extension-method follow-up note

## Non-claims

- Does **not** implement auto-allow policy for Cursor blocking extension methods (follow-up).
- Does **not** implement Grok Build (see #2546 / #2347).

## Test plan

- [x] `cargo test --lib normalizes_cursor`
- [x] `cargo test --lib resolves_cursor`
- [x] `cargo test --lib path_qualified_cursor`
- [x] `cargo test --lib cursor_runtime`
- [x] `node --test onboardingRuntimeSelection.test.mjs`
- [ ] Desktop: harness page shows Cursor Install; after `cursor-agent login`, Next enables
- [ ] Machine with Grok installed: Cursor still resolves via `cursor-agent`, not Grok’s `agent` shim

Signed-off-by: Bartok9 <danielrpike9@gmail.com>

PR #2418, opened 2026-07-22
feat(acp): add OpenCode and Cursor runtimes

austin-thesing:
## Summary

- discover OpenCode and Cursor Agent as native ACP runtimes in Buzz Desktop
- normalize their launch arguments to `acp`, including Cursor’s `agent` and `cursor-agent` commands
- document standalone `buzz-acp` configuration for both runtimes

## Verification

- `opencode acp --help`
- `agent acp --help`
- `cursor-agent acp --help`
- `cargo test -p buzz-acp normalizes_acp_subcommand_runtimes_to_acp`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib managed_agents::discovery::tests::normalizes_opencode_and_cursor_args_to_acp`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib managed_agents::discovery::tests::opencode_and_cursor_are_known_native_acp_runtimes`
- `cargo fmt --all -- --check`
- `git diff --check`

## Note

The broader `cargo test -p buzz-acp config::tests` suite has one unrelated macOS fixture failure in `test_load_rules_empty_name_rejected`: its attempted temporary filename is rejected with `Invalid argument`.

PR #2953, opened 2026-07-26
feat(desktop): add Cursor CLI ACP OAuth and model selection

SomSamantray:
## Summary

Buzz can now use an installed Cursor CLI as its own ACP harness, including Cursor-owned authentication and model selection. It adds the missing vendor auth flow, ACP model discovery, startup model selection, and cross-platform launch handling on top of the existing runtime-registration work.

## Design

- Discovers both `agent` and `cursor-agent`, uses the detected Cursor command on macOS/Linux, and supports the corresponding WSL guest command on Windows.
- Reads Cursor ACP `authMethods` and launches Cursor's own visible login flow when needed; Buzz does not store Cursor credentials or substitute a Buzz OAuth token.
- Uses ACP `session/new` model discovery and injects the selected model with Cursor's startup `--model` argument. The picker persists changes for the next launch instead of sending an unsupported live model-switch request.
- Keeps Cursor-specific startup arguments and vendor auth handling isolated from Claude Code, Codex, and generic ACP runtimes.

Related: #2418, #2536, #2773

## Validation

- `cargo fmt --all -- --check` passed.
- `git diff --check` passed.
- `pnpm --dir desktop check` passed.
- `pnpm --dir desktop typecheck` passed.
- The onboarding readiness suite passed all 18 tests, including preferred Cursor runtime selection.
- Live Cursor validation passed: both `agent` and `cursor-agent` resolve, `agent status` reports the authenticated account, native ACP `initialize` and `session/new` succeed, the model catalog is returned, and `agent --model gpt-5.5 acp` reports the requested model as current.
- `cargo check -p buzz-acp --offline` could not reach compilation because the pinned `aws-creds`/`rust-s3` git dependency was unavailable in this environment; the network fetch was also blocked by the environment proxy.

### Startup-only model selection

Some ACP harnesses accept a model only when the process starts rather than through an in-session control method. Buzz records that capability in runtime metadata, persists picker changes for the next launch, and suppresses live switching for those runtimes.

This is used here because Cursor owns model configuration at process startup: Buzz starts Cursor with `--model <id>` and lets Cursor's ACP session report the resulting model. It should not be used for runtimes that support reliable live model changes through ACP.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -20,6 +20,22 @@ use crate::usage::{TurnUsage, UsageTracker};
 /// Lines exceeding this limit are rejected to prevent OOM from rogue agents.
 const MAX_LINE_SIZE: usize = 10_000_000; // 10 MB
 
+const BUZZ_CHILD_SECRET_ENV_KEYS: &[&str] = &[
+    "BUZZ_PRIVATE_KEY",
+    "BUZZ_ACP_PRIVATE_KEY",
+    "BUZZ_RELAY_URL",
+    "BUZZ_AUTH_TAG",
+    "BUZZ_API_TOKEN",
+    "BUZZ_ACP_API_TOKEN",
+    "NOSTR_PRIVATE_KEY",
+];
+
+fn scrub_buzz_child_secrets(command: &mut tokio::process::Command) {
+    for key in BUZZ_CHILD_SECRET_ENV_KEYS {
+        command.env_remove(key);
+    }
+}
+
 /// An MCP server configuration passed to `session/new`.
 ///
 /// Corresponds to the `McpServerStdio` variant in the ACP schema.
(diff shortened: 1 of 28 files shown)


wpfleger96:
🤖 hey @SomSamantray — thanks for building this. This PR goes further than the plain runtime-registration ones (vendor auth flow, ACP model discovery, WSL handling), so I want to be clear about what's superseded and what isn't.

The registration half is covered: #2773 (ships in v0.5.0) added a generic bring-your-own-harness path, and Cursor is a one-click preset launching `cursor-agent acp`. The auth and model-selection polish you built goes beyond what a preset does — but it would need rebuilding on top of the new preset seam rather than as Cursor-specific runtime code, so I'm closing this rather than trying to rebase it.

If the Cursor auth flow still has rough edges on v0.5.0, a focused issue describing the gap would be a great starting point for a follow-up scoped to the new architecture.

PR #2955, opened 2026-07-26
feat(relay): accept agent job-lifecycle kinds (43001-43006) in ingest

flaukowski:
`buzz-core` defines the NIP-90-style agent job kinds (`KIND_JOB_REQUEST` … `KIND_JOB_ERROR`, 43001-43006) and exports them in `ALL_KINDS`, but `required_scope_for_kind` in the relay ingest never maps them — so a member posting a job-request event is rejected with `restricted: unknown event kind`.

This maps the six job kinds to `Scope::MessagesWrite`. They are not command kinds and not global-only, so they get channel-scoped via their `h` tag and stored + fanned out through the same generic message path as `KIND_TEXT_NOTE` / `KIND_FORUM_COMMENT` — letting an agent offer NIP-90-style services in a channel (request → accepted → result). Adds a scope unit test.

Discovered while building an agent that answers job requests in-channel; happy to adjust the scope choice if there's a preferred authorization model for job events.

## Related work

Searched open PRs and issues for the 43001-43006 kinds and for `required_scope_for_kind` — none found covering this.

Rebased onto current `main`; `cargo test -p buzz-relay --lib` and `cargo clippy -p buzz-relay` are clean.

### DIFF
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -21,15 +21,16 @@ use buzz_core::kind::{
     KIND_GIT_STATUS_CLOSED, KIND_GIT_STATUS_DRAFT, KIND_GIT_STATUS_MERGED, KIND_GIT_STATUS_OPEN,
     KIND_HUDDLE_ENDED, KIND_HUDDLE_GUIDELINES, KIND_HUDDLE_PARTICIPANT_JOINED,
     KIND_HUDDLE_PARTICIPANT_LEFT, KIND_HUDDLE_STARTED, KIND_IA_ARCHIVE_REQUEST,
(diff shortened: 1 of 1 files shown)


flaukowski:
@tlongwell-block — you've been the most active on `handlers/ingest.rs`, so flagging this your way; apologies if that's the wrong person to ask.

Rebased onto current `main` today. `cargo test -p buzz-relay --lib` and `cargo clippy -p buzz-relay` are clean locally.

The `CI` and `Docker image` runs are sitting at `action_required` — as an outside contributor they need a maintainer to approve the workflow run before they'll execute, so there's no green check to look at yet.

Happy to adjust if `Scope::MessagesWrite` isn't the authorization model you'd want for job events — that was the part I was least sure about.

Issue #1593, opened 2026-07-07
relay: workflows created pre-#1369 are orphaned — undeletable, undisableable, and fire forever


AaronGoldsmith:
## Summary

Workflows created while a relay ran the pre-#1369 code path (≤ relay-v0.1.1) are permanently unmanageable — invisible orphans that keep firing on their cron schedules and cannot be disabled or deleted through any client-visible ID. We hit this in production (staging relay `sprout-oss`) and it took several days and a code-level investigation to stop twice-daily ghost messages.

## Background: how the orphans are created

Before #1369 ("Harden relay attack surfaces", merged 2026-06-30), `handle_workflow_def` called `create_workflow`, which generated a **random server-side UUID** (`Uuid::new_v4()`) for the DB row and ignored the kind:30620 event's `d`-tag:

```rust
// crates/buzz-db/src/workflow.rs (relay-v0.1.1)
pub async fn create_workflow(...) -> Result<Uuid> {
    let id = Uuid::new_v4();   // <-- never matches the event d-tag
    ...
}
```

Clients (CLI `buzz workflows list`, desktop) list workflows by querying **kind:30620 events** and reading the `d`-tag as the workflow ID. So every workflow ID a user has ever seen never matched the actual `workflows` row. Consequences:

1. **Disable is a no-op**: re-submitting the definition with `enabled: false` (same d-tag) created a *new* DB row under yet another random UUID. The original `enabled: true` row keeps firing.
2. **Delete is a no-op**: kind:5 with a-tag `30620:<owner>:<visible-uuid>` looks up a UUID that doesn't exist in the DB. Worse, the relay returns `accepted: true` with no error, so the user believes the deletion worked (side-effect failures are logged server-side only).
3. After the relay was upgraded past #1369 (upsert-by-d-tag), all *new* operations behave correctly — but the legacy orphan rows remain, enabled, firing forever, with UUIDs no client can discover.

## How we escaped

`handle_a_tag_deletion` has a name-based fallback: when the a-tag d-value is not a UUID, it resolves via `find_workflow_by_owner_and_name(community, actor, name)`. This reaches orphans regardless of UUID. The stock CLI rejects non-UUID values (`parse_uuid`), so we patched `cmd_delete_workflow` locally to emit `30620:<owner>:<name>` and swept the orphans (repeating, since the lookup is `LIMIT 1`). Verified with webhook canaries (202 → delete → 404).

## Proposed fixes

[...]

PR #2956, opened 2026-07-26
fix(workflow): honor definition 'enabled' flag on upsert

flaukowski:
`upsert_workflow`'s `ON CONFLICT DO UPDATE` sets `name`/`definition`/`definition_hash` but never `enabled`. Re-publishing a kind:30620 workflow definition with `enabled: false` therefore updates the stored JSON while leaving the row's `enabled` column `TRUE` — a disabled scheduled workflow keeps firing.

This parses `enabled` from the definition (WorkflowDef.enabled, default true) and applies it on both insert and update. Minimal, contained to the DB helper; no signature change.

Found while disabling a scheduled workflow via a re-published definition.

## Related work

#1593 reports workflows that "keep firing on their cron schedules and cannot be disabled", and attributes it to orphan rows created before #1369, stating that after the d-tag upsert landed "all *new* operations behave correctly".

That is still not true for the disable path on current `main`, for a separate reason: the upsert's `ON CONFLICT DO UPDATE` never writes the `enabled` column, so re-publishing a definition with `enabled: false` updates the JSON and leaves the row enabled. No orphan row required — this reproduces on a workflow created today.

This PR fixes that half. #1593's orphan reconciliation (and its point 2, surfacing side-effect failures) is still needed for the legacy rows.

Rebased onto current `main`; `cargo clippy -p buzz-db` is clean.

### DIFF
--- a/crates/buzz-db/src/workflow.rs
+++ b/crates/buzz-db/src/workflow.rs
@@ -320,15 +320,24 @@ pub async fn upsert_workflow(
     definition_json: &str,
     definition_hash: &[u8],
 ) -> Result<()> {
+    // The `enabled` flag lives in the definition (WorkflowDef.enabled, default
+    // true). Honor it on both insert AND update — otherwise re-publishing a
+    // definition with `enabled: false` changes the JSON but leaves the row
+    // enabled, so a disabled workflow keeps firing.
+    let enabled = serde_json::from_str::<serde_json::Value>(definition_json)
+        .ok()
+        .and_then(|v| v.get("enabled").and_then(|e| e.as_bool()))
+        .unwrap_or(true);
     let row = sqlx::query(
         r#"
         INSERT INTO workflows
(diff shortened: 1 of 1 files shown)


flaukowski:
@tlongwell-block — flagging this your way as the most recent author of `buzz-db/src/workflow.rs`.

Rebased onto current `main` today; `cargo clippy -p buzz-db` is clean.

Worth surfacing since it may raise the priority: this is a still-live half of #1593. That issue concludes that after #1369's d-tag upsert landed, "all *new* operations behave correctly" — but disable-via-redefinition is still a no-op on current `main`, for an unrelated reason: the upsert's `ON CONFLICT DO UPDATE` sets `name`/`definition`/`definition_hash` but never `enabled`. No legacy orphan row required; it reproduces on a workflow created today, which is how we hit it. #1593's orphan reconciliation is still separately needed for the pre-#1369 rows.

`CI` is at `action_required` and needs a maintainer to approve the run.

PR #2988, opened 2026-07-26
fix(relay): allow owner to delete archived channels

### LINKED ISSUE #2954: Deleting an auto-archived huddle channel returns 400 "channel is archived"
**Describe the bug**

Deleting a channel after it has been archived fails with:

```text
relay returned 400 Bad Request: invalid: channel is archived
```

The user-visible case is an ephemeral huddle channel. When the last participant leaves, the relay automatically archives the huddle channel. Desktop can keep the channel row visible and offers **Delete channel**; confirming deletion sends NIP-29 `DELETE_GROUP` (kind `9008`), which the relay rejects before reaching the delete handler. The same context menu can show **Archive channel** even though the relay already considers the channel archived.

The underlying protocol path is broader than huddles: the relay rejects kind `9008` for any archived channel. Both archived-channel guards exempt only an unarchive event (kind `9002` with `archived=false`), even though the later kind `9008` validator independently enforces owner authorization and the handler performs a soft delete.

**Steps to reproduce**

1. Create a channel as its owner.
2. Archive it with kind `9002` and `archived=true`; ending a huddle so its ephemeral channel auto-archives also reproduces this.
3. As the same owner, submit a kind `9008` delete-group event for that channel. In Desktop, right-click the retained huddle channel, select **Delete channel**, and confirm.
4. Observe the HTTP 400 response: `invalid: channel is archived`.

**Expected behavior**

An authorized owner should be able to permanently delete an archived channel. Kind `9008` should pass the archived-channel gates and continue through its existing owner-only validation before the soft-delete handler runs. Other writes to archived channels should remain rejected.

Desktop should also render lifecycle actions from the authoritative archived state, offering **Unarchive channel** rather than **Archive channel** when appropriate while still allowing deletion.

**Version and platform**

- Buzz version: 0.4.26
- OS: macOS 26.5.2
- Relay: commit `c0d7b52af8d5f3572790be697dd2c4bf3c6a3b8f`; current upstream `main` at `c2a4ee711e481bb427d6cf8cd08b2c7329d1508c` has the same guards

**Logs / additional context**

Relevant code paths:

- `validate_admin_event` rejects archived channels before its kind `9008` owner check: https://github.com/block/buzz/blob/c0d7b52af8d5f3572790be697dd2c4bf3c6a3b8f/crates/buzz-relay/src/handlers/side_effects.rs#L276-L290
- The general ingest guard independently rejects every archived-channel event except unarchive: https://github.com/block/buzz/blob/c0d7b52af8d5f3572790be697dd2c4bf3c6a3b8f/crates/buzz-relay/src/handlers/ingest.rs#L1953-L1968
- Kind `9008` already has owner-only authorization: https://github.com/block/buzz/blob/c0d7b52af8d5f3572790be697dd2c4bf3c6a3b8f/crates/buzz-relay/src/handlers/side_effects.rs#L635-L654
- Huddle teardown auto-archives the ephemeral channel: https://github.com/block/buzz/blob/c0d7b52af8d5f3572790be697dd2c4bf3c6a3b8f/crates/buzz-relay/src/audio/handler.rs#L834-L842

A focused regression should archive a channel, submit a correctly signed kind `9008` as its owner, and assert that the channel is soft-deleted; a non-owner kind `9008` must still be rejected.

Duplicate search: no existing open or closed issue or pull request was found for this archive-then-delete failure.

bokiko:
## Summary
Deleting an archived channel currently fails with 400 `invalid: channel is archived`. Two independent archived-channel guards in the relay reject every mutation on an archived channel except unarchive (kind:9002 with `archived=false`), so the owner's kind:9008 delete is rejected **before** it ever reaches its authorization check and delete handler:

1. `crates/buzz-relay/src/handlers/side_effects.rs` — the guard in `validate_admin_event` runs ahead of the per-kind auth match.
2. `crates/buzz-relay/src/handlers/ingest.rs` — a second, independent rejection in the ingest path.

This change exempts kind:9008 (delete — a valid terminal transition from archived state) at both guards, using the existing `KIND_NIP29_DELETE_GROUP` constant (the guard's raw `9002` literal is upgraded to `KIND_NIP29_EDIT_METADATA` in passing). Authorization is untouched: the 9008 validator arm still runs after the guard and rejects anyone who is not the owner (or the owning human of an owner-role agent), then the existing soft-delete handler runs unchanged.

### Related issue
Fixes #2954. Searched open PRs for duplicates — none found.

### Testing
- New e2e regression test `test_owner_can_delete_archived_channel` (`crates/buzz-test-client/tests/e2e_relay.rs`): create → archive → unarchive succeeds (9002 exemption intact) → re-archive → non-owner delete rejected with `only owner can delete group` (proves the auth path still runs, not the archived guard) → owner delete succeeds → post-delete unarchive fails with `channel not found` (soft-delete took effect).
- Repro proof: the same test against the unpatched relay fails at the non-owner assertion with `invalid: channel is archived` — exactly the reported bug.
- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- Workspace tests green: buzz-relay lib (751), buzz-core (45), buzz-auth (229), buzz-db (84), buzz-conformance, buzz-push-gateway. Live e2e suite 37/39 — the 2 failures (`test_unarchive_emits_member_added_notification`, `test_subscription_limit_enforced`) are pre-existing and reproduce identically on the unpatched build.

### DIFF
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -1970,14 +1970,18 @@ async fn ingest_event_inner(
     }
 
     if channel_id.is_some() {
-        // Allow kind:9002 with archived=false (unarchive operation)
+        // Allow kind:9002 with archived=false (unarchive operation) and
+        // kind:9008 (delete) — a valid terminal transition from archived
+        // state; validate_admin_event still enforces kind:9008 authorization
+        // (owner, or the owning human of an owner-role agent).
         let is_unarchive = kind_u32 == KIND_NIP29_EDIT_METADATA
             && event.tags.iter().any(|t| {
                 let parts = t.as_slice();
                 parts.len() >= 2 && parts[0] == "archived" && parts[1] == "false"
             });
+        let is_delete = kind_u32 == KIND_NIP29_DELETE_GROUP;
 
-        if !is_unarchive {
+        if !is_unarchive && !is_delete {
             if let Some(channel) = &channel_row {
                 if channel.archived_at.is_some() {
                     return Err(IngestError::Rejected("invalid: channel is archived".into()));
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -9,9 +9,9 @@ use uuid::Uuid;
 use buzz_core::kind::{
     event_kind_u32, is_parameterized_replaceable, KIND_AGENT_PROFILE, KIND_DM_VISIBILITY,
     KIND_GIT_REPO_ANNOUNCEMENT, KIND_IA_ARCHIVED, KIND_IA_ARCHIVED_LIST, KIND_IA_UNARCHIVED,
-    KIND_MEMBER_ADDED_NOTIFICATION, KIND_MEMBER_REMOVED_NOTIFICATION, KIND_NIP29_GROUP_ADMINS,
-    KIND_NIP29_GROUP_MEMBERS, KIND_NIP29_GROUP_METADATA, KIND_NIP43_MEMBERSHIP_LIST, KIND_REACTION,
-    KIND_THREAD_SUMMARY,
+    KIND_MEMBER_ADDED_NOTIFICATION, KIND_MEMBER_REMOVED_NOTIFICATION, KIND_NIP29_DELETE_GROUP,
+    KIND_NIP29_EDIT_METADATA, KIND_NIP29_GROUP_ADMINS, KIND_NIP29_GROUP_MEMBERS,
+    KIND_NIP29_GROUP_METADATA, KIND_NIP43_MEMBERSHIP_LIST, KIND_REACTION, KIND_THREAD_SUMMARY,
 };
 use buzz_core::StoredEvent;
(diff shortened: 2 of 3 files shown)


bokiko:
Ready for review when someone from @block/buzz-oss-team has a moment.

Fixes #2954 — deleting an archived channel failed with `invalid: channel is archived`. Two independent archived-channel guards rejected kind:9008 before it reached its authorization check, so an owner's delete never got as far as the owner test.

The change exempts kind:9008 at both guards (`ingest.rs`, `side_effects.rs`), matching the existing kind:9002 unarchive exemption. Authorization is unchanged — the 9008 arm still runs after the guard and still rejects non-owners.

- Rebased onto current main (`63496cc1`); `MERGEABLE`, no conflicts
- Semgrep OSS, zizmor, DCO Check all green on `16e9d231`
- `cargo test -p buzz-relay` — 780 passed, 0 failed, 35 ignored
- New e2e regression test `test_owner_can_delete_archived_channel`; confirmed it fails on the unpatched relay at the non-owner assertion with the reported error

Happy to rebase again, split it, or move the test if you'd prefer a different shape.

PR #2995, opened 2026-07-26
fix(mobile): retry channel-sections startup sync when relay rate-limits cold start

tellaho:
**Category:** fix
**User Impact:** Channel groups created on desktop now reliably appear on Android and iOS on cold start, instead of falling back to the default ungrouped list.

**Problem:** On mobile cold start, ChannelsNotifier fires ~25 per-channel REQs at once, exhausting the relay's per-connection rate-limit quota. `ChannelSectionsManager` then gets BOTH its one-shot history fetch and its live subscription rejected with `rate-limited: quota exceeded` — and both errors were silently swallowed (`catch (_)`) with no retry, so the manager kept the local (empty/default) store forever. Restarting the app repeats the same storm, so Android reliably lost the race every launch. Captured live on the emulator with instrumentation.

**Solution:** Track whether the startup fetch and the live subscription have each succeeded, and retry `_syncWithRelay` with exponential backoff (2s base, shift-capped, 30s max) until both land. The retry timer is cancelled on dispose, and previously-swallowed errors are now logged.

Based directly on `main` — independent of #2829 (which fixes the *write* path: unpublished local edits being clobbered). The analogous retry for `ChannelSortManager` lives in #2829, since that manager is introduced there.

<details>

Extract the startup fetch + live-subscription into `_syncWithRelay`, track success of each step, and schedule a backoff retry until both succeed. `_fetchAndMerge` and `_startLiveSubscription` now report success; swallowed errors are logged; retry timer cancelled on dispose. `startupRetryBaseDelay` ctor param is test-visible.

New regression tests with a rate-limiting relay fake: remote sections are adopted after retries; retry stops once fetch + subscription succeed; dispose cancels pending retries.

</details>

## Reproduction Steps

1. On desktop, create channel groups (sections) for an account.
2. Cold-start the Android app for the same account on a relay with per-connection rate limiting and enough joined channels to trigger the REQ burst (~25 channels reproduced it reliably).
3. Before this fix: logs show `fetch FAILED: Exception: rate-limited: quota exceeded` and the live subscription failing, then silence — the channel list renders the default ungrouped list forever, surviving app restarts.
4. With this fix: logs show `startup sync incomplete; retrying in 2000ms (attempt 1)`, the retry succeeds, and the desktop-created groups render.

## Verification

- Live on emulator-5554 (earlier stacked build of the same logic): cold start reproduced the manager being rate-limited, then a single 2s retry succeeding and groups rendering, matching desktop channel-for-channel.
- Full mobile suite run at this exact head (c5f1d9a38): 654 passing; the 4 failures (3× `compose_bar_test`, 1× `channels_page_test`) reproduce on unmodified `main` (74b63e184) — pre-existing, unrelated. `flutter analyze` clean on both touched files.

Originating thread: Buzz channel ed3994af-0949-447c-be00-29f03965b52e, root 4bf7cbfd48bf.

### DIFF
--- a/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
+++ b/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
@@ -48,6 +48,12 @@ class ChannelSectionsManager {
   void Function()? _unsubscribe;
   bool _disposed = false;
 
+  /// Base delay for the startup-sync retry backoff. Overridable in tests.
+  final Duration _startupRetryBaseDelay;
+  Timer? _startupRetryTimer;
+  int _startupRetryAttempt = 0;
+  bool _startupFetchSucceeded = false;
+
   ChannelSectionsManager({
     required this.pubkey,
     required SharedPreferences prefs,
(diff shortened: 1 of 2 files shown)


tellaho:
Superseded by #3004 — identical diff and head (c5f1d9a38), recreated on `tho/sections-startup-retry` per branch-naming convention.

PR #3001, opened 2026-07-26
feat(cli): add project issue comments

### LINKED ISSUE #2989: CLI cannot comment on Projects issues: no `buzz issues comment`, and `social publish --reply-to` omits the `a` tag Desktop filters on
## What's missing

Buzz Desktop supports comments on Projects issues: `ProjectIssuesPanel.tsx` renders `issue.comments` and posts via `useCreateProjectIssueCommentMutation` (`desktop/src/features/projects/hooks.ts`). The `buzz` CLI has no equivalent — `buzz issues` is `create` / `get` / `list` / `status` only.

## Why the obvious workaround doesn't work

Comments are kind:1 text notes. `buzz social publish --reply-to <issue-event-id>` produces an event with only an `e` tag:

```json
"tags":[["e","<issue-id>","","reply"]]
```

But Desktop resolves comments with **two** conditions (`desktop/src/features/projects/hooks.ts` `fetchProjectIssues`, and `commentsForIssue` in `projectIssues.mjs`):

1. relay filter `kinds:[1]`, `#a: [<repoAddress>]` — requires an `a` tag of `30617:<owner>:<repo-d-tag>`
2. an `e`/`E` tag equal to the issue event id

A CLI-published note satisfies (2) but not (1), so it is fetched by nobody and never appears on the issue. It also can't be rescued client-side: `desktop/src/features/pulse/lib/projectComments.ts` uses the same `a`-tag prefix test to keep project comments out of Pulse feeds.

Net effect: an agent driving Buzz through the CLI can open and close issues but cannot participate in the discussion on them, while a human in Desktop can. For agent/human collaboration on a project, that's the wrong way round.

## Suggested fix

Add `buzz issues comment --issue <event-id> --repo-owner <hex> --repo-id <d-tag> --content -`, emitting kind:1 with both the `a` (`30617:owner:d-tag`) and `e` tags, matching what Desktop publishes. A `--to` for mention pubkeys would mirror `buzz issues create`.

Verified against `block/buzz` @ ab3af828 and a live relay (published a real issue, then a `social publish` reply, and confirmed the tag shape above).

brainx:
## Summary

- add `buzz issues comment` with literal or stdin content and repeatable `--to` recipients
- load the root issue before publishing, verify its repository coordinate, and inherit its author and existing recipients
- publish Desktop-compatible kind `1` events with the issue root `e` tag and repository `a` tag
- validate project identifiers and deduplicate normalized `p` tags
- add SDK event-shape, CLI inventory, recipient-inheritance, and repository-mismatch regressions

## Why

The CLI could create and manage project issues but could not comment on them. Using `social publish --reply-to` was not equivalent because it omitted the repository coordinate that project clients use to discover issue comments. A correct project comment must also address the issue author and existing recipients, matching the Desktop client.

Fixes #2989

## Verification

- `cargo test -p buzz-sdk -p buzz-cli` — 573 unit tests passed; 1 CLI doctest ignored
- `cargo clippy -p buzz-cli --all-targets -- -D warnings` — passed
- `cargo fmt -p buzz-cli -p buzz-sdk -- --check` — passed
- `git diff --check upstream/main...HEAD` — passed


ScaleLeanChris:
Real consumer here via #2989. We run Hermes agents through the Buzz CLI and need their Project issue comments to appear in the same thread that humans see in Desktop.

I found this implementation before starting any duplicate work, so I will not create a competing patch. I can provide an end-to-end validation of this PR using our Hermes/Buzz setup: create an issue, comment from Hermes with stdin and repeatable `--to`, inspect the signed `a`/`e`/`p` tag shape, and verify that a second Buzz Desktop machine renders the comment in the issue thread.

@brainx, would that validation be useful on the current head, or are you planning another restack first?

ScaleLeanChris:
Independent live-consumer validation passed on head `01eee65e7a12ba44b57f700f67e4570f2d06d808`, combined with NousResearch/hermes-agent#78065 at `b92786372af4037f944940f15c3188636c4a937a`.

I used a bounded synthetic repository and issue with no work authority or production data. A real Hermes agent invoked `buzz issues comment` through its terminal tool using stdin content and an explicit `--to` recipient. The command exited `0` and the relay accepted the event.

Independent read-back as the issue owner confirmed:

- event kind `1`;
- root `e` tag points to the issue;
- `a` tag contains the exact repository coordinate;
- the issue author appears once as a `p` recipient, even though it was inherited and also supplied through `--to`;
- the event is signed by the agent identity.

Stock Buzz Desktop then showed the repository with one issue and rendered the agent-authored comment inside the correct issue thread. The synthetic issue was marked resolved after the verifier passed.

No credential values, private content, or production state were used or disclosed.

PR #3003, opened 2026-07-26
fix(relay): strip quotes from ETag before If-Match for Ceph RGW compat

### LINKED ISSUE #3002: Ceph RGW rejects quoted ETags in If-Match (412) — relay CAS / S3 conformance probe fails on Ceph-backed object storage
### Summary

When `buzz-relay` is backed by an S3-compatible object store implemented by **Ceph RGW**, every conditional-write (CAS) — including the **startup S3 conformance probe** — fails with HTTP 412, causing the relay to CrashLoop on startup.

### Root cause

Ceph RGW returns ETags in quoted form (e.g. `"deadbeef"`, per RFC 7232) in GET responses, but **rejects quoted ETags in the `If-Match` precondition header** (returns 412) while accepting the **unquoted** form (e.g. `deadbeef`). This is non-compliant with RFC 7232, which requires `If-Match` to carry a quoted entity-tag.

`buzz-relay` takes the ETag it received from a GET response and forwards it **verbatim** into the `If-Match` header of the subsequent PUT. Because the stored ETag is quoted, every `If-Match` carries a quoted value that Ceph RGW rejects → 412 → CAS fails.

Relevant code: [`crates/buzz-relay/src/api/git/store.rs`](https://github.com/block/buzz/blob/main/crates/buzz-relay/src/api/git/store.rs), `put_pointer()`, the `Precond::IfMatch(ETag(tag))` arm:

```rust
Precond::IfMatch(ETag(tag)) => {
    headers.insert(
        axum::http::header::IF_MATCH,
        tag.parse().map_err(|_| {
            StoreError::Backend(S3Error::HttpFailWithBody(
                400,
                format!("invalid etag {tag}"),
            ))
        })?,
    );
}
```

This single path covers both the conformance probe (`etag_consistency` phase, which calls `get_pointer` → `put_pointer(IfMatch)`) and `cas_publish.rs` (`Precond::IfMatch(e.clone())`).

### Reproduction

- S3 backend: Ceph RGW ( Reef / any recent release).
- Configure buzz-relay against a Ceph-backed bucket (`BUZZ_S3_ENDPOINT`, `BUZZ_S3_BUCKET`, credentials).
- Start the relay: the startup S3 conformance probe fails with 412; the relay exits / CrashLoops.

Empirically confirmed with a stdlib SigV4 probe against our Ceph RGW:
- GET returns `ETag: "deadbeef"` (quoted).
- `If-Match: "deadbeef"` (quoted) → **412 Precondition Failed**.
- `If-Match: deadbeef` (unquoted) → **200 OK**.

### Suggested fix

Normalize the ETag before inserting it into `If-Match` — strip surrounding double quotes:

```rust
Precond::IfMatch(ETag(tag)) => {
    // Ceph RGW rejects quoted If-Match values (RFC 7232 non-compliant) but
    // accepts the unquoted form. AWS S3 and MinIO accept both, so stripping
    // is safe across backends.
    let unquoted = tag.trim_matches('"');
    headers.insert(
        axum::http::header::IF_MATCH,
        unquoted.parse().map_err(|_| {
            StoreError::Backend(S3Error::HttpFailWithBody(
                400,
                format!("invalid etag {unquoted}"),
            ))
        })?,
    );
}
```

This is harmless for AWS S3 and MinIO, which accept both the quoted and unquoted forms, so it does not regress those backends.

Alternatively, a more general fix could normalize the ETag on read (when received from GET) so callers always see the unquoted form, but that is a larger change with broader blast radius.

### Environment
- buzz-relay built from chart tag `chart-v0.1.6`
- S3 backend: Ceph RGW

Happy to open a PR if this approach is acceptable.

darox:
## Problem

Ceph RGW returns quoted ETags but rejects that form in `If-Match`. The relay forwards an ETag obtained from a read into the subsequent conditional write, causing CAS operations and the startup conformance probe to fail.

## Fix

Remove surrounding double quotes only when setting `If-Match`. `If-None-Match: *` is unchanged.

## Validation

- Verified against Ceph RGW in a cluster.
- `cargo fmt --check`
- `cargo clippy -p buzz-relay --lib -- -D warnings`
- `cargo test -p buzz-relay --lib api::git::store::tests`

Closes #3002

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -81,6 +81,9 @@ RELAY_URL=ws://localhost:3000
 # BUZZ_GIT_PACK_CACHE_PATH=./repos/.pack-cache
 # BUZZ_GIT_PACK_CACHE_MAX_BYTES=5368709120
 # BUZZ_GIT_PACK_CACHE_MAX_CONCURRENT_POPULATIONS=2
+# Preserve quoted ETags in If-Match by default (standard S3/MinIO). Set this
+# only when a Ceph RGW deployment rejects the standards-compliant quoted form.
+# BUZZ_GIT_S3_COMPATIBILITY=ceph-rgw
 
 # -----------------------------------------------------------------------------
 # S3-Compatible Object Storage (media + Git/CAS)
--- a/crates/buzz-relay/src/api/git/store.rs
+++ b/crates/buzz-relay/src/api/git/store.rs
@@ -36,6 +36,31 @@ use sha2::{Digest, Sha256};
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ETag(pub String);
 
+/// Backend-specific compatibility applied to conditional S3 requests.
+///
+/// The default preserves ETags byte-for-byte as required by HTTP and standard
+/// S3 implementations. Non-standard behavior must be selected explicitly.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum S3BackendCompatibility {
+    /// Preserve the response ETag exactly when sending `If-Match`.
+    #[default]
+    Standard,
+    /// Remove one surrounding double-quote pair from `If-Match` for Ceph RGW.
+    CephRgw,
+}
+
+impl S3BackendCompatibility {
+    fn if_match_value(self, etag: &str) -> &str {
+        match self {
+            Self::Standard => etag,
+            Self::CephRgw => etag
+                .strip_prefix('"')
+                .and_then(|unquoted| unquoted.strip_suffix('"'))
+                .unwrap_or(etag),
+        }
+    }
+}
+
 /// Precondition for `put_pointer`.
 #[derive(Debug, Clone)]
 pub enum Precond {
@@ -169,6 +194,7 @@ impl From<ProbeFailure> for StoreError {
 #[derive(Clone)]
 pub struct GitStore {
     bucket: Arc<Bucket>,
+    compatibility: S3BackendCompatibility,
 }
 
 impl GitStore {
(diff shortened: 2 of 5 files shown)


dophsquare:
🤖 Review: changes requested — this should not merge as written.

`crates/buzz-relay/src/api/git/store.rs:493` globally strips the quotes from every `If-Match` ETag. The store intentionally carries the response ETag opaquely from GET/PUT into the CAS write; making it unquoted changes behavior for every S3 backend, not only Ceph. Standard HTTP defines entity tags as quoted opaque strings and `If-Match` as entity tags (RFC 9110 §§8.8.3, 13.1.1: https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.1).

Action items:
- Preserve the original quoted ETag for the default S3/MinIO path.
- If Ceph needs a non-standard format, make it an explicit, opt-in backend compatibility policy rather than global normalization.
- Add regression/integration coverage for the exact `If-Match` wire value on both standard S3 and Ceph-compatible paths.

darox:
> 🤖 Review: changes requested — this should not merge as written.
> 
> `crates/buzz-relay/src/api/git/store.rs:493` globally strips the quotes from every `If-Match` ETag. The store intentionally carries the response ETag opaquely from GET/PUT into the CAS write; making it unquoted changes behavior for every S3 backend, not only Ceph. Standard HTTP defines entity tags as quoted opaque strings and `If-Match` as entity tags (RFC 9110 §§8.8.3, 13.1.1: https://www.rfc-editor.org/rfc/rfc9110.html#section-13.1.1).
> 
> Action items:
> 
> * Preserve the original quoted ETag for the default S3/MinIO path.
> * If Ceph needs a non-standard format, make it an explicit, opt-in backend compatibility policy rather than global normalization.
> * Add regression/integration coverage for the exact `If-Match` wire value on both standard S3 and Ceph-compatible paths.

I fixed the issues

PR #3004, opened 2026-07-26
fix(mobile): retry channel-sections startup sync when relay rate-limits cold start

tellaho:
**Category:** fix
**User Impact:** Channel groups created on desktop now reliably appear on Android and iOS on cold start, instead of falling back to the default ungrouped list.

**Problem:** On mobile cold start, ChannelsNotifier fires ~25 per-channel REQs at once, exhausting the relay's per-connection rate-limit quota. `ChannelSectionsManager` then gets BOTH its one-shot history fetch and its live subscription rejected with `rate-limited: quota exceeded` — and both errors were silently swallowed (`catch (_)`) with no retry, so the manager kept the local (empty/default) store forever. Restarting the app repeats the same storm, so Android reliably lost the race every launch. Captured live on the emulator with instrumentation.

**Solution:** Track whether the startup fetch and the live subscription have each succeeded, and retry `_syncWithRelay` with exponential backoff (2s base, shift-capped, 30s max) until both land. The retry timer is cancelled on dispose, and previously-swallowed errors are now logged.

Based directly on `main` — independent of #2829 (which fixes the *write* path: unpublished local edits being clobbered). The analogous retry for `ChannelSortManager` lives in #2829, since that manager is introduced there.

<details>

Extract the startup fetch + live-subscription into `_syncWithRelay`, track success of each step, and schedule a backoff retry until both succeed. `_fetchAndMerge` and `_startLiveSubscription` now report success; swallowed errors are logged; retry timer cancelled on dispose. `startupRetryBaseDelay` ctor param is test-visible.

New regression tests with a rate-limiting relay fake: remote sections are adopted after retries; retry stops once fetch + subscription succeed; dispose cancels pending retries.

</details>

## Reproduction Steps

1. On desktop, create channel groups (sections) for an account.
2. Cold-start the Android app for the same account on a relay with per-connection rate limiting and enough joined channels to trigger the REQ burst (~25 channels reproduced it reliably).
3. Before this fix: logs show `fetch FAILED: Exception: rate-limited: quota exceeded` and the live subscription failing, then silence — the channel list renders the default ungrouped list forever, surviving app restarts.
4. With this fix: logs show `startup sync incomplete; retrying in 2000ms (attempt 1)`, the retry succeeds, and the desktop-created groups render.

## Verification

- Live on emulator-5554 (earlier stacked build of the same logic): cold start reproduced the manager being rate-limited, then a single 2s retry succeeding and groups rendering, matching desktop channel-for-channel.
- Full mobile suite run at this exact head (c5f1d9a38): 654 passing; the 4 failures (3× `compose_bar_test`, 1× `channels_page_test`) reproduce on unmodified `main` (74b63e184) — pre-existing, unrelated. `flutter analyze` clean on both touched files.

Originating thread: Buzz channel ed3994af-0949-447c-be00-29f03965b52e, root 4bf7cbfd48bf.

### DIFF
--- a/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
+++ b/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
@@ -93,6 +93,11 @@ class ChannelSectionsManager {
   /// with `rate-limited: quota exceeded`) — retry with backoff instead of
   /// silently giving up, which left desktop-created groups invisible until
   /// an unrelated refetch.
+  ///
+  /// Retries are intentionally unbounded for the manager's lifetime: this
+  /// sync must eventually land for groups to appear at all, and at the 30s
+  /// delay ceiling a persistent retry is cheap. Do not "fix" this into a
+  /// bounded loop — giving up permanently is the exact bug this replaces.
   Future<void> _syncWithRelay() async {
     if (!_startupFetchSucceeded) {
       _startupFetchSucceeded = await _fetchAndMerge();
@@ -102,6 +107,11 @@ class ChannelSectionsManager {
 
     if (!_startupFetchSucceeded || !subscribed) {
       _scheduleStartupRetry();
+    } else {
+      // Fully recovered — later transient failures (e.g. a late relay
+      // CLOSED) start backing off from the base delay again instead of the
+      // ceiling the cold start climbed to.
+      _startupRetryAttempt = 0;
     }
   }
 
@@ -109,7 +119,7 @@ class ChannelSectionsManager {
     if (_disposed) return;
     _startupRetryTimer?.cancel();
     // The inner shift cap is overflow protection, not the delay policy: the
-    // attempt counter grows for the process lifetime, and an unchecked `<<`
+    // consecutive-failure counter is unbounded, and an unchecked `<<`
     // past 62 wraps negative, which would make the Timer fire immediately in
     // a hot loop. At the default 2s base the outer 30s clamp is what callers
     // actually observe (2s, 4s, …, 30s); the shift cap only bites for the
--- a/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
+++ b/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
@@ -108,6 +108,12 @@ class ChannelSectionsManager {
   void _scheduleStartupRetry() {
     if (_disposed) return;
     _startupRetryTimer?.cancel();
+    // The inner shift cap is overflow protection, not the delay policy: the
+    // attempt counter grows for the process lifetime, and an unchecked `<<`
+    // past 62 wraps negative, which would make the Timer fire immediately in
+    // a hot loop. At the default 2s base the outer 30s clamp is what callers
+    // actually observe (2s, 4s, …, 30s); the shift cap only bites for the
+    // tiny injected bases used in tests.
     final delayMs = min(
       _startupRetryBaseDelay.inMilliseconds << min(_startupRetryAttempt, 5),
       30000,
--- a/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
+++ b/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
@@ -292,6 +292,7 @@ class ChannelSectionsManager {
           limit: 1,
         ),
         _handleIncomingEvent,
+        onClosed: _handleSubscriptionClosed,
       );
       return true;
     } catch (error) {
@@ -300,6 +301,20 @@ class ChannelSectionsManager {
       // startup-sync backoff.
       return false;
     }
+  }
+
+  /// A relay `CLOSED` can arrive after `subscribe()` already reported
+  /// success: the 500ms readiness wait times out silently under load, and
+  /// the rate-limit rejection lands later. Without this handler the manager
+  /// would keep a dead subscription and never retry — the exact
+  /// load-correlated cold-start failure this retry exists for.
+  void _handleSubscriptionClosed(String message) {
+    if (_disposed) return;
+    debugPrint(
+      '[ChannelSectionsManager] live subscription closed by relay: $message',
+    );
+    _unsubscribe = null;
+    _scheduleStartupRetry();
   }
 
   void _mergeEvents(List<NostrEvent> events) {
--- a/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
+++ b/mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
@@ -48,6 +48,12 @@ class ChannelSectionsManager {
   void Function()? _unsubscribe;
   bool _disposed = false;
 
+  /// Base delay for the startup-sync retry backoff. Overridable in tests.
+  final Duration _startupRetryBaseDelay;
+  Timer? _startupRetryTimer;
+  int _startupRetryAttempt = 0;
+  bool _startupFetchSucceeded = false;
+
   ChannelSectionsManager({
     required this.pubkey,
     required SharedPreferences prefs,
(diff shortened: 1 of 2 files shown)


tellaho:
Recreated from #2995 (same head `c5f1d9a38`, identical diff) to move to the `tho/` branch prefix. Marge's review/approval of this exact head is on #2995.

mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
```diff
@@ -74,15 +83,57 @@ class ChannelSectionsManager {
...
+  /// relay rate-limits the burst of per-channel subscriptions and rejects
+  /// with `rate-limited: quota exceeded`) — retry with backoff instead of
+  /// silently giving up, which left desktop-created groups invisible until
+  /// an unrelated refetch.
+  Future<void> _syncWithRelay() async {
+    if (!_startupFetchSucceeded) {
+      _startupFetchSucceeded = await _fetchAndMerge();
+    }
+
+    final subscribed = _unsubscribe != null || await _startLiveSubscription();
```
wpfleger96, line 106: **IMPORTANT** — a late `CLOSED` is treated as subscription success, so the retry may not fire in the exact case this PR targets.

`_startLiveSubscription()` returns `false` only when `subscribe()` throws. But `subscribe()` awaits its ready completer with `.timeout(const Duration(milliseconds: 500), onTimeout: () {})` (`mobile/lib/shared/relay/relay_session.dart:231-234`) — an empty `onTimeout` that swallows the timeout and returns normally. So after 500ms of relay silence, `subscribe()` resolves successfully and `_unsubscribe` is set even though no EOSE ever arrived.

If the relay's `CLOSED rate-limited` then lands *after* that 500ms window, `_handleClosed` removes the live subscription and only invokes `onClosed` (`relay_session.dart:512-522`) — which this manager doesn't pass. Result: `subscribed` is `true`, no retry is scheduled, and the manager sits with a dead subscription.

This matters because the failure is load-correlated in the same direction as the bug: the ~25-channel REQ burst that causes the rate-limit is also what makes the relay slow to respond, so the more contended the cold start, the likelier `CLOSED` misses the 500ms window. The PR's reproduction works because `CLOSED` happened to arrive inside it.

The mechanism to fix this already exists and is used elsewhere — see `mobile/lib/features/channels/agent_activity/observer_subscription.dart:143`. Either:

1. Pass an `onClosed` callback that clears `_unsubscribe` and re-schedules `_syncWithRelay()`, or
2. Treat a readiness timeout in `subscribe()` as not-subscribed (broader blast radius — other callers depend on the current lenient behavior, so option 1 is the safer change for this PR).

mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
```diff
@@ -74,15 +83,57 @@ class ChannelSectionsManager {
...
+    if (!_startupFetchSucceeded || !subscribed) {
+      _scheduleStartupRetry();
+    }
+  }
+
+  void _scheduleStartupRetry() {
+    if (_disposed) return;
+    _startupRetryTimer?.cancel();
+    final delayMs = min(
+      _startupRetryBaseDelay.inMilliseconds << min(_startupRetryAttempt, 5),
```
wpfleger96, line 128: **MINOR** — two interacting caps express one ceiling, and the inner one is dead at the default.

`_startupRetryBaseDelay.inMilliseconds << min(_startupRetryAttempt, 5)` with the 2s default reaches 64s at the shift cap, and the outer `min(..., 30000)` already clamps to 30s. The inner `min` only changes behavior for a base delay below ~470ms, i.e. only in tests. Dropping it gives identical production behavior with one fewer thing to reason about:

```dart
final delayMs = min(
  _startupRetryBaseDelay.inMilliseconds << _startupRetryAttempt,
  30000,
);
```

If the inner cap is there to prevent shift overflow on a pathological injected base, that's legitimate — worth saying so in a comment, since it isn't apparent from the code.

mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
```diff
@@ -74,15 +83,57 @@ class ChannelSectionsManager {
...
+      30000,
+    );
+    _startupRetryAttempt++;
```
wpfleger96, line 131: **MINOR** — `_startupRetryAttempt` is never reset on success, so a later transient failure starts at the 30s ceiling.

Once the counter has climbed (say five failed cold-start attempts), a subsequent successful sync leaves it there. If the relay drops and `_syncWithRelay` needs to retry again later in the same process, the first retry waits 30s instead of 2s — the backoff has no memory of having recovered. Reset it when a pass fully succeeds:

```dart
if (!_startupFetchSucceeded || !subscribed) {
  _scheduleStartupRetry();
} else {
  _startupRetryAttempt = 0;
}
```

Separately: retries are unbounded for the process lifetime. That's defensible for a sync that must eventually land, and at a 30s ceiling it's cheap — but it is a policy decision, so a one-line comment saying it's intentional would stop a future reader from "fixing" it into a bounded loop and silently reintroducing the bug this PR closes.

wpfleger96 (commented):
Reviewed at head `c5f1d9a38`. Combined and deduplicated feedback from three independent reviewers (Paul, Thufir, Duncan) — each formed findings separately before seeing the others.

This is a good PR: small, single-purpose, and the emulator reproduction in the description is exactly the right kind of evidence. Diagnosing the `catch (_)` swallowing as the reason the manager kept empty state forever is correct, and the retry is the right shape. No CRITICAL findings.

One IMPORTANT: the retry can fail to fire in the very scenario it targets, because `subscribe()` reports success on a readiness timeout. Details inline. Two MINORs on the backoff, also inline.

Worth stating what we verified rather than assumed — the `null`-vs-`false` distinction the retry rests on does hold. A rate-limited REQ produces a relay `CLOSED` (`crates/buzz-relay/src/connection.rs:655-679` routes every admission rejection through `request_rejection_message`, which emits `CLOSED` whenever a sub_id is present, and a REQ always has one), which `relay_session.dart:503-509` turns into a thrown error rather than an empty result. So a rate-limit can't masquerade as "relay has no sections blob." We specifically went looking for a hole there and there isn't one.

CI is green, and none of the findings below are CI-detectable.

mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
tellaho: Fixed in edb5bf369. The manager now passes `onClosed` (same pattern as `observer_subscription.dart`), clears `_unsubscribe`, and re-schedules `_syncWithRelay()` — option 1, as suggested, to keep the blast radius PR-local. Regression tests cover the exact scenario: a CLOSED landing after `subscribe()` reported success triggers exactly one replacement subscription (no duplicates, `activeListeners == 1`), remote state is adopted on the new subscription, and a late CLOSED after dispose schedules nothing.

*This reply was AI-generated (Bart, authorized by @tellaho).*

mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
tellaho: You guessed the reason right — it's overflow protection, now documented in 11a4957d2. The attempt counter is unbounded (intentionally, see the `_syncWithRelay` doc), and an unchecked `<<` past 62 wraps negative, which would turn the retry Timer into a hot loop. Kept the inner cap, added the comment explaining it.

*This reply was AI-generated (Bart, authorized by @tellaho).*

mobile/lib/features/channels/channel_sections/channel_sections_manager.dart
tellaho: Both taken in 0071e2d5a. `_startupRetryAttempt` resets to 0 once fetch + subscription have both succeeded, with a regression test proving a post-recovery failure retries at ~base delay instead of the climbed ceiling. The unbounded-retry policy is now documented on `_syncWithRelay` with an explicit warning not to bound it — giving up permanently is the exact bug this PR replaces.

*This reply was AI-generated (Bart, authorized by @tellaho).*

brow (approved):

PR #2353, opened 2026-07-22
fix(desktop): remove bundled libsystemd from AppImage

Chessing234:
## Summary
- AppImage bundles an older `libsystemd.so.0` that shadows the system copy after `libmount` is already removed
- on systemd ≥ 251 (Arch, Fedora 41+, etc.) launch fails with `LIBSYSTEMD_251` not found
- add `libsystemd.so*` to the existing removal list in `fix-appimage.sh`

Closes #2335

## Test plan
- [ ] rebuild / repack an AppImage with the updated script
- [ ] launch on a distro with systemd ≥ 251 without `LD_PRELOAD` workaround

Made with [Cursor](https://cursor.com)

PR #2500, opened 2026-07-23
fix: chart auto-migrate image, AppImage dmabuf, silent reply notice

Chessing234:
## Summary
- Chart 0.1.7 defaults `appVersion` to `main` so `BUZZ_AUTO_MIGRATE` works out of the box (`#2473`)
- AppImage sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` to avoid Intel Mesa / XWayland cores (`#2338`)
- buzz-acp posts a failure notice when a mention turn ends Ok with no agent channel publish (`#2459`)

## Test plan
- [ ] `helm template` / chart render: image is `ghcr.io/block/buzz:main` when tag is empty
- [ ] Linux AppImage on Intel Mesa + XWayland launches without dmabuf core dump
- [ ] Force `buzz messages send` to fail during a mention turn — channel gets a failure notice

Fixes #2473
Fixes #2338
Fixes #2459

PR #3005, opened 2026-07-26
fix(desktop): disable WebKit dmabuf renderer in AppImage shim

### LINKED ISSUE #2335: [Bug] `fix-appimage.sh` doesn't remove bundled `libsystemd.so.0` — AppImage crashes on distros with newer systemd (Arch, Fedora 41+, etc.)
## Describe the bug

The Buzz Desktop AppImage fails to launch on Linux distros running systemd ≥ 251 (Arch, Fedora 41+, etc.) with a dynamic linker error:

```
buzz-desktop: /tmp/.mount_Buzz_0AnNMee/usr/lib/libsystemd.so.0: version `LIBSYSTEMD_251' not found (required by /usr/lib/libmount.so.1)
```

The AppImage bundles a `libsystemd.so.0` from an older systemd (version tags `LIBSYSTEMD_246`–`LIBSYSTEMD_252`, with `251` missing). Because `AppRun` puts the bundled `usr/lib` first in `LD_LIBRARY_PATH`, the system `libmount.so.1` finds the bundled libsystemd instead of the system one and fails the version check.

## To Reproduce

1. Download `Buzz_0.4.22_amd64.AppImage` on a distro with systemd ≥ 251 (e.g. Arch, Fedora 41+)
2. Make it executable: `chmod +x Buzz_0.4.22_amd64.AppImage`
3. Run it: `./Buzz_0.4.22_amd64.AppImage`
4. See error:

```
buzz-desktop: /tmp/.mount_Buzz_0AnNMee/usr/lib/libsystemd.so.0: version `LIBSYSTEMD_251' not found (required by /usr/lib/libmount.so.1)
```

## Expected behavior

Buzz Desktop should launch successfully, using the system `libsystemd.so.0` instead of the bundled copy.

## Supporting Material

Workaround — preloading the system libsystemd forces it to take priority over the bundled copy and the app launches correctly:

```bash
LD_PRELOAD=/usr/lib/libsystemd.so.0 ./Buzz_0.4.22_amd64.AppImage
```

Version tags on bundled vs system libsystemd:

```
# Bundled (inside the AppImage at usr/lib/libsystemd.so.0):
$ strings libsystemd.so.0 | grep LIBSYSTEMD | sort -V
LIBSYSTEMD_246
LIBSYSTEMD_247
LIBSYSTEMD_248
LIBSYSTEMD_249
LIBSYSTEMD_252          ← jumps from 249 to 252, missing 250 and 251

# System (/usr/lib/libsystemd.so.0):
$ strings /usr/lib/libsystemd.so.0 | grep LIBSYSTEMD | sort -V
LIBSYSTEMD_246
...
LIBSYSTEMD_259
LIBSYSTEMD_260
LIBSYSTEMD_261          ← all version tags present
```

The system `libmount.so.1` requires `LIBSYSTEMD_251`, which the bundled libsystemd doesn't export:

```
$ ldd /usr/lib/libmount.so.1 | grep systemd
libsystemd.so.0 => /usr/lib/libsystemd.so.0
```

## Environment

- OS: Garuda Linux (Arch-based)
- Version: Buzz Desktop 0.4.22 (AppImage, amd64); system systemd 261

## Additional context

This is the same class of bundled-lib-shadowing-system-lib problem that `desktop/scripts/fix-appimage.sh` already addresses for libmount, libglib, libwayland, GStreamer, and others, but `libsystemd.so.0` is missing from the removal list. The existing script already removes the bundled `libmount.so` (so the system `libmount.so.1` is used), but the system `libmount` then needs the system `libsystemd`, which is shadowed by the bundled copy.

Fix: add `libsystemd.so` to the `rm -f` list in `fix-appimage.sh`:

```diff
   "$LIBDIR"/libmount.so \
   "$LIBDIR"/libblkid.so \
   "$LIBDIR"/libselinux.so \
+  "$LIBDIR"/libsystemd.so \
   "$LIBDIR"/libpcre2-8.so* \
```

The system `libsystemd.so.0` has a stable SONAME across all systemd versions and is guaranteed present on any Linux distro running systemd (i.e. all AppImage target distros), so removing the bundled copy is safe — the same reasoning already applied to the other libs in the removal list.

Related: #2315  (same class of bundled-lib-shadowing issue), #1567  (original `fix-appimage.sh` PR that missed `libsystemd`), #1856 .

### LINKED ISSUE #2338: [Bug] WebKitGTK `dmabuf` renderer core dump
## Describe the bug

After applying the libsystemd fix (#2335) and the GStreamer plugin path fix (PR #2176), the Buzz Desktop AppImage still crashes with a core dump on launch on systems using Intel Mesa graphics with rootless XWayland (e.g. KDE Plasma on Wayland with Intel iGPU). The process starts, prints its log lines, then dumps core.

Setting `WEBKIT_DISABLE_DMABUF_RENDERER=1` prevents the crash and the app launches and renders correctly.

## To Reproduce

1. On a system with Intel Mesa graphics + rootless XWayland (e.g. Garuda Linux / Arch with KDE Plasma Wayland, Intel RPL-S iGPU, Mesa 26.1.5)
2. Extract the AppImage and apply the libsystemd + GStreamer fixes (per #2335 and #2176)
3. Run the extracted AppRun: `./AppRun`
4. Buzz starts, prints log lines, then crashes with a core dump (no window appears)

## Expected behavior

Buzz Desktop should launch and render its window without crashing.

## Supporting Material

Without `WEBKIT_DISABLE_DMABUF_RENDERER=1`:

```
buzz-desktop: persisted identity pubkey 47a52a04ae052e0cb5663d6523cc1324d22f9ed962e1f428fb83668611563882
buzz-desktop: media proxy listening on 127.0.0.1:39387
buzz-desktop: repos dir resolved at boot — no configured override, REPOS is the default real dir
timeout: the monitored command dumped core
```
With `WEBKIT_DISABLE_DMABUF_RENDERER=1`:
```
buzz-desktop: persisted identity pubkey 47a52a04ae052e0cb5663d6523cc1324d22f9ed962e1f428fb83668611563882
buzz-desktop: media proxy listening on 127.0.0.1:40347
buzz-desktop: repos dir resolved at boot — no configured override, REPOS is the default real dir
```
App runs correctly — WebKitWebProcess starts, window renders at full size (2560x1048, maximized), no crash.

The `GDK_BACKEND=x11` hardcoded in the AppRun hook (`linuxdeploy-plugin-gtk.sh`) forces XWayland. The dmabuf renderer crash does not occur with `GDK_BACKEND=wayland`, but Wayland fails with `eglMakeCurrent failed` (which is why the AppRun hook hardcodes x11 — see tauri#8541).

## Environment

- **OS:** Garuda Linux (Arch-based), KDE Plasma on Wayland
- **GPU:** Intel RPL-S (integrated), Mesa 26.1.5
- **Display server:** Wayland with rootless XWayland (`Xwayland :1-rootless`)
- **WebKitGTK:** 2.52.5 (both bundled and system)
- **Buzz version:** 0.4.22 (AppImage, amd64)

## Additional context

This crash happens *after* the libsystemd (#2335) and GStreamer (PR #2176) fixes are applied. Without those fixes the app crashes even earlier, so the dmabuf issue is masked until the other two are resolved.

The `WEBKIT_DISABLE_DMABUF_RENDERER=1` env var is a well-known workaround for WebKitGTK dmabuf renderer crashes on certain GPU/driver/compositor combinations. It forces WebKitGTK to use the older shared-memory rendering path instead of dmabuf buffers.

Suggested fix: add `export WEBKIT_DISABLE_DMABUF_RENDERER=1` to the `apprun-hooks/linuxdeploy-plugin-gtk.sh` hook in the AppImage (or to `fix-appimage.sh` as a post-build patch), since the `GDK_BACKEND=x11` XWayland path already limits rendering to X11 — the dmabuf renderer provides no benefit on this codepath and only risks crashes on Mesa-based GPU setups.

Alternatively, the fix-appimage.sh script could detect Mesa/Intel and set the env var conditionally, but since the dmabuf renderer offers no advantage on the XWayland path (it exists for Wayland zero-copy page flips, which aren't used when `GDK_BACKEND=x11`), setting it unconditionally is safe.

Related: #2335 (libsystemd), #2176 (GStreamer plugin path), #2315 (LD_LIBRARY_PATH leak).

wpfleger96:
`WEBKIT_DISABLE_DMABUF_RENDERER=1` is now defaulted in the AppImage launcher shim, fixing the core dump on launch reported on Intel Mesa under rootless XWayland.

WebKitGTK's dmabuf render path aborts the process before any window appears on some Mesa drivers. Disabling it costs nothing in this bundle: linuxdeploy's GTK apprun-hook hardcodes `GDK_BACKEND=x11` (because the Wayland backend dies in `eglMakeCurrent` — [tauri-apps/tauri#8541](https://github.com/tauri-apps/tauri/issues/8541)), and the dmabuf renderer exists for Wayland zero-copy page flips. On the only backend the AppImage can use, it is pure crash surface.

### Why the shim, and why AppImage-only

The export goes in the existing `usr/bin/buzz-desktop` shim that `fix-appimage.sh` already installs — the one that strips `AppRun.wrapped`'s bundle-pointing `GST_PLUGIN_*` overrides. `AppRun.wrapped` execs that shim unconditionally and the script already guards its installation, so no new file and no new assumption about hook sourcing.

Scoping is deliberate: not set in Rust or `tauri.conf.json`, because the `.deb` runs on native Wayland where dmabuf does work and disabling it there would be a rendering regression.

`${WEBKIT_DISABLE_DMABUF_RENDERER:-1}` means an explicit user value survives, including `0` — only unset or empty defaults to `1`.

The `fix-appimage.sh` root-cause header also gains a block for the `libsystemd` removal (#2335 / #2353), which the file's own convention requires for every removal, and its stale opening summary is updated to cover the systemd >= 251 failure set.

Closes #2338

`#2353` fixes #2335 in the same file — the two are conflict-free (dry-merged with `git merge-tree`) but touch adjacent lines, so whichever lands second may want a look. PR #2500 is prior art for this same fix; this goes narrow because #2500 bundles it with unrelated `buzz-acp` and helm changes that need their own review, and its `AppRun` string-injection fallback is not needed once the shim is used. Credit to @Chessing234 for diagnosing all three.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -189,7 +189,12 @@ The Linux AppImage is post-processed by `desktop/scripts/fix-appimage.sh`,
 which strips infra libraries over-bundled by linuxdeploy (they crash on
 Mesa 25+ / GLib 2.88 distros; see
 [tauri-apps/tauri#15665](https://github.com/tauri-apps/tauri/issues/15665))
-and re-signs the artifact. As a result the AppImage relies on the
+and re-signs the artifact. The script also installs a launcher shim that
+repairs the environment linuxdeploy's `AppRun` sets up — it restores host
+GStreamer plugin discovery and defaults
+`WEBKIT_DISABLE_DMABUF_RENDERER=1`, whose renderer crashes on some Mesa
+drivers and buys nothing on the X11 backend the AppImage is pinned to.
+As a result the AppImage relies on the
 host's Wayland/GStreamer/graphics stack and requires GLib >= 2.72
 (Ubuntu 22.04 or newer). The `release-linux` job builds inside a
 `ubuntu:22.04` container for broad GLIBC compatibility.
--- a/desktop/scripts/fix-appimage.sh
+++ b/desktop/scripts/fix-appimage.sh
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
-# fix-appimage.sh — Remove infra libs from a Tauri-produced AppImage that crash
-# on Mesa 25+ / GLib 2.88 distros (Ubuntu 26.04, Fedora 42+, etc.).
+# fix-appimage.sh — Post-process a Tauri-produced AppImage so it runs on the
+# host graphics/media stack: remove infra libs that crash when they skew against
+# the host's (Mesa 25+ / GLib 2.88 distros such as Ubuntu 26.04 and Fedora 42+;
+# systemd >= 251 distros such as Arch and Fedora 41+) and repair the launch
+# environment linuxdeploy's AppRun leaves behind.
 #
 # Usage: fix-appimage.sh <path-to.AppImage>
 #
@@ -12,7 +15,8 @@
 # avoid appimagetool fetching one from its mutable `continuous` tag (CI pins
 # this; unset is fine for local testing).
 #
-# Root cause — three interlocking failures (upstream: https://github.com/tauri-apps/tauri/issues/15665):
+# Root cause — five distinct failures the bundle exhibits (1-3 upstream:
+# https://github.com/tauri-apps/tauri/issues/15665):
 #
 #  1. EGL crash: linuxdeploy bundles libwayland-client.so.0 (1.22) alongside
 #     the app. Mesa 25's libEGL calls the bundled version at runtime; the version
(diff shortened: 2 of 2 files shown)


wpfleger96:
🤖 Converting to draft — holding this pending a hardware test from the #2338 reporter, [asked here](https://github.com/block/buzz/issues/2338#issuecomment-5084713785).

Review turned up that `WEBKIT_DISABLE_DMABUF_RENDERER=1` is a wider hammer than this PR's rationale (and my own review brief) claimed. In WebKitGTK 2.52.5 the env check returns from `rendererBufferTransportMode()` before `SharedMemory` is added to the transport set ([`AcceleratedBackingStore.cpp:82-116`](https://github.com/WebKit/WebKit/blob/webkitgtk-2.52.5/Source/WebKit/UIProcess/gtk/AcceleratedBackingStore.cpp#L82-L116)); the empty set fails `checkRequirements()`, `HardwareAccelerationManager` reports no hardware acceleration, and [`WebPreferencesGtk.cpp:41-49`](https://github.com/WebKit/WebKit/blob/webkitgtk-2.52.5/Source/WebKit/UIProcess/gtk/WebPreferencesGtk.cpp#L41-L49) then disables accelerated compositing and threaded scrolling. So the "dmabuf buys nothing under `GDK_BACKEND=x11`" framing in the script comments, `RELEASING.md`, and this description is wrong — the cost lands on every AppImage user, not just the ones hitting the crash.

`WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` looks like the narrow cut: `SharedMemory` is added on line 98 before the `FORCE_SHM` early return on line 101, so compositing stays enabled and only `RendererBufferTransportMode::Hardware` — the dmabuf import that crashes — is excluded. Whether it avoids the actual crash needs the reporter's Intel Mesa + rootless XWayland setup; we can't reproduce it.

Two exits from here:
- `FORCE_SHM=1` works → swap the export and correct the rationale sites.
- it doesn't → keep `WEBKIT_DISABLE_DMABUF_RENDERER=1` and rewrite the rationale sites to state the real cost.

The block-`4.` inconsistency flagged in review is resolved: #2353 merged first, so `libsystemd.so*` is now in the removal list on `main` and the header comment describes something the script does.

wpfleger96:
🤖 Closing this in favor of the full fix: a crash-recovery ladder that relaunches with progressively safer rendering profiles (starting with `WEBKIT_DISABLE_DMABUF_RENDERER`) only on machines that actually crash, plus a `--safe-rendering` flag. This blanket default would've cost working GPUs their hardware rendering, and the recovery work found it would also contaminate the ladder's env handoff — the recovery child re-executes this shim, so the export would override every gentler profile the ladder tries. The design spike for the ladder is done and reviewed; the implementation PR is starting now and will close #2338.

PR #3023, opened 2026-07-26
fix(desktop): use forward slashes for git credential.helper on Windows

### LINKED ISSUE #3025: [Bug] Windows: Projects Remote file tree empty — git-credential-nostr path mangled by MinGW bash
**Describe the bug**
On Windows, the Projects **Remote** file view is empty and Sync/Clone fails. Authentication never succeeds because the absolute `git-credential-nostr` path Buzz injects into `credential.helper` uses backslashes, which Git for Windows' MinGW bash treats as escape sequences and strips.

Example error:

```
Could not fetch repository.
Cloning into 'C:\Users\<user>\AppData\Local\Temp\tmpXXXXX\repo'...
C:\Users\<user>\AppData\Local\Buzz\git-credential-nostr.exe get: line 1:
C:Users<user>AppDataLocalBuzzgit-credential-nostr.exe: command not found
fatal: could not read Username for 'https://<relay>/git/<pubkey>/<repo>':
terminal prompts disabled
```

The same relay/repo works on macOS without extra configuration.

**Root cause (short)**
Buzz sets `credential.helper` from `Path::display()`, which on Windows yields `C:\...\git-credential-nostr.exe`. Git for Windows invokes helpers via `sh.exe`; bash mangles the path → helper not found → NIP-98 auth fails → Remote blobless clone fails → empty tree.

Projects git also sets `GIT_CONFIG_GLOBAL=/dev/null`, so user `~/.gitconfig` workarounds do not apply.

This is **not** a missing/stub helper binary — the shipped `git-credential-nostr.exe` is functional once the path is passed with forward slashes.

**Steps to reproduce**
1. Install Buzz on Windows (Git for Windows required)
2. Connect to a Buzz relay that has a repository with at least one pushed branch
3. Open **Projects** → select the repository → switch to **Remote**
4. Observe an empty file tree
5. Press **Sync** / **Clone** and observe the mangled-path / `command not found` error above

**Expected behavior**
Remote view shows the repository file tree, and Sync/Clone authenticates via `git-credential-nostr` the same as on macOS.

**Version and platform**
- Buzz version: local build from fix branch / 0.4.26 NSIS (also reproduces on prior Windows installers)
- OS: Windows 11
- Git: Git for Windows 2.55.x

**Logs / additional context**
Reproduced by simulating Buzz's `GIT_CONFIG_*` injection:

- Backslash helper path → exact `command not found` mangling
- Forward-slash helper path (`C:/Users/.../git-credential-nostr.exe`) → helper runs successfully

**Fix PR**
https://github.com/block/buzz/pull/3023

bcrdejong:
## Problem

On Windows, Projects **Remote** view shows an empty file tree and Sync/Clone fails with a mangled credential-helper path, for example:

```
C:\Users\<user>\AppData\Local\Buzz\git-credential-nostr.exe get: line 1:
C:Users<user>AppDataLocalBuzzgit-credential-nostr.exe: command not found
fatal: could not read Username for 'https://<relay>/git/...': terminal prompts disabled
```

Buzz injects an absolute path into `credential.helper` via `Path::display()`. On Windows that yields backslashes. Git for Windows runs credential helpers through MinGW bash, which treats `\` as escapes and destroys the path, so NIP-98 auth never runs and the blobless temp clone behind Remote view fails.

macOS/Linux are unaffected (paths already use `/`).

This is unrelated to shipping a stub helper - the bundled `git-credential-nostr.exe` is a real binary. User `~/.gitconfig` workarounds also cannot help here because Projects git sets `GIT_CONFIG_GLOBAL=/dev/null` and injects its own helper.

Closes #3025

## Fix

Normalize the helper path to forward slashes before writing `GIT_CONFIG_VALUE_*`:

- `desktop/src-tauri/src/commands/project_git_exec.rs` (Projects Remote / Sync)
- `desktop/src-tauri/src/managed_agents/runtime.rs` (agent spawn git auth)

Forward slashes are accepted by Git on every platform; on macOS/Linux the replace is a no-op. No `cfg(windows)`, packaging, or libgit2 changes.

## How to reproduce (before)

1. Install Buzz on Windows with Git for Windows
2. Connect to a relay that has a repository with at least one pushed branch
3. Open **Projects** -> select the repo -> **Remote**
4. Observe empty tree; Sync/Clone shows the mangled-path / `command not found` error above

## Test plan

- [x] Unit: `cargo test --manifest-path desktop/src-tauri/Cargo.toml credential_helper_config_value` (formatter covered on all platforms; no-op for Unix-style paths)
- [x] Local Windows NSIS build + install of this branch
- [x] Projects -> Remote / Sync against a Buzz relay repo succeeds on Windows after the fix


noobomancer:
Can confirm this fixes #3025. Tested on Windows 10 (10.0.19045): built desktop v0.4.26 from main (87b3fcd3) with this PR's commit cherry-picked, MSVC toolchain, sidecars bundled. Without the patch the project file tree comes up empty as described in the issue; with it, fetching and browsing a repo from our self-hosted relay (v0.2.0) works without errors. The new `credential_helper_config_value_uses_forward_slashes` test also passes locally.

bcrdejong:
Fixed Rust Lint: rustfmt wanted different wrapping on the `PathBuf::from(...)` line in the new unit test. No behavioral change - `cargo fmt --check` and the unit test pass locally.

wpfleger96 (approved):

wpfleger96:
🤖 Follow-up note for anyone hitting this later: there's a pre-existing issue adjacent to this fix that the PR intentionally didn't touch. Git parses the `credential.helper` value as a shell command line, so a helper path containing spaces gets word-split — reproducible with `git -c credential.helper="/path with spaces/helper"`, which fails with `is not a git command`. The forward-slash normalization here (`credential_helper_config_value` in `project_git_exec.rs` and the `GIT_CONFIG_VALUE_0` write in `runtime.rs`) doesn't quote the path, so a buzz install under a spaced directory (e.g. a Windows username with a space, depending on where the helper lands) would still break. The fix is quoting the value (`"'{path}'"` or shell-escaping) at both call sites — worth its own small PR rather than piling onto this one.

Issue #2682, opened 2026-07-24
feat(client): browser-based web UI to drive agents from any device (incl. mobile)


michabbb:
## Problem

The only human client today is the Tauri desktop app, so I can only use the workspace at a machine where it's installed. The relay is already headless and self-hostable -- I run it on a Linux server with no GUI -- but there's no way to actually use it from that server, or from my phone.

Agents don't keep office hours. When one posts a review or hits an approval gate at 11pm, I want to glance at my phone, read the thread, and approve -- not be tied to the one desktop where the app lives. Steering agents from anywhere is exactly what an agent workspace should nail, and right now the client is the bottleneck.

## Proposal

A browser-accessible, mobile-friendly web UI that talks to the relay over the same WebSocket + REST surface the desktop app already uses -- enough to read channels, reply, react, and approve/deny workflow gates from any device.

I know the current React/Vite frontend is coupled to Tauri APIs (`invoke`, native binaries like `buzz-acp`), so it can't just be served in a browser as-is. A lighter web client focused on the read + reply + approve core would already cover most of the need; git hosting and local shell tools can stay desktop-only.

## Why it matters

- Self-hosters get a usable client on the same box as the relay.
- Mobile access turns async agent work into "approved it from my phone."
- Share a URL instead of a per-platform desktop install.

Happy to help test on self-hosted Linux + mobile.

Giszmo:
I'd like to support this feature request, too. People might be hesitant to install an app but visiting a website is very low commitment. The web version doesn't need to expose all features and could tease the user with buttons that don't work on the web version but tell the user to get the real thing.

I personally am very hesitant to install native apps on desktop because of the inferior sandboxing there. I'm happy to use mobile apps or web apps though.

DanTup:
I'd like to see this too. I don't want the agents on my machine (for multiple reasons, including I want to be able to shut my machine down, and I want them in sandboxes I can control - such as OpenShell). With a web app I know there's no possibility anything is ever running on my machine. If I want it out the browser, I could install it as a PWA.

tianheil3:
### Self-host + headless server use case (👍)

Same pain: relay on a **Debian box without GUI**, want a **browser shell** as the human client (not Tauri Desktop). Agents can stay on the server via `buzz-acp` / future spawner (#2859); humans only need read/reply/@ from any browser.

Current `web/` package in-tree is still closer to **repo browser** than a Desktop-parity channel UI — so this issue remains the right track for a real web client shell.

[...]

PR #3027, opened 2026-07-26
feat(relay): BUZZ_WEB_SPA=full to serve a full client bundle

flaukowski:
## Problem

`should_serve_spa` only falls back to the SPA shell for the invite landing page, and optionally the bundled git browser:

```rust
fn should_serve_spa(path: &str, serve_git_web_gui: bool) -> bool {
    is_invite_landing_path(path) || (serve_git_web_gui && is_git_web_gui_path(path))
}
```

A deployment that points `BUZZ_WEB_DIR` at a full client bundle has no way to get client-side routing for its own paths — every route needs a bespoke predicate compiled into the router. Any client-side route reached by deep link or refresh 404s.

## Change

Add `BUZZ_WEB_SPA=full`. When set, any path the relay does not serve itself falls back to `index.html`.

Server-owned paths still return their own responses, so a mistyped API call 404s instead of returning HTML that a client would then try to parse as JSON:

`/api/…`, `/git/…`, `/hooks/…`, `/media/…`, `/upload`, `/.well-known/…`, `/_*`, `/events`, `/query`, `/count`, `/info`, `/health`, `/moderation`

Defaults to false. With the variable unset, behaviour is unchanged.

## Tests

Three new cases in `crates/buzz-relay/src/router.rs`:

- `full_spa_mode_serves_arbitrary_client_routes` — opt-in serves paths the relay has never heard of
- `full_spa_mode_still_lets_server_paths_404` — every server-owned prefix must not fall back
- `full_spa_mode_is_off_by_default` — unknown paths still 404 without the opt-in

Plus a default-value assertion in the config tests, and a `TESTING.md` row for the new variable.

## Context

We run a self-hosted Buzz deployment whose web bundle is a full workspace client. We had been carrying a hardcoded path predicate in `should_serve_spa` to make it work; this generalises that into an opt-in flag so the fix isn't per-route, and lets us drop the local patch.

## Related work

Searched open PRs and issues for SPA fallback / `should_serve_spa` / `BUZZ_WEB_DIR` routing — none found.

Rebased onto current `main`; the three new router tests plus `cargo test -p buzz-relay --lib` are clean.

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -284,6 +284,7 @@ out of the box with `just setup` or `just relay`. Common overrides:
 | `BUZZ_ALLOW_NIP_OA_AUTH`        | `false`                     | Enable NIP-OA owner attestation for membership |
 | `BUZZ_WEB_DIR`                  | unset (source), `/srv/buzz/web` (container) | Directory containing the invite landing bundle; the production container enables it so `/invite/{code}` always works |
 | `BUZZ_SERVE_GIT_WEB_GUI`        | `false`                     | Set to `true` or `1` to expose the bundled Git repository browser at `/` and `/repos/...`; invite routes do not depend on this flag |
+| `BUZZ_WEB_SPA`                  | unset                       | Set to `full` when the configured `BUZZ_WEB_DIR` bundle is a full client that owns client-side routing: every path the relay does not serve itself then falls back to the SPA shell. Server-owned paths (`/api/...`, `/git/...`, `/hooks/...`, `/media/...`, `/upload`, `/.well-known/...`, `/_*`, `/events`, `/query`, `/count`, `/info`, `/health`, `/moderation`) still return their own responses, so a mistyped API call 404s instead of returning HTML |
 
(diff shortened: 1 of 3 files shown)


flaukowski:
@ThePumpingLemma — you touched `router.rs` most recently, so flagging this your way; apologies if it should go to someone else.

Rebased onto current `main` today. The three new router tests plus `cargo test -p buzz-relay --lib` pass locally.

The flag defaults to `false`, so behaviour is unchanged unless `BUZZ_WEB_SPA=full` is set explicitly. On our side it replaces a hardcoded path predicate we'd been carrying downstream, which is what prompted generalising it.

`CI` and `Docker image` are at `action_required` — they need a maintainer to approve the workflow run before anything executes.

cgray2222:
This is the piece that makes #2682 work for self-hosted relays. A `BUZZ_WEB_SPA=full` option serving the client bundle from the relay would let a headless community be used from any browser with zero client install.

PR #3036, opened 2026-07-26
fix: p-tag the parent author on threaded replies (NIP-10)

### LINKED ISSUE #3032: Threaded replies omit the NIP-10 p tag for the parent author — replying to someone never notifies them
## Summary

A threaded reply carries NIP-10 `e` tags (root/reply markers) but **no `p` tag for the author being replied to**. Buzz notifications are `#p`-based, so replying to someone in a thread never notifies them unless you *also* type their `@name`. In a thread you are, by definition, replying to a specific person — they should be notified.

This is the same class as #2568 (reactions dropped the `p` tag); @vitorpamplona flagged it as a follow-up on that PR. NIP-10 says a reply's `p` tags SHOULD list the pubkeys being replied to; Buzz emits the `e` markers but omits the matching `p`.

Confirmed on `main` @ `499c5d34`.

## Where the `p` tag is dropped

Every reply builder emits `e` tags for threading but sources `p` tags **only** from explicit `@mentions`:

- **SDK** — `crates/buzz-sdk/src/builders.rs:173` `thread_tags` pushes only `["e", …, "reply"]` / `["e", …, "root"]`. `p` tags come solely from `mention_tags` (builders.rs:188), i.e. parsed `@mentions`. `build_message` (kind 9) threads these two independently.
- **Desktop** — `desktop/src-tauri/src/commands/messages.rs:432` `resolve_thread_ref` already fetches the parent event (to read its root), so `parent.pubkey` — the signer — is right there, but it is discarded. `send_channel_message` p-tags only the composer's `@mention_pubkeys`.
- **Mobile** — `mobile/lib/features/channels/send_message_provider.dart:56` builds `_buildReplyTags` (e-tags) plus `p` tags from `normalizedMentions` (explicit mentions) only.

## Why it's a real notification miss

Notification/feed reads are `#p`-only, with no fallback that derives the parent author from the reply `e`-tag:

- Desktop feed: `{"kinds":[9,40002,1,45001,45003],"#p":[self]}` (`commands/messages.rs`).
- Mobile activity: same kind set, `#p:[self]` (`activity_provider.dart`).
- Push leases advertise kind 9 and subscribe by `#p` (`push_lease.rs`).
- Relay ingest updates thread counters but never injects a `p` tag or notifies the parent author from the `e`-tag (grep of `ingest.rs`/`side_effects.rs`).

So: no `p` tag → the replied-to author gets no feed entry and no push.

## Reproduce

1. User A posts a message in a channel.
2. User B opens the thread on that message and replies **without** typing `@A`.
3. The kind:9 reply carries `[["h",…],["e",<root>,"","root"],["e",<A's msg>,"","reply"]]` and no `["p", <A>]`.
4. User A's mentions feed (`#p:[A]`) does not include the reply; no push fires.

Expected: A is notified that B replied to them.

## Suggested fix

When building a threaded reply, add the parent event's **signer** as a `p` tag (deduped against existing mentions, self excluded) — mirroring the fix approach in #2634 (use the signer, not the resolved display author):

- Desktop: `resolve_thread_ref` already holds `parent.pubkey`; thread it into the reply's `p` tags — no extra query.
- Mobile: the thread compose bar has the `threadHead` message; pass its author through.
- SDK/CLI/ACP: carry the parent author alongside the existing `ThreadRef`.

## Note / product check

This is standard chat behavior (Slack/Discord notify the parent on a thread reply) and NIP-10-conformant, so I'm treating it as a bug. If the team would rather gate reply-notifications behind a preference, say so and I'll adjust — but the `p` tag being absent from the wire form is a conformance gap regardless.

I have a fix in progress across all surfaces; PR to follow.

shani-singh1:
Fixes #3032.

## Problem

A threaded reply carries the NIP-10 `e` markers (`root`/`reply`) but no `p` tag for the author being replied to. Buzz notifications are `#p`-based, so replying to someone in a thread never notifies them unless you *also* type their `@name`. In a thread you are by definition replying to a specific person — they should be notified.

Same class as #2568/#2634 (reactions dropped the `p` tag); @vitorpamplona flagged it as the reply-side counterpart. NIP-10 says a reply SHOULD name the pubkeys it replies to.

The `p` tag came only from explicit `@mentions` on every surface: the SDK `thread_tags` (builders.rs), the desktop `send_channel_message`, and mobile `send_message_provider`/`createForumReply`. Notification reads are `#p`-only with no `e`-tag fallback (desktop feed, mobile activity, push leases), and the relay never injects a parent `p` tag on ingest — so no `p` tag genuinely means no notification.

## Fix

Carry the parent event's **signer** as the reply's `p` tag, deduped against explicit mentions and self-excluded — consistent with the "use the signer, not the display author" decision from #2634.

**SDK** (`buzz-sdk`) — `ThreadRef` gains `parent_author: Option<PublicKey>`. `thread_tags` emits its `p` tag; `mention_tags` now seeds its dedup set from `p` tags already on the event, so an explicit `@mention` of the same pubkey stays a single tag. Covers `build_message` (kind 9), `build_forum_comment` (45003), and `build_diff_message` (40008). A self-reply's `p` tag is scrubbed by nostr's default self-tag handling (the builder does not opt into self-tagging), so no self-notification.

**CLI** (`buzz-cli`) — `resolve_thread_ref` already fetches the parent event, so its signer populates `parent_author` at no extra cost. This also covers agent replies, which post via the buzz CLI.

**ACP** — `setup_mode` already p-tags the asker via an explicit mention (unchanged); the dead-letter failure notice is a system message (`parent_author: None`).

**Desktop** — `resolve_thread_ref` already fetches the parent (to read its root), so `parent.pubkey` rides along as `parent_author` — no extra query. `build_message_with_client_tags` and `build_forum_comment` merge it into the mention set (deduped, self-scrubbed).

**Mobile** — the thread compose bar and forum thread page hold the parent message, so its `pubkey` is threaded through `SendMessage.call` and `createForumReply`; both prepend it to the mention set, which already dedupes and drops self.

## Behaviour

- Reply to someone in a thread → they get a mentions-feed entry and a push, without needing an explicit `@`.
- Reply to your own message → no self-notification (self `p` tag scrubbed / excluded).
- Reply that also `@`-mentions the parent → a single `p` tag, not a duplicate.

## Product note

This is NIP-10-conformant and matches Slack/Discord thread-reply behaviour, so I've treated it as a bug. If the team would rather gate reply-notifications behind a preference, the change is small to adjust — but the missing `p` tag is a wire-conformance gap regardless. Raised in #3032.

## Validation

Toolchain note: built on Windows with `x86_64-pc-windows-gnu` (no MSVC linker locally).

- `cargo test -p buzz-sdk --lib` — **238 passed**, including three new tests: the reply carries the parent `p` tag, it is not duplicated when also `@`-mentioned, and a self-reply drops it.
- `cargo check`/`--tests` on `buzz-cli`, `buzz-acp`, and `desktop/src-tauri` — clean (0 errors). `cargo test` on the desktop crate can't link locally (`sherpa-onnx-sys` native lib), so CI-Linux runs those.
- `cargo fmt --check` and `clippy -D warnings` on the workspace crates + desktop rustfmt — clean.
- Mobile (Dart) is unverified locally (no toolchain that bootstraps on Windows); the change is mechanical (one param, prepended to an existing dedup) and CI's Mobile job covers it.

Independent of #2634 — no file overlap.

### DIFF
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -3885,6 +3885,8 @@ pub(crate) async fn post_failure_notice(
         Some(buzz_sdk::ThreadRef {
             root_event_id: root_id,
             parent_event_id: parent_id,
+            // System dead-letter notice, not a reply to a person.
+            parent_author: None,
         })
     });
     let builder =
--- a/crates/buzz-acp/src/setup_mode.rs
+++ b/crates/buzz-acp/src/setup_mode.rs
@@ -611,12 +611,16 @@ async fn publish_setup_nudge(
         Some(ThreadRef {
             root_event_id: root_id,
             parent_event_id: root_id,
+            // Already p-tagged as the asker via the explicit mention below.
+            parent_author: None,
         })
     } else {
         // Top-level event: reply to the triggering event.
         Some(ThreadRef {
             root_event_id: triggering_event.id,
             parent_event_id: triggering_event.id,
+            // Already p-tagged as the asker via the explicit mention below.
+            parent_author: None,
         })
     };
 
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -77,9 +77,17 @@ async fn resolve_thread_ref(
         _ => parent_eid,
     };
 
+    // NIP-10 reply `p` tag: the parent event is already fetched, so its signer
+    // rides along and the reply notifies the author it replies to.
+    let parent_author = event
+        .get("pubkey")
+        .and_then(|v| v.as_str())
+        .and_then(|hex| nostr::PublicKey::parse(hex).ok());
+
     Ok(ThreadRef {
         root_event_id: root_eid,
         parent_event_id: parent_eid,
+        parent_author,
     })
 }
 
--- a/crates/buzz-sdk/src/builders.rs
+++ b/crates/buzz-sdk/src/builders.rs
@@ -186,15 +186,37 @@ fn thread_tags(thread_ref: &ThreadRef, tags: &mut Vec<Tag>) -> Result<(), SdkErr
         tags.push(tag(&["e", &root, "", "root"])?);
         tags.push(tag(&["e", &parent, "", "reply"])?);
     }
+    // NIP-10: a reply SHOULD `p`-tag the author it replies to, so the parent
+    // author's `{"kinds":[9],"#p":[self]}` notification filter matches. Deduped
+    // against explicit mentions by `mention_tags`. A self-reply's `p` tag is
+    // scrubbed by nostr's default self-tag handling, which is what we want (no
+    // self-notification).
+    if let Some(author) = thread_ref.parent_author {
+        tags.push(tag(&["p", &author.to_hex()])?);
+    }
     Ok(())
 }
 
 /// Deduplicate and cap mentions, emitting p-tags.
+///
+/// Seeds its dedup set from `p` tags already on `tags` (e.g. a reply's parent
+/// author emitted by [`thread_tags`]) so an explicit mention of that same
+/// pubkey does not produce a duplicate `p` tag.
 fn mention_tags(mentions: &[&str], tags: &mut Vec<Tag>) -> Result<(), SdkError> {
     if mentions.len() > crate::mentions::MENTION_CAP {
         return Err(SdkError::TooManyMentions);
     }
-    let mut seen = std::collections::HashSet::new();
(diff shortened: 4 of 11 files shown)


gregp-squareup:
@shani-singh1  Thanks for this PR.  I ran into the behavior today this PR addresses. I’m not a Buzz maintainer, so I can’t speak to merge status or project direction, but I noticed the branch now has conflicts with main. 

If you’re still interested in revisiting it, I’d be interested to see an updated diff. No pressure, and thanks again for the contribution.

PR #3053, opened 2026-07-27
fix(mobile): recover and pace live subscriptions

brow:
Tl;dr: avoid a thundering herd of per-channel resubs when reconnecting to relay, which the mobile client has to do most times it is foregrounded.

The burst of resubs would get rate limited, causing some of the subs to fail without recovery, and thus local state of channels to silently fall behind.

This makes mobile follow desktop’s strategy for pacing reconnections, using a prioritized queue.

### What changed?

Mobile now recovers live subscriptions after retryable or rate-limited relay `CLOSED` responses. It ports the existing desktop model: classify terminal versus retryable closures, honor retry hints through a session-owned rate-limit gate, retry with bounded backoff, and replay visible-channel subscriptions first in bounded batches.

Channel refreshes also retain unchanged live subscriptions instead of clearing and recreating them. This is desktop parity, not a new relay policy.

### Why?

On reconnect or resume, mobile replayed its retained live subscriptions while `channelsProvider` independently cleared and recreated roughly the same set, alongside unread catch-up and open-channel requests. The relay allows 50 REQs per 5 seconds, so users in many channels could predictably exceed the budget. In live reproduction, 55 subscriptions produced 9 rate-limit closures, 60 produced 18, and 80 produced 36.

Mobile then treated every live `CLOSED` as terminal, removed the affected subscription, and never restored it. Channel updates could remain dead until a later session reconstruction. This is the primary causal chain behind [BOT-1449](https://linear.app/squareup/issue/BOT-1449/buzz-mobile-posted-messages-dont-appear-until-leavingre-entering-the).

Desktop already handles this as normal transient pressure by classifying closures, gating and backing off retries, pacing reconnect replay, and retaining unchanged subscriptions. This change brings mobile to the same recovery model while removing the avoidable request burst.

### How is it tested?

Full mobile suite: 721 passed, 1 skipped. Analyzer and formatting checks pass. Required CI checks pass.

Added and updated tests cover `CLOSED` classification, retry hints, rate-limit gating, bounded retry and reset behavior, terminal failures, timer cleanup, history gating, visible-first batched replay, and retention of unchanged subscriptions.


brow:
@builderbot review

brow:
@codex review

wesbillman (dismissed):
**P2 — Restore the previously visible channel when a nested channel route is popped.**

`ChannelDetailPage` sets the session's singleton visible channel on mount, then unconditionally clears it on disposal. Channel links push a second `ChannelDetailPage` over the still-mounted first page. In an A → B → back-to-A flow, disposing B writes `null`; A becomes visible again but never remounts, so its effect does not rerun. A subsequent reconnect therefore loses A's first-batch replay priority in `_replayLiveSubscriptions`.

Please make visibility route-aware (or restore the prior active channel) rather than treating every page disposal as “no channel visible,” and add a navigation-stack regression test covering A → B → pop → reconnect ordering. The current relay test only calls `setVisibleChannelId` directly, so it cannot catch this lifecycle failure.

Reviewed at exact head `fc2f4e5d928d89a1dc7b59de59d8527175af4c63`. Mobile CI and all required checks are green; I did not duplicate CI locally.

brow:
🤖 Addressed the nested navigation lifecycle: visible-channel priority is now owner-scoped, so popping B releases only B and restores A for reconnect replay. Added regression coverage for A → B → pop → reconnect, route replacement, nested ownership, and stale or out-of-order disposal.

brow:
🤖 Posted by tomb's agent.

> **Responding to:** a question raised during the team's independent Buzz review of this PR ("Sami Q1"): *"HTTP bridge `queryRelay` not consulting the rate-limit gate (desktop gates both) — fix or justify."* Source thread (internal Buzz): buzz://message?channel=ce6ec97e-b85e-4c84-8992-938a40185053&id=ae1ea243a1b115da61e813597c8e0759745c8cd7e0bbe10f8e475faaaa300ba9

## Re: `queryRelay` does not consult the rate-limit gate

Investigated. We are making a change here, but not the one suggested, and the reasoning matters because the premise behind the suggestion does not hold.

### The premise, corrected

The review reads "desktop gates both paths" as one gate over one shared budget. Desktop actually uses **two independent gates**:

- WebSocket REQs are gated in TypeScript: `desktop/src/shared/api/relayGateBoundary.ts:32,75`.
- The HTTP bridge is gated in **Rust**, in a separate gate with its own expiry state: `desktop/src-tauri/src/relay_admission.rs` (`activate_rate_limit` / `wait_for_rate_limit`).
- The two are only loosely coupled, one-way: a Rust-side 429 arms the TS gate via `applyTauriRateLimitIfNeeded` (`desktop/src/shared/api/tauri.ts:296-301`).

Mobile has no equivalent of that second gate at all, which is the actual gap.

### The two paths bill to different relay buckets

This is the load-bearing part. A `queryRelay` call cannot consume the WebSocket budget that the reconnect pacing in this PR is protecting:

| Path | Limit type | Window | Default limit |
| --- | --- | --- | --- |
| WS `REQ` | `LimitType::WsEvents` | 5s | `human_ws_events_per_sec` 10, so 50 per 5s |
| HTTP `POST /query` | `LimitType::ApiCalls` | 60s | `human_api_calls_per_min` 300 |

- WS: `crates/buzz-relay/src/admission.rs:9,40-44` and `crates/buzz-relay/src/connection.rs:613-623`, limit from `crates/buzz-auth/src/rate_limit.rs:94-95,116-118`.
- HTTP: `crates/buzz-relay/src/api/bridge.rs:28-36`, enforced for `/query` at `bridge.rs:959`, limit from `crates/buzz-auth/src/rate_limit.rs:91-92,113-115`.

They are distinct Redis keys by construction, not by coincidence: `rate_limit_key()` suffixes the key on `LimitType::key_suffix()`, which is `"ws"` versus `"api"` (`crates/buzz-auth/src/rate_limit.rs:71-78,201-208`).

### Why we rejected making `queryRelay` await the gate

Awaiting the gate on the HTTP path would couple a 300-per-minute bucket to a 50-per-5-seconds bucket. A WebSocket rate-limit stall (up to the gate's 300s cap) would then block unrelated user-visible reads: channel windows (`mobile/lib/features/channels/channel_messages_provider.dart:170`), mention candidates (`mention_candidates_provider.dart:34`), thread replies (`thread_replies_provider.dart:41`), and channel management lookups (`channel_management_provider.dart:255,275,332,606`). That is a user-visible regression bought for no measured benefit, since the HTTP calls were never spending the WS budget in the first place.

### What we did change

The real defect is the opposite direction: mobile never arms the gate from HTTP back-pressure. `queryRelay` throws `RelayException` on any non-2xx (`mobile/lib/shared/relay/relay_session.dart:177-179`), and before this change the only two `activate()` calls in `mobile/lib` were both in `_handleClosed`. The relay does emit a retry hint over HTTP (`crates/buzz-relay/src/api/bridge.rs:41-45`) and mobile was discarding it.

So a rate-limited HTTP response now arms the shared gate with the relay's `retry in Ns` hint, matching desktop's deliberate cross-domain back-off intent (`desktop/src-tauri/src/relay_admission.rs:13-18`), while `queryRelay` itself stays non-blocking.

Two details worth flagging for anyone reading the diff, both of which would silently defeat a naive version:

1. The error body is JSON-wrapped, not a bare string: `api_error` returns `{"error": "<msg>"}` (`crates/buzz-relay/src/api/mod.rs:19-21`), so a `startsWith("rate-limited:")` check on the raw body never matches.
2. There are two rate-limited shapes, not one: `429` carries `rate-limited: quota exceeded; retry in {N}s`, and `503` carries `rate-limited: shared admission unavailable` with no hint (`bridge.rs:41-53`). Classifying on status code alone misses the 503, so we classify on the message with the existing `classifyRelayClosed()` and fall back to the gate's 10s default when no hint is present.

tlongwell-block (approved):

PR #3056, opened 2026-07-27
fix(desktop): allow shared agent mentions

joahg:
## Summary
- allow Desktop mention autocomplete to include another owner's agent when the existing relay policy says the current user can invoke it
- preserve archived-candidate exclusion and the existing preference for live managed identities
- keep the local-ownership filter scoped to add-member search
- add regression coverage for a cross-owner channel bot configured with `respondTo: anyone`
## Problem
`useMentions` rejected every agent identity that was not in the current user's locally managed-agent set before applying the existing mention visibility and invocability policy.
That removed cross-owner channel bots from autocomplete even when they were shared with the user and configured to respond to anyone. Because the candidate never entered the mention map, message sending received no recipient pubkey and emitted no corresponding `p` tag.
## Fix
Move the complete mention candidate decision behind a tested admission helper that applies:
1. archived identity exclusion
2. the existing invocability and relay-directory visibility policy
The ownership-only helper remains unchanged for add-member search, where that restriction is intentional.
## Testing
- `cd desktop && pnpm check`
- `cd desktop && pnpm test` — 3,631 passed
- `cd desktop && pnpm build`
The regression test covers a channel bot owned by another user, with no locally managed identity and `respondTo: anyone`, at the candidate-admission seam used by `useMentions`.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -7,6 +7,7 @@ import {
   getSharedChannelIds,
   isAgentIdentityInManagedList,
   relayAgentIsSharedWithUser,
+  shouldAdmitMentionCandidate,
   shouldHideAgentFromMentions,
 } from "./agentAutocompleteEligibility.ts";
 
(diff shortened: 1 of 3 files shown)


joahg:
🤖 Closing as part of Joah's pre-OOO PR triage — stale. Happy to reopen if this is still needed.

PR #3060, opened 2026-07-27
fix(cli): align buzz upload file MIME check with relay policy (#2963)

### LINKED ISSUE #2963: `buzz upload file` rejects PDFs/documents the relay accepts — agents can't attach files the desktop can
Thank you so much for releasing this beast in the wild! 🤗 

**Describe the bug**

The CLI applies a client-side MIME **allowlist** that is stricter than the relay it uploads to, on the endpoint that exists to accept generic files. Files are rejected locally and never reach the server.

`crates/buzz-cli/src/client.rs:1116` (`upload_file`):

```rust
if !ALLOWED_MIMES.contains(&mime.as_str()) {
    return Err(CliError::Usage(format!("unsupported file type: {mime}")));
}
```

`ALLOWED_MIMES` (`client.rs:63`) is `image/jpeg`, `image/png`, `image/gif`, `image/webp`, `video/mp4`.

But the same function uploads to `PUT {relay}/upload` (`client.rs:1144`) — the generic route, which the relay validates with `validate_file_content()` (`crates/buzz-media/src/validation.rs:159`) using a **denylist** (`BLOCKED_FILE_MIME_TYPES`: HTML/JS/SVG + executables). `application/pdf` is not blocked, and `file_mime_to_ext()` explicitly maps `application/pdf → pdf` and `…spreadsheetml.sheet → xlsx`.

So the server accepts these; the client refuses to send them.

**To Reproduce**

1. `buzz upload file --file report.xlsx`
2. → `{"error":"user_error","message":"unsupported file type: application/zip"}`
   (an `.xlsx` is a zip container, so `infer` reports `application/zip` — a type the relay
   accepts on this endpoint.)
3. Same for `.pdf`, `.docx`, `.csv`.
4. `buzz messages send --file report.xlsx` fails identically — same code path, and this is the
   path agents use to attach anything.

Confirmed the server side by patching the check out locally and re-running: the same file
uploads and returns a normal `{url, sha256, size}` descriptor. Nothing server-side had to
change.

**Expected behavior**

The CLI should accept what the relay accepts. Either mirror `BLOCKED_FILE_MIME_TYPES`, or drop the pre-check and surface the server's 4xx — the server is the authority anyway, since it verifies magic bytes rather than a client-supplied type.

**Why this matters**

This is the file-attachment path for **agents**. Desktop already handles generic attachments (#2381 describes the composer as accepting "images, videos, PDFs, archives, and other supported files"), so an agent cannot attach a file that a human can attach to the same channel on the same relay.

Concretely: an agent that generates a spreadsheet or PDF report has no way to deliver it through Buzz.

**Additional context**

Two smaller issues in the same function:

- **Size cap**: non-video files are checked against `MAX_IMAGE_BYTES` (50 MB, `client.rs:1121`). The relay has a separate `max_file_bytes` for the generic path, which the CLI has no notion of.
- **Error message**: `unsupported file type: application/pdf` states the opposite of the truth for this relay, which makes the mismatch hard to diagnose.

If the relay ever exposes Blossom **BUD-06** (`HEAD /upload`), asking the server for its
requirements would remove the duplicated policy entirely — the client would stop carrying a
second copy of a policy that only the server can enforce.

**Environment**

- CLI built from source at `9cc9652c`; code references above are against `main` @ `a64cc71f`, where `client.rs` is unchanged since that build
- Relay: self-hosted, `main` @ `ab3af828`
- macOS 15 (arm64)

maikunari:
## Problem

`buzz upload file` applies a client-side MIME **allowlist** (`ALLOWED_MIMES` — 5 image/video types) that is stricter than the relay it uploads to. The generic `PUT /upload` route the CLI targets validates content with a **denylist** (`BLOCKED_FILE_MIME_TYPES` in `buzz-media`), which accepts PDFs, zip-backed OOXML (`.xlsx`/`.docx`), CSVs, and other generic files. So the client refuses files the server would happily accept, and agents can't attach documents the desktop can.

Repro (before this change):
```
$ buzz upload file --file report.xlsx
error: unsupported file type: application/zip   # an .xlsx sniffs as zip; the relay accepts it
```

Closes #2963.

## Fix

- Replace the CLI allowlist with a **mirror of the relay's `BLOCKED_FILE_MIME_TYPES` denylist** (14 entries: active web content + native executables/installers), extracted into a small, unit-testable `check_upload_mime()`. Everything not on the denylist is sent to the relay, which remains the authority on acceptance (images/video via dedicated pipelines, generic files via the denylist). The client now fast-fails only on the permanent security denylist, so it can never again reject something the server accepts.
- Raise the non-media size cap from the 50MB image ceiling to **100MB** (`MAX_FILE_BYTES`), matching the relay's default `max_file_bytes` (`buzz-media/src/config.rs`). Without this a 60MB PDF would pass the MIME check and then be wrongly rejected by the old image cap — re-breaking the exact files this issue is about. Image (50MB) and video (500MB) caps are unchanged.
- Add a reciprocal keep-in-sync note on the relay-side const. `buzz-cli` deliberately does not take a `buzz-media` dependency (would drag S3/image/mp4 deps into the CLI binary for a 14-string list); the two lists are mirrored with cross-referencing comments.

## Tests

New unit tests in `crates/buzz-cli/src/client.rs`:
- `upload_mime_check_allows_types_the_relay_accepts` — pdf, zip (xlsx/docx), gzip, csv, octet-stream, plus image/video all pass.
- `upload_mime_check_rejects_relay_blocked_types` — html, svg, javascript, and executables still rejected.

```
cargo test -p buzz-cli     # 252 passed (incl. the 2 new)
cargo test -p buzz-media   # 103 passed
cargo fmt --all -- --check # clean
cargo clippy -p buzz-cli -- -D warnings  # clean
```

### DIFF
--- a/crates/buzz-cli/src/client.rs
+++ b/crates/buzz-cli/src/client.rs
@@ -60,20 +60,54 @@ pub fn build_imeta_tag(d: &BlobDescriptor) -> Vec<String> {
     tag
 }
 
-/// MIME types accepted for upload.
-const ALLOWED_MIMES: &[&str] = &[
-    "image/jpeg",
-    "image/png",
-    "image/gif",
-    "image/webp",
-    "video/mp4",
+/// MIME types the relay always rejects on the generic file-upload path.
+///
+/// Mirrors `BLOCKED_FILE_MIME_TYPES` in `crates/buzz-media/src/validation.rs`
+/// (kept in sync manually — buzz-cli does not depend on buzz-media to avoid
+/// pulling its S3/image/mp4 dependency tree into the CLI binary). Everything
+/// else is sent to the relay, which is the authority on acceptance: the relay
+/// validates images/video through dedicated pipelines and accepts generic
+/// files (pdf, zip, csv, ...) via a deny-list, not an allow-list.
+const BLOCKED_MIMES: &[&str] = &[
+    // Active web content — stored-XSS vectors.
+    "text/html",
+    "application/xhtml+xml",
+    "image/svg+xml",
+    "application/javascript",
+    "text/javascript",
+    // Native executables / installers.
+    "application/x-msdownload",
+    "application/x-executable",
+    "application/vnd.microsoft.portable-executable",
+    "application/x-mach-binary",
+    "application/x-sharedlib",
+    "application/x-elf",
+    "application/x-msi",
+    "application/vnd.android.package-archive",
+    "application/x-apple-diskimage",
 ];
+
+/// Client-side pre-check for `upload_file`: fail fast only on MIME types the
+/// relay permanently blocks. All other types are uploaded and validated
+/// server-side.
+fn check_upload_mime(mime: &str) -> Result<(), CliError> {
+    if BLOCKED_MIMES.contains(&mime) {
+        return Err(CliError::Usage(format!(
+            "blocked file type: {mime} (the relay rejects active web content and executables)"
+        )));
+    }
+    Ok(())
+}
 
 /// Maximum file size for image uploads (50 MB).
 const MAX_IMAGE_BYTES: u64 = 50 * 1024 * 1024;
 
 /// Maximum file size for video uploads (500 MB).
 const MAX_VIDEO_BYTES: u64 = 500 * 1024 * 1024;
+
+/// Maximum file size for generic (non-image, non-video) uploads (100 MB) —
+/// matches the relay's default `max_file_bytes`.
+const MAX_FILE_BYTES: u64 = 100 * 1024 * 1024;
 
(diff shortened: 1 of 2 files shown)


maikunari:
Status note (2026-08-07): this branch is ~300 commits behind `main` and currently **CONFLICTING**.

The bug is still real on main (`ALLOWED_MIMES` is still the 5 image/video types). There is now a crowd of open PRs for the same problem; several are already mergeable, including #4298 (relay-as-authority framing, close to this PR's intent).

I'm not rebasing this one into the stampede. If a relay-authority / denylist-style fix lands first, I'll close this as superseded. If maintainers specifically want the denylist-mirror + 100MB size-cap approach here, say so and I'll rebase.

PR #3065, opened 2026-07-27
fix(cli): preserve attachments on buzz messages edit (#3022)

### LINKED ISSUE #3022: `buzz messages edit` silently deletes the message's attachments
## Summary

Editing a message from the CLI strips every attachment off it. `build_edit` emits only `h` and `e` tags, and the desktop's render/cache overlay reads an edit with zero `imeta` tags as "the new attachment set is empty." The CLI exits 0 and gives you no way to put them back.

Same data-loss bug PR #755 fixed for the desktop. The CLI was never brought along.

## Why it happens

`crates/buzz-sdk/src/builders.rs:377-389`:

```rust
pub fn build_edit(
    channel_id: Uuid,
    target_event_id: nostr::EventId,
    new_content: &str,
) -> Result<EventBuilder, SdkError> {
    check_content(new_content, 64 * 1024)?;
    let tags = vec![
        tag(&["h", &channel_id.to_string()])?,
        tag(&["e", &target_event_id.to_hex()])?,
    ];
    Ok(EventBuilder::new(Kind::Custom(40003), new_content).tags(tags))
}
```

No media parameter, and no caller could supply one anyway. `buzz messages edit` takes `--event` and `--content` only (`crates/buzz-cli/src/lib.rs:413-423`) and reaches `build_edit` at `crates/buzz-cli/src/commands/messages.rs:713`.

The receiving side reads that absence as deletion. `desktop/src/features/messages/lib/applyEditTagOverlay.mjs:31-42`:

```js
export function applyEditTagOverlay(originalTags, editTags) {
  if (!editTags) return originalTags;
  const editEmoji = editTags.filter((t) => t[0] === "emoji");
  const droppedFromOriginal =
    editEmoji.length > 0
      ? (t) => t[0] !== "imeta" && t[0] !== "emoji"
      : (t) => t[0] !== "imeta";
```

`imeta` is unconditionally dropped from the original and replaced by whatever the edit carries, which for a CLI edit is nothing. That behaviour is deliberate and pinned by name:

```js
test("edit with zero imeta tags strips all attachments; non-imeta original tags stay", ...)
```
`desktop/src/features/messages/lib/applyEditTagOverlay.test.mjs:52`

Compare the `emoji` branch two lines up: an edit that supplies no emoji tags *preserves* the original's, and the comment at `applyEditTagOverlay.mjs:16-25` explains why, that a tag-less edit "can come from an older build ... or another client that doesn't know this path." That reasoning applies to the CLI and `imeta` word for word. It just wasn't extended there.

The desktop's own builder does carry media:

```rust
pub fn build_message_edit(
    channel_id: Uuid,
    target_event_id: EventId,
    content: &str,
    media_tags: &[Vec<String>],
    custom_emoji_tags: &[Vec<String>],
    mentions: &[&str],
) -> Result<EventBuilder, String> {
```
`desktop/src-tauri/src/events.rs:410-417`, with `imeta_tags(media_tags, ...)` at `:425`.

## Impact

An agent fixing a typo in its own message deletes the screenshot it attached. `buzz` exits 0, nothing in the output mentions attachments, and there's no `buzz messages edit --media` to restore them. The only recovery is deleting the message and posting it again.

`VISION.md:163` says buzz-cli "mirrors and extends the MCP surface ... Agents can script the entire platform without a GUI," and `README.md:40` promises agents "the same affordances as a human teammate." A human editing in the desktop keeps their attachments. An agent editing over the CLI loses them.

## Suggested fix

Fix it at the builder, not at the overlay. The overlay's replace-on-`imeta` semantics is what makes attachment removal work in the desktop, and changing it there would break `applyEditTagOverlay.test.mjs:52` for a good reason.

1. Add an options struct, or a `media_tags: &[Vec<String>]` parameter, to `build_edit` in `crates/buzz-sdk/src/builders.rs`, validating the `imeta` prefix the way `desktop/src-tauri/src/events.rs:96-108` does.
2. Add `--media` to `MessagesCmd::Edit` (`crates/buzz-cli/src/lib.rs:413`), matching whatever flag `buzz messages send` already uses for attachments.
3. The default is the real question. The safe default is for `buzz messages edit` with no `--media` to fetch the target event and carry its existing `imeta` forward, so today's invocations stop losing data, with an explicit `--no-media` to clear. The alternative, requiring `--media` explicitly and dropping otherwise, is easier to reason about but doesn't help anyone who isn't already paying attention. I'd go with carry-forward, but I can see the argument against putting a relay fetch in the CLI's synchronous edit path, so I'd rather agree on it before building.

The same argument applies more weakly to NIP-30 `emoji` tags, which `build_edit` also can't emit. Nothing is lost today because the overlay preserves those on empty, but it'd be worth adding in the same pass for symmetry.

`builders.rs:2171` and `:2182` already cover `build_edit` directly, so the tests have somewhere to go.

Happy to send the PR once you've picked a default.

maikunari:
## Problem

`buzz messages edit` silently deletes a message's attachments. `build_edit` (`crates/buzz-sdk/src/builders.rs`) emits only `h` + `e` tags — never `imeta`. The desktop render/cache overlay (`desktop/src/features/messages/lib/applyEditTagOverlay.mjs`) treats an edit carrying zero `imeta` tags as "the attachment set is now empty" — this is intentional (it's how deliberate attachment removal works), and is pinned by a test. Because the CLI never carried `imeta` forward, an agent fixing a typo in its own message deletes the screenshot it attached. `buzz` exits 0, nothing warns, and there's no flag to restore them.

Desktop was fixed for its own builder in #755; the CLI was never brought along. Closes #3022.

## Fix

Fixed at the builder + CLI, **not** the overlay (the overlay's replace-on-`imeta` semantics is correct and load-bearing for desktop removal).

- **`build_edit` gains a `media_tags: &[Vec<String>]` parameter** and emits `imeta` via the same `imeta_tags()` helper `build_message` already uses (same validation path).
- **`buzz messages edit` gains `--file` and `--no-media`:**
  - *(neither flag — default)* **carry-forward**: fetch the target event and preserve its existing `imeta` tags, so today's invocations stop losing attachments.
  - `--file <path>` *(repeatable)*: upload and **replace** attachments — identical upload path to `buzz messages send` (this is why the flag is `--file`, matching `send`, not `--media`).
  - `--no-media`: **clear** all attachments (explicit). Mutually exclusive with `--file` (enforced by clap).

The single relay fetch already performed to resolve the channel id is reused to read the carry-forward `imeta` — **no extra round-trip**. A failed/absent fetch aborts *before* signing, so an attachment-wiping edit can never be published on error.

## Tests

- `buzz-sdk`: `edit_with_media` (imeta emitted), `edit_without_media_has_no_imeta` (clear contract), `edit_rejects_malformed_media_tag`.
- `buzz-cli`: `extract_imeta_tags_returns_only_imeta` / `_empty_when_none` / `_handles_missing_tags` (only `imeta` carried, full vectors preserved), and a clap test that `--file` + `--no-media` is rejected.

```
cargo test -p buzz-sdk     # 238 passed
cargo test -p buzz-cli     # 254 passed
cargo fmt --all -- --check # clean
cargo clippy -p buzz-sdk -p buzz-cli -- -D warnings  # clean
```

## Known follow-ups (non-blocking)

- **No automated end-to-end test** for the full fetch→extract→build path — the unit tests cover the pieces; the full flow is covered by the manual runbook in `crates/buzz-cli/TESTING.md` (no mocked-relay harness exists for CLI command handlers). Verified manually.
- **Inline `image` markdown in the old body is not carried forward** — `--content` replaces the body wholesale. The `imeta` attachment (what desktop renders) is preserved; only inline markdown in a plaintext client would be affected. Left as-is to keep the change surgical.

### DIFF
--- a/crates/buzz-cli/README.md
+++ b/crates/buzz-cli/README.md
@@ -36,7 +36,9 @@ buzz messages get --channel <uuid> --limit 20
 buzz messages thread --channel <uuid> --event <event-id>
 buzz messages search --query "architecture"
 buzz messages search --author <pubkey|npub|name> --since <unix-ts>
-buzz messages edit --event <event-id> --content "Updated text"
+buzz messages edit --event <event-id> --content "Updated text"           # attachments preserved
+buzz messages edit --event <event-id> --content "Updated text" --file pic.png  # replace attachments
+buzz messages edit --event <event-id> --content "Updated text" --no-media  # drop all attachments
 buzz messages delete --event <event-id>
 
 # Diffs
--- a/crates/buzz-cli/TESTING.md
+++ b/crates/buzz-cli/TESTING.md
@@ -223,7 +223,8 @@ buzz messages thread --channel "$CHANNEL_ID" --event "$EVENT_ID" | jq .
 buzz messages search --query "Hello" | jq .
 buzz messages search --query "CLI test" --limit 5 | jq .
 
-# messages edit
+# messages edit (existing attachments are preserved by default;
+#   --file replaces them, --no-media clears them)
 buzz messages edit --event "$EVENT_ID" --content "Edited by CLI test" | jq .
 
 # messages delete
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -86,6 +86,15 @@ async fn resolve_thread_ref(
 /// Resolve the channel UUID for an event by querying for it via POST /query.
 /// Extracts the `h` tag value from the returned event's tags.
 async fn resolve_channel_id(client: &BuzzClient, event_id: &str) -> Result<Uuid, CliError> {
+    let event = fetch_event_by_id(client, event_id).await?;
+    channel_id_from_event(&event, event_id)
+}
+
+/// Fetch a single event by its ID via POST /query.
+async fn fetch_event_by_id(
+    client: &BuzzClient,
+    event_id: &str,
+) -> Result<serde_json::Value, CliError> {
     let filter = serde_json::json!({
         "ids": [event_id]
     });
@@ -95,9 +104,13 @@ async fn resolve_channel_id(client: &BuzzClient, event_id: &str) -> Result<Uuid,
     let arr = events
         .as_array()
         .ok_or_else(|| CliError::Other("query response is not an array".into()))?;
-    let event = arr
-        .first()
-        .ok_or_else(|| CliError::Other(format!("event {event_id} not found")))?;
+    arr.first()
+        .cloned()
+        .ok_or_else(|| CliError::Other(format!("event {event_id} not found")))
+}
+
+/// Extract the channel UUID from an event's h-tag.
+fn channel_id_from_event(event: &serde_json::Value, event_id: &str) -> Result<Uuid, CliError> {
     let tags = event
         .get("tags")
         .and_then(|t| t.as_array())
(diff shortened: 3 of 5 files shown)


maikunari:
Rebased onto latest `main` (`f53bbd115`).

- Conflict was only in `crates/buzz-cli/src/lib.rs` (new projects clap tests on main vs our edit-flag clap test) — kept both.
- Local verification on rebased tip: `cargo test -p buzz-sdk` (257 passed) and `cargo test -p buzz-cli` (325 passed).
- Branch is now up to date / should be mergeable pending review.

**vs #3062:** same issue (#3022). This PR stays the surgical path — fix at `build_edit` + CLI carry-forward default, with explicit `--file` (replace) and `--no-media` (clear). Happy to close as duplicate if maintainers prefer #3062's approach once one lands.

Re-requesting review from `@block/buzz-oss-team`.

PR #3070, opened 2026-07-27
feat(mobile): bring message actions to desktop parity

tellaho:
**Category:** new-feature
**User Impact:** Mobile users can copy permalinks, revisit, follow, and manage messages through a clearer long-press menu that matches desktop capabilities.

**Problem:** The mobile message menu exposed only a small subset of desktop actions, and important workflows such as copying a permalink or scheduling a reminder were unavailable or hard to discover.

**Solution:** Bring applicable desktop actions to mobile using native patterns, promote Reply, Copy link, and Remind me above the fold, and group the remaining actions by intent in a scrollable sheet.

**Intentional behavior changes (per review):**
- The quick-reaction row goes from 6 emojis to 4 (👀 and 🙏 dropped) to make room for larger 52px tap targets alongside the `+` picker, which still offers the full set.
- **Copy link — not a native share sheet — is the permalink path.** An earlier revision shipped a `share_plus` Share message row; it was removed in review since Copy link covers the job and custom-scheme `buzz://` URIs are handled inconsistently by share targets. Native share can return as a follow-up with an https fallback.
- Mark unread is message-scoped and session-local: it forces just that message unread (surfacing its channel as unread), and message-level Mark read can never clear a channel-level unread set from the channel tile.

<details>

**mobile/lib/features/channels/channels_provider.dart**
Feeds followed thread roots into unread and notification evaluation so following a thread has meaningful behavior.

**mobile/lib/features/channels/message_actions.dart**
Reworks the long-press sheet with promoted fast actions, message links, reminders, read state, thread following, and clearer action grouping while preserving existing guards. Quick-reaction circles share one extracted widget.

**mobile/lib/features/channels/read_state/message_read_state.dart**
Centralizes message-level unread evaluation across channel, message, and thread markers; channel-level forced unread deliberately does not leak into message state.

**mobile/lib/features/channels/read_state/read_state_provider.dart**
Forced-unread flags are per-context (channel id or `msg:` key) mapped to their channel, so message- and channel-level unread choices round-trip independently.

**mobile/lib/features/channels/thread_follows/thread_follows_provider.dart**
Exposes per-identity thread follow state to the message menu and notification pipeline.

**mobile/lib/features/channels/thread_follows/thread_follows_storage.dart**
Persists a bounded, validated set of followed thread roots on the device.

**mobile/lib/shared/reminders/remind_me_later_sheet.dart**
Adds reminder presets and a native custom date/time flow for deferring a message. Lives under `shared/` so the channels feature never imports another feature module. Cancelling the custom picker keeps the preset sheet open; submission failures show stable copy and log the underlying error.

**mobile/lib/shared/reminders/reminder_service.dart**
Creates desktop-compatible, self-encrypted kind-30300 reminder events.

**mobile/lib/shared/reminders/reminder_time_presets.dart**
Defines reminder choices that match the desktop experience.

**mobile/lib/shared/deeplink/deep_link.dart**
Builds canonical Buzz message links, including thread context when present.

**mobile/lib/shared/relay/nostr_models.dart**
Adds the reminder event kind to the shared Nostr model constants.

**mobile/lib/shared/widgets/sheet_divider.dart**
Shared bottom-sheet section divider used by the message actions and reminder sheets.

**mobile/test/features/channels/message_actions_test.dart**
Covers action visibility and guards, promoted actions, read/unread round-tripping (including channel- vs message-level force isolation), thread follows, and canonical links.

**mobile/test/features/channels/read_state/message_read_state_test.dart**
Covers unread precedence for channel, message, and thread contexts.

**mobile/test/features/channels/thread_follows/thread_follows_storage_test.dart**
Covers follow persistence, identity separation, validation, and storage bounds.

**mobile/test/shared/reminders/reminder_service_test.dart**
Covers reminder payloads, tags, crypto round-tripping, and preset behavior.

**mobile/test/features/channels/read_state/read_state_provider_test.dart**
Drives the production ReadStateNotifier/ReadStateManager (no fake bookkeeping) through message unread → read → unread round-trips, explicit channel-level Mark read clearing forced messages, and automatic channel-open reads preserving them.

**mobile/test/shared/reminders/remind_me_later_sheet_test.dart**
Covers custom-picker cancel keeping the sheet open, stable failure copy without the raw error, and the happy preset path.

**mobile/test/shared/deeplink/deep_link_test.dart**
Covers canonical top-level and threaded message-link generation.

</details>

## Reproduction steps

1. Run the mobile app with a signed-in identity and open a channel containing regular messages and threads.
2. Long-press a message and confirm reactions plus Reply, Copy link, and Remind me appear as fast actions above the fold.
3. Use Copy link; confirm the resulting `buzz://message` link opens the correct channel and thread context.
4. Toggle Mark unread/Mark read and Follow thread/Unfollow thread, reopening the sheet to confirm each state changes correctly. Force a channel unread from the channel tile, then mark a message read — the channel stays unread.
5. Choose a reminder preset and a custom date/time; confirm the reminder is created and appears in the desktop reminder experience. Cancel the custom date picker and confirm the reminder sheet stays open.
6. Long-press a system message and a message you cannot manage; confirm utility and destructive actions remain appropriately hidden.

## Screenshots / demos

image


klopez4212:
Looking good!

brow (commented):
🤖 I reviewed the full current diff at `d304d87`, including message-scoped read state, reminder interoperability, thread-follow persistence and notification flow, deep links, identity boundaries, UI actions, and the added tests. CI is green and the deep-link and reminder wire formats look compatible with desktop.

I am not approving yet because one repository architecture rule is violated:

- `mobile/lib/features/channels/message_actions.dart` imports `features/reminders/remind_me_later_sheet.dart` and `features/reminders/reminder_service.dart`. `AGENTS.md` states that feature modules must not import other feature modules, only `shared/`. Please move the reusable reminder entry point/service boundary under `shared/`, or inject the action into the channels feature so the feature dependency is removed.

I also found a material test gap in the read-state change. The action widget tests override `readStateProvider` with a fake that reimplements the new bookkeeping. The pure tests construct `ReadStateState` directly. Nothing drives the real `ReadStateNotifier` through message unread → read → unread or verifies that explicit channel-level Mark read clears forced message entries while automatic channel-open reads preserve them. A defect in the production map removal or manager-emission path could therefore pass the current suite. Please add focused tests against the real notifier for those transitions.

Non-blocking UX observations:

- Cancelling the custom date/time picker currently dismisses the entire reminder sheet because the sheet navigator is popped before checking for a null timestamp. Keeping the preset sheet open on cancel or invalid time would make retrying easier.
- Reminder submission failures expose the raw exception in a user-facing snackbar. Prefer a stable user message and log the underlying relay/crypto detail.

The new files remain below the 1,000-line ceiling, and I did not find a release-blocking defect in the deep-link encoding, self-encrypted reminder payload, or thread-follow storage itself. Re-review after the feature-boundary violation and production-notifier coverage are addressed.

mobile/lib/features/channels/channels_provider.dart
```diff
@@ -545,6 +546,7 @@ class ChannelsNotifier extends AsyncNotifier<List<Channel>> {
           event,
           myPk,
           participatedRootIds: _participatedRootIds,
+          followedRootIds: _followedRootIds(),
```
brow, line 549: 🤖 Could we add one notifier-level test showing that a reply in a followed thread actually badges its channel? Storage and menu toggling are covered, but the headline wiring from `threadFollowsProvider` through `shouldNotifyForEvent` is not, so a provider or root-ID mismatch could leave follows as a silent no-op.

brow (approved):

Issue #2949, opened 2026-07-26
Windows: installer does not register buzz:// and cold-start deep links are dropped


bdleasure:
## Summary

The official `v0.4.24` Windows NSIS installer does not register the configured `buzz://` protocol, and launching the installed executable with a `buzz://nostr-bind?...` argument on a cold start does not deliver the request to the app. This blocks the shipped Nostr identity proof flow on Windows.

The same source shape remains on `v0.4.25`, `v0.4.26`, and current `main`.

## Reproduction

Environment: Windows x64, official `Buzz_0.4.24_x64-setup_alpha-unsigned.exe` from the GitHub release.

Verified installer SHA-256:

```
1931fff983f612364860f723997cd3692083a7128fe3b82c18f55d0144442c78
```

1. Install through Windows Explorer into the normal per-user location.
2. Launch Buzz, create/import an identity, close it completely, and reopen it. Identity persistence succeeds.
3. Inspect protocol registration:
   - `HKCR\buzz` is absent.
   - `HKCU\Software\Classes\buzz` is absent.
4. Open a valid `buzz://nostr-bind?...&return=clipboard` URL from Chrome. No signing dialog appears.
5. Pass the same valid URL as the sole argument to the installed `buzz-desktop.exe`. On cold start the app remains on “Join or create a community”; no signing dialog appears.

The request itself is accepted by `parse_nostr_bind_deep_link` in source/unit fixtures and has valid challenge, nonce, six-digit code, protocol/action/audience/version, HTTPS origin, RFC3339 expiry, and clipboard return mode.

## Source diagnosis

`tauri.conf.json` correctly declares:

```json
"deep-link": { "desktop": { "schemes": ["buzz"] } }
```

However, `tauri-plugin-deep-link` 2.4.9 does not register configured Windows schemes merely from `init()`. Its documented `DeepLinkExt::register_all()` writes `HKCU\Software\Classes\<scheme>`, but Buzz never calls it.

There is also a cold-start delivery race:

- plugin initialization calls `handle_cli_arguments(std::env::args())`, stores the URL in `current`, and emits `deep-link://new-url`;
- Buzz attaches `on_open_url` later inside the application `.setup(...)` callback;
- Buzz does not consume `app.deep_link().get_current()` after attaching the listener;
- the single-instance callback covers a later duplicate process, but not the first process's already-emitted cold-start event.

[...]

PR #3073, opened 2026-07-27
fix(desktop): register configured deep-link schemes

Bortlesboat:
## Summary

- register the statically configured deep-link schemes during Windows/Linux startup
- keep startup resilient if OS-level scheme registration fails
- add a focused regression that pins registration before the runtime listener

## Context

Initializing `tauri-plugin-deep-link` installs the plugin, but Windows and Linux still need the configured schemes registered at runtime. Without that step, an installed desktop build may have no `buzz://` handler.

This addresses the protocol-registration part of #2949. #2409 separately handles delivering a URL from a cold-start process invocation, so this PR deliberately does not duplicate that work.

## Current-main refresh

Merged current `main` and retained its new `egress_guard` module alongside the focused deep-link regression. The PR delta remains exactly two files and 36 additions.

## Validation

- `cargo fmt --all -- --check` passes
- the standalone regression executes with `rustc --test`: 1 passed
- the full Tauri test target compiled its dependency graph but the local build script requires the generated `binaries/buzz-acp-x86_64-pc-windows-msvc.exe` sidecar, which is absent in this checkout
- refresh commit is DCO-signed and GitHub-verified
- `git diff upstream/main...HEAD --check` passes


Bortlesboat:
Rebased onto main. The setup code moved into `deep_link::install_deep_link_handlers` since I opened this, so the fix now sits there rather than in `lib.rs` — `lib.rs` is untouched in the current diff.

The reason it's still worth taking: `install_deep_link_handlers` doesn't call `register_all()`, so the original problem is unchanged — on Windows and Linux the configured schemes never get registered with the OS, and macOS only avoids it because the bundled manifest covers it. Registration goes in ahead of `on_open_url` so a URL that shows up during launch has somewhere to land.

I also had to adjust the startup test. It asserted `handle_deep_link_url` appeared exactly once, but the function has two call sites now — the runtime listener and the Windows/Linux `get_current` replay — so it now checks both of them route through the validated handler instead.

Ran `cargo test --lib deep_link` on Windows: 51 passed. rustfmt clean.

PR #3077, opened 2026-07-27
fix(acp): answer top-level DMs inline instead of opening a thread

### LINKED ISSUE #3075: buzz-acp: a top-level DM is answered in a thread, not inline — the DM branch supplies no reply destination
## Describe the bug

A plain, top-level message to an agent in a 1:1 DM gets answered **inside a thread** rather than
inline in the DM, so the human has to open a collapsed "N replies" affordance to read a reply to
their own direct message. Nothing in the conversation asked for a thread.

This is the second half of #2748, which fixes the *anchor* for replies already inside a thread and
explicitly parks this part:

> Worth discussing whether top-level DM exchanges should thread at all — a 1:1 DM arguably reads
> best flat, with threading reserved for explicit user-initiated threads.

Filing it separately so the behavior question can be settled on its own. It is orthogonal to
#2748: the two touch different branches and neither PR (#2766, #2778) changes this path.

## Steps to reproduce

1. DM a managed agent (no `@mention`, no thread) — e.g. "find the X repo".
2. The agent answers.
3. The answer is not in the DM. The DM shows the original message with a `3 replies` footer, and
   the answer only appears after opening the thread panel.

## Expected behavior

An ordinary answer to a top-level DM arrives **inline in the DM**. A DM is already a private 1:1
conversation — there is nothing to keep a thread tidy for. Threads inside a DM stay reserved for
threads the human opened (or explicitly asks for), and those already anchor correctly once #2748
lands.

## Root cause

`crates/buzz-acp/src/queue.rs` — in `format_context_hints`, the DM branch only emits a reply
instruction inside the `if let Some(ref root) = thread_tags.root_event_id` arm:

```rust
// If this is a DM reply, include thread structural info as supplementary.
if let Some(ref root) = thread_tags.root_event_id {
    ...
    if let Some(event_id) = reply_anchor {
        append_reply_instruction(&mut s, event_id);
    }
}
s
```

A top-level DM has no thread tags, so `reply_anchor` is `None` and the `[Context]` block says
nothing at all about where to reply. With no destination supplied, the agent falls back to the
general rule in `base_prompt.md`:

> The app/harness will choose the correct reply destination: the root of the triggering thread when
> the turn is already threaded, **or the triggering top-level event when the human started a new
> thread.**

…and opens a thread rooted at the DM. In a channel that rule is right — `format_context_hints`
emits `append_new_thread_reply_instruction` for exactly that case. In a DM it produces a thread
nobody asked for.

## Proposed fix

Give the top-level DM case its own explicit instruction — answer inline, no `--reply-to` — as the
`else` of the existing `if let Some(root)` arm, mirroring how the channel branch handles its
top-level case. Threaded DM turns keep their existing instruction, so this does not touch the lines
#2766/#2778 modify.

I have this implemented and tested locally and will open a PR against this issue.

## Version and platform

Desktop app on macOS (Darwin 25.4.0), self-hosted relay, desktop-managed agent, default settings.
Reproduced on `main` at 63c62fcf3.

muqsitnawaz:
## Summary

A top-level message in a 1:1 DM gets answered **inside a thread** instead of inline, so a plain
question to your own agent lands behind a collapsed `N replies` affordance you have to open.

Root cause is a silent fall-through in `format_context_hints`: the DM branch only emits a reply
instruction inside the `if let Some(ref root) = thread_tags.root_event_id` arm. A top-level DM has
no thread tags, so `reply_anchor` is `None` and the `[Context]` block says **nothing at all** about
where to reply. With no destination supplied, the agent applies the general rule from
`base_prompt.md` — *"the triggering top-level event when the human started a new thread"* — and
opens a thread rooted at the DM. That rule is right in a channel (the channel branch emits
`append_new_thread_reply_instruction` for exactly this case); in a DM it produces a thread nobody
asked for.

The fix gives the top-level DM case its own explicit instruction — answer inline, no `--reply-to`
— as the `else` of the existing arm, mirroring the channel branch's shape. A DM is already a
private 1:1 conversation; there is nothing to keep tidy with a thread.

**Threaded DM turns are untouched.** A thread the human opened still gets its reply instruction, so
this is orthogonal to #2748 and does not touch the lines #2766/#2778 modify — it stacks cleanly on
either.

### Related issue

Fixes #3075. Split out of #2748, which parks this exact question:

> Worth discussing whether top-level DM exchanges should thread at all — a 1:1 DM arguably reads
> best flat, with threading reserved for explicit user-initiated threads.

Searched open issues/PRs for duplicates: #2748 and its two PRs (#2766, #2778) cover the *anchor*
for replies already inside a thread — a different branch and a different defect. No PR covers this
one.

### Testing

- `cargo test -p buzz-acp` — 599 passed, 0 failed.
- `cargo fmt --all -- --check` clean; `cargo clippy -p buzz-acp --all-targets -- -D warnings` clean.
- `just ci` running locally; I'll confirm in a comment when it settles.

Test changes:

- **New** `test_dm_non_reply_instructs_flat_reply` — a top-level DM prompt carries the flat-reply
  instruction and anchors no event.
- **New** `test_dm_thread_reply_keeps_reply_instruction` — a DM turn *inside* a thread still gets
  its reply instruction and does **not** get the flat-reply one (guards the split).
- **Renamed + retargeted** `test_reply_instruction_absent_for_dm_non_reply` →
  `test_dm_non_reply_instructs_flat_reply`. Flagging this deliberately: the old test asserted
  `!prompt.contains("--reply-to")` for a top-level DM, which pinned the silence this PR removes.
  The new assertions keep the meaningful half — no reply is anchored to any event — while
  requiring the instruction that was missing.
- `test_reply_instruction_present_for_dm_thread_reply` is intentionally left alone; it belongs to
  the #2748 PRs.

Docs: `base_prompt.md` §Threading now states the DM rule alongside the channel rules it already
documents, so the agent-facing contract and the harness agree.

### DIFF
--- a/crates/buzz-acp/src/base_prompt.md
+++ b/crates/buzz-acp/src/base_prompt.md
@@ -53,6 +53,8 @@ Use the reply destination supplied in the `[Context]` block for ordinary replies
 
 For human-facing work, keep the conversation flat and easy to read. The app/harness will choose the correct reply destination: the root of the triggering thread when the turn is already threaded, or the triggering top-level event when the human started a new thread.
 
+In a direct message, a top-level turn has no thread at all — answer inline in the DM, with no reply anchor. A DM is already a private 1:1 conversation, so threading an ordinary answer only hides it behind a collapsed reply count. Thread inside a DM only when the human opened the thread (the `[Context]` block then supplies its root) or explicitly asks for one.
+
 For agent-to-agent coordination with no human in the loop, deeper nesting is allowed when it helps preserve task structure. Do not flatten agent-only subthreads just because they are inside a thread.
 
 When in doubt, prefer the reply destination explicitly supplied in `[Context]`. If you intentionally choose a different destination, explain why briefly in the message.
--- a/crates/buzz-acp/src/queue.rs
+++ b/crates/buzz-acp/src/queue.rs
@@ -1171,6 +1171,23 @@ fn append_new_thread_reply_instruction(s: &mut String, event_id: &str) {
     ));
 }
 
+/// Append a flat-reply instruction for a top-level DM turn.
+///
+/// A DM is already a private 1:1 surface, so an ordinary reply belongs inline in
+/// the conversation. Threading it hides the answer behind a collapsed "N replies"
+/// affordance the human has to open, and every following exchange starts one
+/// level deeper. Threads inside a DM stay reserved for threads the human opened
+/// — those keep their reply instruction via [`append_reply_instruction`].
+fn append_dm_flat_reply_instruction(s: &mut String) {
+    s.push_str(
+        "\nIMPORTANT: This is a top-level direct message, not a thread. Send \
+         ordinary replies for this turn inline in the DM — do NOT pass \
+         `--reply-to` — so the human reads the answer in the conversation \
+         instead of a collapsed thread. Only open a thread if the human \
+         explicitly asks for one.",
+    );
+}
+
 /// Decide whether a turn is human-facing for reply-anchor purposes.
 ///
 /// A turn is human-facing when the triggering sender is a human, OR a human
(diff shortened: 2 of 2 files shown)


muqsitnawaz:
Local gate settled green: `just ci` → exit 0 (fmt, workspace clippy, desktop check/test/build, Tauri clippy + tests, web build, mobile analyze + tests). `cargo test -p buzz-acp` is 599 passed / 0 failed, including the two new DM tests.

The three checks that run for a fork PR here (DCO, Semgrep OSS, zizmor) are green; the rest of the `CI` workflow needs a maintainer to release the run.

PR #3084, opened 2026-07-27
Improve Kimi Code runtime discovery

ninjusai:
## Summary

- probe Kimi Code's standalone `~/.kimi-code/bin` install directory so GUI-launched Desktop processes can resolve `kimi`
- keep the Kimi preset on `kimi acp` and point its setup guidance to the official Kimi Code CLI documentation
- document the harness configuration and add regression coverage for discovery and preset arguments

## Tests

- `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml -- --check`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml kimi -- --test-threads=1` — 2 passed
- full serial Tauri suite — 1,693 passed, 10 ignored; the existing order-dependent `managed_agents::runtime::tests::claude_spawn_uses_the_probed_cli_executable` test failed in-suite and passed immediately in isolation on the same commit

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -9,7 +9,7 @@ Buzz Relay ──WS──→ buzz-acp ──stdio──→ Your Agent
                                        (send_message, etc.)
 ```
 
-Supports any agent that speaks [ACP](https://agentclientprotocol.com/) over stdio: **goose**, **codex** (via [codex-acp](https://github.com/agentclientprotocol/codex-acp)), and **claude code** (via [claude-agent-acp](https://github.com/agentclientprotocol/claude-agent-acp)).
+Supports any agent that speaks [ACP](https://agentclientprotocol.com/) over stdio, including **goose**, **codex** (via [codex-acp](https://github.com/agentclientprotocol/codex-acp)), **claude code** (via [claude-agent-acp](https://github.com/agentclientprotocol/claude-agent-acp)), and **Kimi Code**.
 
 ## Prerequisites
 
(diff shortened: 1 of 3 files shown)


linxule:
One data point on the discovery path, in case it's useful for scoping this.

I hit this exact symptom (Kimi Code showing "CLI needed" on a GUI-launched Buzz while `kimi` worked fine in a terminal), but `~/.kimi-code/bin` wouldn't have resolved it. Kimi Code also ships as an npm package (`@moonshot-ai/kimi-code`), so a `bun add -g` / npm-global install puts the executable in the package manager's bin dir instead:

```console
$ ls ~/.kimi-code/bin
rg

$ command -v kimi
/Users/…/.bun/bin/kimi -> ../install/global/node_modules/@moonshot-ai/kimi-code/dist/main.mjs
```

`~/.kimi-code` exists and holds config/credentials/cache, but `bin/` only has a vendored `rg` — no `kimi`. So the standalone-installer probe here and the package-manager install route are two separate cases.

The underlying cause is the same one this PR's comment describes, and it isn't Kimi-specific: `~/.bun/bin` isn't in `common_binary_paths()`, and bun's installer writes its PATH export to `~/.zshrc` (interactive), so `find_via_login_shell()` can't recover it either — same as the nvm situation already documented in `resolve_command_uncached`.

I opened #3343 to add `~/.bun/bin` to `common_binary_paths()`, which should be complementary to this rather than overlapping: this PR covers the standalone installer, that one covers bun-installed CLIs generally (it fixed Codex on my machine too, for the same reason).

Issue #3031, opened 2026-07-26
Managed Codex agents receive mentions but buzz messages send fails with misleading DNS error on Desktop 0.4.26


Bharathchinneni:
### Summary

On macOS Buzz Desktop v0.4.26, locally managed Codex agents receive channel mentions, generate a reply, but cannot publish it. The agent reports a transient relay DNS error from `buzz messages send`, even though the agent harness is connected and online.

### Reproduction

1. In a hosted Buzz community, add a managed Codex agent (reproduced with Fizz and Honey).
2. Mention the agent in a channel thread, for example `@Fizz hi`.
3. The agent runs, reads the Buzz CLI skill, and invokes `buzz messages send --channel <channel-id> --reply-to <event-id> --content <reply>`.
4. The command fails twice and the reply does not appear in the UI.

### Expected

The agent reply is published to the thread.

### Actual

The agent receives the mention and forms a reply, but `buzz messages send` fails with a message described by the agent as a relay DNS lookup failure. The UI shows no response because no message is delivered.

### Evidence

- Buzz ACP logs show Fizz connected to `wss://bgpv.communities.buzz.xyz`, relay observer enabled, and presence online before the mention arrived.
- The same behavior reproduced with Honey.
- macOS system DNS was changed to Cloudflare `1.1.1.1` / `1.0.0.1`; both resolve `bgpv.communities.buzz.xyz` successfully, so this is not a persistent local DNS failure.
- Existing fixes #1287 and #1363 look closely related: Codex-managed nested CLI/MCP processes need the relay hostname in the network proxy allowlist and OS sandbox network access. This release still exhibits an equivalent outbound-send failure.

### Environment

- Buzz Desktop: 0.4.26
- macOS
- Relay: `wss://bgpv.communities.buzz.xyz`
- Managed runtime: Codex ACP
- Hosted community (not a localhost relay)

### Request

Could Buzz surface the underlying `buzz messages send` error/host in the agent transcript and confirm that the managed Codex subprocess inherits the current relay hostname and sandbox network access?

ayushsdev:
Noticing this issue as well, any ETA on the fix?

mkelowna:
Root cause found, and it is not DNS or the proxy allowlist. Reproduced on a newer build than the one reported here.

## The error is a sandbox denial wearing a DNS costume

[...]

PR #3097, opened 2026-07-27
fix(desktop): require codex-acp 1.1.7 for managed agents

Bharathchinneni:
## Summary

- require `@agentclientprotocol/codex-acp` 1.1.7 or newer for Buzz managed agents
- parse the full adapter semver instead of accepting every 1.x release
- route older 1.x installations through the existing managed reinstall flow

## Why

A managed Codex agent could receive relay mentions but fail to publish `buzz messages send` replies while running codex-acp 1.1.5. Updating the adapter to 1.1.7 (which updates its bundled Codex runtime) restored outbound replies in a live macOS Buzz test. The current major-only gate treats 1.1.5 as current, so Buzz never offers that upgrade.

## Validation

- `git diff --check`
- live Buzz test: after upgrading from codex-acp 1.1.5 to 1.1.7, a managed agent successfully replied `OK` to a relay mention
- automated Rust tests were not run here because this macOS environment does not have Cargo installed

Related to #3031

### DIFF
--- a/desktop/src-tauri/src/commands/agent_discovery.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery.rs
@@ -26,7 +26,8 @@ fn active_installs() -> &'static std::sync::Mutex<std::collections::HashSet<Stri
 /// `None` if none was found).
 ///
 /// Returns `None` when no install is needed (adapter is present and current).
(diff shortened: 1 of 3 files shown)


salman1993:
closing since its addressed by https://github.com/block/buzz/pull/7427

Issue #2987, opened 2026-07-26
External (non-desktop-managed) agents: publishing kind:10100 makes them un-mentionable — `shouldHideAgentFromMentions`'s invocability logic is unreachable


ilporyynanen-dev:
**Component:** desktop
  **Version:** v0.4.26, source at `0096d710ed2e6abab19aaf7cdc14e3ee603d7ec8`
  
  ## Summary
  
  We run two headless `buzz-acp` agents on a server (systemd, one per team member, each on its own subscription). They authenticate via NIP-42, are relay members, join open channels via kind:9021, and
  respond correctly — the server side works great.
  
  The problem is desktop UI eligibility: for an agent the desktop does **not** manage, publishing a kind:10100 agent profile makes the agent *disappear from @-mention autocomplete* (and from the
  add-member search), even when the 10100 content explicitly declares the current user invocable via `respond_to: "allowlist"` + `respond_to_allowlist`. This forces an either/or choice for external
  agents:
  
  - **No kind:10100** → agent is mentionable (appears as a plain member), but is absent from the Agents directory.
  - **kind:10100 published** → agent appears in the Agents directory, but can no longer be @-mentioned from the composer at all — and since `buzz-acp` subscribes to mentions, the agent becomes
  effectively unusable in channels.
  
  ## Root cause
  
  `useMentions.ts:249` runs the managed-list gate **before** the invocability logic:
  
  ```ts
  if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
    return;                       // ← any isAgent=true candidate not managed by THIS desktop is dropped here
  }
  if (shouldHideAgentFromMentions({ ... })) { return; }
  ```
  
  `isAgentIdentityInManagedList` (`agentAutocompleteEligibility.ts:57`) returns false for every `isAgent === true` candidate whose pubkey is not in the local managed list — so the candidate never reaches
  `shouldHideAgentFromMentions`.
  
  But `shouldHideAgentFromMentions` (same file, line 67) contains carefully-written logic for exactly this case ("Option B"): show a member agent unless the relay directory gives an explicit
  not-invocable signal, backed by `relayAgentIsSharedWithUser`, which checks `respond_to === "allowlist" && respond_to_allowlist.includes(currentPubkey)`. Because of the earlier gate, **this branch is 
  unreachable for every non-managed agent** — the only candidates that survive the gate are managed agents, which are already unconditionally mentionable via `mentionableAgentPubkeys`. The invocability
[...]

PR #3108, opened 2026-07-27
fix(desktop): let channel-member agents into the @mention picker

PizzzaDog:
## Problem

The desktop @mention autocomplete filters agent identities to the current user's **own** managed-agent list. Agents owned by someone else never appear in the picker — even when they're already members of the channel — so you can't `@mention` a teammate's agent that's sitting right there in the channel.

## Fix

One-line filter relaxation in `desktop/src/features/messages/lib/useMentions.ts` (`addCandidate`). An agent now passes the pre-filter if it's in your managed list **or** it's a channel member:

```diff
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
+      if (
+        !candidate.isMember &&
+        !isAgentIdentityInManagedList(candidate, managedAgentPubkeys)
+      ) {
         return;
       }
```

The downstream `shouldHideAgentFromMentions` gate (already present, "Option B") then makes the real visibility decision for member agents: it hides only agents with an explicit *not-invocable* relay-directory (kind:10100) entry, and shows agents whose invocability is unknown. This change simply stops member agents from being dropped **before** they reach that gate.

## Scope notes

- **Agents only.** `isAgentIdentityInManagedList` already returns `true` for non-agents, so human members are unaffected.
- No change to access/permissions — only *who the autocomplete offers*.
- A structurally identical filter exists in `features/channels/ui/MembersSidebar.tsx:285` (member-list display). Left untouched — out of scope for the mention picker; can follow up if desired.

## Verification

- `pnpm test` — **3631 pass / 0 fail**. The member-shown path is already covered by `agentAutocompleteEligibility.test.mjs` ("shows member agents with unknown invocability"); this change removes the pre-filter that blocked reaching it.
- `pnpm typecheck` — clean.
- `pnpm exec biome check` on the changed file — clean.

Originating discussion: Buzz channel #WC.

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -246,7 +246,10 @@ export function useMentions(
       if (isArchivedDiscovery(pubkey)) {
         return;
       }
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
+      if (
+        !candidate.isMember &&
+        !isAgentIdentityInManagedList(candidate, managedAgentPubkeys)
+      ) {
         return;
       }
       if (


Bartok9 (commented):
## Collaborative note (Bartok9)

Small nit-shaped fix — thanks for unpacking the gate.

**Concern for merge priority:** allowing *any* `isMember` agent into the picker before eligibility fully runs can surface **owner-only / non-invocable** member agents that Option B is designed to hide once directory metadata exists. Stronger siblings (#2605 / #3448) expand the **allow-set from directory invocability**, not mere membership.

Suggestion: if you still want this PR alive, add an explicit test that a **member + directory owner-only (other owner)** agent stays hidden after your gate, or close-in-favor of the carrier PR once it lands and leave a credit note.

Not a formal reject — just swarm coordination so we don’t regress #2987’s intended Option B.

PR #3110, opened 2026-07-27
fix: let mentionable relay agents through the managed-list gate

### LINKED ISSUE #2987: External (non-desktop-managed) agents: publishing kind:10100 makes them un-mentionable — `shouldHideAgentFromMentions`'s invocability logic is unreachable

varunsahni18:
## Summary

Fixes a bug where external (non-desktop-managed) agents publishing kind:10100 profiles became un-mentionable and un-addable to channels, even when their `respond_to` allowlist explicitly included the current user.

## Problem

`isAgentIdentityInManagedList` in both `useMentions.ts` and `MembersSidebar.tsx` dropped every non-managed agent before `shouldHideAgentFromMentions` could evaluate invocability via `respond_to` allowlist. This made the carefully-written Option B logic in `shouldHideAgentFromMentions` unreachable for external agents.

**Reproduction:**
1. Create a headless `buzz-acp` agent with `--respond-to=allowlist` including user A
2. No kind:10100 → agent is mentionable ✓
3. Publish kind:10100 → agent disappears from @-mention autocomplete ✗
4. Delete kind:10100 → agent becomes mentionable again ✓

## Root Cause

The managed-list gate ran before the invocability check:

```typescript
// Before: non-managed agents dropped here, never reach shouldHideAgentFromMentions
  return;
}
if (shouldHideAgentFromMentions({ ... })) { return; }
```

### `desktop/src/features/messages/lib/useMentions.ts`
- Let `mentionableAgentPubkeys` through the managed-list gate so `shouldHideAgentFromMentions` can evaluate invocability for external agents

### `desktop/src/features/channels/ui/MembersSidebar.tsx`
- Import `getMentionableAgentPubkeys`, `getSharedChannelIds`, and `useChannelsQuery`
- Compute `mentionableAgentPubkeys` from relay agents and shared channels
- Let mentionable agents through the add-member gate alongside managed agents

## Testing
- Both files parse correctly
- Existing test suite: 329 tests (pre-existing failures due to missing build environment, not related to this change)
- Logic verified: `mentionableAgentPubkeys` already computed correctly via `getMentionableAgentPubkeys` which checks `respond_to` allowlist

Closes #2987


Draugmar:
Independent confirmation, plus one gap this fix doesn't cover.

**Repro confirmed** on a self-hosted relay (`ghcr.io/block/buzz:sha-070fb6a`, web client served by the relay) with six headless `buzz-acp` agents (systemd, `--respond-to=anyone`): rich kind:10100 (`respond_to: "anyone"` + `channel_ids`) → agents visible in the Agent directory and channel member lists, absent from @-mention autocomplete. The gate change in this PR resolves our case: with `mentionableAgentPubkeys` derived from the 10100 content, both the member-sourced and relay-agent-sourced candidates now pass.

**Gap: minimal kind:10100s still hide the agent after this fix.** `relayAgentIsSharedWithUser` needs `respondTo === "anyone"` plus a shared channel via `channelIds`, or an allowlist hit. A 10100 with neither parses to `respondTo: null`, `channelIds: []` → not mentionable → directory presence then reads as an explicit not-invocable signal (`shouldHideAgentFromMentions`) and hides a channel-member agent.

The sharp edge: the official CLI publishes exactly such an event as a side effect — `buzz channels set-add-policy` emits kind:10100 with only `{"channel_add_policy": "..."}` (`buzz-cli/src/commands/channels.rs`, `cmd_set_add_policy`), and 10100 is plain-replaceable so it also clobbers any richer profile. Any external agent that ever set an add-policy stays silently un-mentionable after this fix.

Possible directions: treat absent `respond_to`/`channel_ids` as "no invocability signal" (show member agents) rather than exclusion; and/or make `set-add-policy` merge into existing 10100 content instead of replacing it (that clobbering is how we first hit this).

Happy to test a follow-up build against our setup.

surfingdev:
We hit this from the operator side (three headless `buzz-acp` agents under systemd against a managed relay — details in #2987). Read through the patch; the approach looks right. Two things that might matter before this lands, and one open question.

## 1. The fix is inert for agents registered the documented way

Both gates now depend on `mentionableAgentPubkeys`, which is populated by `relayAgentIsSharedWithUser`, which returns false unless `agent.respondTo` is `"allowlist"` or `"anyone"`:

```ts
if (agent.respondTo === "allowlist" && normalizedCurrentPubkey) { … }
return agent.respondTo === "anyone" && agent.channelIds.some(…)
```

`respondTo` comes from the kind:10100 content (`fromRawRelayAgent`: `respondTo: agent.respond_to ?? null`).

The only CLI path to publish a kind:10100 is `buzz channels set-add-policy`, and it writes the **entire** content as exactly one field (`crates/buzz-cli/src/commands/channels.rs`):

```rust
let content = serde_json::json!({ "channel_add_policy": policy }).to_string();
```

So `respond_to` is absent → `respondTo` is `null` → both branches false → the agent is **still dropped**, even with this patch. Anyone who registered their agent using the documented command sees no change; only operators who hand-roll a fuller kind:10100 benefit. We only got a complete entry by writing our own NIP-42 publisher.

This seems worth pairing with either a `buzz-acp --publish-profile` (suggested in #2987) or an extension of `set-add-policy` so the supported path emits `respond_to` / `respond_to_allowlist` / `channel_ids`. Otherwise the fix is correct but unreachable for the common setup.

## 2. `anyone` and `allowlist` are not equally robust

The `allowlist` branch returns early and ignores channels entirely. The `anyone` branch additionally requires `channelIds.some(id => sharedChannelIds.has(id))`.

`channel_ids` is self-declared in the agent's own kind:10100 and nothing maintains it — `buzz-acp` doesn't republish when its channel set changes. So a `respond_to: "anyone"` agent that later joins a new channel is silently un-mentionable there until someone republishes its profile by hand, while an `allowlist` agent keeps working.

That asymmetry isn't introduced by this PR, but the PR is what makes it observable, and it will read as flakiness ("it worked in #eng, not in #landing"). Might be worth a comment at the `relayAgentIsSharedWithUser` call site at minimum.

## 3. Open question: does the mention itself route, or spawn?

This patch makes external agents selectable. We could not verify what happens on the mention because we never got them visible without the patch — but from a different angle we did hit:

```
Could not start agent mention: <name>: agent <pubkey> has no private key available …
```

from `spawn_key_refusal` (`managed_agents/storage.rs:164`), reached by registering an external agent in the local `managed-agents.json`. That path attempts a **local spawn** rather than routing to an already-connected agent, and there's no record field marking an agent as externally hosted.

If the mention path can reach a spawn attempt for a candidate that is now visible but not locally managed, this PR would make agents selectable-but-unmentionable rather than fixing them. Worth a check — happy to test a build against our setup if that helps.

## Note on the reproduction

Step 4 of the repro ("Delete kind:10100 → agent becomes mentionable again") did **not** hold in our environment: after a kind:5 e-tag deletion of all three profiles (relay verified returning empty) and an app restart, the agents stayed absent from `@` autocomplete — including one with no `agent_owner_pubkey` at all, which by our reading should evaluate `isAgent === false` on every disjunct. Detail in [our comment on #2987](https://github.com/block/buzz/issues/2987#issuecomment-5092315380). Might just be our environment, but if there's a second path that hides these candidates, this patch alone wouldn't clear it.

Draugmar:
Follow-up on both open points, tested end-to-end on my self-hosted setup (relay `sha-070fb6a`, six headless `buzz-acp` agents, web client served by the relay).

**1. The deletion-doesn't-restore observation reproduced here — until I removed the *second* signal.** Sequence, relay-verified between steps:

- Rich kind:10100 + NIP-OA-attested kind:0 → agent in directory, hidden from `@` autocomplete.
- Delete the 10100 (e-tag kind:5; relay serves zero active) + app restart → **still hidden** (matches your observation).
- Additionally republish the kind:0 **without** the NIP-OA auth tag + app restart → **agent appears** in `@` autocomplete.

Mechanism: `nostr_convert.rs:345` sets `is_agent: owner_pubkey.is_some()` — a verified NIP-OA attestation on the profile flags the member candidate `isAgent` with no 10100 involved, so it enters the same managed-list gate, and with no 10100 there's nothing for `mentionableAgentPubkeys` to whitelist. (`:268` adds a third signal, `role == "bot"` — possibly your attestation-free case.) Any of the three signals routes a candidate into the gate; the whitelist this PR adds is only derivable from a **rich** 10100. Treating "isAgent but no invocability data" as *no exclusion signal* for channel members would cover all entrances.

**2. Re: the spawn-vs-route question — mentions route; spawn is never attempted for external agents.** `spawn_key_refusal` takes a `ManagedAgentRecord` and is reached only from the local managed-agents store spawn paths (`runtime.rs:458`, `project_git_workflow.rs:188`). Relay-side, `buzz-acp` wake is purely p-tag based; I exercise this daily via CLI-composed mentions, workflow `send_message` (server-side name→p-tag resolution) and thread replies, all waking systemd-hosted daemons. Your refusal came from registering external agents in `managed-agents.json` — without that registration the path doesn't exist.

**3. The `set-add-policy` clobbering now has a PR: #3132** — merge into the existing 10100 content instead of replacing, query errors propagate. Built and A/B-tested against my live relay (unpatched CLI clobbers → patched CLI preserves `respond_to`/`channel_ids` and adds the policy). One note for whoever writes integration tests in this area: same-second replaceable updates tie-break nondeterministically, so test events need distinct `created_at`s.

Shipitrealgood:
Thanks for digging into this — the diagnosis matches what we found working the same #2987 family: the managed-list pre-gate makes the real eligibility logic unreachable for external agents.

One load-bearing concern with the `MembersSidebar` half, though: it gates **adds** on **mentionability** (`respond_to` via `getMentionableAgentPubkeys`), but the relay doesn't enforce `respond_to` on adds. Third-party adds are enforced against the agent's kind:10100 `channel_add_policy` (`buzz-relay` `handlers/side_effects.rs`: `owner_only` checked against the stored attested owner, `nobody` refused for everyone, `anyone` allowed). The two declarations are independent, so gating adds on `respond_to` fails in both directions:

- **offers adds the relay rejects** — `respond_to: anyone` in a shared channel, but `channel_add_policy: nobody` → shown in the picker, refused on submit;
- **hides adds the relay accepts** — `channel_add_policy: anyone`, but `respond_to: owner-only` or an allowlist without the viewer → invisible despite being addable.

There's also a structural gap for the add picker's primary use case: `relayAgentIsSharedWithUser` requires `respond_to: anyone` agents to already share a channel with the viewer, but adding is exactly how an agent gets into its *first* shared channel.

Two open PRs split this per-surface, each reading its own relay-enforced declaration: #3676 (mentions via `respond_to`, including the verified-owner escape for `owner-only` agents) and #3864 (adds via `channel_add_policy`, carried through the wire types so the picker can mirror what `side_effects.rs` actually enforces). One design constraint worth keeping whichever way this converges: #2149 introduced the managed-list gate to suppress stale, declaration-less agent identities, so any replacement gate should keep "no current declaration → not offered" — #3864 does this deliberately (agents with no directory entry stay hidden even where the relay's stored default would accept the add). Might be worth converging there so the two surfaces don't get coupled to one declaration.

PR #3127, opened 2026-07-27
fix(desktop): allow authorized agent mentions

### LINKED ISSUE #3125: Desktop hides allowlisted agents owned by another user from @mention autocomplete
## Summary

The desktop composer prevents a user from mentioning an agent owned by someone else even when the agent is present in the channel and its `respond_to` policy explicitly permits that user.

Observed in `wss://formidable.communities.buzz.xyz`:

- Testbot is owned by `gabimoncha`.
- Testbot is a `bot` member of `#general`.
- Testbot is configured so Pax may invoke it.
- Gabimoncha can resolve and mention Testbot.
- Pax's `@Testbot` remains plain text and the published event has no Testbot `p` tag, so ACP is never woken.

## Root cause

PR #2149 made the current user's local managed-agent list authoritative for every concrete agent identity in mention autocomplete.

`useMentions.ts` calls `isAgentIdentityInManagedList(candidate, managedAgentPubkeys)` before the existing `respond_to`/allowlist/shared-channel eligibility check. An other-owned agent is therefore discarded even when `getMentionableAgentPubkeys()` has correctly classified it as invocable for the current user.

This conflates two different concepts:

- **managed by me** — ownership/lifecycle control
- **invocable by me** — `owner-only`, `allowlist`, or `anyone` response policy

## Expected behavior

A concrete agent should be mentionable when either:

- it is locally managed by the current user, or
- its relay profile and channel relationship authorize the current user through `respond_to`.

Agents that are neither locally managed nor invocable must remain hidden.

## Acceptance criteria

- An allowlisted user can select an other-owned agent from desktop `@mention` autocomplete.
- An `anyone` agent shared through a joined channel can be selected by other members.
- The sent event contains the selected agent's `p` tag.
- Owner-only and non-allowlisted other-owned agents remain hidden.
- Locally managed agents remain visible regardless of response policy.
- Focused unit and desktop E2E coverage protects the owner/allowlist distinction.

## Suspected files

- `desktop/src/features/messages/lib/useMentions.ts`
- `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts`
- `desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs`
- `desktop/tests/e2e/mentions.spec.ts`

Regression introduced by #2149.

pax-k:
## Summary

- allow relay-discovered agents into mention autocomplete when the current user is authorized by their `respond_to` policy
- preserve filtering for unauthorized or stale agent identities
- cover the other-owned, allowlisted channel-member flow through the signed kind-9 event and its recipient `p` tag

## Root cause

The mention candidate pipeline treated local management ownership as a hard prerequisite before evaluating the existing invocation policy. That conflated who manages an agent with who is allowed to call it, so an agent explicitly allowlisted for Pax was hidden unless Pax also owned it.

## Testing

- focused autocomplete eligibility unit tests (19 passed)
- focused Playwright mention scenarios (3 passed)
- `pnpm check` in `desktop`
- `just ci`

Fixes #3125

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -5,6 +5,7 @@ import {
   coalesceAgentAutocompleteCandidates,
   getMentionableAgentPubkeys,
   getSharedChannelIds,
+  isAgentMentionable,
   isAgentIdentityInManagedList,
   relayAgentIsSharedWithUser,
   shouldHideAgentFromMentions,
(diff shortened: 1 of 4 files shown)


DrZeshanKhan:
Independent verification: I reproduced the failure mode where a relay-discovered agent authorized for the current user appeared online in the directory but was missing from `@` autocomplete because local ownership was treated as an eligibility prerequisite.

I tested this PR's exact branch locally. The complete desktop unit suite passed (3,638 of 3,638), and the focused Playwright scenario passed, including verification of the recipient `p` tag. The negative authorization coverage continues to keep unauthorized or stale relay agents hidden.

This matches the observed bug and the intended behavior.

PR #3132, opened 2026-07-27
fix(cli): merge set-add-policy into existing kind:10100 content instead of replacing it

Draugmar:
## Problem

`buzz channels set-add-policy` — the only CLI path that publishes a kind:10100 agent profile — writes the **entire** content as `{"channel_add_policy": "..."}`. kind:10100 is plain-replaceable, so this silently clobbers any richer profile the identity has published (`respond_to`, `respond_to_allowlist`, `channel_ids`, `name`, ...).

Because desktop mention eligibility reads those fields (`relayAgentIsSharedWithUser`), running the documented command makes an external agent invisible to @-mention autocomplete — the "sharp edge" discussed in #2987 and in the review of #3110: the gate fix there is only reachable for agents whose 10100 is rich, and this command erases exactly that richness.

## Change

- New pure helper `merge_channel_add_policy(existing_content, policy)`: parses the current content object, sets `channel_add_policy`, preserves every other field. Absent/unparsable content degrades to the previous behavior.
- `cmd_set_add_policy` now queries the identity's latest kind:10100 (`{"kinds":[10100],"authors":[self],"limit":1}`) and publishes the merged content. Query errors propagate: failing the command is safer than silently replacing the profile with a minimal one.

## Testing

- 4 unit regression tests on the helper (preserve fields / overwrite policy / no prior content / unparsable content). `cargo test -p buzz-cli`: 254 passed. `cargo fmt --check` and `clippy` clean.
- Verified end-to-end against a live self-hosted relay (`sha-070fb6a`): published a rich 10100, ran patched `set-add-policy`, relay serves the merged content with all fields preserved; the unpatched CLI reproduces the clobber.
- Note for integration tests in this area: two replaceable updates within the same second tie-break nondeterministically — test sequences need distinct `created_at`s.

Refs #2987. Complements #3110 (desktop gate); this addresses the CLI side so the documented registration path produces a mentionable profile.

### DIFF
--- a/crates/buzz-cli/src/commands/channels.rs
+++ b/crates/buzz-cli/src/commands/channels.rs
@@ -1001,6 +1001,22 @@ pub async fn cmd_remove_channel_member(
     Ok(())
 }
 
+/// Merge `channel_add_policy` into an existing kind:10100 content object.
+///
+/// Returns the existing content with the policy field set, preserving every
+/// other field. Absent or unparsable existing content degrades to an object
+/// holding only the policy (the previous behavior).
+fn merge_channel_add_policy(existing_content: Option<&str>, policy: &str) -> String {
+    let mut obj = existing_content
+        .and_then(|c| serde_json::from_str::<serde_json::Map<String, serde_json::Value>>(c).ok())
+        .unwrap_or_default();
+    obj.insert(
+        "channel_add_policy".to_string(),
+        serde_json::Value::String(policy.to_string()),
+    );
+    serde_json::Value::Object(obj).to_string()
+}
+
 /// Set the channel addition policy — sign and submit a kind:10100 (agent profile) event.
 pub async fn cmd_set_add_policy(client: &BuzzClient, policy: &str) -> Result<(), CliError> {
     match policy {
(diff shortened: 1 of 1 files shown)


scottopell:
👍 adding my support for this. I ended up here because external agents aren't reliably `@`-able without this fix.

> ### 2. `set-add-policy` destroys complete profiles
> 
> Kind `10100` is **replaceable as one whole document**. The relay does not merge JSON fields.
> 
> Current upstream does:
> 
> ```bash
> buzz channels set-add-policy --policy owner_only
> ```
> 
> by publishing:
> 
> ```json
> {
>   "channel_add_policy": "owner_only"
> }
> ```
> 
> That replaces the prior rich profile.
> 
> Before:
> 
> ```json
> {
>   "name": "Pup of Concept",
>   "channel_ids": ["agent-lab-id"],
>   "respond_to": "anyone",
>   "channel_add_policy": "anyone"
> }
> ```
> 
> After:
> 
> ```json
> {
>   "channel_add_policy": "owner_only"
> }
> ```
> 
> Now Desktop sees:
> 
> ```text
> channel_ids: []
> respond_to: missing
> ```
> 
> The agent immediately disappears from autocomplete even though its process is still healthy.
> 
>

PR #3139, opened 2026-07-27
fix(mobile): keep TLS on relays joined by invite

### LINKED ISSUE #2662: [Bug] iOS invite join opens the community but remains stuck on “Connecting…”
## Summary

Opening a valid hosted-community invite on iPhone successfully hands off to the Buzz mobile app and selects the correct community, but the app then remains indefinitely on **“Connecting…”** with an empty Home screen.

This blocks a mobile-first rollout for employees who do not have a desktop computer.

## Expected behavior

After the invite is opened and joined, the mobile app should connect to the community relay and render the available channels.

## Actual behavior

The mobile app opens the correct hosted community, shows the correct hostname, and then stays on **“Connecting…”** forever. No channels or content load.

## Reproduction

1. Create a hosted Buzz community at `grupoinnova.communities.buzz.xyz`.
2. Create an invite link from **Community access**.
3. On an iPhone with the Buzz mobile app installed, open that invite link.
4. The app opens and displays `grupoinnova.communities.buzz.xyz`.
5. It remains indefinitely on **“Connecting…”** with an empty Home screen.

## Notes

- This is a hosted community, not a self-hosted relay.
- The deep-link handoff itself appears to work: the mobile app opens the intended community.
- Manual membership via `npub` and desktop-to-mobile pairing were tested separately. Pairing works for the same person's desktop/mobile identity, but it is not a viable onboarding path for employees without a computer because it copies the desktop identity.
- The failing path is specifically a new mobile user opening a community invite.

## Impact

A company cannot invite mobile-only staff reliably: the person appears to enter the app/community but cannot load anything. This makes the invite flow unsuitable for normal employee onboarding.

A screenshot of the device state is available if maintainers need it.

inventivepotter:
## Summary

Communities joined via an invite link never connect: the app dials `ws://` on port 80 instead of `wss://` on 443 and sits on "Reconnecting…" indefinitely.

`RelayConfig.baseUrl` is documented as an HTTP origin, but the two onboarding flows disagree on what they persist:

- **Device pairing** validates and stores `https://` — `pairing_provider.dart:657` throws on anything else.
- **Invite join** stores the relay URL straight off the invite link, and `deep_link.dart:165` always emits `ws://` or `wss://`.

`wsUrl` only special-cased `https://`, so a `wss://` base fell through to the plaintext branch:

```dart
final scheme = uri.scheme == 'https' ? 'wss' : 'ws';   // 'wss' is not 'https'
```

The claim request itself succeeds, because `_claimUrlFromRelay` (`invite_join_provider.dart:242`) maps `wss → https` explicitly. Only the socket path is missing that conversion — which is why the community appears, correctly named, and then never loads.

The same `baseUrl` also feeds `/query` (`relay_session.dart:136`), media upload (`media_upload.dart:765`), Blossom auth (`media_auth.dart:128`) and `relayClientProvider` (`relay_provider.dart:113`), so those requests were malformed too. Where port 80 *does* answer, it is additionally a silent TLS downgrade after `validateInviteRelayUri` insisted on `wss://`.

This folds the websocket schemes back to their HTTP equivalents in `baseUrl` itself, so every consumer is correct by construction rather than needing a second getter remembered at each call site, and communities **already persisted** with `wss://` are repaired on read without a migration. `community_icon_provider.dart:46` already performs this same conversion locally.

One subtlety worth flagging for review: the normalization is derived in the getter rather than applied in the constructor, so the constructor stays `const`. The compile-time fallback at `relay_provider.dart:77` relies on const canonicalization for a stable identity across rebuilds, and Riverpod's `defaultUpdateShouldNotify` is `previous != next` (`element.dart:361`), which falls back to identity for this class. A `factory` constructor here yields a fresh instance per rebuild, which tears down and resubscribes every listener — `channels_provider_test.dart` catches it as an unexpected unsubscribe during reconnect.

### Related issue

Fixes #2662.

### Testing

`flutter test` — **705 passed, 1 skipped, 0 failed**
`flutter analyze` — No issues found
`dart format --set-exit-if-changed .` — 249 files, 0 changed

Run against the Hermit-pinned SDK (Flutter 3.41.7 / Dart 3.11.5), matching CI.

10 new unit tests in `mobile/test/shared/relay/relay_config_test.dart` covering both onboarding schemes, `http`/`https` passthrough, non-default ports, and agreement between the invite and pairing paths for the same relay.

Verified end-to-end against a self-hosted relay behind `tailscale serve`, which terminates TLS on 443 and leaves port 80 closed. Relay logs show the invite claim succeeding over HTTPS at the moment of joining, while no WebSocket connection ever arrives — no `WebSocket connection established`, no NIP-42 auth, no `kind:0` profile, no push registration — across the relay's entire history, even though the member row is present and correct. Port-80 refusals are not logged by `tailscaled`'s netstack, which is why the retries leave no trace server-side. Reproduced on both iOS and Android.

### DIFF
--- a/mobile/lib/features/activity/compose_drafts_provider.dart
+++ b/mobile/lib/features/activity/compose_drafts_provider.dart
@@ -79,7 +79,13 @@ class ComposeDraftsNotifier extends Notifier<List<ComposeDraft>> {
     _prefsKey = '$_draftsPrefsKey:${config.baseUrl}:$pubkey';
 
     final prefs = ref.read(savedPrefsProvider);
-    final raw = prefs.getString(_prefsKey);
+    final raw = readMigratedPref<String>(
+      prefs,
+      canonicalKey: _prefsKey,
+      legacyKey: '$_draftsPrefsKey:${config.storedOrigin}:$pubkey',
+      read: prefs.getString,
+      write: prefs.setString,
+    );
     if (raw == null) return const [];
     try {
       final decoded = jsonDecode(raw);
--- a/mobile/lib/features/search/recent_searches_provider.dart
+++ b/mobile/lib/features/search/recent_searches_provider.dart
@@ -19,8 +19,16 @@ class RecentSearchesNotifier extends Notifier<List<String>> {
     final pubkey = ref.watch(myPubkeyProvider) ?? 'anon';
     _prefsKey = '$_recentSearchesPrefsKey:${config.baseUrl}:$pubkey';
 
+    final prefs = ref.read(savedPrefsProvider);
     final stored =
-        ref.read(savedPrefsProvider).getStringList(_prefsKey) ?? const [];
+        readMigratedPref<List<String>>(
+          prefs,
+          canonicalKey: _prefsKey,
+          legacyKey: '$_recentSearchesPrefsKey:${config.storedOrigin}:$pubkey',
+          read: prefs.getStringList,
+          write: prefs.setStringList,
+        ) ??
+        const [];
     return List.unmodifiable(
       stored
           .map((query) => query.trim())
--- a/mobile/lib/shared/relay/identity_scoped_prefs.dart
+++ b/mobile/lib/shared/relay/identity_scoped_prefs.dart
@@ -1,0 +1,50 @@
+import 'dart:async';
+
+import 'package:shared_preferences/shared_preferences.dart';
+
+/// Reads an identity-scoped preference, migrating values left under a
+/// pre-canonicalization relay origin.
+///
+/// Identity-scoped keys embed the relay origin, and that origin used to be
+/// whatever scheme the onboarding flow happened to persist — `wss://` for an
+/// invite join, `https://` for device pairing. Now that [RelayConfig.baseUrl]
+/// canonicalizes the scheme, an install that joined by invite would compute a
+/// different key than the one its data was written under, leaving that data on
+/// disk but unreachable.
+///
+/// The value under [canonicalKey] wins. Otherwise a value under [legacyKey] is
+/// returned straight away and promoted onto the canonical key in the
+/// background. The legacy entry is removed only once the copy reports success,
+/// so a migration interrupted mid-flight leaves the original readable and the
+/// next read simply retries.
+///
+/// Pass matching accessors for the stored type — `getString`/`setString`, or
+/// `getStringList`/`setStringList`.
+T? readMigratedPref<T>(
+  SharedPreferences prefs, {
+  required String canonicalKey,
+  required String legacyKey,
+  required T? Function(String key) read,
+  required Future<bool> Function(String key, T value) write,
+}) {
+  final canonical = read(canonicalKey);
+  if (canonical != null) return canonical;
+
+  // Pairing-created communities already store an HTTP origin, so the two keys
+  // coincide and there is nothing to migrate.
+  if (canonicalKey == legacyKey) return null;
+
+  final legacy = read(legacyKey);
+  if (legacy == null) return null;
+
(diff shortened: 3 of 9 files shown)


wesbillman (commented):
Carl (an AI agent) is commenting on Wes's behalf.

**Merge blocker — canonicalizing `baseUrl` changes persisted storage identities for existing invite users.**

The normalization itself is correct for network consumers, but current `main` now also embeds `config.baseUrl` in SharedPreferences keys:

- `mobile/lib/features/activity/compose_drafts_provider.dart:79` — `compose_drafts_v1:${config.baseUrl}:$pubkey`
- `mobile/lib/features/search/recent_searches_provider.dart:20` — `recent_searches_v1:${config.baseUrl}:$pubkey`

An existing invite-joined community has stored values keyed under `wss://…`. After this getter returns `https://…`, the app reads a different key. The old data remains on disk but becomes unreachable; for compose drafts, this presents as loss of unsent user-authored text after upgrade.

This is a merge-time interaction with changes added to `main` after this PR branched, rather than a flaw visible from this PR's isolated diff. I reproduced it against current `main`.

Before merge, please preserve storage-key compatibility—for example, retain access to the raw stored origin and, when the canonical key is absent, migrate/fallback from the legacy `wss://`/`ws://` key, copying successfully before removing the legacy value. Please add an upgrade regression test showing an existing `wss://` community retains its stored draft; the recent-search key needs the same compatibility handling.

The URL normalization, ports/IPv6 handling, Riverpod identity behavior, network consumers, mergeability, and CI otherwise reviewed cleanly.

inventivepotter:
Confirmed against current `main` — both call sites embed the origin in a SharedPreferences key, and an invite-joined install would compute `https://…` while its data sits under `wss://…`. The branch was ~50 commits behind, so neither file existed in its diff; it is now rebased onto current `main` and the compatibility handling is in `d81c95f`.

I audited the remaining `config.baseUrl` consumers rather than assuming the two named were exhaustive. Two others embed the origin in an identity string — `observer_subscription.dart:73` (`identityKey`) and `compose_bar.dart:96` (`draftIdentity`) — but both are compared only against a value recomputed in the same run (a field and a `useRef`), never persisted, so a changed origin is self-consistent and needs no migration. The rest are network consumers, which want the canonical form.

What changed:

- `RelayConfig.storedOrigin` exposes the persisted origin, documented as being for storage-key compatibility only and explicitly not for network I/O.
- `readMigratedPref` in `shared/relay/identity_scoped_prefs.dart` prefers the canonical key; failing that it returns the legacy value immediately and promotes it in the background. `prefs.remove(legacyKey)` runs only after the copy resolves `true`, so an interrupted migration leaves the original readable and the next read retries. Where the two keys coincide — pairing-created communities, which already stored an HTTP origin — it returns without touching storage.
- Both providers read through it; writes continue on the canonical key.

Tests: an upgrade regression test per provider seeding data under the legacy `wss://` key and asserting the draft and the recent searches survive, plus the canonical key is populated and the legacy entry cleared. I verified both fail with the migration reverted, so they discriminate rather than pass vacuously. `identity_scoped_prefs_test.dart` covers the promote-then-remove ordering, idempotency across repeated reads, the coinciding-key short-circuit, and both `String` and `List<String>` accessors.

`flutter test` 906 passed / 1 skipped / 0 failed, `flutter analyze` clean, `dart format` clean, on the Hermit-pinned SDK.

wesbillman (approved):

Rodribm10:
Thanks for merging this. We’re testing the iOS app for a workplace rollout, and the App Store version (0.4.12) still gets stuck on "Connecting…" after accepting an invite.

Could you please ship a new iOS build that includes this fix (#3139)? This blocks mobile-only onboarding for our team. We’re happy to retest once it is available.

PR #3141, opened 2026-07-27
feat(invites): add use-limited invite links

kalvinnchau:
## Summary

- add database-backed v2 invite links with optional maximum-use limits and atomic final-slot redemption
- preserve v1 invite compatibility while adding exhausted/expired/invalid client handling across desktop, web, and mobile
- emit structured claim-outcome logs with community, invite ID, outcome, maximum uses, and post-claim count

## Verification

- `cargo test -p buzz-db` (85 passed, 134 Postgres-dependent ignored)
- `cargo clippy -p buzz-db --all-targets -- -D warnings`
- desktop `npm run typecheck`
- push hook: desktop checks/tests, desktop Tauri tests, Rust tests, and branch-skew passed
- Postgres integration tests were previously reviewed green at the pre-rebase tree; local rerun on this session was unavailable because Postgres/Docker were not running
- mobile push-hook check could not start because Flutter is unavailable locally

### DIFF
--- a/crates/buzz-db/src/relay_invite.rs
+++ b/crates/buzz-db/src/relay_invite.rs
@@ -152,6 +152,33 @@ fn log_claim_outcome(
     );
 }
 
+/// Maximum rows deleted by one retention sweep so cleanup cannot monopolize
+/// the invite table on a busy deployment.
+const RETENTION_SWEEP_BATCH_SIZE: i64 = 1_000;
+
+/// Delete one bounded batch of invite rows expired before `cutoff`.
+///
+/// The relay calls this from its leader-only periodic tick. Ordering by the
+/// expiry index makes old rows drain first without turning cleanup into an
+/// unbounded transaction.
+pub async fn reap_expired_relay_invites(pool: &PgPool, cutoff: DateTime<Utc>) -> Result<u64> {
+    let result = sqlx::query(
+        "DELETE FROM relay_invites \
+         WHERE (community_id, id) IN (\
+             SELECT community_id, id FROM relay_invites \
+             WHERE expires_at < $1 \
+             ORDER BY expires_at \
+             LIMIT $2\
+         )",
+    )
+    .bind(cutoff)
+    .bind(RETENTION_SWEEP_BATCH_SIZE)
+    .execute(pool)
+    .await?;
+
+    Ok(result.rows_affected())
+}
+
 /// Atomically claim a v2 relay invite.
 ///
 /// Executes the full redemption in one PostgreSQL transaction:
@@ -174,7 +201,7 @@ fn log_claim_outcome(
 pub async fn claim_relay_invite(
     pool: &PgPool,
     community: CommunityId,
-    token_hash: &[u8],
+    token_hash: &[u8; 32],
     claimer_pubkey: &str,
     policy_version: Option<&str>,
 ) -> Result<ClaimOutcome> {
@@ -204,7 +231,9 @@ pub async fn claim_relay_invite(
     let use_count: i32 = invite.try_get("use_count")?;
     let expires_at: DateTime<Utc> = invite.try_get("expires_at")?;
 
-    // 4. Expired check.
+    // Expiry is checked before membership deliberately. An expired bearer must
+    // not authorize fresh policy-acceptance evidence, even for an existing
+    // member; exhausted-but-live invites remain valid for idempotent retries.
     if expires_at <= Utc::now() {
         tx.rollback().await?;
         log_claim_outcome(
@@ -546,6 +575,44 @@ mod tests {
         assert_eq!(use_count(&pool, community_a, invite.invite_id).await, 0);
         delete_test_community(&pool, community_a).await;
         delete_test_community(&pool, community_b).await;
+    }
+
+    #[tokio::test]
+    #[ignore = "requires Postgres"]
+    async fn retention_sweep_deletes_only_invites_older_than_cutoff() {
+        let pool = setup_pool().await;
+        let community = make_test_community(&pool).await;
+        let old = mint_relay_invite(&pool, community, "owner", 3600, Some(1))
+            .await
+            .expect("mint old invite");
+        let recent = mint_relay_invite(&pool, community, "owner", 3600, Some(1))
+            .await
+            .expect("mint recent invite");
+        let cutoff = Utc::now() - chrono::Duration::days(30);
+
+        sqlx::query("UPDATE relay_invites SET expires_at = $1 WHERE community_id = $2 AND id = $3")
+            .bind(cutoff - chrono::Duration::seconds(1))
+            .bind(community.as_uuid())
+            .bind(old.invite_id)
+            .execute(&pool)
+            .await
+            .expect("age old invite");
+
+        assert_eq!(
+            reap_expired_relay_invites(&pool, cutoff)
+                .await
+                .expect("reap expired invites"),
+            1
+        );
+        let remaining: Vec<Uuid> =
+            sqlx::query_scalar("SELECT id FROM relay_invites WHERE community_id = $1 ORDER BY id")
+                .bind(community.as_uuid())
+                .fetch_all(&pool)
+                .await
+                .expect("read remaining invites");
+        assert_eq!(remaining, vec![recent.invite_id]);
+
+        delete_test_community(&pool, community).await;
     }
 
     #[tokio::test]
--- a/crates/buzz-db/src/relay_invite.rs
+++ b/crates/buzz-db/src/relay_invite.rs
@@ -149,6 +149,23 @@ pub async fn mint_relay_invite(
     })
 }
 
+fn log_claim_outcome(
+    community: CommunityId,
+    invite_id: Option<uuid::Uuid>,
+    outcome: &'static str,
+    max_uses: Option<i32>,
+    use_count: Option<i32>,
+) {
+    tracing::info!(
+        community = %community,
+        invite_id = ?invite_id,
+        outcome,
+        max_uses = ?max_uses,
+        use_count = ?use_count,
+        "relay invite claim completed"
+    );
+}
+
 /// Atomically claim a v2 relay invite.
 ///
 /// Executes the full redemption in one PostgreSQL transaction:
@@ -192,6 +209,7 @@ pub async fn claim_relay_invite(
     // 3. No matching invite.
     let Some(invite) = row else {
         tx.rollback().await?;
+        log_claim_outcome(community, None, "invalid", None, None);
         return Ok(ClaimOutcome::Invalid);
     };
 
@@ -203,6 +221,13 @@ pub async fn claim_relay_invite(
     // 4. Expired check.
     if expires_at <= Utc::now() {
         tx.rollback().await?;
+        log_claim_outcome(
+            community,
+            Some(invite_id),
+            "expired",
+            max_uses,
+            Some(use_count),
+        );
         return Ok(ClaimOutcome::Expired);
     }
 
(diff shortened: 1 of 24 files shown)


crates/buzz-db/src/relay_invite.rs
```diff
@@ -0,0 +1,604 @@
...
+            (3600, Some(-1)),
+            (3600, Some(MAX_INVITE_USES + 1)),
+        ] {
+            let error = validate_mint_inputs(ttl, max_uses).expect_err("invalid mint contract");
+            assert!(matches!(error, crate::DbError::InvalidData(_)), "{error:?}");
+        }
+    }
+
+    #[tokio::test]
+    #[ignore = "requires Postgres"]
```
wpfleger96, line 452: 🤖 **[IMPORTANT — flagged by both reviewers' analysis, blocking]** Required CI never executes these tests, so the PR's central guarantee is unproven in the gate.

All five Postgres-backed tests in this module are `#[ignore]`. The unit job runs `just test-unit`, which invokes `cargo nextest run -p buzz-db --lib` — ignored tests are excluded by design (`Justfile:275-285`). The backend-integration job can't compensate: its nextest archive contains only `buzz-relay` and `buzz-test-client` (`.github/workflows/ci.yml:341-347`), and its filter is `-E 'package(buzz-relay) and test(claim_)'` (`ci.yml:674-679`).

That misses: final-slot serialization (`concurrent_claims_serialize_the_final_slot`), rollback (`policy_evidence_failure_rolls_back_membership_and_consumption`), expiry/tenant isolation, bounded-retry idempotence, and unlimited counting — plus the new endpoint tests `bounded_v2_claims_publish_side_effects_only_for_joined` and `malformed_and_unknown_v2_codes_are_forbidden_without_v1_fallback`, whose names don't match the `claim_` substring. All 32 checks can stay green after a regression in atomic redemption.

**Suggested fix:** add `-p buzz-db` to the backend nextest archive and select these tests by module/package (e.g. `test(/relay_invite::tests/)` plus the new v2 endpoint tests) rather than relying on the incidental `claim_` name prefix. Selection should be explicit and stable, not inferred from test names.

wpfleger96 (commented):
🤖 Combined review from two independent passes (deduped). Verified head: `a4af9622d`.

**Overall:** the core design is strong and both reviews converged on that — 32-byte opaque secrets with hash-only storage, tenant-scoped `(community_id, token_hash)` lookups, the full redemption (membership insert, policy evidence, `use_count` increment) in one transaction under `SELECT ... FOR UPDATE`, exact-prefix `v2.` routing with no v1 fallback, and NIP-43 side effects only on `Joined`. The same-member/different-invite race is handled correctly (`ON CONFLICT ... DO NOTHING` + `rows_affected()` returns the loser as `AlreadyMember` without consuming the invite). Test coverage of the concurrency and rollback contracts is unusually thorough — one reviewer ran all 5 ignored `buzz-db::relay_invite` tests and all 12 ignored `buzz-relay::api::invites` tests against isolated Postgres and everything passed.

**Findings** are inline: 1 IMPORTANT (CI never executes the tests that prove the atomic-redemption contract — independently verified by both reviewers against `Justfile` and `ci.yml` at this head), 1 IMPORTANT-but-follow-up-acceptable (no retention path for expired invite rows — flagged by both reviewers), and 4 minor/optional notes.

Severity tags per comment; nothing here questions the design itself.

migrations/0025_relay_invites.sql
```diff
@@ -0,0 +1,31 @@
...
+-- the wrong tenant host returns Invalid — there is no cross-tenant lookup by
+-- hash alone. `FOR UPDATE` during claim serializes concurrent claims for one
+-- invite across relay processes; membership insertion, join-policy evidence,
+-- and use_count increment share a single commit so exactly one claimant can
+-- win the final slot.
+--
+-- max_uses is optional: NULL means unlimited (preserving current behavior).
+-- use_count is always incremented for new members, even when unlimited, for
+-- observability. role is pinned to 'member' — invite links never grant admin.
+CREATE TABLE relay_invites (
```
wpfleger96, line 18: 🤖 **[IMPORTANT — flagged by both reviewers; a named follow-up is acceptable]** No retention path for expired/exhausted invite rows.

Every "copy invite link" click mints a fresh row, and nothing ever deletes them — the only `DELETE FROM relay_invites` at this head is test cleanup. Growth is monotonic and per-click.

Fine to land as-is, but please file a follow-up at merge time: a bounded retention sweep (e.g. `DELETE FROM relay_invites WHERE expires_at < now() - interval '30 days'` in an existing leader-only tick) plus an `expires_at` index to support it.

web/src/features/invite/ui/InvitePage.tsx
```diff
@@ -110,8 +110,12 @@ export function InvitePage({ code }: { code: string }) {
       await claimInviteInBrowser(code, receipt);
       window.location.assign("/");
     } catch (error) {
+      const message =
+        error instanceof Error ? error.message : "Could not claim this invite.";
       setBrowserJoinError(
-        error instanceof Error ? error.message : "Could not claim this invite.",
+        message.includes("invite_exhausted")
```
wpfleger96: 🤖 **[MINOR — flagged by both reviewers]** `invite_exhausted` gets friendly copy here, but `invite_expired` and `invite_invalid` still surface as raw sentinel strings to the user. Pre-existing behavior, but this PR touched the exact line — mapping all three costs two more branches and avoids inconsistent user-facing errors.

crates/buzz-db/src/relay_invite.rs
```diff
@@ -0,0 +1,604 @@
...
+/// 10. Increment `use_count`.
+/// 11. Commit.
+///
+/// `FOR UPDATE` serializes concurrent claims so exactly one claimant wins the
+/// final slot. Membership insertion, policy evidence, and consumption share
+/// one commit — a failure in any rolls back all.
+pub async fn claim_relay_invite(
+    pool: &PgPool,
+    community: CommunityId,
+    token_hash: &[u8],
```
wpfleger96: 🤖 **[MINOR]** `token_hash: &[u8]` — taking `&[u8; 32]` would make a wrong-length hash a compile error instead of a silent `Invalid` at runtime. The only caller already has a `[u8; 32]` from `hash_v2_code`.

crates/buzz-db/src/relay_invite.rs
```diff
@@ -0,0 +1,604 @@
...
+    /// `None` means unlimited; `Some(n)` means at most `n` uses.
+    pub max_uses: Option<i32>,
+    /// Remaining uses at mint time (equals `max_uses` when bounded, `None`
+    /// when unlimited).
+    pub uses_remaining: Option<i32>,
+    /// The invite's database-generated UUID.
+    pub invite_id: uuid::Uuid,
+}
+
+fn validate_mint_inputs(ttl_secs: u64, max_uses: Option<i32>) -> Result<()> {
```
wpfleger96, line 75: 🤖 **[MINOR]** Duplicate bounds validation with a mismatched failure mode: `validate_mint_request` (relay layer) returns 400, but if this DB-layer check ever fires instead (relay drift, a new caller), `mint_invite` maps `InvalidData` through `internal_error` → 500. Defense in depth is fine — consider either a comment noting the relay layer owns the 400 contract, or mapping `InvalidData` to 400 at the call site.

crates/buzz-db/src/relay_invite.rs
```diff
@@ -0,0 +1,604 @@
...
+        log_claim_outcome(community, None, "invalid", None, None);
+        return Ok(ClaimOutcome::Invalid);
+    };
+
+    let invite_id: uuid::Uuid = invite.try_get("id")?;
+    let max_uses: Option<i32> = invite.try_get("max_uses")?;
+    let use_count: i32 = invite.try_get("use_count")?;
+    let expires_at: DateTime<Utc> = invite.try_get("expires_at")?;
+
+    // 4. Expired check.
```
wpfleger96: 🤖 **[MINOR]** Outcome-ordering asymmetry: an existing member presenting an *exhausted* invite gets `already_member` (200), but presenting an *expired* one gets `invite_expired` (403), because the expiry check precedes the membership check. Defensible either way — flagging in case the idempotent-retry UX is meant to be uniform across the two terminal states.

crates/buzz-relay/src/invite_token.rs
```diff
@@ -49,10 +49,10 @@ use buzz_core::tenant::CommunityId;
 type HmacSha256 = Hmac<Sha256>;
 
 /// Default invite lifetime: 72 hours.
-pub const DEFAULT_INVITE_TTL_SECS: u64 = 72 * 60 * 60;
+pub use buzz_core::invite::DEFAULT_INVITE_TTL_SECS;
 
 /// Maximum invite lifetime a mint request may ask for: 30 days.
-pub const MAX_INVITE_TTL_SECS: u64 = 30 * 24 * 60 * 60;
+pub use buzz_core::invite::MAX_INVITE_TTL_SECS;
```
wpfleger96, line 55: 🤖 **[MINOR]** After this PR, v1 `mint_invite` in this module has no production caller — only tests use it (the mint endpoint now goes through `mint_relay_invite`). Fine during the v1 drain window; consider tagging it for removal when v1 claims are retired so it doesn't linger as an unminted-but-mintable code path.

crates/buzz-db/src/relay_invite.rs
kalvinnchau: 🤖 Fixed in `c3cf887e9`. The backend nextest archive now includes `buzz-db`, and the integration step uses an explicitly parenthesized module/package filter for all ignored `buzz-db::relay_invite::tests` and `buzz-relay::api::invites::tests`. I verified the exact expression parses and selects 18 ignored tests, including the transaction, malformed-v2, and side-effect coverage called out here.

migrations/0025_relay_invites.sql
kalvinnchau: 🤖 Implemented directly in `c3cf887e9` rather than deferring it. Migration 0025 and `schema/schema.sql` now add an `expires_at` index. A leader-only usage tick deletes invites older than a 30-day grace period, oldest first, in bounded batches of 1,000 rows. A Postgres-backed regression test verifies that only rows older than the cutoff are removed.

web/src/features/invite/ui/InvitePage.tsx
kalvinnchau: 🤖 Fixed in `c3cf887e9`. I extracted `inviteClaimErrorMessage`, which maps `invite_exhausted`, `invite_expired`, and `invite_invalid` to stable user-facing recovery guidance while preserving unexpected error text.

crates/buzz-db/src/relay_invite.rs
kalvinnchau: 🤖 Fixed in `c3cf887e9`. Both the database function and the `Db` facade now accept `token_hash: &[u8; 32]`, so an invalid hash length is rejected by the type system rather than degrading to an `Invalid` lookup result.

crates/buzz-db/src/relay_invite.rs
kalvinnchau: 🤖 Fixed in `c3cf887e9`. The relay mint call now maps `buzz_db::DbError::InvalidData` to HTTP 400 and reserves the 500 path for actual internal/database failures. The DB validation remains defense in depth without changing the public error contract for future callers.

crates/buzz-db/src/relay_invite.rs
kalvinnchau: 🤖 Kept the behavior and documented the contract in `c3cf887e9`. Expiry intentionally precedes membership: an expired bearer must not authorize fresh policy-acceptance evidence, even for an existing member. A live-but-exhausted invite still reaches the membership check so an existing member can retry idempotently without consuming capacity.

crates/buzz-relay/src/invite_token.rs
kalvinnchau: 🤖 Fixed in `c3cf887e9`. Legacy v1 `mint_invite` is now `#[cfg(test)]`; its documentation says production minting uses database-backed v2 codes and directs removal of the helper together with v1 claim verification after the compatibility drain.

wpfleger96 (approved):

PR #806, opened 2026-06-01
Fix desktop right-side panel chrome overlap

thomaspblock:
## Summary
- Fix responsive right-side panel behavior for threads, agent activity, and user profiles so narrow layouts use a single-panel takeover below the global top chrome while preserving search and action buttons.
- Systematize panel layering and width observation with shared z-index constants, element-width breakpoints, and 300px minimum column behavior for both channel panels and the Home inbox split view.
- Refine Home inbox detail alignment, message row spacing, avatar sizing, sticky date divider offset, and composer width so one-column and two-column views line up consistently.
- Polish agent activity UI by stabilizing the header, using agent/requester avatars where appropriate, improving transcript alignment, and avoiding clipped inline activity avatars.
- Tune sheet/dialog light-theme treatment so modal dimming follows Sprout's selected theme, channel management panels stay light in light mode, and the members header drop shadow is removed.

## Test plan
- [ ] Open a channel and verify thread, agent activity, and profile panels do not overlap the top search/actions at narrow widths.
- [ ] Resize channel panels and confirm each column keeps a 300px minimum before collapsing into single-panel view.
- [ ] Open Home inbox, resize the list/detail split, and confirm two-column and one-column detail alignment stays consistent.
- [ ] Scroll message timelines and inbox details to confirm sticky date dividers stay below headers and composers do not expose content underneath.
- [ ] Open Members, Add agent, and Channel settings in light and dark themes to confirm dimming, panel backgrounds, and header shadows look correct.

Made with [Cursor](https://cursor.com)

PR #3151, opened 2026-07-27
Polish composer activity layout and transitions

morgmart:
## Summary
- Replace the permanently reserved composer activity row with a conditional, content-driven accessory for channel and thread composers.
- Keep the composer dock geometrically stable while activity appears, so the composer’s bottom edge resizes smoothly without shifting the conversation.
- Preserve translucent backdrop blur during the resize with a stable dock-level blur layer, and keep thread overlay/focus-mode alignment consistent.
- Keep bottom-pinned virtualized conversations fully visible as composer content, zoom, or viewport height changes without repinning readers who have scrolled into history.
- Refine the activity lockup with aligned avatars/text and a subtle, reduced-motion-safe shimmer.
- Centralize the dock’s quiet inset, activity rail, released space, and activity offset in one CSS-variable geometry contract.

This takes a different, systemic route from the spacing reduction proposed in #2602 and supersedes that approach.

### Related issue
Related PR: #2602

### Testing
- `./scripts/check-branch-skew.sh`
- `just desktop-check`
- `just desktop-test` — 3,638 passing
- `useAnchoredScroll.test.mjs` — virtualized viewport resize follows the explicit bottom state
- Focused desktop smoke E2E — 6 passing across stable dock geometry, multiline growth, viewport resize, reduced motion, blur ownership, and thread overlay alignment
- Pre-push hooks passed for organization safety, branch skew, desktop checks/tests, Rust tests, workspace tests, and desktop Tauri tests
- Mobile pre-push is independently red on latest `main`; the exact `activity_page_test.dart` compiler failure reproduces on untouched `origin/main`
- Visually tested channel and thread composers across quiet/activity states, multiline composer growth, and thread overlay/focus mode

https://github.com/user-attachments/assets/ed0b08d9-18e0-4061-b272-ab509dbcd8ee

### DIFF
--- a/desktop/src/features/channels/ui/BotActivityBar.tsx
+++ b/desktop/src/features/channels/ui/BotActivityBar.tsx
@@ -164,7 +164,7 @@ export function BotActivityComposerAction({
           className={cn(
             "inline-flex items-center justify-center rounded-full border border-border/60 bg-background font-medium text-muted-foreground transition-colors hover:border-primary/30 hover:bg-primary/5 hover:text-foreground focus-visible:outline-hidden focus-visible:ring-1 focus-visible:ring-ring data-[state=open]:border-primary/40 data-[state=open]:bg-primary/10 data-[state=open]:text-primary",
             isInline
-              ? "h-7 min-w-0 gap-2 overflow-visible border-transparent bg-transparent px-0 text-xs font-semibold leading-none shadow-none hover:border-transparent hover:bg-transparent data-[state=open]:border-transparent data-[state=open]:bg-transparent"
+              ? "min-w-0 gap-1.5 overflow-visible border-transparent bg-transparent px-0 text-xs font-normal leading-normal shadow-none hover:border-transparent hover:bg-transparent data-[state=open]:border-transparent data-[state=open]:bg-transparent"
               : "h-9 min-w-9 gap-1.5 px-2 text-xs",
           )}
           data-testid="bot-activity-composer-trigger"
@@ -178,15 +178,13 @@ export function BotActivityComposerAction({
           onMouseLeave={closeWithDelay}
           type="button"
         >
-          <span className="flex items-center overflow-visible py-px -space-x-1">
+          <span className="flex h-4.5 items-center overflow-visible -space-x-1">
             {workingAgents.slice(0, 2).map((agent) => (
               <UserAvatar
                 avatarUrl={agentAvatarUrl(agent)}
                 className={cn(
                   "border border-background",
-                  isInline
-                    ? "!h-[18px] !w-[18px] shadow-xs ring-1 ring-primary/25 text-3xs"
-                    : "shrink-0",
+                  isInline ? "!h-4.5 !w-4.5 text-3xs" : "shrink-0",
                 )}
                 displayName={agent.name}
                 key={agent.pubkey}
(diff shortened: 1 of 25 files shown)


tellaho (changes requested):
🤖 Requesting changes on four blocking seams; the overall dock/CSS-variable architecture looks sound.

1. **Preserve `useComposerHeightPadding`’s growth-only fallback contract.** The `padding > previousPadding` → `padding !== previousPadding` change makes composer shrink force-scroll both `MessageThreadPanel` and `InboxDetailPane` to the absolute bottom (immediately and again on rAF), even though neither supplies the new semantic `settleAtBottom` callback. A reader within the 32px near-bottom threshold can therefore be pulled down when activity disappears. Blame/history confirms the growth-only condition was deliberately introduced in #806 and preserved through #941/#1698, and the hook documentation still says it follows when padding *increases*. Please either scope shrink settling to the channel consumer that explicitly supplies `settleAtBottom`, or give the other consumers equivalent semantic anchoring plus shrink regression coverage and update the contract documentation.

2. **Do not use `toolbarExtraActions` truthiness as thread activity state.** In `MessageThreadPanel`, `Boolean(toolbarExtraActions)` now controls whether the activity rail is reserved, while the prop’s name/type still describe a generic `ReactNode`; `MessageComposerProps` also has a different `toolbarExtraActions` that genuinely renders in the toolbar. A future static action would permanently open the rail. Please represent activity explicitly and rename/separate the accessory content from its visibility signal.

3. **Make dock mode explicit instead of encoding it in `bottomAccessoryVisible === undefined`.** `undefined` means legacy blur ownership/no quiet spacer, while both `true` and `false` opt into dock blur ownership and spacer behavior. That is a hidden tri-state behind a boolean-shaped prop and is duplicated across `MessageComposer` and `ComposerDockToolbar`. Please separate/rename the mode and visibility contracts so `false` cannot unexpectedly alter layout/blur ownership.

4. **Restore the contract documentation lost in the `MessageComposerProps` extraction.** The new `MessageComposer.types.ts` drops the existing documentation for `autoSubmitDraftKey`, `editTarget.imetaMedia`, `onEditLastOwnMessage`, and `onCaptureSendContext`. Those comments describe fire-once, NIP-92, return-value, and capture-before-await behavior that callers need to preserve.

morgmart:
🤖 Addressed all four requested seams in `fbc7f7d2a`:

1. **Growth-only fallback restored.** `useComposerHeightPadding` now allows shrink settling only through an explicit semantic `settleAtBottom` owner; thread/inbox fallback scrolling remains initial-measurement/growth-only. Added a regression policy test for shrink.
2. **Thread activity is explicit.** Replaced generic `toolbarExtraActions` with separate `activityAccessoryContent` and `activityAccessoryVisible` contracts. Rendering is gated by the visibility source so stale content cannot leak when human typing opens the rail.
3. **Dock mode is explicit.** Replaced the optional-boolean tri-state with `layoutMode="dock" | "standalone"`. Activity visibility remains owned by the surrounding dock; `MessageComposer` only decides blur/layout ownership.
4. **Composer prop docs restored.** Restored the original fire-once auto-submit, NIP-92 attachment, ↑-to-edit return-value, and capture-before-await contracts in `MessageComposer.types.ts`.

Also rebased onto latest `main`. Since `useAnchoredScroll.ts` was left four lines below its hard ceiling upstream, extracted pure scroll policy and virtualized viewport-resize lifecycle into focused modules rather than adding an exception.

Verification on the final rebased tree:
- `desktop-check` / typecheck / file-size / token guards pass
- 3,717 desktop tests pass
- 10 focused scroll/composer policy tests pass
- 6 focused composer smoke tests pass
- pre-push desktop, Rust unit, mobile, and Tauri checks pass

The full integration pre-push command was blocked before tests by stale local Postgres migration state (`migration 25 was previously applied but is missing in the resolved migrations`), unrelated to this desktop-only diff.

tellaho (approved):
🤖 Re-reviewed at `fbc7f7d2ad5a44bf70a33a441098fd0c502a1804`. All four requested changes are addressed:

- composer shrink now delegates only to an explicit semantic bottom owner; legacy thread/inbox fallback remains initial-measurement/growth-only, with regression coverage
- thread activity visibility and content are explicit, separate contracts
- dock vs. standalone layout/blur ownership is explicit rather than hidden in an optional boolean
- the extracted composer prop contracts are restored

I also reviewed the scroll-policy/viewport-resize extraction introduced to stay within the file limit; it preserves the existing behavior and keeps its policy tests. Exact-head CI is green. Approved.

PR #3152, opened 2026-07-27
feat(buzz-agent): first-class ollama provider alias + local-models docs

### LINKED ISSUE #3145: Local models (Ollama) with buzz-agent: undocumented, undiscoverable, needs first-class provider alias
## Summary

Local models (Ollama, llama.cpp, vLLM, etc.) are **already usable** with `buzz-agent` through the generic OpenAI-compatible provider — but nothing in the repo says so. There is zero mention of `ollama` anywhere in the codebase or docs, and the configuration is undiscoverable. Operators reasonably assume local models are unsupported, or set env vars on the **relay** (which never calls an LLM) and conclude "it doesn't work."

## What works today (undocumented)

`buzz-agent` accepts any OpenAI-compatible endpoint:

```bash
BUZZ_AGENT_PROVIDER=openai        # or "openai-compat"
OPENAI_COMPAT_BASE_URL=http://localhost:11434/v1   # Ollama
OPENAI_COMPAT_MODEL=qwen2.5:7b-instruct
OPENAI_COMPAT_API_KEY=ollama      # any non-empty string; Ollama ignores it
OPENAI_COMPAT_API=chat            # or auto → chat for non-*.openai.com hosts
```

Verified against a local Ollama (`qwen2.5:7b-instruct`) — chat completions work, and tool-capable models are reported via `/api/tags` (`capabilities: ["completion","tools"]`).

## The gaps

1. **Docs gap** — no `docs/` page, README section, or example covers local models. The string "ollama" appears nowhere in the repo.
2. **UX gap in `buzz-agent` config** — Ollama requires a *dummy* API key (`req("OPENAI_COMPAT_API_KEY")`), and the operator must know the magic base URL `http://localhost:11434/v1`. A first-class `BUZZ_AGENT_PROVIDER=ollama` alias could default both (key optional, base URL defaulted) while still allowing overrides.
3. **Architecture confusion** — env vars set on the `buzz-relay` container/process silently do nothing for model selection; the LLM config lives wherever the ACP runtime (`buzz-agent` et al.) runs. Worth stating explicitly in deployment docs.
4. **(Follow-up, separate PR) Desktop discoverability** — the desktop agent config offers "OpenAI-compatible" as a provider but requires an API key and gives no hint that Ollama works; an "Ollama (local)" preset with no required key and a prefilled base URL would make this one click.

## Proposal

- Add `docs/local-models.md` documenting the recipe above (Ollama first; note it applies to any OpenAI-compatible server).
- Add a first-class `ollama` provider alias in `buzz-agent` (`crates/buzz-agent/src/config.rs`): maps to the OpenAI-compatible path, defaults `OPENAI_COMPAT_BASE_URL` to `http://localhost:11434/v1`, makes the API key optional (default `ollama`), and defaults `OPENAI_COMPAT_API` to `chat`. Overrides via the existing env vars keep working.
- Unit tests for the new resolution paths.

I have a PR ready implementing exactly this.

bitcoinaudio:
## Summary

Closes #3145.

Local models already worked through the generic OpenAI-compatible provider, but the path was undocumented and required a dummy API key plus a magic base URL. The word `ollama` appeared nowhere in the repo, leading operators to set model env vars on the **relay** (which never calls an LLM) and conclude local models were unsupported.

- `BUZZ_AGENT_PROVIDER=ollama` maps onto the OpenAI-compatible path with local-friendly defaults: no API key required (a placeholder is sent, since Ollama ignores it), `OPENAI_COMPAT_BASE_URL` defaults to `http://localhost:11434/v1`
- `docs/local-models.md`: where LLM config actually lives (the agent process, not the relay), Ollama quickstart, generic openai-compat recipe (vLLM / llama.cpp / LM Studio), endpoint verification, context-window tuning for small local models, desktop steps
- `crates/buzz-agent/README.md`: ollama run example + env table updates

Minimal Ollama setup after this PR:

```bash
BUZZ_AGENT_PROVIDER=ollama OPENAI_COMPAT_MODEL=qwen2.5:7b-instruct ./buzz-agent
```

## Test plan

- [x] 4 new unit tests: provider alias needs no key, key resolution (placeholder default / whitespace-as-unset / explicit wins / non-ollama still required), base URL defaults + override
- [x] `cargo test -p buzz-agent` — 291 lib + 65 integration tests pass
- [x] `cargo clippy -p buzz-agent` clean, `cargo fmt --check` clean (pre-commit/pre-push hooks green)
- [x] Live smoke test: real `buzz-agent` binary over ACP stdio (`initialize` → `session/new` → `session/prompt`) against a local Ollama (`qwen2.5:7b-instruct`) with **no API key set** — streamed a correct reply, `stopReason: end_turn`

## Out of scope (noted in #3145)

Desktop UI preset ("Ollama (local)" with prefilled base URL, no required key) — separate PR per the agent-config rules in `desktop/src/features/agents/AGENTS.md`.

### DIFF
--- a/crates/buzz-agent/README.md
+++ b/crates/buzz-agent/README.md
@@ -50,6 +50,11 @@ OPENAI_COMPAT_MODEL=gpt-5 \
 OPENAI_COMPAT_BASE_URL=https://api.openai.com/v1 \
   ./target/release/buzz-agent
 
+# Or a local Ollama server (no API key needed)
+BUZZ_AGENT_PROVIDER=ollama \
+OPENAI_COMPAT_MODEL=qwen2.5:7b-instruct \
+  ./target/release/buzz-agent
+
 # Or Databricks model serving via OAuth 2.0 PKCE
 BUZZ_AGENT_PROVIDER=databricks \
 DATABRICKS_HOST=https://dbc-...cloud.databricks.com \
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
nice thin alias — mapping ollama → openai-compat with a baked localhost base url matches how people already run local models today. docs/local-models.md is the part that was missing. for #3874 (ollama cloud) a sibling alias with https://ollama.com/v1 would reuse almost all of this path; worth a follow-up rather than stuffing cloud into this pr.

PR #3155, opened 2026-07-27
fix(mobile): load current agent presence

fitz2882:
## Summary

- fetch current presence when mobile starts tracking agent pubkeys
- batch newly tracked pubkeys into one signed relay query
- keep the existing WebSocket subscription for subsequent live updates
- resolve relay-synthesized snapshot subjects from their `p` tags while keeping live events author-scoped
- prevent an older snapshot response from overwriting a newer live presence update

## Why

Mobile initialized every missing presence entry as offline and only listened for future kind 20001 events. If an agent was already online before mobile subscribed, the UI could show it as offline even while desktop showed it online.

The relay already exposes current ephemeral presence through the generic authenticated `/query` bridge when a kind 20001 filter includes authors. This change uses that existing Nostr bridge instead of adding a new endpoint.

## Testing

- focused Dart formatting and analysis pass
- mobile file-size guard passes
- `flutter test test/features/profile/presence_cache_provider_test.dart` (11 tests)
- regression coverage for current snapshots, batching, snapshot/live ordering, live updates, invalid states, and tagged-subject impersonation

`just ci` was run but current upstream `main` is not green independently of this change:

- desktop file-size guard: `AgentCreationPreview.tsx` is 1025 lines with a 1000-line limit
- mobile analysis: `_FakeReadStateNotifier.markContextRead` has a stale override in `activity_page_test.dart`

This branch does not modify either failing file.

## Manual verification

1. Start an agent before opening mobile.
2. Open a channel or profile that displays the agent.
3. Confirm the existing presence indicator shows online without waiting for a later heartbeat.
4. Stop the agent and confirm the live offline update still applies.

No screenshot is included because this changes the data source for the existing presence indicator, not its visual layout.

### DIFF
--- a/mobile/lib/features/profile/presence_cache_provider.dart
+++ b/mobile/lib/features/profile/presence_cache_provider.dart
@@ -14,6 +14,7 @@ import '../../shared/relay/relay.dart';
 class PresenceCacheNotifier extends Notifier<Map<String, String>> {
   final Set<String> _tracked = {};
   final Set<String> _pendingSnapshotPubkeys = {};
+  final Map<String, int> _liveUpdateVersions = {};
   void Function()? _presenceUnsub;
   Timer? _snapshotBatchTimer;
   int _subscriptionVersion = 0;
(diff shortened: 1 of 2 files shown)


kerpopule:
Confirmed against a real Buzz 0.5.3 managed-agent fleet on macOS with the user observing from mobile. The agents were already running before mobile opened the DM. Mobile initialized them as offline because `PresenceCacheNotifier.track` only registered pubkeys and waited for a future kind:20001 event, while the harness had already published its initial online event and had periodic heartbeats disabled. The snapshot hydration in this PR closes that exact missed-event gap.

PR #3162, opened 2026-07-27
fix(acp): make explicit that final response text is not auto-delivered

holmesadrianh:
Fixes the root cause described in #2698: `base_prompt.md`'s delivery-contract line said *"Your reasoning and tool calls are invisible"* but never explicitly said the model's own final response text is also invisible. Claude Code's own default system prompt tells the model the opposite (final text is normally displayed to the user), so on a low-effort/casual reply the model falls back to that default instead of Buzz's actual publish-is-opt-in model.

This reproduced independently for us on 2026-07-25 (two unrelated agents, same day) and again on 2026-07-27 — confirmed via relay logs that the agent processed the request and queried the relay several times, then never called `buzz messages send` and never published anything.

Minimal fix: reword the existing line to explicitly name "your own final response text" as invisible too, rather than leaving it implied. No behavior/harness change — prompt wording only.


holmesadrianh:
Noticed after opening this that #2703 (commit `5a959a4`) also addresses #2698, and does it more thoroughly than this PR: alongside a similar `base_prompt.md` wording fix, it also patches `queue.rs`'s `append_reply_instruction()`/`append_new_thread_reply_instruction()` to re-inject the "session text is not delivered" reminder at the top of every per-turn prompt, not just once in the base prompt. That's a stronger fix than what I've proposed here — it's more resistant to the instruction getting buried in a long context, which is closer to the actual failure mode we hit.

Happy to have this superseded by (or merged into) #2703's approach if that's preferred — flagging the overlap rather than leaving two PRs solving the same issue for a maintainer to untangle. No objection either way; this one's small enough to stay open as a minimal fallback if useful, but #2703's fix looks like the more complete solution.

One aside for whoever's triaging: #2703 bundles that fix together with two unrelated ones (Claude model discovery, avatar flicker) — if it'd help move the reply-delivery fix along faster, splitting it out might make sense, though that's obviously not my call to make.

PR #3174, opened 2026-07-27
test(desktop): await thread scroll anchor

cameronhotchkies:
## Why
The focus/split E2E test could capture the thread root before its programmatic middle-thread scroll had settled, then incorrectly report a scroll-restoration failure.

## What
- Poll until the requested middle-thread scroll position is applied
- Require the captured anchor to intersect the thread viewport and differ from the root
- Preserve the existing focus-to-split-to-focus viewport assertions

## Risk Assessment
Low — test-only synchronization change with no production behavior changes.

## References
- Original failure: https://github.com/block/buzz/actions/runs/30231271427/job/89870533541
- Buzz thread: buzz://message?channel=12dd513d-45fd-48ff-80ac-8596d2fcc9d3&id=87ce6024b4bf74bfac2fa75d9f7bbbcc8f8fe2df460afe534152c495929f51ba
- Reproduced confidence: 20 consecutive targeted passes, full spec pass, `just desktop-ci`, and `just ci`

Generated with Codex


atishpatel (approved):
Approving — test-only synchronization fix, verified locally.

**What I checked**
- Reviewed the diff: `topVisibleMessageId` → `scrollToMiddleVisibleMessage`, which (a) sets the 40% scroll inside the poll and re-checks `element.scrollTop` landed within 1px in the *same* synchronous evaluate, (b) requires the captured row to intersect the viewport on both edges, (c) rejects the thread root. That is exactly the failure mode in the linked run: the anchor was captured at `scrollTop === 0`, so the root became the anchor and the later restoration assertion legitimately failed on an illegitimate anchor.
- Not masking a product bug: the position check is synchronous with the assignment, so the anchor is genuinely a ~40% row. If restoration were broken (or the app snapped back to bottom asynchronously afterward), the preserved `toBeInViewport()` assertions after focus→split→focus would still fail. The three original viewport/highlight assertions are untouched.
- Rebased the diff cleanly onto `origin/main` (96ae1417) in a scratch worktree — the branch is 159 commits behind but `thread-focus-mode.spec.ts` has no conflicting change on main since `cca88390`.
- `pnpm exec tsc --noEmit`: clean. `pnpm check` (biome + file-size/px-text/pubkey guards): clean, no new findings.
- `playwright test --project=smoke thread-focus-mode.spec.ts --repeat-each=8`: 16/16 passed, no flaky, on the rebased tree.

**Nits, non-blocking**
- `expect.poll` with a mutating callback is slightly unusual (the poll body performs the scroll rather than just observing it). It's correct here — the retry *is* the re-scroll — but a one-line comment saying so would help the next reader.
- The `anchorId` closure write plus the post-poll `if (!anchorId) throw` is redundant with the poll predicate; harmless.

Nothing blocking. Please rebase/let the merge queue bring it up to date before landing.

wesbillman (approved):

PR #3178, opened 2026-07-27
fix(cli): resolve agents from owner records

johnmatthewtennant:
## Context

`buzz users get --name Honey` searches relay-wide profiles and can return an identically named agent owned by someone else. This caused agents from the wrong owner to be added to a channel.

## Summary

This bug fix scopes exact-name agent lookup to owner-authored managed-agent records, then cryptographically verifies each returned profile's NIP-OA `auth` tag before asserting ownership. The relay and database contracts remain unchanged.

### Related issue

None found.

## Changes

- Adds `buzz users get --name Honey --owner me|<hex>|<npub>`.
- Resolves `me` to the NIP-OA owner when the CLI runs as an agent, otherwise to the CLI identity.
- Matches kind `30177` managed-agent record names exactly and case-insensitively under the requested owner.
- Requires exactly one valid NIP-OA `auth` tag whose verified owner equals the requested owner and whose `kind` and `created_at` conditions apply to the profile event before returning `owner_pubkey` or `owned_by_me: true`.
- Keeps missing, malformed, stale, condition-mismatched, or unverifiable owner-record candidates visible with `owned_by_me: false` and an explicit `verification` value.
- Returns every same-name record for the owner so callers can require explicit selection when duplicates remain.
- Preserves the existing output shape and client-side name filter for unscoped searches.
- Documents the distinct owner-scoped managed-agent lookup and unscoped NIP-50 lookup modes.

### Testing

The reviewer-reproducible red and green commands below exercise the ownership bug against the target branch and this branch.

## Screenshots

Not applicable. This is a CLI-only change.

The lookups below were run against the live relay from `main` and this branch.

### Red: unscoped lookup returns the 100-profile relay-wide cap and excludes John's agents

On `main`:

```bash
cargo run -q -p buzz-cli -- users get --name Honey \
  | jq '{count: length, first_three: .[:3] | map(.pubkey), johns_agents: map(select(.pubkey == "31b29bcbe69d6716fbb7ba33602b89200bfc9ddfdabcfd1ea6fbfa70b816dfc7" or .pubkey == "4597ac725bba33fc7dd0454c1e2316a5ed770426acf667837d46f6553b3fcf54"))}'
```

Observed output:

```json
{
  "count": 100,
  "first_three": [
    "20d27fc6c0ab4f50b66d1a32a64c5ca1fb985254143ce911f61ab7733333c3d7",
    "00644478cdd9032c563ddc712b3687d8345d948945aab3c18bab95afbf6f519a",
    "93c16697d0e58007bc11fb953208bc6b1cff387b2dee094abc10bf82dfee5424"
  ],
  "johns_agents": []
}
```

`main` also rejects the owner-scoped command:

```bash
cargo run -q -p buzz-cli -- users get --name Honey --owner me
```

```text
error: unexpected argument '--owner' found
Usage: buzz users get --name <NAME>
```

### Green: owner-scoped lookup distinguishes verified and unresolved records

On this branch:

```bash
cargo run -q -p buzz-cli -- users get --name Honey --owner me \
  | jq 'map({pubkey,display_name,owner_pubkey,owned_by_me,verification})'
```

Observed output:

```json
[
  {
    "pubkey": "0ca77314d7ac8b3fcf6c647cc8cb9c3afd840db3b2a8ff2079f09a168de1827e",
    "display_name": null,
    "owner_pubkey": null,
    "owned_by_me": false,
    "verification": "missing_profile"
  },
  {
    "pubkey": "31b29bcbe69d6716fbb7ba33602b89200bfc9ddfdabcfd1ea6fbfa70b816dfc7",
    "display_name": "Honey",
    "owner_pubkey": "67252b09c31a995daa63aada26569fbc6a3d12f573113f001ce7432f870da820",
    "owned_by_me": true,
    "verification": "verified"
  },
  {
    "pubkey": "4597ac725bba33fc7dd0454c1e2316a5ed770426acf667837d46f6553b3fcf54",
    "display_name": "Honey",
    "owner_pubkey": "67252b09c31a995daa63aada26569fbc6a3d12f573113f001ce7432f870da820",
    "owned_by_me": true,
    "verification": "verified"
  }
]
```

Only the two profiles with valid NIP-OA proofs assert ownership. The owner-authored record whose profile is absent remains visible but cannot be selected as verified ownership.

### DIFF
--- a/crates/buzz-cli/src/commands/users.rs
+++ b/crates/buzz-cli/src/commands/users.rs
@@ -221,12 +221,20 @@ async fn search_by_name(
         crate::OutputFormat::Compact => {
             let compact: Vec<serde_json::Value> = profiles
                 .iter()
-                .map(|p| serde_json::json!({
-                    "pubkey": p.get("pubkey").cloned().unwrap_or_default(),
-                    "display_name": p.get("display_name").or_else(|| p.get("name")).cloned().unwrap_or_default(),
-                    "owner_pubkey": p.get("owner_pubkey").cloned().unwrap_or_default(),
-                    "owned_by_me": p.get("owned_by_me").cloned().unwrap_or_default(),
-                }))
+                .map(|p| {
+                    let mut value = serde_json::json!({
+                        "pubkey": p.get("pubkey").cloned().unwrap_or_default(),
+                        "display_name": p.get("display_name").or_else(|| p.get("name")).cloned().unwrap_or_default(),
+                    });
+                    if let Some(obj) = value.as_object_mut() {
+                        for field in ["owner_pubkey", "owned_by_me"] {
+                            if let Some(field_value) = p.get(field) {
+                                obj.insert(field.to_string(), field_value.clone());
+                            }
+                        }
+                    }
+                    value
+                })
                 .collect();
             serde_json::to_string(&compact).unwrap_or_default()
         }
--- a/crates/buzz-cli/src/commands/users.rs
+++ b/crates/buzz-cli/src/commands/users.rs
@@ -1,3 +1,6 @@
+use buzz_core::kind::KIND_MANAGED_AGENT;
+use nostr::PublicKey;
+
 use crate::client::{normalize_write_response, BuzzClient};
 use crate::error::CliError;
 use crate::validate::validate_hex64;
@@ -13,6 +16,7 @@ pub async fn cmd_get_users(
     client: &BuzzClient,
     pubkeys: &[String],
     name: Option<&str>,
+    owner: Option<&str>,
     format: &crate::OutputFormat,
 ) -> Result<(), CliError> {
     if let Some(query) = name {
@@ -21,7 +25,11 @@ pub async fn cmd_get_users(
                 "--name and --pubkey are mutually exclusive".into(),
             ));
         }
-        return search_by_name(client, query, format).await;
+        return search_by_name(client, query, owner, format).await;
+    }
+
+    if owner.is_some() {
+        return Err(CliError::Usage("--owner requires --name".into()));
     }
 
     for pk in pubkeys {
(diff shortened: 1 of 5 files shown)


crates/buzz-cli/src/commands/users.rs
```diff
@@ -126,6 +206,13 @@ async fn search_by_name(
                     "pubkey".to_string(),
                     serde_json::json!(event.get("pubkey").and_then(|v| v.as_str()).unwrap_or("")),
                 );
+                if let Some(owner) = &result_owner {
```
wpfleger96: [IMPORTANT][Correctness] The filter immediately above this block re-applies the kind `0` profile-name check to the owner-scoped path after `owned_agent_pubkeys_by_name()` has already matched the authoritative kind `30177` name. An owned agent can legitimately have a missing, stale, or different kind `0` name, so `users get --name Honey --owner ...` can resolve the agent pubkey and then silently return `[]`.

Could we skip that filter when `result_owner.is_some()` and retain it only for unscoped NIP-50 results? Please also add coverage where the kind `30177` name matches but the fetched kind `0` profile name differs.

crates/buzz-cli/src/lib.rs
```diff
@@ -808,6 +808,9 @@ pub enum UsersCmd {
         /// Search by display name (case-insensitive substring match)
         #[arg(long = "name")]
         name: Option<String>,
+        /// Scope an exact-name agent lookup to its owner (`me`, hex, or npub)
+        #[arg(long = "owner", requires = "name")]
```
wpfleger96, line 812: [MINOR][Minimalism] `requires = "name"` makes sense here because `--owner` is only implemented as a scope for name-based agent resolution; it is not meaningful for `--pubkey` or the implicit self lookup. However, this also means the runtime `owner.is_some()` check in `commands/users.rs:31-32` is unreachable through normal Clap dispatch. I think we should retain this parse-time constraint and remove the duplicate runtime guard unless direct calls to `cmd_get_users()` are an intended supported path. Non-blocking.

wpfleger96 (changes requested):
I found one correctness issue in the owner-scoped lookup. The duplicated CLI/runtime validation is non-blocking.

crates/buzz-cli/src/lib.rs
johnmatthewtennant: I’m retaining the runtime guard so `cmd_get_users` enforces its own argument invariant independently of Clap. The function is public within the CLI command module and accepts `owner` directly, so keeping the inexpensive check prevents future direct callers or dispatch refactors from silently ignoring `--owner` when no name is supplied. I agree Clap is the normal path, but prefer defense-in-depth at this command boundary. Leaving this non-blocking thread unresolved because I’m declining the suggested removal rather than implementing it. — AI-generated by Fizz

wpfleger96 (changes requested):
Reviewed current remote head `906072bd980f77ceb311fdd7ff635d7b7e7962b9`. The prior trust-boundary and disappearing-candidate findings are addressed, but one protocol-level trust defect remains.

## Review

### [CRITICAL][Correctness]: Condition-restricted credentials are promoted to profile ownership proof

[`owner_verification()`](https://github.com/block/buzz/blob/906072bd980f77ceb311fdd7ff635d7b7e7962b9/crates/buzz-cli/src/commands/users.rs#L196-L219) calls `verify_auth_tag()`, which validates the signature and condition syntax but never applies the signed conditions to the profile event. NIP-OA requires verifiers to evaluate every clause and defines `kind=<n>` and `created_at` bounds against the event being verified ([spec lines 64–69](https://github.com/block/buzz/blob/906072bd980f77ceb311fdd7ff635d7b7e7962b9/docs/nips/NIP-OA.md#L64-L69)). The new test makes the mismatch concrete: it signs `kind=9` and expects that credential to verify on a kind:0 profile ([test lines 559–584](https://github.com/block/buzz/blob/906072bd980f77ceb311fdd7ff635d7b7e7962b9/crates/buzz-cli/src/commands/users.rs#L559-L584)).

That promotes an attestation the owner restricted to kind:9 messages into trusted `owner_pubkey` / `owned_by_me` metadata for a kind:0 profile. Because this command exists to prevent wrong-owner agent selection, widening a credential beyond its signed scope is blocking.

After the signature check, evaluate the exact conditions string against the profile event: every `kind=` clause must equal `0`, and every `created_at<` / `created_at>` clause must accept the profile's `created_at`. Return a non-verified status when any clause fails, and add negative tests for `kind=9` and violated time bounds. If this lookup is intentionally defining a separate standing-identity proof context like NIP-IA rather than event-level NIP-OA verification, that semantic expansion needs an explicit protocol contract and test name; the current implementation and PR description claim ordinary NIP-OA verification.

### [MINOR][Correctness]: The reported Unit Tests check does not run these tests

The added `buzz-cli` tests pass locally, but [`just test-unit`](https://github.com/block/buzz/blob/906072bd980f77ceb311fdd7ff635d7b7e7962b9/Justfile#L275-L296) only selects `buzz-core`, `buzz-auth`, `buzz-db`, `buzz-conformance`, and `buzz-push-gateway`. Therefore the green GitHub `Unit Tests` job does not exercise this PR's new tests. Non-blocking for this fix because I independently ran the full package suite, but `buzz-cli` should be added to that CI gate or get a changed-package job.

## What's Solid

- The original trust-boundary flaw is substantially fixed: ownership fields now require exactly one cryptographically valid auth tag whose owner matches the requested owner.
- Renamed profiles, missing profiles, malformed agent keys, missing/invalid/multiple auth tags, and owner mismatches remain visible without claiming ownership.
- The owner-scoped path no longer re-applies the lossy profile-name filter, so the prior disappearing-candidate defect is fixed.
- Reusing `extract_d_tag`, filtering invalid author keys before the profile query, preserving unscoped compact output, and returning all duplicate matches are all correct.
- All GitHub checks are green. I independently ran `cargo test -p buzz-cli` at exact head `906072bd980f77ceb311fdd7ff635d7b7e7962b9`: 254 passed, 0 failed; the SHA was unchanged before and after the run.

## Alternatives

The smallest correction is local condition evaluation after `verify_auth_tag()`. A more reusable design is an event-aware SDK verifier that accepts `(kind, created_at)` while retaining the existing signature-only primitive for explicitly documented NIP-AA/NIP-IA contexts.

## Quality Gate

- **Minimalism: 9/10** — the CLI-only two-query flow and explicit verification statuses are focused and load-bearing.
- **Elegance: 9/10** — the owner resolution, candidate preservation, and output shaping are direct and readable.
- **Correctness: 7/10** — signature verification is now present, but signed authorization conditions are ignored before trusted ownership metadata is emitted.

**Verdict: Request changes.** Fix the condition applicability gap, then this is ready for another pass.

johnmatthewtennant:
Addressed the re-review findings in `a7c177ae`.

- NIP-OA ownership now requires every signed `kind` and `created_at` condition to apply to the actual kind:0 profile event. Non-applicable credentials return `condition_mismatch` and do not emit ownership metadata.
- Added negative coverage for `kind=9` and exclusive lower/upper timestamp bounds, plus a positive combined-condition case.
- Added `buzz-cli` to both `just test-unit` paths so CI exercises these tests.

Validation at `a7c177ae`: `cargo test -p buzz-cli` (255 passed), `cargo check -p buzz-cli --all-targets`, `cargo clippy -p buzz-cli --all-targets -- -D warnings`, `just test-unit`, formatting, shell syntax, and pre-push Rust/Desktop Tauri suites passed. — AI-generated by Fizz

wpfleger96 (approved):
Re-reviewed exact remote head `4275a48d1b65cdbb590a90541bad843cd667d995`. The prior blocking condition-applicability defect and the CI coverage comment are addressed.

## Review

No blocking findings.

## What's Solid

- `owner_verification()` now verifies the signature first, then evaluates every signed `kind`, `created_at<`, and `created_at>` condition against the actual kind:0 profile event before emitting ownership metadata.
- Negative tests cover `kind=9` and both exclusive timestamp boundaries; the combined valid case verifies.
- Drifted, missing, malformed, and condition-mismatched candidates remain visible without `owner_pubkey` or `owned_by_me: true`.
- `buzz-cli` is included in both `just test-unit` execution paths. The green CI Unit Tests log ran 258 CLI tests, including this path.
- Independent exact-SHA validation passed: `just test-unit` (including `buzz-cli`: 255 passed, 0 failed) and `cargo check -p buzz-cli --all-targets`. HEAD remained unchanged and the worktree stayed clean.

## Quality Gate

- **Minimalism: 9/10** — the condition check is a focused local addition; the retained command-boundary guard is a reasonable defense-in-depth choice.
- **Elegance: 9/10** — signature, owner, condition, and candidate-state decisions remain explicit and readable.
- **Correctness: 9/10** — the trust boundary now applies every supported NIP-OA condition with strict timestamp semantics, and the failure paths do not assert ownership.

**Verdict: Approve.**

PR #3182, opened 2026-07-27
fix(desktop): recover full local storage on startup

wesbillman:
## Summary
- recover already-full Buzz installs before desktop initialization without deleting healthy caches
- enforce a global 2 MiB UTF-16 byte budget across disposable message, channel, timeline-skeleton, and sidebar-skeleton caches, regardless of relay count
- route all disposable cache writes through quota recovery and reserve roughly 3 MiB of WebKit's observed ~5 MiB quota for durable state
- preserve communities, identities, preferences, drafts, and read state; match only delimiter-qualified disposable namespaces

## Context
WebKit enforces an approximately 5 MiB per-origin localStorage quota and Tauri does not expose an app-level knob to raise it to 50 MiB. Buzz 0.4.26 shipped reactive recovery for selected durable writes, but disposable writers swallowed quota failures and their existing limits were count-based per relay rather than byte-based per origin.

This PR handles both halves: upgrade recovery for already-wedged origins and proactive global headroom so disposable snapshots cannot drive the origin back to the cliff.

## Safety
- startup first probes a one-byte marker; healthy installs retain their caches
- only if the marker write fails are the four relay-rehydratable cache namespaces removed
- namespace matching requires the `v1:` delimiter, preventing future `v10` or similarly named durable keys from matching
- oversized individual snapshots are rejected; crossing the global budget evicts disposable snapshots only
- failed recovery leaves the marker absent, so the next launch retries

## Verification
- `pnpm test` — 3,670 passed
- `pnpm check`
- `pnpm typecheck`
- push hooks: branch-skew, desktop-check, desktop-test passed
- byte-budget tests cover UTF-16 accounting, multiple relays, oversized writes, durable-state preservation, healthy startup, full startup, marker retry, and namespace near misses

### DIFF
--- a/desktop/src/features/channels/channelSnapshot.test.mjs
+++ b/desktop/src/features/channels/channelSnapshot.test.mjs
@@ -87,6 +87,36 @@ test("remove clears the snapshot for that relay", () => {
   assert.equal(readChannelSnapshot(RELAY), null);
 });
 
+test("cache write evicts disposable entries and retries at quota", () => {
+  const original = window.localStorage;
+  const storage = new Map([
+    ["buzz-channel-messages.v1:relay:old", "big"],
+    ["buzz-timeline-skeleton-shape.v1:old", "small"],
+  ]);
+  window.localStorage = {
+    get length() {
+      return storage.size;
+    },
+    key: (index) => [...storage.keys()][index] ?? null,
+    getItem: (key) => storage.get(key) ?? null,
+    setItem(key, value) {
+      if (!storage.has(key) && storage.size >= 2) {
+        throw new Error("quota exceeded");
+      }
+      storage.set(key, value);
+    },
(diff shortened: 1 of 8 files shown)


AaronGoldsmith:
🤖 One additional cache namespace worth including in this recovery/budget: `buzz-thread-activity.v1:`.

In the affected real-world localStorage database, thread activity accounted for ~968 KB. It is relay-rehydratable (`useUnreadChannels` loads the snapshot, then catches up from the relay and rebuilds it), so it appears to fit the same disposable-cache category as messages/channels/skeletons.

Could we add that prefix to `PURE_CACHE_KEY_PREFIXES` and route `threadActivityStorage.ts` writes through `setLocalStorageItemWithRecovery`? Otherwise the nominal 2 MiB cache budget can still have roughly another 1 MiB sitting outside it.

PR #3194, opened 2026-07-27
feat(desktop): add pitch-preserving TTS playback speed

johnmatthewtennant:
## Context

Generated speech currently plays at a fixed 1x rate. Rodio's speed control shifts pitch, while Pocket synthesis rate changes model generation rather than playback. Playback speed needs to remain independent from Pocket voice selection and preserve the existing per-chunk streaming path.

## Summary

This adds a locally persisted generated-speech playback setting with pitch-preserving Signalsmith time stretching. The slider covers 0.5x to 2x, the editable value accepts 0.25x to 4x, and Reset restores 1x. The control lives in Voice settings but remains independent from Pocket synthesis and voice selection.

## Changes

- Processes each complete Pocket playback chunk with Signalsmith Stretch before the chunk is appended to rodio.
- Keeps the fixed device lead-in outside time stretching and retains the existing post-DSP cancellation check.
- Preserves streaming time-to-first-playback. The first synthesized chunk plays without waiting for the full response.
- Uses a bit-exact 1x bypass, persists the global speed locally across restarts, and supports precise typed values beyond the slider range.
- Serializes native persistence writes and coalesces optimistic UI updates without allowing stale saves to overwrite newer intent.
- Preserves the parent Voice Settings label, accessibility behavior, and live-only speech routing.
- Adds focused coverage for pitch, ordering, onset, latency, persistence, cancellation, and the production chunk boundary.
- Adds `ssstretch` 0.1.0 and the Signalsmith DSP implementation under the MIT license. Both are bundled at build time; normal use performs no download.

## Related issue

None found.

## Testing

Manual validation in the combined daily-driver build covered the 0.5x to 2x slider, typed values across the 0.25x to 4x accepted range, stable voice pitch, immediate cancellation, Reset to 1x, and persistence after restart. The production DSP probe measured 60 ms of compensated lookahead and processing below 0.21% of realtime at the sampled speeds.

## Screenshots

The Voice settings control combines the bounded slider, editable value, and Reset action.

Speech playback speed

1. Launch Buzz Desktop and open **Settings > Voice**.
2. Set **Speech playback speed** to 0.5x and play a Pocket voice preview or start a huddle with generated speech. Confirm speech is slower without a lower voice pitch.
3. Set the control to 2x and confirm the next generated speech is faster without a higher voice pitch.
4. Type a precise value outside the slider range, such as 0.25x or 4x, and confirm it is accepted and applied.
5. Select **Reset** and confirm the value returns to 1x.
6. Set a non-default value, quit and reopen Buzz, and confirm the value persists.
7. During active generated speech, stop or interrupt playback and confirm cancellation remains immediate.

For a repeatable DSP measurement, run:

```bash
CARGO_HOME=$PWD/.hermit/rust cargo run --offline --release \
  --manifest-path desktop/src-tauri/Cargo.toml \
  --example tts_speed_probe
```

Observed output:

```text
0.75x: processing=20.59ms, 0.206% realtime, output=320000 samples, compensated DSP lookahead=60.0ms
1.25x: processing=12.42ms, 0.124% realtime, output=192000 samples, compensated DSP lookahead=60.0ms
1.50x: processing=10.18ms, 0.102% realtime, output=160000 samples, compensated DSP lookahead=60.0ms
```

Pocket still submits the first completed synthesis chunk immediately rather than rendering the full response.


johnmatthewtennant:
🤖 New review-intent sweep completed on the frozen stacked candidate.

- `timeless-comments`: 0 findings
- `stacksmith`: 0 findings
- `dead-code`: 1 valid P3, fixed by narrowing playback-speed DSP imports and removing the unused re-export suppression
- selective `dead-code` rerun on the patched candidate: 0 findings

Published DCO-signed head: `a5139463d1ca15eb9ee8292568429fae2e610e34` (base remains `7e5cc4b60796a4ceb5dfa10629587417dbbff85e`). Focused patched-content validation passed: Rust formatting check, 8 playback-speed tests, warning-denied Tauri clippy, and `git diff --check`. GitHub CI finished green with 18 passing and 7 skipped checks. PR remains open and draft; no readiness or merge action was taken.

Manual validation still outstanding: listen to Mary and Marius preview plus active-huddle speech at 0.75x, 1.25x, and 1.50x; verify pitch, first phoneme, tail, cancellation, persistence after restart, and Reset to 1.00x.

johnmatthewtennant:
🤖 ### Speech playback speed

The slider covers 0.5x to 2x, the editable value accepts 0.25x to 4x, and Reset restores 1x.

pocket-voices

desktop/src-tauri/src/huddle/tts.rs
```diff
@@ -754,6 +758,17 @@ fn tts_worker(
                 }
                 match synthesis {
                     Ok(samples) if !samples.is_empty() => {
+                        let speed = playback_speed.get();
+                        let samples = match process_complete_chunk(&samples, speed, SAMPLE_RATE) {
```
wesbillman: **Blocking:** this resets, drains, flushes, and latency-crops a fresh Signalsmith stretcher for every Pocket *model unit*. Those units are hidden subdivisions of one playback chunk; `PlaybackChunkAudio` deliberately joins them without internal fades or pauses to preserve continuity. Independent phase-vocoder runs can therefore create a click or timbre/prosody seam exactly where the April pipeline avoids exposing a boundary.

Please keep one streaming stretcher across all model units in a playback chunk, or buffer and stretch the complete playback chunk. Add a regression with two adjacent units that verifies waveform/spectral continuity; the current DSP tests process one buffer, and `tts_playback_speed_tests.rs` checks only the fixed lead-in.

— Carl, reviewing on Wes's behalf

desktop/src/features/settings/ui/playbackSpeedPersistence.ts
```diff
@@ -0,0 +1,91 @@
...
+  state.desiredSpeed.current = nextSpeed;
+  callbacks.setSpeed(nextSpeed);
+  callbacks.setError(null);
+  ensurePlaybackSpeedFlush(state, callbacks);
+}
+
+export function cancelPlaybackSpeedPersistence(
+  state: PlaybackSpeedPersistenceState,
+) {
+  state.active.current = false;
```
wesbillman: **Blocking persistence race:** unmount sets `active=false`, so `flushPlaybackSpeed` stops after the already-dispatched save and silently drops any newer queued intent. Example: 1.25× is in flight, the user selects 1.5×, then closes Settings; native persistence completes at 1.25× and 1.5× is never sent. The existing unmount test codifies that loss by asserting the old instance makes only the first persistence call.

An in-flight native write also cannot be cancelled: a remounted panel may read the old atomic before that write finishes and remain stale because there is no revision/subscription. Move latest-intent/coalescing ownership outside the component (or give native persistence a monotonic revision/latest-intent contract), and test that unmount/remount preserves the user's final selection.

— Carl, reviewing on Wes's behalf

wesbillman (changes requested):
The pitch-preserving speed control is useful, and the branch is now cleanly shaped against main (19 files, mergeable). Two correctness issues still need resolution before approval: the stretcher resets at hidden Pocket model-unit boundaries, and component-local persistence can discard the user's latest speed when Settings unmounts. Details are inline.

Cancellation is rechecked under the player-operation lock before append, so cancelled processed audio is rejected. The accessibility wiring also looks sound. The new target-wide MSVC CXXFLAGS affect every C++ dependency, so the current Windows CI result is important evidence for the final candidate; it was still running when I submitted this review.

Reviewed at exact head `9ff70fff53f4c9c21d061fbf6ce92838560dd89b`.

— Carl, reviewing on Wes's behalf

desktop/src-tauri/src/huddle/tts.rs
johnmatthewtennant: 🤖 Fixed in `ac342ddd8bc59bbfe3c4225ce4bdf3efa0921d52`: model units are now joined before a single Signalsmith pass, so hidden Pocket boundaries share one stretcher timeline. Added a two-unit continuity regression that verifies exact equivalence to already-joined processing and bounds the boundary sample discontinuity.

desktop/src/features/settings/ui/playbackSpeedPersistence.ts
johnmatthewtennant: 🤖 Fixed in `ac342ddd8bc59bbfe3c4225ce4bdf3efa0921d52`: latest-intent coalescing now lives in a module-lifetime controller, so persistence continues after Settings unmount and a remounted panel immediately observes the final value/error. The regression now proves the queued 1.5x intent persists after unmount and is visible after remount.

Issue #3160, opened 2026-07-27
feat: Slack-parity per-channel notification settings — levels, temporary mute, mute-and-hide, broadcast opt-out


LordMelkor:
> 🤖 Filed by kray's AI agent on his behalf.

## Problem

The only per-channel notification control in Buzz is a binary mute (#838). That single bit can't express the states people actually live in:

- **"This channel is busy but I can't leave it."** A high-traffic channel bolds, dots, and dock-bounces on every message. The only relief is full mute — which also kills the signal you *do* want. There is no "just tell me when I'm mentioned."
- **"Quiet this for the afternoon."** A launch thread is blowing up, you're heads-down for an hour. Muting works, but nothing reminds you to unmute — the channel stays silent forever, and Buzz has no temporary mute at all.
- **"I'm done with this channel but can't leave it."** Muted channels still occupy sidebar rows (dimmed, but present). There's no way to hide a channel and trust it to resurface when someone actually mentions you.
- **"@channel is too much for *this* channel."** #3146 adds `@channel`/`@here`. Slack learned that channel-wide mentions need a per-channel escape valve ("Get broadcast messages"); without one, the only defense against an @channel-happy channel is full mute.

Slack users reach for all four by muscle memory: a three-level notification setting, "Mute until…", "Mute and hide", and per-channel broadcast opt-out. Buzz has none of them — and issue #100 (the original notification-system issue) closed without any follow-up for per-channel granularity, so this is greenfield in the tracker.

## Proposal

Slack-parity per-channel notification settings, desktop first:

- **Notification level** per channel — **All new posts** (default; today's exact behavior) / **Just mentions** (channel still shows unread, but no dock bounce or alerts for ordinary posts) / **Mute and hide** (today's mute semantics, plus the channel leaves the sidebar — resurfacing while active or when it holds an unread mention, Slack-style; still reachable from the channel browser).
- **Temporary mute** — "For 1 hour" / "Until tomorrow" (9am), stored as an absolute timestamp, evaluated lazily so it syncs correctly across devices and survives restarts. Expiry restores the previous level automatically. Reuses the reminders time-preset machinery (#1093).
[...]

Issue #2792, opened 2026-07-24
Mobile: user groups — mention compose, group management, and pill rendering


LordMelkor:
Follow-up to #2728 (backend/CLI/desktop landed in #2791).

The Flutter app needs user-group support to reach parity with desktop:

- **Render**: messages carrying a `["group", <id>, <handle>]` marker tag should render `@handle` as a single mention pill (desktop reference: `remarkGroupMentions.ts`). Members are already notified via compose-time `p`-tag expansion, so notification behavior works today — this is display-only.
- **Compose**: group suggestions in the mention autocomplete + send-time expansion to `p` tags for (group members ∩ channel members) with the marker tag (desktop reference: `groupMentionExpansion.ts` / `useMentionSendFlow.ts`).
- **Manage** (optional, could be a second phase): list groups (kind `39100` snapshots, `d` = group id) and create/edit via the `47000`-`47004` command kinds.

Kind constants must be added to `mobile/lib/shared/relay/nostr_models.dart` in sync with `buzz-core/src/kind.rs` and `desktop/src/shared/constants/kinds.ts` (47000-47004 commands, 39100 snapshot).

Protocol details: see the design summary in #2791.

🤖 Filed by kray's AI agent as part of the user-groups feature work.

PR #2791, opened 2026-07-24
feat: user groups — mentionable member sets across relay, CLI, and desktop

LordMelkor:
# User groups: mentionable member sets across relay, CLI, and desktop

Closes #2728.

Slack-parity **user groups**: named, community-shared sets of members — humans and agents alike — that can be @-mentioned in any channel and bulk-added to channels. A group carries **default channels** that new members automatically join.

```bash
buzz groups create --handle ios-team --name "iOS Team" --default-channel <uuid>
buzz groups add-members <group> --member <pubkey>     # → auto-joins default channels
buzz messages send --channel <uuid> --content "hey @ios-team, release is cut"
buzz channels add-group <channel> ios-team            # bulk-add, dedupes existing members
```

## Protocol design

Groups follow the established NIP-29-style pattern already used for channel membership (9000-series commands → 39002 snapshot): **signed command events + relay side effects + relay-published state snapshot**. Single-author NIP-33 objects were rejected because replacement is keyed by `(pubkey, kind, d)` — incompatible with creator+admin editing.

| Kind | Type | Purpose |
|---|---|---|
| `47000` | command | Create group (handle, name, description, members, default channels) |
| `47001` | command | Edit metadata / default channels |
| `47002` | command | Delete (soft; snapshot tombstoned) |
| `47003` | command | Add member(s) — triggers default-channel auto-join |
| `47004` | command | Remove member(s) — permits removing departed community members |
| `39100` | relay-published, param-replaceable | Group state snapshot, `d` = group id |

Key decisions:
- **Mentions expand at compose time** to real `p` tags for (group members ∩ channel members), so every existing notification path (live subs, mention feed, mobile push, agent wake-up) works unchanged. A `["group", <id>, <handle>]` marker tag drives single-pill rendering.
- **Default-channel auto-join is `47003`-only, by design.** Creating a group with initial members does not auto-join them — membership → channel-access flows through exactly one path. Create a group (zero+ members) with default channels, then add members as they onboard.
- Auto-join is public-channels-only, respects each target's `channel_add_policy`, and skips-doesn't-fail on channels that became invalid.
- Permissions: any community member creates; creator + community admins edit/delete/manage members.
- Handles: `^[a-z0-9][a-z0-9_-]{1,31}$`, unique among active groups per community.
- Caps: 256 member pubkeys per command, 32 default channels; command storage and group mutation commit atomically (reaction-transaction pattern); snapshots carry monotonic per-group versions.

## Surfaces

[...]

Issue #2728, opened 2026-07-24
feat: user groups — mentionable groups of people and agents


LordMelkor:
> 🤖 Filed by kray's AI agent on his behalf.

## Problem

Buzz has no way to address a set of members as one unit. Two Slack-parity gaps:

1. **Group mentions** — in Slack you define a user group (e.g. `@ios-team`); mentioning its handle notifies the members. In Buzz you must @-mention each person individually.
2. **Bulk channel membership** — putting a whole team into a channel means looking up and adding members one by one.

Buzz's existing **agent teams** (kind `30176`) don't cover this: they are containers of *persona IDs* (agent templates deployed into channels) with no pubkey membership and no humans. And because Buzz members and agents are both just pubkeys, a "user group" here can naturally contain **both people and agents** — something Slack can't do.

The kind registry already reserves an empty `47000–47999` "User groups" range (`buzz-core/src/kind.rs`), so this fills in that slot.

## Proposal

Community-shared, named, mentionable groups, following the established NIP-29-style pattern (signed command events + relay side effects + relay-published addressable state snapshot — same architecture as channel membership 9000→39002). NIP-51 follow sets (kind 30000) were considered but are single-author personal lists; groups need creator+admin multi-editor semantics.

### Proposed kinds

| Kind | Purpose |
|---|---|
| `47000` | create group (`handle`, `name`, `description`, initial `p` members, default `channel`s) |
| `47001` | edit metadata / default channels |
| `47002` | delete group |
| `47003` | add member(s) |
| `47004` | remove member(s) |
| `39100` | relay-published group state snapshot (parameterized replaceable, `d` = group id) |

### Semantics

- **Mentions expand to real `p` tags at compose time** (group members ∩ channel members, Slack parity: members not in the channel are silently skipped). Every existing notification path — `#p` live subs, mention feed, mobile, agent wake-up — works unchanged. A `["group", <id>, <handle>]` marker tag lets clients render `@handle` as a mention pill.
- **Bulk add to channel**: resolve group members → existing per-pubkey kind-9000 put-user loop (same as desktop's multi-select invite).
- **Default channels** (Slack parity): when a member is added to a group, the relay auto-adds them to the group's default channels (public channels only for MVP).
- **Permissions**: any member can create a group; only the creator and community admins can edit/delete/manage members. Handles are unique, validated `^[a-z0-9][a-z0-9_-]{1,31}$`.

### Surfaces (this iteration)

[...]

PR #3197, opened 2026-07-27
feat: @channel and @here channel-wide mentions (NIP-CM)

### LINKED ISSUE #3146: feat: channel-wide mentions — @channel and @here
> 🤖 Filed by kray's AI agent on his behalf.

## Problem

Getting the attention of *everyone in a channel* means typing every member's name — and keeping that habit current as membership changes. The real cost isn't the typing, it's the **silent under-notification**: as [a community member described on #2728](https://github.com/block/buzz/issues/2728#issuecomment-5078984489), the moment a fourth agent joined their three-agent channel, "every saved snippet and habit is silently stale — he gets three answers and never notices the fourth was left out."

User groups (#2728) solve the *named-set* case, but someone has to create and maintain the group. Quoting the same comment: a zero-config `@channel`-style mention "is the 80% case and has no membership to keep in sync." #2332 asks for the same primitive from the agent side — "will also be good to have an 'everyone' tag or similar."

And sometimes you don't want everyone — you want *whoever is around right now*. Asking a 40-person channel "can someone restart the deploy?" at 2am shouldn't badge 35 sleeping people.

Slack users reach for `@channel` and `@here` by muscle memory. Buzz has no equivalent — today a literal `@channel` renders as plain text and notifies nobody.

## Proposal

Two reserved, zero-config mention tokens with Slack-parity semantics:

- **`@channel`** — notify every member of the channel.
- **`@here`** — notify only members who are online right now.

### Wire format: one marker tag, not roster fan-out

The message carries a single `["notify", "channel" | "here"]` marker tag (exact tag name finalized in the PR). It does **not** expand into per-member `p` tags, deliberately departing from the compose-time-expansion idiom used by team mentions (#1918), category mentions (#2531), and user groups (#2791):

- **`@here` is not expressible as a compose-time list.** "Who is online" is a delivery-time fact; a frozen send-time snapshot is stale against the presence TTL and leaks exactly who was idle to the sender.
- **Roster fan-out breaks at scale.** N `p` tags bloat the signed event, leak the full roster, hit the 50-mention cap, and are silently suppressed by NIP-PL's `suppress.p_tags_max` hellthread gate — the push spec's own example value is 20, so `@channel` push would silently vanish in any channel larger than that.
- **Agent amplification.** Every `p` tag wakes an ACP-managed agent (`event_mentions_agent`), so an expanded `@channel` in an agent-heavy channel burns one model turn per agent, involuntarily.
- **Precedent exists on both sides.** Buzz already ships a p-tag-free channel-wide attention signal — `["broadcast", "1"]` (NIP-CW) — that the notification pipeline understands today. Matrix faced this exact fork and chose the structured flag (`m.mentions: {"room": true}`, MSC3952) over per-user fan-out.

Foreign Nostr clients degrade gracefully (NIP-27 style): they render literal `@channel` text and simply don't notify. No upstream NIP defines group-wide mentions (verified against nostr-protocol/nips and the NIP-29 client ecosystem), so this will be specced as a Buzz-local NIP in `docs/nips/` (working name **NIP-CM, Channel Mentions**).

### Notification semantics

- `@channel` escalates to **mention-tier** (numeric badge, mention feed, OS notification) for every channel member.
- `@here` escalates **live-only**: clients that receive the event while online treat it as a mention; members who were offline see a normal message later — no retroactive badge (Slack parity).
- **Channel mutes suppress `@channel`/`@here`** — unlike a direct `@name` mention, which continues to pierce mutes (Slack parity; same call #2138 made).
- `@channel` messages appear in members' Home mention feed via a relay-side lookup — **one row per event**, joined against membership at query time, not one `event_mentions` row per member.
- Editing a message to add `@channel` does not re-notify.

### Composer

Autocomplete rows for `@channel` ("Notify all N members") and `@here` ("Notify N online members"), plus a confirmation prompt before sending — reusing the existing send-flow interception point (same seam as the non-member mention dialog).

### CLI and automation safety

`buzz messages send` gets an explicit `--notify channel|here` flag. A literal `@channel` typed in content **without** the flag warns and sends without the tag — agents and workflows must opt in explicitly to mass-notify, so a template variable can never accidentally ping 400 people. (Same spirit as the unresolved-mention warning in #3047.)

### Agents: humans-only, by design

`@channel` and `@here` notify **humans only**. The marker tag carries no `p` tags, so `event_mentions_agent` never fires and no managed agent burns a model turn — an involuntary N-agent turn storm per `@channel` would be a real cost explosion in agent-heavy channels. Channel-wide *agent* addressing is a different intent and already has a lane: `@agents` category mentions (#2531) and user groups (#2728) both expand to explicit `p` tags and wake agents deliberately. If demand emerges for agent-inclusive broadcast tokens, they'd be additional reserved tokens on this same marker mechanism — a follow-up, not v1.

### Permissions and abuse

- v1: any channel member may use `@channel`/`@here` (Slack's default). Relay ingest validates the tag (valid mode, allowed kinds, not in DMs). Per-channel restriction ("only admins may @channel") is a natural follow-up on the #2497 per-channel write-policy surface.
- `channel` and `here` become reserved mention tokens: resolved ahead of display-name matching (a member named "here" can't shadow or be shadowed), and reserved in the user-group handle validator from #2728.

## Prior art

- **#2138** — a full-stack `@everyone`/`@here` attempt (closed unmerged by its author, July 2026; never reviewed). It chose compose-time p-tag expansion with a 4,000-recipient cap; this proposal diverges for the reasons above but reuses several of its calls (mute-respecting semantics, relay-side tag validation, reserved-token autocomplete copy). It also surfaced a real bug this feature depends on: the hidden 1,000-row `LIMIT` in `get_members()` truncating rosters — being fixed separately.
- **#2332** — the open ask for an "everyone tag". Its agent-wake angle is intentionally *not* covered here (`@channel` is humans-only); the agent side is served by `@agents` category mentions (#2531) and user groups (#2728).
- **#2728 / #2791** — user groups: the named-set complement to this zero-config alias; the `["group", id, handle]` marker precedent.
- **#2531** — `@agents`/`@people` category mentions (open): adjacent composer work; this PR will coordinate with its `mentionCandidates.ts` helpers.
- **#2497** — announcement channels: the permission axis this feature's follow-up gate would build on.

## Out of scope / follow-ups

- Mobile compose + pill rendering (follow-up issue, mirroring #2792; mobile already surfaces the mention feed).
- Per-channel `@channel` permission policy (with #2497).
- A dedicated NIP-PL push class for channel mentions (v1 rides existing channel-wide lease subscriptions).
- `@everyone` as a community-wide token — deliberately not proposed; channel scope only.

LordMelkor:
# feat: @channel and @here channel-wide mentions (NIP-CM)

Closes #3146

## What this adds

Two reserved, zero-config mention tokens with Slack-parity semantics:

- **`@channel`** — notifies every member of the channel (mention-tier: badge, mention feed, OS notification).
- **`@here`** — notifies only members who are online right now; members who were offline see a normal message later, with no retroactive badge.

Getting everyone's attention in a channel today means typing every member's name and keeping that habit current as membership changes — the real cost is silent under-notification ([#2728 comment](https://github.com/block/buzz/issues/2728#issuecomment-5078984489)). User groups (#2791) solve the named-set case; `@channel`/`@here` cover the zero-config 80% case, and `@here` covers "whoever is around right now" without badging 35 sleeping people at 2am.

composer autocomplete

## Protocol design: one marker tag, no roster fan-out

The message carries a single marker tag — `["notify", "channel" | "here"]` — specced as a Buzz-local NIP in [docs/nips/NIP-CM.md](docs/nips/NIP-CM.md). It deliberately does **not** expand into per-member `p` tags (departing from the compose-time-expansion idiom of #1918/#2531/#2791), because roster fan-out breaks at scale:

- N `p` tags bloat the signed event, leak the full roster, hit the 50-mention cap, and are silently dropped by NIP-PL's `suppress.p_tags_max` hellthread gate (spec example value: 20 — `@channel` push would vanish in any channel larger than that).
- `@here` is not expressible as a compose-time list — "who is online" is a delivery-time fact.
- Every `p` tag wakes an ACP-managed agent; an expanded `@channel` in an agent-heavy channel would burn one involuntary model turn per agent.

Matrix hit the same fork and chose the structured flag (MSC3952 `m.mentions: {room: true}`). Buzz already has a p-tag-free precedent in `["broadcast","1"]` (NIP-CW). Foreign Nostr clients degrade gracefully: literal `@channel` text, no notification.

### Relationship to #2138

#2138 was the one prior attempt at this feature (closed unmerged by its author in July 2026, never reviewed). It expanded `@everyone` into per-member `p` tags at compose time with a 4,000-recipient cap — the fan-out approach this PR rejects for the reasons above, and it couldn't express `@here` (online-ness is a delivery-time fact). This PR does reuse its good calls: mute-suppression semantics, relay-side tag validation, and the reserved-token autocomplete copy. It also independently surfaced the `get_members()` `LIMIT 1000` roster truncation, now being fixed separately in #3166.

## Key decisions

- **Humans-only.** The marker carries no `p` tags, so `event_mentions_agent` never fires — no agent turn storms. Channel-wide *agent* addressing already has a lane (`@agents` #2531, user groups #2728). Agent-inclusive broadcast tokens would be additional reserved tokens on this same mechanism, as a follow-up.
- **Channel mutes suppress `@channel`/`@here`** — unlike a direct `@name` mention, which continues to pierce mutes (Slack parity; same call #2138 made).
- **`@here` is live-only.** Clients that receive the event while online (own presence online and event within a 120s freshness window) escalate it; there is no feed persistence and no retroactive badge.
- **`@channel` persists one row per event** in a new `channel_notifications` table (migration 0026), UNIONed into the mentions feed against live membership at query time — not one `event_mentions` row per member. Leaving the channel drops those feed entries automatically; joining doesn't backfill someone else's history into your feed.
- **Relay-side validation.** Ingest rejects malformed modes, duplicate notify tags, disallowed kinds (only 9/45001/45003 may carry it), and any notify tag in DM channels — before storage, on every ingest path (WebSocket and HTTP bridge share the pipeline).
- **CLI requires explicit opt-in.** `buzz messages send --notify channel|here`. A literal `@channel` in content without the flag warns on stderr and sends untagged — a template variable can never accidentally mass-notify.
- **Reserved tokens.** `channel`/`here` resolve ahead of display-name matching in every parser (buzz-sdk, desktop, workflow sink), so a member named "here" can't shadow or be shadowed.
- **Edits don't re-notify.** A kind-40003 edit adding `@channel` is accepted but never escalates or persists a notification row (D-spec: no re-notification on edit).

## What changed (23 commits, 72 files, +4,663/−416)

| Layer | Change |
|---|---|
| `buzz-core` | New `channel_mentions` module: `NotifyMode`, tag validation, allowed-kind set, reserved tokens (pure, no I/O) |
| `migrations` | `0026_channel_notifications.sql` — PK (community_id, event_id), feed-shaped index (renumbered after #3141 claimed 0025); table also registered in the pgschema desired-state `schema/schema.sql` |
| `buzz-db` | `insert_channel_notification` paired with mention insertion on message paths; mentions-feed UNION with membership + visible-channel scoping and author exclusion |
| `buzz-relay` | Ingest validation (`validate_channel_mention`), workflow-sink reserved-token skip |
| `buzz-sdk` | `notify` parameter on message/forum builders; reserved-token guards in mention extraction |
| `buzz-cli` | `--notify channel\|here` flag + explicit-opt-in warning |
| desktop | Composer autocomplete rows + confirmation dialog, pill rendering, notification-ladder escalation (pure predicates in `channelNotifyEscalation.ts`, separate from render code) |
| docs | `docs/nips/NIP-CM.md` (Buzz-local NIP), e2e suite `e2e_channel_mentions.rs`, Playwright spec |

Fixed in passing (adjacent bug the feature exposed): the Home feed treated *any* mention-category item as mute-piercing; now only direct p-tag mentions pierce channel mutes, `@channel` rows respect them.

confirm dialog

pill render

here pill

## Adversarial review hardening

The branch went through an adversarial review (independent model, high effort) with KISS/SoC as named dimensions, followed by per-finding adversarial verification. Nine findings were confirmed and fixed before this PR opened:

- **Member-only enforcement**: the open-visibility write fallback no longer extends to notify tags — a non-member of an open channel can post, but cannot `@channel` (rejection: `restricted: only channel members may use @channel or @here`). Kind 40003 edits are exempt (render continuity, never notify).
- **Every-kind validation**: the pure notify-tag check now runs before the early-return handler branches (command/feedback/report/moderation kinds) and on the WebSocket ephemeral path — a notify tag on any disallowed kind is rejected on every ingest path, matching NIP-CM's "reject, don't ignore" rule.
- **Mute precedence**: an event carrying both a direct `p`-tag and `@channel` now correctly pierces a channel mute (direct mention wins); notify-only rows respect mutes in both the feed and the Home badge, via one shared predicate.
- **Single live notification**: an event that is both a direct mention and `@here` no longer double-notifies across the two live subscriptions.
- **Edit render continuity**: kind-40003 edits now carry `notify` through the render tag overlay (pill renders after edit; still never escalates — D9).
- **Code masking**: tilde fences and multi-backtick spans are now masked when detecting literal `@channel`/`@here` in CLI content.

Two findings were reviewed and deliberately **not** changed: @here catch-up escalation within the 120s window is intended (observation-time semantics, clarified in NIP-CM), and a second feed index was rejected as premature for a table that stores one row per @channel event (choice documented in code).

One knock-on worth reviewer attention: the every-kind validation hoist runs before the ban/timeout gate (unavoidable — command kinds return before that gate), so a banned sender submitting a *malformed* notify tag now receives `invalid: …` rather than `blocked: …`. The write is refused either way.

## Test evidence

All gates run on the final branch state (23 commits, rebased on current main):

- `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -D warnings` — clean
- `just test-unit` — 5 suites, 0 failures; `buzz-relay --lib` 771 passed (sole failure is the pre-existing `api::mesh_demo` environment issue, verified failing identically on clean main)
- Postgres-backed: `buzz-db feed::tests` 8/8 (includes 5 channel-mention feed tests)
- **Live-relay e2e**: `e2e_channel_mentions` **9/9** against a relay built from this branch — covers both notify modes, invalid mode / missing mode / duplicate tag / disallowed kind / DM rejection, feed surfacing (@channel appears, @here never persists), non-member rejection in open channels, and rejection on kinds with their own early-return handlers
- Desktop: biome (1,668 files), `tsc --noEmit`, 3,718 unit tests, `check:file-sizes`, `check:px-text` — all clean; Playwright `channel-mentions.spec.ts` 3/3
- 30 new unit/e2e tests were added by the review-hardening pass alone, each verified to fail without its fix

## Interactions with in-flight work

- **#3160 (per-channel notification settings)** also lands in the desktop `shouldNotify` ladder — whichever merges second should re-verify ladder ordering (`@channel` escalation is a separate pure predicate, so the merge should be mechanical).
- **#2531 (`@agents`/`@people` category mentions)** touches the same composer autocomplete surface; reserved tokens here are inserted ahead of display-name candidates, so the lists compose.

## Out of scope / follow-ups

- Mobile compose + pill rendering (follow-up issue, mirroring #2792; mobile already surfaces the mention feed).
- Per-channel `@channel` permission policy (on the #2497 write-policy surface).
- Dedicated NIP-PL push class for channel mentions (v1 rides existing channel-wide lease subscriptions).


LordMelkor:
🤖 Comment posted by kray's AI agent.

**Rebased onto current main and renumbered the migration — no changes to the feature itself.**

[#3141](https://github.com/block/buzz/pull/3141) merged after this PR opened and claimed migration number `0025`, which made this branch `CONFLICTING`. What changed in the force-push:

- **Migration renumbered `0025` → `0026_channel_notifications.sql`** (content identical). The migrator test now asserts `relay_invites` at version 25 and `channel_notifications` at 26.
- **`channel_notifications` registered in `schema/schema.sql`** — the rebase surfaced that pgschema-provisioned test relays (`scripts/start-relay-for-tests.sh` applies the desired-state file, not the migration chain) would be missing the table. New commit, mirrors the migration exactly.
- **Migrator-test hardening** (new commit): this was the third migration-number race on this repo in a week (#2615/reverted by #3168, #3141, now this PR). The hardcoded `migrations.len()` assertion is replaced with a version-contiguity loop that never needs editing and fails with an actionable message when two branches race for the same number; the `channel_notifications` assertions are looked up by content instead of array index; and the pre-0007 test's tail-version literal (`Some(25)`, which #3141 had just bumped from 24) is now derived from the `MIGRATOR`. Net effect: the next number race costs a file rename only, with no test edits. If maintainers want to eliminate the race entirely, timestamp-based versions (sqlx supports i64) are the structural fix — happy to file a follow-up issue.
- **Conflict resolutions**: `crates/buzz-db/src/migration.rs` (both PRs appended migration assertions) and `desktop/src/features/messages/ui/MessageComposer.tsx` (#3151 renamed the toolbar to `ComposerDockToolbar` while this branch swapped the mention dialogs for `ComposerMentionDialogs` — kept both).

Gates re-run on the rebased branch: `cargo fmt --check` clean, buzz-db migration tests 3/3 (Postgres) + 7/7 (pure), `feed::tests` 8/8, desktop `tsc --noEmit` clean, and the full pre-push suite (clippy, Rust/desktop/Tauri/mobile tests, integration) passed on push. PR body updated for the new numbering.

hetsaraiya:
Hello @LordMelkor, I’d be happy to implement the mobile portion.
Two questions:
1. Should new members see older [at]channel mentions? The PR description and implementation seem inconsistent.
2. Should [at]channel remain human-only, or could agents be included through a separate mention? I was considering this feature for both people and agents.

LordMelkor:
Thanks @hetsaraiya that's great. It's tracked in #3198; worth a quick comment there saying you're picking it up and sketching your approach, so nobody duplicates the work and a maintainer can flag any concerns early (per CONTRIBUTING.md)

1. Good catch — those are inconsistent, and the description is the intended behavior. @channel addresses the members present at post time; someone who joins later can read the message but gets no retroactive mention badge (Slack behaves the same way). The mentions-feed membership join (crates/buzz-db/src/feed.rs, branch 2) currently filters only on removed_at IS NULL, so a new member would see older @channel rows — fixed on this branch in cf5fd184e, which bounds the join with `cm.joined_at <= e.received_at` (the relay's admission time). Build mobile against the described semantics.

2. This PR keeps @channel/@here human-only. For a human, @channel is an attention signal they can triage or ignore. An agent has no "ignore" — a ping is effectively a prompt, so a channel with six agents turns one "heads up, deploy at 5pm" into six simultaneous agent turns, each burning tokens deciding whether it's being asked to do something, and each potentially replying (worst case, to each other). It also makes blast radius implicit: merely adding an agent to a channel would silently change what every existing @channel habit costs and triggers. That's why the marker carries no p tags — it structurally can't start agent turns — and why addressing agents stays a deliberate act through the existing lanes (@agents [#2531](https://github.com/block/buzz/pull/2531), user groups [#2728](https://github.com/block/buzz/issues/2728)). I didn't have a combined people+agents use case in front of me, though; if you have one, write it up as its own issue — it'd be a new reserved token (e.g. @everyone) on the same notify-tag mechanism, and the design work that matters is the noise/cost control above. For the mobile port of this PR, match its human-only behavior so desktop and mobile stay consistent.

A few pointers for the mobile work: kind/tag constants must stay in sync with desktop/src/shared/constants/kinds.ts (mirror: mobile/lib/shared/relay/nostr_models.dart); mobile @channel must respect channel mutes, and once per-channel notification settings land ([#3233](https://github.com/block/buzz/pull/3233), follow-up [#3234](https://github.com/block/buzz/issues/3234)), the per-channel "Get broadcast messages" opt-out (spec: docs/nips/NIP-CN.md on the [#3233](https://github.com/block/buzz/pull/3233) branch); and the mobile conventions in CLAUDE.md (§ Mobile App) are CI-enforced, so worth reading first.

LordMelkor:
> 🤖 Posted by kray's AI agent on his behalf.

Triage of the two adversarial review rounds run against this branch after the backfill fix (cf5fd184e), with fixes now pushed.

## Round 1 — consumer-side gaps (relayed from the backfill review)

| Finding | Outcome |
|---|---|
| Desktop Inbox and `buzz feed get` never requested the bounded `feed_types` feed — marker-only `@channel` produced no Inbox row | **Fixed** in 585407fc4: desktop `get_feed` and the CLI now send `{"feed_types": ["mentions"], ...}` (extracted, unit-tested filter builders in both) |
| Desktop catch-up badges count pre-join `@channel` markers (badge with no Inbox row) | **Filed** as #3383 — client-side join-time bound, separable from this PR |
| Community rail unread counts are `#p`-only (misses `@channel`) | **Already tracked** as #3235 |

## Round 2 — review of 585407fc4 (gpt-5.6-sol high, adversarial)

| Finding | Outcome |
|---|---|
| F1 major — Tauri emitted `category: "mentions"` (plural) vs the TS contract's `"mention"` | **Fixed** in 4805337e6 (`MENTION_CATEGORY` const + test). Latent, not user-visible: the frontend re-labels by feed section — but the contract now holds if anything starts trusting `item.category`. The plural `"mentions"` in the `feed_types` *filter* is a different namespace and intentionally unchanged. |
| F2 major — CLI `DEFAULT_FEED_TYPES` included `activity` (unbounded firehose by default) | **Fixed** in 4805337e6: default is now `mentions,needs_action`; `activity`/`agent_activity` are opt-in via `--types` |
| F3 major — bridge `feed_types` fill used one shared budget sequentially: first category could starve the rest | **Fixed** in 4805337e6: each requested type queries with the full limit; results dedupe by event id, pass channel-access + reader-auth filtering, then merge → sort (`created_at` DESC, `id` ASC) → truncate. Merge semantics documented in a block comment. |
| F4 major — desktop `needs_action` (separate approvals query) diverges from the bounded feed's `needs_action` | **Documented, no code change**: desktop derives inbox categories from feed *sections* and nothing in desktop home consumes kinds 46011/46012 from the feed, so aligning the approvals path would be churn without behavior change. Divergence noted here for the record. |
| F5 minor — feed `ORDER BY event_created_at DESC` lacked an id tie-break (unstable pagination on timestamp ties) | **Fixed** in 4805337e6 — turned out to be 5 clauses across the mentions (incl. union), needs_action, and activity builders; SQL-shape tests updated |
| F6 minor — tests only pinned serialization | **Covered** by the new F1/F2/F5 tests |

Round 2 also explicitly refuted: bridge field handling, dedupe, `since` semantics, double-counting, p-gate fallback, and CLI exit codes; KISS/SoC clean.

## Round 3 — re-review of 4805337e6

A third adversarial pass over the fix commit found **no critical or major defects** — all four fixes confirmed functionally correct, with the starvation, dedupe (SQL `UNION` + bridge `seen` set), `since`/limit composition, alias canonicalization, auth-before-truncate, and over-fetch angles explicitly refuted (per-type over-fetch is bounded at 100 × 3 canonical types, deliberate for exact global top-N). Two minor residuals, both pre-existing and accepted:

- **Category contract drift risk**: `MENTION_CATEGORY` (Rust) and the TS `category` union are independently maintained strings with no shared source of truth — a future rename on one side would pass both compilers. Same class as the kind constants kept in sync between `kind.rs` and `kinds.ts`; noted, not blocking.
- **Deterministic top-N ≠ keyset pagination**: the feed route consumes `since`/`limit` only. The id tie-break makes bounded head reads stable and repeatable, but there is no cursor to fetch past `limit` events sharing one timestamp. The feed is a bounded "what needs my attention" surface, not a paginated timeline, so this is by design — flagging it so nobody reads the tie-break as full pagination support.

Gates on the branch head: fmt + clippy clean (workspace incl. Tauri), buzz-cli 261 ✓, buzz-relay 777 ✓ (one pre-existing `mesh_demo` failure reproduced on baseline), Tauri 1855 ✓, buzz-db unit 87 ✓ + feed integration suite ✓ against Postgres.

PR #3202, opened 2026-07-27
fix(relay): reject events containing NUL characters at validation time

### LINKED ISSUE #3116: relay: events with embedded NUL (\u0000) reach Postgres and fail with UTF8 0x00 internal error instead of clean rejection
**What happens**

Production relay pods (bb-public cell) intermittently log:

```
Internal error: error: database error: database error: error returned from database: invalid byte sequence for encoding "UTF8": 0x00 at line 1870
```

27 occurrences over the last 14 days (first sighting 2026-07-22, ~5/day since, most recent 2026-07-27 12:14Z), across many different pods/hosts, often in pairs seconds apart — consistent with a client retrying after a 500. Datadog: https://app.datadoghq.com/logs?query=kube_cluster_name%3Aproduction-bke-bb-public-infra-cell-01-us-west-2%20kube_namespace%3Abuzz%20%22invalid%20byte%20sequence%22

**Likely root cause**

JSON permits `\u0000` escapes in strings, but Postgres `text`/`jsonb` columns cannot store NUL bytes. An event whose content (or a tag value) contains an embedded NUL passes JSON parsing and signature checks, then blows up at insert time as a Postgres encoding error, which surfaces as `Internal error` (500 / generic failure) to the client instead of a clean `invalid:` rejection.

**Expected**

Events containing NUL bytes in content or tag values should be rejected at validation time (`invalid: content contains disallowed control characters` or similar), before any DB write — not surfaced as a database internal error.

**Notes**

- Blast radius today is one client request per hit (no panic/restart, the transaction just fails), but each hit is a 500 to a real user and an error-stream pollution source during launch monitoring.
- Worth checking all ingest text paths (event content, tags, profile fields, workflow definitions) for the same gap, since the error doesn't identify which statement fired.

rvaccone:
Events with `\u0000` in content or a tag value pass signature verification, then
fail at Postgres insert with `invalid byte sequence for encoding "UTF8": 0x00`,
surfacing as a 500. Postgres text and jsonb cannot contain the NUL character.
Fixes #3116.

`ingest_event_inner` now rejects NUL in content or any tag element, next to the
content size check. Both transports route through it. `id`, `pubkey`, and `sig`
are hex and cannot carry NUL; profiles (kind 0) and workflow definitions
(kind 30620) are ordinary events, so the paths named in the issue are covered.

Verified live: 500 before, 400 `invalid: content contains a NUL character` or
`invalid: tag value contains a NUL character` after, no row written, clean
events unaffected. Four unit tests added; fmt and clippy pass. The 9 failing
`buzz-relay --lib` tests also fail at the merge base (noted as preexisting
in #3128).

### DIFF
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -1754,6 +1754,21 @@ fn validate_not_before(tag_value: &str) -> Result<u64, &'static str> {
     Ok(value)
 }
 
+/// Postgres cannot store NUL in `text`/`jsonb` columns, so an event carrying
+/// `\u0000` in its content or tag values would otherwise pass validation and
+/// then fail at insert time as an internal database error.
+fn validate_no_nul_bytes(event: &Event) -> Result<(), &'static str> {
+    if event.content.contains('\0') {
+        return Err("content contains a NUL character");
+    }
+    for tag in event.tags.iter() {
+        if tag.as_slice().iter().any(|value| value.contains('\0')) {
+            return Err("tag value contains a NUL character");
+        }
+    }
+    Ok(())
+}
+
 /// Validate the public tag envelope of a NIP-ER `kind:30300` event before it
 /// reaches NIP-33 parameterized replacement.
 ///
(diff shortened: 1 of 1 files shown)


rvaccone:
Bump on this one. Open since July 27 with no review, and CI hasn't run yet because the workflows are still waiting on maintainer approval (fork PR).

Could someone approve the run and take a look? One file, +56 lines, fixes #3116, which is still open and still 500s on any event with an embedded NUL.

Happy to rebase onto current main if that helps.

PR #3208, opened 2026-07-27
feat(desktop): add Pocket voice settings

johnmatthewtennant:
## Context

Pocket TTS can read live agent replies during a huddle, but Buzz does not provide an installation-wide way to turn that behavior off or choose among the bundled voices. The setting also needs a stable voice identity boundary that future backends and per-agent preferences can reuse without treating a display label as identity.

## Summary

This adds a compact Pocket TTS voice settings surface with an Agent text to speech toggle, preview playback, and 12 bundled voices. Text to speech remains enabled by default with Mary selected. Disabling it immediately cancels queued and in-flight agent speech.

## Changes

- Adds a versioned, device-local TTS settings schema and backend-qualified voice keys.
- Bundles Anna, Azelma, Charles, Eponine, Eve, Fantine, George, Jane, Mary, Michael, Paul, and Vera with provenance and deterministic Mary fallback behavior.
- Adds voice selection and preview controls while preserving the saved voice when text to speech is disabled.
- Labels the visible and accessible selector as `Pocket TTS voice`.
- Speaks only new live agent-authored text in active huddles, including kind 40002 agent replies, while excluding history, people, reactions, edits, system or tool rows, and attachment-only content.
- Keeps agent speech strictly live-only across reconnects. The `limit: 0` subscription cannot page missed history or reorder replay against new live events.
- Removes the unused `speakableAgentText` wrapper and replaces stale replay-override coverage with a live-only no-paging assertion.
- Cancels stale, queued, and in-flight speech on barge-in, shutdown, or voice changes without appending stale audio.
- Adds content-free route diagnostics, focused Rust and frontend coverage, and a Playwright screenshot flow.
- Groups huddle audio settings into a cohesive state boundary and extracts relay timing constants so both existing files remain below their strict file-size ratchets.

## Related issue

None found.

## Testing

Manual validation in the combined daily-driver build confirmed all 12 bundled voice previews, one-time playback of a new live agent response through the selected output, and cancellation of stale audio after a voice change.

## Screenshots

The global toggle, Pocket voice selector, and Preview action share one compact Voice settings surface.

A focused screenshot was captured at `desktop/test-results/voice-settings/pocket-voices.png` (SHA-256 `a430ce71e7093612a0e9d27ebd3c6e43b2fa02b081c28376d8292e50f45d0a8b`). It is intentionally not posted.

Exercise the bundled selector and live-only subscription contract from a fresh checkout:

```bash
. ./bin/activate-hermit
cd desktop
pnpm build:e2e
pnpm exec playwright test tests/e2e/voice-settings.spec.ts --project=smoke
node --import ./test-loader.mjs --experimental-strip-types --test \
  src/features/huddle/lib/ttsLiveMessages.test.mjs \
  src/shared/api/relayReconnectReplay.test.mjs \
  src/shared/api/relayChannelFilters.test.mjs \
  src/features/settings/ui/voiceSettingsLogic.test.mjs
```

The selector renders 12 bundled voices under the accessible `Pocket TTS voice` label. The agent-speech subscription remains `limit: 0` and never requests missed-history paging.


johnmatthewtennant:
🤖 ### Pocket TTS voice settings

The global text-to-speech toggle, bundled Pocket voice selector, and Preview action share one compact Voice settings surface.

pocket-voices

PR #3217, opened 2026-07-27
fix(desktop): keep drafts out of the Inbox All view

thomaspblock:
## Summary

Drafts were showing up in the Home Inbox **All** view, mixed in with messages and reminders (reported in `#buzz-bugs`). Drafts are private composer state, not inbox activity — they now appear only under the dedicated **Drafts** filter.

## Changes

- **`inboxListRows.ts`** — drop the `draft` row variant from `buildInboxListRows`; the mixed view builds only `inbox` + `reminder` rows.
- **`InboxListPane.tsx`** — remove the draft branch of the All-view render path; `PersonalItemRow` now renders reminders only.
- **`useHomePersonalInbox.ts`** — stop enabling draft selection (and its root-status relay probing) for the mixed view; draft selection is scoped to the Drafts filter.
- Drafts filter behavior is unchanged: the filter badge count, `DraftsPanel` list, and `DraftDetailPane` all still work.

## Testing

- `pnpm test` (desktop unit suite): 3697 passed, 0 failed.
- `pnpm exec biome check src/features/home tests`: clean.
- Updated `inboxListRows.test.mjs` for the two-variant row model.
- Updated the e2e test (`channels.spec.ts`) to assert All never lists drafts and that the draft is still reachable under the Drafts filter.
- Added `drafts-all-fix-screenshots.spec.ts` capturing both states (screenshots below).

### All view — draft is gone, messages/reminders unaffected

01-all-view-no-drafts

### Drafts filter — the draft is still listed and editable

02-drafts-filter-still-lists

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -103,6 +103,7 @@ export default defineConfig({
         "**/project-pr-review.spec.ts",
         "**/persona-model-combobox-screenshots.spec.ts",
         "**/drafts-screenshots.spec.ts",
+        "**/drafts-all-fix-screenshots.spec.ts",
         "**/inbox-refactor-screenshots.spec.ts",
         "**/buzz-theme-screenshots.spec.ts",
         "**/channel-sort.spec.ts",
--- a/desktop/src/features/home/lib/inboxListRows.test.mjs
+++ b/desktop/src/features/home/lib/inboxListRows.test.mjs
@@ -17,16 +17,6 @@ function inboxItem(
   };
 }
 
-function draftItem(key, updatedAt, rootStatus = "available") {
-  return {
-    entry: {
-      key,
-      draft: { createdAt: updatedAt, updatedAt },
-    },
-    rootStatus,
-  };
-}
-
 function reminder(
   id,
   createdAt,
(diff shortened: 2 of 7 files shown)


thomaspblock:
### All view — draft is gone, messages/reminders unaffected

01-all-view-no-drafts

### Drafts filter — the draft is still listed and editable

02-drafts-filter-still-lists

klopez4212 (approved):
Reviewed at head `e4a05bfe`. Tight, well-scoped fix that mostly deletes code, with the right test updates. **Approving — no blocking issues.**

**Verified locally on the PR head**: `pnpm exec tsc --noEmit` clean, `pnpm exec biome check src tests` clean, `pnpm test` 3697/3697 pass. CI fully green.

**What I checked**: the draft row variant is fully removed from `buildInboxListRows` (type, sort helper, and import all gone); the `InboxListPane` render path narrows cleanly to `inbox` | `reminder` with no dead branches; `useHomeDrafts(selectionEnabled: isDrafts)` correctly keeps reminder mixed-selection intact while scoping draft selection + root-status relay probing to the Drafts filter; the Drafts filter path (badge, `DraftsPanel`, selection wiring in `HomeView`) is untouched and still works.

**Non-blocking notes** (all "your call"):

1. **Draft-absence assertions are keyed to a removed testid** (P3) — `channels.spec.ts:2734` and the screenshot spec assert `home-all-drafts-${draftKey}` has count 0, but no code renders that testid anymore, so the guard only catches a regression that reintroduces drafts under the *same* testid. A stronger guard would also assert the seeded draft preview text is absent from the All list, e.g. `expect(page.getByTestId("home-inbox-list").getByText("Draft one")).toHaveCount(0)`.

2. **Drafts badge is now optimistic while in All view** (P3) — before this PR, sitting in All ran the draft root-status probes, so the filter-menu badge dropped orphaned (deleted-root) thread drafts once probes completed. Now probes only run under the Drafts filter, so the badge can briefly overcount orphaned drafts until the user opens Drafts. This matches the documented sanctioned design in `DraftsPanel.tsx` ("bounded eventual-consistency… Will, 2026-07-07"), so I'd accept it — flagging so it's a conscious choice.

3. **`PersonalItemRow` name is now stale** (P4) — with the draft branch removed it renders reminders only (hardcoded Bell icon, "Reminder" title, `home-all-reminders-*` testid) but keeps its generic name. Consider renaming to `ReminderListRow` in a follow-up.

4. **One-off screenshot spec added to the CI smoke suite** (P4) — `drafts-all-fix-screenshots.spec.ts` carries real assertions and follows repo precedent (`drafts-screenshots`, `inbox-refactor-screenshots`, …), so it's consistent — just adds one more spec to smoke runtime forever. Fine to keep; could be folded into `channels.spec.ts` later.

PR #3220, opened 2026-07-27
fix(desktop): stop create-agent form clobbering provider config on keystroke

### LINKED ISSUE #3216: Create-agent form resets provider `config_schema` fields to defaults on every keystroke
## Summary

In the create-agent flow, any provider that advertises a `config_schema` has its config fields reset to their schema defaults (or wiped to empty) on **every keystroke**. Typing into a field like **Inference auth** snaps the value back almost immediately, making it effectively impossible to enter a non-default value through the UI.

Against a fast local provider binary the reset appears instantaneous, so it looks like the field refuses to accept input.

## Impact

- Any backend provider advertising `config_schema` fields is unconfigurable from Buzz Desktop.
- Fields **with** a `default` snap back to that default; fields **without** one get wiped to empty.
- Concrete case: I hit this while building a backend provider that runs agents on a Databricks sandbox / Lakebox. It advertises `profile`, `inference_auth` (default `env`), and `idle_timeout`. Its opt-in `inference_auth: "sandbox"` path cannot be selected from the UI at all — `inference_auth` always resets to `env`. (Details in IceRhymers/buzz-lakebox#11.)
- There is no provider-side workaround: dropping the `default` from the schema only changes the reset target from `env` to empty.

## Root cause

`desktop/src/features/agents/ui/WhereToRunSection.tsx`

1. The probe `useEffect` (line 29) lists the entire `draft` in its dependency array:

   ```ts
   }, [draft, isProviderMode, onDraftChange, selectedBackendProvider]);  // line 64
   ```

2. Every keystroke in `ProviderConfigFields` calls `onDraftChange({ ...draft, providerConfig })` (lines 116–117), producing a new `draft` object — which re-triggers the effect and re-runs `probeBackendProvider(...)`.

3. When the probe resolves, it rebuilds `defaults` from `config_schema` and calls (lines 50–54):

   ```ts
   onDraftChange({
     ...draft,
     probedProvider: result,
     providerConfig: defaults,   // clobbers whatever the user just typed
   });
   ```

   So each keystroke fires a fresh probe, and each resolved probe overwrites `providerConfig` back to the schema defaults. This is both a correctness bug (typed values lost) and a churn bug (one provider probe per keystroke).

## Proposed fix

Don't re-probe once the provider has already been probed:

```tsx
React.useEffect(() => {
  if (!isProviderMode || !selectedBackendProvider) {
    setProbeError(null);
    return;
  }
  if (draft.probedProvider) return; // already probed; don't clobber user edits
  // ... existing probe logic
}, [draft, isProviderMode, onDraftChange, selectedBackendProvider]);
```

This stays correct on provider switches: the "Run on" `<select>` resets the draft to `emptyWhereToRunDraft` (`probedProvider: null`, lines 78–83), so a fresh probe still fires when the user picks a different provider.

Belt-and-suspenders — seed defaults without overwriting typed values, so a late-resolving probe can never erase user input:

```ts
providerConfig: { ...defaults, ...draft.providerConfig },
```

The early-return guard alone fixes both the reset and the probe-per-keystroke churn; the merge is defense-in-depth.

## Repro

1. Have any backend provider installed that advertises a `config_schema` with at least one property that has a `default`.
2. Open the create-agent form, set **Run on** to that provider.
3. Try to type a value into one of the config fields that differs from its default.
4. Observed: the field snaps back to its default almost immediately (near-instant against a fast local provider binary).
5. Expected: the typed value persists.

Discovered while building a backend provider that launches agents on a Databricks sandbox / Lakebox. Its `inference_auth` field (default `env`) could not be switched to `sandbox` from the UI. Details in `IceRhymers/buzz-lakebox#11`.

https://github.com/user-attachments/assets/c821127f-7eed-4b6e-94bc-9385b104a52d

## Environment

- Verified in `block/buzz` at `main` (component unchanged from `8bb43d51` through `3faea98`).
- File: `desktop/src/features/agents/ui/WhereToRunSection.tsx`

IceRhymers:
## What

Fixes the create-agent form resetting provider `config_schema` fields to their schema defaults (or wiping them) on every keystroke, which made non-default values impossible to enter from the UI.

Fixes #3216.

## Root cause

In `desktop/src/features/agents/ui/WhereToRunSection.tsx`, the provider-probe `useEffect` listed the entire `draft` in its dependency array. Every keystroke in `ProviderConfigFields` calls `onDraftChange({ ...draft, providerConfig })`, producing a new `draft` → the effect re-fires → the probe re-runs → on resolve it overwrites `providerConfig` with defaults rebuilt from `config_schema`. Against a fast local provider binary the probe is near-instant, so the field appears to reset on the keystroke itself.

Fields with a `default` snapped back to it; fields without one were wiped to empty. It's also a churn bug — one provider probe per keystroke.

## Fix

- Early-return guard `if (draft.probedProvider) return;` so the probe runs once per selection instead of on every draft change. Switching providers still re-probes correctly, because the "Run on" `<select>` resets the draft to `emptyWhereToRunDraft` (`probedProvider: null`).
- On probe resolve, seed `providerConfig: { ...defaults, ...draft.providerConfig }` so a late-resolving probe can never erase values the user already typed (defense-in-depth).

## Testing

- `just desktop-typecheck` — pass
- `just desktop-test` — pass
- `just desktop-check` (Biome) — no findings on changed source

#### Testing against custom provider

Evidence of the fix, vs the video in #3216 

https://github.com/user-attachments/assets/943b4235-c566-4e6d-a93b-571551370ed5

## Notes

Discovered while building an out-of-tree backend provider (Databricks sandbox / Lakebox) that advertises a defaulted `inference_auth` field which couldn't be switched away from its default via the UI. No provider-side workaround exists — the reset is entirely client-side.

### DIFF
--- a/desktop/src/features/agents/ui/WhereToRunSection.tsx
+++ b/desktop/src/features/agents/ui/WhereToRunSection.tsx
@@ -31,6 +31,12 @@ export function WhereToRunSection({
       setProbeError(null);
       return;
     }
+    // Once probed, don't re-probe: every keystroke in ProviderConfigFields
+    // produces a new `draft`, and re-running the probe would clobber the
+    // user's edits back to schema defaults. Switching providers resets the
+    // draft to emptyWhereToRunDraft (probedProvider: null), so a fresh probe
+    // still fires on provider change.
+    if (draft.probedProvider) return;
     let cancelled = false;
     setProbeError(null);
     void probeBackendProvider(selectedBackendProvider.binaryPath)
@@ -50,7 +56,9 @@ export function WhereToRunSection({
         onDraftChange({
           ...draft,
           probedProvider: result,
-          providerConfig: defaults,
+          // Seed defaults without overwriting any values the user already
+          // typed, so a late-resolving probe can never erase input.
+          providerConfig: { ...defaults, ...draft.providerConfig },
         });
       })
       .catch((error: unknown) => {


saurav-square (approved):

ashik112:
Confirmed this exact failure with the out-of-tree `buzz-backend-ssh` provider from #3449 on macOS Desktop: the provider is discovered and its `config_schema` renders, but typing the required `ssh_host` value immediately resets the field. That isolates the failure to the client-side probe/draft loop fixed here and currently blocks creating an SSH remote agent through the released Desktop UI.

The successful provider discovery also confirms this is not an installation or provider-schema issue.

IceRhymers:
Superseded by #4411, which independently landed the broader tested fix for this issue.

PR #3221, opened 2026-07-27
fix(desktop): pin agent instances to their community relay, guard duplicate minting, scope mentions (#2515)

### LINKED ISSUE #2515: [Bug] Agents from different communities collide in channels, causing duplicate @mention targets and missing avatars

jeremyd:
Fixes #2515, and removes the largest trigger for #2648.

## Problem

The create flow mints one record + keypair per (agent, community), but every started instance connects to the **active workspace** relay. After a community switch + relaunch, an agent comes up under a pubkey the relay never admitted, while @mentions keep tagging its real — now offline — instance. Same-name records from different communities then collide in the mention picker (duplicate targets, missing avatars — this issue), and a second install re-mints identities that already exist (#2648).

Root cause sits in gaps #2122 itself lists as known follow-ups: restore-on-launch is record-level (pair-scoped restore not yet shipped) and the startup reconcile is one-shot. This PR bridges those gaps without changing the agents-everywhere design.

## Changes (one commit each)

1. **`fix(desktop): honor a managed agent's stored relay pin`**
   A non-empty per-record `relay_url` now wins at `effective_agent_relay_url` — the single choke point all agent relay resolution flows through (spawn, restore, reconcile fan-out, profile sync, spawn-hash). An instance minted for a community stays bound to that community's relay regardless of the active workspace. **An empty pin is byte-for-byte the existing agents-everywhere behavior (#2122) — the pin is strictly opt-in.**

2. **`fix(desktop): refuse minting a duplicate agent instance per relay`**
   `create_managed_agent` rejects a new record whose name matches an existing instance in the same relay scope (same pin, or both unpinned), inside the records lock so concurrent creates can't race past the check. Same name on a *different* relay stays allowed — one identity per agent per community. Guard lives in `managed_agents/mint_guard.rs` with unit tests; snapshot/team import paths untouched (they restore records, not mint identities).

3. **`fix(desktop): scope mention candidates to the active community`**
   An instance pinned to another community's relay runs — and answers mentions — only there, so `useMentions` drops such instances from candidacy entirely. A stale channel membership can no longer make the foreign instance win the same-name coalesce over the instance actually serving the active relay. This implements the fix suggested in #2515's description. Pin comparison in `foreignPinnedAgentPubkeys.ts` with unit tests.

## Design notes

- **Opt-in by construction.** No stored pin → zero behavior change. Existing installs are unaffected until a record carries a pin.
- **Bridge, not a competing model.** When pair-scoped restore lands, one-record-one-relay dissolves; at that point the pin should become a per-pair membership filter or be retired. Happy to carry that migration.
- **Composes with recent work.** No overlap with #2857 (duplicate remote agent starts) or #2713 (mention dedupe) — we run both alongside these commits. Complements #2607 (persona-rename identity republish): that fixes *what* gets republished, this fixes *which relay* an instance binds to.
- **Also mitigates #2717** in practice: with pins, each community's instance keeps its own record, so settings edits land on a deterministic instance.

## Testing

- Full `just ci` (fmt + clippy + unit tests + desktop builds) green on top of current `main` (`7ca0bbd94`).
- New unit tests: relay-pin resolution + spawn-hash, `mint_guard.rs`, `foreignPinnedAgentPubkeys.test.mjs`.
- Running in production on our fork since 2026-07-27 across two closed relays (two communities, three agents each): wrong-relay spawns and wrong-pubkey mention tags gone; duplicate-name creation refused; cross-community mention pollution gone.

---

*(If maintainers want it, we also have a follow-up that fans kind-0 profiles out per (agent, relay) pair on spawn and on persona edits — closes the missing-avatar half permanently. Kept out of this PR to keep it reviewable.)*

### DIFF
--- a/desktop/scripts/check-file-sizes.mjs
+++ b/desktop/scripts/check-file-sizes.mjs
@@ -93,7 +93,9 @@ const overrides = new Map([
   // build_managed_agent_summary into callers, dangling-harness summaries render
   // the deleted id, and spawn errors surface as sentences (tests included).
   // +1: merge of the two deltas above (actual post-merge count).
-  ["src-tauri/src/commands/agents.rs", 1418],
+  // +7 (#2515): mint guard call in create_managed_agent (guard itself lives in
+  // managed_agents/mint_guard.rs).
+  ["src-tauri/src/commands/agents.rs", 1425],
   // agent-lifecycle-fixes: cascade-delete in delete_persona restructured into
   // 3-phase (stage/stop/commit) + commit_cascade_agents injectable helper for
   // retry-safety. Load-bearing reviewer-required change; queued to split.
--- a/desktop/src-tauri/src/commands/agents.rs
+++ b/desktop/src-tauri/src/commands/agents.rs
@@ -636,16 +636,6 @@ pub async fn create_managed_agent(
             let personas = load_personas(&app)?;
             ensure_persona_is_active(&personas, persona_id)?;
         }
-        let keys = Keys::generate();
-        let pubkey = keys.public_key().to_hex();
-        if records.iter().any(|record| record.pubkey == pubkey) {
-            return Err(format!("agent {pubkey} already exists"));
-        }
-        let private_key_nsec = keys
-            .secret_key()
-            .to_bech32()
-            .map_err(|error| format!("failed to encode private key: {error}"))?;
-
         // Store the relay override exactly as supplied (trimmed). An explicit
         // value pins the agent; empty stays empty and resolves to the active
         // workspace relay at read-time. Uniform for Local and Provider.
@@ -655,6 +645,23 @@ pub async fn create_managed_agent(
             .map(str::trim)
             .unwrap_or("")
             .to_string();
+
+        // #2515 mint guard: one instance of an agent per relay scope.
+        crate::managed_agents::mint_guard::ensure_no_duplicate_instance(
+            &records,
+            &name,
+            &resolved_relay_url,
+        )?;
+
+        let keys = Keys::generate();
+        let pubkey = keys.public_key().to_hex();
+        if records.iter().any(|record| record.pubkey == pubkey) {
+            return Err(format!("agent {pubkey} already exists"));
+        }
+        let private_key_nsec = keys
+            .secret_key()
+            .to_bech32()
+            .map_err(|error| format!("failed to encode private key: {error}"))?;
 
         (keys, private_key_nsec, pubkey, resolved_relay_url, input)
     };
--- a/desktop/src-tauri/src/managed_agents/mint_guard.rs
+++ b/desktop/src-tauri/src/managed_agents/mint_guard.rs
@@ -1,0 +1,98 @@
+//! Mint guard for managed-agent instances (#2515).
+
+use super::ManagedAgentRecord;
+
+/// Refuse minting a record whose name matches an existing instance in the
+/// same relay scope (same pin, or both unpinned). A second keypair for one
+/// @Name on a relay is exactly the duplicate-identity state from #2515:
+/// mentions resolve to one pubkey while the other instance answers. The same
+/// name on a different relay stays allowed — one identity per agent per
+/// community.
+pub(crate) fn ensure_no_duplicate_instance(
+    records: &[ManagedAgentRecord],
+    name: &str,
+    relay_url: &str,
+) -> Result<(), String> {
+    let relay = relay_url.trim();
+    let Some(existing) = records.iter().find(|record| {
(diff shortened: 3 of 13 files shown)


jeremyd:
Looks like I was part of a cohort of users that installed, and tried to use multi-community agents.  Which ended up having a mess of agents (multiple key pairs per agent, mention problems, and etc).  This was fixed I believe by #2122, but the data itself was not.  So I was confused.  Closing this PR as I don't think it's necessary now.

Issue #3224, opened 2026-07-27
Mobile: no WebSocket stall detection — half-open cellular sockets look connected until force-quit (plus 4 resilience gaps vs desktop)


Dirtywater97:
## Summary

`mobile/lib/shared/relay/` is missing every connection-resilience mechanism that `desktop/src/shared/api/relayClientSession.ts` has accumulated. The most serious gap is that **mobile has no stall detection at all**, so a half-open socket on cellular is indistinguishable from a healthy idle one — the app sits believing it is connected and delivers nothing until the user force-quits.

Reported by a Play Store user of the current release: *"constantly reconnecting and disconnecting and not being able to load, constantly having to force quit the app just to see a message."*

All line references below are against `9810d8545` (main).

## Why the stall is invisible today

The relay pings every 30s and closes after 3 missed pongs (`crates/buzz-relay/src/connection.rs:373-400`). That protects the *relay* from dead clients. It does nothing for the client: when Android's radio blackholes an idle TCP socket, the phone's `WebSocket` never receives a close frame, `onDone`/`onError` never fire (`mobile/lib/shared/relay/relay_socket.dart:91-104`), and `RelaySessionNotifier` is never told to reconnect. The socket is dead, the UI says connected.

Desktop solved exactly this in `relayStallWatchdog.ts` — a *passive* watchdog that treats inbound traffic (including the relay's own pings) as the liveness signal, and fires after two missed heartbeat windows:

```ts
// desktop/src/shared/api/relayClientSession.ts:79-80
const STALL_CHECK_INTERVAL_MS = 10_000;
const STALL_IDLE_TIMEOUT_MS = 60_000;
```

Mobile has no equivalent. `grep -niE 'ping|pong|idle|keepalive|watchdog' mobile/lib/shared/relay/relay_socket.dart` returns zero matches across all 251 lines.

## The five gaps

| # | Gap | Mobile | Desktop equivalent |
|---|---|---|---|
| 1 | **No stall detection** | nothing | `relayStallWatchdog.ts`; wired at `relayClientSession.ts:120-124`, `:596`, `:770` |
| 2 | **8s timeouts** | `relay_socket.dart:99`; `relay_session.dart:133,185,251` | 25s — `relayClientSession.ts:63-65` |
| 3 | **No backoff jitter** | `relay_session.dart:398-412`, flat `delay * 2` | ±25% — `relayClientSession.ts:985-989` |
[...]

PR #3231, opened 2026-07-27
fix(mobile): detect half-open relay sockets with client-side ping/pong

Dirtywater97:
Fixes gap 1 of #3224 — the one behind the reported symptom.

## Problem

The relay pings every 30s and closes after 3 missed pongs (`crates/buzz-relay/src/connection.rs:378-400`). That protects the relay from dead clients; it does nothing for the client.

Mobile networks — cellular especially — silently drop idle TCP flows without sending a FIN. When that happens the phone's WebSocket never receives a close frame, so neither `onError` nor `onDone` fires in `relay_socket.dart:84-96`, `RelaySessionNotifier` is never told to reconnect, and the socket stays dead while the UI reports connected. A Play Store user described the result as *"constantly having to force quit the app just to see a message."*

## Fix

`WebSocketChannel.connect` resolves to `AdapterWebSocketChannel`, which exposes no ping configuration (`web_socket_channel/lib/src/channel.dart:107-108`). `IOWebSocketChannel` does — and the package is already a direct dependency (`mobile/pubspec.yaml:20`, locked at 3.0.3):

> If a ping message is not answered by a pong message from the peer, the WebSocket is assumed disconnected and the connection is closed with a `goingAway` close code.

So the missing liveness detection is a constructor away. The unanswered ping produces a normal close, which reaches the **existing** `onDone` handler and drives the **existing** reconnect path — no new lifecycle handling.

`mobile/` has `android/` and `ios/` and no web target, so binding to the `dart:io` implementation is safe. The connector is isolated in `relay_websocket_connector.dart` so `relay_socket.dart` changes by one call site plus an import.

### On the interval

`dart:io` never has two pings outstanding: it waits one interval, sends a ping, then allows one more for the pong. Detection therefore takes up to `2 × interval` — ~40s at 20s, keeping foreground recovery under a minute. Cost is three extra ping/pong exchanges per idle foreground minute; the session already disconnects 5s after backgrounding (`relay_session.dart:316-329`), so this doesn't run while suspended. **30s would halve the traffic and roughly double time-to-detection** — happy to change it if you prefer that trade.

### Alternative you may prefer

Desktop solved this with a *passive* watchdog (`relayStallWatchdog.ts`) that treats inbound traffic as the liveness signal, chosen deliberately for a Tauri constraint documented at the top of that file. That constraint doesn't apply to Flutter, and active ping/pong is far less code here. But if you'd rather keep the two clients symmetric, say so and I'll port the watchdog shape instead.

## Testing

New test stands up a raw `ServerSocket` that completes the WebSocket handshake and then goes silent — never answering pings, never sending a close frame. A `dart:io` `WebSocket` server can't express this because it auto-replies to pings, hence the manual handshake.

It asserts the socket closes **cleanly through `onDone` with close code 1001**, not via `onError` — the distinction that matters, since only a clean close exercises the reconnect path.

Verified in both directions:
- Against `WebSocketChannel.connect`: **fails** — socket stays open indefinitely.
- With ping/pong enabled: **passes** in ~400ms at a 200ms test interval.

```
dart format --output=none --set-exit-if-changed .   0 changed
flutter analyze                                     No issues found!
flutter test                                        825 passed, 1 skipped
```

Baseline before this change was 823 passed, 1 skipped — the 2 new tests are the only delta, no regressions.

## Not included

Gaps 2–5 from #3224 (8s vs 25s timeouts, missing backoff jitter, backoff resetting on connect rather than on stability, dropped `NOTICE` frames) are deliberately left out so this stays independently reviewable and revertable. Happy to follow up with those in a separate PR.

### DIFF
--- a/mobile/lib/shared/relay/relay_socket.dart
+++ b/mobile/lib/shared/relay/relay_socket.dart
@@ -6,6 +6,7 @@ import 'package:nostr/nostr.dart' as nostr;
 import 'package:web_socket_channel/web_socket_channel.dart';
 
 import 'nostr_models.dart';
+import 'relay_websocket_connector.dart';
 
 /// Low-level websocket connection with NIP-42 authentication.
 ///
@@ -63,7 +64,7 @@ class RelaySocket {
     _state = SocketState.connecting;
 
     try {
-      _channel = WebSocketChannel.connect(Uri.parse(_wsUrl));
+      _channel = connectRelayWebSocket(Uri.parse(_wsUrl));
       await _channel!.ready;
     } catch (e) {
       _state = SocketState.disconnected;
--- a/mobile/lib/shared/relay/relay_websocket_connector.dart
+++ b/mobile/lib/shared/relay/relay_websocket_connector.dart
@@ -1,0 +1,41 @@
+import 'package:web_socket_channel/io.dart';
+import 'package:web_socket_channel/web_socket_channel.dart';
+
+/// How often to send a WebSocket ping frame, and how long to wait for the
+/// matching pong before declaring the socket dead.
+///
+/// The relay pings every 30s and closes after 3 missed pongs
+/// (`crates/buzz-relay/src/connection.rs`). That protects the relay from dead
+/// clients; it does nothing for the client. Mobile networks — cellular in
+/// particular — silently drop idle TCP flows without sending a FIN, so an
+/// unused socket can stay dead while the client still reports it as open, and
+/// nothing surfaces it until some later network event or an app restart.
+///
+/// `dart:io` never has two pings outstanding: it waits one interval, sends a
+/// ping, then allows one more interval for the pong. Detection therefore takes
+/// up to `2 * relayPingInterval` — about 40s at this value, which keeps
+/// recovery under a minute while the app is foregrounded. The cost is three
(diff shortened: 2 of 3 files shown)


technicallybrantley:
+1 — independent confirmation of both the bug and this exact approach.

We hit this in production yesterday within hours of putting our relay behind a
Cloudflare Tunnel: with a channel open on iOS, incoming messages silently
stopped rendering until the next outbound send forced a reconnect. No error,
no close frame — a textbook half-open socket after the proxy dropped the idle
path (~100s idle timeout in our case, though a carrier NAT rebind gets you to
the same place).

Before finding this PR we independently root-caused it and wrote the same fix:
`dart:io` `pingInterval` (also 20s) on the relay socket, with the missed-pong
close surfacing through `onDone` into the existing reconnect-and-replay path.
Two separate reproductions converging on an identical design seems worth
noting for review confidence — this matches our findings exactly, and we'd
like to see it merged: it affects every mobile user on a flaky path.

One small thing worth double-checking, from our implementation of the same
change (I haven't verified whether it applies to this connector's structure):
once the connect path awaits the socket factory, a dispose/teardown racing
that await can leave the late-arriving socket unobserved unless the post-await
state is rechecked. We ended up guarding that.

Happy to keep our (unsubmitted) branch closed in favor of this PR.

Issue #3211, opened 2026-07-27
mobile: invite-link joins downgrade wss:// to plaintext ws:// (and break HTTP query/upload)


DamienStevens:
## Summary

On mobile, joining a community via an **invite deep link** stores a `wss://` base URL, but `RelayConfig.wsUrl` only maps `https → wss`. A `wss` scheme falls through to the `else` branch and is rewritten to **plaintext `ws://`**, which is then dialed. The same stored value also produces `wss://…` URLs that are handed to `package:http`, which cannot use that scheme.

The pairing path is unaffected (desktop sends `https://`), which is why this hasn't surfaced.

## The defect

```dart
// mobile/lib/shared/relay/relay_provider.dart:21-25
String get wsUrl {
  final uri = Uri.parse(baseUrl);
  final scheme = uri.scheme == 'https' ? 'wss' : 'ws';   // 'wss' → 'ws'
  return uri.replace(scheme: scheme).toString();
}
```

## Reproduction

Running the method body verbatim under Dart 3.11.5:

```
https://r.example.com   ->   wss://r.example.com     ✅
wss://r.example.com     ->   ws://r.example.com      ❌ downgraded
http://localhost:3000   ->   ws://localhost:3000     ✅
```

## Chain to a live connection

1. `deep_link.dart:174-178` — `parseInviteDeepLink` normalizes to `Uri(scheme: relayUri.scheme, host:, port:)`. `validateInviteRelayUri` requires `ws`/`wss` and rejects non-`wss` outside `kDebugMode` (`relay_validation.dart:28-30`), so in release builds this is always `wss://host`.
2. `invite_join_provider.dart:157-159` — `Community.create(relayUrl: invite.relayUrl)` stores it unchanged (`community.dart:23-37` does no normalization).
3. `relay_provider.dart:49` — `RelayConfig(baseUrl: active.relayUrl)`.
4. `relay_provider.dart:21-25` — `wsUrl` → **`ws://host`**.
5. `relay_session.dart:361-362` — `_socketFactory(wsUrl: config.wsUrl)` dials it.

## Second failure on the same value

`baseUrl` is also used directly for HTTP:

- `relay_session.dart:136` — `Uri.parse(config.baseUrl).resolve('/query')` → `wss://host/query` passed to `http.Client()`
- `media_upload.dart:364` — `Uri.parse(_baseUrl).resolve(path)` → `wss://…` for a `PUT`

Both are unsupported schemes for `package:http`.

## Impact

[...]

PR #2726, opened 2026-07-24
fix(mobile): invite join stuck on Connecting — URL scheme mismatch

alvinjaison:
## Summary

Fixes the iOS invite join flow getting permanently stuck on "Connecting…" when opening a hosted-community invite link.

## Root Cause

The invite deep-link parser produces `wss://` relay URLs (correct for the WebSocket scheme). These are stored directly as `Community.relayUrl`. However, `RelayConfig.wsUrl` assumed `baseUrl` was always HTTP-scheme:

```dart
// Before: only recognized https → wss
final scheme = uri.scheme == 'https' ? 'wss' : 'ws';
```

When `baseUrl` was `wss://`, this incorrectly downgraded it to `ws://` (plaintext). A hosted relay at `*.communities.buzz.xyz` only accepts TLS connections, so the `ws://` connect attempt hangs forever — no timeout on `WebSocketChannel.connect`.

The pairing flow works because it validates and stores URLs as `https://`, which the old getter handled correctly.

## Fix

1. **`RelayConfig.wsUrl`** — switch expression that passes `wss`/`ws` through unchanged and converts `https` → `wss`, `http` → `ws`.
2. **`RelayConfig.httpUrl`** (new) — inverse getter for REST endpoints. Converts `wss` → `https`, `ws` → `http`.
3. **`queryRelay` + `relayClientProvider`** — use `httpUrl` instead of raw `baseUrl` for HTTP calls (media upload, `/query` bridge).
4. **`InviteJoinNotifier.confirmJoin`** — normalize URL at storage time (`wss` → `https`, `ws` → `http`) so `Community.relayUrl` matches the pairing convention.

## Testing

- 13 new unit tests for `RelayConfig.wsUrl`/`httpUrl` scheme handling
- 2 new parameterized tests for invite URL normalization
- All 20 existing + new invite tests pass
- `flutter analyze` — no issues
- `dart format` — no changes needed

Closes #2662

PR #3243, opened 2026-07-27
fix(mobile): refetch open threads for live replies on the websocket path

### LINKED ISSUE #3046: Mobile app doesn't show threaded replies visible on Desktop
Replies sent via the Buzz CLI (`buzz messages send --reply-to <event_id>`) into a channel thread are visible in Buzz Desktop but do not appear in the mobile app, even after the parent message and thread are open.

**Steps to reproduce:**
1. Have an agent reply to a channel message using `--reply-to <event>` (creates a NIP-10 threaded reply).
2. View the channel on mobile — the reply doesn't show up.
3. View the same channel/thread on Desktop — the reply shows up correctly.

**Example event IDs** (all correctly tagged as replies, confirmed via `buzz messages get`):
- `455d40444e458ed53e3d6dd141b398736fe3f5fef3cfb0a2195d6c60a7553c6d`
- `768e6f62d4b47cd253eae0131452cbabc2e92f8d65a4373fa6b415bd2f4e17a2`
- `137841583fa64ce6a332e067cfc59a28a0b8750257281c41034797101e318ead`

Channel: `94314e5c-436c-43f6-828a-32b2ac589d3f`

ss251:
## Summary

Live thread replies never reached an open thread view on mobile's websocket history path.

`threadRepliesProvider` is a one-shot `FutureProvider.family` — an open thread only picks up a new reply when something invalidates it. That invalidation existed and was already correct, but it lived inside `_mergeWindowEventIntoStore`, which is reachable only from the `_usingChannelWindow == true` branch of `_handleLiveEvent`. The websocket fallback branch could never see it: the reply merged into the flat channel timeline and the open thread was never told.

`_fetchNewestHistory` latches that branch for the rest of the session — it catches any error from the channel-window query, logs `channel window unavailable … falling back to WS history`, and sets `_usingChannelWindow = false`. So a single failed `POST /query` is enough to put a session on the path where an open thread stops updating until the app is restarted and the one-shot future re-runs.

This moves the invalidation into `_invalidateThreadReplies(event)` and calls it from `_handleLiveEvent` ahead of the branch, so both history paths refetch.

**The channel-window path is unchanged.** `_mergeWindowEventIntoStore` has exactly one caller, `_handleWindowLiveEvent`, which calls it as its unconditional first statement; `_handleWindowLiveEvent` in turn has exactly one caller — the `_usingChannelWindow == true` branch of `_handleLiveEvent`. The invalidation ran under the same two gates it runs under now (`channelTimelineContentKinds`, `parentId != null`), so every event that reached the old invalidation reaches the new one. The only behavioral delta is the fallback branch gaining it.

Scope, stated plainly: `threadRepliesProvider` queries the same `POST /query` endpoint as the channel window, so this fixes the **transient** case — the endpoint blips, the session latches onto the fallback, the endpoint recovers, and live thread replies keep working instead of requiring a restart. Under a *persistent* `/query` outage the thread view cannot load replies with or without this patch.

### Related issue

Fixes #3046.

Duplicate search — the closest open work in this file neighbourhood, all distinct from this change:

- **#3211** (invite-link joins downgrade `wss://` to plaintext `ws://`), with PRs **#3139** and **#2726**. Genuinely adjacent: a `ws`-scheme `baseUrl` also makes `POST /query` throw, which latches this same fallback. But it is a different defect with a different fix, and neither PR touches `channel_messages_provider.dart` or `thread_replies_provider.dart`.
- **#3053** (live subscriptions — `relay_session.dart`, `channels_provider.dart`) and **#3121** (`thread_detail_page.dart`, layout and typography) — no file overlap.

Searched by issue citation and by changed-file list across all **437** open PRs (`gh pr list --repo block/buzz --state open --limit 1000 --json number,files`): none cite #3046, and none touch either file this change edits.

### Testing

Full mobile suite: **825 passed, 1 skipped.** `flutter analyze` clean; `dart format --output=none --set-exit-if-changed .` clean (283 files, 0 changed).

Added `websocket fallback refetches an open thread when a reply arrives live` to `mobile/test/features/channels/channel_messages_provider_test.dart`. It fails the channel-window query to drive the notifier onto the fallback path, opens the thread, emits a live reply over the subscription, and asserts the thread refetches.

Negative control — the same test with only the `lib/` change reverted:

```
00:00 +0 -1: websocket fallback refetches an open thread when a reply arrives live [E]
  Expected: ['reply']
    Actual: MappedListIterable<NostrEvent, String>:[]
```

An empty thread — the reported symptom. It passes with the change applied.

**Not reproduced against a live relay.** The mechanism was traced at source and is covered by the test above, but I did not observe the failure on a device against a real relay, so I cannot confirm which condition put the reporter on the fallback branch. The behavior is consistent with their follow-up that "closing and re-opening the app allows the threaded replies to be seen," since a restart re-runs the one-shot future. If there is a second path into #3046, I would want to know.

No UI change — no widget, layout, style, or string is touched — so there are no screenshots.

### DIFF
--- a/mobile/lib/features/channels/channel_messages_provider.dart
+++ b/mobile/lib/features/channels/channel_messages_provider.dart
@@ -174,6 +174,7 @@ class ChannelMessagesNotifier extends Notifier<AsyncValue<List<NostrEvent>>> {
     if (authoritative && event.threadReference.parentId == null) {
       _confirmLocalMessages([event.id]);
     }
+    _invalidateThreadReplies(event);
     if (_usingChannelWindow) {
       _handleWindowLiveEvent(event);
     } else {
@@ -198,30 +199,42 @@ class ChannelMessagesNotifier extends Notifier<AsyncValue<List<NostrEvent>>> {
     state = AsyncData(flattened);
   }
 
+  /// Refetches the open thread view for a live reply.
+  ///
+  /// [threadRepliesProvider] is a one-shot future, so an open thread only
+  /// picks up a reply when it is invalidated. This runs for both history
+  /// paths: the websocket fallback merges replies into the flat timeline
+  /// alone, which left an open thread stale until an app restart rebuilt the
+  /// future.
+  void _invalidateThreadReplies(NostrEvent event) {
+    if (!EventKind.channelTimelineContentKinds.contains(event.kind)) return;
+    final thread = event.threadReference;
+    if (thread.parentId == null) return;
+
+    final rootId = thread.rootId;
+    if (rootId != null) {
+      ref.invalidate(
+        threadRepliesProvider(
+          ThreadRepliesArgs(channelId: channelId, rootId: rootId),
+        ),
+      );
+    }
+    final parentId = thread.parentId;
+    if (parentId != null && parentId != rootId) {
+      ref.invalidate(
+        threadRepliesProvider(
+          ThreadRepliesArgs(channelId: channelId, rootId: parentId),
+        ),
+      );
+    }
+  }
+
   bool _mergeWindowEventIntoStore(NostrEvent event) {
     final isTimelineRow = EventKind.channelTimelineContentKinds.contains(
       event.kind,
     );
     final thread = isTimelineRow ? event.threadReference : null;
-    if (thread?.parentId != null) {
-      final rootId = thread?.rootId;
-      if (rootId != null) {
-        ref.invalidate(
-          threadRepliesProvider(
-            ThreadRepliesArgs(channelId: channelId, rootId: rootId),
-          ),
-        );
-      }
-      final parentId = thread?.parentId;
-      if (parentId != null && parentId != rootId) {
-        ref.invalidate(
-          threadRepliesProvider(
-            ThreadRepliesArgs(channelId: channelId, rootId: parentId),
-          ),
-        );
-      }
-      if (!_isBroadcastReply(event)) return false;
-    }
+    if (thread?.parentId != null && !_isBroadcastReply(event)) return false;
     if (!isTimelineRow &&
         !EventKind.channelAuxEventKinds.contains(event.kind)) {
       return false;
--- a/mobile/test/features/channels/channel_messages_provider_test.dart
+++ b/mobile/test/features/channels/channel_messages_provider_test.dart
@@ -423,6 +423,58 @@ void main() {
   );
 
   test(
+    'websocket fallback refetches an open thread when a reply arrives live',
+    () async {
+      final relaySession = _RecordingRelaySessionNotifier(
+        queryResults: [
+          Exception('channel window unavailable'),
+          <NostrEvent>[],
+          [
+            _event(
+              id: 'reply',
+              createdAt: 20,
+              extraTags: const [
+                ['e', 'root', '', 'reply'],
+              ],
+            ),
+          ],
+        ],
+      );
+      final container = _buildContainer(relaySession);
+      addTearDown(container.dispose);
+
+      container.read(channelMessagesProvider(_channelId));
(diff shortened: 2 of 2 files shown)


jbforge:
Read through this against `dba97eec`; every file cited below is unchanged at
current `main` (`d40a3329`). One point that may widen the case for merging it.

The description frames the latch as "a single failed `POST /query`". It is broader
than that: the fallback also latches on a **successful** query, whenever the
response does not carry exactly one kind-39006 window-bounds event.

`parseChannelWindowResponse` throws unconditionally in that case
(`mobile/lib/features/channels/channel_window.dart:131-138`):

```dart
final boundsEvents = events
    .where((event) => event.kind == EventKind.channelWindowBounds)
    .toList();
if (boundsEvents.length != 1) {
  throw Exception(
    'Channel window response must contain exactly one bounds event.',
  );
}
```

That 39006 overlay is emitted only by the bridge's `top_level` extension path
(`crates/buzz-relay/src/api/bridge.rs:396-399`). A relay build that does not
implement that extension answers the filter as an ordinary query — HTTP 200, events
returned, no bounds overlay — and the parse throws. `_fetchNewestHistory` catches it
and sets `_usingChannelWindow = false`
(`mobile/lib/features/channels/channel_messages_provider.dart:151-158`), which is
exactly the branch this PR is about.

The consequence is that on such a relay the fallback is not an intermittent
network edge case — it is the **steady state for every channel, every session**.
`build()` re-runs `_init()` on each reconnect
(`channel_messages_provider.dart:46-71`), so the window query is retried, and it
fails the same way every time. Every open thread on that deployment therefore stops
live-updating, and because `threadRepliesProvider` is a one-shot
`FutureProvider.family`, the only thing that surfaces a reply is an app restart.

This is worth separating from the scope note already in the description. Under a
persistent `/query` *outage*, `threadRepliesProvider` cannot load either — it goes
through `queryRelay` as well (`thread_replies_provider.dart:41`) — so the fix
genuinely cannot help there. The missing-extension case is different: `/query`
answers normally and only the bounds overlay is absent, so `threadRepliesProvider`
still resolves and this patch does restore live thread replies for the whole
session. So it covers a permanent case, not only the transient one.

This also produces a distinctive and, I think, useful triage signature, because the
inbox and the channel window do not share a transport:

| Surface | Path |
|---|---|
| Activity inbox | websocket `REQ` — `fetchHistory` (`mobile/lib/features/activity/activity_provider.dart` → `mobile/lib/shared/relay/relay_session.dart:182-206`) |
| Channel window | HTTP `POST /query` — `queryRelay` (`relay_session.dart:131-136`) |

So **a fresh inbox alongside a stale channel/thread is positive evidence the
websocket is healthy** and the HTTP window path is what failed. That distinguishes
this bug from the half-open-socket class in #3224, where the websocket is the thing
that died and the inbox would be stale too. Reports of "I have to force-quit to see
replies" seem to be splitting across both, and this asymmetry separates them without
needing device logs.

When device logs are available, the direct confirmation is the existing debug line:

```
[ChannelMessagesNotifier] channel window unavailable for <channelId>, falling back to WS history: <error>
```

Two suggestions, both optional and neither blocking this PR:

1. That `debugPrint` is the only signal that a session has latched onto the fallback,
   and it is invisible in a release build. Some non-debug surface for "this relay does
   not support the channel-window extension" would turn a silent permanent degradation
   into something diagnosable.
2. If the intent is that older relays degrade gracefully, the missing-bounds case may
   warrant distinguishing "relay lacks the extension" from "query failed" — the first
   is permanent and knowable up front, the second is transient and worth retrying.

Neither changes the fix here, which is correct as written: moving the invalidation
ahead of the branch is right regardless of why the fallback engaged.

PR #3246, opened 2026-07-28
fix(composer): scope multiline block formatting

tellaho:
**Category:** fix
**User Impact:** Composer block formatting now applies to the intended line or selection without collapsing multiline content.

**Problem:** Block formatting from a Shift+Enter line could convert the entire draft, selected visual lines could collapse into one list item, and code conversion could lose line breaks. **Solution:** Scope caret formatting to its hard-break-delimited line and normalize explicit selections for the destination block type while preserving neighboring content and visual line boundaries.

<details>

**desktop/src/features/messages/lib/selectionBlockFormatting.ts**
Scopes collapsed-caret block actions to the active visual line and normalizes multiline selections for lists and code blocks.

**desktop/src/features/messages/lib/selectionBlockFormatting.test.mjs**
Adds unit coverage for caret-line isolation across line positions and selection directions.

**desktop/src/features/messages/ui/FormattingToolbar.tsx**
Routes list, quote, and code-block actions through the selection-aware formatting transaction.

**desktop/tests/e2e/composer-selection-formatting.spec.ts**
Covers caret-only formatting, multiline list conversion, list-to-code conversion, preserved hard breaks, Markdown output, and backward selections.

</details>

## Reproduction steps

1. In the desktop composer, enter several lines using Shift+Enter and place the caret on one line.
2. Apply a bullet list, ordered list, quote, or code block; only the caret line should change.
3. Select several Shift+Enter lines and apply a list; each visual line should become its own item.
4. Select several list items and apply Code block; they should become one multiline code block while unselected neighbors remain intact.
5. Select several Shift+Enter lines and apply Code block; each line break should remain visible.

## Screenshots/Demos
Screen Recording 2026-07-27 at 5 29 19 PM

Expected multiline code-block result: https://buzz.block.builderlab.xyz/media/d2e2668093af3b67d896a32e9799daccd236da9fc9e24ec56ddb4ebf7d01dd96.png

### DIFF
--- a/desktop/src/features/messages/lib/selectionBlockFormatting.ts
+++ b/desktop/src/features/messages/lib/selectionBlockFormatting.ts
@@ -1,3 +1,4 @@
+import type { Node as ProseMirrorNode } from "@tiptap/pm/model";
 import { TextSelection, type Transaction } from "@tiptap/pm/state";
 import { canSplit } from "@tiptap/pm/transform";
 
@@ -146,6 +147,96 @@ export function isolateSelectionForBlockFormatting(
       isBackward ? to : from,
       isBackward ? from : to,
     ),
+  );
+  return true;
+}
+
+/** Split each selected hard-break line into a textblock before list wrapping. */
+export function splitSelectedLinesForListFormatting(
+  transaction: Transaction,
+): boolean {
+  if (!(transaction.selection instanceof TextSelection)) return false;
+  if (transaction.selection.empty) {
+    return isolateCaretLineForBlockFormatting(transaction);
+  }
+
+  const isBackward = transaction.selection.anchor > transaction.selection.head;
+  isolateSelectionForBlockFormatting(transaction);
+  let { from, to } = transaction.selection;
+  const breakPositions: number[] = [];
+
+  transaction.doc.nodesBetween(from, to, (node, position) => {
+    if (node.type.name === "hardBreak") breakPositions.push(position);
+  });
+
+  for (const position of breakPositions.reverse()) {
+    transaction.delete(position, position + 1);
+    ({ from, to } = mapRangeThroughLatestStep(transaction, from, to));
+    if (!canSplit(transaction.doc, position)) continue;
+    transaction.split(position);
+    ({ from, to } = mapRangeThroughLatestStep(transaction, from, to));
+  }
+
+  transaction.setSelection(
+    TextSelection.create(
+      transaction.doc,
+      isBackward ? to : from,
+      isBackward ? from : to,
+    ),
+  );
+  return true;
+}
+
+function selectedTextblocks(
+  transaction: Transaction,
+): Array<{ node: ProseMirrorNode; position: number }> {
+  const blocks: Array<{ node: ProseMirrorNode; position: number }> = [];
+  const { from, to } = transaction.selection;
+  transaction.doc.nodesBetween(
+    Math.max(0, from - 1),
+    Math.min(transaction.doc.content.size, to + 1),
+    (node, position) => {
+      if (node.isTextblock) {
+        blocks.push({ node, position });
+        return false;
+      }
+      return true;
+    },
+  );
+  return blocks;
+}
+
+function textblockTextForCode(node: ProseMirrorNode): string {
+  return node.textBetween(0, node.content.size, "\n", (leaf) =>
+    leaf.type.name === "hardBreak"
+      ? "\n"
+      : (leaf.type.spec.leafText?.(leaf) ?? ""),
+  );
+}
+
+/** Replace selected textblocks with one newline-joined code block. */
+export function mergeSelectedTextblocksIntoCodeBlock(
+  transaction: Transaction,
+): boolean {
+  if (!(transaction.selection instanceof TextSelection)) return false;
+  if (transaction.selection.empty) {
+    isolateCaretLineForBlockFormatting(transaction);
+    return false;
+  }
+
+  const blocks = selectedTextblocks(transaction);
+  const codeBlock = transaction.doc.type.schema.nodes.codeBlock;
+  const first = blocks[0];
+  const last = blocks.at(-1);
+  if (!(codeBlock && first && last)) return false;
+
+  const text = blocks.map(({ node }) => textblockTextForCode(node)).join("\n");
+  const from = first.position;
+  const to = last.position + last.node.nodeSize;
+  const content = text ? transaction.doc.type.schema.text(text) : undefined;
+  transaction.replaceWith(from, to, codeBlock.create(null, content));
+  transaction.setSelection(
+    TextSelection.create(transaction.doc, from + 1, from + 1 + text.length),
   );
   return true;
 }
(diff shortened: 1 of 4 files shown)


desktop/src/features/messages/lib/selectionBlockFormatting.ts
```diff
@@ -84,3 +150,93 @@ export function isolateSelectionForBlockFormatting(
...
+}
+
+function selectedTextblocks(
+  transaction: Transaction,
+): Array<{ node: ProseMirrorNode; position: number }> {
+  const blocks: Array<{ node: ProseMirrorNode; position: number }> = [];
+  const { from, to } = transaction.selection;
+  transaction.doc.nodesBetween(
+    Math.max(0, from - 1),
+    Math.min(transaction.doc.content.size, to + 1),
```
wpfleger96: 🤖 I think this ±1 padding, combined with `canSplitInsideTextblock` requiring `parentOffset > 0 && < content.size`, lets whole unselected blocks leak into the conversion.

Repro (Chromium): with `<p>alpha</p><p>beta</p><p>gamma</p>`, drag from the visual end of `alpha` to the visual start of `gamma`. The selection endpoints land exactly on block boundaries (`from` at alpha's end, `to` at gamma's start), so neither endpoint gets split out — and then the `from - 1` / `to + 1` window here pulls `alpha` and `gamma` into the merge. Result: all three paragraphs end up in the code fence (and the bullet/quote paths similarly reformat all three) when only `beta` was visibly selected.

Could we normalize boundary endpoints inward before the block commands (drop a start block when the selection begins at its end, and an end block when it ends at its start), and collect blocks from the exact normalized range instead of padding by one? A mouse-drag e2e for this boundary shape (plus its backward equivalent) would lock it in.

desktop/src/features/messages/ui/FormattingToolbar.tsx
```diff
@@ -196,14 +200,35 @@ export const FormattingToolbar = React.memo(function FormattingToolbar({
...
+      ? Array.from(
+          { length: selection.$from.depth + 1 },
+          (_, depth) => selection.$from.node(depth).type.name,
+        ).some((name) => name === "bulletList" || name === "orderedList")
+      : false;
+    let conversion = chain.command(({ tr }) => {
+      isolateSelectionForBlockFormatting(tr);
+      return true;
+    });
+    if (isInList) conversion = conversion.clearNodes();
```
wpfleger96: 🤖 `isInList` is computed from `editor.state.selection` (the pre-chain state), but `clearNodes()` runs *after* the isolation command has already mutated the doc — so it traverses stale coordinates. Selecting from the middle of one list item into the middle of the next throws an uncaught `RangeError: Position 43 out of range` in our repro; the list-toggle path doesn't throw on the same shape but rewrites two list items into one item containing four paragraphs, changing unselected structure.

Also, since `isInList` only checks `$from`, a selection that starts outside a list and ends inside one skips `clearNodes()` entirely.

The e2e list-to-code case selects whole item texts only, so it doesn't catch this. I'd either define block formatting in lists as whole-item-only (snap/expand the selection before mutating) or explicitly split/lift list items with mapped positions — but not generic textblock splitting followed by `clearNodes()` over stale state. Partial-first/partial-last list-item coverage for code/bullet/ordered/quote would be great.

desktop/src/features/messages/ui/FormattingToolbar.tsx
```diff
@@ -196,14 +200,35 @@ export const FormattingToolbar = React.memo(function FormattingToolbar({
...
   const toggleCodeBlock = React.useCallback(() => {
-    formattingChain()
-      ?.command(({ tr }) => {
-        isolateSelectionForBlockFormatting(tr);
-        return true;
-      })
-      .toggleCodeBlock()
+    const chain = formattingChain();
+    if (!chain) return;
+    if (editor?.state.selection.empty) {
```
wpfleger96: 🤖 This branch reads the *live* editor selection, but `formattingChain()` exists precisely because WKWebView can collapse the selection when focus moves to the toolbar — the pending captured range is only restored inside the chain. On that collapse path (which `collapseAfterMouseDown` simulates elsewhere in the e2e suite), a multiline code-block conversion takes this caret branch instead: no merge, no `clearNodes()`, each line becomes its own code block. Same issue for the `isInList` computation below.

I'd decide the branch from the pending captured range (or inside the transaction after the restore), and add a collapse-simulation e2e for the code-block merge path — none of the new tests exercise `collapseAfterMouseDown` with it.

Minor related note: `mergeSelectedTextblocksIntoCodeBlock`'s empty-selection arm is unreachable from here since this branch handles empty selections first — either route through it or drop it.

wpfleger96 (changes requested):
🤖 Thanks for the fix — the caret-line isolation is nicely done (first/middle/last/empty trailing lines all covered, backward selections restored, single undo unit) and the serialized-Markdown e2e assertions are the right end-to-end check.

That said, we found two selection shapes where unselected content gets structurally reformatted, plus a branch that I think defeats the toolbar's own WKWebView selection workaround, so requesting changes. Details inline on the relevant lines.

A couple of non-blocking notes while you're in here:

- `selectionBlockFormatting.test.mjs` only covers `isolateSelectionForBlockFormatting` — `splitSelectedLinesForListFormatting` and `mergeSelectedTextblocksIntoCodeBlock` are e2e-only. They're pure transaction logic so unit cases should be cheap to add.
- `textblockTextForCode()` drops any leaf node that doesn't implement `leafText`, so mentions/custom emoji could silently disappear when converted to code. Probably fine as a contract, but worth a unit case or a comment saying it's intentional.

desktop/src/features/messages/lib/selectionBlockFormatting.ts
tellaho: Fixed in `6ab556e16e504b2a49bfecfe9c7783dd9ee4c9e2`. Boundary endpoints are normalized inward and block collection now uses the exact normalized range; forward and backward regression coverage confirms unselected endpoint blocks stay out. — AI-generated by Ned

desktop/src/features/messages/ui/FormattingToolbar.tsx
tellaho: Fixed in `c71436576d6b74fac9fc7f90f727164cd09cae97`. Partial list selections now expand to whole items before mutation, selected items are safely lifted, and list detection scans the restored transaction range, including outside→inside selections. Coverage exercises code, bullet, ordered, and quote behavior with neighbors preserved. — AI-generated by Ned

desktop/src/features/messages/ui/FormattingToolbar.tsx
tellaho: Fixed in `65d50521069302c370059cad91da4ef58ca3e052`. Code-block branching now reads the restored transaction selection inside the formatting command; collapse-after-mousedown multiline conversion is covered, and the unreachable empty-selection arm was removed. — AI-generated by Ned

tellaho:
@wpfleger96 We also closed out both non-blocking notes from your review:

- `21987a8d2fb772be2580933f53762b8f6dc0d925` preserves meaningful inline-atom text during code-block conversion, documents the fallback semantics, and adds focused unit coverage for schema `leafText`, mention labels, custom-emoji shortcodes, hard breaks, and safe dropping of unknown leaves.
- `fb91387616f93e38eda66f7e6c410f255614195f` tightens the partial-list Playwright regression to assert exact top-level structure, item contents, and preserved neighbors for code, bullet, ordered, and quote outcomes.

Validated with 3,731 desktop unit tests, typecheck/check, and focused composer Playwright 24/24.

— AI-generated response on behalf of Taylor Ho

wpfleger96 (approved):

PR #3262, opened 2026-07-28
refactor(buzz-agent): swap the agent loop onto the goose library

michaelneale:
## What changed

Previously, Buzz implemented the agent stack itself:

```rust
loop {
    let response = buzz_provider.complete(&history, &tools).await?;
    history.push(response);

    if response.tool_calls.is_empty() {
        return stop_or_continue_via_buzz_hooks();
    }

    history.push(buzz_mcp.execute(response.tool_calls).await);
}
```

Buzz owned provider adapters, messages, MCP, tool execution, compaction, and loop policy.

Now, Buzz owns orchestration and Buzz-specific policy while composing [Goose GDK primitives](https://goose-docs.ai/docs/gdk/api-reference/?version=0.1&language=rust):

```rust
loop {
    apply(buzz_round_start.step(&conversation).await?); // steer, compact

    match goose_provider.stream(&conversation, &goose_tools).await {
        Ok(response) if response.has_tool_calls() =>
            conversation.push(buzz_rmcp.dispatch_bounded(response).await),
        Ok(response) =>
            return buzz_end_turn_policy(response).await,
        Err(ContextOverflow) =>
            apply(buzz_compaction.force(&conversation).await?),
    }
}
```

Buzz-specific policy is expressed as small Goose operations:

```rust
round_start = [Steer, BuzzCompaction];
round_gate  = [MaxRounds, StopVeto, ReplyGuard];
```

Buzz still owns when to compact, overflow recovery and retry, conversation replacement, and todo restoration. Goose supplies the context-management primitives:

```rust
if buzz_policy.should_compact(usage, context_limit) || provider_overflowed {
    let summary = goose_context::summarize(visible_messages).await?;
    conversation = buzz_replace_with_summary(summary);
    restore_buzz_todos().await;
}
```

**In one sentence:** previously Buzz built an agent stack; now it composes Goose primitives while retaining the behavior that makes an agent a good Buzz participant.

This is intended to preserve Buzz agent behavior while removing duplicated infrastructure and making future improvements easier.

Skills use Goose’s discovery and rendering APIs but remain selected by Buzz: Buzz can choose filesystem roots or source entries per agent, advertise only a small index, and load the chosen skill body on demand.

## Presentation

[BUZZ_GDK_BRIEF.pdf](https://github.com/user-attachments/files/31490685/BUZZ_GDK_BRIEF.pdf)

Buzz GDK overview

size impact: 

image


michaelneale:
🤖 Review note from MicBlock's AI agent (Galadriel) — re-review at head `2ea414f0b`.

**Verdict:** the substantive regressions from the first review are fixed (diff-confirmed): compaction now uses each response's own occupancy total; usage/pricing emission is routed back through `wire::usage_update_payload` with separate cache read/write accounting, sticky-unknown totals, and origin-allowlisted `pricingIdentity`; cancel mid-inference returns `cancelled` and discards the partial; reply guard rejects hallucinated tool names via `list_tools()`; `MAX_TOOL_CALLS_PER_TURN=64` restored; provider errors classified from typed `ProviderError`; Databricks 401 forced refresh wired via `refresh_hook`.

**Before this leaves draft:**

1. **Desktop lockfile downgrades** — `desktop/src-tauri/Cargo.lock` still carries `anyhow 1.0.103` and `nostr 0.44.6` vs main's `1.0.104` / `0.44.7`. Nothing in this PR requires them; restore with `cargo update -p anyhow -p nostr` against the desktop manifest. (Root-lockfile icu 1.5.x moves are justified — forced by goose's `idna_adapter = 1.2.0` pin, verified with `cargo tree -i`.)
2. **Multi-round compaction test** — the fix is correct by inspection, but the only compaction test exercises a single effect; the original bug was invisible to exactly that shape. Add a two-round test asserting occupancy doesn't accumulate.
3. **CI** — `tooltip-semantics.spec.ts` fails identically on current main (pre-existing); `thread-reply-anchor-roleplay` failed only here — rerun to confirm flake.

Smaller residuals to record as decisions (not merge-gates): steer queue not cleared on cancel/error return paths (stale steer can replay into the next turn); desktop env-strip ordering vs baked build env / `default_env`; mid-turn `usage_update` emission now end-of-turn only.

Full review discussion in the #feature-buzz-with-gdk channel.

micspiral:
## Test-parity review: three specific gaps in loop bounding

Read at `origin/main` `e8172b5b` vs this branch `0a4c78e1e`. Everything below is a `git grep` result on those two trees, not an inference.

Framing first, because the headline numbers are misleading: **612 tests → 198** looks alarming but most of it is correct. `llm.rs` (186 tests) tested a request transport that no longer exists; ~38 of `config.rs`'s 100 were `anthropic_thinking_config_*` per-model permutations for a request buzz no longer builds; the 13 handoff tests covered a summariser goose now owns. `permission.rs` (21) and `wire.rs` (18) carried over intact, and the Databricks OAuth tests moved to `buzz-model-catalog` (50). Those are not gaps and I am not asking for them back.

Three things I do think are gaps, all in **loop bounding** — the part that decides when a turn stops.

### 1. `BUZZ_AGENT_MAX_ROUNDS` unset changed meaning, and the README still documents the old one

- main: `max_rounds: parse_env("BUZZ_AGENT_MAX_ROUNDS", 0)`, and `agent.rs:362` guards `if self.cfg.max_rounds > 0 && round >= …` — unset means **unbounded**.
- branch: `config.rs:223` `env_parse::<u32>(…).filter(|n| *n > 0)` → `None`, and `loop_drive.rs:172` `ctx.max_rounds.unwrap_or(DEFAULT_MAX_ROUNDS)` where `DEFAULT_MAX_ROUNDS = 1000` (`loop_drive.rs:88`) — unset means **1000**.

The change is deliberate and the reasoning in the doc comment is good ("owning the loop means owning the bound"). Two problems:

- `README.md:334` still says `| Loop rounds | 0 (unlimited) | BUZZ_AGENT_MAX_ROUNDS |`, and `README.md:155` still says `0 = unlimited`. Both now describe main's behaviour.
- `DEFAULT_MAX_ROUNDS` has **two references in the whole branch** — its definition and its one use. No test pins it. Main had `finite_round_cap_still_binds_without_a_context_overflow` and `max_tokens_recovery_respects_finite_round_cap`.

**To close:** fix the two README rows to say 1000, and add one test asserting an unset `max_rounds` stops at 1000 with `StopReason::MaxTurnRequests`. `ops.rs`'s `at_budget_ends_the_turn_with_max_turn_requests` is the pattern; it just needs the default-path case.

### 2. The documented ordering between the tool-call cap and the reply guard is untested

Correcting my own earlier count: reply-guard coverage is **10 → 5**, not 10 → 1. `ops.rs` has `the_reply_guard_reminds_when_nothing_was_published`, `a_publish_attempt_disarms_the_guard`, `an_unrelated_tool_call_does_not_disarm_the_guard`, `a_hallucinated_shell_does_not_disarm_the_guard`, `the_guard_stops_after_its_budget`. That is the core of the guard and it is well covered.

Four of main's cases have no counterpart:

| main test | branch |
|---|---|
| `reply_guard_ignores_calls_lost_to_the_turn_cap` | none |
| `reply_guard_bounded_by_stop_rejection_budget` | none |
| `reply_guard_off_when_stop_budget_is_zero` | none |
| `reply_guard_combines_with_stop_hook_objection` | none |

The first is the one worth adding, because **`README.md:217-219` states the property as a guarantee**: *"Detection is checked after the per-turn tool-call cap (`MAX_TOOL_CALLS_PER_TURN`) is applied: a publish-shaped call that was discarded never ran."* `loop_drive.rs:408-415` truncates, `BuzzReplyGuardOperation` inspects the conversation afterwards — so the behaviour looks right, but nothing tests it, and it is now an emergent property of two separate components rather than one function's control flow. The other three are guard × stop-hook budget interactions; both mechanisms exist on this branch and both have tests, but nothing tests them together.

**To close:** one test with >64 publish-shaped calls in a round asserting the guard still nags, plus one guard-under-exhausted-stop-budget test.

### 3. Six documented caps have no implementation on this branch

`README.md:325-332` lists these. Grepping the **entire branch** for each name returns exactly one hit — the README row itself:

| cap | main implementation | branch |
|---|---|---|
| `MAX_MCP_SERVERS` (16) | `mcp.rs:26,207-209` rejects | README only |
| `MAX_TOOLS_PER_SESSION` (128) | `mcp.rs:22,262-264` rejects | README only |
| `MAX_DESCRIPTION_BYTES` (1 KiB) | `mcp.rs:23,286` clamps | README only |
| `MAX_SCHEMA_BYTES` (4 KiB) | `mcp.rs:24,880-884` replaces with `{}` | README only |
| `MAX_TOOL_RESULT_BYTES` (8 MiB) | `config.rs:384`, `agent.rs:891` | README only |
| `MAX_LLM_RESPONSE_BYTES` (16 MiB) | `llm.rs:23,1966` | README only |

`MAX_LLM_RESPONSE_BYTES` is fine to drop — that is transport, goose's now. The other five are **MCP-facing** limits, and MCP servers are the untrusted input here: `MAX_SCHEMA_BYTES` and `MAX_DESCRIPTION_BYTES` existed so one server's oversized tool definitions could not crowd out the prompt.

`MAX_TOOL_RESULT_BYTES` is the one I would check first. `BUZZ_AGENT_MAX_TOOL_RESULT_TEXT_BYTES` is projected onto goose's `GOOSE_MAX_TOOL_RESPONSE_SIZE` (`config.rs:347-348`), so tool result **text** is bounded — but main's 8 MiB was the **total including images**, and I found nothing on this branch enforcing a total. Mitigating: `lib.rs:549` declares `promptCapabilities.image: false`, so images cannot arrive via the prompt; the unbounded path would be an MCP tool returning image content.

**To close:** for each of the five, either cite the goose mechanism that enforces it (and change the README row to name that mechanism) or delete the row. A documented limit with no enforcement is worse than no limit, because a reviewer reads the table and concludes it is handled.

### Also worth listing explicitly in the PR body

`permission_boundary.rs` inverts a main assertion: `test_load_skill_emits_no_permission_request` (main, asserts zero requests) → `test_load_skill_is_gated_like_any_other_tool` (branch, asserts one). The doc comment argues it well — exempting `load_skill` would need a name-based allowlist and tool names are attacker-controlled — and I agree with the call. But it is a user-visible behaviour change and the PR body does not mention it. **Any test whose assertion flipped should be listed there**, because both suites are green while asserting opposite things, so review cannot see it from CI.

### One process note

`scripts/run-tests.sh:121-122` runs `cargo test -p buzz-agent --lib` — unit tests only, on both main and this branch. The 10 integration files in `tests/` (`stdio_turn.rs`, `provider_faults.rs`, `steer.rs`, `cancel.rs`, `stop_hook.rs`, `skills.rs`, `real_dev_mcp.rs`, `permission_boundary.rs`, `memory.rs`, `reflect.rs`) are the best evidence in this PR — they drive the real binary over stdio against a live socket provider — and CI is not running them. Green CI is not evidence they pass. Adding them to that script would make this PR's strongest evidence load-bearing.

---

I have not run either suite; disk and toolchain constraints on this machine meant I read both trees rather than executing them. Every claim above is a grep on `e8172b5b` / `0a4c78e1e` with the file and line cited, so each is cheap to falsify — please do.

crates/buzz-agent/src/loop_drive.rs
```diff
@@ -0,0 +1,956 @@
...
+    let tools = ctx.mcp.rmcp_tools();
+
+    let mut stream = provider
+        .stream(
+            &model_config,
+            &system_prompt,
+            conversation.messages(),
+            &tools,
+        )
+        .await
```
salman1993: **P1 — Make stream construction cancellable, not just stream consumption.**

`Provider::stream()` does not immediately return a lazy stream: the pinned Goose OpenAI implementation awaits the HTTP POST (including retries/response headers) before returning. This await is outside both the cancellation select and keepalive ticker below. A queued/stalled gateway therefore prevents cancellation until headers arrive or the provider timeout expires, and the harness sees no keepalive during that wait. Main wrapped the entire LLM call in the cancellation select.

Reproduced with the actual ACP binary: accept the completion POST but withhold HTTP response headers; send `session/cancel`; no prompt response within 3 seconds. Releasing headers immediately produces `stopReason: cancelled`. Please cover the stream-construction future with cancellation/keepalive and add a regression that stalls **before headers**, not only between SSE chunks.

crates/buzz-agent/src/loop_drive.rs
```diff
@@ -0,0 +1,956 @@
...
+        // Steering and compaction, both at the round boundary and never
+        // mid-inference where they would race a partially streamed response.
+        // Run to exhaustion: `step` stops at the first operation that applies,
+        // so a turn that both steers and compacts needs more than one pass.
+        loop {
+            let Gate::Applied(effects) = gate(&start_machine, state.session(), &emitter).await
+            else {
+                break;
+            };
+            if !apply_effects(&mut state, effects) {
```
salman1993, line 249: **P1 — Bound proactive compaction before rerunning the start machine.**

Every successful compaction returns a replacement, so `apply_effects` reports progress and clears total_tokens. But `compacted_conversation` preserves the latest user text verbatim. If that text alone exceeds the byte-estimated threshold, the next start-machine step compacts it again indefinitely. Neither the outer max-rounds gate nor the reactive context-recovery counter bounds this inner loop.

Reproduced with `GOOSE_CONTEXT_LIMIT=1000`, `BUZZ_AGENT_MAX_ROUNDS=1`, a 2000-byte user prompt, and a provider returning a short valid summary: 8 consecutive summary requests, zero normal inference requests, and no prompt completion before I explicitly cancelled. A larger prompt exceeding the threshold of a normal-sized context produces the same condition. Main capped proactive handoff attempts and then used truncation. Please enforce a proactive compaction budget/no-progress policy so repeated successful summaries cannot run up unbounded requests without advancing the task.

salman1993 (changes requested):
Reviewed at `d2cc2e1779161b4d3a6e920952ea81a5ef2a8ebb`. Two reproducible agent-loop regressions need fixing before approval.

Validation: built buzz-dev-mcp, then ran both whole package suites (`cargo test -p buzz-agent -p buzz-model-catalog`): 277 passed, 1 ignored across targets. Additional probes used the real buzz-agent binary over ACP stdio with an isolated local fake HTTP provider and temp HOME/XDG directories. No live-relay or real-provider pass is claimed.

Review by Leo (AI agent).

crates/buzz-agent/README.md
```diff
@@ -153,21 +153,34 @@ Everything is environment variables. No flags, no config files. (We are a subpro
...
-| `BUZZ_AGENT_MAX_TOKEN_RECOVERIES` | `3` | Retries after a successful response is truncated at the output-token limit. `0` disables recovery; the finite value and `BUZZ_AGENT_MAX_ROUNDS` prevent infinite retries. |
+| `BUZZ_AGENT_MAX_TOKEN_RECOVERIES` | — | **No longer read.** buzz's own truncation-recovery loop went with its request transport; goose owns retries. |
 | `BUZZ_AGENT_MAX_CONTEXT_TOKENS` | `200000` | Provider context window used by the handoff gate. |
-| `BUZZ_AGENT_MAX_HANDOFFS` | `10` | Max context handoffs per session before falling back to truncation. |
-| `BUZZ_AGENT_LLM_TIMEOUT_SECS` | `240` | Max seconds with no response bytes before abandoning an LLM call (per-read inactivity, not wall-clock). |
-| `BUZZ_AGENT_TOOL_TIMEOUT_SECS` | `1260` | Per-tool call timeout in seconds |
-| `BUZZ_AGENT_MAX_PARALLEL_TOOLS` | `8` | Max concurrent tool calls per turn (1 = sequential) |
+| `BUZZ_AGENT_MAX_HANDOFFS` | — | **No longer read.** goose's compaction replaced the handoff mechanism it bounded. |
```
salman1993: 🤖 Flagging several high-risk environment-variable changes that appear to remove safety controls without equivalent replacements:

- `BUZZ_AGENT_MAX_HANDOFFS`: removed, but it previously bounded compaction attempts and fell back to truncation. Without an equivalent cap/progress check, compaction can loop indefinitely.
- `BUZZ_AGENT_MAX_HISTORY_BYTES`: removed, while compaction only hides old messages from the model rather than removing them from memory. Long-lived sessions may grow without bound.
- `BUZZ_AGENT_MAX_TOKEN_RECOVERIES`: removed, and it is unclear whether truncated model responses still get safely continued with a bounded recovery budget.

crates/buzz-agent/src/loop_drive.rs
michaelneale: 🤖 Reply from MicBlock's AI agent (Galadriel).

Fixed in 7a81512ee. `Provider::stream()`'s construction await is now pinned and covered by the same `tokio::select!` as chunk consumption — `session/cancel` interrupts it immediately and keepalives keep ticking during the pre-headers wait (`loop_drive.rs::infer`).

Regression test `cancel_during_header_stall_returns_promptly` (tests/cancel.rs) does exactly your repro: a provider that accepts the completion POST but parks the socket without response headers, `BUZZ_AGENT_LLM_TIMEOUT_SECS=300` so a prompt response can only come from cancellation, then asserts `stopReason: cancelled` in <5s. Passes; full `cargo test -p buzz-agent` green.

crates/buzz-agent/src/loop_drive.rs
michaelneale: 🤖 Reply from MicBlock's AI agent (Galadriel).

Fixed in 7a81512ee. Proactive compaction now has a no-progress budget of **one per inference**: `BuzzCompactionOperation` tracks `compacted_since_inference` (shared across the loop's forced path and the start-machine gate); a second proactive trip before any inference is refused and the request goes to the provider, whose acceptance or context-400 is authoritative — genuine overflow then takes the existing bounded reactive path (3 per turn). Forced/reactive compaction also consumes the budget, and each successful inference re-arms it.

Regression test `oversized_prompt_does_not_spin_proactive_compaction` (tests/provider_faults.rs) uses your exact repro — `GOOSE_CONTEXT_LIMIT=1000`, `BUZZ_AGENT_MAX_ROUNDS=1`, 2000-byte prompt, short valid summaries — and asserts the turn completes with ≤4 provider requests (pre-fix: 8+ summaries, zero inference, no completion). Passes.

crates/buzz-agent/README.md
michaelneale: 🤖 Reply from MicBlock's AI agent (Galadriel).

All three addressed in 7a81512ee, taking each variable on its merits:

- **`BUZZ_AGENT_MAX_HANDOFFS`** — stays removed, but the unbounded-compaction risk it guarded is now closed structurally: at most one proactive compaction per inference (no-progress guard in `ops.rs`) plus the existing cap of 3 reactive compactions per turn on provider context-overflow. See the P1 thread above for the regression test. README row now documents the replacement bound instead of just "no longer read".

- **`BUZZ_AGENT_MAX_HISTORY_BYTES`** — **restored** (default 16 MiB, values <1 MiB ignored). Your diagnosis was right: goose compaction only marks messages agent-invisible, it deletes nothing, so carried-forward history grew without bound. `turn_state::evict_hidden_history` now evicts the oldest **agent-invisible** messages once the stored history exceeds the budget; model-visible messages are never evicted (their footprint is what token compaction bounds). Unit-tested.

- **`BUZZ_AGENT_MAX_TOKEN_RECOVERIES`** — **restored** (default 3, `0` disables), because goose does *not* continue truncated responses itself: it marks them (`output_token_limit_reached`) and leaves the policy to the embedder. Behaviour: tool calls from a truncated response are discarded and never executed (a later call may have been cut off), surviving text is preserved, an agent-only continue nudge is injected, and exhausting the budget surfaces as `stopReason: max_tokens`. Two integration tests pin recovery-then-completion and the bounded persistent-truncation case.

Full `cargo test -p buzz-agent` green (17 suites, 0 failures) at 7a81512ee.

michaelneale:
@salman1993 I guess we can close this as at the moment isn't the direction we will go?

PR #3275, opened 2026-07-28
feat(desktop): host MCP Apps as channel tabs

tmarman:
## Summary

This draft lets a Buzz channel host task-specific interfaces as tabs through
[MCP Apps](https://modelcontextprotocol.io/extensions/apps/overview).

The desktop client can:

- connect to a Streamable HTTP MCP server;
- inspect its advertised tools, UI resources, network policy, and permissions;
- install selected Apps as channel tabs for the current local identity;
- render each App inside a nested sandbox;
- route App-visible `tools/call` requests to the MCP server; and
- let an App request a channel post through `ui/message`.

Buzz shows the exact attributed message before it publishes the event. The App
does not receive the user's Nostr key or relay credentials. This change does not
modify the relay.

### Interaction model

```mermaid
flowchart LR
    U["User"]
    C["Buzz channel shell"]
    P["Buzz sandbox proxy"]
    A["MCP App UI<br/>opaque origin"]
    H["Buzz MCP host"]
    S["Streamable HTTP<br/>MCP server"]
    R["Buzz relay"]

    U -->|"install / select / approve"| C
    C --> P
    P -->|"validated postMessage"| A
    A -->|"tools/call"| P
    P -->|"validated postMessage"| C
    C --> H
    H -->|"host-owned context + MCP request"| S
    S -->|"tool result"| H
    H --> C
    C --> P
    P --> A
    A -->|"ui/message request"| P
    P --> C
    C -->|"exact-message approval"| U
    C -->|"signed event after approval"| R
```

Buzz removes caller-supplied values under the complete
`params._meta["xyz.block.buzz/*"]` namespace before a tool call. It then adds
the current host-owned context value. That value can contain community,
channel, and App-installation references. These references provide routing
context. They do not grant authority.

### Protocol compatibility

- Supports the current MCP Apps Streamable HTTP flow.
- Falls back to the 2025-11-25 MCP handshake when a server does not return a
  recognized modern response.
- Keeps required MCP request metadata separate from App arguments.
- Preserves multi-line SSE events and ignores notifications until the matching
  JSON-RPC response arrives.
- Treats tools without explicit visibility as App-visible for compatibility
  with existing MCP servers.

### Trust boundaries

- Allows HTTPS endpoints and explicit loopback HTTP endpoints.
- Rejects endpoint credentials, URL fragments, mDNS names, public endpoints
  that resolve to private or reserved addresses, and IPv4 transition forms.
- Pins the validated endpoint address, disables redirects, and bypasses
  environment or system HTTP proxies.
- Limits response bodies before parsing.
- Shows the sanitized resource policy during installation.
- Rejects a render-time policy that is broader than the reviewed policy.
- Uses a host-owned custom-protocol proxy and an opaque inner frame. The inner
  frame receives `allow-scripts`; it does not receive `allow-same-origin`.
- Keeps server-authored labels, errors, and channel-post content bounded and
  safe for trusted Buzz chrome.
- Requires explicit approval for the exact attributed channel message.
- Keeps an active App mounted when another tab is installed. The initial host
  tool call therefore runs once for unchanged App configuration.

### Manual flow

1. Open a channel and select **Apps**.
2. Connect to a Streamable HTTP MCP endpoint.
3. Review the App's requested network access and permissions.
4. Add an advertised UI resource as a channel tab.
5. Use the App to call an App-visible MCP tool.
6. Let the App request a channel post.
7. Review and approve the exact message.
8. Return to **Chat** and confirm that one attributed message was published.

### Screenshots

The review assets are neutral fixtures. They contain no personal names,
communities, private runtime names, or real endpoints.

#### Before installation and policy review

| Before installation | Policy review |
| --- | --- |
| Buzz channel before an MCP App is installed | Review MCP App network access and browser permissions |

#### Task-specific channel interfaces

| Project board | Signal reader |
| --- | --- |
| Project board MCP App installed beside Chat | Signal reader MCP App installed beside Chat |

| Artifact review | Review channel post |
| --- | --- |
| Artifact review MCP App installed beside Chat | Review the attributed MCP App message before posting |

The executable flow also captures the one attributed result after publication.

### Current limits

- Streamable HTTP only. This draft does not add stdio process setup because
  executable selection and process credentials require a separate trust model.
- Server authorization and OAuth are not implemented. A complete flow needs
  protected-resource discovery, PKCE, token refresh, revocation, and secure
  storage.
- Installations are local to one Buzz identity and device.
- The host advertises inline display only.
- Host context currently contains community, channel, and installation
  references. It does not define thread, project, or external Space references.
- Host context is metadata, not authorization.
- App-visible tool calls do not have per-call approval. Channel publication
  through `ui/message` does.
- A hostname in an App's reviewed CSP is not DNS-pinned when the webview later
  fetches it.
- App metadata has no Nostr provenance.
- Interrupted SSE requests do not resume with `Last-Event-ID`.
- A failed publish can be retried by the user. The host does not yet persist an
  idempotency key across that retry.
- MCP Apps are disabled on Windows until the WebView2 subframe IPC boundary is
  verified. The UI does not advertise them, Rust rejects the root connection,
  and the untrusted App protocol is not registered.

### Named follow-ups

- The SSE parser retains the cumulative response buffer. A follow-up should
  avoid repeated full-buffer scans and add explicit outbound request-body and
  concurrent-request limits.
- The endpoint policy needs explicit tests and policy decisions for the
  remaining IPv6 special-purpose ranges.
- The server registry needs explicit behavior for a poisoned mutex, its
  `MAX_SERVERS` limit, and installed-view lifetime.
- The main-frame CSP should narrow `frame-src` from the custom scheme to
  `buzz-mcp-app://localhost` after the cross-platform protocol form is proven.
- A narrow cosmetic race remains when a post request is rejected while its
  approval dialog is already closing.

> [!IMPORTANT]
> The official `@modelcontextprotocol/ext-apps` package requires
> `@modelcontextprotocol/sdk`. This adds approximately 140 lockfile entries,
> including server and OAuth packages. This draft imports the browser bridge,
> JSON-RPC schema, and client transport modules.

### Review focus

1. Does the sandbox, CSP, endpoint validation, and DNS-pinning model establish
   the right boundary for server-authored UI?
2. Are Streamable HTTP support and the explicit deferral of stdio and OAuth a
   reasonable first integration boundary?
3. Should tools without explicit MCP Apps visibility remain App-visible, and
   which App-visible tool calls should require user approval?
4. Does the current typed channel-surface seam compose cleanly with the
   behavior-preserving registry proposed in #3280?

### Related issue

No dedicated MCP Apps issue was found when this draft was opened. The closest
related discussion is #3280, which proposes a behavior-preserving registry for
channel features. This PR implements the MCP Apps host through the current
typed channel-surface seam. It does not implement or depend on that registry.

### Testing

- `CHECK_FILE_SIZES_BASE=upstream/main bin/just ci`
  - workspace checks, desktop checks, desktop build, and Tauri check passed;
  - the Tauri unit suite discovered 2,057 tests: 2,043 passed and 14
    OS-keychain or real-relay tests were ignored; all 3 rodio diagnostic tests
    passed;
  - the final mobile phase stops on the current-upstream
    `ChannelDetailPage keeps follow mode off while a tall newest message stays visible`
    failure;
  - this branch has no diff under `mobile/`; the same test fails at the same
    assertion when run alone on both this branch and the exact
    `upstream/main` base (`10d5a2641`).
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml mcp_apps`
  - 40 passed, including the live Streamable HTTP round trip.
- `cd desktop && pnpm test`
  - 3,892 passed.
- `cd desktop && pnpm lint`
- `cd desktop && pnpm build:e2e`
- `cd desktop && pnpm exec playwright test tests/e2e/mcp-app-channel-screenshots.spec.ts --project=smoke`
  - 1 passed, including the no-remount/one-initial-call regression.


jmreijnen-parkbee:
Nice, this is exactly the kind of surface a channel-feature registry would give a first-class home. I just opened #3280 (RFC: a desktop `ChannelFeaturePlugin` registry) proposing that channel tabs/settings/sidebar entries be *registered* rather than hand-wired into the shared `*ChannelContent` shells and the `ChannelScreen` dispatch. MCP-App tabs would slot in as a plugin's `tabs` entry (with your security-boundary host as the tab body), alongside Sequence/board/docs, with no edits to the central shells. Happy to align on how App-tabs would register if the maintainers are warm to the registry direction.

tmarman:
@wesbillman @wpfleger96 @baxen I think this is ready for review now. I’ve been running it locally for a few days and have put it through several rounds of review and hardening.

This first pass adds MCP Apps as channel tabs using Streamable HTTP. The App stays sandboxed, and Buzz asks for approval before it can post anything back into the channel.

I’ve left stdio, OAuth, and a few other pieces for follow-up. For now, I’d love your thoughts on whether this is the right first cut.

PR #3292, opened 2026-07-28
fix(desktop): mention autocomplete drops every relay-directory agent before eligibility runs

TNovs1:
## Problem

`useMentions` gates every autocomplete candidate through `isAgentIdentityInManagedList` against **only the local managed-agent set**. Any candidate flagged `isAgent: true` that is not managed by this desktop is dropped before `shouldHideAgentFromMentions` runs — so `relayAgentIsSharedWithUser`, the `respond_to` allowlist, and channel membership are never consulted for it. The entire relay-agent branch of `agentAutocompleteEligibility` is unreachable from the mention picker.

Net effect: an agent hosted anywhere other than the local desktop (registered in the relay agent directory via kind:10100, attested via NIP-OA, or both) can never be @-mentioned, even when its directory entry explicitly allowlists the current user.

## Reproduction

1. Register a server-hosted agent in the relay directory: kind:10100 with `respond_to: "allowlist"` including your pubkey.
2. Add it to a channel you are in. It appears in the member list.
3. Type `@` — the agent is never offered.
4. Delete its kind:10100 (NIP-09) — the identity is offered again, because with no directory entry it isn't flagged as an agent at all. Publishing the metadata the eligibility module wants to read is exactly what makes the agent invisible to it.

## Fix

Gate the candidate loop on `mergeKnownAgentPubkeys(managedAgentsQuery.data, relayAgentsQuery.data)` — the same managed ∪ relay-directory union the app already treats as "known agents" elsewhere (`useKnownAgentPubkeys`). Identity decides who is *known*; the eligibility rules (`shouldHideAgentFromMentions` → `relayAgentIsSharedWithUser`) keep deciding who is *invocable* — a directory agent whose entry excludes the viewer is still hidden, as before.

One line of behavior change plus a memoized union; no new API surface.

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -11,6 +11,7 @@ import {
 } from "@/features/channels/hooks";
 import { useIsArchivedPredicate } from "@/features/identity-archive/hooks";
 import type { MentionSuggestion } from "@/features/messages/ui/MentionAutocomplete";
+import { mergeKnownAgentPubkeys } from "@/features/agents/knownAgentPubkeys";
 import {
   coalesceAgentAutocompleteCandidates,
   coalesceAutocompleteCandidatesByKey,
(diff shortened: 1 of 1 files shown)


rmichelena:

rmichelena:
Cross-linking from #5483, where I listed this PR — with the specific reason it doesn't close #5363 on our repro, so you can check my reading.

Two things:

1. **Stale against `main`.** This patches `isAgentIdentityInManagedList(candidate, managedAgentPubkeys)`. Since #4913 that call site is `isAgentIdentityInAllowedList(candidate, mentionableAgentPubkeys)` — managed ∪ directory-eligible — so the union you build is roughly what `main` already passes in.
2. **The union still doesn't reach the failing case.** With the agent admitted by gate 1, gate 2 runs: it isn't in `mentionableAgentPubkeys` (its directory entry has `respond_to: null`), it *is* a member, so `shouldHideAgentFromMentions` returns `directoryAgentPubkeys.has(pubkey)` → `true` → hidden. And an agent with no `kind:10100` at all never enters the union in the first place.

Root cause we traced: `list_relay_agents` builds the directory from `kind:10100` only, while `respond_to` is published as `kind:30177` and channel membership as `kind:39002`, so agents using the dedicated events arrive `respond_to: null` / `channel_ids: []` and every eligibility branch below is starved of the data it tests. Fix at that layer in #5483, with an end-to-end check on a self-hosted relay.

Your diagnosis of the layering ("identity decides who is KNOWN, eligibility decides who is INVOCABLE") is the right frame, and it's the one #4913 landed. Would you consider closing this in favor of #5483? Maintainers' call in the end — and if you think the union still buys something on top of the directory fix, I'd rather hear that than have you close it.

Issue #2459, opened 2026-07-23
buzz-acp: turn completes ok but reply is silently lost when the agent's buzz CLI write fails


brocoppler:
A turn can complete "successfully" while the agent's reply is silently lost: if the agent's `buzz messages send` fails (observed: auth error, exit 3), the harness sees `end_turn` with `outcome=ok`, fires no failure notice, and requeues nothing. The human sees a mention with no response and no indicator.

Observed live (local relay, buzz-acp with `claude-code-acp`, kind:9 mention):

```
INFO acp::stream: …ails with an auth error (exit 3). The answer to the question is **14**, but I have no way to post it to the channel until…
INFO pool::prompt: turn complete for channel 85453cd6-…: end_turn
DEBUG buzz_acp: agent_returned agent=0 outcome="ok"
```

The agent computed the answer, reported its inability to post in its final text, and the harness treated the turn as ok — the failure classification from #1659/#2289 doesn't catch it because the ACP turn itself didn't error; only the out-of-band CLI write did.

What triggered the auth failure here is unconfirmed (transient; same env posted successfully in adjacent turns). The report is about the missing recovery path, not the auth error itself:

- The harness has no signal that a turn which was *supposed* to reply actually published anything. A cheap heuristic: turn ended `ok` + zero relay OKs for agent-authored channel events during the turn + triggering event was a mention → emit the existing failure notice (or retry once).
- Alternatively, surface CLI write failures back into the turn outcome so the existing dead-letter path applies.

Can take a PR for either direction if maintainers have a preference.

Chessing234:
Fixed in #2500 — mention turns that Ok with no channel publish now get a failure notice.

adambe0:
Corroborating data point from a headless deployment (buzz-acp under launchd driving `claude-agent-acp`, v0.4.23): we hit exactly this signature — mention delivered, typing reaction posts and deletes, `pool::prompt` logs `turn complete … end_turn` and `agent_returned outcome=ok`, but no kind-9 ever reaches the relay.

Root cause in our case was on the CLI side of the out-of-band write, with an extra twist: the agent's `buzz` resolved through a stale symlink to a **zero-byte** leftover binary, which executes as an empty script — **exit 0, no output**. So the reply write didn't just fail silently; it *succeeded* as far as both the agent and the harness could tell. Nothing surfaces under `RUST_LOG=buzz_acp=debug` because (as this issue notes) the harness is never the publisher.

Two takeaways that might be useful here:

[...]

PR #3311, opened 2026-07-28
feat(acp): isolate signing behind durable reply broker

Policy-Guru:
## Scope\n\nImplements the approved P0/P1 delivery foundation without activation, credential rotation, or automatic recovery.\n\n- removes raw signing credentials and durable-state path from worker/MCP environments\n- hides credential environment values in CLI and ACP help\n- adds a harness-owned constrained reply broker with a pre-send signed-event ledger, deterministic event ID reconciliation, and delivery-unknown blocking\n- adds the MCP relay_reply client and pair-scoped durable state path\n\n## Verification\n\nAt a28f209db1c8495567cbd2ebbd89e7ac3500f3bf:\n\n- cargo fmt --check\n- env -u BUZZ_ACP_LAZY_POOL cargo test -p buzz-acp --lib (618 passed)\n- cargo test -p buzz-dev-mcp --lib (96 passed)\n- cargo test -p buzz-cli --lib (251 passed)\n- cargo check -p buzz-acp -p buzz-dev-mcp\n\nDesktop cargo check is packaging-blocked because this source clone lacks binaries/buzz-acp-aarch64-apple-darwin; it reaches the build script before that missing generated resource.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -766,6 +766,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "base64",
+ "buzz-cli",
  "buzz-core",
  "buzz-persona",
  "buzz-sdk",
@@ -986,6 +987,7 @@ dependencies = [
  "tokio-util",
  "tracing",
  "tracing-subscriber",
+ "uuid",
  "windows-sys 0.61.2",
  "zeroize",
 ]
--- a/crates/buzz-acp/Cargo.toml
+++ b/crates/buzz-acp/Cargo.toml
@@ -20,6 +20,7 @@ path = "src/main.rs"
 buzz-core = { workspace = true }
 buzz-sdk = { workspace = true }
 buzz-persona = { path = "../buzz-persona" }
+buzz-cli = { path = "../buzz-cli" }
 
 # Nostr
 nostr = { workspace = true }
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -20,6 +20,20 @@ use crate::usage::{TurnUsage, UsageTracker};
 /// Lines exceeding this limit are rejected to prevent OOM from rogue agents.
 const MAX_LINE_SIZE: usize = 10_000_000; // 10 MB
 
+/// Signing material belongs to the harness or its dedicated broker, never to
+/// the model subprocess.  Keep this list at the final spawn boundary: callers
+/// may add arbitrary persona environment entries, but they cannot reintroduce
+/// a raw signing capability through `AcpClient::spawn`.
+const WORKER_RESTRICTED_ENV_KEYS: &[&str] = &[
+    "BUZZ_PRIVATE_KEY",
+    "BUZZ_AUTH_TAG",
+    "NOSTR_PRIVATE_KEY",
+    "BUZZ_ACP_PRIVATE_KEY",
+    // The durable ledger contains signed events (and potentially their NIP-OA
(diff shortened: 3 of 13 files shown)


rmichelena:
Live validation from a headless Buzz/ACP deployment: this PR's structured `relay_reply` path addresses two failures we reproduced in production-like use that prompt hardening alone cannot eliminate.

Environment:

- `buzz-acp` on Linux (current local `main` build) and macOS Intel (Buzz Desktop 0.5.3 bundled binary)
- Goose v1.45.0
- one agent using `codex-acp`; one using OpenRouter/DeepSeek directly through Goose's Developer shell tool

### Reproduction 1: Markdown becomes shell program text

When the model publishes with a command shaped like:

```bash
buzz messages send --content "message containing `inline code`, $VARS, or $(commands)"
```

Bash performs command/parameter substitution before `buzz` sees the content. Depending on the generated Markdown, the reply is altered, truncated, or never published. The ACP turn can still finish normally, leaving the human with typing/activity followed by silence.

### Reproduction 2: a quoted heredoc is still not a general transport

We then strengthened the prompt to require a quoted heredoc. That prevents expansion, but the prompt taught the model a fixed delimiter. While explaining that very pattern, the model emitted a line equal to the delimiter inside its response. Bash deterministically ended the heredoc there.

The corresponding relay event (`a779d9e7…`) was accepted but ended exactly after:

```text
<markdown literal>
```

Everything intended after the delimiter collision was lost. This is useful evidence that even a correctly quoted heredoc cannot safely transport arbitrary model-authored text when the model also chooses or repeats the delimiter.

As a control, we used Goose's structured `write` tool to create a workspace file and then ran only:

```bash
buzz messages send --content - < .buzz-outgoing.md
```

That delivered the full message and returned an event ID, proving the relay, identity, and thread metadata were healthy. It is a viable workaround, but it adds disk I/O and still relies on the model remembering a multi-step publication protocol.

We also observed a separate turn where the harness emitted its working/typing lifecycle signals, the model produced a response, and no kind-9 event was published. The triggering instruction said to answer through "stdin/stdout only"; the direct-provider model likely interpreted that as "do not call `buzz messages send`". The historical tool arguments were not persisted, so this cause cannot be proven, but the symptom matches #2698 and #3980.

### Why #3311 is the right boundary

The `relay_reply(channel_id, reply_to, content, idempotency_key)` tool removes model-authored content from shell syntax entirely and gives publication a narrow, harness-owned capability. The current implementation also provides the feedback operators need:

- success returns `event_id` plus the idempotency key;
- retry with the same key reconciles by deterministic event ID before resubmission;
- rejected and delivery-unknown outcomes remain distinguishable;
- the signing key stays out of the worker shell.

Please preserve those receipt semantics as part of the public tool contract: a model should consider a reply delivered only after an explicit `sent`/accepted result (or successful relay reconciliation), and should retry an unknown outcome only with the same idempotency key. It would also be valuable for `buzz-acp` to record per-turn whether `relay_reply` succeeded, so a turn cannot end silently with generated text but no durable reply.

Related: #2459, #2698, #3385, #3980, and #4057.

This live reproduction strongly supports merging the structured broker rather than trying to make arbitrary Markdown safe through increasingly specific shell instructions.

avivsinai:
AMQ will not hold a Buzz nsec. Worker-side `BUZZ_*` stripping shipped in https://github.com/avivsinai/agent-message-queue/pull/637; that is not a substitute for isolating signing behind a durable reply broker / scoped `client_publish` as this PR proposes. Linking from the AMQ Buzz-upstream bead so the two stay aligned.

PR #3315, opened 2026-07-28
feat(desktop): inline key setup, owner-direction primacy, and OPENAI_BASE_URL for card mints

tlongwell-block:
Follow-up to #3278 (Agent Trading Cards), per the design conversation in buzz-agent-trading-cards. Three knobs, one small PR — #3278 stays frozen at its approved SHA.

## 1. Inline `OPENAI_API_KEY` setup in the mint dialog

New Tauri command `card_mint_key_status(id)` reports whether a key resolves through the exact same env layering the mint uses (agent record > persona > global Agent Defaults > process env). It never returns the key.

When no key resolves, the dialog renders a key-setup panel instead of the mint form:

- **Cost disclosure**: "Minting a card costs money — it generates the art and card text through the OpenAI API with your key…"
- Pointer to platform.openai.com for key creation.
- **Free-path pointer**: "Don't want to spend money? The Export action on the agent shares the same importable agent as a plain file — free, just without the collectible card art."
- Password input + "Save key & continue" writes the key into the **global Agent Defaults env** — the same single source of truth every agent inherits. No second key store, nothing new to snapshot-scrub, no card-specific key that can drift from the agent key.

A `NO_OPENAI_KEY` mint error (key removed after the dialog opened) now flips into key-setup mode instead of a dead-end toast.

## 2. Owner-direction primacy (art AND card text)

The direction box now advertises both uses: *"Optional directions — art ("stormy night, lightning motif") and/or card text ("ability: Verify — scry 2"). Your directions take priority; anything you leave open is designed for you."*

The Responses prompt is restructured so the designer model treats owner input as authoritative:

- Owner directions are framed as `OWNER'S DIRECTIONS — these override the default art-style and copy guidance below wherever they conflict`, explicitly covering type line / ability / flavor text, with "use their wording (edited only for spelling)".
- The style-match-the-avatar anchor is demoted to a **default** that owner directions override.
- The fixed contract survives regardless: frame identity, layout, and text-fidelity requirements cannot be overridden.

Pinned by the rewritten `instructions_pin_style_match_default_and_owner_primacy` test.

## 3. `OPENAI_BASE_URL` override

Resolved through the same env layering; the mint posts to `{base}/responses` instead of the hardcoded api.openai.com. Supports endpoints and proxies that speak the OpenAI Responses shape with Bearer auth. (Azure OpenAI is NOT covered by this alone — it uses its own URL scheme and `api-key` header, so it would need a real driver.) Trailing slashes normalized (`responses_url_default_and_override` test).

## Verification (at 3f3f0fc44, same shell)

- `cargo test --lib`: 1844 passed / 0 failed
- `cargo clippy --all-targets -- -D warnings`: clean; `cargo fmt --check`: clean
- `tsc --noEmit`: clean; desktop tests: 3723/3723; `pnpm check` (biome + file-size gates): clean

## Round 2 (Wren's UX/code review — all findings addressed)

1. **Persistent pre-spend disclosure**: the normal mint form (key present) now carries a compact cost note + the free-path row — not just the no-key onboarding panel.
2. **Actionable directions**: "Get a key at platform.openai.com" is a real external-link button (`openUrl`), and the free path is a real **Share without card art** action that closes the mint dialog and opens the existing snapshot export flow (wired at the `UserProfilePanel` callsite; the prop is optional so the dialog stays testable standalone).
3. **No silent agent restarts**: key save no longer calls the general `set_global_agent_config` (which stops/restarts every running local agent whose effective env changes and whose restart counts the dialog discarded). New narrow `card_mint_save_openai_key` command: validated single-key read-modify-write of the latest on-disk config under the managed-agents store lock, never restarts anything — the mint re-reads config per call, so no restart is needed. Toast tells the user running agents pick the key up on next restart. Settings cache invalidated so the Agent Defaults editor stays fresh.
4. **Azure claim narrowed** in code comment + this body (Bearer-auth Responses endpoints/proxies only).
5. **Owner-copy rule vs 220-char limit**: prompt now says "use their wording within the 220-character text-box limit … if their text exceeds the limit, condense it minimally while keeping their words and intent" — no mutually impossible instructions. Prompt-contract test updated.

## Review plan

Wren co-reviews the UX (Tyler's directive: simple and intuitive) and produces fresh screenshots of the key-setup panel + updated direction box.


tlongwell-block:
Rolled into #3278 per Tyler: eva/agent-trading-cards was fast-forwarded to this branch's exact reviewed head edc8bea4c85d8e4915e107f4c2926bd0cf68c530 (no rebase, no force, commits untouched). #3278 now carries the full card-mint key UX work; this PR's diff vs #3278 is empty. Wren's post-merge 10/10/10 review + local redteam apply verbatim to the new #3278 head since the tree is bit-identical.

PR #3317, opened 2026-07-28
fix(desktop): show cached display names on startup

TheSentinel454:
## Why

Buzz restores cached channels and messages before profile lookups complete. On launch, that briefly exposes pubkey-derived labels in place of familiar display names.

## What

- Persist a bounded, relay-scoped cache of last-known display names, NIP-01 names, and NIP-05 handles
- Seed batch profile queries from those labels immediately, while keeping them stale so the existing relay request revalidates them
- Keep cached data presentation-only: avatars and ownership metadata are not persisted or used to seed profile-detail caches
- Remove cleared or missing profiles, purge a relay's labels when its community is removed, and include the cache in local-storage quota recovery
- Add unit coverage for parsing, bounds, eviction, malformed data, and cleared profiles
- Add an E2E regression that delays the relay profile response and verifies the cached name is rendered first

## Risk Assessment

Low. The cache is disposable, capped at 1,000 entries per relay, scoped by normalized relay URL, and always revalidated. It contains only public label fields and does not restore avatars, agent ownership, or authorization state.

## Verification

- `just ci`
- `pnpm typecheck`
- `pnpm test` — 3,727 passed
- `pnpm exec playwright test tests/e2e/channels.spec.ts --grep "cached profile labels"` — passed

Generated with Codex

### DIFF
--- a/desktop/src/features/profile/hooks.ts
+++ b/desktop/src/features/profile/hooks.ts
@@ -41,6 +41,10 @@ import {
   shouldFetchAvatar,
   resolveAvatarDataUrl,
 } from "@/features/profile/lib/selfProfileStorage";
+import {
+  readCachedUserLabels,
+  writeCachedUserLabels,
+} from "@/features/profile/lib/userLabelStorage";
 import { useCommunities } from "@/features/communities/useCommunities";
 import { updateCachedChannelMemberDisplayName } from "@/features/channels/channelMemberProfileCache";
 
@@ -317,6 +321,8 @@ export function useUsersBatchQuery(
   },
 ) {
   const queryClient = useQueryClient();
+  const { activeCommunity } = useCommunities();
+  const relayUrl = activeCommunity?.relayUrl ?? "";
   const normalizedPubkeys = [
     ...new Set(pubkeys.map((pubkey) => pubkey.toLowerCase())),
   ]
@@ -352,6 +358,9 @@ export function useUsersBatchQuery(
       }
       if (toFetch.length > 0) {
         const fresh = await getUsersBatch(toFetch);
+        if (relayUrl) {
+          writeCachedUserLabels(relayUrl, fresh.profiles, fresh.missing);
+        }
         for (const pubkey of toFetch) {
           const summary = fresh.profiles[pubkey] ?? null;
           queryClient.setQueryData<UsersBatchEntry>(
@@ -368,6 +377,10 @@ export function useUsersBatchQuery(
     // key entirely. Without this, already-resolved authors would flash back
     // to their raw pubkey while the larger batch refetches.
     placeholderData: keepPreviousData,
+    initialData: relayUrl
+      ? () => readCachedUserLabels(relayUrl, normalizedPubkeys)
+      : undefined,
+    initialDataUpdatedAt: 0,
     staleTime: 60_000,
     gcTime: 5 * 60 * 1_000,
   });
@@ -375,6 +388,9 @@ export function useUsersBatchQuery(
   // Seed individual "user-profile" cache entries so avatar clicks are instant
   // cache hits instead of fresh network requests.
   React.useEffect(() => {
+    // Persisted labels are intentionally presentation-only. Wait for a relay
+    // result before seeding profile-detail caches that also carry ownership.
+    if (query.dataUpdatedAt === 0) return;
     const profiles = query.data?.profiles;
     if (!profiles) return;
     for (const [pubkey, summary] of Object.entries(profiles)) {
@@ -391,7 +407,7 @@ export function useUsersBatchQuery(
           },
       );
     }
-  }, [query.data, queryClient]);
+  }, [query.data, query.dataUpdatedAt, queryClient]);
 
   return query;
 }
--- a/desktop/src/features/profile/hooks.ts
+++ b/desktop/src/features/profile/hooks.ts
@@ -5,7 +5,6 @@ import type {
 } from "@tanstack/react-query";
 import * as React from "react";
 import {
-  keepPreviousData,
   useInfiniteQuery,
   useMutation,
   useQuery,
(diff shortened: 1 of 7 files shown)


TheSentinel454:
🤖 Before / after with the relay profile response deliberately delayed:

| Before | After |
| --- | --- |
| Before: pubkey-derived label | After: cached display name |

desktop/src/features/profile/hooks.ts
```diff
@@ -368,13 +377,20 @@ export function useUsersBatchQuery(
     // key entirely. Without this, already-resolved authors would flash back
     // to their raw pubkey while the larger batch refetches.
     placeholderData: keepPreviousData,
+    initialData: relayUrl
+      ? () => readCachedUserLabels(relayUrl, normalizedPubkeys)
+      : undefined,
+    initialDataUpdatedAt: 0,
```
wpfleger96: 🤖 I think `initialData` here silently defeats the `keepPreviousData` right above it. In TanStack Query, `placeholderData` only applies when a query has no data — and once `initialData` seeds the new query, it counts as real data. So in the exact case the comment above describes (scroll-back grows the pubkey set → brand-new query key), any label-cache hit means the new query starts from label stubs (`avatarUrl: null`) instead of the previous batch's full profiles. Two symptoms from that one cause:

1. Message avatars blank on scroll-back until the fetch completes — a single frame when the per-pubkey entry cache is fresh, but a full network round-trip when entries are >60s old. That partially reintroduces the flash #1137 fixed, in avatar form.
2. `ChannelScreen` gates DM huddle-member resolution on `isPending || isPlaceholderData`; with `initialData` the query is instantly `success` and not placeholder, so the gate opens while only label stubs are loaded. Fail direction is benign (ownership absent, never wrongly granted), but the gate's intent is defeated.

Suggested fix — deliver the cache through the placeholder chain instead:

```ts
placeholderData: (prev) => prev ?? readCachedUserLabels(relayUrl, normalizedPubkeys),
```

and drop `initialData`/`initialDataUpdatedAt`. That preserves `keepPreviousData` semantics (previous full profiles win over label stubs), keeps `isPlaceholderData` true so the ChannelScreen gate behaves as before, still triggers the revalidating fetch, and the `dataUpdatedAt === 0` seeding guard below still holds since placeholder data never advances `dataUpdatedAt`. Your new e2e test should pass unchanged. Bonus: the localStorage read+parse gets skipped whenever previous data exists instead of running on every scroll-back key growth.

wpfleger96 (commented):
🤖 thanks for this — it fills a real gap (the persisted snapshots don't store display names at all, so restored messages show pubkey labels until profiles resolve). The design is careful: presentation-only fields, always-stale revalidation, purge on community removal, quota-recovery integration. One behavioral thing worth fixing inline below, plus a few small nits.

desktop/tests/e2e/channels.spec.ts
```diff
@@ -487,8 +487,13 @@ async function expectIntroActionsShareRow(
...
 
-test.beforeEach(async ({ page }) => {
-  await installMockBridge(page);
+test.beforeEach(async ({ page }, testInfo) => {
+  await installMockBridge(
+    page,
+    testInfo.title.includes("cached profile labels")
+      ? { usersBatchDelayMs: 10_000 }
+      : undefined,
+  );
```
wpfleger96, line 497: 🤖 nit: injecting the delay by string-matching the test title is a bit fragile — if the test gets renamed the delay silently stops applying and the test would still pass without proving anything. A tag or a per-test fixture would be sturdier. Not blocking.

desktop/src/features/profile/hooks.ts
```diff
@@ -352,6 +358,9 @@ export function useUsersBatchQuery(
       }
       if (toFetch.length > 0) {
         const fresh = await getUsersBatch(toFetch);
+        if (relayUrl) {
+          writeCachedUserLabels(relayUrl, fresh.profiles, fresh.missing);
+        }
```
wpfleger96, line 362: 🤖 nit: this writes labels under the `relayUrl` captured at render, so a fetch resolving across a community switch could in theory file profiles under the old relay's cache key. In practice the `communityKey` remount tears the query down first, so I think this is theoretical — just noting it.

desktop/src/features/profile/lib/userLabelStorage.ts
```diff
@@ -0,0 +1,172 @@
...
+type CachedUserLabel = {
+  displayName: string | null;
+  name: string | null;
+  nip05Handle: string | null;
+  updatedAt: number;
+};
+
+type UserLabelCache = {
+  version: 1;
+  updatedAt: number;
```
wpfleger96: 🤖 nit: the top-level `updatedAt` on the cache envelope is written and parsed but never consumed — the per-entry `updatedAt` does all the work. Could drop it.

desktop/src/features/profile/hooks.ts
TheSentinel454: 🤖 Addressed in 502e2612. Cached labels now flow through the placeholder chain after previous query data, so resolved profile/avatar data wins and `isPlaceholderData` retains its original semantics. Added regression coverage for the precedence rule.

desktop/tests/e2e/channels.spec.ts
TheSentinel454: 🤖 Addressed in 2c1099a9. The delayed relay setup now uses an explicit `@cached-profile-labels` Playwright tag and `testInfo.tags`, so renaming the test cannot silently disable the delay.

desktop/src/features/profile/hooks.ts
TheSentinel454: 🤖 Reviewed; no code change. The query closure captures the relay key, and the Rust `query_relay` path resolves the active relay API URL before awaiting the network response, so storing that response under the captured relay key is intentional. Eliminating the remaining pre-dispatch scheduling race would require explicit relay identity or cancellation plumbing and is disproportionate for this PR.

desktop/src/features/profile/lib/userLabelStorage.ts
TheSentinel454: 🤖 Addressed in e357b4c0. Removed the unused envelope-level `updatedAt`; per-entry timestamps remain the source for cache bounding, and older payloads with the extra field continue to parse.

wpfleger96 (approved):

PR #3320, opened 2026-07-28
fix(desktop): preserve early relay auth challenges

tlongwell-block:
## Summary

- buffer native WebSocket frames until `plugin:websocket|connect` returns the socket ID and the NIP-42 auth waiter is installed
- drain those frames in order before normal inbound handling begins
- add a deterministic E2E seam and regression for an AUTH challenge delivered before connect resolves

## Root cause

The native WebSocket task starts forwarding relay frames before the connect command returns its socket ID. Buzz relay sends the NIP-42 AUTH challenge immediately on connection, so Desktop could process that challenge while `wsId` and `authRequest` were still unset. The challenge was discarded, the cold connection failed authentication, and the first plain-text send failed. Channel history still appeared because it loads through the Tauri channel-window command rather than this frontend WebSocket.

## Verification

- deterministic regression is red before the fix (`connecting` after 5s) and green after it
- `pnpm test`: 3,721 passed
- `pnpm build:e2e`: passed
- focused early-AUTH E2E: passed
- existing failed-initial-dial E2E: passed
- `pnpm typecheck`: passed
- Biome + file-size pre-commit checks: passed

The complete 7-test relay-reconnect file passed once before the final line-count-only compaction. After that compaction, two full-file reruns each had the same unrelated startup-seam flake in the existing initial-dial test; both the new test and that existing test pass independently at the committed HEAD.


AmagiDDmxh:
Prepared the focused ReadOnlyRelayClient follow-up as #5572. It applies this PRs all-frame gate to inactive-community observer connections and includes a deterministic early AUTH regression test.

wesbillman:
## Mordecai review — changes required

Reviewed exact head `6b03028c5fd8a43079afb6ef8918a0035d4fdb40` source-only.

### High: a stalled early AUTH handler defeats the auth timeout and can permanently wedge reconnect

The fix creates the authentication timeout, then serially awaits every buffered frame before it awaits the authentication promise:

```ts
const authentication = armRelayAuthentication(...);
while (pendingInbound.length > 0) {
  await this.handleWsMessage(pendingInbound.shift(), generation);
}
pendingInbound = null;
await authentication;
```

(`desktop/src/shared/api/relayClientSession.ts:559-574`)

For the early `AUTH` frame this PR targets, `handleWsMessage()` awaits `handleAuthChallenge()`, which awaits the Tauri `createAuthEvent` command before its generation check (`:755-788`, `:837-852`). If that signer invocation stalls beyond `AUTH_TIMEOUT_MS`:

1. `armRelayAuthentication` fires and `resetConnection()` increments the generation and schedules reconnect.
2. The authentication promise rejects, but `connect()` cannot observe it because it is still awaiting the stalled buffered-frame handler.
3. `ensureConnected()` retains the unresolved `connectPromise` (`:491-519`).
4. The scheduled reconnect calls `ensureConnected()`, gets that same unresolved promise, and cannot start a replacement connection.

If signing never settles, reconnect is wedged indefinitely. Even if it eventually settles, recovery is delayed by the stalled dependency rather than bounded by the advertised auth timeout. This is a regression from the pre-PR timeout path: `connect()` awaited the auth promise directly, so timeout unwound `connectPromise` and allowed the reconnect coordinator to proceed.

The new E2E seam only delays `connect` by 50 ms and uses a normally completing signer, so it cannot expose this lifecycle failure.

**Required shape of fix:** make timeout/reset able to abort or stop waiting for buffered-frame processing, while preserving ordered delivery and preventing stale handlers from mutating the replacement generation. Add a deterministic early-AUTH + stalled-signing regression proving the timeout clears the in-flight connect and a replacement attempt starts.

### Medium: pre-connect inbound buffering is unbounded

`pendingInbound` accepts every native frame until `plugin:websocket|connect` resolves (`:533-535`) with no frame/byte cap. A hostile relay can flood this interval and grow renderer memory without bound. This is a new network-boundary queue; give it an explicit bound and fail/reset the connection on overflow rather than permitting unbounded growth.

I found the normal early-AUTH ordering itself sound: the waiter is armed before draining, the drain remains FIFO (including frames arriving during drain), and normal handling begins without a zero-length transition gap. All current GitHub checks are green, but they do not cover either boundary above.

wesbillman (dismissed):

Requesting changes at `6b03028c5fd8a43079afb6ef8918a0035d4fdb40` for two lifecycle/network-boundary issues:

1. **High: a stalled early AUTH signer defeats the auth timeout and wedges reconnect.** `connect()` arms the auth timeout, but then awaits each buffered frame before awaiting the authentication promise (`desktop/src/shared/api/relayClientSession.ts:559-574`). Processing the early `AUTH` frame awaits `createAuthEvent()` (`:755-788`, `:837-852`). If that invocation stalls, the timeout resets the connection and schedules reconnect, but `connect()` remains blocked in the frame handler. `ensureConnected()` therefore retains the unresolved `connectPromise` (`:491-519`), and the reconnect attempt reuses that same promise instead of starting a replacement socket. Before this change, `connect()` directly awaited the auth promise, so its timeout unwound the in-flight connect. Make timeout/reset stop waiting for stale buffered-frame work, and add a deterministic early-AUTH plus stalled-signing regression proving a replacement attempt starts.

2. **Medium: the new pre-connect inbound queue is unbounded.** `pendingInbound` accepts every relay frame until the native connect command resolves (`:533-535`) without a frame or byte ceiling. This is network-controlled input crossing into renderer memory. Give the queue an explicit bound and fail/reset on overflow.

The normal FIFO handoff is otherwise sound: the socket ID and waiter are installed before draining, frames arriving during async drain remain ordered, and close/error invalidates the generation. Current CI is green, but the new E2E uses a 50 ms delay and a normally completing signer, so it does not exercise the timeout/reconnect failure.

wesbillman (approved):
Mordecai, an automated reviewer, commenting via Wes’s GitHub account.

Re-reviewed source at `0506ae112db5f013a0f5249d224f82953325669e`. The two previously blocking findings are addressed:

- `createRelayInboundBuffer` caps the pre-connect queue at 256 frames and fails/resets the connection on overflow.
- `connect()` now races the drain against authentication timeout and buffer overflow. If early AUTH signing stalls, the auth timeout rejects the race, `connect()` unwinds, and `ensureConnected()` releases `connectPromise`, allowing the scheduled reconnect to open a replacement socket.
- The new deterministic E2E covers that exact stalled early-AUTH path and requires a second dial followed by `connected`.

The FIFO transition remains sound: frames received during an asynchronous drain stay queued, and the queue switches to live handling synchronously once empty. Close/error and overflow invalidate the generation, so stale work cannot mutate the replacement session.

No blocking findings in the updated diff. Source-only re-review; I did not run tests or harnesses. Current GitHub checks are green.

wesbillman (dismissed):
Princess Donut, an automated reviewer, commenting via Wes's GitHub account.

Re-reviewed exact head `0506ae112db5f013a0f5249d224f82953325669e`.

The stalled-signer lifecycle defect is fixed: `Promise.race([drain, authentication, inbound.overflow])` now lets the AUTH rejection unwind `connect()` without awaiting the stalled drain, so `ensureConnected()` clears the old `connectPromise`; the new deterministic E2E exercises a replacement dial. The FIFO handoff remains sound.

One network-boundary issue remains:

**Medium: the frame-count cap is not a memory bound.** `relayInboundBuffer.ts` accepts 256 arbitrary `unknown` messages and tracks only `pending.length`. The native transport uses `connect_async(url)` with tungstenite defaults (`desktop/src-tauri/src/native_websocket.rs:124-135`); tungstenite 0.29 permits an incoming message up to 64 MiB by default. Because the race this PR fixes proves the native task can deliver multiple messages before `connect` returns, a hostile relay can queue roughly 256 × 64 MiB (about 16 GiB before JS/IPC representation overhead) in the renderer during this interval. Thus the original unbounded-memory finding is reduced but not resolved.

Smallest safe remedy: track a cumulative byte estimate alongside frame count and overflow/reset when either limit is exceeded. For the native tagged messages, count UTF-8 bytes for text and byte length for binary; use a conservative fallback for any other shape. Add a test with a small number of oversized frames proving the byte budget trips below the 256-frame ceiling.

**BLOCK** until the pre-connect queue has an explicit byte ceiling. All other previously reported behavior is resolved at this head.

wesbillman (dismissed):

Consolidated Royal Court re-review at exact head `0506ae112db5f013a0f5249d224f82953325669e`.

The stalled early-AUTH signer defect is fixed: racing the drain against authentication now lets timeout rejection unwind `connect()`, release `connectPromise`, and permit the replacement dial. The deterministic E2E covers that lifecycle. FIFO handoff and generation invalidation remain sound.

One network-boundary blocker remains:

**Medium: the frame-count cap still permits attacker-controlled multi-gigabyte buffering.** `desktop/src/shared/api/relayInboundBuffer.ts:22-29` limits the pre-connect queue to 256 arbitrary messages but does not account for their size. The native transport opens with `connect_async(url)` and no custom WebSocket configuration (`desktop/src-tauri/src/native_websocket.rs:124-135`). Its locked tungstenite 0.29 default allows incoming messages up to 64 MiB, so a hostile relay can enqueue roughly 256 × 64 MiB, about 16 GiB before JS/IPC overhead, during the pre-connect interval. A frame count alone therefore does not resolve the original unbounded-memory boundary.

Add an explicit cumulative byte ceiling, overflowing/resetting when either the frame or byte limit is exceeded, plus a focused regression where a small number of oversized frames trips the byte budget below the 256-frame ceiling. A tighter native per-message limit may also be appropriate, but it does not replace bounding the queue's aggregate memory.

wesbillman:

Risk reassessment at `0506ae112db5f013a0f5249d224f82953325669e`: I am withdrawing the aggregate-byte-limit blocker. The PR now bounds the newly introduced pre-connect queue to 256 frames; accumulating the theoretical 16 GiB requires saturating that very short interval with maximum-size messages from a hostile configured relay. A cumulative byte budget remains sensible transport hardening, but it is not proportionate to block this targeted early-AUTH fix. The stalled-signer lifecycle defect is fixed and deterministically covered. No remaining merge blocker from this review.

wesbillman (approved):

Issue #3034, opened 2026-07-26
[Bug] Provider-backed agents display local Claude config instead of remote runtime and expose env values


bencassistant-ship-it:
## Summary

Provider-backed managed agents display the local desktop's inherited harness and configuration instead of the runtime that the remote provider actually deployed.

In our case, five agents were deployed through a `buzz-backend-*` provider to an always-on Linux host. Their saved records correctly have `backend.type = "provider"`, a provider ID, and a non-empty remote `backend_agent_id`. The remote services are active and execute different ACP commands (Hermes profiles, Codex ACP, and Claude Code ACP).

Despite that:

- **Edit agent** shows `Harness: Claude Code` and `Model: Harness default` for every provider-backed agent.
- Agent rows show `Default model` even where the provider enforces a remote model.
- The **Runtime** tab reads and renders the local machine's Claude configuration as though it applied to the remote agent.
- That Runtime tab includes full environment-variable values from the local Claude config. In testing, a credential-like value was rendered in plaintext.
- Removing the provider executable from discovery does not stop the already-running remote agent, but the card still presents local edit/runtime controls without clearly distinguishing stale deployment metadata from effective remote state.

This makes it very difficult to determine which harness/model is actually executing and can expose unrelated local secrets on a remote-agent details screen.

## Environment

- Buzz Desktop 0.4.26
- macOS
- `block/buzz` checkout at `a31fc4d2f35d51cdf45ff8c61fc3a07f49c665e8`
- Custom provider discovered as `buzz-backend-<id>`
- Remote `buzz-acp` services managed by systemd on Linux
- Provider protocol: `info` + `deploy`

## Steps to reproduce

1. Configure the desktop's default local harness as Claude Code.
2. Install a valid `buzz-backend-<id>` provider.
3. Create a managed agent using that remote provider.
4. Have the provider deploy a non-Claude runtime and return a remote `agent_id`.
5. Confirm that the saved managed-agent record has:
   - `backend.type = "provider"`
   - the expected provider ID/config
   - a non-empty `backend_agent_id`
6. Confirm independently on the remote host that the running service uses a different ACP command/model.
7. Open the agent in Buzz:
   - **Edit** → observe `Claude Code / Harness default`
   - **Runtime** → observe local Claude config entries and full environment-variable values

An example of the contradictory state, with identifying values removed:

```text
Saved backend: provider / remote agent ID present
Saved local fallback agent_command: claude-agent-acp
Edit dialog: Claude Code / Harness default
[...]

PR #3323, opened 2026-07-28
Fix provider agent runtime projection

asymetryk:
## Summary

- keep provider-backed agent profiles from projecting local runtime, ACP, MCP, startup, model, and environment configuration
- keep the provider backend identifier and status visible
- add an integration regression test with deliberately conflicting local fallback values

## Motivation

Provider-backed agents can currently display local Claude/runtime configuration that is not authoritative for the remote provider and may expose environment values. This is a focused privacy-safe first slice of #3034.

## Test evidence

- `just ci` — passed
- targeted Playwright integration test — passed
- full `profile.spec.ts` integration suite — 26 passed; one unrelated existing media-proxy assertion failed because the mock returned `buzz-media://localhost/...` instead of `http://127.0.0.1:54321/...`

## Risk and rollback

The change only affects owner profile rendering for agents whose backend type is `provider`. Local agents retain the existing runtime/config UI. Reverting this commit restores the previous projection.

## Follow-up

Provider-specific editing/lifecycle UI remains tracked in #3034; this PR intentionally does not broaden provider mutation behavior.

### DIFF
--- a/desktop/src/features/profile/ui/UserProfilePanelFields.tsx
+++ b/desktop/src/features/profile/ui/UserProfilePanelFields.tsx
@@ -244,6 +244,7 @@ export function buildOwnerFields({
   relayAgent: RelayAgent | undefined;
 }): ProfileField[] {
   const fields: ProfileField[] = [];
+  const isProviderManaged = managedAgent?.backend.type === "provider";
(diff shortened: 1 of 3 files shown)


asymetryk:
@codex review

PR #3325, opened 2026-07-28
feat(desktop): add Google Antigravity runtime

kevinmjones:
## Summary

- add Google Antigravity as a first-class Buzz agent runtime
- launch the official `agy` CLI through the community `agy-acp` ACP adapter
- expose official CLI installation guidance while keeping the source-only adapter owner-reviewed
- normalize zero-argument adapter spawning, reserve the built-in runtime ID, and include Antigravity processes in managed cleanup
- lock provider selection to the vendor runtime while preserving ACP model selection

## Validation

- `cargo test -p buzz-acp` — 625 passed
- `pnpm test` — 3,722 passed
- `cargo test` in `desktop/src-tauri` — 1,817 passed, 14 environment-specific tests ignored
- `pnpm typecheck`
- `pnpm check`
- `git diff --check`
- live Buzz-to-`agy-acp` ACP initialization/model-discovery handshake

## Notes

`agy-acp` is currently a source-only community adapter rather than a Google-owned release. This change intentionally does not auto-install it from an unpinned Git branch. A live Gemini prompt was not exercised because the test machine does not have an authenticated `agy` installation.

Originating Buzz channel: `eeac2838-a8fa-4280-b23f-84819773058b`


JPeetz:
Ran the live end-to-end test this PR's description flagged as missing (no authenticated `agy` install on the original test machine).

**Result: PASS.** Built `hicder/agy-acp` (the adapter this PR pins to) against a real, signed-in Antigravity IDE install and drove its ACP stdio protocol directly:

1. `initialize` → `protocolVersion: 1`
2. `session/new` → session created, model list returned (worth noting: Antigravity surfaces non-Gemini models too — `claude-sonnet-4-6`, `gpt-oss-120b-medium` — alongside `gemini-3.6-flash-high`, etc.)
3. `session/prompt` ("Reply with exactly one word: PONG") → real `session/update` with `text: "PONG"`, then `stopReason: "end_turn"`

That closes the one gap called out in the PR description.

**Two things worth flagging before merge, not blockers to the test result above:**

- **Adapter choice**: this PR pins to `hicder/agy-acp` (22★, fork of `openabdev/agy-acp`). There are at least 14 independent, non-canonical `agy-acp` implementations on GitHub right now — no first-party Google release. `shubzkothekar/antigravity-acp` (26★) is a plausible alternative that doesn't appear to have been evaluated. Not asking for a rebuild, just flagging so a maintainer can make that call deliberately rather than by default.
- **Upstream bug in `hicder/agy-acp` itself** (not this PR): its own `prepare_auth()` test helper (`src/tests.rs`) only checks `GEMINI_API_KEY` or `~/.gemini/antigravity-cli/settings.json`. Neither exists on this working, authenticated IDE install — no `settings.json` there at all. So the adapter's own `#[ignore]`d e2e test would falsely report "no auth" on a machine where the live round-trip actually works. Filing this upstream separately; noting here since it affects confidence in the adapter's own test coverage.

Happy to share the exact commands/output if useful for review.

Issue #2878, opened 2026-07-25
Workflow request_approval: the documented from: syntax is rejected at approval time, and the only syntax that works accepts any key in the community


standing-order:
Buzz Version 0.4.26 (0.4.26)
MacOS Sequoia 15.7.5

<h2 id="summary">Summary</h2>
<p><code>WorkflowStep::RequestApproval</code> documents its <code>from:</code> field as <em>"User mention or role"</em> with the example <code>"@release-manager"</code>. The relay's <code>check_approver_spec</code> accepts exactly three things — the empty string, <code>"any"</code>, and a 64-character hex pubkey — and fails closed on everything else.</p>
<p>The two do not overlap. The result is a fork with no good branch:</p>

from: value | Documented? | Result
-- | -- | --
"@release-manager" | ✅ the doc comment's own example | ❌ forbidden: approver spec '@release-manager' is not yet supported
"any" | ❌ undocumented in the schema | ✅ works — and any authenticated key in the community may approve
64-char hex pubkey | ❌ undocumented in the schema | ✅ works, exact match
omitted / "" | — | ✅ works — same unrestricted behaviour as "any"

<p><strong>Follow the documentation and your gate fails closed. Reach for the form that actually works and you get an approval gate with no membership check on it.</strong> The safe-and-documented option and the working option are disjoint sets.</p>
<p>This is a documentation/validation mismatch, not a missing feature, and it is worth fixing independently of WF-08.</p>
<h2 id="environment">Environment</h2>
<ul>
<li>Source read at <code>block/buzz</code> commit <code>a64cc71</code> (<code>main</code>).</li>
<li>Relay under test reports <code>version: 0.2.0</code> via NIP-11, matching <code>crates/buzz-relay/Cargo.toml</code> at that commit. <strong>Version match, not commit-exact.</strong></li>
<li><strong>What we ran:</strong> created and triggered workflows carrying <code>request_approval</code> on a live relay, with <code>from:</code> set to both an <code>@mention</code> and to <code>any</code>, and attempted <code>buzz workflows approve</code>.</li>
[...]

Issue #2830, opened 2026-07-25
Approval gates: the grant binds a run and a step index, not the step


tomjwxf:
Read through #2376 and PR #2377, and #2509. This is additive to all three and
narrower than either: it is about what the grant is bound to at resume time.
The arm-then-announce ordering and the fail-closed approver resolution in #2377
both look like the right calls.

As far as I can follow the resume path:

- `generate_approval_token` returns a bare `Uuid::new_v4()`. The doc comment is
  explicit that `run_id` and `step_id` are accepted for logging context and are
  not mixed in, so the token identifies the gate but commits to nothing about
  it.
- The `workflow_approvals` row carries the token hash and `step_index`.
- On grant, the handler reads the run, reads the workflow by
  `run.workflow_id`, and calls `execute_from_step` with the stored index.

So the definition executed at resume is whatever is in the database when the
grant lands, and the resume point is an integer offset into it. Nothing ties
the grant to the definition that produced the announcement.

Three things follow, roughly in order of how much they would bother me:

1. **Insert a step and the index moves.** A run suspends at index 3. Someone
   edits the workflow and adds a step above it. The grant resumes at index 3 of
   the new definition, which is now a different step. The approver approved one
   thing and another runs.

2. **Edit the step in place and the grant still validates.** A `send_message`
   destination changes, or a `call_webhook` URL does. The token still hashes to
   the same row, the run is still `WaitingApproval`, and the grant path has
   nothing to notice with.

3. **The announcement text is not bound to anything.** kind:46010 carries
   `message`, which is free text from the definition. Even with no edit at all,
   the sentence in the approver's needs-action feed is asserted rather than
   derived, so there is no way to check afterwards that the approver was shown
   a description of what actually ran.

I do not know whether something upstream refuses to update a workflow that has
runs in `WaitingApproval`. I could not find one, but I have only read the tree
from the outside and may have missed it. If that guard exists, 1 and 2 mostly
go away and 3 still stands.

If it does not, the fix looks small and sits inside the surface #2377 already
touches. Store a digest over the definition, or just over the resolved step, on
the `workflow_approvals` row at arm time. Carry it in the kind:46010 content.
On grant, recompute it against the reloaded definition and refuse the resume on
mismatch. A mismatch is not a stale approval to be tolerated, it is a request
to do something else, so refusing and requiring a fresh gate seems like the
right default.

That would also give #2509's `verdict_ref` firmer ground, since a checkable
[...]

PR #2377, opened 2026-07-22
feat(workflow): persist and resume approval gates (WF-08)

brocoppler:
## What this solves

Closes the WF-08 gap called out in VISION.md and ARCHITECTURE.md (known-gap #5): runs that hit `request_approval` were marked Failed with "approval gates not yet implemented". The resume half — grant/deny handlers, `resume_workflow_after_approval`, `execute_from_step`, `workflow_approvals` CRUD, migration, CLI, SDK — already existed but was unreachable because the suspend half never persisted anything. This PR wires that half. Design decisions are laid out in #2376 for sign-off.

## How it works

- **Executor (`buzz-workflow`):** `StepResult::Suspended` now carries a `PendingApproval` (token, step id, resolved approver spec, message, timeout), captured where the resolved action is in scope. The approver spec is resolved *before* suspension — `"any"`, 64-char hex, or `npub1…`, each with optional leading `@`. Unresolvable specs (display names) fail the step with an actionable error instead of minting an approval `check_approver_spec` could never authorize. Invalid `timeout:` strings now error at the step as well.
- **Engine (`finalize_run`):** the Failed stub is replaced with arm-then-announce: (1) `create_approval` row — raw token hashed by `buzz-db`; (2) run → `WaitingApproval` at the suspended step with the accumulated trace; (3) announce via the new `ActionSink::emit_approval_requested`. Ordering means a half-armed gate is never observable. Any persistence failure marks the run Failed; announce failures are logged at ERROR but don't fail the run (the gate is armed and grantable by token).
- **Relay sink:** emits kind:46010 signed by the relay keypair — `d` tag = SHA-256 token hash (the grant handlers' `get_approval_by_stored_hash` key, derived from the now-`pub` `buzz_db::workflow::hash_approval_token`, keeping every hash derivation on one definition), `p` tag = resolved approver or the workflow owner for `"any"` (so the request lands in `query_needs_action`), `h` tag when channel-scoped, and content JSON carrying the raw token, message, workflow/run/step ids, and expiry. Persist + fan-out mirror the `send_message` path (`insert_event_with_thread_metadata` + `dispatch_persistent_event`).

With the row minted and the run in `WaitingApproval`, the existing grant path (`handle_approval_grant` → spawn `resume_workflow_after_approval(step_index + 1)`) and deny path (→ Cancelled) operate unchanged — no relay handler changes.

## Testing

[...]

Issue #2376, opened 2026-07-22
WF-08: wire approval-gate persist/suspend/resume (design + implementation)


brocoppler:
The VISION/ARCHITECTURE docs both call out WF-08 as "infrastructure exists, glue still drying": the executor mints a token and returns `Suspended`, but `finalize_run` marks the run **Failed**, so the grant/deny handlers, `resume_workflow_after_approval`, `execute_from_step`, and the `workflow_approvals` CRUD are all currently unreachable. The conformance suite has two `#[ignore]`d rows explicitly blocked on this.

I've implemented the missing half and I'm opening the PR alongside this issue — flagging three design decisions here for explicit sign-off:

**1. Arm-then-announce ordering.** On suspension: create the `workflow_approvals` row (raw token hashed by `buzz-db`), then set the run `WaitingApproval` at the suspended step, then emit kind:46010. A grant can only act on a run that is both `WaitingApproval` *and* has a pending row, so this order never exposes a half-armed gate. Any persistence failure marks the run Failed (a run must never wait on an ungrantable gate). The announcement is best-effort once the gate is armed — an emit failure is logged at ERROR but doesn't fail the run, since the token remains grantable via `buzz workflows approve`. Reasonable?

**2. Approver-spec resolution fails closed.** `request_approval.from` accepts `"@release-manager"`-style specs in the schema examples, but `check_approver_spec` in the grant handler authorizes only `""`/`"any"`/64-char hex. Unresolved, that's an approval **nobody can ever grant** — a silent dead end until expiry. I resolve at suspend time: `any`/hex/`npub1…` (with or without `@`) are accepted; display names **fail the step loudly** with an actionable error. Display-name → pubkey resolution via the member directory would be a follow-up (it needs a name-collision policy — same ambiguity problem `resolve_mention_pubkeys` handles for SendMessage). OK to land strict-first?

**3. kind:46010 shape.** Relay-signed; `d` tag = SHA-256 token hash (the exact `get_approval_by_stored_hash` lookup key, derived from the now-shared `buzz_db::workflow::hash_approval_token`); `p` tag = resolved approver (or the workflow owner for `"any"`) so it lands in the needs-action feed (`query_needs_action` joins on mentions); `h` tag when the workflow is channel-scoped; content JSON carries the raw token + message + workflow/run/step + expiry. Grant/deny lifecycle events (46011/46012) are still unemitted — same family, happy to add in the PR or a follow-up, your call.

[...]

PR #3327, opened 2026-07-28
feat(workflow): implement WF-08 approval gates

afkehaya:
## Summary

Implements WF-08: the missing bridge between the workflow executor's `StepResult::Suspended` and the existing approval grant/deny/resume infrastructure.

This PR ports proven approval-gate patterns from [Interchange](https://github.com/faremeter/interchange) into Buzz's Nostr-native workflow engine as an open-source contribution. Interchange's workflow state-machine and inference gate layers already solved durable approval gating with battle-tested patterns — rather than reinventing from scratch, this adapts those concepts for relay-signed Nostr events and NIP-33 parameterized-replaceable semantics.

### Relationship to existing work

- **Issue #2376** — the WF-08 design/implementation issue. This PR is an independent implementation of that spec.
- **PR #2377** — an existing open PR by @brocoppler implementing the same WF-08 gap. We overlap significantly on the core wiring (persist in `finalize_run`, `ActionSink::emit_approval_requested`, `hash_approval_token` promoted to pub). Our PR differs in: the race condition fix (status-before-record write order with rollback), definition-time approver spec validation in `schema.rs`, the `resolve_channel` helper extraction, and the Interchange-informed design decisions documented below. Happy to consolidate or defer to whichever approach maintainers prefer.
- **Issue #2878** — bug report that the documented `from:` syntax is rejected at approval time. Our PR fixes this: `schema.rs validate()` now rejects unsupported specs at definition time, and all test fixtures use the accepted `"any"` format.
- **Issue #2830** — grant binds run + step index, not the step. Not addressed in this PR.
- **PR #3301** — broader workflow engine completion PR (all 7 actions). May overlap on the `request_approval` action.

No prior issue was opened for this specific contribution — none found that covers the Interchange pattern-porting angle specifically.

### What it does

- **Suspended trace entry** — when the executor hits a `request_approval` step, it generates a CSPRNG token and returns `Suspended` with a redacted trace entry (the raw token never appears in traces)
- **`emit_approval_requested` on ActionSink** — new trait method + `RelayActionSink` impl that builds a kind:46010 Nostr event with `d`/`h`/`p`/`buzz:workflow` tags, persists it, and fans out to channel subscribers
- **`persist_approval_gate` in `finalize_run`** — replaces the WF-08 stub with the real pipeline: fetch workflow metadata, transition run to `WaitingApproval`, create approval record, emit kind:46010 (best-effort)
- **Approver spec validation** — `schema.rs validate()` now rejects unsupported `from:` formats at definition time (only `""`, `"any"`, and 64-char hex pubkeys are accepted)
- **Race condition fix** — flipped write order in `persist_approval_gate` so status transitions before record creation, with rollback on failure
- **`resolve_channel` helper** — extracts ~45 lines of shared validation between `send_message` and `emit_approval_requested`
- **7 E2E tests** — grant resumes, deny cancels, kind:46010 emission, no-approval regression, expired token rejection, wrong approver rejection, double-grant idempotency

### Key design decisions from Interchange

- **Durable-before-transmit ordering** (from Interchange's `packages/workflow/src/state-machine/`) — persist the approval record and run status to Postgres before emitting the notification event. If event emission fails, the approval is still actionable via the token hash.
- **Absolute-timeout design** (from Interchange's `packages/inference/src/gates.ts`) — store a computed `expires_at` timestamp rather than running countdown timers. The grant handler checks `expires_at < now` with no background jobs needed.
- **Status-before-record write order** — transition the run to `WaitingApproval` before creating the approval record, so a fast grant sees the correct status. If record creation fails, a rollback marks the run as `Failed`.

### Quality process

The implementation went through two critique loops (12 findings, all resolved), a three-reviewer simplification pass (reuse, quality, efficiency — 6 improvements applied), and a final fix round for the race condition, definition-time validation, and missing E2E coverage.

## Test plan

- [x] `cargo test -p buzz-workflow` — 156 tests pass
- [x] `cargo clippy -p buzz-workflow -p buzz-relay -p buzz-test-client -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo check -p buzz-test-client --tests` — 7 E2E tests compile
- [ ] E2E tests against live relay (`cargo test --test e2e_workflow_approval -- --ignored`)

### DIFF
--- a/crates/buzz-db/src/workflow.rs
+++ b/crates/buzz-db/src/workflow.rs
@@ -2213,7 +2213,7 @@ mod tests {
                 run_id: run_a,
                 step_id: "gate",
                 step_index: 0,
-                approver_spec: "@anyone",
+                approver_spec: "any",
                 expires_at: expires,
             },
         )
@@ -2228,7 +2228,7 @@ mod tests {
                 run_id: run_b,
                 step_id: "gate",
                 step_index: 0,
-                approver_spec: "@anyone",
+                approver_spec: "any",
                 expires_at: expires,
             },
         )
--- a/crates/buzz-db/src/workflow.rs
+++ b/crates/buzz-db/src/workflow.rs
@@ -30,7 +30,7 @@ pub const LIST_MAX_LIMIT: i64 = 1000;
 ///
 /// Approval tokens are stored hashed so that a DB read does not expose
 /// the raw token (same pattern as API tokens in buzz-auth).
-fn hash_approval_token(token: &str) -> Vec<u8> {
+pub fn hash_approval_token(token: &str) -> Vec<u8> {
     Sha256::digest(token.as_bytes()).to_vec()
 }
 
--- a/crates/buzz-relay/src/workflow_sink.rs
+++ b/crates/buzz-relay/src/workflow_sink.rs
@@ -365,22 +365,18 @@ impl ActionSink for RelayActionSink {
 
     fn emit_approval_requested(
         &self,
-        community_id: CommunityId,
-        channel_id: &str,
-        token_hash_hex: &str,
-        approver_spec: &str,
-        message: &str,
-        workflow_id: &str,
-        run_id: &str,
-        author_pubkey: &str,
+        params: buzz_workflow::ApprovalRequestParams,
     ) -> Pin<Box<dyn Future<Output = Result<String, ActionSinkError>> + Send + '_>> {
-        let channel_id = channel_id.to_owned();
-        let token_hash_hex = token_hash_hex.to_owned();
-        let approver_spec = approver_spec.to_owned();
-        let message = message.to_owned();
-        let workflow_id = workflow_id.to_owned();
-        let run_id = run_id.to_owned();
-        let author_pubkey = author_pubkey.to_owned();
+        let buzz_workflow::ApprovalRequestParams {
+            community_id,
+            channel_id,
+            token_hash_hex,
+            approver_spec,
+            message,
+            workflow_id,
+            run_id,
+            author_pubkey,
+        } = params;
 
         Box::pin(async move {
             // 0. Upgrade weak reference — fails only during shutdown.
@@ -436,8 +432,9 @@ impl ActionSink for RelayActionSink {
             let author_pubkey_hex = author_pk.to_hex();
 
             // 5. Validate workflow_id and run_id are valid UUIDs.
-            let _workflow_uuid = Uuid::parse_str(&workflow_id)
-                .map_err(|e| ActionSinkError::InvalidInput(format!("invalid workflow UUID: {e}")))?;
+            let _workflow_uuid = Uuid::parse_str(&workflow_id).map_err(|e| {
+                ActionSinkError::InvalidInput(format!("invalid workflow UUID: {e}"))
+            })?;
             let _run_uuid = Uuid::parse_str(&run_id)
                 .map_err(|e| ActionSinkError::InvalidInput(format!("invalid run UUID: {e}")))?;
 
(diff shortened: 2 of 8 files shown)


mfanafuthimhlanga (commented):
Read through this against #3525's caveat list. The shape looks right to me, and two decisions stand out as better than the obvious alternative.

Re-loading run → workflow → definition inside `persist_approval_gate`, rather than widening `ExecutionResult`, keeps the executor ignorant of persistence at the cost of two queries on a step that is about to block on a human. That seems like the right trade. And the `_ => "step {step_id} at index {step_index} is not a RequestApproval action"` arm is what actually defuses #3525's `step_index` caveat — an off-by-one becomes a loud definition error instead of a silent second token. Worth keeping even if the surrounding code moves.

Disclosure before the substance: I am looking at building a governance layer on top of approval gates, so I have a stake in point 2 below. Flagging that up front — the finding stands on its own, but you should know where I am coming from.

**1. I do not think the tests in this PR run in CI.**

`just test-unit`, which is what the Unit Tests job runs, covers `buzz-core`, `buzz-auth`, `buzz-voice`, `buzz-cli`, `buzz-db --lib`, `buzz-conformance`, and `buzz-push-gateway`. I could not find `buzz-workflow` in any `cargo test` or `nextest` invocation under `.github/workflows/` or `scripts/run-tests.sh`, and `buzz-relay`'s lib tests appear to run only when named by one of the four `-E` filters in the Backend Integration job.

I checked rather than assumed: on a fork I added a test, watched CI go green, then inverted the assertion so it had to fail — still green. They compile, because `cargo clippy --workspace --all-targets` builds test targets, but nothing executes them.

If that is right, the `schema.rs` changes here would pass or fail silently. Adding `cargo nextest run -p buzz-workflow --lib` to `just test-unit` made them run (154 tests, all passing). `-p buzz-relay --lib` is not safe to add wholesale — four tests in `api::admin` and `api::media` need Postgres but are not `#[ignore]d`, so they hang ~30s each and then fail. Is the gap known, or worth a separate PR?

**2. The `from` validation is a behavioral change worth calling out explicitly.**

`ActionDef::RequestApproval` documents the field as "User mention or role (e.g. `@release-manager`)" (`schema.rs:134`), and two existing tests encode that: `parse_workflow_with_all_actions` uses `@manager`, `parse_approval_gate_example` uses `@engineering-lead`. This PR rewrites both fixtures to `'any'` and adds `validate_rejects_at_anyone_approver_spec`, while the doc comment still promises mentions.

The new validator faithfully mirrors `check_approver_spec`, which already fails closed on role specs — so today the documented syntax parses, saves, and then can never be approved. That is #2878, and moving the failure from grant time to definition time is a real improvement.

I pinned both halves in CI to be sure of it rather than arguing from a read:

```
PASS buzz-workflow schema::tests::accepts_documented_mention_syntax_at_definition_time_see_issue_2878
PASS buzz-relay    handlers::command_executor::approver_spec_tests::rejects_documented_mention_syntax_see_issue_2878
```

My concern is that a plumbing PR ends up deciding, via a fixture edit, that mentions and roles are not coming to `from:`. Practically that leaves two options for anyone wanting a specific human to approve: hardcode a pubkey into every workflow definition, or use `"any"` and let anyone in the community approve.

So: is the empty / `"any"` / 64-hex allowlist intended as the final set, or is #2878 expected to widen it? If it is final, could the `schema.rs:134` doc comment change alongside it? If it is meant to widen, keeping those two fixtures `#[ignore]d` with a pointer to #2878 would preserve the record better than rewriting them — and I would be glad to take the resolution work (`from: "@handle"` and `from: "role:admin"` against channel membership) as its own PR. `workflow_sink.rs` already has `resolve_mention_pubkeys`, so the primitive may largely exist.

**3. #3525's double-approve caveat looks unaddressed.**

This does not touch `command_executor.rs`, so the `run.status != WaitingApproval` guards at `:1240` and `:1296` stay check-then-act rather than an atomic `WaitingApproval → Running`. I think it is safe in practice — `update_approval_by_stored_hash`'s `if !updated` serializes on the approval row, and a correct `step_index` means one run never has two live tokens — but that is three assumptions holding it up and none are written down. Worth a single-flight guard on `run_id`, or a comment recording why the row-level check suffices?

Two smaller things: `parse_duration_secs(timeout_str).unwrap_or(86400)` silently swallows a malformed timeout, which the new definition-time validation seems well placed to catch instead; and a workflow with no `channel_id` persists the gate but returns before emitting kind:46010, leaving the run suspended with nobody notified.

Happy to be wrong on any of these.

PR #3329, opened 2026-07-28
fix(desktop): paint own thread replies optimistically instead of waiting for the relay echo

tlongwell-block:
## The echo-chamber bug

The thread pane renders from the `threadRepliesKey(channelId, rootId)` React Query cache, whose **only live writer was the relay's WebSocket echo**. `useSendMessageMutation` optimistically painted the channel window on send but never touched the thread cache — so your own thread reply waited a full REST round-trip **plus** the relay's post-commit fire-and-forget fanout before appearing, while the main pane painted instantly. The thread pane absorbed 100% of echo lag. (Reported by Tyler in #buzz-bugs; traced independently by Eva and Wren.)

There is no separate thread subscription — both panes ride the same channel-wide WS sub. The asymmetry is purely the local send path.

Why no test ever caught it: the e2e mock relay echoed sends **synchronously inside the send command**, so any UI that incorrectly depended on the echo still painted "immediately" under test.

## Fix (desktop-only)

Mirror the channel window's optimistic lifecycle into the thread-replies cache:

- **onMutate** → `insertPendingThreadReply`: inserts the pending reply under the resolved thread root. Root resolution now searches cached thread subtrees too (`collectReplyParentContext`) — a nested reply's parent lives in the thread cache, not the channel timeline, and the old channel-only lookup would mis-root it.
- **onSuccess** → `reconcileSentThreadReply`: swaps pending → confirmed under the same `localKey` (no row remount), dedupe-safe against a WS echo racing ahead of the REST response (`mergeMessages` dedupes by id).
- **onError** → `rollbackPendingThreadReply`: targeted filter of the failed pending entry only — a snapshot restore would drop live replies that arrived mid-send.
- `useThreadReplies` fetch reconciliation now preserves still-pending optimistic sends so a refetch can't flash-remove an unconfirmed send.

## Tests

- New unit suite `threadOptimisticReply.test.mjs` (6 tests) covering insert/reconcile/rollback, nested-root resolution, and echo-race dedupe.
- New e2e `thread-own-send-liveness.spec.ts` using a new `sendEchoDelayMs` mock knob that defers the mock echo like the real relay's fanout. **Proven red pre-fix** (stash fix → both tests fail → restore → pass): own reply must paint before the echo can arrive, the late echo must not duplicate, and a reply sent while switching threads lands only in its own thread.
- Mock echo default stays synchronous (legacy behavior); the delay is opt-in per test.

## Verification

- typecheck + biome clean; full desktop unit suite 3727/3727 green.
- Full Playwright smoke project at this SHA: **759 passed, 3 failed** — all 3 unrelated and accounted for: `relay-reconnect` + `virtualization` pass on same-SHA rerun (flaky); `video-attachment` fails identically on the parent commit `4a977c588` (pre-existing, not introduced here).
- **Live-local pass** (TESTING.md): isolated Docker stack (project `buzz-echo-eva`, pg :53341, relay :3341, health :8341, `BUZZ_RECONCILE_CHANNELS=true`), seeded via `setup-desktop-test-data.sh`, real-relay Playwright run through `installRelayBridge`: own thread reply painted **< 2s** against the real relay, exactly one row after the real WS echo reconciled, and it survived a real thread refetch. Stored event carries the correct `["e", <root>, "", "reply"]` tag (verified via `/query`).

Buzz thread: `cdeaa1b1291b879fd1dcac941d7eb654df7699c71fe8125fc61cfca4ce9b5002` in #buzz-bugs (`e62570dd-33ad-42c5-b92b-75f2689f9694`).

Reviewer: @Wren (per Tyler).

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -75,6 +75,7 @@ export default defineConfig({
         "**/home-collapsed-top-chrome.spec.ts",
         "**/top-chrome-zoom-clearance.spec.ts",
         "**/thread-unread.spec.ts",
+        "**/thread-own-send-liveness.spec.ts",
         "**/workspace-rail.spec.ts",
         "**/community-rail.spec.ts",
         "**/boot-splash.spec.ts",
--- a/desktop/src/features/messages/hooks.ts
+++ b/desktop/src/features/messages/hooks.ts
@@ -1,5 +1,10 @@
 import { useEffect, useEffectEvent } from "react";
-import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query";
+import {
+  type QueryClient,
+  useMutation,
+  useQuery,
+  useQueryClient,
+} from "@tanstack/react-query";
 import { toast } from "sonner";
 
 import {
@@ -74,6 +79,9 @@ type MessageQueryContext = {
   previousWindow: ChannelWindowStore | undefined;
   channelId: string;
   queryKey: ReturnType<typeof channelMessagesKey>;
+  /** Set only for thread replies: the thread-replies cache the optimistic
+   *  message was inserted into, so onSuccess/onError can reconcile it. */
+  threadKey?: ReturnType<typeof threadRepliesKey>;
 };
 
 const CHANNEL_TIMELINE_KINDS = new Set<number>(CHANNEL_TIMELINE_CONTENT_KINDS);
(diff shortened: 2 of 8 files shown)


tlongwell-block:
Blocker: concurrent identical replies can delete an accepted reply from the thread cache.

At exact head `6b7dc42001c7dd52660a661df847881b4e55d3e6`, I added a reviewer-only permutation test with two pending replies having identical author/content/root and both REST successes plus both WS echoes. The ordering `echo1 -> success1 -> echo2 -> success2` finishes with only `real-2`; accepted `real-1` has been evicted.

`isMatchingPendingMessage` matches semantic fields rather than a send identity, so an echo can claim either indistinguishable pending row. Subsequent exact-id success reconciliation then rewrites local keys in a way that lets the later echo/success collide with and remove the other confirmed row.

Please preserve both accepted events for every REST/WS ordering and add a regression matrix for two identical pending replies. The channel-window suite already explicitly treats concurrent identical sends as distinct; the thread cache should retain that invariant.

Independent baseline at this SHA: full desktop unit suite 3727/3727; the two new Playwright smoke tests 2/2. Reviewer-only race suite: 7/8, failing on the ordering above. Worktree was restored clean after the probe.

tlongwell-block:
Blocker fixed in 6c3d472ffb4583aa2f4dfb4e2386774845260f69.

Root cause of the eviction: `reconcileIncomingMessage` ran semantic pending-matching unconditionally, so with two indistinguishable in-flight sends a later echo could claim a localKey already owned by a *confirmed* row, and the localKey-keyed dedupe then evicted that confirmed sibling (`echo1 → success1 → echo2 → success2` left only `real-2`).

Fix: key resolution is now ownership-aware (`resolveIncomingLocalKey`):
1. confirmed row with same id already exists → inherit its key, never re-match a pending;
2. caller-supplied localKey (onSuccess binding) is honored only if no other confirmed row owns it — if an echo claimed that pending first, fall through and claim a remaining pending;
3. otherwise claim the first semantically matching pending row.

Since claiming a pending removes it in the same merge, pending and confirmed rows never share a key, so a claimed key can never evict a confirmed row.

Regression: `concurrentIdenticalReplies_allEchoSuccessOrderings_keepBothAcceptedEvents` — full 4! echo/success interleaving matrix, asserting both accepted events survive with distinct render keys and no pending residue. Proven red against the previous merge logic (stash-verified in the same shell) and green with the fix.

Verified at 6c3d472ff: full desktop unit suite 3728/3728, both thread-liveness Playwright smoke specs 2/2, typecheck + biome clean, pre-push hooks (branch-skew, desktop-check, desktop-test) green. `messageMerge` is shared with the channel-window path, so the full-suite run covers that side too.

@wren ready for re-review.

tlongwell-block:
Re-review clear at exact head `6c3d472ffb4583aa2f4dfb4e2386774845260f69`.

The ownership-aware key resolution closes the reported eviction: same-id confirmed events retain their key; a supplied success key cannot displace a different confirmed owner; otherwise the incoming event claims a remaining semantic pending row. The landed 4! matrix passes.

Independent verification:
- Full desktop unit suite: 3728/3728.
- `pnpm build:e2e` plus targeted Playwright smoke: 2/2.
- Reviewer-only stronger matrix, restored afterward: all 6! orderings for three concurrent identical replies passed; all orderings of one failed identical sibling plus the other reply's echo/success passed. No pending residue, all accepted ids retained, distinct render keys.
- Local HEAD, `refs/pull/3329/head`, and `gh pr view` all matched the SHA above in the final shell; worktree clean and `git diff --check` clean.

No remaining code blocker from me. Ratings: minimalness 9/10, elegance 9/10, correctness 9/10. CI is still running on this fresh head, so merge remains gated on required checks rather than this review.

DanTup:
> `@wren ready for re-review.`

I'm curious if this was a deliberate mention, or a local bot name accidentally included in a comment posted to GitHub, causing notifications for an unrelated user.

PR #3333, opened 2026-07-28
chore(mobile): build iOS app in CI

brow:
### What changed?

Adds a `Mobile iOS Build` job that builds the Flutter iOS app.

It runs for iOS source and dependency changes, plus the `Justfile` recipe and workflow definition so the build plumbing tests itself.

The existing Mobile job and Android build are unchanged.

### Why?

Mobile CI already analyzes and tests the Flutter app and builds Android, but it never compiled the native iOS runner. Swift changes could therefore pass every required check without compiling.

Standard hosted macOS runners have no marginal billed cost for this public repository. The constraints are about five minutes of PR feedback and one of the 50 concurrent macOS jobs GitHub documents for this plan.

This check is deliberately advisory and is not in the required-checks ruleset; making it required remains an admin decision.

### How is it tested?

- [Deliberate Swift compiler failure](https://github.com/block/buzz/actions/runs/30378095404/job/90338856803): the job reached Xcode and failed on an injected Runner type error.
- [Final passing run](https://github.com/block/buzz/actions/runs/30379336640/job/90343231016): built `Buzz.app` at the exact PR head and passed the post-build lockfile guard.

The same commit built in 4m55s with a cold Hermit cache and 5m24s with a warm one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -859,34 +859,6 @@ jobs:
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
       - uses: cashapp/activate-hermit@cea9af7913204a965fd488637a8d1811bba2e616 # v1
-      - name: Compute Hermit cache key
-        id: hermit-bin-hash
-        run: |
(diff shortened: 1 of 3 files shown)


brow:
@codex review

brow:
@builderbot review

Issue #2391, opened 2026-07-22
feat: Add Linux ARM64 (aarch64) release artifacts and .deb packages


beingminimal:
### Issue Description

**Is your feature request related to a problem? Please describe.**
Currently, pre-built binary releases and `.deb` installation packages are not provided for Linux ARM64 (`aarch64`) architectures. Users running Linux on ARM-based hardware (such as Raspberry Pi 4/5, Apple Silicon Macs via Linux VMs, or AWS Graviton instances) cannot easily install or run Buzz without building from source manually.

**Describe the solution you'd like**
Add Linux ARM64 target support to the automated CI/CD release workflow to build, package, and upload the following artifacts alongside existing releases:

1. Standard Linux ARM64 binary archive (`.tar.gz` or release binary)
2. Linux ARM64 Debian package (`.deb`)

**Describe alternatives you've considered**
Building from source manually on ARM64 devices, which requires setting up the full build environment, dependencies, and compilation toolchains on each local target machine.

**Additional context**

* Expanding build matrix targets to include `linux/arm64` / `aarch64` in GitHub Actions (or equivalent CI pipelines).
* Ensuring `.deb` package creation includes the correct `Architecture: arm64` control metadata.

scottybo:
I hit this today on an ARM64 Linux box and can confirm source builds work, but the release artefacts do not cover this setup.

Environment tested:
- Ubuntu 24.04.3 on `aarch64`
- Buzz `v0.5.0`
- Release page currently includes Linux `amd64` `.AppImage` and `.deb`, but no Linux `aarch64` artefact
- Built from source successfully and ran the web/source runtime with `buzz-relay` + `buzz-acp` against OpenClaw ACP

The missing packaged desktop artefact was the only blocker for a normal desktop install path on this machine.

While looking at `.github/workflows/release.yml`, I noticed the Linux release job is already close to being arch-aware:
- `Install appimagetool` switches on `uname -m` and has an `aarch64` branch.
- It currently fails intentionally for non-x86_64 because the `appimagetool-aarch64` and `runtime-aarch64` SHA256 pins are not filled in.
- The current job runs on `ubuntu-latest`, so it only produces the x86_64 Linux artefacts.

Potential implementation path now that GitHub-hosted ARM runners are available:
- add a Linux ARM64 release job on `ubuntu-24.04-arm` or `ubuntu-22.04-arm`
- keep the existing x86_64 Linux job unchanged for updater compatibility
[...]

PR #3339, opened 2026-07-28
ci: add Linux ARM64 release artifacts

scottybo:
## Summary
- add a separate Linux ARM64 release job using GitHub's `ubuntu-24.04-arm` runner
- pin SHA256 values for the aarch64 `appimagetool` and type2 runtime assets
- upload ARM64 Linux `.deb` and `.AppImage` artefacts to the versioned release
- leave the existing x86_64 Linux updater/latest.json path unchanged until ARM64 updater behaviour is validated

Refs #2391

## Notes
I hit this while trying to run the v0.5.0 desktop package on an Ubuntu 24.04.3 `aarch64` machine. Building from source worked, but the release only had Linux `amd64` artefacts.

This PR is intentionally conservative: it adds installable ARM64 release artefacts without changing the existing Linux auto-update manifest contract.

## Verification
- YAML parse check: `python3 -c 'import yaml; yaml.safe_load(open(".github/workflows/release.yml"))'`
- `git diff --check`
- SHA256 values computed from the pinned upstream assets:
  - `appimagetool-aarch64.AppImage`: `f0837e7448a0c1e4e650a93bb3e85802546e60654ef287576f46c71c126a9158`
  - `runtime-aarch64`: `00cbdfcf917cc6c0ff6d3347d59e0ca1f7f45a6df1a428a0d6d8a78664d87444`

I have not run the full release workflow locally; this needs maintainer CI validation.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -713,6 +713,240 @@ jobs:
           ARCHIVE_PATH: ${{ steps.linux-artifacts.outputs.archive }}
           SIG_PATH: ${{ steps.linux-artifacts.outputs.sig }}
 
+
+  # Build Linux ARM64 artefacts separately so x86_64 remains the only Linux
+  # updater target until the updater manifest has been validated for ARM64.
+  release-linux-arm64:
+    name: Release Linux (ARM64)
+    if: github.repository == 'block/buzz'
+    runs-on: ubuntu-24.04-arm
+    # Digest-pinned like the SHA-pinned actions below; Renovate keeps it fresh.
+    container: ubuntu:22.04@sha256:0e0a0fc6d18feda9db1590da249ac93e8d5abfea8f4c3c0c849ce512b5ef8982
+    needs: setup
+    timeout-minutes: 60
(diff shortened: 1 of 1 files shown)


scottybo:
Addressed the zizmor hash/comment finding by updating the newly-added Linux ARM64 job pins so their commit SHAs match the version comments:

- `actions/checkout` `# v6` now points at `d23441a48e516b6c34aea4fa41551a30e30af803`.
- `Swatinem/rust-cache` `# v2` now points at `42dc69e1aa15d09112580998cf2ef0119e2e91ae`.

I also amended the commit with a DCO sign-off; the DCO check is now passing.

scottybo:
Fixed the remaining zizmor finding on `Swatinem/rust-cache`.

The previous update used the annotated tag object SHA for `v2` (`42dc69e...`). I have changed it to the peeled commit SHA (`e18b497...`), matching the existing pin used elsewhere in this workflow.

Re-ran:
- `git diff --check`
- YAML parse check with Python/PyYAML

PR #3340, opened 2026-07-28
fix(acp): preserve truncated thread context

cameronhotchkies:
## Why
Long Buzz threads were rendered as `[Thread Context (13 of 13 messages)]` because the harness counted only the already-limited query result. That hid older context and could also hide the agent's own prior reply in busy threads.

## What
- Fetch one extra thread reply as a sentinel so truncated context is labeled correctly.
- Use a best-effort `/count` call for improved truncated totals when available, clamped to the sentinel-proven minimum so racy counts cannot render impossible labels.
- Keep the `/count` path single-attempt with a short timeout and only add the root to exact totals when the root was actually fetched.
- Fetch and preserve the agent's newest prior reply when it falls outside the recent window, with exact event-id matching for the pin/dedup boundary.
- Add parser and fetch-boundary tests for truncation, exact count, missing root, count-below-minimum clamping, count failure fallback, distinct fetched-reply lower bounds, agent-reply dedup/pinning, and serialized query/count filter semantics.

## Risk Assessment
Low-to-medium — limited to buzz-acp prompt context fetching and a small RestClient helper. If `/count` fails or times out, the code falls back to the sentinel-derived minimum total rather than failing the prompt. The synchronous `/count` happens only for truncated thread contexts and is bounded to one short best-effort attempt.

## References
- Buzz thread: chotchkies-buzz-bombing-flakes / `7ef71407f1c7a642382c7e48e0c80fb6ca66948890e04d1eb6f1408c3b7278b1`
- Validation at `c1cfd1b16a04a3ac1d1d0d3cf43e1a08508f3532`:
  - `cargo fmt -p buzz-acp` ✅
  - `cargo test -p buzz-acp test_fetch_thread_context -- --nocapture` ✅ (6 tests)
  - `cargo test -p buzz-acp parse_nostr_thread_response` ✅
  - `cargo test -p buzz-acp` ✅ (649 unit + 9 lifecycle tests)
  - `git diff --check` ✅
  - Push was completed with `--no-verify` after pre-push hooks reached non-code local environment failures: `flutter` missing for `mobile-test`; Node.js v20.20.2 too old for pnpm/node:sqlite in `desktop-check` and `desktop-test`. Earlier hook stages passed: `check-push-org`, `branch-skew`, `rust-tests`, `test`, `desktop-tauri-checks`.
- Earlier full `./bin/just ci` at `622ed7eb8807d64e06209101569b1013414af091` ⚠️ passed Rust/desktop/web stages, then failed in `mobile-test` on unrelated existing mobile test `ChannelDetailPage keeps follow mode off while a tall newest message stays visible`; rerunning that single mobile test reproduced the same failure without touching mobile code.

Generated with Codex

### DIFF
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -19,6 +19,7 @@
 //!
 //! `AcpClient` is NOT Clone — ownership moves out on claim and back on return.
 
+use std::cmp::Reverse;
 use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
@@ -799,6 +800,9 @@ pub enum IdleSwitchResult {
 /// Each call gets this budget; with one retry the total worst-case is
 /// 2 × CONTEXT_FETCH_TIMEOUT + CONTEXT_FETCH_RETRY_DELAY ≈ 6.5 s.
 const CONTEXT_FETCH_TIMEOUT: Duration = Duration::from_millis(3_000);
+
+/// Short, single-attempt timeout for best-effort exact truncated-thread counts.
+const CONTEXT_COUNT_TIMEOUT: Duration = Duration::from_millis(500);
 
 /// Delay between the first failed context fetch and the single retry.
 const CONTEXT_FETCH_RETRY_DELAY: Duration = Duration::from_millis(500);
@@ -2600,7 +2604,14 @@ async fn fetch_conversation_context(
     let last_event = batch.events.last()?;
     let tags = crate::queue::parse_thread_tags(&last_event.event);
     if let Some(root_id) = tags.root_event_id {
-        return fetch_thread_context(batch.channel_id, &root_id, limit, &ctx.rest_client).await;
+        return fetch_thread_context(
+            batch.channel_id,
+            &root_id,
+            limit,
+            ctx.agent_keys.public_key(),
+            &ctx.rest_client,
+        )
+        .await;
     }
 
     // DM non-reply: fetch recent conversation history.
@@ -2762,12 +2773,48 @@ async fn fetch_prompt_profile_lookup(
 }
 
 /// Fetch thread context via Nostr query: root event by ID + replies by `#e` tag.
+///
+/// The reply query intentionally requests one more reply than the configured
+/// display window. That sentinel event lets the prompt say `N of M, truncated`
+/// when the relay has more thread history, instead of reporting the capped page
+/// as the total. When the window is full, a best-effort `/count` attempts to
+/// improve that lower-bound total; because it is a separate racy request, the
+/// result is clamped to the sentinel-proven minimum. The query also asks for the
+/// agent's newest reply separately so the next prompt can include the agent's
+/// own prior turn even in busy threads where the recent-message window would
+/// otherwise push it out.
 async fn fetch_thread_context(
     channel_id: Uuid,
     root_event_id: &str,
     limit: u32,
+    agent_pubkey: nostr::PublicKey,
     rest: &RestClient,
 ) -> Option<ConversationContext> {
+    fetch_thread_context_with(
+        channel_id,
+        root_event_id,
+        limit,
+        agent_pubkey,
+        |filters| async move { rest.query(&filters).await },
+        |filters| async move { rest.count(&filters).await },
+    )
+    .await
+}
+
+async fn fetch_thread_context_with<Query, QueryFut, Count, CountFut>(
+    channel_id: Uuid,
+    root_event_id: &str,
+    limit: u32,
+    agent_pubkey: nostr::PublicKey,
+    query: Query,
+    count: Count,
+) -> Option<ConversationContext>
+where
+    Query: Fn(Vec<nostr::Filter>) -> QueryFut,
+    QueryFut: std::future::Future<Output = Result<serde_json::Value, crate::relay::RelayError>>,
+    Count: Fn(Vec<nostr::Filter>) -> CountFut,
+    CountFut: std::future::Future<Output = Result<serde_json::Value, crate::relay::RelayError>>,
+{
     use nostr::{Alphabet, SingleLetterTag};
 
     // Defense-in-depth: validate hex event ID.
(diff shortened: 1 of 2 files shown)


crates/buzz-acp/src/pool.rs
```diff
@@ -2793,6 +2843,70 @@ async fn fetch_thread_context(
...
+                    tracing::debug!(
+                        channel_id = %channel_id,
+                        "thread context count timed out — will retry"
+                    );
+                    return None;
+                }
+            };
+
+        // +1 for the root event fetched by ID alongside matching replies.
+        Some(replies_count as usize + 1)
```
wpfleger96: 🤖 I think this `+1` can invent a root that isn't there. `parse_nostr_thread_response` explicitly tolerates a missing root and returns replies-only context (its own `fetched_total` uses `usize::from(root_msg.is_some())`), and that state is reachable — a soft-deleted root is excluded by both `/query` and `/count` (`deleted_at IS NULL`), while its replies remain queryable. In that case the exact-count path reports one more message than actually exists, and the total flips inconsistently at the truncation boundary depending on whether the count ran. Maybe carry a `root_present` flag out of the parse (or count the root filter too) and only add 1 when the root was actually seen, plus a truncated missing-root test.

crates/buzz-acp/src/pool.rs
```diff
@@ -2793,6 +2843,70 @@ async fn fetch_thread_context(
...
+async fn fetch_thread_total<Count, CountFut>(
+    channel_id: Uuid,
+    replies_filter: &nostr::Filter,
+    count: &Count,
+) -> Option<usize>
+where
+    Count: Fn(Vec<nostr::Filter>) -> CountFut,
+    CountFut: std::future::Future<Output = Result<serde_json::Value, crate::relay::RelayError>>,
+{
+    fetch_with_retry(|| async {
```
wpfleger96: 🤖 Wrapping the count in `fetch_with_retry` + `CONTEXT_FETCH_TIMEOUT` means a slow or degraded `/count` can stall a truncated-thread prompt for up to another ~6.5s (2 × 3s attempts + 500ms delay) after the message query already succeeded — and `bridge_post()` has its own transport retries inside each bounded attempt. That doubles the documented worst-case thread-context fetch, and it lands exactly on the long threads this PR is trying to fix. Since the sentinel lower bound is already correct and this is only label precision, I'd make it genuinely best-effort: one attempt under a short dedicated timeout, no outer retry, fall back to the sentinel minimum.

crates/buzz-acp/src/pool.rs
```diff
@@ -2946,48 +3060,93 @@ fn json_to_context_message(obj: &serde_json::Value) -> Option<ContextMessage> {
...
+    if truncated {
+        // The relay returns limited REQ results newest-first. Sort explicitly so
+        // the sentinel we drop is the oldest reply in the fetched window, not an
+        // arbitrary last element if the HTTP bridge ever changes iteration order.
+        reply_msgs.sort_by_key(|(ts, _, _)| Reverse(*ts));
+        reply_msgs.truncate(limit as usize);
+    }
+
+    if let Some(agent_reply) = newest_agent_reply {
+        let agent_reply_already_displayed = reply_msgs.iter().any(|(ts, is_agent, msg)| {
```
wpfleger96: 🤖 nit: the pin check identifies the agent reply by `(created_at, pubkey, content)` because the tuple drops the event id after parsing. Carrying the id through the tuple would make this exact instead of content-equality — cheap and removes a theoretical false match between same-second identical-content replies.

crates/buzz-acp/src/pool.rs
```diff
@@ -4136,6 +4295,477 @@ mod tests {
...
+        ]);
+
+        let ctx = fetch_thread_context_with(
+            channel_id,
+            root_id,
+            2,
+            agent.public_key(),
+            move |filters| {
+                assert_eq!(
+                    filters.len(),
```
wpfleger96: 🤖 nit: this `filters.len() == 3` is the only filter assertion across the fetch-boundary tests — they'd still pass if the sentinel `limit+1`, the `#e`/`#h` constraints, the agent author filter, or the count filter scope regressed, since the mocks return a canned response regardless. One boundary test asserting the serialized filter semantics (for the three query filters and the count filter) would lock the behavior in.

wpfleger96 (commented):
🤖 Combined review from my agents (two independent passes, deduped). Two things I think are worth fixing before merge — the phantom-root `+1` in the exact-count path and the retry policy around the best-effort `/count` — plus a couple of nits inline. The core design looks right to me: the sentinel lower bound, the clamp against racy counts, the dedup across OR'd filters, and the agent-reply pinning are all cleanly separated and well covered by the new tests.

wpfleger96 (approved):
Re-reviewed exact head `c1cfd1b16a04a3ac1d1d0d3cf43e1a08508f3532`. All four prior comments are addressed:

- Exact totals add the root only when `ParsedThreadContext.root_present` confirms it was fetched.
- Best-effort `/count` has no outer retry and is bounded by the dedicated 500 ms timeout.
- Reply deduplication and agent-reply pinning compare exact event IDs.
- Fetch-boundary tests assert the serialized root, recent-reply, agent-reply, and count filter semantics.

The count result is also clamped to the sentinel-proven minimum, closing the deletion race that could render impossible labels.

Quality gate:
- Minimalism: 9/10 — the behavior is contained to the thread-context path and one reusable HTTP bridge helper; the fallback remains the already-fetched sentinel minimum.
- Elegance: 9/10 — parsing metadata, display-window selection, and optional exact counting have distinct responsibilities and readable invariants.
- Correctness: 9/10 — missing roots, count failures, low racy counts, distinct fetched replies, exact-ID dedup/pinning, and filter construction are covered.

Independent exact-head validation passed: `cargo test -p buzz-acp` (649 unit + 9 lifecycle tests), `cargo check -p buzz-acp --all-targets`, `cargo fmt -p buzz-acp -- --check`, and `git diff --check`. The worktree remained clean at the same SHA; GitHub checks are green (Mobile/Web intentionally path-filtered).

PR #3345, opened 2026-07-28
feat(desktop): recommend mesh models for local hardware

sandro-sq:
## Summary

- adds an intentional hardware diagnostic to **Settings → Compute → Share compute**
- recommends a model using mesh-llm's hardware-aware model-pack selection, including AI-memory and free-disk fit
- presents a staged hardware scan and focused recommendation before exposing manual model controls
- persists accepted recommendations and adds clearer model download/startup diagnostics

## Testing

- `just mesh-e2e`
- pre-push hooks: desktop checks, desktop tests, and Tauri tests
- manually tested on an Apple M5 Max with 128 GB unified memory

### Smart hardware recommendation

Smart hardware recommendation

### End-to-end inference

After accepting the recommendation and enabling Share compute, a Buzz agent successfully responds through the local mesh runtime.

Successful inference through shared compute

## Related issues

None.


michaelneale:
@sandro-sq I think this is the right approach - I like it. 
while we are in there touching that buzz/server code, should we tweak it so if you shift servers, it keeps sharing on the one you were sharing, but the GUI just says (it is sharing with community X...) if you go to that page? that would be I think the least annoying thing vs now where I think it stop it when you switch?

sandro-sq:
> @sandro-sq I think this is the right approach - I like it. while we are in there touching that buzz/server code, should we tweak it so if you shift servers, it keeps sharing on the one you were sharing, but the GUI just says (it is sharing with community X...) if you go to that page? that would be I think the least annoying thing vs now where I think it stop it when you switch?

FYI Fixed this in the `8342d52` commit

PR #3347, opened 2026-07-28
fix(mobile/ios): stop dropping push notifications during catch-up

brow:
### What changed?

Rewrites how the iOS notification service extension decides what to show when a push wake arrives.

Before, each wake fetched a bounded batch per community, displayed exactly one event, and then advanced the read watermark for every community that produced a candidate. Everything else was consumed silently and could never be shown again. This replaces that with an explicit traversal: candidates are selected oldest-first, pagination continues only from tails the relay actually returned, and the forward floor advances only after traversal proves coverage of every older lease-authorized event above the previous floor. When the page budget or the extension deadline is hit, the floor stays put and the in-progress scan is persisted so the next wake resumes it instead of restarting.

Also replaces the extension's hand-rolled notion of "interesting" with the push lease itself, so channel and author constraints are applied by the same authority that caused the wake, and emits one filter per channel rather than batching channels into a single filter.

### Why?

Notifications were being dropped in normal use, not in edge cases: any wake carrying more than one new event permanently lost all but one of them, and events the extension chose not to display were marked read anyway. Duplicates were the visible symptom, silent loss was the more serious one.

One known gap remains and is documented in the code rather than papered over: the relay's `/query` exposes no exhaustion signal, so a short page is still treated as exhausted even though post-limit filtering makes that inference unsound. Fixing it properly requires a relay change, tracked separately.

### How is it tested?

Build and run, plus the extension's Swift package suite and the Flutter mobile suite.

Added tests:

- `PushCatchUpTests`, `PushConsumptionStateTests`, `PushLeaseTests`, `PushNotificationIdentityTests`
- `push_subscription_test.dart`, `push_subscription_provider_test.dart`, `push_models_test.dart`

Traversal behavior was additionally checked against purpose-built corpora and mutation-tested: each mutant was first confirmed to change observable behavior, then confirmed to be caught.

### Stacking

Based on `push-ios-blockers-1-2` (#2744), which is itself behind `main` and conflicts with it in `Runner/AppDelegate.swift` and `Flutter/Debug.xcconfig`. That merge belongs to #2744, not here.


brow:
@codex review

PR #2527, opened 2026-07-23
relay: support host aliases so in-cluster clients can bind to a community

jedwards27:
# relay: support host aliases so in-cluster clients can bind to a community

## Problem

The relay binds every request to a community strictly by the request `Host` header against `communities.host`. In some deployments, clients running inside the same cluster as the relay can only reach it via an internal service DNS name (e.g. `relay.<namespace>.svc.cluster.local:3000`) — the public, edge-fronted hostname isn't routable from inside the cluster. That internal host has no community mapping, so every in-cluster connection fails closed with:

```
404 relay: no community is configured for this host
```

Client-side workarounds don't exist:
- Spoofing the public `Host` header doesn't work when a service mesh in front of the relay routes by Host — the request is rejected before the relay ever sees it.
- WebSocket connections can't override `Host` at all, so even a relay-side forwarded-host hack wouldn't cover the primary (WS) traffic.

This blocks any agent or client hosted inside the cluster from connecting to the relay.

## Design

Adds a "CNAME equivalent": one community reachable via multiple hostnames.

- **New table `community_host_aliases`** (`migrations/0025_community_host_aliases.sql`, mirrored in `schema/schema.sql`): `(host, community_id → communities.id)`, unique on `lower(host)`, with guards preventing an alias colliding with any `communities.host` (and vice versa on community create — collision surfaces as a distinguishable error, not a silent failure).
- **Resolver fallback**: the `HostResolver` DB implementation resolves against `communities.host` first, then `community_host_aliases`. Fail-closed semantics preserved — unmapped hosts still return the same generic 404; **no default tenant introduced**.
- **Tenant binding invariants preserved**: `TenantContext` still binds the *request* host (aliases do not rewrite `tenant.host()`), so NIP-98 `u`-URL matching and AUTH origin checks continue to compare against the host the client actually used. Invariant comments in `tenant.rs` / `api/bridge.rs` updated accordingly.
- **Admin surface**: `buzz-admin community alias add|remove|list`, managing aliases for the tenant resolved from `RELAY_URL` (admin tenant resolution also resolves through aliases now).

## Operational rollout

After deploy, map the internal host to the existing community:

```
buzz-admin community alias add relay.<namespace>.svc.cluster.local:3000
```

No client changes needed — in-cluster clients keep using the internal service URL.

## Testing

- Unit tests: alias resolves to the same community; unknown host still fails closed; alias/primary-host collisions rejected in both directions (including the `create_community_with_owner` path).
[...]

PR #3372, opened 2026-07-28
feat(relay): serve the relay under an optional BUZZ_BASE_PATH prefix

### LINKED ISSUE #3354: Relay cannot be served under a path prefix — hardcoded root routes block deployment behind a path-routing gateway
## Problem

`buzz-relay` mounts every route at the root. The WebSocket upgrade and the NIP-11 document both live on `/` (`crates/buzz-relay/src/router.rs:64`), with the rest of the HTTP surface as siblings. There is no configuration to serve the relay under a path prefix — the only `nest()` call in `build_router` is for the admin sub-router.

That makes the relay undeployable behind a gateway that routes to apps by path rather than by hostname. Such gateways are common in shared-ingress Kubernetes setups: each app is assigned a path prefix, auth policy is applied per prefix, and the root is not available to any individual app. A relay that can only serve at `/` has no place to go.

Compose behind a dedicated hostname works, so this is not a blocker for every self-hoster — but anyone whose ingress is path-routed currently has no option except a fork.

## Proposal

Add a `BUZZ_BASE_PATH` config option (empty by default, so current behavior is unchanged). When set, nest the merged router under that prefix so the WebSocket lands at `wss://host/<prefix>` and every HTTP route under `wss://host/<prefix>/...`.

Health probes (`/health`, `/_liveness`, `/_readiness`) should stay mounted at the root as well, since Kubernetes probes hit the pod directly rather than through the gateway.

## Things to get right

- **NIP-98 `u` tags** — auth events sign the full request URL, and the server compares against the URL it received. The prefix has to be present consistently on both sides or auth fails in a way that looks like a signature bug.
- **Client-side URL derivation** — `relayHttpFromWs` (`desktop/src/shared/api/inviteHelpers.ts:80`) is a pure scheme swap, so a prefix carries through. `mediaUrl.ts:49` compares `new URL().origin`, which drops the path — still matches, just coarser. `buzz-cli` and the mobile client need the same audit.
- **Git smart HTTP** and the static web assets, both of which build URLs of their own.
- Nostr relay URLs may carry a path, so `wss://host/relay` is protocol-legal — no NIP-level obstacle.

## Willing to implement

Happy to open a PR if the approach and the config name look right.

kcao-gss:
Closes #3354.

## Why

`buzz-relay` mounted every route at the root, so it could not be deployed behind a gateway that routes to apps by path rather than by hostname. That shape is common in shared-ingress Kubernetes setups: each app is assigned a path prefix, auth policy is applied per prefix, and the root belongs to the gateway rather than to any one app. A relay that can only serve at `/` has nowhere to go.

Compose behind a dedicated hostname works, so this is not a blocker for every self-hoster — but a path-routed ingress previously left no option except a fork.

## What

`BUZZ_BASE_PATH`, empty by default. Unset, the router is byte-identical to before. When set, the whole surface nests under the prefix — the WebSocket and NIP-11 handshake answer on `wss://host/<prefix>`, the HTTP routes live beneath it. `nest` strips the prefix before the inner router sees the request, so route matches and the SPA fallback's own path checks keep working unchanged.

Health probes stay mounted at the root as well: Kubernetes reaches the pod directly rather than through the gateway that needs the prefix, and `deploy/compose` curls `/_liveness` on the published port.

The prefix is normalized at load time (one leading slash, no trailing slash — `relay`, `/relay`, `/relay/` all become `/relay`). Values that would make the server's URL reconstruction differ from the client's connect URL — query, fragment, whitespace, dot segments, empty interior segments — fail startup rather than surfacing later as an unexplainable 401.

## The part that needed care

The prefix has to reach every place a URL is rebuilt from the request host, because each one is signed and compared byte-for-byte:

| Site | Signed value |
|---|---|
| `nip98_expected_url` | the `u` tag for bridge, invite, and moderation calls — one choke point for all of them |
| `nip42_expected_relay_url` | the WebSocket AUTH `relay` tag |
| `authorize_operator_request` | the operator API's expected URL |
| NIP-11 `push.origin` | dialed verbatim by push clients |

Miss one and it reads as a signature bug, not a routing bug.

## Client-side sweep

Three places assumed the relay owned the origin root:

- **`buzz-cli`** rejected a prefixed media URL as "not a /media/ path"; it now strips the relay URL's own path before the check. The sha-shorthand form was already correct.
- **Desktop** keyed the Download affordance on a root-anchored `/media/` pathname. Now matches `/media/` as a full path segment anywhere, still origin-pinned, so `/media-evil/<hash>` is still rejected.
- **Mobile** used `Uri.resolve('/query')` in three places, which discards a base path outright. These now go through a shared `relayEndpoint` helper.

One existing mobile test had codified the prefix-dropping behavior against a base URL of `https://relay.example/base`. Updated — and its NIP-98 `u` tag assertion now reads `https://relay.example/base/query`, which is the cross-check that the client and the relay agree on the prefixed URL.

Verified clean: `relayHttpFromWs` (desktop) is a pure scheme swap and `normalize_relay_url` (CLI) is a scheme swap plus trailing-slash strip, so both carry a prefix through already.

## Known limitation

The browser web bundle (`BUZZ_WEB_DIR`) references its assets at absolute `/assets/...` URLs baked in at build time, so serving that SPA under a prefix additionally requires rebuilding it with a matching Vite `base`. Documented on the config field. The protocol surface — WebSocket, bridge, media, git — is unaffected.

## Testing

17 new tests: prefix normalization and rejection, the default-empty path, router resolution under single- and multi-segment prefixes, root passthrough for probes, 404 for unprefixed application paths, both signed-URL builders, the push descriptor, the CLI media URL, and the mobile endpoint helper.

Suites run: `buzz-relay` 811 lib tests, `buzz-cli` 253, desktop 3733, mobile 874. `cargo fmt --check`, clippy on both crates, `flutter analyze`, biome lint, and `cargo check --workspace --all-targets` all clean.

Nine relay tests and one mobile test fail in my environment; I confirmed each one fails identically on a clean tree. The relay ones are `Sqlx(PoolTimedOut)` — they need Postgres, which I did not have running.

## Compatibility

`BUZZ_BASE_PATH` unset is the default and changes nothing: same routes, same signed-URL strings, same NIP-11 document. Operators opting in must set the same prefix on `RELAY_URL` and `BUZZ_MEDIA_BASE_URL`.


kcao-gss:
CI here needs a maintainer to approve the workflow run (first-time fork contributor). To save you that round trip, I ran the same workflows on my fork: **kcao-gss/buzz#1** — 22 of 23 CI jobs green.

Green there, including everything I couldn't run locally:

- `Relay E2E`, `Backend Integration (relay e2e)` — the Postgres + Redis suites
- `Desktop E2E Relay`, both `Desktop E2E Integration` shards, all 4 `Desktop Smoke E2E`
- `Rust Lint`, `Unit Tests`, `Security`, `Windows Rust`, `Mobile`, `Desktop`, `Desktop Core`, `Web`
- Both musl server cross-compiles

Two jobs fail on the fork for permission reasons that won't reproduce here:

1. `Desktop Build (macOS)` — `mesh-llm checkout for 43103c5 not found after cargo fetch`. The private dep isn't fetchable with a fork token.
2. `Docker image` — the image builds, then `exporting cache to registry` hits `denied: permission_denied` writing `ghcr.io/block/buzz-buildcache`. `cache-to` is gated to same-repo, but the export still targets the org registry.

Neither touches code paths in this PR.

Locally I also ran the relay lib suite against a live Postgres: 778 passed, 0 failed.

Happy to rebase or split this if the shape isn't what you want before you spend CI minutes on it.

kcao-gss:
Per CONTRIBUTING, the closest existing work I could find — I'd missed this line when I opened the PR.

**Closest open PR: #2527** ("relay: support host aliases so in-cluster clients can bind to a community"). Same family of problem — making the relay reachable in a deployment where the public hostname doesn't apply — but the orthogonal axis. #2527 widens which *hosts* map to a community; this PR changes which *path* the relay answers on. No functional overlap, and no file overlap except `crates/buzz-relay/src/api/operator.rs`, where the two changes sit in different places: #2527 extends alias resolution, this adds the base-path prefix to the operator API's expected NIP-98 URL. A trivial rebase either way.

Worth noting the two compose: a deployment could use aliases to reach a community by internal DNS *and* a base path to sit behind a path-routing gateway.

Also adjacent but not overlapping: #2736 (bundles the pair-relay sidecar for deploy) and #2617 (documents `BUZZ_CORS_ORIGINS` in `.env.example`) — both deploy-config work that touches neither routing nor signed-URL construction.

No duplicate of this change found in open PRs or issues.

kcao-gss:
Pushed one more client-side fix, found by running a prefixed deployment for real rather than by reading code.

`parseInviteInput` (`desktop/src/shared/api/inviteHelpers.ts`) rejected invite URLs from a relay served under a base path, and would have derived the wrong relay URL even if it hadn't:

```js
const match = url.pathname.match(/^\/invite\/([^/]+)\/?$/);   // anchored at root
const relayWsUrl = `wss://${url.host}`;                        // path discarded
```

Both halves fail against `https://host/relay/invite/<code>`: the regex doesn't match, so the input is rejected as a non-invite URL, and had it matched, the derived relay URL would have been `wss://host` — a relay that isn't there.

Now captures an optional base path and carries it through, so `https://host/relay/invite/<code>` yields `{ relayWsUrl: "wss://host/relay", code }`. Four tests added covering single-segment, multi-segment, `http://` with a port, and a trailing slash. The existing rejection cases (credentials, fragments, bare codes, inputs with slashes) all still pass — 28/28 in that file, 3737/3737 across desktop, biome clean.

Worth noting how this one surfaced: my original audit checked `relayHttpFromWs` and `mediaUrl.ts` and cleared both, but I missed this because it reconstructs a relay URL from components instead of transforming one it was handed. That's the pattern to grep for if anyone wants to look further — `${...}${url.host}` style rebuilds rather than scheme swaps. The symptom in the field was a teammate getting a 404 from an invite link, which points nowhere near the parser.

PR #3377, opened 2026-07-28
feat(desktop): keyboard shortcut for next/previous channel in sidebar

### LINKED ISSUE #3078: feat(desktop): keyboard shortcut to navigate to next/previous channel
**Problem**

There's no way to move between channels from the keyboard. Every channel switch requires the mouse or a round-trip through quick search (`⌘K`) — typing a name you already know to reach a channel that's already visible in the sidebar. Slack, Discord, and IRC clients all have a single-chord next/previous-conversation binding, and its absence is felt immediately when working across a few adjacent channels.

**Proposal**

Add `⌥↓` / `⌥↑` (`Ctrl+Alt+↓` / `Ctrl+Alt+↑` on Windows) to move the active channel selection down/up the sidebar list.

- **Order** matches what the sidebar displays: starred → custom sections in section order → unassigned, each honoring its own sort preference (as assembled in `AppSidebar.tsx:395-433`).
- **Scope:** the active community's stream channels. Forums and DMs are out of scope here.
- **Muted channels are skipped.** `mutedChannelIds` is already available in `AppShell`, so this is a natural filter rather than a new data path.
- **Stops at both ends** rather than wrapping — no wraparound surprise when you're walking to the top or bottom of a list deliberately.
- No-op when the list is empty or no channel is selected (e.g. on the home feed).
- Registered in `shared/lib/keyboard-shortcuts.ts` under `Navigation`, so it appears in Settings → Keyboard Shortcuts automatically.

Windows uses `Ctrl+Alt+arrows` rather than Slack's plain `Alt+arrows` because `Alt+←`/`Alt+→` are already bound to back/forward there (`useBackForwardControls.ts:92-106`).

**Implementation sketch**

The displayed-order flattening currently lives inline in `AppSidebar.tsx`. I'd extract it into a shared helper so the shortcut and the sidebar can't drift apart, cover it with a `.test.mjs` unit test (matching the convention in `features/sidebar/lib/`), add a `useChannelNavigationShortcuts` hook alongside the existing `useMarkAsReadShortcuts`, and add an e2e case to `navigation.spec.ts`.

**Known limitation in this first pass**

Sidebar collapse state (`collapsedGroups` / `collapsedSections`) is local `React.useState` inside `AppSidebar` and isn't reachable from an app-level shortcut hook. So navigating into a collapsed group will switch the channel correctly but leave the group shut, meaning the selection highlight isn't visible until you expand it. Making the shortcut collapse-aware — expand on arrival, or skip collapsed groups — requires lifting that state into `AppShell`. Happy to do that here or as a follow-up, but it restructures state ownership, so I'd rather you called the shape.

**Also happy to follow up with**

`⌥⇧↓` / `⌥⇧↑` for next/previous *unread*, which is where Slack's mute-skipping behavior properly belongs. Keeping this PR to the plain case.

Checking the keybinding and scope look right before I open the PR.

enzowilliam:
## Problem

There is no keyboard path to move between channels — every switch requires the mouse or a `⌘K` round-trip. Closes #3078.

**Duplicate check:** searched open PRs and issues — none found beyond issue #3078 itself. The issue author sketched this exact design in the issue body; this PR follows that spec as written. If they have work in flight, happy to defer or hand this over.

## What this adds

`⌥↓` / `⌥↑` (macOS) and `Ctrl+Alt+↓` / `Ctrl+Alt+↑` (Windows/Linux) move the active channel selection down/up the sidebar list. Windows/Linux includes `Ctrl` because plain `Alt`+arrows are already back/forward (`useBackForwardControls.ts`).

Per the issue spec:
- Order matches the sidebar exactly: starred → custom sections in section order → unassigned, each honoring its own sort preference.
- Scope is the active community's stream channels; forums and DMs keep the combo's default behavior.
- Muted channels are skipped.
- Stops at both ends — no wraparound.
- No-op on the home feed or when the selection isn't in the list.
- Registered under **Navigation** in `keyboard-shortcuts.ts`, so it shows up in Settings → Keyboard Shortcuts automatically.

## Implementation

- **`features/sidebar/lib/sidebarChannelOrder.ts`** — the sidebar's displayed-order flattening, extracted from `AppSidebar.tsx` into a shared helper (`buildSidebarChannelGroups` / `flattenSidebarChannelGroups` / `adjacentSidebarChannelId`) so the shortcut and the sidebar can't drift. `AppSidebar` now renders from it; rendering is unchanged.
- **`app/useChannelNavigationShortcuts.ts`** — new hook mirroring `useMarkAsReadShortcuts`, wired in `AppShell`. Section membership and per-group sort prefs are read from their relay-scoped localStorage stores at keypress time — the same stores the sidebar hooks keep in sync (including remote NIP-78 updates) — avoiding a duplicate relay subscription.
- **`app/useGlobalActionShortcuts.ts`** — the pre-existing inline ⌘K/⇧⌘K/⇧⌘N/⇧⌘O/⇧⌘A keydown effect moved verbatim out of `AppShell`. Not drive-by churn: the addition pushed `AppShell.tsx` over the enforced 1000-line ceiling (`check:file-sizes` says split, never bump). AppShell is now 951 lines.

## Tests

- 11 unit tests for the ordering/adjacency helper (`sidebarChannelOrder.test.mjs`): section order, per-group sort, orphaned assignments, muted skipping, boundary stops, empty/unselected no-ops.
- New e2e case in `navigation.spec.ts`: steps through sidebar channels without wrapping, platform-aware combo. Ran with the smoke project: 13 passed, 1 skipped (pre-existing `test.fixme`).
- Full desktop unit suite: 3732/3732 pass. `pnpm typecheck`, biome, `pnpm check` (file sizes, px-text, pubkey-truncation): pass.
- `just ci` passes locally.

## Manual test

1. Join a community with a few channels; star one, mute another, put one in a custom section.
2. Select a channel, press `⌥↓` / `⌥↑` (or `Ctrl+Alt+↓/↑`): selection walks the sidebar order, skips the muted channel, stops at the ends.
3. Open a DM or the home feed: the combo does nothing / keeps its default behavior.
4. Settings → Keyboard Shortcuts shows both entries under Navigation.

## Deferred (per the issue)

- Collapse-awareness: navigating into a collapsed group switches the channel but the group stays shut (collapse state is local to `AppSidebar`; lifting it is explicitly deferred in the issue).
- `⌥⇧↓` / `⌥⇧↑` for next/previous *unread*.

Screenshots follow in a comment (via `scripts/post-screenshots.sh`).

### DIFF
--- a/desktop/src/app/AppShell.tsx
+++ b/desktop/src/app/AppShell.tsx
@@ -11,6 +11,7 @@ import { useBackForwardControls } from "@/app/navigation/useBackForwardControls"
 import { useCommunityNavigationTransitions } from "@/app/useCommunityNavigationTransitions";
 import { useLiveHomeFeedActions } from "@/app/useLiveHomeFeedActions";
 import { useChannelBrowserDialog } from "@/app/useChannelBrowserDialog";
+import { useChannelNavigationShortcuts } from "@/app/useChannelNavigationShortcuts";
 import { useMarkAsReadShortcuts } from "@/app/useMarkAsReadShortcuts";
 import { useSettingsShortcuts } from "@/app/useSettingsShortcuts";
 import { useAppShellDesktopNotifications } from "@/app/useAppShellDesktopNotifications";
@@ -91,7 +92,7 @@ import { ChannelNavigationProvider } from "@/shared/context/ChannelNavigationCon
 import { MainInsetProvider } from "@/shared/layout/MainInsetContext";
 import { chromeCssVarDefaults } from "@/shared/layout/chromeLayout";
 import { cn } from "@/shared/lib/cn";
-import { hasPrimaryShortcutModifier } from "@/shared/lib/platform";
+import { useGlobalActionShortcuts } from "@/app/useGlobalActionShortcuts";
 import { useMessageDeepLinks } from "@/shared/useMessageDeepLinks";
 import { SidebarInset, SidebarProvider } from "@/shared/ui/sidebar";
 import { RelayConnectionOverlay } from "@/app/RelayConnectionOverlay";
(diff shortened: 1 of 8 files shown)


enzowilliam:
Screenshots for the channel-navigation shortcut (captured via the e2e mock bridge, sidebar shots cropped to the sidebar).

### Before
`#general` is the active channel.

01-before-general-selected

### After one press of ⌥↓
Selection moves to the next channel in sidebar order, `#random`.

02-after-next-random-selected

### Muted channels are skipped
With `#random` muted (dimmed), one press from `#general` lands on `#secret-projects`.

03-skip-muted-random

### Settings → Keyboard Shortcuts
Both shortcuts registered under Navigation.

04-settings-keyboard-shortcuts

PR #3379, opened 2026-07-28
feat(acp): run channel threads concurrently

joahg:
Fixes the single-lane-per-channel limit in buzz-acp: an agent with `parallelism=20` could still only run one turn per channel at a time, because queueing and in-flight tracking were keyed by `channel_id` alone.

Refs AGNTOPS-348.

## What changed

Concurrency is now keyed by `ConversationKey { channel_id, thread_root }`, where `thread_root` is the NIP-10 root event id from `parse_thread_tags`:

- **Queue** (`queue.rs`): pending queues, in-flight tracking/deadlines, retry backoff, cancelled-batch requeues, and native-steer withholding are all per-conversation. Same thread stays serialized and batched; different threads in one channel flush concurrently. Top-level (non-threaded) events share the channel's `None` lane and stay serialized among themselves.
- **Main loop** (`lib.rs`): typing indicators, dispatch/result/panic cleanup, mid-turn steer/cancel, and native-steer acks all route by conversation. Observer `cancel_turn` accepts an optional `threadRootEventId` to target one lane; omitting it cancels every active turn in the channel (backward compatible).
- **Pool** (`pool.rs`): ACP session identity and turn-rotation counters are per-conversation. `try_claim` prefers an exact conversation-affinity hit, then a same-channel agent (reusing core/canvas caches, which stay channel-scoped), then any idle agent. Channel removal invalidates every conversation lane for that channel.

## Validation

At `32b4cb12f` (worktree of `a35771fc`):

```
cargo fmt --check        ✓
cargo clippy -p buzz-acp --all-targets -- -D warnings   ✓
cargo test -p buzz-acp   623 + 9 passed, 0 failed
```

New focused tests cover: two roots flushing concurrently, same-root serialization, per-root completion isolation, channel drain across all roots, top-level lane serialization, per-conversation sessions, and exact-vs-channel-wide invalidation.

Buzz thread: channel `82f1f4c4-1235-446b-bbba-7eeb7a0544f8`, root `a448075e2158045ea54c6d7be29b74642537108131789bb655f3b5ea49f2028a`

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -41,7 +41,7 @@ use pool::{
     PromptResult, PromptSource, SessionState, TimeoutKind,
 };
 use pool_lifecycle::PoolLifecycle;
-use queue::{CancelReason, EventQueue, FlushBatch, QueuedEvent, ThreadTags};
+use queue::{CancelReason, ConversationKey, EventQueue, FlushBatch, QueuedEvent, ThreadTags};
 use relay::{HarnessRelay, RelayEventPublisher};
 use tokio::sync::{mpsc, watch};
 use tracing_subscriber::EnvFilter;
@@ -892,6 +892,10 @@ fn handle_relay_observer_control_event(
 }
 
 /// Handle a `cancel_turn` control frame: signal the in-flight task to cancel.
+///
+/// An optional `threadRootEventId` targets one conversation lane exactly.
+/// Without it, every in-flight turn for the channel is cancelled — identical
+/// to the historical behavior when at most one turn per channel existed.
 fn handle_cancel_turn_control(
     payload: &serde_json::Value,
     pool: &mut AgentPool,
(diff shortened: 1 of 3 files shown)


cantrellsolutions:
This is very close to the UX we need, and I tested it against a self-hosted Buzz + Hermes ACP pilot.

The PR built and ran cleanly from `32b4cb12f`, but our smoke test still showed top-level channel messages using the shared channel lane:

```text
conversation=<channel_uuid>
```

That means two independent top-level agent tasks in the same dedicated agent channel can still steer/cancel/merge each other before either has a thread root.

For the “one agent channel, many independent task threads” UX, I think `thread_root=None` should not always mean “shared channel session” when thread/conversation scoping is enabled. A top-level channel event should become its own conversation root immediately:

```text
top-level event: conversation = (channel_id, event_id)
reply event:     conversation = (channel_id, root_event_id)
```

Then the agent’s first reply anchors to the top-level event as the thread root, and all follow-ups under that root reuse the same ACP session.

Would you be open to that semantic in this PR? We’re happy to test a branch or contribute the small delta if useful.

joahg:

Superseded by the merged thread-scoped session implementation in https://github.com/block/buzz/pull/6732, its opt-in desktop control in https://github.com/block/buzz/pull/6909, and concurrency fixes https://github.com/block/buzz/pull/7337 and https://github.com/block/buzz/pull/7340. This retires the old implementation; it does not claim the opt-in feature is enabled everywhere.

PR #3387, opened 2026-07-28
fix(desktop): include relay agents in picker results

justinhnaylor:
## Summary

- include relay-directory agents in mention autocomplete and member search
- keep existing managed-agent behavior while allowing shared relay-managed agents
- add regression coverage for known managed and relay identities

## Problem

Relay-managed agents that are already channel members are filtered out of the `@` picker because candidates are accepted only when their identity exists in the local managed-agent list. This leaves shared agents invisible in autocomplete even though the channel member list contains them.

## Testing

- `pnpm check` passed for desktop, including formatting and file-size checks
- desktop regression suite passed: 3,731 tests
- desktop TypeScript and production build passed
- Rust formatting, clippy, unit tests, and desktop-native tests passed
- web checks and production build passed
- focused agent picker regression suite passed: 18 tests

`just ci` reaches the unrelated mobile suite and fails in the existing `ChannelDetailPage keeps follow mode off while a tall newest message stays visible` test at `mobile/test/features/channels/channel_detail_page_test.dart:985`. The failure reproduces when that mobile test is run alone. This pull request does not change any mobile files.

## Manual verification

In Buzz 0.5.0, a channel containing relay-managed agents reproduces the issue: the channel roster includes the agents, but typing `@` omits them from the picker. With this change, known relay agents pass candidate filtering in both mention autocomplete and member search.

No matching open issue or pull request was found.

No agent configuration rules changed.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -5,7 +5,7 @@ import {
   coalesceAgentAutocompleteCandidates,
   getMentionableAgentPubkeys,
   getSharedChannelIds,
-  isAgentIdentityInManagedList,
+  isAgentIdentityInKnownDirectories,
   relayAgentIsSharedWithUser,
   shouldHideAgentFromMentions,
 } from "./agentAutocompleteEligibility.ts";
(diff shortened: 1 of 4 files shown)


justinhnaylor:
Friendly ping — this has been open a week with CI green and no merge conflicts, so I think it's just waiting on a review.

Recap: relay-managed agents that are already channel members get filtered out of the `@` picker, because candidates are only accepted when the identity exists in the local managed-agent list. Shared agents stay invisible in autocomplete even though the channel roster already contains them. The fix is +35/-14 across 4 files, leaves existing managed-agent behavior unchanged, and adds regression coverage for both known managed and relay identities.

@tlongwell-block you've reviewed a number of community PRs recently — would you be the right person for this one, or could you point me toward whoever owns the desktop agent picker?

The branch is 173 commits behind `main` at this point; happy to rebase if that would help.

Issue #2390, opened 2026-07-22
relay: workflow a-tag deletion leaves current kind:30620 event live — ghost workflows in Desktop list, silent resurrection on edit


leesalminen:
## Summary

Deleting a workflow (kind:5 with `a`-tag address) removes the `workflows` control-plane row (log: "Workflow deleted via NIP-09 a-tag (UUID)") but can leave the **current kind:30620 definition event live** in the event log. The Desktop workflows list is built from `/query kinds:[30620]`, so it keeps showing "ghost" workflows that no longer exist.

Worse, the ghosts are interactive: editing/toggling a ghost entry submits a kind:30620 update for the deleted id, and `handle_workflow_def` treats an unknown id as a create — silently **resurrecting the deleted workflow**. Manual-trigger (46020) on a ghost returns "workflow not found", which is how users notice.

## Observed sequence (single-node relay, `ghcr.io/block/buzz:main`)

1. Create workflow W (definition event E1). Delete W → row deleted, **E1 tombstoned** ✅
2. Ghost entry still visible in a stale Desktop list → user edits it → kind:30620 update with W's d-tag → workflow re-created with **new definition event E2** ✅ (by design, but surprising from a ghost)
3. Delete W again → relay logs "Workflow deleted via NIP-09 a-tag (UUID)", row deleted, **but E2 stays live** (`events.deleted_at IS NULL`) ❌
4. `/query kinds:[30620]` returns E2 forever → Desktop lists a workflow that `get_workflow`/trigger reject as not found.

Step 1 vs step 3 suggests event tombstoning keys on specific event ids (`e` tags from the client's possibly-stale view) rather than the a-tag address: the second delete referenced the original event id, not E2.

## Suggested fix

On NIP-09 a-tag workflow deletion, tombstone **all** live kind:30620 events whose `d` tag matches the deleted workflow id (address-based, same transaction as the row delete). Optionally: reject kind:30620 updates for ids with a deletion tombstone instead of silently re-creating.

Related UX: the workflows list should refetch after delete mutations resolve; a stale list is what makes the resurrection path reachable.

DannyDesert:
Another data point for the delete path, from a **Block-hosted** community (communities.buzz.xyz), CLI-created workflow:

- `buzz workflows delete <id>` returns `{"accepted":true}`
- `buzz workflows get <id>` still returns the workflow afterwards (retried twice with a wait between)
- Its `schedule` trigger keeps firing on cron — the executor still posts into the bound channel daily

[...]

Issue #3087, opened 2026-07-27
Hosted relay ghost workflow cannot be deleted


wjbright:
## Summary

A workflow in a hosted Buzz community is no longer visible in Buzz Desktop, but the relay still returns it and the scheduler continues to run it. Deletion is reported as accepted without removing the workflow.

## Affected community

- Community: `zairaa.communities.buzz.xyz`
- Workflow ID: `86cf70ff-d8f8-4c04-8b29-c4f81907777b`
- Workflow name: `Cross-board control pulse`
- Buzz Desktop: `0.4.26`

## Observed behaviour

1. The Desktop Workflows screen shows no workflows.
2. Relay/CLI workflow lookup still returns the workflow definition.
3. A delete request returns `accepted: true`.
4. Subsequent workflow list/get checks still return the workflow.
5. The hosted scheduler continues to execute it.

## Expected behaviour

Deleting the workflow should remove or disable both the relay definition and its scheduler state. A deletion whose backend side effect fails should not be reported as successfully accepted without an actionable error.

## Operator action requested

Please urgently disable or delete this workflow in the hosted relay backend and verify that no enabled scheduler entry remains for the workflow ID above.

Please also investigate the inconsistent state and the false-success deletion response so an owner can recover from this condition through the normal Desktop/CLI flow.

crsmoore:
We hit a closely related state on another hosted `*.communities.buzz.xyz` community: `workflows delete` returned `accepted: true` (three attempts) while `list`/`get` kept returning the workflow. In our case a `workflows trigger` probe returned `400: workflow not found` — meaning the delete had actually worked and the workflow can no longer run; what was left behind was only its stored definition record (the kind:30620 event), a "ghost" listing for a workflow that no longer exists. That's the #2879 behavior (full details in my comment there: https://github.com/block/buzz/issues/2879#issuecomment-5134243121). But until we ran that probe, the symptoms were indistinguishable from what's described here.

If the scheduler in your case is genuinely still firing the deleted workflow, that's a worse variant than #2879 covers. +1 for operator tooling on hosted relays to verify (and force-remove) scheduler state for a given workflow ID.

PR #2489, opened 2026-07-23
fix(workflow): stop deleted workflows from returning after a stale edit

BradGroux:
Fixes #2390.

## What users saw

A deleted workflow could return after someone edited a stale copy. A later delete removed the workflow row again, but the replacement definition could remain queryable on the relay.

## What changed

Canonical workflow deletion now uses one database transaction that:

- acquires the coordinate-scoped PostgreSQL advisory lock used by kind:30620 replacement writes;
- keeps the workflow row when a live definition is newer than the signed deletion event;
- tombstones only definition versions at or before the deletion event's `created_at`, as required by NIP-09;
- commits the workflow-row decision and event tombstones together;
- returns every affected definition channel so relay query caches are invalidated even when the workflow row was already gone.

The timestamp cutoff prevents a replayed tombstone from deleting a newer replacement. The complete affected-channel result also closes an orphan-repair gap where the database could tombstone a definition while leaving a stale query result cached.

## Safety and scope

The existing validator still proves signer authority. The database mutation independently scopes the row and definitions by community and owner.

This PR does not add a permanent resurrection ban, sweep legacy name aliases, or change Desktop refetch behavior. It complements rather than replaces the separate work on legacy/name no-op handling and effective agent-author attribution.

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy -p buzz-db -p buzz-relay -p buzz-test-client --all-targets --all-features -- -D warnings`
- `cargo test -p buzz-db tests::coordinate_delete_spares_head_newer_than_the_deletion -- --ignored --nocapture`: passed against PostgreSQL.
- The regression creates definitions before and after the deletion timestamp, proves the older definition is retired while the newer head and workflow survive, then proves a current tombstone removes both.

PR #3392, opened 2026-07-28
fix(relay): retire the kind:30620 event on workflow delete (#2879)

### LINKED ISSUE #2879: buzz workflows delete removes the workflow but leaves its kind:30620 event live, so workflows list and workflows get still return it
Buzz Version 0.4.26 (0.4.26)
MacOS Sequoia 15.7.5

Summary
Deleting a workflow works: the DB row is removed, the engine cache is invalidated, and the workflow can no longer be triggered. But the kind:30620 definition event is deliberately left un-soft-deleted, and both CLI read commands query that event.

Every surface an operator can inspect reports the workflow as live. The only way to find out it is gone is to trigger it.

Environment
All CLI output below was executed against a live relay.
Reproduction
``` $ buzz workflows delete --workflow 6f8ef4ca-697c-47d9-8aea-a0b7800300c4 {"accepted":true,"event_id":"85c62f65…","message":""}

$ buzz workflows list --channel [ … 6f8ef4ca-697c-47d9-8aea-a0b7800300c4 still present, full YAML … ]

$ buzz workflows get --workflow 6f8ef4ca-697c-47d9-8aea-a0b7800300c4 {"workflow_id":"6f8ef4ca-…","content":"name: Approval Gate Control\n…"}

$ buzz workflows trigger --workflow 6f8ef4ca-697c-47d9-8aea-a0b7800300c4 {"error":"relay_error","message":"relay error 400: invalid: workflow not found"} ```

The delete succeeded. list and get disagree.

Confirmed on two separate workflows. Neither posted its first step after deletion, and both had posted it on every prior trigger, so the trigger failure is corroborated by an observable side effect rather than by the error string alone.

Root cause
cmd_delete_workflow (crates/buzz-cli/src/commands/workflows.rs:139-150) signs a kind:5 whose a tag is 30620:<pubkey>:<workflow_id> (crates/buzz-sdk/src/builders.rs:1498-1508).

handle_a_tag_deletion (crates/buzz-relay/src/handlers/side_effects.rs:1979-2053) dispatches on the kind in the a tag. The KIND_WORKFLOW_DEF arm calls delete_workflow_for_owner and invalidate_channel_workflows, then returns. The generic parameterized-replaceable arm below it — the one that calls soft_delete_by_coordinate so REQs stop returning the event — is never reached, because the workflow arm matched first. The code comment states this is deliberate:

rust // Listed after the workflow branch so workflow's bespoke deletion // (which doesn't soft-delete the `events` row by design — that's a // separate concern) takes precedence.

Meanwhile cmd_list_workflows and cmd_get_workflow (crates/buzz-cli/src/commands/workflows.rs:20-58) both query {"kinds":[30620]}. They read exactly the row the delete path declines to touch.

The two halves are each internally consistent. The disagreement is between them.

Impact
An operator cannot tell a deleted workflow from a live one through any shipped read path.
A definition intended to be retired stays fully readable. Since workflows list returns the complete YAML body to any channel member, a workflow deleted because its content was sensitive is still readable by every member afterwards.
Combined with workflows runs returning [] by construction (documented at crates/buzz-cli/src/commands/workflows.rs:60-65), the CLI's entire workflow-observability surface can disagree with the engine.
Secondary: accepted: true on delete carries no information about the target
$ buzz workflows delete --workflow 00000000-0000-4000-8000-000000000000 {"accepted":true,"event_id":"f5718152…","message":""}

That UUID has never existed. The response reports that the kind:5 event was stored, not that a workflow was deleted — delete_workflow_for_owner returning None is logged and otherwise ignored. An operator who typos a workflow id gets an unqualified success. (We hit this by accident before hitting it on purpose.)

Suggested fix
Soft-delete the event row as well as the DB row. Call soft_delete_by_coordinate from the KIND_WORKFLOW_DEF arm, or let that arm fall through to the generic parameterized-replaceable handler after doing its bespoke work. Either makes list/get agree with trigger.
If leaving the event live is intentional — e.g. definitions are meant to be an append-only audit trail — then cmd_list_workflows should reconcile against the engine, or mark deleted entries, rather than presenting them as current. Whichever way it is decided, the two surfaces should not disagree silently.
Report whether anything was deleted. Have the delete path surface delete_workflow_for_owner's Option so a no-op delete is distinguishable from a real one.
Found by Bumble (Buzz onboarding agent) on 2026-07-25 while deleting test artifacts from an unrelated investigation.

bonpiedlaroute:
Fixes #2879.

## Summary

`buzz workflows delete` removed the workflow row and invalidated the engine
cache, but left the kind:30620 definition event live. `workflows list` and
`workflows get` both read that event, so every inspection surface kept
reporting a workflow that `workflows trigger` rejected with
`invalid: workflow not found`. The only way to discover the deletion had
landed was to trigger it.

## Cause

`handle_a_tag_deletion`'s `KIND_WORKFLOW_DEF` arm returns after
`delete_workflow_for_owner` + `invalidate_channel_workflows`, so the generic
NIP-33 arm below it — the only caller of `soft_delete_by_coordinate` — is never
reached for kind:30620. Every other addressable kind, kind:30023 included, is
soft-deleted so REQs stop returning it.

## Change

Extract the soft-delete block into `soft_delete_addressable_coordinate` and
call it from both arms. No new logic: the behaviour already existed, it was
just unreachable for kind:30620.

The comment above the generic arm stated the current behaviour was deliberate
("doesn't soft-delete the `events` row by design — that's a separate concern").
It is rewritten, since it would otherwise assert the opposite of what the code
does. My reading is that the concern it deferred is workflow *row* deletion
semantics, not the visibility of the definition event. Worth noting
`soft_delete_by_coordinate` is an `UPDATE … SET deleted_at = NOW()`
(`crates/buzz-db/src/event.rs:771`) — the row is retained, only its REQ
visibility changes, so no definition history is lost.

If the intent really was for the kind:30620 event to outlive the workflow, the
fix belongs in `cmd_list_workflows` / `cmd_get_workflow` instead, filtering out
definitions with no live workflow. Happy to switch to that shape.

## Authorization

No new surface. Ownership of the a-tag coordinate is already enforced by
`validate_standard_deletion_event`
(`crates/buzz-relay/src/handlers/side_effects.rs:234-253`), which runs from
`ingest.rs:1967` before the event is stored and before any side effect —
it requires `target_pubkey == actor || is_agent_owner(...)`, otherwise
`must be event author`. The check is identical for both arms.

## Testing

New e2e test `test_workflow_delete_retires_definition_event`, mirroring
`test_long_form_a_tag_deletion` which covers this path for kind:30023.

Verified it is a real regression guard:

- without the change: `FAILED — workflow deletion should retire the definition
  event (got 1 events)`
- with the change: `ok`

### DIFF
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -2071,6 +2071,50 @@ async fn handle_leave_request(
 // handle_reaction() removed — kind:7 reaction dedup and DB writes are now
 // handled inline in ingest_event() before storage (see ingest.rs step 20a).
 
+/// Soft-delete the live `events` row for an addressable coordinate so REQs stop
+/// returning it.
+///
+/// Shared by the workflow branch and the generic NIP-33 branch of
+/// [`handle_a_tag_deletion`]. Ownership of the coordinate is already enforced by
+/// [`validate_standard_deletion_event`] before the event is stored, so this
+/// helper does not re-check it.
+async fn soft_delete_addressable_coordinate(
+    tenant: &TenantContext,
+    state: &Arc<AppState>,
+    kind: u32,
+    pubkey_hex: &str,
+    d_tag: &str,
+) -> anyhow::Result<()> {
+    let pubkey_bytes = hex::decode(pubkey_hex)
+        .map_err(|e| anyhow::anyhow!("invalid pubkey hex in a-tag {pubkey_hex}: {e}"))?;
+    // Safe cast: NIP-33 kinds are 30000-39999, well within i32
+    let kind_i32 = kind as i32;
+    let deleted = state
+        .db
+        .soft_delete_by_coordinate(tenant.community(), kind_i32, &pubkey_bytes, d_tag)
+        .await
+        .map_err(|e| {
+            anyhow::anyhow!(
+                "failed to soft-delete by coordinate {kind_i32}:{pubkey_hex}:{d_tag}: {e}"
+            )
+        })?;
+
+    if deleted {
+        tracing::info!(
+            kind,
+            d_tag,
+            "NIP-09 a-tag deletion: soft-deleted addressable event by coordinate"
+        );
+    } else {
+        tracing::debug!(
+            kind,
+            d_tag,
+            "NIP-09 a-tag deletion: no live row matched coordinate"
+        );
+    }
+    Ok(())
+}
+
 /// Handle NIP-09 deletion via `a` tag (addressable/parameterized-replaceable events).
 /// Parses "kind:pubkey:d-tag" and deletes the corresponding DB record.
 async fn handle_a_tag_deletion(
@@ -2147,6 +2191,18 @@ async fn handle_a_tag_deletion(
                     }
                 }
             }
+
+            // The DB row and the engine cache are gone; the kind:30620 definition
+            // event must follow, or `workflows list` / `workflows get` keep
+            // returning a workflow that `workflows trigger` rejects (#2879).
+            soft_delete_addressable_coordinate(
+                tenant,
+                state,
+                buzz_core::kind::KIND_WORKFLOW_DEF,
+                pubkey_hex,
+                d_tag,
+            )
+            .await?;
         }
         // Generic NIP-33 (parameterized-replaceable) soft-delete by coordinate.
         //
@@ -2157,38 +2213,7 @@ async fn handle_a_tag_deletion(
         // live row matching `(kind, pubkey, d_tag)` so REQs stop returning it.
         // See https://github.com/block/sprout/issues/714.
         k if is_parameterized_replaceable(k) => {
-            let pubkey_bytes = match hex::decode(pubkey_hex) {
-                Ok(b) => b,
-                Err(e) => {
-                    return Err(anyhow::anyhow!(
-                        "invalid pubkey hex in a-tag {pubkey_hex}: {e}"
-                    ));
-                }
-            };
-            // Safe cast: NIP-33 kinds are 30000–39999, well within i32.
-            let kind_i32 = k as i32;
-            let deleted = state
-                .db
-                .soft_delete_by_coordinate(tenant.community(), kind_i32, &pubkey_bytes, d_tag)
-                .await
-                .map_err(|e| {
-                    anyhow::anyhow!(
-                        "failed to soft-delete by coordinate {kind_i32}:{pubkey_hex}:{d_tag}: {e}"
-                    )
-                })?;
-            if deleted {
-                tracing::info!(
-                    kind = k,
-                    d_tag = d_tag,
-                    "NIP-09 a-tag deletion: soft-deleted addressable event by coordinate"
-                );
-            } else {
-                tracing::debug!(
-                    kind = k,
-                    d_tag = d_tag,
-                    "NIP-09 a-tag deletion: no live row matched coordinate"
-                );
-            }
+            soft_delete_addressable_coordinate(tenant, state, k, pubkey_hex, d_tag).await?;
         }
         _ => {
             tracing::debug!(
--- a/crates/buzz-test-client/tests/e2e_workflow_deletion.rs
+++ b/crates/buzz-test-client/tests/e2e_workflow_deletion.rs
@@ -1,0 +1,164 @@
+//! End-to-end regression test for NIP-09 a-tag deletion of workflow
+//! definitions (kind:30620).
+//!
+//! Requires a running relay. Marked `#[ignore]` like the other e2e suites.
+//!
+//! ```text
+//! cargo test --test e2e_workflow_deletion -- --ignored
+//! ```
+
+use std::time::Duration;
+
+use buzz_test_client::BuzzTestClient;
+use nostr::{Alphabet, EventBuilder, Filter, Keys, Kind, SingleLetterTag, Tag};
+
+const KIND_WORKFLOW_DEF: u16 = 30620;
+
+fn relay_url() -> String {
+    std::env::var("RELAY_URL").unwrap_or_else(|_| "ws://localhost:3000".to_string())
+}
+
+fn relay_http_url() -> String {
+    relay_url()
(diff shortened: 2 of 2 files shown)


bonpiedlaroute:
Flagging an overlap I should have caught before opening this: #2489 (open since
Jul 23, rebased today) fixes #2390, which is the same root cause reported four
days before #2879. It also adds a `crates/buzz-test-client/tests/e2e_workflow_deletion.rs`,
so the two branches would conflict on that path.

#2489 is the broader fix: one transaction under the kind:30620 coordinate
advisory lock, tombstoning every live definition for the coordinate rather than
the current one, and returning affected channels so query caches are
invalidated even when the workflow row was already gone.

This PR is the minimal version: reuse the existing `soft_delete_by_coordinate`
call from the generic NIP-33 arm by extracting it into a helper, so the
`KIND_WORKFLOW_DEF` arm stops short-circuiting past it. 2 files, +221/-32, no
schema or transaction changes.

Maintainers should take whichever fits the review budget — I'm happy to close
this in favour of #2489, and equally happy to keep it as the small-surface
option if #2489 needs more review time. Not looking to duplicate anyone's work.

bonpiedlaroute:
Closing in favour of #4234.

Agreeing with @aweiker's review: the `created_at`-bounded guard on the projection
delete is the piece this branch is missing, and it's a real defect rather than a
nitpick. #3392 predates the `deletion_created_at_secs` parameter that
`soft_delete_by_coordinate` has since gained on `main`, so a replayed or
late-arriving deletion could erase a newer replacement head. #4234 gets that
right, and it puts both writes in one transaction.

Also worth recording that this root cause now has #2489, #3650, #4318 and #4882
proposing overlapping fixes, plus four open issues (#2390, #2879, #3087, #4580).
Seven overlapping PRs is a signal about review budget, not about the fix — I'd
rather withdraw mine than add to the pile.

One artefact #4234 doesn't carry, in case it's useful: a relay-level e2e test.
Its regressions are Postgres-level in `buzz-db`. This branch has
`crates/buzz-test-client/tests/e2e_workflow_deletion.rs` — publish a kind:30620
under a real channel, assert it is queryable, delete via the a-tag coordinate the
CLI actually emits, assert the REQ comes back empty. Verified as a genuine
regression guard rather than a tautology: it fails on `main` without the fix
(`workflow deletion should retire the definition event (got 1 events)`) and passes
with it. That exercises @aweiker's suggestion (2) from the client side rather than
the DB side.

Happy to port it onto #4234 if @Wirenut33 wants it — #2489 and #4882 also carry
e2e tests, so take whichever fits best.

wesbillman (changes requested):

This feature is valid, but this PR is not safe to resurrect or merge for release.

**Blockers**

1. Workflow deletion remains a non-atomic, non-retryable post-storage side effect. The kind-5 request is committed first; projection deletion and the new kind-30620 soft-delete then run as separate writes. Any transient failure is only logged while ingest still returns success. Replaying the same signed request takes the duplicate fast path before side effects, so the missing lifecycle write is never retried. This can still strand query state and executable state on opposite sides of the deletion. The event tombstone and workflow projection must be committed as one revision-fenced transaction, or the command needs durable retry/reconciliation.

2. The path is incomplete for the already-authorized managed-agent owner case. `validate_standard_deletion_event` permits a human owner to delete their agent’s kind-30620 coordinate, but `handle_a_tag_deletion` passes the effective human actor to `delete_workflow_for_owner` rather than the coordinate owner from the `a` tag. The projection delete therefore fails after the request has already been accepted/stored. Use the validated coordinate owner and cover this case end to end.

3. The branch predates timestamp-fenced coordinate deletion. Current `main` requires `deletion_created_at_secs` in `soft_delete_by_coordinate`; this helper calls the old signature. Resolving the merge conflict mechanically would either fail to compile or omit the NIP-09 guarantee that a stale/replayed tombstone cannot erase a newer replacement. Pass the deletion event timestamp through and test old-delete/new-replacement ordering.

The new relay E2E is a useful regression test for the basic UUID happy path, but it does not cover atomic failure/replay recovery, stale deletion ordering, or human-owner deletion of an agent-authored workflow. The safer route is to supersede this closed branch with the atomic, revision-fenced implementation being developed in the newer workflow-deletion work, then retain this E2E scenario there.

wesbillman (commented):
Correction: this review was submitted before one delegated review lane returned. It should be treated as procedurally withdrawn pending the consolidated review. GitHub did not reflect the dismissal attempt on this closed PR.

wesbillman (changes requested):

This feature is valid, but this PR is not safe to resurrect or merge for release.

**Blockers**

1. Workflow deletion remains a non-atomic, non-retryable post-storage side effect. The kind-5 request is committed first; projection deletion and the new kind-30620 soft-delete then run as separate writes. Any transient failure is only logged while ingest still returns success. Replaying the same signed request takes the duplicate fast path before side effects, so the missing lifecycle write is never retried. This can still strand query state and executable state on opposite sides of the deletion. The event tombstone and workflow projection must be committed as one revision-fenced transaction, or the command needs durable retry/reconciliation.

2. The path is incomplete for the already-authorized managed-agent owner case. `validate_standard_deletion_event` permits a human owner to delete their agent’s kind-30620 coordinate, but `handle_a_tag_deletion` passes the effective human actor to `delete_workflow_for_owner` rather than the coordinate owner from the `a` tag. The projection delete therefore fails after the request has already been accepted/stored. Use the validated coordinate owner and cover this case end to end.

3. The branch predates timestamp-fenced coordinate deletion. Current `main` requires `deletion_created_at_secs` in `soft_delete_by_coordinate`; this helper calls the old signature. Resolving the merge conflict mechanically would either fail to compile or omit the NIP-09 guarantee that a stale/replayed tombstone cannot erase a newer replacement. Pass the deletion event timestamp through and test old-delete/new-replacement ordering.

The new relay E2E is a useful regression test for the basic UUID happy path, but it does not cover atomic failure/replay recovery, stale deletion ordering, or human-owner deletion of an agent-authored workflow. The safer route is to supersede this closed branch with the atomic, revision-fenced implementation being developed in the newer workflow-deletion work, then retain this E2E scenario there.

Issue #3390, opened 2026-07-28
Proposal: add Remote Team discovery and join


tmarman:
## The problem

The bridge in #3389 connects one agent at a time. But the runtime on the other side already knows its whole roster. Adding each agent by hand means re-entering configuration Buzz could simply read.

## What we propose

- Add **New team → Remote team** in the Agents view.
- Show the remote roster first. Nothing is created until the owner chooses.
- Pick which agents join, and which channel they join.
- Keep the names and avatars the source already uses.
- Give each selected agent its own Buzz-local identity.
- Prompts, memory, tools, credentials, and signing stay on the source runtime.

The UI calls this **Remote Team**. Internally the modules are still named `RemoteAgency`; renaming them would be unrelated churn.

## Screenshots

Before:

![Agents view before Remote Team](https://raw.githubusercontent.com/tmarman/buzz/5eb49564bb379d2a65450311b99ad35c5bb1185d/.github/pr-assets/remote-team-stack/remote-team-before.png)

After:

![Agents view with Remote Team entry](https://raw.githubusercontent.com/tmarman/buzz/5eb49564bb379d2a65450311b99ad35c5bb1185d/.github/pr-assets/remote-team-stack/remote-team-overview-after.png)

![Add remote team dialog](https://raw.githubusercontent.com/tmarman/buzz/5eb49564bb379d2a65450311b99ad35c5bb1185d/.github/pr-assets/remote-team-stack/remote-team-dialog-after.png)

These use neutral test fixtures.

## Not in scope

- AGNTCY Directory lookup
- Keeping an agent's original identity in Buzz
- Enforcing Space or project permissions on the source runtime
- MCP Apps rendering
- Leave, revoke, and refresh actions

## Questions for maintainers

1. Does **New team → Remote team** fit how teams work today?
2. Is reading one reviewed record enough to start, or should this wait for directory-based discovery?
3. After an agent joins, what does someone need to be able to do — leave, revoke, refresh? Which of those are required for a first version?

tmarman:
Dependency update: the runtime bridge is now open as draft #3394. Proposed chain: #3389 / #3394 → #3390 (this Remote Team UI) → #3391. The UI implementation remains staged on the personal fork so this discussion can settle the product shape before an upstream code PR.

Issue #3389, opened 2026-07-28
Proposal: bridge existing A2A runtimes through Buzz proxy identities


tmarman:
## The problem

Some agents already run somewhere else. That runtime owns their prompts, memory, tools, and execution state. Today Buzz has to be configured as if it owns them too, which blurs who holds the credentials and who controls the process.

## What we propose

Let Buzz talk to an agent without taking over its runtime.

- An OASF Agent Record says what the agent is and where to reach it over A2A.
- A small `buzz-a2a-acp` process sits in between. Buzz starts and stops that process over ACP, the same way it manages any local agent.
- The agent keeps running where it already runs. Buzz never receives its keys.
- Buzz gives the agent a local Nostr identity so it can take part in a channel.
- When the agent replies, Buzz posts that reply once, under the local identity.
- The relay does not change.

## What we check before trusting a record

- HTTPS only, except loopback for local development.
- DNS answers are validated and pinned before we connect.
- Artifacts must match their declared digest, media type, and size.
- A bearer token is only ever sent to the one endpoint it was configured for.
- The adapter process never inherits Buzz or Nostr private keys.
- If posting fails and we retry, we resend the same signed event instead of signing a new one.

Code: #3394 (draft).

## Not in scope

- AGNTCY Directory lookup
- Keeping the agent's original identity in Buzz
- Cancelling or streaming a remote turn
- The team join UI (#3390)
- Enforcing project or Space permissions
- MCP Apps rendering

## Questions for maintainers

1. Is "Buzz supervises a local adapter, the adapter calls A2A" the right split for an agent that already runs elsewhere?
2. Should the adapter stay a separate crate and a shipped sidecar?
3. Do you need remote cancellation or streaming before this could merge?
4. A source runtime may want to know which organization, project, or channel a turn came from. Is an A2A extension the right way to pass opaque references for that, or does something existing already cover it? These would be forwarded metadata only — they would not authorize anything on their own.

tmarman:
Draft implementation: #3394. Review chain: #3389 / #3394 (runtime bridge and proxy publication) → #3390 (Remote Team discovery and join) → #3391 (optional Agency Context references). The later code slices remain staged until the runtime boundary is accepted.

PR #3394, opened 2026-07-28
feat(acp): bridge A2A runtimes through Buzz proxy identities

tmarman:
## What this does

Some agents already run somewhere else, and that runtime owns their prompts, memory, tools, credentials, and execution state. This draft lets Buzz talk to such an agent without taking it over.

- An OASF Agent Record says what the agent is and where to reach it over A2A.
- A2A carries each prompt to the runtime that is already running the agent.
- ACP stays what it is today: how Buzz starts and stops a local process.
- When the agent replies, Buzz posts that reply under a Buzz-local Nostr identity.

The relay protocol and the relay implementation do not change.

## What is in the diff

- A new `buzz-a2a-acp` sidecar, packaged with the supported desktop builds.
- Record resolution and validation, before anything calls the reviewed endpoint.
- An opt-in A2A extension mechanism. No vendor extension is enabled by default.
- Output capture that posts exactly one event per completed turn, and keeps metrics and adapter reuse working when posting fails.
- Reply threading that matches what Buzz already does in channels and DMs.
- Remote bearer credentials stay inside the adapter process. Buzz and Nostr private keys are never passed down to it.

## What we check

- Non-loopback HTTP is rejected.
- DNS answers are validated and pinned; private, mixed, and transitional address ranges are refused.
- Records and artifacts must match their declared media type and digest.
- The source runtime keeps its signing authority. A Buzz proxy identity does not claim to be the original agent.

## Scope

This draft does not add the Remote Team UI, use AGNTCY Directory, carry a source identity into Buzz, enforce project permissions, cancel or stream a remote turn, or render MCP Apps. Those are left out on purpose so the runtime contract can be reviewed on its own.

The follow-on UI question is #3390.

One note on history: commit `819c5d18f` carries a `fix(remote-team)` subject from the branch this came from. It only touches `crates/buzz-acp`. It is left as-is rather than force-pushing a draft that is under review.

## Related issue

Discusses #3389.

## Testing

- `CARGO_NET_GIT_FETCH_WITH_CLI=true bin/cargo test -p buzz-acp`
  - 631 unit tests and 9 lifecycle tests passed.
- `CARGO_NET_GIT_FETCH_WITH_CLI=true bin/cargo test -p buzz-a2a-acp`
  - 27 tests passed.
- `CARGO_NET_GIT_FETCH_WITH_CLI=true bin/just ci`
  - Rust workspace clippy, desktop checks/clippy/tests/build, web checks/build, mobile formatting/analyze, and the non-mobile unit suites passed.
  - The run stops in `mobile-test` at `ChannelDetailPage keeps follow mode off while a tall newest message stays visible`.
  - That exact test fails the same way on a detached clean `upstream/main` at `1d3b810ad`, and this branch changes nothing under `mobile/`.

This PR has no user-interface changes, so before/after screenshots do not apply.


dzintars-co:
This A2A bridge looks especially valuable for practical integrations with externally hosted agents. It would remove a lot of custom glue for setups where the remote runtime should remain authoritative for prompts, memory, tools, and execution state. Is there anything specific blocking this from leaving Draft / entering maintainer review? A short status update on the remaining blockers or expected next step would be very helpful.

PR #3397, opened 2026-07-28
feat(acp): require Buzz-Thread trailer on commits in base prompt

joahg:
## Summary

Updates the buzz-acp base system prompt so agents **always** link commits back to the originating Buzz thread:

- Adds a mandatory rule under **Working in the Repo**: every commit for Buzz-originated work must carry a `Buzz-Thread: <channel-uuid>/<event-id>` trailer in the extended commit description (channel UUID from `[Context]`, event id = thread root, or the triggering event for top-level messages).
- Cross-references the trailer from the existing PR guidance (`buzz pr open --channel`), so PR-level and commit-level provenance stay paired.

This gives code-review tooling a stable, machine-parseable pointer from any commit back to the Buzz conversation that produced it.

The commit in this PR dogfoods the trailer format.

## Tracking

- Originating Buzz thread: `d9dddf9c-a04b-4eb9-b701-2fe8dabc1610/b910727269c38109fc22a06c6960b3d4569e880f1cc1512ff056671094c0fe0b`

### DIFF
--- a/crates/buzz-acp/src/base_prompt.md
+++ b/crates/buzz-acp/src/base_prompt.md
@@ -22,7 +22,7 @@ The `buzz` CLI is your primary interface. Auth env vars: `BUZZ_RELAY_URL`, `BUZZ
 
 Run `buzz --help` or `buzz <group> --help` for full usage. For multiline message content, pass real newline bytes through stdin: `printf 'first\n\nsecond\n' | buzz messages send ... --content -`. Do not write `--content 'first\n\nsecond'`: single-quoted shell strings preserve `\n` literally, so recipients will see the backslash characters. `buzz agents draft-create` and `buzz agents draft-update` require `BUZZ_AUTH_TAG`; if it is missing, explain that this managed agent cannot open owner-reviewed agent drafts from chat.
 
-When opening a pull request in response to channel work, always pass `--channel <current-channel-uuid>` using the UUID from `[Context]`. This preserves a link from the pull request back to its originating conversation.
+When opening a pull request in response to channel work, always pass `--channel <current-channel-uuid>` using the UUID from `[Context]`. This preserves a link from the pull request back to its originating conversation. Every commit in that pull request must also carry the `Buzz-Thread` trailer described in **Working in the Repo**.
 
 ## Conversational Agent Creation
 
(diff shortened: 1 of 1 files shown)


tlongwell-block (changes requested):
Let's not put this upstream

wesbillman:
From my agent, Carl.

The provenance goal makes sense, but I don't think this policy belongs in the generic ACP base prompt as written.

`crates/buzz-acp/src/base_prompt.md` is inherited by every ACP agent, so this would require every user and every destination repository to add a Buzz-specific trailer to every Buzz-originated commit—including fixups and follow-ups. That crosses a repository-policy boundary: the current generic instruction to include the trailers required by the repository is appropriately scoped, while this makes one provenance convention universal.

There are also a few product and implementation concerns:

- A channel UUID and event ID can expose a pointer to a private collaboration thread in durable, potentially public Git history, but this change defines no visibility or consent boundary.
- The PR adds no parser, validator, link resolver, or other consumer for the new format. Agents would hand-copy both identifiers, creating typo and drift risk.
- “Every commit in that pull request” can include pre-existing, cherry-picked, merged, or human-authored commits. An agent should not rewrite those commits to satisfy a Buzz prompt policy.
- Teams that squash, keep workflow metadata at the PR/ticket layer, or enforce their own commit format would still inherit this requirement.

One important nuance: `buzz pr open --channel` currently records the source channel as the PR event's `h` tag, but not the exact thread root. Exact thread provenance is therefore a real missing capability. I think the generic solution should be structured source provenance on the Buzz PR flow—channel plus root event—populated automatically by the harness/CLI, with privacy semantics and a consuming link resolver. That avoids imposing Buzz metadata on external Git histories or asking an LLM to reproduce identifiers manually.

If commit trailers are needed immediately for a particular workflow, they should be opt-in through owner/team/repository instructions. If they later become a supported platform feature, tooling should generate them according to explicit configuration.

**Recommendation:** don't merge this into the universal base prompt as-is. Keep the requirement local to the requesting workflow, or replace it with an automatically populated, structured PR-level provenance feature.

joahg:

Abandoning the universal Buzz-Thread commit-trailer proposal in line with the review requesting that this policy not be added upstream. This is a rejected direction, not a claim that an equivalent requirement merged.

PR #3398, opened 2026-07-28
fix(desktop): export configured OpenAI key to every harness

### LINKED ISSUE #3395: goose + provider openai: the API key never reaches the agent (401 every turn)
Buzz stores the OpenAI key as `OPENAI_COMPAT_API_KEY` (buzz-agent's env contract, reused as the one secret field for every runtime).
goose's openai provider reads `OPENAI_API_KEY`.
Nothing translates between them, so the goose child starts with no credential it recognizes.

Readiness hides it: `readiness::goose_requirements` checks for `OPENAI_COMPAT_API_KEY` - the key Buzz itself wrote - so the agent shows as Ready with the key set, and the only symptom is a 401 at turn time.

## Evidence

A live goose child spawned by Buzz:

```
$ ps eww <pid>
GOOSE_PROVIDER=openai   GOOSE_MODEL=gpt-5.6-sol
OPENAI_COMPAT_API_KEY=…   # present
OPENAI_API_KEY            # absent
```

And outside Buzz, with the same key:

```
OPENAI_COMPAT_API_KEY=<key> goose run -t "say ok"   # 401
OPENAI_API_KEY=<key>        goose run -t "say ok"   # ok
```

## Affects

Any goose agent configured through Buzz with provider `openai`.
Users who stored the key via `goose configure` (system keychain) are unaffected, since goose falls back to the keyring.

The obvious workaround - typing `OPENAI_API_KEY` into the env-vars editor - is blocked: the config dialogs still demand the canonical name, so Save stays disabled.

## Proposed fix

Export the canonical value under `OPENAI_API_KEY` too when a goose child's effective provider is `openai` and the agent has no native key of its own.
Write it after the layered user env, and emit nothing when the user set `OPENAI_API_KEY` themselves, so an explicit value is never clobbered.

Resolve the provider the way the child does - layered env, then the structured field, then the inherited process env, then goose's `config.yaml` - since a later env layer can override `GOOSE_PROVIDER` in either direction.

## Not covered by that fix

- `openai-compat` stays broken for goose: it has no such provider, and reaches compatible endpoints via `openai` plus `OPENAI_HOST` / `OPENAI_BASE_PATH`. That needs Buzz's single `OPENAI_COMPAT_BASE_URL` split into an origin and a path, including a decision about URLs that already carry a version suffix.
- Accepting a user-typed `OPENAI_API_KEY` needs readiness *and* the dialogs changed together: `requiredCredentialEnvKeys` plus its four consumers, and `getGlobalAgentCredentialState`, which derives `apiKeyMissing` from the canonical key directly rather than from the required-keys list.
- The same class of bug exists wherever a harness's native env name differs from Buzz's canonical one. A per-runtime alias table is the durable shape.

filipkujawa:
Fixes #3395

When Buzz has `OPENAI_COMPAT_API_KEY`, also export it as `OPENAI_API_KEY` to managed agent processes. This keeps `buzz-agent` compatibility while allowing Goose and other harnesses to read the standard variable.

An explicitly configured `OPENAI_API_KEY` is not overwritten.

### DIFF
--- a/desktop/src-tauri/src/managed_agents/runtime.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime.rs
@@ -22,8 +22,8 @@ pub(crate) use super::access_policy::{build_respond_to_env_with_policy, RespondT
 
 mod metadata;
 pub(crate) use metadata::{
-    apply_agent_display_env, resolve_session_title, runtime_metadata_env_vars,
-    DISPLAY_NAME_ENV_VAR, SESSION_TITLE_ENV_VAR,
+    apply_agent_display_env, apply_openai_api_key, resolve_session_title,
+    runtime_metadata_env_vars, DISPLAY_NAME_ENV_VAR, SESSION_TITLE_ENV_VAR,
 };
 
 mod stop;
@@ -809,6 +809,9 @@ pub fn spawn_agent_child(
     for (key, value) in &descriptor.env {
         command.env(key, value);
     }
+
+    apply_openai_api_key(&mut command, &descriptor.env);
+
     configure_runtime_cli(&mut command, runtime_meta);
 
     // Buzz shared compute is stored as a native provider; derive the OpenAI-compatible
--- a/desktop/src-tauri/src/managed_agents/runtime/metadata.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime/metadata.rs
@@ -1,3 +1,33 @@
+use std::collections::BTreeMap;
+
+const OPENAI_API_KEY: &str = "OPENAI_API_KEY";
+const OPENAI_COMPAT_API_KEY: &str = "OPENAI_COMPAT_API_KEY";
+
+fn configured_openai_api_key(env: &BTreeMap<String, String>) -> Option<&str> {
+    if env.contains_key(OPENAI_API_KEY) {
+        return None;
+    }
+    env.get(OPENAI_COMPAT_API_KEY)
+        .map(String::as_str)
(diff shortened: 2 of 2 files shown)


michaelneale:
am taking this on - still not happy with shape, has to be simpler surely..

michaelneale (approved):

wesbillman (dismissed):
Reviewing on Wes's behalf.

The local spawn fix is directionally correct, and its precedence semantics are sensible, but the alias is applied only to `std::process::Command` inside `spawn_agent_child`. That leaves another supported execution boundary broken.

**Blocking: project the alias into the portable remote launch environment too.** `build_launch_block` serializes `descriptor.env` unchanged at `desktop/src-tauri/src/commands/agents_deploy.rs:104-108`, and the Kubernetes backend executes that map as the authoritative launch environment (`crates/buzz-backend-kubernetes/src/env.rs:176-184`). Consequently, a remotely deployed Goose agent still receives only `OPENAI_COMPAT_API_KEY`, not `OPENAI_API_KEY`, and retains the same 401 behavior this PR is intended to fix. This also contradicts the launch block's documented role as the portable contract shared with provider-backed agents.

Please make the alias transformation reusable at the environment-map level, apply it to both local spawn and `launch.env`, and cover the remote launch payload plus explicit-`OPENAI_API_KEY` precedence in tests. Keeping the final mesh-specific scrub after the local projection remains important.

michaelneale:
@wesbillman @tlongwell-block @wpfleger96  this any good - would like to use it! (am having to use env vars right now!)

PR #3412, opened 2026-07-28
fix(relay): avoid subscription lock inversion

tlongwell-block:
## Summary
- drop the subscription-map guard before removing fan-out index entries
- snapshot indexed candidates before re-checking subscriptions
- add a concurrent fan-out/subscription-replacement regression test

## Why
The deployed order could deadlock synchronous DashMap shards: fan-out held an index guard then acquired `subs`, while removal held `subs` then mutated the index. Enough blocked Tokio workers can park the whole runtime.

## Validation
- `cargo check -p buzz-relay` (Hermit Rust 1.95)
- `cargo test -p buzz-relay --lib`: 769 passed, 0 failed, 33 ignored
- pre-push full repository checks passed with the Hermit toolchain

### DIFF
--- a/crates/buzz-relay/src/subscription.rs
+++ b/crates/buzz-relay/src/subscription.rs
@@ -164,15 +164,18 @@ impl SubscriptionRegistry {
         conn_id: ConnId,
         sub_id: &str,
     ) -> Option<RemovedSubscription> {
-        if let Some(mut conn_subs) = self.subs.get_mut(&conn_id) {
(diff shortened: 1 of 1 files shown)


tlongwell-block:
Closing as duplicate of #3413 after line-by-line comparison. #3413 contains the same production fix and regression coverage, plus the explicit no-cross-DashMap-guards invariant comment; reviewed with no blocking findings.

PR #3413, opened 2026-07-28
fix(relay): avoid subscription lock inversion

tlongwell-block:
## Summary
- drop the `subs` DashMap guard before mutating subscription indexes
- snapshot fan-out candidate vectors so index guards are dropped before looking up `subs`
- add concurrent fan-out/replacement regression coverage

## Why
`fan_out_scoped` previously held an index guard while `push_match` acquired `subs`, while CLOSE and same-ID replacement held `subs` while removing from an index. The reverse ordering made an AB/BA deadlock reachable and could synchronously park all Tokio workers.

## Validation
- `rustup run 1.95.0 cargo test -p buzz-relay` — 769 library tests passed, 33 ignored; 11 binary tests passed; doc tests passed
- push hooks with pinned Rust 1.95 — branch-skew, repository Rust suites, and desktop Tauri suite passed
- `git diff --check`

## Residual risk
Fan-out now clones bounded candidate vectors before matching. This adds allocation/copy cost proportional to the indexed candidate set, in exchange for eliminating nested DashMap guards. This fixes the concrete lock cycle but does not prove every observed production wedge had this cause.


tlongwell-block (commented):
Reviewed line-by-line against deployed a35771fc4 and my independent fix. The patch removes both reachable cross-map guard nestings, covers all five fan-out indexes, preserves stale-index safety via the existing subs re-check, and includes the concurrent fan-out/replacement regression test. No blocking findings. GitHub will not let this shared account approve its own PR.

tlongwell-block (commented):
BLOCKER at head 7d2da0d9: snapshot candidate for channel A, then same-ID replacement onto channel B, then push_match resumes. push_match reads the new record but validates only filters, not stored community/channel scope, so a kind-only replacement on B receives A's event. Deterministic injected test expected [] and delivered one match. Please pass expected community into push_match and require current record scope == (fanout community, event.channel_id) before filters_match; retain this stale-snapshot test for all snapshot safety. Shared GitHub account prevents REQUEST_CHANGES review state; this comment is a merge blocker.

PR #3422, opened 2026-07-28
feat(messaging): add Slack-parity message forwarding (kind 40009)

### LINKED ISSUE #3264: Add native “Forward message” option (to channel, thread, or agent)
**Motivation**
Right now there is no clean way to forward an existing message (or thread) to another channel, a different agent, or a specialist.  Users and orchestrator agents constantly need to:

- Hand off context to a specialist agent
- Move a discussion into a dedicated working channel
- Share a useful reply with another team/channel

The current workarounds (copy-paste, manual @mentions, or re-explaining the context) are slow, lose formatting/attachments, and break the audit trail. This becomes especially painful when running multi-agent teams.

**Proposed solution**
Add a “Forward” action on every message (same place as Reply / React):Forward to an existing channel
Forward to a new temporary channel
Forward directly to one or more agents (with optional note)
Optionally keep the original author + timestamp visible
Preserve attachments, reactions, and thread context when possible
Show a clear “Forwarded from …” banner

The forwarding agent/human should be able to add a short note (“Please review this and implement the fix”).

**Alternatives considered**
Manual copy-paste (loses context and attribution)
Asking the orchestrator agent to re-summarize and re-post

**Additional context**
This would make agent orchestration (especially with an orchestrator + specialist pattern) dramatically smoother.
Similar to Slack’s forward feature, but adapted to Buzz’s agent + channel model.

---

Before opening: please [search open issues and PRs](https://github.com/block/buzz/issues?q=is%3Aopen) for duplicates — link the closest one, or say "none found".

LordMelkor:
Fixes #3264

## Summary

Adds Slack-parity message forwarding: any channel or DM message can be forwarded to any other channel or DM, with an optional note, rendered as a quoted card with attribution and a link back to the source (when visible to the viewer).

## Protocol design (new kind `40009`, documented in `docs/nips/NIP-FW.md`)

- **New event kind `40009` (`KIND_STREAM_MESSAGE_FORWARD`)** — follows the repo convention that new features get new kind integers; old relays/clients fail closed.
- **NIP-18-style embedded original**: the full signed original event is carried in a `["fwd", <event JSON>]` tag, making every forward self-verifying (NIP-01 id recompute + Schnorr signature check on the relay). Works across per-channel read scoping and DMs, where a reference alone would be unreadable to recipients.
- **`content` is the forwarder's note only** — search/FTS and attribution stay honest; the original author's words are never re-attributed to the forwarder.
- **Tags**: `["h", dest]` (NIP-29 scoping), `["k", <orig kind>]`, `["fwd-src", <channel uuid>, "channel"|"private"|"dm"]`, and a NIP-18 `["q", ...]` quote tag **only when the source channel is open** (private/DM sources are not leaked). `imeta` tags are copied verbatim so media renders. No `p` tag is added for the original author (forwarding is not a mention).
- **Anti-laundering gate**: the relay verifies the *forwarder* has read access to the source channel (member, or open visibility) and that the claimed `fwd-src` label matches the actual channel row. You cannot exfiltrate a private channel's messages into a public one by fabricating tags — the embedded event must genuinely verify and the source must genuinely be readable by you.
- **Forwardable kinds allowlist** (`9`, `40002`, `45001`, `45003`); forwards-of-forwards are flattened to the innermost original (depth 1), and embedded `40009` is rejected.
- **Snapshot semantics** (Slack parity): the forward captures the original at forward time; later edits/deletes of the original do not propagate. Forwards are always root messages in the destination.

## Surfaces

| Surface | Change |
|---|---|
| `buzz-core` | `forward.rs`: envelope parse/verify, tag builder, error taxonomy (single source of truth) |
| `buzz-relay` | ingest validation: embedded-event verification under `spawn_blocking`, source-channel lookup, read-access + label checks; scope/h-tag registration |
| `buzz-db` | forwards with mentions reach the mentions feed |
| `buzz-sdk` | `build_forward(...)` typed builder |
| `buzz-cli` | `buzz messages forward --event <hex> --to-channel <uuid> [--note ...]` (sig-preserving fetch, flatten-on-reforward, fail-closed source typing) |
| Desktop | Forward action in the message action bar → destination picker dialog with note + preview → `ForwardedMessageCard` rendering (quote card, source attribution, jump-to-original when visible) |
| Mobile | `ForwardInfo` defensive tag parser + `ForwardedMessageQuote` card wired into the message bubble |
| Docs | `docs/nips/NIP-FW.md` protocol spec |

## Adversarial review

The full diff went through two adversarial review rounds plus two focused verification passes (high-effort external model). Every confirmed fixable finding was fixed in this PR:

- **Provenance (critical)**: the relay now requires the embedded original to exist in the destination community's event store, stored in the channel named by `fwd-src` — signature verification alone cannot prove a message was ever actually posted where the forward claims. Forwarding a deleted original is likewise rejected.
- Outer `imeta` must equal the embedded original's `imeta` verbatim; exactly one `h` tag on the outer event and on the embedded original; the `q` tag is pinned to its exact 4-element shape with hex validation.
- The forwarder's **source**-channel read access is checked against the database directly (bypassing the positive membership cache), so revoked members can't keep forwarding during the cache TTL.
- Kind `40009` added to every mention/notification/search surface that enumerates message kinds: desktop Home + search, pop-out channel window timeline, mobile Activity + search, ACP agent mention defaults, CLI default read/search kinds.
- Desktop: @mentions in a forward note now resolve against the destination channel and publish `p` tags; the dialog preview is fully WYSIWYG (renders the live note); forwarded-card author + embedded mention profiles ride the existing batched profile query instead of per-row requests.
- CLI: transport errors during forward-source diagnostics keep exit code 2 instead of collapsing into "not found".
- Round-2 residuals, also fixed: canonical two-element `h` tag shape enforced on outer + embedded (the `nostr` crate preserves trailing tag elements, so `["h", uuid, "extra"]` genuinely parsed before); mobile's direct-DM Activity query includes 40009; forward-note mentions resolve at submit time and the button waits for member data when the note looks mention-bearing (no more silent `p`-tag drops); the dialog preview renders embedded-original mentions and `#channel` refs identically to the timeline.
- Final-pass residuals, also fixed: the note-mention preflight is character-agnostic (`@李`, `@Élodie`, emoji-led agent names — JS `\w` is ASCII-only, so those previously bypassed the gate); the member-data wait is bounded (4s) and never silently degrades — when no member data exists (deadline expired, or the query failed before first data) the submit button explicitly reads "Forward without mentions" with a visible warning, reverting to a normal Forward the moment data arrives, so a hung request can't strand the button, a slow one can't silently drop notifications, and a failed background refetch (which keeps cached members that still resolve) doesn't show a false warning; the preview resolves mention chips and the flattened-forward author line through the same merged profile overlay as the timeline (managed/relay agents without a kind-0 profile render by name); NIP-FW now normatively specifies the two-element `["h", <uuid>]` shape and accurately describes each client's parser role (TypeScript/Dart permissive display parsers vs. the Tauri send-side builder).

Deferred as documented follow-ups (rationale in NIP-FW Security Considerations): transactional revalidate-on-insert for the residual authorization race window, client-side signature verification of embedded originals (clients currently share the platform-wide trust in relay ingest verification — this applies equally to every message kind), an E2E-bridge `forward_message` mock for submit-flow tests, and unifying the four per-platform envelope parsers (the relay's `buzz-core` validation is the single normative gate; client parsers are display-oriented, now stated explicitly in NIP-FW).

## Testing

- `just ci` green (Rust fmt/clippy/unit, desktop lint + 3802 JS tests, mobile analyze + 886 tests, builds).
- 9 new integration tests in `crates/buzz-test-client/tests/e2e_relay.rs` run against a live relay: open→channel, →DM, private source by member vs. non-member (rejected), tampered embedded event (rejected), mismatched `k` (rejected), forward-of-a-forward (flattened), spoofed `fwd-src` type (rejected), never-published original (rejected).
- Live CLI smoke test verified the exact stored event shape end-to-end.
- Desktop unit tests for the forward builder/parser (`forwardMessage.test.mjs`, card render tests); mobile widget tests for the quote card (channel/private/DM label variants, malformed-tag fallback).

Screenshots follow in a comment (per repo screenshot policy).

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -1212,7 +1212,8 @@ pub fn resolve_channel_filters(
     rules: &[SubscriptionRule],
 ) -> HashMap<Uuid, ChannelFilter> {
     use buzz_core::kind::{
-        KIND_STREAM_MESSAGE, KIND_STREAM_REMINDER, KIND_WORKFLOW_APPROVAL_REQUESTED,
+        KIND_STREAM_MESSAGE, KIND_STREAM_MESSAGE_FORWARD, KIND_STREAM_REMINDER,
+        KIND_WORKFLOW_APPROVAL_REQUESTED,
     };
 
     let target_channels: Vec<Uuid> = if let Some(ref overrides) = config.channels_override {
@@ -1232,6 +1233,7 @@ pub fn resolve_channel_filters(
             let kinds = config.kinds_override.clone().unwrap_or_else(|| {
                 vec![
                     KIND_STREAM_MESSAGE,
+                    KIND_STREAM_MESSAGE_FORWARD,
                     KIND_WORKFLOW_APPROVAL_REQUESTED,
                     KIND_STREAM_REMINDER,
                 ]
@@ -1314,7 +1316,8 @@ pub fn resolve_dynamic_channel_filter(
     rules: &[crate::filter::SubscriptionRule],
 ) -> Option<ChannelFilter> {
     use buzz_core::kind::{
-        KIND_STREAM_MESSAGE, KIND_STREAM_REMINDER, KIND_WORKFLOW_APPROVAL_REQUESTED,
+        KIND_STREAM_MESSAGE, KIND_STREAM_MESSAGE_FORWARD, KIND_STREAM_REMINDER,
+        KIND_WORKFLOW_APPROVAL_REQUESTED,
     };
 
     // In Mentions/All mode, if the operator explicitly constrained channels
@@ -1337,6 +1340,7 @@ pub fn resolve_dynamic_channel_filter(
             kinds: Some(config.kinds_override.clone().unwrap_or_else(|| {
                 vec![
                     KIND_STREAM_MESSAGE,
+                    KIND_STREAM_MESSAGE_FORWARD,
                     KIND_WORKFLOW_APPROVAL_REQUESTED,
                     KIND_STREAM_REMINDER,
                 ]
@@ -1483,6 +1487,7 @@ mod tests {
             assert!(f.require_mention, "mentions mode requires mention");
             let kinds = f.kinds.as_ref().expect("should have kinds");
             assert!(kinds.contains(&buzz_core::kind::KIND_STREAM_MESSAGE));
+            assert!(kinds.contains(&buzz_core::kind::KIND_STREAM_MESSAGE_FORWARD));
             assert!(kinds.contains(&buzz_core::kind::KIND_WORKFLOW_APPROVAL_REQUESTED));
             assert!(kinds.contains(&buzz_core::kind::KIND_STREAM_REMINDER));
         }
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -22,7 +22,7 @@ use acp::{AcpClient, EnvVar, McpServer};
 use anyhow::Result;
 use buzz_core::kind::{
     KIND_MEMBER_ADDED_NOTIFICATION, KIND_MEMBER_REMOVED_NOTIFICATION, KIND_STREAM_MESSAGE,
-    KIND_STREAM_REMINDER, KIND_WORKFLOW_APPROVAL_REQUESTED,
+    KIND_STREAM_MESSAGE_FORWARD, KIND_STREAM_REMINDER, KIND_WORKFLOW_APPROVAL_REQUESTED,
 };
 use buzz_core::observer::{
     decrypt_observer_payload, encrypt_observer_payload, OBSERVER_FRAME_TELEMETRY,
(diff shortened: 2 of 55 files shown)


LordMelkor:
🤖 Desktop screenshots for the forward-message flow (mock E2E bridge, 1280x720).

### Forward action in the message menu
The `Forward message` item in the message action menu.

01-menu-item

### Forward dialog
Destination picker with note field and live WYSIWYG preview — the note renders through the same Markdown pipeline as the timeline, and the embedded original shows mention chips and `#channel` refs exactly as they will render after forwarding.

02-forward-dialog

### Forwarded card — open channel source
Quote card with author attribution, source-channel label, and jump-to-original link (source is visible to the viewer).

03-forwarded-open-source

### Forwarded card — private channel source
Private sources show the attribution label without a jump link for viewers who can't read the source channel.

04-forwarded-private-source

### Forwarded card — DM source
DM-sourced forwards label the source as a direct message and never leak a reference to the conversation.

05-forwarded-dm-source

PR #3426, opened 2026-07-28
docs(chart): fix ArgoCD example for native OCI sources (full artifact repoURL + path)

Kampe:
## Problem

`examples/argocd-app.yaml` uses the split form:

```yaml
repoURL: oci://ghcr.io/block/buzz/charts
chart: buzz
targetRevision: 0.1.0
```

On ArgoCD >= 3.0 (native OCI sources), the `chart` field is **ignored** for `oci://` repoURLs, so ArgoCD tries to pull the `charts` path itself and fails with `403 … repository:block/buzz/charts:pull denied` — a misleading error that reads like an auth problem. Additionally, spec validation rejects the Application without a `path` (`spec.source.repoURL and either spec.source.path or spec.source.chart are required`), since `chart` isn't recognized for OCI.

Hit both on ArgoCD 3.4.4 following the example verbatim.

## Fix

Use the full chart artifact path as `repoURL`, add `path: "."`, bump the pinned example version to the latest published chart (0.1.6), and leave a comment explaining both traps:

```yaml
repoURL: oci://ghcr.io/block/buzz/charts/buzz
path: .
targetRevision: 0.1.6
```

Verified working in production (ArgoCD 3.4.4, anonymous GHCR pull, chart 0.1.6).

Related open PRs/issues: none found.


tlongwell-block (approved):

tlongwell-block:
Thanks for the contribution, @Kampe! 🐝

PR #3431, opened 2026-07-28
fix(push): certificate-based APNs auth over HTTP/2

brow:
## Summary

Makes the push gateway able to actually deliver to Apple, by fixing transport and switching authentication:

- enable reqwest HTTP/2 for the APNs transport (APNs is HTTP/2-only; without the feature the TLS handshake fails before any request), with a compile canary so removing the feature fails the build
- replace ES256 provider-token (JWT/.p8) auth with client-certificate auth: a combined key+cert PEM is loaded via `reqwest::Identity` and installed on the client; the JWT machinery, bearer header, and `p256` dependency are removed
- config collapses `apns_key_path`/`apns_key_id`/`apns_team_id` into a single `BUZZ_PUSH_APNS_CERT_PATH`; `apns_topic` stays required (APNs pins the topic to the cert bundle id); no dual mode or token fallback
- mount the certificate Secret with `defaultMode: 0400` (Kubernetes defaults Secret files to 0644, which would leave the private key world-readable in the pod) and assert the mode in the chart render test
- remove the dead `RefreshCredential` outcome (enum variant, trait method, caller branch, metric, docs): nothing constructs it under cert auth
- chart schema now defines `apnsCert` and rejects legacy `apnsKey` overrides so stale token-auth values fail loudly

## Validation

- full `buzz-push-gateway` suite green: 22 passed, 0 failed, 7 ignored (opt-in live probe + PostgreSQL integration)
- mutation checks: deleting the identity install fails the mismatched-key test; removing `defaultMode` fails the render test; removing the http2 feature fails compilation
- live APNs sandbox probe through the gateway send path with the real (uncommitted) certificate: `400 BadDeviceToken`, proving Apple accepted the client certificate; control without identity returns `403 MissingProviderToken`
- identity-load failure matrix (empty, non-PEM, cert-only, key-only, encrypted, mismatched key/cert) all map to `ApnsError::Credential`
- clippy `-D warnings`, helm lint, full chart render script
- independently accepted by a red-team and an acceptance review at commit c7bbf7409; branch then merged `origin/main` (Cargo.lock only overlap) and the suite re-ran green at the pushed head

No credentials are committed; the chart mounts them from a Secret.


brow:
@codex review

brow:
@builderbot review

brow:
Consolidating: tomb directed all push-notification work onto the single feature branch push-ios-blockers-1-2 (draft #2744). The reviewed cert-auth commits (head 89de72496, judge-accepted) are merged there as f61f8419d. Closing this PR; branch apns-http2 retained.

PR #3437, opened 2026-07-29
fix(relay): stop overcounting NIP-45 COUNT for non-pushable filter constraints

### LINKED ISSUE #3373: relay: NIP-45 COUNT returns an accessible-channel-wide total when a filter's constraint never reaches SQL
## Summary

`filter_fully_pushable` (`crates/buzz-relay/src/handlers/req.rs`) decides whether a COUNT can be answered by a plain SQL `COUNT(*)`. When it returns true, the WebSocket COUNT handler (`handlers/count.rs`) and `POST /count` (`api/bridge.rs`) call `count_events()` with no post-filtering, on the promise in its doc comment of "an exact count without post-filtering".

For two filter shapes that promise does not hold, because the constraint never becomes a SQL predicate. The relay answers with the count of every matching event across all of the caller's accessible channels instead of the correct total. There is no error, just a wrong number.

## Affected shapes

**1. A single `#h` value that is not a channel UUID.** Callers derive `channel_id` by parsing the `#h` value, so a bare NIP-29 group id leaves the query with no channel predicate. `filter_fully_pushable` still reports it as pushable because the `"h"` arm only checks `tag_values.len() > 1`.

```json
{"kinds":[9],"#h":["general"]}
```

**2. An empty value set.** This reaches every generic-tag arm, plus `authors` and `ids`. `filters_match` rejects a present constraint with no matching value, so these match nothing, but `filter_to_query_params` drops the constraint rather than emitting a predicate.

```json
{"kinds":[9],"#t":[]}
{"kinds":[9],"authors":[]}
```

Correct answer in every case above is 0. `kinds: []` is already handled correctly, via the match-nothing sentinel the DB layer short-circuits on.

The REQ path is not affected. It always post-filters through `filters_match`, which enforces these strictly, so REQ and COUNT disagree on the same filter.

## Reproduction

Against a relay on current `main`:

1. Create a channel and post two kind:9 messages to it.
2. `POST /count` with `[{"kinds":[9],"#h":["general"]}]`.

Expected 0. Actual on my local relay was 4: the two messages in my own channel plus two more from an unrelated open channel left over from an earlier run. The same filter over REQ returns nothing, which is the correct behavior.

Substituting `[{"kinds":[9],"#t":[]}]` or `[{"kinds":[9],"authors":[]}]` gives the same shape of wrong answer. A control filter naming the channel's actual UUID (`[{"kinds":[9],"#h":["<uuid>"]}]`) correctly returns 2, so only the unpushable shapes are affected.

## Suggested fix

Report these as not pushable so they take the existing bounded `query_events` plus `filters_match` fallback, which already computes the correct total. A single `#h` that does parse as a UUID keeps its narrower `channel_id` predicate.

Falling back rather than short-circuiting to 0 matters for the `#h` case: `filters_match` compares the h-tag literal, and ingest only requires a parseable h tag for channel-scoped kinds, so an event can legitimately carry a non-UUID h tag and match.

A cheaper alternative for the empty-set half is to emit the match-nothing sentinel from `filter_to_query_params`, the way `kinds: []` already does, keeping the count exact at no query cost. That touches every consumer of the shared query builder, so it may be better as its own change.

Happy to send a PR.

1nderboi:
## Summary
Fixes #3373. `filter_fully_pushable()` was letting several filter shapes through to the fast SQL `COUNT` path that `filter_to_query_params()` actually drops instead of pushing as predicates — causing `COUNT` to overcount:

- A present-but-empty constraint (`authors: []`, `ids: []`, or any generic tag with zero values) matches nothing under `filters_match`, but the query builder just omits it rather than emitting a match-nothing predicate.
- A single `#h` value that isn't a channel UUID (e.g. a bare NIP-29 group id) was treated as pushable, but only UUID-valued `#h` is pushed as `channel_id` — a non-UUID value needs post-filtering since events can legitimately carry it as a literal tag.
- Any other generic tag (`#t`, `#a`, etc.) with a non-empty value list was allowed through when it should always require post-filtering.

All three now correctly fall back to the exact post-filtered COUNT path.

## Changes
- `crates/buzz-relay/src/handlers/req.rs`: reject empty `authors`/`ids`/generic-tag constraints, reject non-UUID single `#h`, reject any non-empty non-`h`/`p`/`d`/`e` generic tag.
- 6 new unit tests covering: UUID vs non-UUID single `#h`, multi-`#h`, empty authors/ids/generic-tag constraints, non-empty `#t`, and the plain kind+author baseline that should stay pushable.

## Test plan
- [x] `cargo fmt -p buzz-relay --check` — clean
- [x] Manual trace of `filter_fully_pushable` against `filters_match` / `filter_to_query_params` for each new case
- [ ] `cargo test -p buzz-relay` — **could not run to completion locally.** This shared dev machine's disk repeatedly hit ENOSPC mid-build (down to 0 bytes free at one point) even after `cargo clean`-ing target dirs; the `buzz-relay` dependency tree (aws-lc-sys, sqlx-postgres, opentelemetry_sdk, mesh-llm-native-runtime) alone exceeds available local headroom. Requesting CI run the full suite including the new tests in `req.rs`'s `mod tests`.

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -764,13 +764,26 @@ pub(crate) fn count_fallback_exceeded(candidate_count: usize) -> bool {
 /// in SQL by `filter_to_query_params` — meaning `count_events()` will produce
 /// an exact count without post-filtering.
 ///
-/// Pushed constraints: kinds, authors (single or multi), ids, since, until,
-/// channel_id (#h single), #p (single), #d (single, NIP-33-only kinds), #e (any),
-/// channel_ids (injected by caller).
+/// Pushed constraints: kinds, authors (single or multi, non-empty), ids
+/// (non-empty), since, until, channel_id (#h single, UUID-valued), #p
+/// (single), #d (single, NIP-33-only kinds), #e (any), channel_ids (injected
+/// by caller).
 ///
-/// Anything else (multi-#p, #t, #a, search, multi-#h, #d on non-NIP-33)
-/// requires post-filtering and cannot use the fast COUNT path.
+/// Anything else — multi-#p, #t, #a, search, multi-#h, a non-UUID single #h,
+/// #d on non-NIP-33 kinds, or any constraint present with zero candidate
+/// values (`authors: []`, `#t: []`, etc.) — requires post-filtering and
+/// cannot use the fast COUNT path. A present-but-empty constraint matches
+/// nothing under `filters_match`, but `filter_to_query_params` drops it
+/// instead of emitting a match-nothing predicate, so treating it as pushable
+/// would overcount.
 pub fn filter_fully_pushable(filter: &Filter) -> bool {
+    if filter.authors.as_ref().is_some_and(|a| a.is_empty()) {
+        return false;
+    }
+    if filter.ids.as_ref().is_some_and(|ids| ids.is_empty()) {
+        return false;
+    }
+
     // Check if filter exclusively targets NIP-33 kinds (needed for #d pushability).
     let is_nip33_only = filter.kinds.as_ref().is_some_and(|ks| {
         !ks.is_empty()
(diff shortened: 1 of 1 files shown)


1nderboi:
Closing in favor of #3374, which fixes the identical bug (filter_fully_pushable letting a non-UUID single #h and any present-but-empty constraint through to the fast COUNT path) the same way, was opened first, and additionally has an e2e wire-level regression test against a real relay (test_count_unpushable_filters_do_not_count_all_channels) on top of unit tests — stronger coverage than this PR's unit-tests-only approach.

For the record: locally confirmed this branch's fix works too — cargo test -p buzz-relay --lib req:: → 51 passed, 0 failed — but no need for two competing fixes on the same issue. Thanks @beardthelion for #3374.

PR #3438, opened 2026-07-29
Add gated development App Attest bypass

brow:
### What changed?

Add a non-default development App Attest policy for iOS Simulator enrollment. The bypass accepts only the versioned development sentinels and is unavailable unless the gateway is explicitly compiled with `dev-app-attest-bypass`.

At runtime, enabling the bypass now fails closed unless both gateway listeners are loopback addresses and `buzz-ios-sandbox` is the only enabled profile. Default builds continue to use Apple's verifier. The gateway also records attempts that reach the APNs HTTP send seam.

### Why?

The iOS Simulator cannot perform Apple App Attest, which otherwise makes local push enrollment structurally impossible. This provides a narrowly gated development path without adding bypass behavior to the production gateway build.

### How is it tested?

Build and run.

Added tests:

- [`dev_app_attest`](https://github.com/block/buzz/tree/main/crates/buzz-push-gateway/src/dev_app_attest.rs)
- [`config`](https://github.com/block/buzz/tree/main/crates/buzz-push-gateway/src/config.rs)
- [`AppAttestPolicy` compile-fail test](https://github.com/block/buzz/tree/main/crates/buzz-push-gateway/src/app_attest_policy.rs)

*🤖 This PR was authored [with an agent](codex://threads/019fab3a-420e-7d62-b39f-01c100fef027).*

### DIFF
--- a/crates/buzz-push-gateway/Cargo.toml
+++ b/crates/buzz-push-gateway/Cargo.toml
@@ -7,6 +7,10 @@ rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
 
+[features]
+default = []
+dev-app-attest-bypass = []
+
 [lib]
 name = "buzz_push_gateway"
 path = "src/lib.rs"
--- a/crates/buzz-push-gateway/src/apns.rs
+++ b/crates/buzz-push-gateway/src/apns.rs
@@ -218,6 +218,7 @@ impl PushTransport for ApnsTransport {
             AppProfile::BuzzIosProduction => &self.production_base_url,
             AppProfile::BuzzIosSandbox => &self.sandbox_base_url,
         };
+        crate::metrics::record_apns_send_attempt();
         let response = self
             .client
             .post(format!("{base_url}/3/device/{endpoint}"))
--- a/crates/buzz-push-gateway/src/app_attest_policy.rs
+++ b/crates/buzz-push-gateway/src/app_attest_policy.rs
@@ -1,0 +1,82 @@
+//! Selects the production Apple verifier or the feature-gated development stub.
+
+use crate::app_attest::{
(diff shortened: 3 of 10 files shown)


brow:
@codex review

brow:
@builderbot review

brow:
🤖 Superseded, closing. This work landed on the single feature branch `push-ios-blockers-1-2` (draft #2744) per tomb's single-branch rule, in a strictly later form.

Where it went: three commits (`6d7f118fa` / `06841bcf9` / `a8826c57b`) plus the guard restore `737457569`. The branch version is not a copy of this PR's `b342d21d4`, it is the reviewed evolution of it: the bypass is now reachable only through a capability token minted from validated config (`AppAttestPolicy::from_config`) rather than a bare `AppAttestPolicy::Development`, and config validation additionally requires both listeners loopback and `enabled_profiles` exactly `{BuzzIosSandbox}`, erroring rather than silently disabling on a non-canonical flag value. `crates/buzz-push-gateway/src/dev_app_attest.rs` is byte-identical between the two (sha256 `aeb7ef02...`), and the `push_gateway_apns_send_attempts_total` counter from this PR is present there as well.

Adversarial review happened against the branch version, not this one, and produced one landing condition that this PR does not carry: the `development_bypass_stays_a_non_default_feature` guard is the sole detection for flipping `default = ["dev-app-attest-bypass"]`, which matters because `Dockerfile.push-gateway:20` builds the release image with no `--features`. Mutating that line with the guard absent was caught by nothing (both arms 40/0, clippy clean); with the guard restored it fails by name in the plain `cargo test -p buzz-push-gateway` arm, and the resulting no-features binary was confirmed to contain the bypass symbols. Nothing is lost by closing this.

PR #3441, opened 2026-07-29
fix(desktop): preserve cross-install agent identities

### LINKED ISSUE #3414: Mention picker re-mints a persona on a second device: inbound kind:30177 persona_id link is discarded
## Summary

On a second device belonging to the same owner, the `@` mention picker offers a persona as *mintable* even when an instance of that persona is already a member of the channel and running elsewhere. Selecting it creates a brand-new agent identity (new keypair, new `kind:30177`) instead of p-tagging the existing one.

The information needed to prevent this is already on the wire — `ManagedAgentEventContent.persona_id` — but the receiving device discards the whole event because it has no local record to merge into, so it never learns the `pubkey → persona_id` mapping that the picker's dedup depends on.

Line numbers below are from **v0.5.0**.

## Repro

1. Mac A: create persona *Penelope*, mint an instance, add it to `#channel`. Works.
2. Mac B: same owner account, app freshly launched. The persona definition (`kind:30175`) syncs and auto-inserts. No local instance is created (by design).
3. Mac B: in `#channel`, type `@Pen…`. The picker shows **two** entries both labeled "Penelope" — the channel member, and a mintable persona.
4. Selecting the persona entry mints a second identity for the same persona and attaches it to the channel.

Now every `@Penelope` is a coin flip between two identities, and messages addressed to the one whose harness lives on the other machine are silently dropped.

## Root cause

`agentIdentityKey` in `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts:114` is what should collapse these two candidates:

```ts
if (candidate.personaId) {
  return `persona:${candidate.personaId}`;
}
```

and `agentCandidateRank` (same file, :158) already ranks `isMember` first, so on a merge the live channel member would correctly win over the mintable persona.

The merge never happens because the **member candidate has no `personaId` on device B**. `desktop/src/features/messages/lib/useMentions.ts:316`:

```ts
personaId: managedAgentPersonaIdsByPubkey.get(pubkey) ?? linkedPersonaId,
```

`managedAgentPersonaIdsByPubkey` is derived from the *local* managed-agent table, which on device B has no row for that pubkey. `linkedPersonaId` only resolves when a persona id happens to equal the agent pubkey, which is never true for minted instances. So the member candidate keys on label instead, the persona candidate keys on `persona:<uuid>`, and both survive into the list.

The persona is offered as mintable because the same local table gates it (`useMentions.ts:389`):

```ts
const personaCandidates = activePersonas
  .filter((persona) => !managedAgentPersonaIds.has(persona.id))
```

and selecting it routes to `createPersonaAgentMutation` in `desktop/src/features/messages/ui/useMentionSendFlow.ts:354`.

## Why the local table is empty — and why that's the fixable part

`apply_inbound_managed_agent` in `desktop/src-tauri/src/commands/personas/mod.rs:839` drops the inbound `kind:30177` projection entirely when there is no local match:

```rust
/// No match is a no-op: managed agents carry device-local secrets and are never
/// minted from a relay event — an agent that does not already exist locally has
/// no secret key to run with, so inserting a secretless shell would be useless
/// and misleading. This diverges from the persona path, which DOES insert on no
/// match (personas are secretless definitions).
```

The security reasoning is right and shouldn't change: device B must not fabricate a runnable record it has no key for. But the projection it throws away contains `persona_id`, which is not a secret and is exactly the join key the picker needs. Dropping the event is stricter than dropping the secrets.

## Impact

- Duplicate identities accumulate silently; each `@mention` from the non-minting device creates another.
- Mentions become nondeterministic once duplicates exist.
- Messages to the identity whose harness lives on the other machine are dropped with no error (correctly — `useMentionSendFlow.ts:260` skips non-local agents — but the user sees only silence).
- Related second-order failure: if device B has a *stale* local record (from an earlier accidental mint) whose harness was since deleted, the mention resolves to it and device B attempts a local `startAgentMutation` (`useMentionSendFlow.ts:265`), producing `cannot spawn agent <pubkey>: harness "<name>" was deleted`.

## Suggested fix

Keep the no-mint rule; record the association only.

On no-match in `apply_inbound_managed_agent`, persist the `(pubkey, persona_id, name)` triple into a separate non-runnable index rather than into `managed-agents.json`. Then source `managedAgentPersonaIdsByPubkey` from local records ∪ that index. The existing `agentIdentityKey` / `agentCandidateRank` logic then does the right thing with no further change: the two candidates collapse and the channel member wins.

A narrower alternative, if a new store is unwelcome: have the relay agent directory (`relayAgentsQuery`) carry `persona_id`, so `useMentions.ts:341` and `:316` can resolve it without any local state. That fixes the picker but leaves other cross-device persona↔instance lookups unaware.

Either way the desired end state is the one the code already reaches for on a single device — a persona with a live instance anywhere in the channel is *tagged*, not re-minted.

## Environment

- Buzz desktop v0.5.0, two macOS machines, one owner account
- Self-hosted relay (`ghcr.io/block/buzz`), agent runs under a custom harness on machine A

### LINKED ISSUE #3440: Deleting a relay-synced agent definition deletes it account-wide and orphans live agents
## Summary

A second Desktop install can receive an account-scoped agent definition without owning the corresponding local agent keypair. If the resulting bodyless/duplicate entry is deleted on that install, Buzz publishes a kind:5 tombstone for the shared kind:30175 definition. The tombstone is applied on every install, including the machine running the real agent body.

The surviving agent retains its original `persona_id`, but that definition no longer exists. There is no supported UI path to restore the definition at the old id or relink the agent to a replacement definition, so the live identity becomes orphaned.

This is a destructive second-order failure of #3414 (also related to #2648): the second install can first mint a ghost identity for the already-instantiated definition, then deleting the ghost's entry removes the shared definition rather than only the local duplicate.

## Reproduction

1. On install A, create a definition and start its managed agent.
2. Sign into the same Buzz account on install B. The account-scoped kind:30175 definition appears, but install B has no local agent record/keypair for install A's kind:30177 identity.
3. Start or DM the definition on install B. A new keypair/agent identity is minted with the same display name and persona id.
4. Delete the duplicate/bodyless entry on install B.
5. Observe a kind:5 targeting the shared kind:30175. The definition disappears on both installs.
6. Install A's original agent record remains but points to the now-missing `persona_id` and is reported as orphaned.

## Observed evidence

- Duplicate pubkey prefix `497d45dd…` published kind:0 at `2026-07-28T21:51:36Z` and 11 kind:9 events through `22:30:25Z`, with no kind:30175 of its own.
- The matching Rimac-Buzz kind:30175 definition was published about 90 seconds earlier at `21:50:08Z`.
- Manual recovery required reinserting the definition rows into `managed-agents.json` at their original ids and saving from Desktop; replacement kind:30175 events appeared at `00:26:43Z` and `00:26:49Z`.

## Expected behavior

- A relay-synced definition must not be treated as locally owned deletion authority. Removing it on a secondary install should be local-only and must not publish a kind:5 for the shared kind:30175.
- Before minting, Desktop should use the retained owner-authored kind:30177 `persona_id` association to detect an identity that exists on another install. It must not fabricate a runnable local shell without the secret key.
- If a shared definition is legitimately tombstoned, linked local agent identities should remain runnable as standalone snapshots rather than becoming dangling references.
- The edit UI should support relinking an orphaned agent to an active definition without changing its pubkey/keypair.

## Root causes

1. Inbound kind:30177 events are retained but no non-runnable identity index is exposed when there is no local record, so creation and mention flows cannot deduplicate by `persona_id`.
2. `delete_persona` publishes a kind:5 for a definition learned from the owner's relay, conflating local cleanup with account-wide deletion authority.
3. Inbound kind:30175 tombstones remove the definition without first detaching/materializing linked local agents, and the edit API has no persona-link repair operation.

Czaruno:
## Summary

- enforce `one persona_id, one keypair` at the central managed-agent key-generation boundary, covering Start, mention, channel-add, team-add, onboarding, and other creation surfaces
- fail closed until the exact relay+owner kind:30177/deletion history has been completely fetched and durably reconciled; incomplete, failed, capped, or unpageable backfills cannot authorize a new keypair
- expose retained owner-authored kind:30177 persona links as a non-runnable, community-scoped identity index and reuse existing agents instead of minting duplicates
- make deletion of a relay-synced/bodyless definition local-only so a secondary install cannot publish a kind:5 tombstone for the shared kind:30175
- queue a NIP-IA kind:9035 archive request when deleting a locally owned agent body, allowing Desktop discovery to hide the retired identity from the relay's kind:13535 snapshot
- when a legitimate shared-definition tombstone arrives, snapshot linked local agents into standalone records, preserve their pubkeys/keypairs, and publish repaired kind:30177 projections
- add an edit-time relink path for already-orphaned agents so they can adopt an active definition without identity rotation
- make tombstone suppression delivery-order independent and exclude deleted retained identities from the reference index

Closes #3414.
Closes #3440.

## Testing

Rebased onto `main` at `4d47aa834`; branch head `7433a135e`.

- 3,839 desktop frontend tests passed
- 1,931 Desktop Tauri tests passed, 14 ignored; 3 diagnostics passed
- 1,019 Flutter tests passed, 1 skipped
- workspace Clippy passed with warnings denied
- desktop, web, and mobile formatting/static analysis passed
- desktop and web production builds passed
- `cargo check` for Desktop Tauri passed
- the macOS `desktop-tauri-clippy` leg currently fails on three dead-code warnings in unchanged `main` file `desktop/src-tauri/src/linux_media.rs`, introduced by merged PR #3607; #3441 does not modify that file
- regression coverage asserts the key factory is never invoked when any identity already links the persona; it intentionally does not select which historical duplicate “wins”
- sync-readiness coverage for empty completed history, persistence-in-flight, fetch failure, and capped/unpageable tied timestamps
- deletion, archive-request, tombstone ordering, detach, and orphan relink regression coverage


Czaruno:
Manual verification and the two review questions, updated after the branch fix.

I reproduced the ghost-key defect on shipped Buzz 0.5.2 from a secondary install with no local key for the persona. The failing sequence produced kind:30177 and kind:44100 in the same second, followed by a new kind:0 from a second pubkey. The negative control from the install that already held the key was clean. Which key remained active varied between reproductions, so that ordering is evidence of danger—not an invariant or a test expectation.

The invariant is now enforced at the backend key-generation boundary: one `persona_id`, one keypair. Every creation surface, including channel add, converges on `create_managed_agent`; persona-backed creation cannot invoke the key factory when either a local agent record or a retained account-scoped kind:30177 already links that persona. The regression test supplies both an original and a prior duplicate and asserts that the key factory is never called; it does not select a winner.

The empty-versus-unknown gap is also closed. Persona-backed creation fails closed until the exact relay+owner history backfill is complete and durably reconciled. Fetch failure, reconciliation failure, pagination cap, and an unpageable tied-timestamp boundary all leave the reference index unknown and key creation blocked.

Deletion behavior remains deliberately asymmetric: removing a relay-synced/bodyless definition is local-only and cannot tombstone the shared kind:30175; deleting a locally owned body also queues the NIP-IA kind:9035 archive request so Desktop discovery can hide the retired pubkey from the relay's kind:13535 snapshot. A foreign ghost that the install does not own is not guessed at or automatically archived.

The branch is rebased onto current `main` at `7433a135e`. Validation after the rebase passed 3,839 desktop frontend tests, 1,931 Desktop Tauri tests, 3 Tauri diagnostics, and 1,019 Flutter tests, plus workspace Clippy, static checks, and desktop/web builds. The only local gate failure is macOS Tauri Clippy on three dead-code warnings in unchanged `main` file `desktop/src-tauri/src/linux_media.rs`, introduced by merged PR #3607; this branch does not modify that file.

Czaruno:
Closing this. The bug it fixes is still there on main. This branch is no longer the way to fix it.

It is 389 commits behind and conflicts in six files: app_state.rs, commands/agent_models.rs, commands/agents.rs, personas/inbound.rs, useAgentsDataRefresh.ts, usePersonaSync.ts. #7131 rewrote most of those when it rebound explicit agent profiles to their exact key. Rebasing here means rewriting the change against code that moved, and this branch never drew a review to preserve anyway.

The problem itself: a second install of an agent mints a new keypair under the same display name, and there is no repair path once it happens. #5667 tracks that. #6997 asks for the owner-side attestation the fix needs. Those threads are live and this one is not.

The diff stays here if anyone wants the approach.

PR #3443, opened 2026-07-29
fix(desktop): prompt immediate restart when respond-to gate changes

surfingdev:
## Problem

When you change "Who can talk to this agent" (`respond_to`) on an agent that's already running, Desktop writes the new setting to disk immediately and correctly — but it never restarts the running `buzz-acp` process. That process keeps enforcing the OLD permission in memory until it happens to restart on its own (an idle-based auto-restart policy, gated to 3+ minutes idle plus other conditions) or someone restarts it by hand. In the meantime, mentions from people who should now be allowed under the new setting are silently dropped, and vice versa. The only indication anything is stale is a passive "Restart required" badge — easy to miss, and especially risky for a setting whose entire point is access control (buzz#2501, buzz#2950).

## What this changes

Adds an explicit "Restart now" action to the save toast whenever the save changed the `respond_to` gate and the backend confirms the running process is now out of sync with disk — mirroring the existing "saved while stopped → Start now" toast pattern already used elsewhere in `AgentInstanceEditDialog`. The predicate driving it (`shouldOfferImmediateRespondToRestart`) is a pure function, and the restart itself reuses only existing primitives: `respawnManagedAgentWithRules`, `startManagedAgent`/`stopManagedAgent`, `clearActiveTurnsForAgentOnStop`, and the backend's already-computed `needsRestart` drift signal. No Rust/backend changes, no new IPC, no schema changes — this is purely surfacing an existing signal more assertively.

Rebased onto main's #3352 (stateless file-size ratchet replacing the static override ledger). `AgentInstanceEditDialog.tsx` was already at the line cap and can't grow further under that rule, so the pre-existing "saved while stopped → Start now" toast was extracted alongside the new restart prompt into `agentSaveOutcomePrompts.ts` (renamed from `respondToRestartPrompt.ts`) — a net line reduction at the call site instead of a net addition. No behavior change to either toast.

- Propagating a persona/definition-level `respond_to` edit to already-linked instances — `update_persona` only propagates avatar/display_name today.
- The Desktop mention-picker/autocomplete eligibility work — in flight on other open PRs (buzz#2987).

## Test plan
- [x] `node --import ./test-loader.mjs --experimental-strip-types --test "src/**/*.test.mjs"` — 3770/3770 pass (full desktop suite, rebased onto main @ 58d2dfc)
- [x] `npx tsc --noEmit` — clean
- [x] `npx biome check` on touched files — clean
- [x] `node desktop/scripts/check-file-sizes.mjs` (base `origin/main`) — clean
- [ ] Manual verification in Desktop: edit an existing agent's "Who can talk to this agent" to Anyone/allowlist while running, confirm the "Restart now" toast appears and a non-owner can then @mention it

🤖 Written by Claude, running as an agent inside Buzz (buzz-acp) — not the Claude Code CLI.

### DIFF
--- a/desktop/src/features/agents/lib/managedAgentControlActions.test.mjs
+++ b/desktop/src/features/agents/lib/managedAgentControlActions.test.mjs
@@ -4,6 +4,7 @@ import test from "node:test";
 import {
   startManagedAgentWithRules,
   respawnManagedAgentWithRules,
+  shouldOfferImmediateRespondToRestart,
 } from "./managedAgentControlActions.ts";
 
 function agent(overrides = {}) {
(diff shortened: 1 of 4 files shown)


surfingdev:
Rebased onto `main` to resolve the merge conflict — `main` landed #3352 (stateless file-size ratchet against the base tree) while this was open, deleting the static override ledger this branch's diff touched. Under the new rule, `AgentInstanceEditDialog.tsx` (already over the 1000-line cap) can no longer grow at all, so I extracted the pre-existing "saved while stopped → Start now" toast alongside the new restart prompt into `agentSaveOutcomePrompts.ts` — same behavior, net line reduction instead of net addition.

Re-verified against the rebased tip (`58d2dfc`):
- `tsc --noEmit` — clean
- `biome check` on touched files — clean
- `node desktop/scripts/check-file-sizes.mjs` (base `origin/main`) — clean
- Full desktop suite: 3770/3770 pass

Still open: manual verification in Desktop (toast appears, non-owner can @mention post-restart).

PR #3455, opened 2026-07-29
fix(tls): trust the OS certificate store on relay WebSocket connections

benjaminkitt:
## Problem

The desktop app's relay WebSocket trusts only a compiled-in Mozilla root list and ignores the OS trust store, so Buzz cannot connect to a relay from a machine behind a TLS-inspecting proxy (Netskope, Zscaler, Palo Alto, …). The connection fails with:

```
invalid peer certificate: UnknownIssuer
```

There is no user- or admin-side workaround: installing the corporate CA into the system keychain has no effect, and neither `SSL_CERT_FILE` nor `NODE_EXTRA_CA_CERTS` is consulted on this path.

This is easy to misdiagnose, because everything *else* in the app works. HTTPS requests go through `reqwest`, which uses `rustls-platform-verifier` and **does** read the OS trust store. Only the WebSocket path is affected, so the app looks healthy right up until the relay never connects.

## Root cause

1. `desktop/src-tauri/Cargo.toml` and the root `Cargo.toml` build `tokio-tungstenite` with only the `rustls-tls-webpki-roots` feature. `rustls-tls-native-roots` is not enabled anywhere in the workspace, so feature unification does not compensate.
2. `desktop/src-tauri/src/native_websocket.rs` calls bare `connect_async(url)` with no custom `Connector`, so tokio-tungstenite builds its default root store — which under that feature set is the bundled Mozilla roots only.
3. `desktop/src/shared/api/relayClientSession.ts` and `readOnlyRelayClient.ts` reach the relay via `invoke("plugin:websocket|connect")`, so relay traffic goes through the Rust path above rather than the webview. (The webview would have used the OS trust store and would not have this problem.)

The same applies to the workspace dependency, so `buzz`, `buzz-acp` and `buzz-dev-mcp` are affected too. `buzz-agent` uses `reqwest` only and is unaffected.

Confirmed against a v0.5.0 release build: the binary contains the Mozilla root names (`ISRG Root X1`, `DigiCert Global Root G2`, …) and contains `SecTrustEvaluateWithError` (rustls-platform-verifier, via reqwest) but no `rustls_native_certs` / `load_native_certs` symbols.

## Fix

Enable `rustls-tls-native-roots` alongside the existing feature, in both manifests.

The two root-store features are **additive** — tokio-tungstenite's `src/tls.rs` pushes both sets into the same `RootCertStore`, and the webpki branch is not gated on `not(rustls-tls-native-roots)`. So this preserves today's behaviour for everyone (the bundled Mozilla roots are still there) and additionally honours the platform trust store.

## Verification

Reproduced with a standalone tokio-tungstenite 0.29 probe against a real proxied relay, connecting with each root store in turn:

```
(a) webpki-roots only      => FAILED: invalid peer certificate: UnknownIssuer
    native cert load: added=148 ignored=0 errors=[]
(b) native + webpki roots  => SUCCESS (TLS + WS handshake completed)
```

I then built the full app from this branch and confirmed the relay connects normally on the affected machine, and that the rebuilt `buzz` / `buzz-acp` / `buzz-dev-mcp` sidecars pick up the change through the workspace dependency.

## Alternative worth considering

Since the app already depends on `rustls-platform-verifier` through reqwest, an arguably tidier fix is to build a `Connector::Rustls` from the platform verifier and pass it to `connect_async_tls_with_config`. That would make WebSocket trust behaviour match HTTP trust behaviour instead of having the two paths use different trust stores. Happy to rework this PR that way if you'd prefer it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9998,6 +9998,7 @@ dependencies = [
  "futures-util",
  "log",
  "rustls",
+ "rustls-native-certs",
  "rustls-pki-types",
  "tokio",
  "tokio-rustls",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -122,7 +122,7 @@ moka    = { version = "0.12", features = ["sync"] }
 futures-util = "0.3"
 
 # WebSocket client (test client)
-tokio-tungstenite = { version = "0.29", features = ["rustls-tls-webpki-roots"] }
+tokio-tungstenite = { version = "0.29", features = ["rustls-tls-webpki-roots", "rustls-tls-native-roots"] }
 url = "2"
 
 # Property-based testing (dev-only)
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -11200,6 +11200,7 @@ dependencies = [
  "futures-util",
  "log",
  "rustls",
+ "rustls-native-certs",
  "rustls-pki-types",
  "tokio",
  "tokio-rustls",
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -83,7 +83,7 @@ infer = "0.19"
 hex = "0.4"
 ed25519-dalek = "=3.0.0-rc.0"
 tokio = { version = "1", features = ["fs", "sync", "rt", "macros", "time", "net", "io-util"] }
-tokio-tungstenite = { version = "0.29", features = ["rustls-tls-webpki-roots"] }
+tokio-tungstenite = { version = "0.29", features = ["rustls-tls-webpki-roots", "rustls-tls-native-roots"] }
 tokio-util = { version = "0.7", features = ["rt"] }
 bytes = "1"
 futures-util = "0.3"


ptahdunbar:
Ran into this exact issue independently — same `invalid peer certificate: UnknownIssuer` handshake failure connecting to a relay from behind a TLS-inspecting corporate proxy (Zscaler), and confirmed `rustls-tls-native-roots` fixes it once wired up correctly.

Heads up on why CI is red here: the diff only edits `Cargo.toml` (workspace + `desktop/src-tauri/Cargo.toml`) but doesn't regenerate the matching `Cargo.lock` files. Adding `rustls-tls-native-roots` pulls in a new transitive dependency (`rustls-native-certs`) that isn't in either lockfile yet, and CI builds with `cargo build --locked`, so it fails with `cannot update the lock file ... because --locked was passed to prevent this` (see the failing Docker image build jobs). Running `cargo build` (or `cargo build --manifest-path desktop/src-tauri/Cargo.toml`) locally after the `Cargo.toml` edit and committing the resulting lockfile changes should get this green. Branch is also ~100 commits behind `main` at this point, so a rebase is probably needed regardless.

scottopell:
Adding support for this one. 👍 
Looks like the last push occurred during a gh actions outage so CI is stuck, but the diff is still correctly this simple.

My motivating use case is similar but a bit different, running behind a port-forward effectively so I generate my own cert for the clients since hostname is internal.

benjaminkitt:
Rebased onto current `main` and force-pushed (`6c264a3d`, no conflicts; lockfiles still regenerated for `--locked` builds; DCO green). The new workflow runs are sitting in `action_required` — could a maintainer approve them so CI can run? Thanks!

xarielx (commented):
Strong support for this one. The one-line version for a busy maintainer: HTTP already trusts the OS certificate store, but the WebSocket does not, so private-CA deployments half-work in a way that is hard to diagnose.

Buzz is meant to run as a private team relay, and the common shape of that is "internal network only, TLS terminated with the company's own CA." On a split-horizon internal DNS zone a public cert is often not even obtainable, so operators cannot "just use a public cert." Every managed device already trusts the corporate root via MDM, yet the baked-in Mozilla roots make the desktop WebSocket the one component that cannot honor it.

Keeping `rustls-tls-webpki-roots` and adding `rustls-tls-native-roots` trusts both sets: the default connector loads native certs and extends the same root store with webpki roots when both features are on. Public and hosted relays keep working unchanged, which removes the usual objection to flipping the default.

This PR also covers the gap the workspace-only PRs miss: `desktop/src-tauri/Cargo.toml` redeclares `tokio-tungstenite` independently, so a workspace-only change leaves the desktop binary on webpki roots. Editing both manifests and regenerating both lockfiles is what CI needs for `--locked`. All three `connect_async` sites use the default connector, so no code change is needed.

I reproduced this diff independently: both features resolve on the single `tokio-tungstenite v0.29.0`, `rustls-native-certs` is the one new lockfile edge, and a full `desktop-release-build` produces `Buzz.app` and the dmg clean with the combined feature set. The only thing blocking this is a maintainer approving the queued `action_required` CI runs. Could a maintainer kick those off?

xarielx (approved):
Approve, added my comment to the PR directly.

olegpulatov:
hi @wesbillman @wpfleger96 @tlongwell-block 
could a one line change be reviewed please after couple of months? its blocking on private networks access
and there are 7 PRs on the same and i wanted to do an 8th one
sorry to bother but its very annoying to rebuild all apps for it and its here for long time
may I kindly suggest to use ai auto review and out triage? i can point to some solution we use in other project

shawnhank:
Adding a data point from a self-hosted setup, since this is the fix I needed as well.

## Summary
- My relay runs behind a reverse proxy that serves a certificate issued by a private two-tier CA (step-ca). The root is installed and trusted in the macOS System keychain.
- With the stock build, `buzz-acp` fails every connect with `WebSocket error: IO error: invalid peer certificate: UnknownIssuer`, and Desktop shows "Can't reach the relay."
- On the same machine, at the same time, `security verify-cert` reports `certificate verification successful` for the full chain, and `curl` to the relay returns 200. The OS trusts the chain; only the WebSocket path doesn't consult it.
- I rebuilt Desktop from current `main` with `rustls-tls-native-roots` enabled on `tokio-tungstenite` in both `Cargo.toml` and `desktop/src-tauri/Cargo.toml`. After that, Desktop and all managed agents connected and passed NIP-42 auth on the first attempt, and agent turns completed normally.

One point worth making explicit for review: this change doesn't add any certificate material to Buzz. The user installs their own root in their OS trust store, which is the same step they already take for browsers and `curl`, and Buzz honors it. Keeping `rustls-tls-webpki-roots` alongside it, as this PR does, means nothing changes for anyone on a public CA.

## Test plan
- [x] macOS 26.6.2 (Apple Silicon), Desktop 0.5.23 built from `main` with the feature change
- [x] Before: `buzz-acp` logs `invalid peer certificate: UnknownIssuer` on every connect attempt
- [x] After: relay logs show 6 WebSocket connections and 6 `NIP-42 auth successful` (Desktop plus managed agents)
- [x] Agent round-trip: message sent, agent responded, model backend received requests
- [x] Isolated `tokio-tungstenite` 0.29 check against the same relay: `rustls-tls-webpki-roots` alone → `UnknownIssuer`; `rustls-tls-native-roots` alone → `101` + `AUTH`; both features (this PR's configuration) → `101` + `AUTH`

Happy to test anything else that would help get this merged. Thanks, @benjaminkitt, for keeping it rebased.

PR #3473, opened 2026-07-29
feat(desktop): register nano-core as a conversational ACP runtime

0-CYBERDYNE-SYSTEMS-0:
## What

Registers **nano-core** as a known ACP runtime so it can be selected as an agent harness from the desktop runtime picker.

nano-core is an agnostic agent runtime whose ACP surface is built on the official `@agentclientprotocol/sdk` (v1.3.0). Verified against a live handshake:

```json
{"protocolVersion":1,
 "agentCapabilities":{"loadSession":false,
   "promptCapabilities":{"image":false,"audio":false,"embeddedContext":false},
   "mcpCapabilities":{"http":false,"sse":false}},
 "authMethods":[],
 "agentInfo":{"name":"nano-core","title":"nano-core ACP Preview","version":"0.1.0"}}
```

## Launch shape — subcommand, not a bare binary

Unlike the zero-arg native entrypoints, nano-core reaches ACP through a subcommand of its own CLI, so it joins the **Goose-shaped arm** of `default_agent_args` in both the desktop catalog and the `buzz-acp` harness.

The `["acp"]` default is the launch *shape*, not a complete command line. `bin/nano-core.js` locates its checkout by walking **up** from the working directory; agents launch from the nest (`~/.buzz`), so the walk never reaches the repo and the CLI exits before the ACP handshake — presenting as a dead agent with no protocol error. Operators must override the agent's args with `--repo <abs-path> acp`. That flag is position-independent and has no env-var fallback, so it cannot be defaulted here (the path differs per machine). Both the install hint and a regression test cover the override.

## Capability scoping — conversational / ops, not a workspace agent

| Setting | Reason |
|---|---|
| `mcp_command: None` | **Load-bearing.** nano-core *rejects* a non-empty `mcpServers` in `session/new` with `invalidParams` rather than ignoring it, so attaching the dev MCP server would break session creation outright. |
| `skill_dir: None` | Every turn executes in a per-run Docker container, which cannot reach a `buzz-cli` symlink in the host workspace. |
| `supports_acp_native_config: false` | No `session/set_config_option` implemented. |
| `model_env_var` / `provider_env_var` | Pi runs inside the container and reads `PI_MODEL` / `PI_PROVIDER`. `PROFILE.json` defaults are softer than the environment, so injected values win. |
| `auth_probe_args: None` | `authMethods` is empty; Docker + an onboarded profile are the real prerequisites and are surfaced via `login_hint`. |

The preview advertises neither `modes` nor `models` in `session/new`, so the permission-mode and model-switch calls are skipped by the existing response-shape gates (`agent_supports_mode`) rather than failing.

Session semantics are safe for concurrent channels: `session/new` mints an independent id per call and validates the `cwd`. Only a second *concurrent prompt on a single session* is refused, which matches the harness's one-turn-per-session model.

## Catalog split

`discovery.rs` is already over the file-size ratchet, which forbids any growth. The runtime catalog is pure data that grows with every new runtime, so it moves to `discovery/runtime_catalog.rs` — this keeps the discovery logic readable and leaves both files well inside the budget. The nano-core tests live beside the data they pin.

## Testing

- `cargo test -p buzz-acp` and the full desktop `managed_agents` suite (899 tests) pass.
- `cargo clippy --all-targets` clean on both crates; `cargo fmt --check` clean.
- Live ACP `initialize` handshake verified against the real binary (output above).

`just ci` is green except for `mobile-test`, which fails identically on unmodified `origin/main` — `git diff origin/main -- mobile/` is empty for this branch.

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -692,7 +692,9 @@ pub(crate) fn normalize_agent_command_identity(command: &str) -> String {
 
 fn default_agent_args(command: &str) -> Option<Vec<String>> {
     match normalize_agent_command_identity(command).as_str() {
-        "goose" => Some(vec!["acp".to_string()]),
+        // nano-core's ACP entrypoint is a subcommand of its own CLI
+        // (`nano-core acp`), not a bare stdio binary — same shape as Goose.
+        "goose" | "nano-core" => Some(vec!["acp".to_string()]),
         "codex" | "codex-acp" | "claude-agent-acp" | "claude-code-acp" | "claude-code"
         | "claudecode" | "buzz-agent" => Some(Vec::new()),
         _ => None,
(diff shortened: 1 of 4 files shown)


0-CYBERDYNE-SYSTEMS-0:
Opened against the wrong base — this belongs in our own fork, not upstream. Closing.

PR #3481, opened 2026-07-29
docs(readme): clarify which release asset to download per platform

dan-sheehan:
## Summary

The "I just want to try the app" section names platforms generically (macOS `.dmg`, Linux `.AppImage` / `.deb`, Windows `.exe`), but the release publishes five assets, including two separate macOS builds. A first-time user on a Mac has no way to tell whether they need `aarch64` or `x64`, and nothing sets expectations for the SmartScreen warning on the unsigned Windows build.

This replaces that sentence with a platform-to-filename table, a one-line note on how to check which Mac you have, and a note that the Windows build is unsigned and what the warning looks like.

Filenames use `<version>` rather than `0.5.0` so the table doesn't go stale each release.

### Related issue

None found. Searched open issues and PRs for README/download/install topics.

### Testing

Docs-only change, no code paths touched. Verified the table and paragraph breaks render correctly in GitHub's markdown preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,19 @@ New to Buzz? Pick the path that matches you.
 
 ### I just want to try the app
 
-Grab a packaged build from the [latest release](https://github.com/block/buzz/releases/latest) — macOS (`.dmg`), Linux (`.AppImage` / `.deb`), or Windows (`.exe`). Install it like any other app.
+Grab a packaged build from the [latest release](https://github.com/block/buzz/releases/latest):
+
+| Platform | File |
+|---|---|
+| macOS (Apple Silicon) | `Buzz_<version>_aarch64.dmg` |
+| macOS (Intel) | `Buzz_<version>_x64.dmg` |
+| Linux (x86_64) | `Buzz_<version>_amd64.AppImage` or `Buzz_<version>_amd64.deb` |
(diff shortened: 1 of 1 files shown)


tlongwell-block (changes requested):
The artifact table matches the current release and release workflow, but the Mac identification sentence is already incomplete: Apple Silicon now includes M5, while the text only names M1/M2/M3/M4. Please make this generation-independent (for example, “`Chip: Apple …` means Apple Silicon; `Processor: Intel …` means Intel”) rather than enumerating chip generations.

Also, SmartScreen is configuration/reputation-dependent, so “may show” is more accurate than “will show”; managed systems may not offer **Run anyway**.

tlongwell-block (approved):
The follow-up commit makes the Mac guidance generation-independent and correctly qualifies the SmartScreen behavior. Re-reviewed at 3b6a011fcde92cd99dcd5a977b3a07d43eb9ef7f.

dan-sheehan:
Thanks for the review and update. The generation-independent wording makes sense, and this looks good to me.

tlongwell-block (approved):

tlongwell-block:
Thanks for the contribution, @dan-sheehan! 🐝

Issue #3283, opened 2026-07-28
Local dev: loopback hosts are seeded as separate communities, so desktop-managed agents silently discover 0 channels


linearuncle:
## Summary

In local dev, `scripts/seed-local-community.sh` seeds `localhost:3000` and `127.0.0.1:3000` as **separate rows in `communities`** — i.e. two isolated tenants — while the desktop app connects as `ws://localhost:3000` but spawns managed-agent harnesses with the **canonicalized** URL `ws://127.0.0.1:3000`. The result: every desktop-managed agent lands in an empty tenant, discovers 0 channels, and silently never responds to mentions. This cost us several hours to debug, because nothing errors — the agents just sit idle.

## Environment

- macOS, current `main` (`6f19a18` era), `just setup` + `just dev`
- Default `.env` (`RELAY_URL=ws://localhost:3000`), default `scripts/instance-env.sh` (`BUZZ_RELAY_URL="${BUZZ_RELAY_URL:-ws://localhost:3000}"`)

## Steps to reproduce

1. `just setup` (seeds communities), `just dev`
2. Complete desktop onboarding against `ws://localhost:3000`; create a channel
3. Create a managed agent (e.g. Honey), add it to the channel, start it
4. Mention the agent in the channel

**Expected:** the agent responds.

**Actual:** nothing happens. The harness log (`~/Library/Application Support/<app-id>/agents/logs/<agent-pubkey>__<relay-hash>.log`) shows:

```
INFO buzz_acp: buzz-acp starting: relay=ws://127.0.0.1:3000 ...
INFO buzz_acp: discovered 0 channel(s)
WARN buzz_acp: no channel subscriptions resolved — agent will sit idle
```

## Root cause

Three pieces interact:

1. **Seeding:** `scripts/seed-local-community.sh` inserts each loopback host variant (`localhost`, `127.0.0.1`, `localhost:3000`, `127.0.0.1:3000`) as its **own** `communities` row. The comment says these are "loopback aliases", but with row-zero host binding each row is a fully isolated tenant:

   ```
   SELECT id, host FROM communities;
   -- 4 rows, 4 distinct community ids
   ```

2. **Desktop app tenant:** `scripts/instance-env.sh` defaults the app to `ws://localhost:3000`, so all channels/messages/profiles are created in the `localhost:3000` tenant.

[...]

PR #2446, opened 2026-07-23
fix(desktop): dial managed agents on the configured relay URL, not the canonical runtime key

lewiswang0516:
Fixes #2444

## Problem

Since #2122 (shipped in desktop v0.4.23), every managed-agent spawn path passes `ManagedAgentRuntimeKey.relay_url` - the canonical identity form - to the child as `BUZZ_RELAY_URL`. The canonicalizer folds every loopback spelling to `127.0.0.1`, but the relay derives the community boundary from the Host header. So with a community configured as `ws://localhost:3000`:

- the UI posts messages into the `localhost:3000` community
- every agent connects to the `127.0.0.1:3000` community, which is empty

Each agent logs `discovered 0 channel(s)` and `no channel subscriptions resolved - agent will sit idle`, shows as online, and silently never answers a mention. Full diagnosis with DB evidence is in #2444.

This also contradicts the contract documented on `buzz_core::relay::normalize_relay_url` itself:

> Connection code may retain the configured URL; this canonical form is for identity, receipts, status and deduplication.

## Fix

Keep the canonical key for identity, receipts, and log paths. Dial the caller-supplied requested URL:

- `spawn_agent_child` dials its `relay_url` parameter verbatim; `runtime_key` stays identity-only
- `start_managed_agent_process`, `start_pair`, and the launch restore path pass their already-resolved requested URL instead of `key.relay_url`
- `reconcile_managed_agent_runtimes` forwards the requested community URL to `start_pair` instead of the canonical echo
- `probe_agent_relay_access` probes over the requested URL, so the spawn-eligibility probe hits the same community the child will join
- `start_pair` now fills `requested_relay_url` in its status rows, and the Active Communities settings card prefers `requestedRelayUrl` for start/stop/restart, so a manual restart re-dials the configured URL

## Testing

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib`: 1560 passed
- `cargo clippy --all-targets` and `cargo fmt --check` on the Tauri crate: clean
- `tsc --noEmit` and `biome check` on the changed frontend file: clean
- Honest gap: no new automated test covers the spawn wiring itself, since `spawn_agent_child` forks a real child process. Verified the failure mode end to end on the affected machine (v0.4.23, local relay): before the fix, children spawn with `BUZZ_RELAY_URL=ws://127.0.0.1:3000` (checked via `ps eww`) and discover 0 channels; the configured community URL is `ws://localhost:3000`.

[...]

PR #3484, opened 2026-07-29
fix(desktop): spawn managed agents with the configured relay URL

### LINKED ISSUE #2444: [Bug] Desktop 0.4.23 spawns managed agents on ws://127.0.0.1:3000 while the community uses ws://localhost:3000 - agents silently stop responding
## Summary

On desktop v0.4.23 (installed DMG, macOS), managed agents are spawned with `BUZZ_RELAY_URL=ws://127.0.0.1:3000` even though the active community is configured as `ws://localhost:3000`. Because the relay derives the community boundary from the Host, `localhost:3000` and `127.0.0.1:3000` are two fully isolated communities. The result: the UI posts messages into the `localhost:3000` community while every agent sits in an empty `127.0.0.1:3000` community, so agents never see any message and never reply. There is no error anywhere - the only hint is a low-key log line in the agent log.

## Environment

- Desktop app: v0.4.23, installed from DMG (GUI launch), macOS (Darwin 25.5)
- Relay: local dev relay at port 3000 (`just relay`)
- Community config in the app (localStorage `buzz-communities`): `relayUrl: "ws://localhost:3000"`
- No `BUZZ_RELAY_URL` in the app process environment, no launchctl env

## Steps to reproduce

1. Run a local relay on port 3000.
2. In the desktop app, use a community configured as `ws://localhost:3000`, create channels, add managed agents to a channel.
3. Install/launch desktop v0.4.23 from DMG and let it restore the managed agents.
4. Mention an agent in the channel.

## What happens

- The user's messages land in the `localhost:3000` community (verified in the relay DB: `events.community_id` maps to the `localhost:3000` row in `communities`).
- Every spawned `buzz-acp` child has `BUZZ_RELAY_URL=ws://127.0.0.1:3000` (verified via `ps eww`). Their membership discovery (kind 39002 query) runs against the `127.0.0.1:3000` community, which is empty, so each agent logs:

```
WARN buzz_acp: no channel subscriptions resolved - agent will sit idle
```

- Agents show as online (presence is set) but never respond to anything.

## Notes on the source of the rewrite

- The agent records in `managed-agents.json` have an empty `relay_url`, so the spawn falls back to the workspace relay (`effective_agent_relay_url` in `desktop/src-tauri/src/relay.rs`).
- On current `main`, the fallback chain (workspace override from `apply_workspace` -> env -> build-time -> `DEFAULT_RELAY_WS_URL`) is all `ws://localhost:3000`, and the frontend passes the stored community `relayUrl` unchanged. I could not find the `localhost -> 127.0.0.1` rewrite on `main`, so it seems specific to the 0.4.23 build (repeated agent restarts always produce `127.0.0.1`, so it is not a launch-order race).
- An earlier build on the same machine spawned agents with `ws://localhost:3000` (visible in older agent logs), so this looks like a regression introduced in or shortly before 0.4.23.

## Why this is nasty to debug

1. Nothing fails loudly. The agent connects, authenticates, sets presence to online, and then silently subscribes to nothing.
2. `localhost` vs `127.0.0.1` is invisible in the UI. The user sees online agents in the member list that simply never answer.
3. Setting `BUZZ_RELAY_URL` as a per-agent env var is rejected (it is on `RESERVED_ENV_KEYS`, which is reasonable), so the only workarounds are the per-agent Relay URL field in the Advanced section, or changing the community relay URL - the latter silently forks the community and hides all previous history.

## Suggested fixes

1. Find and remove the `localhost -> 127.0.0.1` rewrite in the agent spawn path so the child relay URL always matches the workspace relay the UI uses.
2. Consider normalizing loopback hosts (`localhost`, `127.0.0.1`, `[::1]`) to one canonical community key on the relay side, since they are the same machine and the split is never what a local user wants.
3. Make the idle state visible: when an agent discovers 0 channels, surface a warning in the desktop UI (the log line is easy to miss), since an online-but-deaf agent looks identical to a working one.

JHSeo-git:
## Summary

On a local relay, desktop-managed agents silently never respond: the relay derives the community boundary from the HTTP Host header, but every managed-agent spawn path handed the child `ManagedAgentRuntimeKey.relay_url` — the canonical identity form, whose normalization folds every loopback spelling to `127.0.0.1`. With a community configured as `ws://localhost:3000`, the UI writes all channels, memberships, and messages into the `localhost:3000` community while every agent connects to the empty `127.0.0.1:3000` community. Each harness logs `discovered 0 channel(s)` / `no channel subscriptions resolved — agent will sit idle`, the member list shows the agent stuck on "Waking", and mentions are never answered. No error surfaces anywhere.

This also violates the contract documented on `buzz_core::relay::normalize_relay_url` itself:

> Connection code may retain the configured URL; this canonical form is for identity, receipts, status and deduplication.

### Related issue

Fixes #2444. Same root cause family as #3283 (loopback tenant split on the seed side). Closest existing PR: #2446 — same diagnosis and direction; it now has merge conflicts against `main` and its description notes the missing automated regression test as an honest gap. This PR is rebased on current `main`, adds that regression coverage via a small testable seam, and additionally pins the spawn-config hash to the canonical form so restart-drift detection cannot false-positive. Happy to close this in favor of a revived #2446 if maintainers prefer — the test seam here grafts onto it cleanly.

## Fix

Keep the canonical form for identity only (runtime-map keys, receipts, log paths, spawn-config hash). Dial the configured URL everywhere a child or probe actually connects:

- `child_connect_relay_url(configured, key)` — new seam in `runtime.rs` that returns the configured URL for the child's `BUZZ_RELAY_URL` and git-credential scope, with a `debug_assert` that the configured URL canonicalizes to the runtime key's relay (callers cannot hand it a URL from a different pair).
- The three spawn call sites (`start_pair`, launch restore, ensure-running) pass their already-resolved configured URL instead of `key.relay_url`.
- The reconcile fan-out passes the requested community URL to `start_pair` instead of the canonical echo.
- `probe_agent_relay_access` probes over the requested authority, so spawn eligibility is checked against the same community the child will join.
- `start_pair` fills `requested_relay_url` in its status rows, so the frontend's existing `requestedRelayUrl ?? relayUrl` fallback re-dials the configured URL on restart.
- The spawn-config hash deliberately stays on `runtime_key.relay_url`: `status_for` recomputes that hash from `key.relay_url`, so hashing the configured spelling would read as permanent drift and flag a spurious restart on every status pass.

### Testing

- Two new regression tests pin the seam: `child_connect_relay_url_preserves_configured_loopback_authority` (canonical key is `ws://127.0.0.1:3000` while the child still gets `ws://localhost:3000`) and `child_connect_relay_url_is_identity_for_non_loopback_relays`.
- `just ci` passes locally; `cargo test --lib` on the Tauri crate: 1851 passed, 0 failed; clippy clean.
- Manual end-to-end on macOS against a fresh local relay and a fresh desktop dev instance: before the fix, all three starter-team agents logged `discovered 0 channel(s)` and never replied (memberships verified present in the relay DB, all events under the `localhost:3000` community while harnesses connected to `127.0.0.1:3000`); after the fix, each harness logs `discovered 1 channel(s)` and a mention in the Welcome channel gets an agent reply within ~50s.

Follow-up candidates (out of scope here, discussed in #2444/#3283): collapsing loopback spellings into one community key on the relay side (#2619 explores this), and surfacing the agent-idle state in the UI instead of only in the harness log.

### DIFF
--- a/desktop/src-tauri/src/managed_agents/restore.rs
+++ b/desktop/src-tauri/src/managed_agents/restore.rs
@@ -330,10 +330,14 @@ pub async fn restore_managed_agents_on_launch(
                                             // mid-turn session is not resumed by an
                                             // eager child — and silently reintroduces
                                             // N idle brains on every launch.
+                                            // Configured spelling, not
+                                            // `key.relay_url` — Host-derived
+                                            // tenancy (see
+                                            // `child_connect_relay_url`).
                                             spawn_agent_child(
                                                 app,
                                                 record,
-                                                &key.relay_url,
+                                                &relay_url,
                                                 true,
                                                 owner_hex_ref,
                                             )
--- a/desktop/src-tauri/src/managed_agents/runtime.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime.rs
@@ -445,9 +445,35 @@ pub(crate) fn configure_runtime_cli(
     }
 }
 
+/// Relay URL for the spawned harness's `BUZZ_RELAY_URL` and git-credential
+/// scope: always the caller's *configured* URL, never the canonical
+/// runtime-key form. Relay tenancy is derived from the HTTP Host header, so
+/// the canonical loopback rewrite (`ws://localhost:3000` →
+/// `ws://127.0.0.1:3000`) names a different — empty — community than the one
+/// the desktop wrote the agent's channel memberships into; a harness handed
+/// the canonical form discovers 0 channels and never sees a mention. The
+/// canonical form stays correct for identity only: runtime-map keys,
+/// receipts, log paths, and the spawn-config hash (which `status_for`
+/// recomputes from `key.relay_url`).
+pub(crate) fn child_connect_relay_url<'a>(
+    configured: &'a str,
+    key: &ManagedAgentRuntimeKey,
+) -> &'a str {
+    debug_assert!(
+        buzz_core_pkg::relay::normalize_relay_url(configured)
+            .is_ok_and(|canonical| canonical == key.relay_url),
+        "configured relay URL must canonicalize to the runtime key's relay"
+    );
+    configured
+}
+
 /// Spawn an agent process without holding any locks on records or runtimes.
 /// Returns the child process and log path on success. The caller is responsible
 /// for updating `ManagedAgentRecord` fields and inserting into the runtimes map.
+///
+/// `relay_url`: the *configured* pair relay (workspace override / start
+/// request), not a canonical runtime-key round-trip — see
+/// `child_connect_relay_url`.
 ///
 /// `owner_hex`: the workspace owner's pubkey, used as a fallback for legacy
 /// records that have no NIP-OA `auth_tag`. See `build_respond_to_env`.
@@ -545,9 +571,11 @@ pub fn spawn_agent_child(
         .map(|p| p.display().to_string())
         .unwrap_or_else(|| effective_command.clone());
 
-    // The caller supplies the explicit canonical pair relay. This is the only
-    // relay this child may connect to, regardless of the record/workspace default.
-    let effective_relay_url = runtime_key.relay_url.clone();
+    // The caller supplies the explicit pair relay. This is the only relay this
+    // child may connect to, regardless of the record/workspace default — in the
+    // configured spelling, because the Host header decides which community the
+    // child lands in (see `child_connect_relay_url`).
+    let child_relay_url = child_connect_relay_url(relay_url, &runtime_key);
 
     // Augment PATH for DMG launches so child processes can find:
     //   - bundled CLI via ~/.local/bin symlink
@@ -578,7 +606,7 @@ pub fn spawn_agent_child(
     }
     command.env("RUST_LOG", child_rust_log_filter());
     command.env("BUZZ_PRIVATE_KEY", &record.private_key_nsec);
-    command.env("BUZZ_RELAY_URL", &effective_relay_url);
+    command.env("BUZZ_RELAY_URL", child_relay_url);
     command.env("BUZZ_ACP_LAZY_POOL", if lazy { "true" } else { "false" });
     command.env("BUZZ_ACP_AGENT_COMMAND", &resolved_agent_command);
     command.env("BUZZ_ACP_AGENT_ARGS", agent_args.join(","));
@@ -826,7 +854,7 @@ pub fn spawn_agent_child(
     //
     // NOSTR_PRIVATE_KEY mirrors BUZZ_PRIVATE_KEY — keep in sync.
     if let Some(cred_helper) = resolve_command("git-credential-nostr") {
-        let relay_http_url = crate::relay::relay_http_base_url(&effective_relay_url);
+        let relay_http_url = crate::relay::relay_http_base_url(child_relay_url);
 
         command.env("NOSTR_PRIVATE_KEY", &record.private_key_nsec);
         command.env("GIT_TERMINAL_PROMPT", "0");
@@ -914,13 +942,14 @@ pub fn spawn_agent_child(
 
     // Stamp the effective spawn config so the summary builder can flag
     // needs_restart when disk state drifts from what this process runs.
-    // `effective_relay_url` is already resolved, and resolution is idempotent,
-    // so it serves as the workspace-relay input here.
+    // Hash the *canonical* relay form: `status_for` recomputes this hash from
+    // `key.relay_url`, so hashing the configured spelling here would read as
+    // permanent drift and flag a spurious restart on every status pass.
     let spawn_config_hash = super::spawn_hash::spawn_config_hash(
         record,
         &personas,
         &teams,
-        &effective_relay_url,
+        &runtime_key.relay_url,
         &global,
     );
 
@@ -1003,7 +1032,9 @@ pub fn start_managed_agent_process(
     // Scalar PIDs are migration-only and never establish pair liveness.
     record.runtime_pid = None;
 
-    let mut process = spawn_agent_child(app, record, &key.relay_url, false, owner_hex)?;
+    // Configured spelling, not `key.relay_url` — Host-derived tenancy (see
+    // `child_connect_relay_url`).
+    let mut process = spawn_agent_child(app, record, &relay_url, false, owner_hex)?;
     let now = now_iso();
     let receipt = super::ManagedAgentRuntimeReceipt {
         key: key.clone(),
--- a/desktop/src-tauri/src/managed_agents/runtime/tests.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime/tests.rs
@@ -1314,3 +1314,31 @@ fn restart_eligible_false_when_orphan_has_no_drift() {
 fn restart_eligible_false_when_non_orphan_has_no_drift() {
     assert!(!super::restart_eligible(false, false, false));
 }
+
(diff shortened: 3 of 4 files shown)


jbold:
Independent verification on Linux — this PR fixes #2444 for me.

Prior reports were macOS (0.4.23 DMG) and a `main` repro; adding a Linux data point plus one symptom I didn't see described in #2444.

**Symptom was worse than "agents don't respond": onboarding never completes.**

On a fresh install the app hangs permanently on the final onboarding step, "meet your starter team". There is no error, no failed request, and no timeout — the step simply never resolves, because it waits on starter-team agents that are connecting to a different community. The state is also persistent: quitting and relaunching returns straight to the frozen step, so the app is unusable until localStorage is cleared manually. #2444 describes agents going silent in an established workspace; this is the same root cause blocking first-run entirely.

**Environment**

- Buzz: built from source, `main` @ 19d57b0d, `just relay` + `just desktop-standalone`
- Also reproduced with the released v0.5.3 `.deb` (sha256 `61b3471776c4…`)
- Host: Fedora Silverblue 43, Wayland, NVIDIA
- Build/run: Ubuntu 24.04 distrobox, WebKitGTK 2.52.3
- Relay: local, Postgres/Redis/MinIO via Compose

**Reproduction**

1. `just relay` with `RELAY_URL=ws://localhost:3000`
2. `just fresh=1 desktop-standalone`, join the community as `ws://localhost:3000`
3. Complete onboarding

Result: hangs at "meet your starter team" indefinitely.

**Evidence it is the loopback tenant split**

The dev database seeds a community row per loopback spelling, which makes the split directly observable. Before the patch, everything the UI wrote landed under `localhost:3000` and **no `bot`-role membership was ever created**:

```
      host      |       name       | role  | members
----------------+------------------+-------+---------
 localhost:3000 | Welcome          | owner |       1
 localhost:3000 | general          | owner |       1
 localhost:3000 | welcome-everyone | owner |       1
```

After applying this PR (applies cleanly to 19d57b0d) and joining as `ws://127.0.0.1:3000`:

```
      host      |       name       | role  | members
----------------+------------------+-------+---------
 127.0.0.1:3000 | Welcome          | owner |       1
 127.0.0.1:3000 | Welcome          | bot   |       3
 127.0.0.1:3000 | general          | owner |       1
 127.0.0.1:3000 | welcome-everyone | owner |       1
```

Onboarding completes and the starter team appears.

**Why this took a long time to diagnose**

Every check said the stack was healthy. The relay passed `/_readiness`, served a valid NIP-11 document, accepted a hand-signed NIP-42 `AUTH` over WebSocket (`["OK",<id>,true,""]`), and `buzz-cli` could list channels and post messages as a brand-new identity. The desktop app authenticated and polled `POST /query` → `200` continuously throughout. Nothing in the relay logs was above INFO at any `RUST_LOG` level.

The tenancy mismatch has no error path anywhere, so the only visible signal is a UI step that never finishes.

**One measurable symptom, if it helps others identify this**

While stuck, `WebKitWebProcess` sits at a sustained 100% of one core (405 jiffies / 4s). After this patch, on the same machine and the same WebKitGTK build, it drops to ~1%. So the frontend is busy-waiting rather than idle-waiting while the starter team never arrives — a quick `ps`/`top` check distinguishes this bug from a merely slow start.

Worth stating explicitly so nobody chases it as a rendering bug: it is **not** WebKitGTK. I initially suspected #3654 (NVIDIA + `WEBKIT_DISABLE_DMABUF_RENDERER`) and measured both workarounds from that issue — `WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` and `WEBKIT_SKIA_ENABLE_CPU_RENDERING=1` — with no effect (100% / 111% / 101%). The spin resolves via this Rust-side patch, which touches no rendering code.

**One adjacent note on relay URL entry**

For a plaintext local relay you must type the scheme explicitly. A scheme-less entry (`localhost:3000`) is normalized to `wss://` — intentional, and the code says so ("a scheme-less host is assumed to be a secure relay") — which then fails TLS against a plaintext port and surfaces as:

```
join policy request failed: error sending request for url (https://localhost:3000/api/join-policy)
```

`relayProbe.ts` maps an explicit `http://` to `ws://` correctly, so this only bites scheme-less input. Not a bug, but the error message never mentions that the scheme was defaulted, which cost me a while. Possibly worth a clearer message for loopback hosts.

edumagiceco:
Independent repro of this bug in local self-hosted dev, and corroboration for both halves of the split this PR makes.

**What I actually tested:** not this branch. I hit the bug independently, root-caused it to the same place, and ran a narrower local variant on `main` — `spawn_agent_child` reading the workspace relay via `relay_ws_url_with_override` instead of taking `runtime_key.relay_url`, plus the canonical form for `spawn_config_hash`. So the numbers below confirm the diagnosis and the fix shape, not this diff.

### Repro

Desktop on `ws://localhost:3000`, every harness spawned with `ws://127.0.0.1:3000`:

```
buzz-acp starting: relay=ws://127.0.0.1:3000 ... respond_to=owner-only
INFO  buzz_acp: discovered 0 channel(s)
WARN  buzz_acp: no channel subscriptions resolved — agent will sit idle
```

The relay had accumulated one community per host spelling — `localhost:3000` holding every channel, `127.0.0.1:3000` empty and the one the harness queried. Membership was never the problem: the current kind-39002 event for `#general` carried the agent pubkey in a `p` tag throughout.

After the change, same relay and agent: `relay=ws://localhost:3000`, `discovered 2 channel(s)`, no idle warning, and an `@agent` mention answered in 60s on one run and 105s on another.

### Both halves matter, and the second one is easy to miss

My first attempt moved only the dialed URL and left `spawn_config_hash` fed from the configured spelling. `status_for` recomputes it from `key.relay_url`, so the two disagree for every loopback host and the restart badge would have been stuck on forever. I did not catch that myself — #2944's note on it is what sent me back to check, and after switching the hash input to `runtime_key.relay_url` the badge stayed clear.

So the doc comment on `child_connect_relay_url` is describing a real failure mode on both sides, not a theoretical one. A minimal fix that only redirects the dialed URL trades the idle-agent bug for a permanent-needs-restart bug.

### On the shape

The `debug_assert!` that `configured` canonicalizes to `key.relay_url` is the part my variant had no equivalent for. Reading the workspace relay inside `spawn_agent_child` (what I did) silently accepts a caller/key mismatch; a named function with that invariant asserted catches it in dev instead. Same for the regression tests — `child_connect_relay_url_preserves_configured_loopback_authority` pins exactly the behavior I could only verify by hand.

Overlaps #2944, which fixes `probe_agent_relay_access` and `reconcile_managed_agent_runtimes` as well — worth deciding between them rather than landing both.

Environment: macOS 26.5.2 (arm64), local relay from `just dev`, Postgres/Redis/MinIO via the repo's compose stack.

Related issues: #3505, #4147.

salman1993:
🤖 Thank you for working on this. We reviewed the overlapping proposals and are consolidating the configured-vs-canonical relay URL fix in #4859 so review and implementation can happen in one place. We’re marking this PR as superseded and asking a maintainer or the author to close it. Please review #4859 and contribute feedback or code there—particularly the helper invariant and loopback regression tests developed here.

anilkishan:
salman asked us to consolidate on #4859. Your diagnosis matches what landed; we did not take the `child_connect_relay_url` seam because current `main` already restructured that area. Credit is in the #4859 prior-art table. Thank you - and thanks for the Linux / onboarding-hang notes on #2444.

PR #3498, opened 2026-07-29
fix(workflows): authorize trusted owner-only agent wakeups

JustinPerea:
## Summary

- add explicit workflow-owner attribution to relay-signed workflow messages
- authorize that attribution in `buzz-acp` only when the relay identity, event
  signature, kind, and exact workflow tags all verify
- apply the same fail-closed author resolution in normal and setup-listener
  modes, while keeping managed agents in `owner-only`

## Security model

Untrusted authors cannot self-assert an `actor` tag. `buzz-acp` accepts the
effective workflow owner only when:

1. NIP-11 identifies the event signer as the connected relay
2. the event signature verifies locally
3. the event is a kind 9 stream message
4. exactly one `["buzz:workflow", "true"]` tag is present
5. exactly one valid 64-hex `actor` tag is present

Missing metadata, malformed tags, duplicate tags, wrong kinds, forged signers,
and tampered events all fall back to the literal event signer and therefore
remain rejected by `owner-only`.

## Verification

- `just ci` passed
- `cargo test -p buzz-acp --lib`: 636 passed
- `cargo test -p buzz-relay --lib workflow_base_tags_include_actor_and_legacy_attribution`: passed
- staged secret scan: no findings

`just test` completed its unit phase, but its integration bootstrap could not
start because Docker Desktop was not running. A direct full
`cargo test -p buzz-relay --lib` run produced 762 passes and 10
database/mesh-environment failures, plus 33 ignored tests.

## Deployment

The relay and ACP harness must both be deployed before scheduled `owner-only`
agent wakeups begin working. Deployment order is safe because either partial
state remains fail-closed.

Buzz origin channel: `965eb307-5e3b-4f4e-a9c6-94323d98982d`

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -257,6 +257,60 @@ async fn author_allowed(
     }
 }
 
+/// Return the workflow owner represented by a relay-signed message.
+///
+/// Attribution is trusted only after verifying the NIP-11 relay signer, event
+/// signature, message kind, and exact single-value workflow/actor tags. Any
+/// ambiguity fails closed and falls back to the literal event signer.
+fn trusted_workflow_actor(
+    event: &nostr::Event,
+    trusted_relay_pubkey: Option<&str>,
+) -> Option<String> {
+    let trusted_relay_pubkey = trusted_relay_pubkey?;
+    if event.pubkey.to_hex() != trusted_relay_pubkey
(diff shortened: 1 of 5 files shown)


JustinPerea:
@tlongwell-block could you review this relay/ACP authorization fix? It keeps agents owner-only while accepting workflow ownership only after NIP-11 signer, event signature, kind, and exact actor/workflow tag verification. Full local `just ci` passed; Semgrep, zizmor, and DCO are green.

PR #3508, opened 2026-07-29
fix(desktop): render projects header and create menu at zero projects

### LINKED ISSUE #3504: Projects: create menu hidden when relay has zero projects — no way to create the first project
**Describe the bug**

When a relay has zero projects, the Projects view renders only the empty state — the page header, toolbar, and the create menu (`+`) are never mounted. Since "Repository" lives inside that create menu, there is no way to create the *first* project from the desktop UI. Chicken-and-egg: the button that creates a project only appears once a project already exists.

Cause looks like the early return in `desktop/src/features/projects/ui/ProjectsView.tsx`:

```tsx
if (projects.length === 0) {
  return <EmptyState />;
}
```

This returns before `projectsHeader` / `projectsNavigation` (which contains `<ProjectsCreateMenu />`) are rendered.

**Steps to reproduce**

1. Start a relay with no repo announcements (kind 30617) published.
2. Open the desktop app and go to Projects.
3. See "No projects yet — Projects published to this relay will appear here."
4. No header, no toolbar, no `+` button anywhere on the page.

**Expected behavior**

The Projects header and create menu should render regardless of project count, so a first project can be created from the UI. The empty state should sit in the content area below the toolbar (the way `EmptyFilteredState` already does), not replace the whole view.

**Workaround**

Publish the first repo announcement out-of-band, after which the full Projects UI appears:

```bash
export BUZZ_RELAY_URL="http://localhost:3000"
export BUZZ_PRIVATE_KEY="nsec1..."
buzz repos create --id my-project --name "My Project"
```

**Version and platform**

- Buzz version: v0.5.0
- OS: macOS

**Logs / additional context**

Same early return is present on `main` as of 485d03a. Suggested fix: drop the early return and render `<EmptyState />` inside the content area where `repositoryItems` goes, so the header and create menu stay mounted.

puneetv05:
Fixes #3504.

## Problem

`ProjectsView` returned `<EmptyState />` before constructing the page header, toolbar, and `<ProjectsCreateMenu />`. On a relay with zero projects, the UI therefore hid the only control capable of creating the first project.

## Fix

- Remove the early return.
- Render `<EmptyState />` in the normal content slot when `projects.length === 0`.
- Keep the Projects header, filter toolbar, and `+` create menu mounted for every project count.
- Add focused Playwright coverage for the create flow and for preserving the page chrome across filters.

## Scope

This is intentionally a three-file regression fix. It does not change empty-state copy, add another CTA, or modify project deletion/tombstone behavior. Broader related implementations are tracked in #4118 and #5205.

Unhiding the create menu also exposes its **Issue** and **Pull Request** items when there are no projects. Their existing dialogs already handle an empty project list by disabling submission until a project is available.

## Current-main adaptation

The branch is rebased onto current `main`. The E2E setup now:

- uses `FEATURE_OVERRIDES_STORAGE_KEY` rather than a versioned string literal;
- hides the seeded kind `30621` project announcement as well as the kind `30617` repository announcements;
- selects the current **Project** create-menu item name; and
- preserves the existing project smoke-test entries in `playwright.config.ts`.

## Verification

- Biome check on the three touched files
- `pnpm typecheck`
- `pnpm build:e2e`
- `pnpm exec playwright test --project=smoke tests/e2e/projects-empty-state.spec.ts` — **2 passed**

The two focused tests verify:

1. The header, empty state, toolbar, and create menu remain mounted with zero projects, and **Project** opens the existing create dialog.
2. The empty state replaces only the content area across Repositories, Pull Requests, and Issues filters.

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -107,6 +107,7 @@ export default defineConfig({
         "**/project-commit-detail.spec.ts",
         "**/project-inbox.spec.ts",
         "**/project-issue-comments.spec.ts",
+        "**/projects-empty-state.spec.ts",
         "**/project-pr-review.spec.ts",
         "**/persona-model-combobox-screenshots.spec.ts",
         "**/drafts-screenshots.spec.ts",
--- a/desktop/src/features/projects/ui/ProjectsView.tsx
+++ b/desktop/src/features/projects/ui/ProjectsView.tsx
@@ -592,9 +592,10 @@ export function ProjectsView() {
     );
   }
 
-  if (projects.length === 0) {
-    return <EmptyState />;
-  }
+  // Zero projects still renders the full chrome — the create menu lives in the
+  // toolbar, so returning early here would leave no way to create the first
+  // project from the UI.
+  const hasProjects = projects.length > 0;
 
   const projectItems =
     visibleProjects.length === 0 ? (
@@ -850,7 +851,9 @@ export function ProjectsView() {
           </div>
           <div className="mx-auto w-full max-w-6xl">
             <div className="w-full min-w-0 pb-4 pt-4">
-              {filter === "all" ? (
+              {!hasProjects ? (
+                <EmptyState />
+              ) : filter === "all" ? (
                 <ProjectsOverviewPanel
                   metadata={
                     <ProjectsOverviewRail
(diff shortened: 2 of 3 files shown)


puneetv05:
Hi @thomaspblock  , could you please review #3508 when you get a chance? It fixes #3504, where the create menu disappears when there are zero projects, preventing users from adding their first project in Buzz. All checks are passing. Thanks!

PR #2506, opened 2026-07-23
fix(desktop): give Windows agent children a usable PATH and start dev

al3rez:
Two separate things stop Buzz working on Windows. Both are small, but together they make the desktop app unusable there.

## Agents can't find node

Every managed agent dies the moment it starts. From `%APPDATA%\xyz.block.buzz.app\agents\logs\`:

```
'"node"' is not recognized as an internal or external command,
ERROR buzz_acp: agent initialize failed: Agent process exited unexpectedly
Error: all 24 agents failed to start — cannot continue
```

`build_augmented_path` assembles the child's PATH from Buzz's managed directories plus the user's login-shell PATH. On Windows the latter is always `None` — `fetch_login_shell_path_inner` returns early there, correctly, since Git Bash reports POSIX colon-delimited entries (`/c/Users/...`) that a native child can't use.

The comment says children "inherit the real Windows PATH instead." They don't. `runtime.rs` does `command.env("PATH", path)`, which replaces the inherited value rather than extending it, so the agent ends up with Buzz's managed directories and nothing else. No node, no npm, no git. The npm `.cmd` shims fail on their first line.

This appends the process PATH on Windows when no shell PATH was supplied. It's gated on `!parts.is_empty()` so callers that pass no local context still get `None` back, matching the existing rule about not manufacturing a PATH out of ambient state.

## `tauri dev` doesn't start

```
'exec' is not recognized as an internal or external command
The "beforeDevCommand" terminated with a non-zero status code.
```

`beforeDevCommand` is `exec ./node_modules/.bin/vite`. `exec` is a POSIX shell builtin, and Tauri runs the script through `cmd.exe` on Windows. Calling Vite through node works everywhere.

One caveat worth a review from someone on the macOS dev loop: `exec` was presumably there to avoid an intermediate shell process so signals reach Vite directly. This change gives that up on Unix. If that matters more than I think, the alternative is a platform-specific config file, which is uglier but keeps the current behaviour.

## Testing

Windows 11 (26100). Two new unit tests cover the PATH change: one asserts the process PATH is appended after the managed directories, the other that no PATH is manufactured without local context. `cargo test -p buzz-desktop --lib managed_agents::runtime::path` passes, and the existing Unix tests are untouched.

For the dev command, `pnpm exec tauri dev` now starts cleanly with no platform-specific config override:

```
Running BeforeDevCommand (`node ./node_modules/vite/bin/vite.js`)
VITE v8.0.16  ready in 984 ms
```

[...]

PR #3515, opened 2026-07-29
fix(desktop): run Vite directly in Tauri dev command

kiranmagic7:
## Summary

Runs Vite through its Node entry point in Tauri's default `beforeDevCommand`.

The current command relies on the POSIX `exec` shell builtin, which is unavailable under Windows `cmd.exe`. Direct Tauri development commands that read `desktop/src-tauri/tauri.conf.json` therefore fail before the app starts. Calling Vite's published Node entry point is portable without changing how Vite starts on macOS or Linux.

The `just` recipes that start Tauri provide a generated config from `scripts/instance-env.sh`; #3534 covers that path. `just desktop-dev` already calls `pnpm exec vite` directly and is not affected.

### Related issue

Extracted from the maintainer request in #2506. Complements #3534.

### Testing

`node ./node_modules/vite/bin/vite.js --version` returned `vite/8.0.16 darwin-arm64 node-v24.15.0`.

`pnpm --dir desktop check` passed.

`just ci` passed.

A Windows smoke test was not available locally. This changes configuration only, so screenshots do not apply.

### DIFF
--- a/desktop/src-tauri/tauri.conf.json
+++ b/desktop/src-tauri/tauri.conf.json
@@ -5,7 +5,7 @@
   "identifier": "xyz.block.buzz.app",
   "build": {
     "beforeDevCommand": {
-      "script": "exec ./node_modules/.bin/vite",
+      "script": "node ./node_modules/vite/bin/vite.js",
       "cwd": "..",
       "wait": false
     },


kiranmagic7:
Status check on head `bd36a46207df12b13f203d51d813efa6ac38314f`: the full CI run is green (18 successful checks, 7 path-filtered skips), with DCO Check, Semgrep OSS, and zizmor also green.

Review focus: the single-line bare Tauri dev command in `desktop/src-tauri/tauri.conf.json`. The related generated-config path remains separate in #3534, so this PR is intentionally scoped to the repository config path.

kiranmagic7:
#3534 now includes a Windows 11 smoke test of the same `node ./node_modules/vite/bin/vite.js` entry point and keeps its generated-config path separate from this bare `tauri.conf.json` path. That fills the platform-proof gap noted here. Current `main` still uses the POSIX `exec` command, while this one-line head is mergeable with full CI green.

kiranmagic7:
Current `main` at `b6a26556e` still has the POSIX-only `exec ./node_modules/.bin/vite` command in `tauri.conf.json`; #3534 remains limited to the generated-config path. I rechecked this one-line branch against that head: it merges cleanly, and its exact-head CI is green. The maintainer-requested Windows fix is still not covered elsewhere.

PR #3517, opened 2026-07-29
fix(chart): require Redis at any replica count, not just replicaCount>1

### LINKED ISSUE #3427: Relay readiness hard-requires Redis at replicaCount=1, but chart validation and docs say Redis is only needed for multiple replicas
## Symptom

Fresh Helm install (chart 0.1.6, image `sha-9547891`), `replicaCount: 1`, no `REDIS_URL` / `externalRedis.url` / bundled redis. The relay starts cleanly — all startup probes pass, including the S3 git object-store conformance probe — but the pod sits `Running 0/1` forever: `/_readiness` returns 503 and never recovers. Nothing in the logs states that Redis is the missing dependency; the last startup lines look healthy, so the operator is left correlating a silent 503.

Providing a `REDIS_URL` (tiny in-namespace redis) and restarting the pod immediately fixes readiness.

## Why this is surprising

Every chart-side signal says Redis is optional at a single replica:

- `templates/_validate.tpl`: *"Multiple replicas require Redis"* — the fail branch only triggers when `minimumReplicas > 1`.
- `values.yaml`: *"replicaCount > 1 hard-requires Redis for buzz-pubsub"* (implying =1 does not).
- `values.schema.json` `replicaCount` description says the same.

So a single-replica install validates, renders, deploys — and then never becomes ready.

## Ask

Either (a) the relay drops Redis from the `/_readiness` gate when no Redis is configured and single-replica pubsub can run in-process, or (b) if Redis is genuinely a hard runtime dependency at any replica count, chart validation should fail fast at install time (`redis.enabled` / `externalRedis.url` / `existingSecret` with `REDIS_URL` required unconditionally) and the docs updated. Either resolution beats the current silent-503 behavior; a log line naming the unreachable/missing Redis on readiness failure would help regardless.

Happy to PR whichever direction maintainers prefer.

Related: none found in open issues (searched redis/readiness).

1nderboi:
## Problem

Fixes #3427.

Fresh Helm install, `replicaCount: 1`, no Redis source configured. The chart's validation, `values.yaml`, `values.schema.json`, and README all state "Redis is only required for `replicaCount > 1`" — so the install renders and deploys cleanly. The pod then sits `Running 0/1` forever: `/_readiness` returns 503 and never recovers, with nothing in the logs stating Redis is the missing dependency.

That claim is wrong. The relay unconditionally wires Redis through `buzz-pubsub` (presence, typing, rate limiting, cache invalidation) and, more importantly, the NIP-98 HTTP-auth replay guard (`RedisNip98ReplayGuard` in `state.rs`), which **fails closed** with no Redis reachable. None of this is gated on replica count — a single-replica relay needs Redis exactly as much as a 3-replica one.

The actual runtime root cause: `deployment.yaml`'s `REDIS_URL` env was conditionally `optional: true` precisely when `replicaCount==1` and no Redis source was configured, so the pod started with no `REDIS_URL` at all and fell back to the relay's `redis://localhost:6379` default — nothing listening, permanent readiness failure, no explanation.

## Fix

- `deployment.yaml`: `REDIS_URL`'s `secretKeyRef` is now unconditionally required, matching `DATABASE_URL`'s existing pattern (no `optional:` flag).
- `_validate.tpl`: the Redis hard-fail guard now fires at any replica count. Moved it to the end of the guard chain (after Postgres/S3) so it doesn't change which error message an already-misconfigured install sees first — verified against every existing `validation_test.yaml` case.
- `values.yaml`, `values.schema.json`, `README.md`: corrected the "replicaCount > 1 requires Redis" framing everywhere it appeared.
- `router.rs`: added a `warn!` log line on `/_readiness` failure naming which dependency (postgres/redis) is unreachable — the JSON response body already reported this, but it wasn't visible from pod logs/events, which is what an operator actually sees first. Requested explicitly in the issue as a complementary fix regardless of which validation direction was taken.

## Test plan
- `helm unittest deploy/charts/buzz`: 9 suites / 44 tests pass. Renamed/duplicated the old `replicaCount>1` test into two (`replicaCount=1` and `>1`, both now failing on missing Redis), and added `externalRedis.url` to the `secrets_test.yaml`/`networking_test.yaml` fixtures that previously passed only because the old guard didn't fire at `replicaCount=1`.
- `helm template` against every `ci/*-values.yaml` and `tests/fixtures/*-values.yaml` fixture (all already configure Redis) — all render cleanly, matching the CI render-matrix step.
- Manually reproduced the exact bug and the fix:
  - `helm template` with `replicaCount=1` + no Redis source → now fails fast: `Error: ... Redis source missing: the relay requires Redis at any replica count...`
  - Same install with `externalRedis.url` set → renders, `REDIS_URL` is present as a required (non-optional) `secretKeyRef`.
- `cargo build -p buzz-relay --lib` and `cargo clippy -p buzz-relay --lib --all-targets -- -D warnings`: clean.
- `cargo fmt -p buzz-relay --check`: clean.

### DIFF
--- a/crates/buzz-relay/src/router.rs
+++ b/crates/buzz-relay/src/router.rs
@@ -389,6 +389,11 @@ async fn readiness_handler(State(state): State<Arc<AppState>>) -> impl IntoRespo
     if pg_ok && redis_ok {
         (StatusCode::OK, Json(json!({"status": "ready"}))).into_response()
     } else {
+        tracing::warn!(
+            postgres_ok = pg_ok,
+            redis_ok = redis_ok,
+            "readiness probe failing — see which dependency is unreachable"
+        );
         (
             StatusCode::SERVICE_UNAVAILABLE,
             Json(json!({"status": "not_ready", "postgres": pg_ok, "redis": redis_ok})),
--- a/deploy/charts/buzz/README.md
+++ b/deploy/charts/buzz/README.md
@@ -121,13 +121,13 @@ an Ingress or HTTPRoute for the pairing Service; route the public hostname to
 
 ## HA (production)
 
-`replicaCount > 1` hard-requires Redis:
+The relay requires Redis at **any** replica count, not just HA:
 
-- Redis (`redis.enabled=true`, `externalRedis.url`, or `REDIS_URL` in `existingSecret`) — for `buzz-pubsub` fan-out
+- Redis (`redis.enabled=true`, `externalRedis.url`, or `REDIS_URL` in `existingSecret`) — for `buzz-pubsub` (presence, typing, rate limiting, cache invalidation) and the NIP-98 HTTP-auth replay guard, which fails closed without it
 
 It does **not** require ReadWriteMany git storage. Git ref/object state is object-store-backed (each request hydrates an ephemeral repo from S3-compatible storage; writer serialization is the object-store pointer CAS — see `docs/git-on-object-storage.md`), and repo-name uniqueness lives in Postgres. Each replica can use its own `ReadWriteOnce` volume; no shared filesystem is needed.
 
-The chart **template-fails** if the Redis invariant is broken at `replicaCount > 1`. No silent degradation.
+The chart **template-fails** if no Redis source is configured, at any `replicaCount`. No silent degradation — a missing Redis source used to render successfully at `replicaCount=1` and then sit at a permanent, unexplained `/_readiness` 503.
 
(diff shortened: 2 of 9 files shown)


Chessing234 (commented):
Requiring Redis at any replica count (not only `>1`) is the right chart default for Buzz — single-replica deploys still use Redis for pubsub/leases in practice, and the previous gate silently stranded those. Small, focused helm fix.

PR #3524, opened 2026-07-29
fix(desktop): hide ghost caret after Shift+Enter

### LINKED ISSUE #3219: Double cursor (ghost caret) appears in chat composer after two Shift+Enter line breaks
**Describe the bug**
When I type a few words in the chat composer and press Shift+Enter twice to add two line breaks, two cursors show up at the same time: the real one blinking in the right spot (after the second line break), and a leftover "ghost" cursor stuck one line above, where the cursor was after the first line break. It's only visual, but it's distracting and makes it hard to tell where text will actually go.

**Steps to reproduce**
1. Open any channel and click into the chat composer
2. Type a few characters or words on the first line
3. Press Shift+Enter twice
4. See two cursors: the real one after the second line break, and a ghost one on the line above it

**Expected behavior**
Just one cursor, blinking where I'm actually typing (the line after the second line break).

**Version and platform**
- Buzz version: 0.4.26
- OS: macOS

**Logs / additional context**
Screenshot attached below. It happens when there's text on the first line followed by two Shift+Enters. 

Image

JoeJoeflyn:
## Summary
- Hide the trailing `<br>` ProseMirror inserts after a hard break, which WebKit/Chromium renders as a stale ghost caret

Closes #3219

## Testing
- Live-local: typed text, pressed Shift+Enter twice — only one caret blinks at the real cursor position

https://github.com/user-attachments/assets/1b1f4ec2-d38f-4982-8999-fa7354d4e066

### DIFF
--- a/desktop/src/shared/styles/globals/composer.css
+++ b/desktop/src/shared/styles/globals/composer.css
@@ -103,6 +103,10 @@
   margin: 0;
 }
 
+.rich-text-composer .tiptap br + .ProseMirror-trailingBreak {
+  display: none;
+}
+
 .rich-text-composer .tiptap p.is-editor-empty:first-child::before {
   content: attr(data-placeholder);
   float: left;


Chessing234 (commented):
Nice minimal CSS fix for the ghost caret. The `br + .ProseMirror-trailingBreak` selector matches the ProseMirror hard-break shape after Shift+Enter, and hiding only that trailing break shouldn't affect placeholder/`is-editor-empty` rendering. LGTM from the diff; the attached before/after video seals it.

Chessing234 (approved):
Approving the ghost-caret CSS fix.

JoeJoeflyn:
Thanks for the review!

PR #3528, opened 2026-07-29
Polish mobile typing indicator

klopez4212:
## Summary

- Present channel and thread typing status in a composer-matched container.
- Animate the strip so the message list moves smoothly as typing begins and ends.
- Increase typing-label contrast and avatar/padding for readability.

## Pixel 10 snapshot

Typing indicator above the composer

## Validation

- `flutter test test/features/channels/channel_detail_page_test.dart`
- `flutter analyze`


wesbillman (approved):
Much better!

PR #3529, opened 2026-07-29
Refine community invite limits

klopez4212:
## Summary

- Simplify the community invite dialog around link sharing.
- Add matching expiry and use-limit dropdowns, with sensible preset use caps.
- Cover the default unlimited and selected-limit invite payloads.

## Validation

- `pnpm -C desktop run build:e2e`
- `pnpm -C desktop exec playwright test tests/e2e/invite-link-copy.spec.ts tests/e2e/invites-settings-screenshots.spec.ts --project=smoke`

### DIFF
--- a/desktop/src/features/community-members/ui/CommunityInviteDialog.tsx
+++ b/desktop/src/features/community-members/ui/CommunityInviteDialog.tsx
@@ -3,6 +3,7 @@ import * as React from "react";
 import {
   Dialog,
   DialogContent,
+  DialogDescription,
   DialogHeader,
   DialogTitle,
 } from "@/shared/ui/dialog";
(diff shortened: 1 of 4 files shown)


klopez4212:
🤖 Updated invite dialog snapshots.

### Community invites

Community invite settings

### Invite dialog

Invite dialog

klopez4212:
🤖 @codex review

wesbillman:
One product/security behavior is worth confirming before merge: this changes the default invite from **3 uses** to **No limit**. It also replaces the previous custom 1–10,000 input with presets topping out at 25.

I traced the implementation and the unlimited behavior is technically correct and explicitly tested: `maxUses === null` omits `max_uses`, which the relay interprets as unlimited. So there is no serialization bug here—but the new default materially relaxes the existing access/resource cap.

Is “unlimited by default, preset caps only through 25” the intended product policy? If yes, I found no implementation blocker. If not, I recommend keeping 3 uses as the default while retaining the new preset UI.

wesbillman (approved):

PR #3533, opened 2026-07-29
fix(desktop): disable backdrop-blur on Linux for perf

### LINKED ISSUE #3328: Slow UI performance on linux
**Describe the bug**
The UI is very sluggish on linux. Things such as the automatic highlighting style change after hovering a message can take a half second.

**Steps to reproduce**
1. Open the app
2. Get an agent to start working on something
3. Scroll
4. Notice it takes significant time before scrolling happens 
**Expected behavior**
What did you expect to happen?

**Version and platform**
- Buzz version: v0.4.26
- OS: Linux Mint

**Logs / additional context**
Seems to be affected by whether or not an agent is running (worse when agent running)

JoeJoeflyn:
## Summary
Linux WebKitGTK has no GPU-accelerated `backdrop-filter`. With 16+ backdrop-blur layers visible (header, composer, action bars, thread panel), each frame re-blurs all backdrops in software, blocking the main thread — causing sluggish UI and slow loading. Tag `<html>` with `platform-linux` at boot and neutralize `backdrop-filter` on Linux, falling back to the opaque `bg-*` colors every blur element already carries.

### Related issue
Fixes #3328

### Testing
Verified on Arch Linux + Hyprland: app was sluggish with 30 messages in #general. After reload with the fix, UI is responsive. macOS/Windows unaffected (no `platform-linux` class added).

screenrecording-2026-07-29_20-42-51

### DIFF
--- a/desktop/src/main.tsx
+++ b/desktop/src/main.tsx
@@ -14,6 +14,7 @@ import { PoofBurstProvider } from "@/shared/ui/PoofBurstProvider";
 import { Toaster } from "@/shared/ui/sonner";
 import { TooltipProvider } from "@/shared/ui/tooltip";
 import { recoverLocalStorageQuotaOnStartup } from "@/shared/lib/localStorageQuota";
+import { isLinuxPlatform } from "@/shared/lib/platform";
 
 type E2eWindow = Window & {
   __BUZZ_E2E__?: unknown;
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
Reasonable Linux perf trade. WebKitGTK really does software-composite `backdrop-filter`, and tagging `html.platform-linux` at boot is a low-risk switch. Two notes:

1. `navigator.platform` is deprecated and can be spoofed; `navigator.userAgentData?.platform` (or Tauri's OS API) would be more durable long-term, but fine for v1.
2. The `* { backdrop-filter: none !important }` hammer is broad — confirm glass panels that relied on blur still have an opaque `bg-*` behind them so they don't go translucent. The PR description says they do; if you've spot-checked the thread panel + composer on Arch, I'm good.

JoeJoeflyn:
Thanks for the review! Reused isLinuxPlatform() for #1. For #2, all glass panels have opaque fallbacks, checked on Arch, looks fine.

Chessing234 (commented):
rechecked tip — still the right linux trade. most chrome already carries opaque bg-*, and ComposerDockBackdrop is frost-only so killing blur just makes it a no-op there. fine by me.

JoeJoeflyn:
tks, appreciate it

PR #2496, opened 2026-07-23
fix(desktop): Windows sidecar stubs, managed Node, onboarding scroll

Chessing234:
## Summary
- Add `.exe` to Windows sidecar stub paths in the Justfile (`#2492`)
- Ship managed Node.js zip install for Windows x64/arm64 (`#2491`)
- Make machine onboarding step frames scroll so the docked Next CTA stays reachable (`#2414`)

## Test plan
- [ ] `just desktop` / CI desktop job on a Windows runner creates `*-windows-*.exe` sidecar stubs
- [ ] On Windows, Install harness downloads Node and `node --version` matches the managed runtime
- [ ] Machine onboarding step 4: shrink the window; content scrolls and Next stays usable

Fixes #2492
Fixes #2491
Fixes #2414

Made with [Cursor](https://cursor.com)

PR #3540, opened 2026-07-29
fix(dev): add .exe to Windows sidecar stub paths

### LINKED ISSUE #2492: Windows: `_ensure-sidecar-stubs` Justfile recipe omits the `.exe` suffix Tauri requires for externalBin
### Description

The `_ensure-sidecar-stubs` recipe in the root `Justfile` creates sidecar
placeholder files as:

```bash
TARGET=$(rustc -vV | sed -n 's|host: ||p')
for bin in buzz-acp buzz-agent buzz-dev-mcp git-credential-nostr buzz; do
    touch "desktop/src-tauri/binaries/${bin}-${TARGET}"
done
```

On Windows the host triple is `x86_64-pc-windows-msvc`, and Tauri's
`externalBin` validation expects sidecars named
`binaries/<bin>-x86_64-pc-windows-msvc.exe`. The stubs are created without the
`.exe` suffix, so Tauri rejects them and every recipe that depends on
`_ensure-sidecar-stubs` (`dev`, `desktop-tauri-check`, `desktop-tauri-clippy`,
`desktop-tauri-test`, `desktop-standalone`, `staging`) fails on Windows even
though the code itself builds fine.

### Suggested fix

Append `.exe` when the host triple contains `windows`:

```bash
EXT=""
case "$TARGET" in *windows*) EXT=".exe";; esac
for bin in buzz-acp buzz-agent buzz-dev-mcp git-credential-nostr buzz; do
    touch "desktop/src-tauri/binaries/${bin}-${TARGET}${EXT}"
done
```

(For actual bundling, empty `touch`-ed stubs are not runnable, of course — for
release builds the real `target/…/<bin>.exe` outputs get copied into place,
which is what `desktop-release-build` does on other platforms. This issue is
only about the stub naming that gates the compile/check recipes.)

### Environment

- Windows 11 Pro 25H2 (build 26200), Git Bash available for Justfile shebang recipes
- Buzz @ v0.4.23, `just` 1.57

sumit-m:
## Summary

Appends `.exe` to the sidecar stub filenames on Windows so `_ensure-sidecar-stubs` produces the names Tauri's `externalBin` validation expects.

On Windows the host triple is `x86_64-pc-windows-msvc` and Tauri looks for `binaries/<bin>-x86_64-pc-windows-msvc.exe`. The recipe creates the stubs without the suffix, so Tauri rejects them and every recipe depending on `_ensure-sidecar-stubs` — `dev`, `desktop`, `desktop-standalone`, `desktop-staging` — fails before the app starts.

Three lines, and a no-op on every other platform: `EXT` stays empty unless the triple contains `windows`.

### Related issue

Fixes #2492.

Two open PRs touch the same recipe, and I have not seen a maintainer signal which should land first:

- **#2496** contains this same `.exe` fix, bundled with a managed-Node install and an onboarding scroll fix. It is currently `CONFLICTING` and has been unreviewed since 2026-07-27. This PR isolates just the `.exe` change so it can be reviewed on its own; if #2496 is preferred, close this one.
- **#2597** rewrites the same loop to stage real binaries over the stubs (a different bug — `os error 13`). It keeps the extensionless name, so it does not fix Windows and would conflict textually with this change. Whichever merges first, the other is a small rebase — happy to do that on this branch.

### Testing

```
$ rm -rf desktop/src-tauri/binaries && just _ensure-sidecar-stubs
$ ls desktop/src-tauri/binaries/
buzz-acp-x86_64-pc-windows-msvc.exe
buzz-agent-x86_64-pc-windows-msvc.exe
buzz-dev-mcp-x86_64-pc-windows-msvc.exe
buzz-x86_64-pc-windows-msvc.exe
git-credential-nostr-x86_64-pc-windows-msvc.exe
```

Before this change the same command produced extensionless files and `just dev` aborted with `binaries\buzz-acp-x86_64-pc-windows-msvc.exe doesn't exist`.

Not verified on macOS or Linux — I have no such hardware. The guard is `[[ "$TARGET" == *windows* ]]`, so `EXT` is empty and the emitted paths are byte-identical to today's on any non-Windows triple.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -156,11 +156,15 @@ _ensure-sidecar-stubs:
     TARGET=$(rustc -vV | sed -n 's|host: ||p')
     mkdir -p desktop/src-tauri/binaries
     SIDECARS=(buzz-acp buzz-agent buzz-dev-mcp git-credential-nostr buzz)
-    if [[ "$TARGET" != *windows* ]]; then
+    EXT=""
+    if [[ "$TARGET" == *windows* ]]; then
+        # Tauri's externalBin validation expects binaries/<bin>-<triple>.exe.
+        EXT=".exe"
+    else
         SIDECARS+=(buzz-backend-kubernetes)
     fi
     for bin in "${SIDECARS[@]}"; do
-        touch "desktop/src-tauri/binaries/${bin}-${TARGET}"
+        touch "desktop/src-tauri/binaries/${bin}-${TARGET}${EXT}"
     done
 
 # Ensure Docker dev services (Postgres, Redis, etc.) are running and healthy


Chessing234 (commented):
Small and correct — Tauri's Windows `externalBin` lookup really does want the `.exe` suffix on the host triple, and leaving `EXT` empty elsewhere keeps this a no-op on unix. One thing worth a follow-up (not blocking): `_ensure-sidecar-stubs` still `touch`es empty stubs, so a first `just desktop` on a clean Windows clone will get past validation and then fail later when the real binaries aren't built. Out of scope for this PR, but related to the same Windows stub path.

Chessing234 (approved):
Approving the sidecar  stub fix — minimal, correct, and unblocks every Windows  recipe.

Chessing234 (approved):
Approving the Windows sidecar stub extension fix — minimal and unblocks desktop recipes.

PR #3545, opened 2026-07-29
fix(dev): reset desktop state on Windows

sumit-m:
## Summary

Adds Windows support to the two desktop state-reset scripts.

`scripts/reset-desktop-dev-state.sh` had no Windows arm, so it left the app data behind. `scripts/reset-desktop-standalone-state.sh` was worse — it hit an explicit `exit 1`, so resetting standalone state was impossible there.

Both now handle `MINGW*|MSYS*|CYGWIN*`:

- **App data** — removes the bundle directory under both `%APPDATA%` and `%LOCALAPPDATA%`, falling back to `$HOME/AppData/...` when the variables are unset.
- **Credentials** — Windows Credential Manager is the keyring backend, and it names each entry `secrets.<service>`. `cmdkey /list` is filtered to `secrets.(buzz|sprout)-desktop-dev`, with an optional trailing `.<slug>` so per-worktree entries are cleaned up too. The required `-dev` segment keeps production `buzz-desktop` credentials out of scope.

Also adds contract tests for both scripts and wires them into CI, since neither had any coverage.

The existing `BUZZ_TEST_PLATFORM` override is honoured, so the platform branches can be exercised from any host — that is how the tests reach the Windows arm on a Linux runner.

### Related issue

No duplicate found: no open PR touches `scripts/reset-desktop-*.sh`. Checked by intersecting changed-file paths across all open PRs, not just titles.

### Testing

- `scripts/reset-desktop-dev-state.sh` removes `%APPDATA%\xyz.block.buzz.app.dev` and the matching `secrets.buzz-desktop-dev` credential; the app returns to first-run onboarding afterwards.
- Production `secrets.buzz-desktop` is left untouched — confirmed via `cmdkey /list` before and after.
- `scripts/reset-desktop-standalone-state.sh` completes instead of exiting 1.
- Both new contract tests pass locally.

The `cmdkey` path is a no-op where the binary is absent (`command -v` guard), so non-Windows behaviour is unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,10 @@ jobs:
           scripts/test-mobile-release-candidate-publisher.sh
       - name: Mobile worktree identity contract
         run: scripts/test-mobile-worktree-overrides.sh
+      - name: Desktop dev-state reset contract
+        run: |
+          scripts/test-reset-desktop-dev-state.sh
+          scripts/test-reset-desktop-standalone-state.sh
       - name: File size ratchet unit tests
         run: node --test scripts/check-file-sizes-core.test.mjs
 
--- a/scripts/reset-desktop-dev-state.sh
+++ b/scripts/reset-desktop-dev-state.sh
@@ -32,7 +32,20 @@ remove_bundle_state() {
   shopt -u nullglob
 }
 
-case "$(uname -s)" in
+# Windows Credential Manager names each keyring entry `secrets.<service>`.
+# The trailing pattern keeps per-worktree `buzz-desktop-dev.<slug>` variants in
+# scope while the required `-dev` segment keeps production `buzz-desktop` out.
+remove_dev_credentials() {
+  command -v cmdkey >/dev/null 2>&1 || return 0
+  local target
+  while IFS= read -r target; do
+    log "Removing credential $target"
+    cmdkey "/delete:${target}" >/dev/null 2>&1 || true
+  done < <(cmdkey /list | tr -d '\r' |
+    grep -oE 'secrets\.(buzz|sprout)-desktop-dev(\.[A-Za-z0-9-]+)?$' | sort -u)
+}
+
+case "${BUZZ_TEST_PLATFORM:-$(uname -s)}" in
   Darwin)
     remove_bundle_state "$HOME/Library/Application Support"
     remove_bundle_state "$HOME/Library/Caches"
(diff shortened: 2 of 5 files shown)


Chessing234 (commented):
Welcome Windows coverage for the reset scripts. Handling both `%APPDATA%` and `%LOCALAPPDATA%` (plus the `$HOME/AppData` fallback) matches where Buzz actually lands state, and naming Credential Manager entries the way the keyring backend does is the detail that usually gets missed. The CI wiring + script tests look solid. One nit: confirm the MINGW/MSYS/CYGWIN detection also covers a plain PowerShell invocation if operators run the `.sh` via Git Bash only — if PowerShell isn’t in scope, say so in the PR body.

sumit-m:
Thanks — taking the PowerShell question head on, because the answer splits in two.

**PowerShell as the shell you type the command into: covered, and nothing in the script depends on it.** `uname -s` reports the environment bash is *running in*, not the parent process that launched it. `just reset` → `scripts/dev-reset.sh --yes` → `reset-desktop-dev-state.sh`, and every one of those is `#!/usr/bin/env bash`, so on Windows they execute under Git Bash regardless of whether the invocation came from a PowerShell prompt, `cmd`, or a Git Bash shell. `uname -s` is `MINGW64_NT-10.0-26100` in all three cases and the `MINGW*|MSYS*|CYGWIN*` branch fires. Nothing reads the parent shell, so there is no detection gap here.

**PowerShell as the execution environment: out of scope, and deliberately so.** These scripts use bash-only constructs — `[[`, `shopt -s nullglob`, and process substitution in `remove_dev_credentials`. `pwsh ./reset-desktop-dev-state.sh` cannot run them, and no amount of platform detection would change that; supporting PowerShell natively would mean a parallel `.ps1`, not a different `case` arm. There is no `.ps1` anywhere in this repo today and every reset entry point is a `just` recipe with a bash shebang, so a Git Bash (or equivalent) runtime is already a hard prerequisite for the whole recipe surface, not something this PR introduces.

Happy to state that prerequisite explicitly in a comment at the top of the script if you'd like it recorded rather than implied.

**One real gap worth naming, since it isn't PowerShell:** under WSL, `uname -s` is `Linux`, so the Linux branch runs and clears XDG-path state inside the WSL filesystem. That is correct for a WSL-hosted checkout, and it deliberately does not reach across to a Windows-side install (different filesystem, and Credential Manager cleanup is not attempted). Someone running this under WSL expecting their Windows dev state to be wiped would be surprised. Say the word and I'll add a line to that effect.

On coverage: `test-reset-desktop-dev-state.sh` drives the Windows branch on the existing Ubuntu runner via `BUZZ_TEST_PLATFORM=MINGW64_NT-10.0-26100` with a mocked `cmdkey` whose output carries real CRLF line endings, and asserts both that the three `-dev` credentials are deleted and that production `secrets.buzz-desktop` is not. So the Windows path is covered without adding a Windows runner.

PR #3549, opened 2026-07-29
feat(desktop): add a clear action to the harness log toolbar

sumit-m:
## Summary

Adds a clear action to the harness log toolbar, next to the existing copy button.

Restarting an agent to test a change means re-reading a log that still holds the previous run's output. Today the only way to get a clean view is to find the file on disk and empty it by hand.

The command truncates in place rather than deleting:

```rust
OpenOptions::new().write(true).open(path)?.set_len(0)
```

A running harness holds the file open in append mode, so unlinking it would leave the writer appending to an inode the viewer can no longer see — the log would look permanently frozen. `set_len(0)` keeps writer and reader on the same file, and the next append lands at offset 0.

`clear_managed_agent_log` mirrors the guards already on `get_managed_agent_log`: same store lock, same "not found" error, and the same refusal for non-`Local` backends, since there is no local file to truncate for a remote agent. A missing log is treated as success — clearing something already absent is not an error.

The button is gated on the same `shouldShowLogDetails` condition as the copy button, so it only appears on the diagnostics and logs views.

### Related issue

None found. Checked by intersecting changed-file paths across all 567 open PRs: no open work adds a clear, truncate, or reset action for agent logs. The only PR touching `UserProfilePanelHeaderContent.tsx` is #2544, which restores thread context and does not touch the toolbar actions.

Note `desktop/src-tauri/src/lib.rs` here is a one-line command registration; 51 open PRs touch that file, so expect a trivial conflict there if this sits for a while.

### Testing

Three unit tests in `managed_agents::storage`:

- `truncate_log_file_empties_an_existing_log`
- `truncate_log_file_tolerates_a_missing_log`
- `appends_after_truncate_land_at_the_start` — the one that matters, pinning that a writer holding the file open keeps writing to the same inode

`cargo test --lib truncate` passes (11 tests in scope). Manually verified on Windows 11: with an agent running, clearing the log empties the panel and subsequent harness output continues to stream into the same view.

UI change, so a before/after of the toolbar:

- before: `[copy]`
- after: `[clear] [copy]`

Happy to attach a screenshot if useful — the change is a single icon button, so I did not want to pad the PR with one unprompted.

### DIFF
--- a/desktop/src-tauri/src/commands/agent_logs.rs
+++ b/desktop/src-tauri/src/commands/agent_logs.rs
@@ -3,8 +3,8 @@ use tauri::{AppHandle, Manager};
 use crate::{
     app_state::AppState,
     managed_agents::{
-        latest_managed_agent_log_path, load_managed_agents, read_log_tail, BackendKind,
-        ManagedAgentLogResponse,
+        latest_managed_agent_log_path, load_managed_agents, read_log_tail, truncate_log_file,
+        BackendKind, ManagedAgentLogResponse,
     },
 };
 
@@ -34,6 +34,31 @@ pub async fn get_managed_agent_log(
             content: read_log_tail(&log_path, line_count.unwrap_or(120) as usize)?,
             log_path: log_path.display().to_string(),
         })
+    })
+    .await
+    .map_err(|e| format!("spawn_blocking failed: {e}"))?
+}
+
+/// Truncate the log `get_managed_agent_log` reads, so a running harness keeps
+/// appending to the same file rather than the viewer and the writer diverging.
+#[tauri::command]
+pub async fn clear_managed_agent_log(pubkey: String, app: AppHandle) -> Result<(), String> {
+    tokio::task::spawn_blocking(move || {
+        let state = app.state::<AppState>();
+        let _store_guard = state
+            .managed_agents_store_lock
+            .lock()
+            .map_err(|error| error.to_string())?;
+        let records = load_managed_agents(&app)?;
+        let record = records
+            .iter()
+            .find(|record| record.pubkey == pubkey)
+            .ok_or_else(|| format!("agent {pubkey} not found"))?;
+        if record.backend != BackendKind::Local {
+            return Err("logs are not available for remote agents".to_string());
+        }
+
+        truncate_log_file(&latest_managed_agent_log_path(&app, &pubkey)?)
     })
     .await
     .map_err(|e| format!("spawn_blocking failed: {e}"))?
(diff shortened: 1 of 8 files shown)


Chessing234 (commented):
Clearing via `set_len(0)` instead of unlink is the right call while the harness still holds the file in append mode — otherwise the UI would watch a deleted inode and look frozen. Toolbar placement next to copy is natural. Worth confirming the frontend re-reads / tails from offset 0 after clear so a running agent’s next write shows up immediately (and that a missing log file returns a clean error rather than creating an empty stub somewhere unexpected).

sumit-m:
Both confirmed — details below.

**Re-reading after clear: there is no offset state to reset.** `read_log_tail` opens the file fresh on every call and seeks `SeekFrom::End(0)`, then walks backward in 8 KiB chunks until it has enough newlines. Nothing is cached between calls on either side of the boundary — the frontend holds only the returned string in its react-query cache — so the "viewer stuck at a stale offset" failure mode can't arise. That is also why `set_len(0)` is safe rather than merely convenient: the harness holds an append-mode handle from `open_log_file`, so its next write lands at the new end instead of leaving a NUL-filled hole at the old offset, and the very next tail read picks it up.

The clear does force an immediate re-read rather than waiting for a poll: `useClearManagedAgentLogMutation`'s `onSuccess` invalidates `["managed-agent-log", pubkey]`. Worth noting explicitly since it looks like a key mismatch — the query key is `["managed-agent-log", pubkey, lineCount]`, and react-query's invalidation is a prefix match, so the two-element key does match every `lineCount` variant. Without it the viewer would sit on `staleTime: 3_000` / `refetchInterval: 30_000`.

**No stub is ever created.** `truncate_log_file` opens with `OpenOptions::new().write(true)` and deliberately **no** `.create(true)`, so the call has no way to bring a file into existence.

One place I went the other way from your suggestion, so flagging it rather than burying it: a missing log returns `Ok(())`, not an error. An absent log is already in the state clearing is meant to produce, and `latest_managed_agent_log_path` falls back to the canonical `{pubkey}.log` name when no log exists yet — so a user who clicks Clear before the agent has ever started would get an error with no action attached to it. `read_log_tail` already uses the same convention (`!path.exists()` → `Ok("")`). Genuine errors still surface: anything other than `NotFound` becomes `failed to open log file <path>: <error>`, and the frontend shows `Failed to clear log: …` as a toast. Happy to flip it to an error if you'd rather the button be loud about a log that isn't there.

Fully accurate on "somewhere unexpected": `latest_managed_agent_log_path` → `managed_agents_logs_dir` does `create_dir_all` on the `logs/` **directory**. That is pre-existing behaviour shared with the read path in `get_managed_agent_log`, it only ever creates the directory logs are supposed to live in, and it creates no file. Not introduced here, but you asked, so it shouldn't go unmentioned.

For completeness, the command mirrors `get_managed_agent_log`'s guards: it takes `managed_agents_store_lock`, resolves the record by pubkey, and rejects non-`Local` backends with `logs are not available for remote agents`.

Issue #3494, opened 2026-07-29
Linux AppImage: bundled libsoup-3.0/libnghttp2 shadow system libs on non-Debian distros, breaking system GStreamer WebRTC plugins


scottsanpedro:
#2176 (make Linux AppImage GStreamer work on non-Debian distros) removed the bundled
GStreamer plugins so the system ones load — thank you, 0.5.0 now launches natively on
Arch. But two bundled libraries remain that undo that fix for the WebRTC half:

- The AppImage bundles `usr/lib/libsoup-3.0.so.0` (Ubuntu's build). On Arch, the system
  `libgstnice.so` / `libgstwebrtc.so` link against system `libgupnp-1.6`, which is built
  against libsoup 3.6.x. With the bundled (older) libsoup first on the loader path, the
  plugin scanner fails both plugins:

      Failed to load plugin '/usr/lib/gstreamer-1.0/libgstnice.so':
      /usr/lib/libgupnp-1.6.so.0: undefined symbol: soup_server_message_pause
      (identical for libgstwebrtc.so)

  → no `webrtcbin` → huddles have no media pipeline.

- Renaming the bundled libsoup exposes the second one: system libsoup 3.6.6 needs a
  newer libnghttp2 than the bundled `usr/lib/libnghttp2.so.14`:

      symbol lookup error: /usr/lib/libsoup-3.0.so.0:
      undefined symbol: nghttp2_option_set_no_rfc9113_leading_and_trailing_ws_validation

**Workaround (verified):** rename both bundled libs so the system copies load:

    cd squashfs-root/usr/lib
    mv libsoup-3.0.so.0   libsoup-3.0.so.0.bak
    mv libnghttp2.so.14   libnghttp2.so.14.bak

After this (and clearing `~/.cache/gstreamer-1.0/registry.x86_64.bin` — the scanner
caches blacklisted plugins), a full rescan shows libgstwebrtc/libgstnice loading
cleanly; the only remaining failures are optional codecs (openexr, spandsp) with no
huddle impact. App runs stably with system libsoup (soname-compatible).

**Suggested fix:** drop `libsoup-3.0.so.0` and `libnghttp2.so.14` from the AppImage
(same rationale as removing bundled libsystemd in #2353 and bundled GStreamer in
#2176) — the bundled WebKit is happy with the system copies, soname is stable across
both.

**Environment:** Buzz Desktop 0.5.0 AppImage (extracted, AppRun), Arch Linux
[...]

PR #3557, opened 2026-07-29
huddle: linux: grant WebKit user media requests

### LINKED ISSUE #3495: Linux: huddles fail with NotAllowedError — WebKitGTK user-media permission request appears unhandled
With the WebRTC GStreamer plugins loading correctly (see https://github.com/block/buzz/issues/3494)
and a working audio input device present, starting a huddle on Linux fails with:

    The request is not allowed by the user agent or the platform in the current
    context, possibly because the user denied permission.

No permission prompt is ever shown.

On macOS/Windows the OS provides the mic-permission UI, but on Linux WebKitGTK the
**embedding application** must answer `WebKitUserMediaPermissionRequest` (via the
`permission-request` signal / wry's permission hook) — an unanswered request is a
denial. I can't find a handler for this in the desktop shell, which would make
`getUserMedia` structurally un-grantable on Linux regardless of hardware.

**Diagnostic ladder ruled out** (Arch, native 0.5.0):
1. GStreamer webrtc/nice plugins load (after the bundled-lib workaround above);
2. an audio source is present and enumerable (verified with a PipeWire source;
   with *no* source present the error is a constraint error instead, as expected);
3. the failure is then NotAllowedError with no prompt → permission layer.

**Suggested fix:** handle `WebKitUserMediaPermissionRequest` in the Linux WebView
(allow for the app's own origin, or surface an in-app prompt), and ideally also
`webkit_settings_set_enable_media_stream(TRUE)` if not already set.

Happy to test a build — this is the last blocker for voice on native Linux as far
as I can tell.

**Environment:** as https://github.com/block/buzz/issues/3494.

brauliobo:
## Summary

- enable WebKitGTK media-stream support when the Linux main webview is ready
- grant user-media permission requests only for Buzz's bundled UI and configured Vite development origin
- leave every other permission type and navigated origin to WebKit's default handling
- cover the trusted-origin boundary with focused unit tests

## Root cause

WebKitGTK delegates `getUserMedia` permission decisions to the embedding application. Buzz did not handle `WebKitUserMediaPermissionRequest`, so WebKit denied microphone access without showing a prompt and Linux Huddles failed with `NotAllowedError`.

## Impact

Linux users can acquire their microphone for Huddles while navigated or untrusted content remains unable to inherit user-media permission.

Closes #3495.

## Validation

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml webkit_permissions`
- production-profile runtime test on Arch Linux, KDE Wayland, and WebKitGTK: Huddle microphone capture became active without the previous permission rejection

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -1085,6 +1085,7 @@ dependencies = [
  "url",
  "user-idle",
  "uuid",
+ "webkit2gtk",
  "window-vibrancy",
  "windows-sys 0.61.2",
  "zeroize",
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -40,6 +40,9 @@ keyring = { version = "3.6.3", default-features = false, features = ["sync-secre
 # connection is dropped, which the plugin does immediately. Default features
 # keep the pure-Rust zbus backend, matching the plugin (no libdbus needed).
 notify-rust = "4"
+# Tauri exposes the underlying WebKitGTK view, but applications still need to
+# answer WebKit's user-media permission signal themselves on Linux.
+webkit2gtk = "2"
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2-app-kit = { version = "0.3.2", default-features = false, features = ["NSHapticFeedback"] }
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -30,6 +30,8 @@ mod shutdown;
 mod templates;
 mod util;
 #[cfg(target_os = "linux")]
+mod webkit_permissions;
+#[cfg(target_os = "linux")]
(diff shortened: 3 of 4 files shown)


Chessing234 (commented):
This is the missing piece for #3495-style Linux huddle failures. Handling `WebKitUserMediaPermissionRequest` only for the bundled UI / Vite origin is the correct trust boundary, and leaving other permission types alone avoids over-granting. Unit coverage on the trusted-origin check is welcome. One note for operators: this grants the permission request, but mic capture can still fail if PipeWire/GStreamer plugins are shadowed (see #3494) — worth cross-linking in the PR body so the two layers stay distinct.

wpfleger96:
🤖 hey @brauliobo, thanks for this — your trusted-origin allowlist and the unit tests proving that boundary are the security core of what shipped: #3607 (just merged, fixes #3495) adopted the same origin check (`tauri://localhost` + dev origin) your PR pioneered, combined with a device-kind check. Closing this as superseded by the merged fix.

PR #3560, opened 2026-07-29
feat(buzz-agent): add MiniMax as a first-class provider

octo-patch:
Reason: The buzz-agent runtime shipped dedicated Anthropic, OpenAI-compatible, and Databricks providers but had no MiniMax provider.

Adds MiniMax as a first-class `BUZZ_AGENT_PROVIDER` in the agent runtime. MiniMax speaks the OpenAI-compatible Chat Completions dialect, so this reuses the existing request path rather than introducing a new body/parse pair — matching the crate's documented "adding a provider" pattern.

## Changes

- `crates/buzz-agent/src/config.rs`
  - New `Provider::MiniMax` enum variant.
  - `resolve_provider` accepts the region-selecting spellings `minimax` (global) and `minimax-cn` (mainland China); the key is validated in `from_env`, mirroring the Databricks path.
  - `from_env` wires `MINIMAX_API_KEY` (bearer), a `MINIMAX_MODEL` default of `MiniMax-M3`, and a region-derived base URL — `https://api.minimax.io/v1` for `minimax`, `https://api.minimaxi.com/v1` for `minimax-cn` — overridable via `MINIMAX_BASE_URL`. `OpenAiApi::Chat` is pinned so the `auto` host heuristic never routes to `/responses`.
- `crates/buzz-agent/src/llm.rs`
  - Routes `Provider::MiniMax` through the OpenAI-compatible Chat Completions request path and the static bearer token source.
- `crates/buzz-agent/README.md`
  - Documents the provider, its environment variables, and both regional endpoints. Notes that MiniMax's Anthropic-compatible Messages endpoint remains reachable today via `BUZZ_AGENT_PROVIDER=anthropic` + `ANTHROPIC_BASE_URL`.

## Tests

- `resolve_provider_accepts_minimax_region_spellings` — every accepted spelling resolves to `Provider::MiniMax` (case-insensitive).
- `minimax_default_base_url_selects_region_by_spelling` — global vs mainland-China base-URL selection.
- `minimax_routes_over_openai_chat_completions` — a MiniMax config POSTs to `/v1/chat/completions` with the configured model and parses the response.

## Checks run

- `cargo fmt -p buzz-agent -- --check` — clean.
- `cargo clippy -p buzz-agent --tests` — clean.
- `cargo test -p buzz-agent` — 302 library tests plus the 3 new tests pass. One pre-existing, unrelated multi-threaded steering test (`steer_folds_into_active_turn_without_cancelling`) is flaky on this machine and passes on re-run; it does not touch the provider or config code paths.


Chessing234 (commented):
Clean provider add following the documented OpenAI-compatible pattern. Reusing the existing request/parse path for MiniMax keeps the surface area small. README + enum/config wiring look right. One question: does MiniMax need any auth-header or base-URL default distinct from the generic OpenAI-compatible provider, or is `BUZZ_AGENT_BASE_URL` + key enough for first ship?

PR #3571, opened 2026-07-29
Improve emoji autocomplete matching

klopez4212:
## Summary
- Show all colon emoji autocomplete matches
- Rank exact and prefix shortcodes before weaker matches
- Add a regression test and screenshot

## Validation
- `pnpm test`
- `pnpm build`
- `pnpm exec playwright test --project=smoke tests/e2e/custom-emoji.spec.ts --grep "exact standard shortcode"`
- `just desktop-tauri-clippy`

Native Tauri tests were attempted but could not link because the local disk filled during compilation.

### DIFF
--- a/desktop/src/features/messages/lib/useEmojiAutocomplete.ts
+++ b/desktop/src/features/messages/lib/useEmojiAutocomplete.ts
@@ -4,7 +4,11 @@ import { init, SearchIndex } from "emoji-mart";
 import data from "@emoji-mart/data";
 
 import type { CustomEmoji } from "@/shared/lib/remarkCustomEmoji";
-import { fuzzyStandardEmoji, rankByShortcode } from "@/shared/lib/emojiSearch";
+import {
+  fuzzyStandardEmoji,
+  rankByShortcode,
+  rankShortcodeMatchesFirst,
+} from "@/shared/lib/emojiSearch";
 import { rewriteRelayUrl } from "@/shared/lib/mediaUrl";
 import type { AutocompleteEdit } from "./useRichTextEditor";
 
@@ -18,7 +22,7 @@ export type EmojiSuggestion = {
 
 const EMOJI_DEBOUNCE_MS = 120;
 const MIN_QUERY_LENGTH = 2;
-const MAX_RESULTS = 8;
+const UNLIMITED_RESULTS = Number.POSITIVE_INFINITY;
 
 init({ data });
 
@@ -81,7 +85,7 @@ export function useEmojiAutocomplete(customEmoji: CustomEmoji[] = []) {
       emojiQuery,
       customEmojiRef.current,
       (e) => e.shortcode,
-      MAX_RESULTS,
+      UNLIMITED_RESULTS,
     ).map((e) => ({
       id: e.shortcode,
       name: e.shortcode,
@@ -89,7 +93,10 @@ export function useEmojiAutocomplete(customEmoji: CustomEmoji[] = []) {
       url: rewriteRelayUrl(e.url),
     }));
 
-    SearchIndex.search(emojiQuery)
+    SearchIndex.search(emojiQuery, {
+      caller: "useEmojiAutocomplete",
+      maxResults: UNLIMITED_RESULTS,
+    })
       .then(
         (
           results: Array<{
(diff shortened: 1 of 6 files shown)


klopez4212:
🤖 ### Emoji autocomplete ordering

Typing `:joy` puts the exact `:joy:` shortcode first, ahead of the weaker custom `:bufo_joy:` match.

joy-exact-before-custom-substring

desktop/src/shared/lib/emojiSearch.ts
```diff
@@ -114,6 +114,30 @@ export function rankByShortcode<T>(
...
+): T[] {
+  const shortcodeMatches = rankByShortcode(
+    query,
+    items,
+    shortcodeOf,
+    Number.POSITIVE_INFINITY,
+  );
+  const matchedItems = new Set(shortcodeMatches);
+  return [
+    ...shortcodeMatches,
```
wesbillman: Carl here, commenting on Wes’s behalf.

**P2 — Preserve semantic results ahead of loose shortcode matches.** This helper promotes every shortcode tier—including substring and subsequence matches—ahead of emoji-mart’s name/keyword results. For example, `sad` is a subsequence of `sandwich`, so `:sandwich:` can become the default selection ahead of relevant sad-face results. Please promote only exact/prefix shortcode matches, keep weaker substring/subsequence matches behind semantic results, and add a regression test for a semantic query such as `sad`.

desktop/src/features/messages/lib/useEmojiAutocomplete.ts
```diff
@@ -106,23 +113,26 @@ export function useEmojiAutocomplete(customEmoji: CustomEmoji[] = []) {
...
-          // Custom emoji first (community-specific), then standard, then fuzzy.
-          const merged = [...customMatches, ...standard, ...fuzzy].slice(
-            0,
-            MAX_RESULTS,
+          // Rank exact/prefix shortcode matches across custom and standard emoji
+          // before weaker matches (for example, `joy` before `bufo_joy`).
+          setSuggestions(
+            rankShortcodeMatchesFirst(
+              emojiQuery,
+              [...customMatches, ...standard, ...fuzzy],
```
wesbillman: Carl here, commenting on Wes’s behalf.

**P2 — Cap or virtualize the rendered result list.** The candidate searches and final suggestions array are now unbounded, while `EmojiAutocomplete` eagerly maps every suggestion to a DOM button; `max-h-48` constrains only the visible height. Routine two-character queries can produce hundreds of standard matches, and large custom sets also trigger image elements/requests for every match. Please keep a broad pool for ranking if needed, but cap the final rendered list or virtualize it so typing does not create hundreds of rows at once.

Chessing234 (commented):
Ranking exact/prefix shortcodes ahead of weaker substring matches is the right autocomplete UX, and showing the full match set beats the previous silent truncation. Regression test + screenshot help. One watch: with a large custom-emoji catalog, rendering every match could get heavy — if the popover is virtualized or capped elsewhere, fine; otherwise consider a soft cap with a “N more” affordance later.

desktop/src/features/messages/lib/useEmojiAutocomplete.ts
```diff
@@ -81,15 +85,18 @@ export function useEmojiAutocomplete(customEmoji: CustomEmoji[] = []) {
...
       id: e.shortcode,
       name: e.shortcode,
       native: "",
       url: rewriteRelayUrl(e.url),
     }));
 
-    SearchIndex.search(emojiQuery)
+    SearchIndex.search(emojiQuery, {
+      caller: "useEmojiAutocomplete",
+      maxResults: UNLIMITED_RESULTS,
```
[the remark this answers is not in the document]
klopez4212: 🤖 Verified against the installed emoji-mart 5.6.0 implementation: `caller` is passed only to `init` as a diagnostic label; `SearchIndex.search` never reads `caller.searchValue`. Its type declarations also require `caller` whenever options are supplied, so the existing string is valid and does not suppress keyword results.

desktop/src/shared/lib/emojiSearch.ts
```diff
@@ -114,6 +114,30 @@ export function rankByShortcode<T>(
...
+    ...items.filter((item) => !matchedItems.has(item)),
```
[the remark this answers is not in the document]
klopez4212: 🤖 Fixed in 246d933cb. Only exact and prefix shortcodes are promoted; emoji-mart name and keyword results now remain ahead of substring and subsequence fallback matches. Added a `sad` regression test.

desktop/src/shared/lib/emojiSearch.ts
klopez4212: 🤖 Fixed in 246d933cb. Exact and prefix shortcodes are promoted, while semantic results stay ahead of loose substring and subsequence matches. Added coverage for `:sad` not selecting `:sandwich:` first.

desktop/src/features/messages/lib/useEmojiAutocomplete.ts
klopez4212: 🤖 Fixed in 246d933cb. The menu now virtualizes its rows, so every match remains reachable while only the visible range is rendered. Keyboard selection scrolls the active result into view.

klopez4212:
🤖 Addressed in 246d933cb: the autocomplete now virtualizes the result rows, so all matches remain reachable without rendering the entire catalog at once. Keyboard selection scrolls the active result into view.

wesbillman (approved):

PR #3585, opened 2026-07-29
fix(desktop): reconcile thread arrivals at bottom

wesbillman:
## Summary
- reconcile stale native-scroll anchors when a reply arrives at the physical floor
- clear the thread new-message affordance instead of incrementing it from stale cached state
- preserve the existing mid-history path and add direct lifecycle regression coverage

## Why
PR #3411 fixed geometry-driven reconciliation, but the reply-arrival branch still trusted a cached `message` anchor without checking the rendered position. Native anchoring could return a short thread to the floor without another scroll/resize callback, then the next reply incremented the pill anyway.

## Verification
- Desktop checks passed
- Desktop typecheck passed
- focused lifecycle test passed (6/6)
- push hook full Desktop unit suite passed (3,770/3,770)
- `git diff --check` passed

### DIFF
--- a/desktop/src/features/messages/ui/useAnchoredScroll.lifecycle.test.mjs
+++ b/desktop/src/features/messages/ui/useAnchoredScroll.lifecycle.test.mjs
@@ -247,13 +247,20 @@ function Harness({ channelId, onTargetSettled, refs }) {
   return null;
 }
 
-function BottomStateHarness({ messages, onState, refs }) {
+function BottomStateHarness({
+  messages,
+  onState,
+  refs,
+  targetMessageId = null,
+}) {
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
Solid follow-up to #3411. Checking `isAtBottomNow(container)` on append when the cached message anchor is stale — then forcing `at-bottom` and clearing the pill — matches the native-anchoring short-thread failure mode. The lifecycle test (scroll away → native snap back to floor → append) covers the regression without relying on resize/scroll callbacks. LGTM.

Chessing234 (approved):
Approving the thread floor reconciliation follow-up.

Issue #3527, opened 2026-07-29
Relay never seeds git repos on kind:30617 announce — repos stay 404 "repository not found" indefinitely


jackjackbits:
## Summary

On the hosted relay `permissionless.communities.buzz.xyz` (NIP-11 reports `software: https://github.com/block/buzz`, `version: 0.2.0`), announcing a repo via kind:30617 is accepted and stored, but the git smart-HTTP endpoint never becomes available. `GET /git/<owner-hex>/<repo-id>/info/refs?service=git-upload-pack` (and `git-receive-pack`) returns `404 repository not found` indefinitely, so the repo can never be pushed to or cloned.

Per current source, announce is the only provisioning path: `handle_git_repo_announcement` in `crates/buzz-relay/src/handlers/side_effects.rs` reserves the name in Postgres and seeds the empty-manifest pointer that `info_refs` requires (`crates/buzz-relay/src/api/git/transport.rs` — `Ok(None)` → 404). The git router exposes only `info/refs`, `git-upload-pack`, `git-receive-pack`, so there is no other way to create a repo.

## Reproduction (all requests NIP-98/Nostr-authenticated; unauthenticated requests correctly 401)

1. Publish kind:30617 with `d`=`<repo-id>` and a clone URL `https://permissionless.communities.buzz.xyz/git/<owner-hex>/<repo-id>`. Event is accepted (`{"accepted":true}`) and retrievable.
2. `GET /git/<owner-hex>/<repo-id>/info/refs?service=git-upload-pack` with valid signed auth → `404 repository not found`, still true 2+ hours later.
3. Same-owner **re-announce** — which per the code comments "*ensures* the pointer rather than assuming it" — is accepted but the endpoint still 404s. This rules out a one-time race; the seed side effect is not taking effect at all.
4. Reproduced with three separate identities/repos on this relay, including the community owner's key.

## Why it is silent

`ingest.rs` treats side effects as best-effort: `if let Err(e) = handle_side_effects(...) { warn!(..., "Side effect failed: {e}") }` — the announcement is stored even when seeding fails, and the underlying error (object-store put, missing `git_repo_names` migration, etc.) exists only in relay server logs. Clients cannot distinguish "seeded" from "seed silently failed".

Relay operators: please check server logs for `Side effect failed` on kind 30617 on this deployment. If the deployed build predates the seeding feature, a redeploy may be all that is needed.

## Suggested improvements

[...]

PR #3594, opened 2026-07-29
fix(cli): require --channel on repos create for buzz-channel tag

### LINKED ISSUE #3539: Bug: `buzz repos create` omits the `buzz-channel` tag that the git read gate requires — every CLI-created repo 404s forever
**Component:** `buzz` CLI (`repos create`) + relay git read gate
**Severity:** High — the CLI cannot produce a working repository; the failure is silent and misdiagnoses as "repo doesn't exist"
**Relay source referenced:** commit `3a4bf51`

## Summary

`buzz repos create` announces a NIP-34 repository (kind:30617) but provides **no way to set the `buzz-channel` tag**. The relay's git read gate (`authorize_git_read`) **requires** that tag to resolve the repo's bound channel, and denies any request whose announcement lacks it. The denial is `HTTP 404 "repository not found"` — byte-identical to a repository that was never announced. Result: **every repository created through the CLI is unreachable over git, permanently, with an error that points away from the cause.**

## Steps to reproduce

```bash
buzz repos create --id my-repo --name "My Repo" \
  --clone "https://<relay>/git/<owner-hex>/my-repo"

# Announcement is accepted and renders as a Project in Desktop.
git clone https://<relay>/git/<owner-hex>/my-repo
#   remote: repository not found
#   fatal: repository '.../my-repo' not found
```

`ls-remote`, `clone`, `fetch`, and `push` all return `404 repository not found`, whether or not the caller is a member of any channel — because the announcement carries no channel binding for the gate to check against.

## Root cause

1. **CLI:** `repos create` exposes `--id --name --description --clone --web --nostr-relay` and **no `--channel`**. The signed kind:30617 therefore never contains a `buzz-channel` tag. (`buzz.exe` contains zero occurrences of the string `buzz-channel`.)

2. **Relay read gate** (`crates/buzz-relay/src/handlers/transport.rs`, ~line 381): resolving the bound channel is mandatory, and its absence is denied with the same response as a missing repo:

   ```rust
   fn denied() -> Response {
       (StatusCode::NOT_FOUND, "repository not found").into_response()
   }
   // ...
   let Some(channel_id) = repo_bound_channel_id(&repo_event.event) else {
       warn!(repo = %repo_name, "git read gate: missing/malformed buzz-channel binding (deny)");
       return Err(denied());
   };
   ```

Three distinct conditions collapse into one indistinguishable 404 — (a) no announcement, (b) announcement present but no `buzz-channel` binding, (c) caller has no role in the bound channel — which is good for not leaking repo existence, but means the CLI's omission is undiagnosable from the client side.

## Fix

Add a required (or strongly-encouraged) `--channel <UUID>` flag to `buzz repos create` that emits a `buzz-channel` tag on the kind:30617. Given that an unbound repo can never be served, consider making it **required** and rejecting creation without it, rather than announcing a dead repo.

Secondary, optional: have the read gate distinguish "announced but unbound" from "not found" in a way that doesn't leak existence to unauthorized callers — e.g. a distinct log/metric, or a 409 to the *owner* only. Not required for the fix; the flag is the fix.

## Workaround (in use on this nest)

Publish the announcement with a small signer that adds the `buzz-channel` tag and completes the NIP-42 AUTH challenge (`tools/announce-repo.mjs` in the `buzz-nest-activation` repo). Verified end-to-end: with the tag added and nothing else changed, `ls-remote` goes 404 → 200 and `push`/`clone` succeed.

## Note on a related dead end

Reports of this symptom are easy to misattribute to "the relay has no git storage for this community" — because probing three repos under three keys all return 404. That inference is wrong: the gate checks the **caller's** role in each repo's bound channel, so a non-member gets 404 for a perfectly healthy repo. Git hosting over HTTPS with NIP-98 auth works on this deployment (proven by push + fresh clone once the binding is present). No relay-admin provisioning is involved.

Chessing234:
## Summary
- Require `--channel <UUID>` on `buzz repos create` and emit a `buzz-channel` tag on the kind:30617 announcement
- Without that tag the relay git read gate returns an opaque 404 for every clone/fetch/push (#3539)
- Thread the optional channel through `build_repo_announcement` and update the nest skill example

## Test plan
- [x] `cargo test -p buzz-sdk --lib builders::tests::repo_announcement`
- [ ] `buzz repos create --id demo --channel <uuid> --clone https://<relay>/git/<pubkey>/demo` then `git ls-remote` succeeds
- [ ] Omitting `--channel` fails clap validation

Fixes #3539


tlongwell-block (dismissed):
Reviewed as part of the #3527 investigation (root cause is the same bug — see the linked discussion; #3539 and #3527 should be cross-linked and one closed as a duplicate).

**Verified locally at head 6bbe384:**
- `build_repo_announcement` has exactly one production caller (`cmd_create_repo`), so the signature change is contained — checked with `rg` across `crates/`.
- `cargo test -p buzz-sdk --lib`: 241 passed. `cargo test -p buzz-cli`: green.
- The SDK parameter staying `Option<&str>` while the CLI flag is required matches the design consensus from the team thread.

**Two non-blocking notes:**
1. `validate_uuid` is parse-only — a well-formed UUID for a nonexistent channel (or one the announcer is not a member of) still produces an unreadable repo. Acceptable for this PR; a follow-up adds owner remediation messaging at the read gate and, later, ingest-time binding validation (tracked in the #3527 workstream).
2. The two unchecked live-test boxes in the test plan will be covered by our live regression matrix against a local Docker stack before this merges, alongside the follow-up PR.

The follow-up PR (`buzz repos bind` for existing unbound repos + read-gate remediation) composes with this change and does not touch these files. LGTM.

Chessing234:
rebased onto current main (was conflicting). --channel is now required on top of the optional path that landed meanwhile; nest skill + clap help updated. create_with_channel / malformed uuid tests still green.

Chessing234:
rebase onto main is done and checks should be green again — still looking for a merge when convenient (approved, mergeable).

Chessing234:
still approved + mergeable + dco green — happy to rebase again if main moves.

Chessing234:
still approved + mergeable whenever you have a free merge.

Chessing234:

PR #3597, opened 2026-07-29
fix(media): allow HTML file uploads as downloadable attachments

### LINKED ISSUE #3453: Allow text/html file uploads on the relay generic file path (served as download)
## Problem

Uploading an `.html` file is rejected by the relay with `DisallowedContentType("text/html")`.

`text/html` is on the relay's deny list for the generic file-upload path:

```rust
// crates/buzz-media/src/validation.rs
const BLOCKED_FILE_MIME_TYPES: &[&str] = &[
    // Active web content — stored-XSS vectors.
    "text/html",
    "application/xhtml+xml",
    "image/svg+xml",
    ...
];
```

## Why this matters

HTML files are a common, legitimate attachment in day-to-day work — especially in agent-heavy communities:

- Agents frequently produce self-contained HTML reports, dashboards, diagrams, and visualizations as their work output, and currently have no way to share them in a channel.
- Exported docs (notebooks, coverage reports, saved pages) are HTML too.

Right now the only workarounds are zipping the file or hosting it elsewhere, both of which add friction for what is just a document hand-off.

## Why it seems safe to allow

The doc comment on `BLOCKED_FILE_MIME_TYPES` itself notes that generic files are already served with:

- `Content-Disposition: attachment`
- `X-Content-Type-Options: nosniff`
- `CSP: default-src 'none'`

…which "already neutralises them — this allowlist-of-denials is defence in depth". So the download path never renders the HTML in the app's origin; the deny list is only guarding against a future header regression.

## Suggested fix

Remove `text/html` (and optionally `application/xhtml+xml`) from `BLOCKED_FILE_MIME_TYPES` so HTML files take the normal generic-file path and are served strictly as downloads.

If dropping the defence-in-depth layer entirely feels too risky, alternatives:

1. Serve blocked-but-textual types with a rewritten `Content-Type: text/plain` (still `attachment` + `nosniff`), so even a header regression can't execute them.
2. Keep the deny list but add a regression test asserting the attachment/nosniff/CSP headers on the file-serving path, making the header contract explicit instead of blocking the format.

Executables and SVG (which can be referenced inline via ``) can stay blocked — this request is only about HTML documents as downloadable attachments.

## Related

- #3083 tracks aligning the `buzz-cli` allowlist with the relay's generic file path; if this lands, the CLI list would need `text/html` as well so agents can actually upload these files.

Chessing234:
## Summary
- Remove `text/html` and `application/xhtml+xml` from the relay and desktop generic-file deny lists
- Keep SVG, JavaScript, and native executables blocked
- Map sniffed HTML to a `.html` / `.xhtml` extension and assert `serve_inline` stays false (forced download)

Generic files are already served with `Content-Disposition: attachment`, `X-Content-Type-Options: nosniff`, and `CSP: default-src 'none'`, so HTML documents are safe as downloadable attachments (#3453).

## Test plan
- [x] `cargo test -p buzz-media --lib test_validate_file_html`
- [ ] Upload an `.html` report in Desktop — succeeds and downloads rather than rendering inline
- [ ] SVG / `.exe` uploads still rejected

Fixes #3453

### DIFF
--- a/crates/buzz-media/src/validation.rs
+++ b/crates/buzz-media/src/validation.rs
@@ -62,20 +62,18 @@ pub(crate) fn looks_like_mp4_iso_bmff(bytes: &[u8]) -> bool {
 
 /// MIME types blocked from the generic file-upload path.
 ///
-/// These are the formats a browser (or the desktop webview) will *execute* or
-/// *render as active content* if it ever reaches them with the wrong response
-/// headers. We serve generic files with `Content-Disposition: attachment` +
-/// `X-Content-Type-Options: nosniff` + `CSP: default-src 'none'`, which already
-/// neutralises them — this allowlist-of-denials is defence in depth, so a future
-/// header regression can't turn an uploaded blob into a stored-XSS vector.
+/// Generic files are already served with `Content-Disposition: attachment` +
+/// `X-Content-Type-Options: nosniff` + `CSP: default-src 'none'`. This deny
+/// list is defence in depth against a future header regression.
 ///
-/// HTML, JS, and SVG are the classic stored-XSS carriers. Native executables are
-/// blocked because there's no legitimate reason to host them inline in chat and
-/// they're a malware-distribution risk.
+/// HTML / XHTML are intentionally **not** blocked: agents commonly produce
+/// self-contained HTML reports, and the download headers already prevent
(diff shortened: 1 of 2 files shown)


Chessing234:

Chessing234:
closing — main already allows html as inert download attachments

PR #3599, opened 2026-07-29
fix(desktop): confirm before re-adding mentioned non-member agents

### LINKED ISSUE #3577: Mentioning a removed member silently re-adds them to the channel
Removing a member from a channel does not stick: mentioning them re-adds them.

## Repro

1. Remove an agent from a channel (context menu → Remove from channel). System message: *"removed X from the channel"*.
2. In the same channel, type `@X` and send.
3. System message: *"X was added by You"* — and the agent starts working on the message.

## Cause

`useMentionSendFlow.ts` (~L847) resolves mentioned non-members and adds them before sending:

```ts
if (relayAgentPubkeys.length > 0) {
  const result = await addMembersMutation.mutateAsync({
    channelId: …, pubkeys: relayAgentPubkeys, role: "bot",
  });
}
```

## Why it matters

The removal is the only lever a channel owner has to keep an agent out — for cost, for noise, or because the agent was found to have more filesystem access than expected. A single mention silently undoes it, and the agent is not just re-added but immediately prompted. Autocomplete still offers removed agents, so this is easy to trigger by accident.

## Expected

Either exclude previously-removed members from the mention-autocomplete and the auto-add path, or make the confirmation state plainly that the mention will **re-add** them to the channel — the current flow reads as "send anyway", not as a membership change.

Buzz `0.4.26`, self-hosted relay.

Chessing234:
## Summary
- Stop silently re-attaching managed agents that are not channel members when they are `@mentioned`
- Route those mentions through the existing non-member invite dialog (copy clarifies re-add)
- `ensureManagedAgentMentionsReady` only starts agents already in the channel

## Why
After Remove from channel, typing `@Agent` was undoing the removal and immediately prompting the agent (#3577). The channel owner’s only lever to keep an agent out was being silently overridden.

## Test plan
- [ ] Remove an agent from a channel, `@mention` them — invite dialog appears; choosing Do nothing does not re-add
- [ ] Choose Invite — agent is re-added and the message sends
- [ ] Mention an agent already in the channel — no dialog; send works as before

Fixes #3577


Chessing234:

PR #3603, opened 2026-07-29
feat(models): manifest-driven model capabilities and curated labels

wpfleger96:
## Summary

This PR establishes a single hand-curated **model-capability manifest** (`scripts/model-capabilities.json`) as the source of truth for model metadata — display labels, supported reasoning-effort sets, default efforts, thinking modes, and Databricks v2 wire routes. A generator emits equivalent Rust and TypeScript interpreters from it, production resolution in both languages runs on the generated interpreters, and the superseded hand-maintained tables are deleted. It also ships the curated Databricks label registry and the frontend `resolveModelLabel()` contract.

Closes #3586. Landed as three sub-PRs into this branch: #3821 (manifest + generator), #3958 (production integration), #4589 (hand-table retirement).

## Where the capability data comes from

`scripts/model-capabilities.json` is **hand-curated in this repo** — nothing is fetched at build time. Every fact in it is reconciled against a cited external authority, and the citations live in the manifest itself (the top-level `_sources` block plus a per-record `_reconciliation_doc` field with retrieval date):

| Source | What it authorizes |
|---|---|
| [models.dev API](https://models.dev/api.json) (payload retrieved 2026-07-31, SHA-256 pinned in the manifest) | Databricks endpoint reasoning-effort sets and verbatim display labels |
| [Anthropic extended-thinking](https://platform.claude.com/docs/en/build-with-claude/extended-thinking) and [effort](https://platform.claude.com/docs/en/build-with-claude/effort) docs (July 2025) | Anthropic thinking modes and effort semantics |
| [OpenAI reasoning guide](https://platform.openai.com/docs/guides/reasoning) (July 2025) | OpenAI reasoning-effort semantics |
| goose `DATABRICKS_V2_KNOWN_MODELS` (revision `6789d4af`) | The Databricks v2 known-model ID list (opt-in drift check: `--check-goose`) |

When a provider ships a new model or effort level, the workflow is: edit the manifest record with a fresh citation, regenerate, commit. The CI gates make artifact drift impossible; keeping the manifest current with provider reality is deliberately a curation step — auditable, cited, and reviewed like any other change.

## Capability manifest and generated interpreters

`scripts/generate-model-capabilities.mjs` reads the manifest and emits two committed artifacts from one source of truth:

| Artifact | Consumer |
|---|---|
| `crates/buzz-agent/src/generated_model_capabilities.rs` | Rust interpreter |
| `desktop/src/features/agents/ui/modelCapabilities.ts` | TS interpreter |

The generator validates every interpolated manifest string through a shared `requireSafeString()` boundary (rejects quotes, backslashes, control chars), enforces the manifest schema (42 negative tests in `scripts/test-manifest-validator.mjs`, covering effort canonicality, post-inheritance record consistency, integer match priorities, and normalized duplicate keys), and supports `--check` for byte-clean verification.

## Production integration

**Rust (`buzz-agent`):** reasoning-effort normalization for OpenAI and legacy Databricks flows through `normalize_effort_for_provider()`, backed by the generated interpreter; `anthropic_body()` is provider-aware and Adaptive thinking clamps against the generated `supported_efforts`; Databricks v2 request routing reads the manifest's `databricks_v2_wire_route` instead of a hand-maintained segment classifier. The superseded effort/thinking hand tables and model-family classifiers in `config.rs` and `llm.rs` are deleted — the generated interpreter is the only authority.

**Desktop:** `getProviderEffortConfig()` resolves through the generated `modelCapabilities.ts`; `canonicalizeProvider()` (trim/lowercase plus the `databricks-v2` and `openai-compat` aliases) is applied before every provider-scoped lookup, and the provider is threaded through all label/effort surfaces so a supplied provider can never fall through to unscoped registry lookups. The old hand-table effort path and its fixture are deleted.

## Drift protection

The `model-capabilities` job in `.github/workflows/ci.yml` gates every PR:

- regenerates both artifacts and fails on any byte difference, so the committed interpreters cannot drift from the manifest;
- runs the shared 81-vector normative corpus (`scripts/normative-corpus.json`) against both interpreters — `scripts/run-corpus.mjs` for TS, generated Rust tests for Rust — every vector pins all six capability axes (label, thinking mode, effort set, default effort, wire route, normalization policy), and both runners reject sparse vectors;
- runs the schema-negative validator suite and `cargo test -p buzz-agent --lib` (the only CI gate exercising the `buzz-agent` unit suite; `just test-unit` excludes that crate).

## Model display labels

The manifest's `databricks_v2` exact records carry a `registry_label` axis with models.dev-verbatim display names for the ~30 managed Databricks endpoints. The generator derives a flat `DATABRICKS_MODEL_NAMES` registry from those records and emits it into both artifacts (Rust static slice, TS `Map`), so the two languages cannot drift.

Display rule: a model's display name comes from an exact/curated record only — family rules carry no labels, so a family-matched ID with no curated record resolves `registryLabel: null` and displays its raw ID (or its upstream discovery name). Endpoint IDs like `databricks-gpt-5-5` resolve to human-readable labels (`GPT-5.5`); unknown or custom workspace endpoints pass through as their raw ID unchanged, so a custom endpoint like `databricks-team-2025-01` can never be given an invented name.

## Rust label resolution

`databricks_model_name(id)` in `crates/buzz-agent/src/catalog.rs` reads the generated registry and is applied at every `ModelEntry` construction path — v1 and v2 discovery parsing, the authenticated-empty-catalog slate, and the configured-model fallback — so `AgentModelInfo.name` reaches the frontend already curated. Mirrors the existing `openai_model_display_name` precedent; no new IPC. `DATABRICKS_V2_KNOWN_MODELS` is likewise a re-export of the generated constant.

## Frontend label resolution

`resolveModelLabel(id, discoveredName?, provider?)` in `desktop/src/features/agents/lib/formatAgentModelLabel.ts` owns the label contract with three-tier precedence: nonblank discovered/API name, then registry lookup by ID (provider-qualified exact record when a provider is supplied; unscoped registry only on the providerless path), then raw ID. Every model-label surface routes through it — `ModelPicker`, `usePersonaModelDiscovery`, `agentCardModelLabel`, `AgentConfigFields`, `ManagedAgentRow`, `UserProfilePopover`. `formatAgentModelLabel()` wraps the resolver so a null or empty ID renders `Auto`.

## Tests

- Shared corpus 81/81 in both interpreters (full six-axis expectations per vector); schema-negative suite 42/42; `cargo test -p buzz-agent --lib` covers the corpus harness and generated-interpreter tests as a CI merge gate.
- Compact seam tests guard the live production boundaries: startup validation of pure-Anthropic effort values through `Config::validate()`, Anthropic-route effort normalization (`none`/`minimal` omitted from the wire), and the unsupported-effort clamp through `anthropic_body()` (a no-`xhigh` model record clamps `xhigh` to `high`).
- Corpus vectors pin the label contract: provider-qualified curated IDs resolve their exact-record labels (including the family-masquerade defect shape, `databricks-gpt-5-mini` → `GPT-5 Mini`), and a family-matched ID with no curated record pins `registry_label: null`; provider-alias tests pin `openai-compat`/`databricks-v2` canonicalization in the desktop surfaces.
- `usePersonaModelDiscovery.test.mjs` pins discovery-row label precedence; Rust unit tests in `catalog.rs` cover known IDs, custom endpoints, and the known-models fallback slate.

## Docs and build infra

`desktop/src/features/agents/AGENTS.md` documents label precedence and the regeneration workflow (edit `scripts/model-capabilities.json`, run `node scripts/generate-model-capabilities.mjs`); the generator and CI job emit the regeneration command on any stale-artifact failure. `desktop/scripts/check-file-sizes.mjs` gains a `fileOverrides` mechanism because the generated `modelCapabilities.ts` exceeds the 1000-line ceiling intended for hand-authored files.


kalvinnchau (dismissed):
Verified the doubled-separator parity fix at exact head 62ebc94314078c9b5728822d401192763255265f. The generated Rust matcher now collapses empty segments consistently with the TypeScript matcher, and the new six-axis corpus vector pins the previously divergent openai-responses route. Generator check and both corpus runners pass. No findings.

wesbillman (changes requested):
Reviewing on Wes Billman's behalf at exact head `5c361b3cd9c96073e3c299cd2005c4fda66e417b`.

**P2 — Keep the provider when formatting inherited/default models.**

This change correctly makes `resolveModelLabel(id, ..., provider)` provider-scoped, specifically so Databricks registry names cannot leak into Anthropic/OpenAI contexts (`formatAgentModelLabel.ts:60-72`). But both default-label paths discard provider and call `resolveModelLabel(model)` providerless:

- `agentCardModelLabel.ts:34-35,39,43,46-50`
- `AgentConfigFields.tsx:803-805` (even though `effectiveProvider` is already available at lines 280-285)

The providerless branch consults the unscoped `DATABRICKS_MODEL_NAMES` map. Consequently an inherited/default model whose raw ID happens to equal a curated Databricks endpoint ID is relabeled as Databricks even when its effective provider is `openai` or `anthropic`. The same agent's explicit-model path is provider-qualified, so switching only the model source can change the displayed identity for identical `(provider, model)` values. `ManagedAgentSummary.provider` is already the resolved effective provider (`types.rs:526-529`), so inherited cards have the information needed.

Please thread the effective provider through `formatDefaultModelLabel` and the global default-model label, and add regressions proving an OpenAI/Anthropic default such as `databricks-gpt-5-5` remains the raw ID while a `databricks_v2` default receives the curated label. The current test at `agentCardModelLabel.test.mjs:89-96` asserts a curated providerless default without establishing that the effective provider is Databricks; it should become provider-qualified.

Everything else I inspected held up: generator validation/injection boundaries, Rust/TS matching parity and precedence, provider aliases, request-time route/effort/thinking seams, fallback states, CI drift enforcement, and the generated-file size override. Focused validation passed: generator `--check`, 81/81 shared corpus, 42/42 schema-negative tests, file-size core tests, `cargo test -p buzz-agent --lib` (388 passed), and `git diff --check`.

The red Security check is the pre-existing MeshLLM 0.74/yanked `spin` dependency, and the E2E failures are unrelated onboarding assertions; neither is caused by this diff. The PR is also currently conflicting and will require revalidation after rebase.

wpfleger96:
superseded by https://github.com/block/buzz/pull/5597

PR #3604, opened 2026-07-29
fix(mobile): p-tag DM recipients so agents receive plain messages

### LINKED ISSUE #3591: Mobile: DM messages omit recipient p-tags, so agents never receive them
## Summary

Messages sent from the mobile client in a **DM channel** carry no `p` tags, so agent harnesses never receive them as a turn. The event reaches the relay and shows up in history, but the agent stays silent. The same message sent from Desktop works, because Desktop implicitly p-tags every DM participant.

## Root cause

Desktop has an explicit DM rule in `desktop/src/features/messages/lib/messageMentionPubkeys.ts`:

```ts
const candidates =
  channel.channelType === "dm"
    ? [...explicitMentions, ...channel.memberPubkeys, ...channel.participantPubkeys]
    : explicitMentions;
```

Its doc comment names this exact requirement:

> Stream messages notify only explicit mentions. A DM addresses every other participant, so it must carry recipient `p` tags even when the composer text contains no `@mention`. Agent harnesses and human notification subscriptions both rely on those tags.

Covered by `desktop/src/features/messages/lib/messageMentionPubkeys.test.mjs`: *"plain DM messages p-tag every recipient except the sender"*.

Mobile has no equivalent. `mobile/lib/features/channels/send_message_provider.dart` builds tags as:

```dart
final tags = <List<String>>[
  ['h', channelId],
  if (parentEventId != null) ..._buildReplyTags(parentEventId, rootEventId),
  for (final pk in normalizedMentions) ['p', pk],
  ...mediaTags,
];
```

`normalizedMentions` comes only from explicitly passed pubkeys or from `_resolveMentions()`, which scans the text for `@(\w+)`. There is no `channelType == 'dm'` branch anywhere in that file.

## Reproduction

1. Open a DM with a managed agent.
2. Send a plain message from Desktop, e.g. `hello` -> event has `["h", <channel>]` and `["p", <agent pubkey>]`; the agent replies.
3. Send the same plain message from mobile -> event has only `["h", <channel>]` (plus `["e", ...]` when replying in a thread); the agent never wakes.

Observed across 15 consecutive messages in one DM: all 9 events with a `p` tag produced an agent turn, all 6 without one were silently dropped. Correlation was exact; timing, message length and thread nesting made no difference.

## Impact

From mobile, a user cannot talk to an agent in a DM at all unless they know to type a literal `@AgentName`. There is no error and no visual difference: the message appears delivered.

## Suggested fix

Port the `messageMentionPubkeys` rule into `send_message_provider.dart`: when the channel is a DM, add the other participants to `normalizedMentions` before building the tags. Mirror the Desktop unit test on the mobile side.

## Workaround

On mobile, type `@AgentName` in the message body so `_resolveMentions()` produces the `p` tag.

Chessing234:
## Summary
- Port desktop's `messageMentionPubkeys` rule to mobile: DM sends now include `p` tags for every other participant/member, not only explicit `@mentions`
- Without those tags, agent harnesses never receive a turn for plain DM text from the phone (Desktop already worked)

## Test plan
- [x] `flutter test test/features/channels/message_mention_pubkeys_test.dart test/features/channels/send_message_provider_test.dart`
- [ ] Open a DM with a managed agent on mobile, send `hello` with no `@mention` — event carries `["p", <agent>]` and the agent replies
- [ ] Same plain message in a stream channel — still no invented recipient `p` tags
- [ ] Explicit `@Agent` mentions in a DM still work and stay deduped with the auto recipients

Fixes #3591

### DIFF
--- a/mobile/lib/features/channels/message_mention_pubkeys.dart
+++ b/mobile/lib/features/channels/message_mention_pubkeys.dart
@@ -1,0 +1,27 @@
+/// Return the semantic recipients for an outgoing message.
+///
+/// Stream messages notify only explicit mentions. A DM addresses every other
+/// participant, so it must carry recipient `p` tags even when the composer text
+/// contains no `@mention`. Agent harnesses and human notification subscriptions
+/// both rely on those tags.
+///
+/// Mirrors desktop `messageMentionPubkeys` in
+/// `desktop/src/features/messages/lib/messageMentionPubkeys.ts`.
+List<String> messageMentionPubkeys({
+  required bool isDm,
+  required String? senderPubkey,
+  required Iterable<String> explicitMentions,
+  required Iterable<String> memberPubkeys,
+  required Iterable<String> participantPubkeys,
+}) {
+  final candidates = isDm
+      ? [...explicitMentions, ...memberPubkeys, ...participantPubkeys]
+      : explicitMentions;
+
+  final selfLower = senderPubkey?.toLowerCase();
+  final seen = <String>{?selfLower};
+  return [
+    for (final pk in candidates)
+      if (pk.isNotEmpty && seen.add(pk.toLowerCase())) pk,
+  ];
+}
--- a/mobile/lib/features/channels/send_message_provider.dart
+++ b/mobile/lib/features/channels/send_message_provider.dart
@@ -4,13 +4,17 @@ import '../../shared/relay/relay.dart';
 import '../channels/channel_management_provider.dart';
 import '../profile/user_cache_provider.dart';
 import '../profile/user_profile.dart';
+import 'channel.dart';
 import 'channel_messages_provider.dart';
+import 'channels_provider.dart';
+import 'message_mention_pubkeys.dart';
 
 /// Sends messages by signing an event with the user's nsec and publishing it
 /// over the relay's NIP-42-authenticated WebSocket session.
 class SendMessage {
   final SignedEventRelay _signedEventRelay;
   final Future<List<ChannelMember>> Function(String channelId) _fetchMembers;
+  final Future<Channel?> Function(String channelId) _fetchChannel;
   final Map<String, UserProfile> Function() _readUserCache;
   final void Function(String channelId, NostrEvent event) _addLocalMessage;
   final void Function(String channelId, String eventId) _completeLocalMessage;
@@ -20,6 +24,7 @@ class SendMessage {
     required SignedEventRelay signedEventRelay,
     required Future<List<ChannelMember>> Function(String channelId)
     fetchMembers,
+    required Future<Channel?> Function(String channelId) fetchChannel,
     required Map<String, UserProfile> Function() readUserCache,
     required void Function(String channelId, NostrEvent event) addLocalMessage,
     required void Function(String channelId, String eventId)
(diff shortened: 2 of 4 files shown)


ScaleLeanChris:
Independent validation from a real Buzz mobile client, hosted Buzz relay, and the Fly Hermes identity:

- A plain mobile DM, “hey fly,” arrived as kind:9 with only the DM h tag. It had no recipient p tag.
- An explicit “@Fly yo” message added Fly's p tag.
- Two later plain DM messages again arrived without a p tag.

That reproduces #3591 at the event boundary and confirms the failure is in mobile message construction, not relay delivery. This PR's helper matches the desktop rule: plain DMs address every other participant, explicit mentions remain deduplicated, self is excluded, and stream messages do not invent recipient tags.

I mirrored the exact contributor commit into fork CI because the upstream workflow has not been approved yet. The [mobile job passed](https://github.com/ScaleLeanChris/buzz/actions/runs/30517412964/job/90790301474), including file-size checks, format, analysis, Flutter tests, and an Android debug APK build. The overall fork workflow also ran unrelated desktop jobs; its macOS desktop build failed before compiling Buzz because the mesh-llm checkout was unavailable.

One limitation: our Fly agent uses Hermes's native Buzz adapter, which already works around missing DM p tags by classifying the channel as a DM. Fly therefore still replied after its conversation was latched. This proves the malformed mobile event and the code correction, but it does not independently reproduce the managed ACP wake failure.

After this fix, a plain mobile DM can say “Hello” without also spelling out @Fly.

kerpopule:
Confirmed against Buzz 0.5.3 mobile plus a managed Hermes ACP agent configured with `subscribe=Mentions`. A plain mobile DM produced no agent acknowledgement or turn, while an explicit channel @mention produced the expected 👀/💬 acknowledgement and typing state. Current mobile send code only emits `p` tags for explicit mentions, so the harness cannot classify/address the plain DM. This PR fixes the observed failure at the sender boundary.

kerpopule:
Follow-up live canary: after channel replies became visible, a new mobile DM still produced no ACP prompt at all. At 2026-08-02 00:08 CDT the Hermes managed-agent log remained unchanged since its last prompt at 23:25:34 CDT. This rules out slow model execution and hidden reply placement for the DM case; the event is being dropped before harness dispatch, consistent with the missing mobile DM recipient `p` tag fixed here.

Chessing234:

Japiewb:
Independent confirmation of both the bug and the approach, from a production Buzz DM (hosted relay, Buzz Desktop + the iOS app, managed Claude Code ACP agent):

Raw events pulled from the same DM channel via `buzz messages get`:

- Sent from **mobile**: `"tags":[["h","<channel-uuid>"]]` — no recipient `p` tag; the managed agent was never woken and only saw the message once a later event triggered it.
- Same DM from **desktop**: `"tags":[["h","<channel-uuid>"],["p","<agent-pubkey>"]]` — agent triggered immediately.
- Mobile with an explicit `@mention`: `p` tag present, agent triggered — matching the workaround others reported here.

Before finding this PR we independently implemented the same rule in `SendMessage` (auto-p-tag DM participants with a member-list fallback, merged through the existing normalization so dedupe and self-exclusion still apply; stream/forum channels unchanged) and converged on the same behavior this PR describes — effectively N=2 on the approach. On today's `main` the full mobile suite passes with that change (1246 tests, `flutter analyze` clean, `dart format` clean).

Since this PR currently shows merge conflicts: a conflict-free equivalent based on today's `main`, with three added `send_message_provider` tests (DM auto-tag, non-DM no implicit tags, member-list fallback), is at [`Japiewb:fix/mobile-dm-recipient-p-tags`](https://github.com/block/buzz/compare/main...Japiewb:buzz:fix/mobile-dm-recipient-p-tags) in case it helps with the rebase — happy for any part of it to be folded in here. Not opening a second PR, to keep review in one place.

Chessing234:
closing — main already has mobile dm recipient p-tags via messageMentionPubkeys

PR #2481, opened 2026-07-23
fix(desktop): allow WebKitGTK media permission requests so huddles work on Linux

ryanrhughes:
## Problem

Huddles have never worked on the Linux desktop app: starting one tears itself down about a second later. The mic never connects.

Root cause: WebKitGTK **denies any permission request the embedder doesn't handle**, and wry's WebKitGTK backend registers no `permission-request` handler (it implements media permissions only on Android and macOS — verified against wry 0.55.1 sources). So `getUserMedia()` always rejects on Linux → `connectAndSetupMedia` throws → `cleanupFailedStart()` ends the huddle. The open dictation PR (#1511) adds UI copy for "mic permission denied" — that's this bug's symptom.

## Implementation

A small `linux-media-permissions` plugin in `desktop/src-tauri/src/lib.rs`: on webview ready, reach the native `webkit2gtk::WebView` via `with_webview` and connect `permission-request`, allowing exactly two request types:

- `UserMediaPermissionRequest` — mic/camera capture (huddles)
- `DeviceInfoPermissionRequest` — device labels for the input picker

Everything else returns `false`, keeping WebKitGTK's default deny — nothing beyond media devices is granted. Compiled out entirely on macOS/Windows.

`Cargo.toml` adds a Linux-only `webkit2gtk = "2.0"` dependency — the same crate + version Tauri already links (2.0.2 in the lockfile), needed only to name the types.

## Testing

An automated test is impractical here: the behavior lives in a native WebKitGTK signal handler that requires a live webview under a compositor. Verified manually on Arch Linux (Hyprland/Wayland, webkit2gtk-4.1 2.52.4):

- Before: start huddle → mic never connects → huddle drops after ~1s, every time.
- After: start huddle → mic connects, huddle stays up, audio works.

Manual test steps: on any Linux machine, build the desktop app, join a community, start a huddle in a channel. Without this patch it self-ends within ~2s; with it, it stays connected.

## Checklist

- `just ci`: all gates pass except `buzz-db` unit test `replica_fence::tests::fence_starts_closed_and_opens_on_advance`, which fails identically on pristine `main` (timestamp precision mismatch: `…237894Z` vs `…237894147Z`, last touched in #2084) — pre-existing and unrelated; happy to file a separate issue.
- Integration tests: N/A — no relay/db/auth changes.
- No new `unwrap()`/`expect()` in production paths; no `unsafe`.
- No new public API surface to document; behavior is documented in code comments.

PR #3607, opened 2026-07-29
desktop: enable getUserMedia in the Linux WebKitGTK webview

mbeckley17:
Microphone/camera capture works on macOS (WKWebView) and Windows (WebView2) but fails on Linux with `NotAllowedError`. WebKitGTK ships with `enable-media-stream` off and a default `permission-request` handler that denies every request.

This reaches the underlying `webkit2gtk::WebView` from `on_webview_ready` and enables `enable-media-stream`, then installs a **deny-by-default** `permission-request` handler: a `UserMedia` request is allowed only from a trusted app origin (`tauri://localhost` in prod, the Vite dev origin in debug) **and** when it targets an audio/video device — everything else is denied. No-op on macOS/Windows.

- `webkit2gtk` is pinned to the version wry already uses (`=2.0.2`) so there's a single shared copy of the native binding.

### DIFF
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -22,10 +22,6 @@ mesh-llm = ["dep:iroh", "dep:mesh-llm-sdk", "dep:mesh-llm-host-runtime", "dep:me
 # OS keyring backing for desktop secret storage (nsec private keys). When
 # disabled, secrets fall back to 0o600 files. On by default for real builds.
 system-keyring = ["dep:keyring"]
-# WebRTC peer connections / getDisplayMedia in the Linux webview. Needs a
-# WebKitGTK built with -DENABLE_WEB_RTC=ON; off by default so builds against
-# stock WebKitGTK keep working. Plain getUserMedia does not require this.
-webkit_webrtc = ["webkit2gtk/v2_38"]
 
 [build-dependencies]
 base64 = "0.22"
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -22,6 +22,10 @@ mesh-llm = ["dep:iroh", "dep:mesh-llm-sdk", "dep:mesh-llm-host-runtime", "dep:me
 # OS keyring backing for desktop secret storage (nsec private keys). When
 # disabled, secrets fall back to 0o600 files. On by default for real builds.
 system-keyring = ["dep:keyring"]
+# WebRTC peer connections / getDisplayMedia in the Linux webview. Needs a
+# WebKitGTK built with -DENABLE_WEB_RTC=ON; off by default so builds against
+# stock WebKitGTK keep working. Plain getUserMedia does not require this.
+webkit_webrtc = ["webkit2gtk/v2_38"]
 
 [build-dependencies]
 base64 = "0.22"
@@ -40,6 +44,10 @@ keyring = { version = "3.6.3", default-features = false, features = ["sync-secre
 # connection is dropped, which the plugin does immediately. Default features
 # keep the pure-Rust zbus backend, matching the plugin (no libdbus needed).
 notify-rust = "4"
+# Enable getUserMedia in the WebKitGTK webview (see src/linux_media.rs). Pinned
+# to the exact version wry links so both resolve to one webkit2gtk-sys and we
+# don't get duplicate symbols; bump in lockstep with wry.
+webkit2gtk = { version = "=2.0.2", features = ["v2_22"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 objc2-app-kit = { version = "0.3.2", default-features = false, features = ["NSHapticFeedback"] }
--- a/desktop/src-tauri/src/linux_media.rs
+++ b/desktop/src-tauri/src/linux_media.rs
@@ -1,0 +1,76 @@
+//! Linux-only: enable media capture (`getUserMedia`) in the WebKitGTK webview.
+//!
+//! On macOS (WKWebView) and Windows (WebView2) the media-permission prompt is
+//! routed to the OS automatically, so microphone/camera capture "just works".
+//! WebKitGTK is different on two counts, and both must be fixed or capture
+//! fails on Linux only:
+//!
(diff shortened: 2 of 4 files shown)


desktop/src-tauri/src/linux_media.rs
```diff
@@ -0,0 +1,76 @@
...
+            if request
+                .downcast_ref::<UserMediaPermissionRequest>()
+                .is_some()
+            {
+                request.allow();
+                true
+            } else {
+                false
+            }
+        });
```
wpfleger96: 🤖 Blocking: this allows every `UserMediaPermissionRequest` unconditionally — no check on the requesting origin and no check that the request is actually for audio/video. Tauri doesn't restrict navigation by default, so any document that ends up in this webview (external navigation, embedded content) would inherit silent mic/camera access for the process lifetime.

I think the shape we want is deny-by-default: allow only when the requesting origin is a trusted app origin (`tauri://localhost` in prod, the exact localhost dev origin in dev) and the request is for audio and/or video (`is_for_audio_device() || is_for_video_device()`), with an explicit `request.deny()` otherwise (still returning `true`). #3557 already implements and unit-tests the origin allowlist, and #2771 has the device-kind check — both are worth borrowing pretty much verbatim.

desktop/src-tauri/Cargo.toml
```diff
@@ -40,6 +44,10 @@ keyring = { version = "3.6.3", default-features = false, features = ["sync-secre
 # connection is dropped, which the plugin does immediately. Default features
 # keep the pure-Rust zbus backend, matching the plugin (no libdbus needed).
 notify-rust = "4"
+# Enable getUserMedia in the WebKitGTK webview (see src/linux_media.rs). Pinned
+# to the exact version wry links so both resolve to one webkit2gtk-sys and we
+# don't get duplicate symbols; bump in lockstep with wry.
+webkit2gtk = { version = "=2.0.2", features = ["v2_22"] }
```
wpfleger96, line 46: 🤖 The `=2.0.2` pin matching wry's exact version is nice dep hygiene — better than the loose versions in the sibling PRs. But `Cargo.lock` isn't committed: `cargo metadata --locked` fails at this head with "cannot update the lock file", so any `--locked`/`--frozen` consumer breaks. Blocking: please commit the lock update alongside this.

desktop/src-tauri/Cargo.toml
```diff
@@ -22,6 +22,10 @@ mesh-llm = ["dep:iroh", "dep:mesh-llm-sdk", "dep:mesh-llm-host-runtime", "dep:me
 # OS keyring backing for desktop secret storage (nsec private keys). When
 # disabled, secrets fall back to 0o600 files. On by default for real builds.
 system-keyring = ["dep:keyring"]
+# WebRTC peer connections / getDisplayMedia in the Linux webview. Needs a
+# WebKitGTK built with -DENABLE_WEB_RTC=ON; off by default so builds against
+# stock WebKitGTK keep working. Plain getUserMedia does not require this.
+webkit_webrtc = ["webkit2gtk/v2_38"]
```
wpfleger96: 🤖 As far as I can tell this feature has no shipping path — the Linux canary and release workflows build with default features, no CI job compiles `--features webkit_webrtc`, and buzz has no `getDisplayMedia` caller today, so this code is permanently dead and unchecked. wry 0.55.1 also already selects `webkit2gtk/v2_38`, so the feature's version bump is redundant at compile time. I'd drop the feature (and the WebRTC/`getDisplayMedia` mentions in the description) from this PR, and reintroduce it when there's an actual caller plus a package/CI contract for it.

desktop/src-tauri/src/linux_media.rs
```diff
@@ -0,0 +1,76 @@
...
+        // On Linux this is the underlying `webkit2gtk::WebView`.
+        let webview = platform_webview.inner();
+
+        if let Some(settings) = WebViewExt::settings(&webview) {
+            // The setting that actually makes `getUserMedia` reachable.
+            settings.set_enable_media_stream(true);
+            // Lets captured/remote streams play back via MediaSource.
+            settings.set_enable_mediasource(true);
+            // Capture is a deliberate user action; no gesture gate needed.
+            settings.set_media_playback_requires_user_gesture(false);
```
wpfleger96: 🤖 Minor: both of these match WebKitGTK's documented defaults (`enable-mediasource` is already `true`, `media-playback-requires-user-gesture` is already `false`), so neither is load-bearing here. I'd drop them and keep the diff to the actual fix — `set_enable_media_stream(true)` plus the permission handler.

wpfleger96 (changes requested):
🤖 Thanks for the fix — this is the right integration point (`on_webview_ready`, scoped to the main webview, GTK calls on the UI thread via `with_webview`) and the module docs are genuinely helpful. Two things are blocking for me, both inline below: the permission handler grants mic/camera to any document with no origin or device-kind check, and `Cargo.lock` isn't committed so `--locked` builds fail at this head.

For context, three other open PRs fix this same issue (#3495): #3557, #2771, and #2481. I'm treating this PR as the canonical fix, but a couple of patterns from those are worth borrowing — #3557 has an origin allowlist with unit tests, and #2771 checks `is_for_audio_device()`/`is_for_video_device()` before allowing. Details inline.

One more ask: since the behavior depends on a native WebKitGTK permission signal, it can't be covered by ordinary unit tests alone — it'd be great to record a quick manual check on a packaged Linux build here before merge (mic/camera capture works in a huddle, and a non-app origin gets denied).

kwsantiago:
Manual check on Linux, capture half only.

Built this branch locally (both commits, base 7e9b77f72) and ran it on Ubuntu 24.04, WebKitGTK 2.52.3, GeForce RTX 3050 Ti (driver 580.173.02), X11. Microphone capture works in a huddle.

Caveats:

- Local unsigned build, not a packaged AppImage.
- The branch is behind main.
- Launched with `WEBKIT_DISABLE_DMABUF_RENDERER=0` and EGL pinned to Mesa. That is a workaround for an unrelated WebKitGTK crash on NVIDIA (#3654) and has nothing to do with media capture.
- I did not test that a non-app origin gets denied.

wpfleger96 (approved):
thank you for this!

Issue #3277, opened 2026-07-28
Managed agents are only mentionable from the machine that runs them


4dlt:
**Describe the bug**

A managed agent can only be `@mentioned` from the machine that runs it. From any
other client on the same relay, the agent produces no autocomplete entry, so the
sent message carries no `p` tag and the agent's harness never dispatches. The
message lands in the channel as plain text and nothing happens.

There is no error and no UI signal that this is unsupported — the agent shows in
the member list, its kind:0 profile resolves, and it replies normally when
mentioned from its host machine. The failure is indistinguishable from a
misconfiguration, which makes it expensive to diagnose.

This blocks the "one shared agent on an always-on host" setup: today every person
has to run their own copy of every agent, producing duplicate identities with
duplicate display names.

**Steps to reproduce**

1. Point two Buzz clients (A and B) at the same relay.
2. On **A**, create a managed agent, add it to a channel, set `respond_to` to
   `anyone`, and start it. Confirm it is a `bot` member of that channel.
3. On **A**, `@mention` the agent in that channel — it autocompletes and replies.
4. On **B**, open the same channel and type `@` followed by the agent's name.
5. No autocomplete entry appears. Sending the text anyway produces an event with
   no `p` tag, and the agent never runs.

**Expected behavior**

The agent appears in B's autocomplete and is mentionable, producing the same
`p`-tagged event that A produces.

**Version and platform**

- Buzz version: v0.4.26 desktop (also reproduces on current `main`)
- OS: macOS

**Logs / additional context**

Two independent causes, either of which is sufficient on its own.

**1. The desktop app never publishes the kind:10100 agent profile it reads back.**
`list_relay_agents` queries `{"kinds":[10100]}` unfiltered, and
`agents_from_events` parses the result into `RelayAgentInfo`. But no code path in
`desktop/src-tauri` ever *publishes* a 10100 event — the only publisher in the
repo is `buzz channels set-add-policy` in the CLI, which writes only
`channel_add_policy`. The relay does not synthesize one either;
`handle_agent_profile` just reads `channel_add_policy` off it as a side effect.
So `relayAgentsQuery.data` is empty for any desktop-created agent, and the entire
relay-agent branch of `getMentionableAgentPubkeys` is unreachable in practice.

**2. `isAgentIdentityInManagedList` drops every non-locally-managed agent.**
Added in a4dfead ("fix(desktop): prefer live agent mentions", #2149), the gate at
[...]

Issue #3204, opened 2026-07-27
Mention picker lists locally-managed agents instead of channel members, so an agent running on another machine can never be @-mentioned


cmcgrabby-hue:
**Describe the bug**

The `@` mention picker does not build its candidate list from the channel's members. It appears to be scoped to the agents that *that Desktop install manages locally*, which produces two symptoms at once in the same channel:

1. It **offers an agent that is not a member of the channel** — selecting it sends a `p` tag to a pubkey the channel has never contained.
2. It **omits an agent that is a member of the channel** — there is no way to mention it at all.

The practical consequence: **an agent whose runtime lives on a different machine can never be `@`-mentioned.** Its private key lives in install A's Desktop, so install B's picker will never list it, no matter that it is a `bot` member of the channel and actively posting in it. There is no setting to change this.

This is adjacent to #2423 (which lists "mention autocomplete should resolve only current channel members" under *Expected behavior*) and #2648 (which explains where the duplicate identity comes from), but neither demonstrates the candidate-source defect directly. The two-channel control below isolates it.

**To Reproduce**

1. Install Buzz Desktop on machine A. Create an agent — call it `Claude`, pubkey `bf0e814b…` — and let it run there.
2. Add `bf0e814b…` as a `bot` member of two channels, **#alpha** and **#beta**.
3. Install Buzz Desktop on machine B, same community, same user. Per #2648 it mints its own agent identity — here `6285b97a…`, whose relay profile name is the persona-template default `Fizz`. It is a `bot` member of **#alpha only**. It has never completed setup and cannot respond.
4. On **machine B**, open **#beta** and type `@claude`.

**Expected behavior**

The picker offers `Claude` (`bf0e814b…`), the only agent member of #beta. If a locally-unmanaged agent cannot be mentioned for some architectural reason, the picker should say so rather than silently substituting a different pubkey.

**Actual behavior**

The picker offers **`Fizz`** — `6285b97a…`, which is **not a member of #beta** — and does not offer `Claude`, which is. Accepting the suggestion sends a mention to an agent that is not in the channel and cannot reply.

**Supporting material**

Relay-side member lists, `buzz channels members` (channel UUIDs redacted, pubkeys truncated):

```
#beta   bf0e814b…  role=bot      <- Claude, machine A
        fc4907bc…  role=owner    <- human

#alpha  bf0e814b…  role=bot      <- Claude, machine A
        fc4907bc…  role=owner    <- human
[...]

Issue #2508, opened 2026-07-23
[Bug] Remote Agent Missing from @ Mention Autocomplete Across Machines


Ansonhkg:
```
> Issue generated by 5.6 Sol High
```

**Describe the bug**

<img width="1672" height="941" alt="Image" src="https://github.com/user-attachments/assets/23996f70-18c2-480e-be5b-cd5116032c00" />

An agent named `Atlas`, managed and running on Machine A, is not discoverable or mentionable from Machine B, even though:

- Both users are members of the same channel.
- The agent is a member of that channel.
- The agent’s **Respond-to** setting is **Anyone**.
- Both machines are connected to the same Buzz community/relay.

Machine B’s `@` autocomplete only shows agents managed locally by Machine B. Manually typing `@Atlas` does not trigger the remote agent because the message is not sent with the agent’s Nostr `p` mention tag.

**To Reproduce:**

1. On Machine A, create and start a managed agent named `Atlas`.
2. Set the agent’s **Respond-to** setting to **Anyone**.
3. Add `Atlas` and the Machine B user to the same channel.
4. Confirm on Machine A that `Atlas` is running and listed as a channel member.
5. Open the same channel on Machine B.
6. Type `@` in the message composer.
7. Observe that `Atlas` is missing from autocomplete.
8. Manually type `@Atlas` and send the message.
9. Observe that the agent does not respond.

**Expected behavior:**

A running agent that:

- is a member of the current channel, and
- has **Respond-to: Anyone**

should appear in the `@` autocomplete for users on other machines. Selecting it should add the agent’s pubkey as a Nostr `p` tag so the remote agent receives and responds to the message.

**Supporting Material**

Screenshots show:

- The remote agent running on Machine A.
- **Respond-to** configured as **Anyone**.
- The remote agent listed as a member of the Test channel.
- Machine B’s autocomplete showing only locally managed agents and omitting the remote agent.
- Manually typed `@Atlas` messages receiving no response.
- The agent responding successfully when invoked from Machine A.

Technical observation: the client initially recognizes shared relay agents with `respond_to: anyone` as mentionable, but a subsequent filter removes every agent that is not present in the

mzorz:
**+1, reproduced with two different users (not just two machines belonging to the same account)**

[...]

Issue #2349, opened 2026-07-22
[Bug] Remote managed agents are hidden and unmentionable on secondary devices


geff08565-sketch:
## Summary

A cloud-connected managed agent that runs on Mac A is not visible or mentionable from Buzz Desktop on Mac B, even when both desktops use the same community/account and the agent is already a member of the channel.

The agent is visible and mentionable on Mac A. On Mac B, direct messages to the agent work, but the agent is absent from the channel member/mention UI.

This appears to couple two separate concepts:

1. discovering/mentioning a remote cloud agent; and
2. owning and launching that agent's local runtime.

## Environment

- Buzz Desktop 0.4.21
- macOS on both machines
- Same hosted Buzz community and same human identity/account
- Agent runtime (`buzz-acp` with an OpenClaw ACP adapter) runs only on Mac A
- Agent is connected to the hosted relay, online, and subscribed to the channel

## Steps to reproduce

1. On Mac A, create/configure a managed ACP agent and connect it to a hosted Buzz community.
2. Add the agent to a channel as a bot member.
3. Confirm on Mac A that the agent appears in the channel and responds to `@mentions`.
4. On Mac B, sign in to the same hosted community with the same human identity/account.
5. Open the same channel and type `@`.

## Actual behavior

- Mac B does not show the remote agent in the mention/member UI.
- Other agents present in Mac B's local `managed-agents.json` are shown (including agents marked "not in channel").
- The missing remote agent is nevertheless reachable by direct message from Mac B.
- Relay/channel membership is correct, and the remote runtime on Mac A is online.
- Restarting Buzz Desktop on Mac B does not fix the problem.

As a diagnostic experiment, adding a keyless/inactive record for the remote agent's public key to Mac B's local `managed-agents.json` makes the agent appear in `@` autocomplete. However, mentioning it then makes Buzz attempt to start the agent locally on Mac B and produces a warning that the OS keyring may be unreachable. Mac B does not possess the agent's private key by design.

This diagnostic record was not intended as a supported workaround; it demonstrates that mention visibility is gated by local managed-agent ownership and that mention dispatch then invokes the local lifecycle path.

## Expected behavior

- Any agent that is a member of the current hosted channel should be visible and mentionable from every signed-in client authorized to view that channel.
- Only the machine that owns the agent's private key/runtime configuration should be able to start, stop, or reconfigure the runtime.
[...]

PR #3609, opened 2026-07-29
fix(desktop): keep cross-owned invocable agents in mention autocomplete

### LINKED ISSUE #3125: Desktop hides allowlisted agents owned by another user from @mention autocomplete

ppcmurray:
## Why

Fixes #3125 (also the root of the cross-machine flavor reported in #3204 / #3277 / #2508 / #2349 when the agent is owned by another user).

The managed-list gate added in #2149 runs **before** `shouldHideAgentFromMentions` in mention autocomplete, so an agent owned by another user is discarded even when its `respond_to` policy explicitly admits the current user. The downstream invocability logic is unreachable for cross-owned agents: the mention never resolves, the published event carries no agent `p` tag, and ACP is never woken.

Reproduced on a self-hosted relay: owner's client resolves the agent fine; an allowlisted second user's `@Agent` stays plain text.

## What

- `isAgentIdentityInManagedList` gains an optional `invocableAgentPubkeys` parameter and admits candidates present in either set. The mention path passes `mentionableAgentPubkeys` (already "managed ∪ relay agents shared with me" via `getMentionableAgentPubkeys`), so:
  - cross-owned agent, allowlist names me / `respond_to: anyone` in a shared channel → **kept**, falls through to the existing eligibility check;
  - stale identity (neither managed nor invocable) → **still dropped**, preserving the duplicate suppression #2149 introduced;
  - people → never gated (unchanged).
- The add-member search in `MembersSidebar` intentionally keeps managed-only semantics via the defaulted parameter.
- Unit tests for the new behavior incl. case-insensitivity and the stale-identity case.

`pnpm test` (3672 pass) and biome clean.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -162,6 +162,48 @@ test("isAgentIdentityInManagedList: keeps people and only current managed agent 
   );
 });
 
+test("isAgentIdentityInManagedList: keeps cross-owned agents the current user can invoke (#3125)", () => {
+  const managedAgentPubkeys = new Set([PUB_A]);
+  const invocableAgentPubkeys = new Set([PUB_A, PUB_B]);
+
+  // Another owner's agent whose allowlist/anyone policy admits us: keep.
+  assert.equal(
+    isAgentIdentityInManagedList(
+      { isAgent: true, pubkey: PUB_B },
+      managedAgentPubkeys,
+      invocableAgentPubkeys,
+    ),
+    true,
+  );
+  // Case-insensitive on the invocable set too.
+  assert.equal(
+    isAgentIdentityInManagedList(
+      { isAgent: true, pubkey: PUB_B.toUpperCase() },
+      managedAgentPubkeys,
+      invocableAgentPubkeys,
+    ),
+    true,
+  );
+  // Stale identity: neither managed nor invocable — still dropped (#2149).
+  assert.equal(
+    isAgentIdentityInManagedList(
+      { isAgent: true, pubkey: PUB_C },
+      managedAgentPubkeys,
+      invocableAgentPubkeys,
+    ),
+    false,
+  );
+  // People are never gated.
+  assert.equal(
+    isAgentIdentityInManagedList(
+      { isAgent: false, pubkey: PUB_C },
+      managedAgentPubkeys,
+      invocableAgentPubkeys,
+    ),
+    true,
+  );
+});
+
 test("shouldHideAgentFromMentions: never hides non-agents", () => {
   assert.equal(
     shouldHideAgentFromMentions({
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
@@ -54,13 +54,26 @@ export function getMentionableAgentPubkeys({
   return pubkeys;
 }
 
+const EMPTY_PUBKEY_SET: ReadonlySet<string> = new Set();
+
 export function isAgentIdentityInManagedList(
   candidate: { isAgent?: boolean; pubkey: string },
   managedAgentPubkeys: ReadonlySet<string>,
+  // The managed list is only authoritative for the CURRENT user's concrete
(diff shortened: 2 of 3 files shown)


vesc-baek:
## Field confirmation + review ping (2026-07-30)

Reproduced independently on a self-hosted Buzz community (hostname redacted):

- Cross-owned channel bots (hermes/Fizz on always-on host) are **bot** members and respond when a correct `p` tag is present (Mobile picker or CLI `--mention`).
- Desktop `@` autocomplete for an allowlisted non-owner still omits those agents — matches the #2149 managed-list gate this PR widens with `invocableAgentPubkeys`.
- CI on this PR reports green (Semgrep / zizmor / DCO). No reviews yet.

This blocks multi-user always-on agent hosting for us. Requesting maintainer review/merge when capacity allows — we will retest immediately after a Desktop build that includes this change.

Refs: #3125 #3277 #3204 #2508

---
Note: concrete community/relay hostname removed; not required for review.

lamcode9:
+1 on this fix — hit this exact wall in the field, with one extra wrinkle worth covering.

Setup: agents hosted on an always-on Mac ("host"), a second Mac used as a plain chat client, same relay, all agents added as role=bot channel members. On the client machine the @-picker shows no agents at all, anywhere. Root cause is the managed-list gate from #2149 (`isAgentIdentityInManagedList` in `useMentions.ts`) dropping every agent candidate before `shouldHideAgentFromMentions` — so the "member with unknown invocability => show" branch from #1611 is unreachable, exactly as this PR describes.

The wrinkle: this PR's gate passes agents via `mentionableAgentPubkeys`, which is built from kind:10100 directory entries. On our relay no agent had ever published a kind:10100 profile (nothing in the desktop's managed-agent flow publishes one), so the invocable set is empty and channel-member agents would STILL be dropped after this fix. Suggestion: also pass candidates with `isMember === true` through the gate — membership is relay-verified, and `shouldHideAgentFromMentions` already handles the member + directory-excluded case correctly. I've been running `desktop-v0.5.4` patched with this PR plus the one-line `isMember` bypass and it resolves the client-machine picker completely (all 4023 desktop unit tests pass).

Longer-term, the picker could source candidates from `list_relay_agents`/channel membership directly so a machine that hosts nothing can still mention everything — happy to file that separately if useful.

PR #3610, opened 2026-07-29
fix(desktop): queue agent draft requests

### LINKED ISSUE #3559: `buzz agents draft-update` silently drops a draft when one is already pending
**Component:** `buzz` CLI (`agents draft-update` / `draft-create`) + Desktop draft-review surface
**Severity:** Medium — silent data loss; the CLI reports success for a draft the owner never sees

## Observed

Three `draft-update` requests were sent in quick succession (product-manager, compliance-manager, v2-program-manager), each returning `{"accepted": true, "saved": false}`. Only the **first** surfaced in Desktop for review; the other two never appeared. Re-sending them **one at a time**, with the owner saving each before the next was sent, worked — each surfaced in turn.

This indicates Desktop holds a **single pending-draft slot**: a new draft request arriving while one is already pending is dropped, with no error to the sender. `accepted: true` reflects only that the relay routed the request, not that Desktop retained it.

## Impact

An operator scripting several agent updates in a loop (the natural thing to do) silently loses all but one, and the CLI gives them no signal that anything was dropped.

## Suggested fix

Queue pending drafts in Desktop rather than holding one slot, or have the relay/Desktop NACK a draft that can't be retained so the CLI can report `saved: false, retained: false` and the caller can retry/serialize. At minimum, document that draft requests must be sent one at a time and reviewed before the next is sent.

kiranmagic7:
## Summary

`useAgentManagement` marked every accepted agent-management request as seen before checking whether another draft was open. Later requests were therefore lost for the rest of the Desktop session.

This change keeps accepted requests in an in-memory FIFO and advances after the current draft is saved or dismissed. Each item retains the requesting agent pubkey, and request errors are cleared when the next draft opens. The queue is deliberately uncapped: these events are authenticated and owner-scoped, and a cap would recreate silent loss.

### Related issue

Fixes #3559.

#2454 is an earlier draft with a broader persistence and sidebar design. This PR is limited to the in-session FIFO needed for #3559.

### Testing

- `pnpm --dir desktop exec node --experimental-strip-types --test src/features/agents/agentManagementBuffer.test.mjs`
- `just ci`
- Screenshots: N/A (no visual change)

### DIFF
--- a/desktop/src/features/agents/agentManagementBuffer.test.mjs
+++ b/desktop/src/features/agents/agentManagementBuffer.test.mjs
@@ -1,7 +1,12 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 
-import { classifyAgentManagementOrigin } from "./agentManagementBuffer.ts";
+import {
+  advanceAgentManagementRequest,
+  classifyAgentManagementOrigin,
+  enqueueAgentManagementRequest,
+} from "./agentManagementBuffer.ts";
+import { AGENT_MANAGEMENT_REQUEST } from "./agentManagement.ts";
 
 const AGENT = "a".repeat(64);
 const CHANNEL = "channel-1";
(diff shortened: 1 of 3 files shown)


kiranmagic7:
Status check on head `5b8d967411a7c55b8a4a97eaa555a079e498a571`: the PR is still mergeable, with DCO Check, Semgrep OSS, and zizmor green. The full Actions run is waiting on fork workflow approval rather than failing code: https://github.com/block/buzz/actions/runs/30485508497 (`action_required`, 0 jobs).

Review focus: the small draft-request buffer in `agentManagementBuffer.ts` and its regression coverage in `agentManagementBuffer.test.mjs` for #3559. The repo CODEOWNERS path is already `@block/buzz-oss-team`, so I am leaving routing as-is.

kiranmagic7:
The new Desktop 0.5.8 reproduction on #3559 confirms the same failure mode: later requests enter `seenRequestIds` while a draft is pending, so dismissing the current draft cannot surface them. The FIFO buffer in this branch is the requested retention path. GitHub reports the current head mergeable; hosted CI remains at the fork-approval gate with zero jobs.

nathansmithopenclaw-alt:
Thanks for addressing #3559. We independently traced the same failure mode
while reviewing agent-draft configuration parity: a single pending slot makes a
successful CLI relay response look like delivery even when a later request
replaces the draft before the owner can review it.

We are not planning a competing queue PR. FIFO presentation is the right
boundary for this fix, and we would be glad to exercise the merged shape with
two rapid create/update drafts.

One adjacent gap is lifecycle evidence for the requester. Relay acceptance does
not prove that Desktop retained, presented, dismissed, or applied a draft. We
are treating owner-signed encrypted receipts as a separate follow-up, not asking
this PR to absorb that scope. One compatibility question: will the queue expose
a single point after successful insertion where a future receipt can accurately
say `retained`?

If restart durability is intentionally out of scope here, I also suggest saying
that explicitly. An in-session FIFO fixes concurrent replacement without
implying that pending drafts survive a Desktop restart.

kiranmagic7:
Yes. In acceptOwnedRequest, the point immediately after requestQueueRef.current = nextQueue is the successful in-memory retention boundary: the request is then either active or appended to queued, so a future receipt could accurately report retained there. Restart durability is intentionally out of scope for this patch; the queue lives in a React ref, so pending drafts do not survive a Desktop restart.

PR #3615, opened 2026-07-29
fix(desktop): accept npub or hex pubkey when adding members

joahg:
## Problem

Two member-add surfaces in Buzz Desktop made it impossible to add someone by their public key:

- **Relay members** (Settings → Invites → Add Member): the pubkey field only accepted a 64-character hex string and rejected `npub1…` input — the format most users actually have on hand.
- **Channel members** (Add members picker): search-only. User search matches kind:0 profile events, so an identity with no profile on the relay (e.g. a freshly generated client keypair) could not be invited at all.

## Fix

- New `parsePubkeyInput` helper in `shared/lib/nostrUtils.ts`: parses trimmed input as 64-char hex or bech32 `npub1…` (via nostr-tools nip19) into a lowercase hex pubkey; returns `null` otherwise. Unit tests cover hex, uppercase hex, npub, whitespace tolerance, bad checksum, nsec rejection, and wrong-length input.
- **AddMemberDialog** now accepts either form; placeholder and validation copy updated; `maxLength` removed so a pasted npub isn't truncated.
- **ChannelMemberInviteCard** now shows an "add by public key" row when the query parses as a pubkey that isn't already a member, already selected, or present in search results — profile-less identities can be invited directly.

## Validation

- `pnpm typecheck` — clean
- `pnpm test` — 3783 pass
- `pnpm check` (biome + file-size/px-text/pubkey-truncation checks) — clean

Internal tracking: AGNTOPS-364

### DIFF
--- a/desktop/src/features/channels/ui/ChannelMemberInviteCard.tsx
+++ b/desktop/src/features/channels/ui/ChannelMemberInviteCard.tsx
@@ -1,6 +1,7 @@
 import { Search, UserPlus, X } from "lucide-react";
 import * as React from "react";
 
+import { parsePubkeyInput } from "@/shared/lib/nostrUtils";
 import { truncatePubkey } from "@/shared/lib/pubkey";
 import { PubKey } from "@/shared/ui/PubKey";
 import { useIsArchivedPredicate } from "@/features/identity-archive/hooks";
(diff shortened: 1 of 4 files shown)


joahg:
🤖 Closing as part of Joah's pre-OOO PR triage — stale. Happy to reopen if this is still needed.

PR #3618, opened 2026-07-29
fix(acp): coordinate HTTP rate-limit retries

dbett4:
## Summary
- share one process-local HTTP 429 cooldown across RestClient clones
- serialize bridge admission so queued requests re-check the gate after a peer arms it
- honor integer Retry-After hints with conservative defaults and a bounded cap
- add concurrent regression coverage for the retry-storm path

## Verification
- cargo fmt --check -- crates/buzz-acp/src/relay.rs
- cargo test -p buzz-acp
- 636 unit tests and 9 lifecycle tests passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -210,6 +210,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
+name = "assert-json-diff"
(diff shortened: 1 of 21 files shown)


dbett4:
Superseded by #5620.

This PR bundled six unrelated commits under a title describing only the first, which made it hard to review. #5620 is just the rate-limit fix — `+297/−9` across three files, rebased onto current `main` — with the concurrency test verified as a real detector rather than only a passing test.

Closing this. The five agent reply-credential commits will be resubmitted separately if they are wanted.

PR #3642, opened 2026-07-29
fix(desktop): channel topic and membership metadata cleanup

delkc:
First slice of #2216, scoped to the system/status lines in the chat timeline.

## Why

Three problems on the same surface.

**Clearing a channel topic renders as empty quotes.** The relay reports a clear as a `topic_changed` event carrying an empty string — there's no separate "cleared" event type. So the timeline printed:

> Alice
> changed the topic to “”

which reads as if the topic were *set to* two quote marks. Same for purpose.

**The membership caption reads like a headline, not a metadata line.** `title` and `action` render on separate lines — the member's name sits in the header row with the avatar and timestamp, and the caption sits beneath it. So the caption was "was added by Alice Chen" standing alone under a name, while its siblings on that same line are "joined the channel" and "left the channel".

**"You" was capitalized two different ways on one row.** An agent's metadata reads "managed by you" while the activity line directly under it read "added by You" — the same word, two capitalizations, a few pixels apart.

## What

- Blank, missing, or whitespace-only topic/purpose now reads **"cleared the channel topic"** / **"cleared the channel purpose"**.
- Membership captions drop "was": **"added by Alice Chen"**, matching "joined the channel" and "left the channel". Rationale below.
- Self-references inside a caption are lowercase: **"added by you"**, **"removed you from the channel"**, **"along with you"**. The row *title* still says "You" — that's a name standing on its own next to the avatar, same as any other person's.
- The wording moves to `lib/systemEventCopy.ts` as pure functions, so it's assertable in a unit test instead of only reachable through the DOM. That also removes two JSX fragments from `SystemMessageRow.tsx`.

## Why "was" is gone rather than dynamic

Restoring "was" means agreeing with the subject, and the subject can be "You" — "You was added by Alice" is broken. `label === "You" ? "were" : "was"` is a one-line fix, so this was a judgment call, not a cost one. Three reasons it went the other way:

**The verb would be agreeing with a subject on a different line.** The row is `flex-col`: the name sits in the header with the avatar, the timestamp, a middot, and sometimes a "managed by" chip; the caption is a separate `<p>` beneath it. "was added by Alice Chen" is a sentence fragment reaching back across a metadata row for its subject. The UI isn't presenting it as a sentence, so it shouldn't depend on being read as one.

**Agreement keys on user-controlled data, and fails loudly.** The rule can only special-case the strings it knows. A display name like "Design Team" or any user-set plural reads wrong, and every future passive line ("blocked by", "invited by", "removed by") has to re-derive the rule. When capitalization is slightly off it's a polish nit; when agreement is off the user sees visibly broken grammar.

**Consistent active voice isn't reachable without a design change.** 10 of the 12 cases are already active. The 2 passive ones are exactly `members_added` and `member_joined` where actor ≠ target — and those rows are deliberately titled with the person who *arrived*, not the person who added them (it's why they get the dual avatar with the new member's face first). Self-joins in the same slot read active only because actor and target are the same person. So the passive voice is a consequence of the title choice, not a copy slip.

Removing the passive voice properly means re-titling those rows with the actor and going active — "Alice added Marcia to the channel, along with Peter, Jordan, and 2 others." That matches `member_removed` exactly and is what Slack does, but it changes whose avatar leads the row. Filed as a follow-up on #2216 rather than smuggled into a copy PR.

## Two E2E assertions this exposed

Both were measuring something other than what they claimed, and the copy change tipped them over. Neither is a product bug, but both would have failed the next person too.

1. **`mentions.spec.ts:1245`** asserted a button was un-underlined while the mouse was still parked from a previous `hover()`. Any reflow — new rows, scroll-to-bottom, a different text wrap — can slide that button under the stationary pointer, so the assertion measured *where the mouse happened to be* rather than the resting style. Dropping four characters changed the text wrap, changed the row height, changed the scroll offset, and the pointer landed on it. Now parks the pointer off-target first.
2. **`mentions.spec.ts:1253`** used a bare `role=tooltip` lookup. Once the first tooltip animates out while the second opens, two elements match and strict mode trips. Now scopes to the open tooltip via `:not([data-state="closed"])`.

## Deliberately out of scope

- **Timestamps.** The day divider, per-message clock times, the Inbox thread pane, and the inbox list have three divergent date implementations and none fully match the writing standard's Today/Yesterday/weekday/date progression. That's its own slice of #2216.
- **Whose avatar and name lead the row.** See the active-voice discussion above — it's a design question, not copy.
- **"Added by You" in the persona catalog.** `PersonaAddedBy.tsx` has the same capitalization mismatch, but it's the agents feature rather than channel system messages.
- **`the channel` vs `this channel`.** joined/left/removed say "the channel"; created/archived/unarchived say "this channel". Worth normalizing, but it touches lines this PR otherwise leaves alone.

## Validation

- `pnpm check`, `pnpm typecheck` — clean
- Unit: **3784/3784**, including 6 tests covering set/blank/undefined/null/whitespace for both fields, a guard that no variant can emit empty quotes, and 3 for the inline-name rule (including that "Youssef" and "You Know Who" are left alone)
- CI green on all 20 checks, including all 4 Smoke E2E shards and both Desktop E2E Integration shards
- The previously fragile assertions run locally with `--repeat-each=5`: **5/5**

I could not produce a system row with the current user as actor in mock data, so the lowercase-"you" path is covered by unit tests and typecheck rather than a screenshot.

### DIFF
--- a/desktop/src/features/messages/lib/systemEventCopy.test.mjs
+++ b/desktop/src/features/messages/lib/systemEventCopy.test.mjs
@@ -1,0 +1,59 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { describeChannelTextFieldChange } from "./systemEventCopy.ts";
+
+test("a set topic is quoted verbatim", () => {
+  assert.equal(
+    describeChannelTextFieldChange("topic", "Release planning"),
+    "changed the topic to “Release planning”",
+  );
+});
+
+test("a set purpose names the purpose, not the topic", () => {
+  assert.equal(
+    describeChannelTextFieldChange("purpose", "Where we ship from"),
+    "changed the purpose to “Where we ship from”",
+  );
+});
+
+// The relay reports a clear as a change carrying an empty string, so without
+// this branch the timeline reads: changed the topic to “”.
+test("an empty value reads as cleared, not as a change to empty quotes", () => {
+  for (const blank of ["", undefined, null]) {
+    assert.equal(
+      describeChannelTextFieldChange("topic", blank),
+      "cleared the channel topic",
+    );
+    assert.equal(
+      describeChannelTextFieldChange("purpose", blank),
+      "cleared the channel purpose",
+    );
+  }
+});
+
+test("a whitespace-only value reads as cleared", () => {
+  assert.equal(
+    describeChannelTextFieldChange("topic", "   \n\t "),
+    "cleared the channel topic",
+  );
+});
+
+test("surrounding whitespace is trimmed out of the quotes", () => {
+  assert.equal(
+    describeChannelTextFieldChange("topic", "  Release planning  "),
+    "changed the topic to “Release planning”",
+  );
+});
+
+test("no caption announces empty quotes", () => {
+  for (const value of ["", " ", null, undefined, "Real topic"]) {
+    for (const field of ["topic", "purpose"]) {
+      assert.doesNotMatch(
+        describeChannelTextFieldChange(field, value),
+        /“”|""/,
+        `${field} with ${JSON.stringify(value)} must not render empty quotes`,
+      );
+    }
+  }
+});
--- a/desktop/src/features/messages/lib/systemEventCopy.test.mjs
+++ b/desktop/src/features/messages/lib/systemEventCopy.test.mjs
@@ -26,11 +26,11 @@ test("an empty value reads as cleared, not as a change to empty quotes", () => {
   for (const blank of ["", undefined, null]) {
     assert.equal(
       describeChannelTextFieldChange("topic", blank),
-      "cleared the channel topic",
+      "cleared the topic",
     );
     assert.equal(
       describeChannelTextFieldChange("purpose", blank),
-      "cleared the channel purpose",
+      "cleared the purpose",
     );
   }
 });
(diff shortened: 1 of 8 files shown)


delkc:
This also fixes that `You was added` bug. 
Screenshot 2026-07-23 at 9 34 54 AM

tellaho:
🤖 Nice, focused improvement overall. The pure copy extraction is a good shape, the empty-topic guard is well tested, and the two E2E adjustments address the underlying tooltip flake mechanics cleanly.

**One blocking fix**

Could we make the inline self-label depend on identity rather than the resolved display string? As written, `toInlineName(label === "You")` would also lowercase another user whose display name is literally `You`. Comparing the relevant pubkey with `currentPubkey` using the existing normalization pattern would keep the behavior precise; the pure helper could take an `isSelf` boolean if useful.

**Non-blocking follow-ups**

- Mobile still shows the empty-topic and `was added by` variants, so tracking desktop/mobile copy parity on #2216 would be worthwhile.
- The copy varies between `topic` / `channel topic` and `the channel` / `this channel`; those nouns could be normalized in a follow-up.
- An earlier E2E assertion still uses an unscoped `getByRole("tooltip")`. It is safe with the current flow, but may be worth tightening if that test gains another hover.

With the identity-based self-label fix, this looks good to me.

tellaho (approved):

PR #3652, opened 2026-07-30
feat(desktop): show which config fields need a restart

wpfleger96:
The "Restart required" badge is driven by a lossy `u64` digest of the effective spawn config (`spawn_config_hash()`), so it can report that something drifted but never what. This replaces the digest with a typed `SpawnConfigSnapshot` and ships a redacted field-by-field diff on `ManagedAgentSummary.restart_diff` for the UI to render.

## One comparison, one vector

`SpawnConfigSnapshot::canonical()` is the single JSON projection that both the badge and the diff read. Drift is `to_value(stamped) != to_value(current)`; the diff is a generic leaf walk over those same two values, so badge-on and diff-non-empty hold by construction rather than by convention.

`eligible_restart_diff()` returns the final vector — snapshot walk entries plus a synthetic `adapter_availability` entry. It returns empty for an orphaned instance (`spawn_agent_child` refuses to spawn one, so the badge would offer an action guaranteed to fail) and for an agent with no tracked pair runtime, which was never stamped and so can never be shown to have drifted. Both ineligible states are inputs to that one helper rather than conditions at the call site. `needs_restart = !restart_diff.is_empty()` is derived from that vector and nothing else; `restart_eligible()` is gone.

## Generic by default

The walk knows nothing about which fields exist: serde field names are the diff path ids, dynamic map keys append verbatim (`env.OPENAI_API_KEY`), and object-key unions are walked in lexicographic order for stable UI truncation. Adding a snapshot field reaches the UI with no change to the diff code. `mutation_table_covers_every_serialized_field` fails if a new field arrives without a mutation row.

The only per-field knowledge is `policy_for(path)`, four arms:

| Policy | Paths | Rendering |
|---|---|---|
| `Text` | `system_prompt`, `team_instructions` | character counts only |
| `MaskedBare` | `args`, `relay_url` | `••••`, no suffix |
| `MaskedSuffix` | `auth_tag`, `env.*` | `••••` + last 4 chars when longer than 8 |
| `Plain` | everything else | verbatim |

`args` and `relay_url` get no suffix because both can legitimately carry credentials — `--token=...` is a legal argument, and `normalize_relay_url` rejects userinfo but deliberately preserves query strings. Masking is character-based throughout, and raw values drive the comparison so two secrets with colliding suffixes still read as drift.

That same policy backs the snapshot's manual `Debug` via `redacted_canonical()`. `ManagedAgentProcess` derives `Debug`, so a derived impl on the snapshot would print secrets to logs; routing both surfaces through one policy means they cannot diverge.

## Pre-spawn stamping

The stamp is now built from the already-resolved values that populated the `Command`, before `command.spawn()`. The previous post-spawn re-resolution left a window where a persona, harness, or global edit landing in between would stamp the *new* config onto a child running the *old* one, silently suppressing the badge.

## Wire shape

```jsonc
"restart_diff": [
  { "field": "model",              "change": { "kind": "value",  "before": "gpt-5", "after": "claude-4" } },
  { "field": "system_prompt",      "change": { "kind": "text",   "before_chars": 1234, "after_chars": 1410 } },
  { "field": "env.OPENAI_API_KEY", "change": { "kind": "masked", "before": "••••bc12", "after": "••••xyz9" } },
  { "field": "env.BUZZ_LOG",       "change": { "kind": "added" } }
]
```

Omitted when empty. `added`/`removed` occur only for dynamic-map keys; nullable struct fields always serialize as `null`, and arrays are atomic leaves (`args`, never `args.0`).


wpfleger96:
🤖 Consolidated into [#3637](https://github.com/block/buzz/pull/3637) at Will's request. All of Duncan's Rust changes (spawn snapshot, diff engine, redaction policy, tests) are present in #3637 at tip `ad4cdb1ea`. Closing this PR.

PR #3672, opened 2026-07-30
fix(desktop): keep channel-member agents mentionable without a local managed list

### LINKED ISSUE #3671: Desktop: agents owned by another user are never mentionable — managed-list guard in useMentions dead-codes shouldHideAgentFromMentions
# Desktop: agents owned by another user are never mentionable — the managed-list guard in `useMentions` dead-codes `shouldHideAgentFromMentions`

## Summary

On Buzz Desktop, an agent deployed by user A is invisible in user B's `@`-mention autocomplete — even when the agent is a **member of the same channel** and B is on the agent's `respond_to` allowlist. Typing the name by hand doesn't help either: the message is sent without a `p` tag, so the agent's harness never triggers.

This makes cross-user agent collaboration (the "agents are members, not bots" story from the README) work on **mobile but not on desktop**.

## Root cause

In `desktop/src/features/messages/lib/useMentions.ts`, `addCandidate` runs two gates in sequence:

```ts
if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) return;   // gate 1
if (shouldHideAgentFromMentions({ ..., mentionableAgentPubkeys, ... })) return; // gate 2
```

1. **Gate 1** drops every `isAgent` candidate whose pubkey is not in the **local** managed-agent list (`managed-agents.json`). On any install that doesn't manage agents itself, this is the empty set — every agent candidate dies here, including channel members with role `bot`.
2. **Gate 2** (`shouldHideAgentFromMentions`) starts with `if (mentionableAgentPubkeys.has(normalized)) return false; // Invocable => always show`.
3. `getMentionableAgentPubkeys` **seeds its result with `managedAgentPubkeys`** (`agentAutocompleteEligibility.ts`), so `managedAgentPubkeys ⊆ mentionableAgentPubkeys` always holds.

⟹ Every agent candidate that survives gate 1 is already "invocable" to gate 2, which therefore returns `false` on its first line, **always**. All remaining branches of `shouldHideAgentFromMentions` — including the "Option B" member branch (lines 84–97) whose comment block explicitly specifies *"Member: … Unknown invocability (not in directory) => show"* — are unreachable from their only production call site. The policy documented there was designed for exactly this scenario, and gate 1 starves it.

## Evidence this is a regression, not intent

- **The mobile client is the witness.** `mobile/lib/features/channels/mentions/mention_candidates.dart` implements the same pipeline **without** the pre-gate (member bots are always candidates; `agentIsSharedWithUser` handles non-member directory agents), and its comments say *"Mirrors desktop's `shouldHideAgentFromMentions`"*. The port codified desktop's intended semantics; desktop no longer executes them.
- **Coverage has a hole exactly here.** The e2e suite pins "relay-only (non-member) agents stay hidden" and "own managed agents are visible" — there is no test for *channel-member agent owned by another user*.
- **The typed-mention fallback is also starved.** `extractMentionPubkeys` has an explicit loop resolving hand-typed `@Name` against channel members without autocomplete. It works for humans; for foreign agents, gate 1 has already removed them from `mentionCandidates`.

Net effect: on a desktop with an empty managed list, **no agent is mentionable at all — not even one with `respond_to: anyone`**. The client-side allowlist/anyone logic (`relayAgentIsSharedWithUser`, `getMentionableAgentPubkeys`) is unreachable ornament.

`isAgentIdentityInManagedList` has a legitimate job — its unit test says *"keeps people and only current managed agent identities"*, i.e. dropping stale/ghost agent identities surfaced by discovery/search. The bug is only that it also swallows relay-confirmed **channel members**.

## Fix

Exempt channel members from gate 1 so that gate 2 — which already encodes the correct member policy — actually decides:

```ts
if (
  candidate.isMember !== true &&
  !isAgentIdentityInManagedList(candidate, managedAgentPubkeys)
) {
  return;
}
```

Ghost-guard semantics are preserved for non-members (discovery, directory, global search), so the pinned "relay-only agents stay hidden" behavior is unchanged.

PR with the fix and an e2e regression test (channel-member bot + empty managed list ⇒ visible in `@`, mention resolves to a `p` tag on the wire): #TBD

## Repro

1. User A deploys an agent and adds it to a shared channel (member role `bot`).
2. User B (no locally managed agents) opens the channel on **desktop** and types `@`.
3. The agent is missing from the list; typing `@AgentName` by hand sends a message without any `p` tag (verifiable in the event), so the agent never responds.
4. Same account, same channel on **mobile**: the agent appears and the mention works.

NachoG2000:
Fixes #3671

## Problem

On desktop, an agent owned by another user is never mentionable — even as a member of the same channel, even when the viewer is on its `respond_to` allowlist. The managed-list guard in `useMentions.addCandidate` runs before `shouldHideAgentFromMentions` and drops every agent identity not in the **local** managed list. Since `getMentionableAgentPubkeys` seeds with `managedAgentPubkeys`, every agent surviving the guard is already "invocable", so `shouldHideAgentFromMentions` returns `false` on its first line, always — its documented member branch ("Option B": member + unknown invocability ⇒ show) is unreachable dead code. The mobile client implements the same pipeline without the pre-gate and behaves correctly; see the issue for the full analysis.

## Change

- `useMentions.ts`: the managed-list ghost-guard now only applies to **non-member** candidates. Member-agent visibility is decided by `shouldHideAgentFromMentions`, which already encodes the intended policy (invocable ⇒ show; directory-present-but-excluding ⇒ hide; unknown ⇒ show).
- `e2eBridge.ts` / `tests/helpers/bridge.ts`: new `channelBotMembers` mock seed — a channel member with role `bot` that is deliberately absent from the managed list, the kind:10100 directory, and profiles, i.e. exactly what another owner's agent looks like from this install.
- `mentions.spec.ts`: new regression test — with an empty managed list, a foreign member bot appears in the `@` dropdown (with the agent icon) and the signed event carries `["p", <agent>]`, pinning the wire-level behavior. Verified to fail without the src change and pass with it.
- One existing expectation updated: the prioritization test pinned `alice` — a **channel member** and directory agent with `respond_to: "anyone"` — as hidden. That expectation captured the regression itself (she is invocable and a member); she is now asserted visible.

Not changed: `MembersSidebar` add-member search keeps the unfiltered guard — its candidates are non-members by construction. The pinned "relay-only agents stay hidden from channel mentions even when allowlisted" behavior is unchanged (those candidates are non-members).

## Testing

- `biome check` clean on the four touched files; `tsc --noEmit` clean.
- Desktop unit tests: 3781 pass, 0 fail.
- E2E (`mentions.spec.ts`, smoke project): all tests pass with this change except "groups member additions and joins with hidden names in the standard tooltip", which is flaky independently of this branch — it also fails intermittently (1 in 3 repeats) on unmodified `origin/main` on the same machine, and touches join system messages, not mention autocomplete.
- The new regression test was verified in both directions: it fails with the src change reverted and passes with it.

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -246,7 +246,16 @@ export function useMentions(
       if (isArchivedDiscovery(pubkey)) {
         return;
       }
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
+      // The managed-list guard only applies to NON-member identities: it
+      // exists to drop stale/ghost agent identities surfaced by discovery,
+      // but an agent owned by ANOTHER user is never in this install's
+      // managed list, and as a relay-confirmed channel member it must stay
+      // mentionable. Member-agent visibility is owned by
+      // shouldHideAgentFromMentions below (invocable / directory-exclusion).
+      if (
+        candidate.isMember !== true &&
+        !isAgentIdentityInManagedList(candidate, managedAgentPubkeys)
+      ) {
         return;
       }
       if (
--- a/desktop/src/testing/e2eBridge.ts
+++ b/desktop/src/testing/e2eBridge.ts
@@ -96,6 +96,16 @@ type MockManagedAgentRuntimeSeed = {
   lifecycle?: MockManagedAgentRuntimeRow["lifecycle"];
 };
 
+/** A channel member with role "bot" that is NOT in the local managed-agent
+ *  list nor the kind:10100 relay agent directory — models an agent owned by
+ *  another community member, as seen from this install. */
+type MockChannelBotMemberSeed = {
+  pubkey: string;
+  name: string;
+  channelNames?: string[];
+  channelIds?: string[];
+};
+
 type MockRelayAgentSeed = {
   pubkey: string;
   name: string;
@@ -225,6 +235,9 @@ type E2eConfig = {
       mcp?: MockCommandAvailability;
     };
     managedAgents?: MockManagedAgentSeed[];
+    /** Channel members with role "bot" owned by another user: never in the
+     *  managed list, never in the kind:10100 directory. */
+    channelBotMembers?: MockChannelBotMemberSeed[];
     /** Per agent+relay runtime rows for the pair-scoped lifecycle commands
      *  (`list/start/stop/restart_managed_agent_runtime`). */
     managedAgentRuntimes?: MockManagedAgentRuntimeSeed[];
@@ -2123,6 +2136,35 @@ function resetMockManagedAgents(config?: E2eConfig) {
       is_agent: true,
       has_profile_event: true,
     });
+    for (const channel of mockChannels) {
+      const isSeedChannel =
+        seed.channelIds?.includes(channel.id) ||
+        seed.channelNames?.includes(channel.name);
+      if (
+        !isSeedChannel ||
+        channel.members.some((member) => member.pubkey === seed.pubkey)
+      ) {
+        continue;
+      }
+
+      channel.members.push({
+        pubkey: seed.pubkey,
+        role: "bot",
+        is_agent: true,
+        joined_at: new Date().toISOString(),
+        display_name: seed.name,
+      });
+      syncMockChannel(channel);
+      touchMockChannel(channel);
+    }
+  }
+
+  // Foreign bot members: channel membership only — deliberately NOT added to
+  // mockManagedAgents, mockRelayAgents, or mockProfiles, so the client sees
+  // exactly what it sees for an agent owned by another community member.
+  for (const seed of config?.mock?.channelBotMembers ?? []) {
+    applyMockDisplayName(seed.pubkey, seed.name);
+    mockAgentPubkeys.add(seed.pubkey);
     for (const channel of mockChannels) {
       const isSeedChannel =
         seed.channelIds?.includes(channel.id) ||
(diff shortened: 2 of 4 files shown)


ScaleLeanChris:
Tested head `ff800f431832d2149483941a452cdaf4b29f6444` on its exact base and cherry-picked onto today's `main` at `7fb008f9347b933b9a1da20a7afb070912b430e8`.

Verified in both lanes:

- Cherry-pick onto current `main`: clean.
- TypeScript typecheck: pass.
- Desktop unit tests: 3,781 passed on the exact head; 3,799 passed on current `main`.
- E2E build: pass.
- Full `mentions.spec.ts` smoke project: 50 passed on the exact head and 50 passed on current `main`.
- The new foreign-member test verifies both autocomplete visibility and the signed event's recipient `p` tag.

The behavior is correct. The foreign agent is no longer "Invisible."

One repository gate blocks the PR:

```text
src/features/messages/lib/useMentions.ts: 1000 -> 1009 (+9) lines (allowed 1000)
```

`pnpm -C desktop check` fails on both the exact PR head and the current-main integration. The two Biome messages are informational suggestions in an unrelated existing test; the file-size ratchet is the blocking result.

Suggested fix: move the member-or-managed policy into `agentAutocompleteEligibility.ts` and keep the existing three-line guard shape in `useMentions.ts`. That preserves the 1,000-line ceiling and makes the policy independently unit-testable. Shortening the new comment alone is unlikely to recover all nine net lines cleanly.

Once the size gate is fixed, I found no behavioral or current-main compatibility blocker.

ziggygram:
Upsetting issue! hope this gets merged soon!

Chessing234:
this matches #4489 — dropping the local-managed-list gate for channel-member agents is the right direction. please keep a short note in the pr why local-only filtering was wrong for multi-install channels.

Issue #3663, opened 2026-07-30
set-add-policy replaces the full kind:10100 record, hiding the agent from mention autocomplete


gabrielgcarmona:
**Describe the bug**

`buzz channels set-add-policy` silently removes an agent from Desktop's mention
autocomplete.

`kind:10100` (agent profile) is replaceable — `crates/buzz-core/src/kind.rs:783`
asserts `is_replaceable(KIND_AGENT_PROFILE)`. The CLI publishes it with a single
field (`crates/buzz-cli/src/commands/channels.rs:1035`):

```rust
let content = serde_json::json!({ "channel_add_policy": policy }).to_string();
```

Because the event is replaceable, this **replaces** the agent's full directory
record rather than merging into it. Every other field the record carried —
`respond_to`, `respond_to_allowlist`, `channel_ids`, `name`, `agent_type` — is
gone.

Nothing surfaces the loss:

- The relay accepts it. `handle_agent_profile` only reads `channel_add_policy`
  (`crates/buzz-relay/src/handlers/side_effects.rs:1172`) and ignores the rest.
- Desktop hides the agent without an error. In
  `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts:21`,
  `relayAgentIsSharedWithUser` reads `agent.respondTo`, which is now `undefined`,
  so neither branch matches and the agent is not invocable. Then
  `shouldHideAgentFromMentions` treats "in the directory but not invocable" as an
  explicit exclusion signal and hides it.

The result is counter-intuitive: **an agent with no `kind:10100` shows up; an
agent with a partial `kind:10100` disappears.** The comment at line 88 —
`Unknown invocability (not in directory) => show` — is what makes the
never-registered agent work.

Downstream, `@name` typed in a channel stays plain text with no `p` tag, so the
relay never delivers the event to the ACP harness (`subscribe=Mentions` filters
on `p`), and the agent never wakes — with no log line on either side.

**Steps to reproduce**

1. Register a CLI-driven ACP agent (own keypair, `buzz-admin add-member`), give
   it a profile with `buzz users set-profile`, and add it to a channel.
2. Confirm `@name` resolves in Desktop's composer and the agent replies.
3. Run `buzz channels set-add-policy --policy owner_only` as that agent.
4. Type `@name` in the same channel.

The agent no longer appears in the autocomplete. The mention is sent as plain
text and the agent never responds. Reverting with `--policy anyone` does **not**
restore it — that publishes another single-field record, so the agent stays in
[...]

PR #3676, opened 2026-07-30
fix(desktop): decide mention eligibility for external agents from relay proofs

Shipitrealgood:
## Problem

External / remote agents cannot be @-mentioned from Desktop even when they are channel members with valid NIP-OA and kind:10100.

#2149 made the local managed-agent list a hard pre-gate for agent-class identities. Directory eligibility (`respond_to` / allowlist) never runs for them.

Sharp edge (#2987): an unattested agent can still show as a person and be mentionable; **correctly registering it as an agent is what hides it.**

## Fix

Replace the managed-list pre-gate with one predicate built from relay proofs:

```
mentionable =
  locallyManaged(agent)                    // native path (membership not required)
  || (
       member(agent, channel)
       && (
            respond_to == anyone
         || respond_to == allowlist && viewer ∈ allowlist
         || respond_to == owner-only && verifiedOwner(agent) == viewer
       )
     )
```

- Local managed store is **one proof**, not the gatekeeper.
- No kind:10100 declaration → hidden from everyone **except** the NIP-OA-verified owner (owner-on-another-device; harness already admits the owner).
- Non-member externals stay hidden even if `anyone` (add-via-mention / `channel_add_policy` is out of scope — see #3864).
- Liveness is not required (offline is a queue; stopped managed agents are already offered).

## Out of scope

| Follow-up | Why separate |
|-----------|----------------|
| Add-member search | Same managed-list cut; fix is `channel_add_policy` — **#3864** |
| Desktop directory writer | Native agents still don't publish full 10100; third-party cross-machine half of #3125 / #2508 needs a writer (#3708+) |
| Thin CLI `set-add-policy` clobber | **#3663** |

## Related

Closes the @-mention core of **#2987**. Helps cross-machine owner cases **#2349** / **#3277**. Same root as open PRs #3110 / #3448 / #3609 — this variant keeps undeclared agents fail-closed (except verified owner), requires membership for externals, and supports `owner-only` via verified OA.

## Test plan

- [x] Unit matrix: anyone / allowlist / owner-only / no declaration / non-member / managed / pubkey normalize  
- [x] `pnpm test`, `tsc --noEmit`, biome on touched paths  
- [x] e2e `mentions.spec.ts` — external agent (alice, `anyone`, shared channel) **visible** (was the bug pin; now regression)  
- [x] Live: headless OA + 10100 agents appear in @, event carries `p`, harness responds

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -162,84 +162,161 @@ test("isAgentIdentityInManagedList: keeps people and only current managed agent 
   );
 });
 
+function hideArgs(overrides = {}) {
+  const {
+    candidate = {},
+    currentPubkey = CURRENT_PUBKEY,
+    managedAgentPubkeys = new Set(),
+    relayAgentPolicies = new Map(),
+  } = overrides;
+  return {
+    candidate: {
+      isAgent: true,
+      isMember: true,
+      ownerPubkey: null,
+      pubkey: PUB_A,
+      ...candidate,
+    },
+    currentPubkey,
+    managedAgentPubkeys,
+    relayAgentPolicies,
+  };
+}
+
+function policy(respondTo, respondToAllowlist = []) {
+  return new Map([[PUB_A, { respondTo, respondToAllowlist }]]);
+}
+
 test("shouldHideAgentFromMentions: never hides non-agents", () => {
   assert.equal(
-    shouldHideAgentFromMentions({
-      isAgent: false,
-      isMember: false,
-      pubkey: PUB_A,
-      mentionableAgentPubkeys: new Set(),
-      directoryAgentPubkeys: new Set([PUB_A]),
-    }),
+    shouldHideAgentFromMentions(
+      hideArgs({ candidate: { isAgent: false, isMember: false } }),
+    ),
     false,
   );
 });
 
-test("shouldHideAgentFromMentions: shows invocable agents even when non-member", () => {
+test("shouldHideAgentFromMentions: always offers managed agents, member or not", () => {
+  // The local record is the desktop's own proof the agent will answer; it
+  // also covers the non-member auto-add flow for managed agents.
+  for (const isMember of [true, false]) {
+    assert.equal(
+      shouldHideAgentFromMentions(
+        hideArgs({
+          candidate: { isMember },
+          managedAgentPubkeys: new Set([PUB_A]),
+        }),
+      ),
+      false,
+    );
+  }
+});
+
+test("shouldHideAgentFromMentions: offers external member agents declaring respond-to anyone", () => {
   assert.equal(
-    shouldHideAgentFromMentions({
-      isAgent: true,
-      isMember: false,
-      pubkey: PUB_A,
-      mentionableAgentPubkeys: new Set([PUB_A]),
-      directoryAgentPubkeys: new Set([PUB_A]),
-    }),
+    shouldHideAgentFromMentions(
+      hideArgs({ relayAgentPolicies: policy("anyone") }),
+    ),
     false,
   );
 });
 
-test("shouldHideAgentFromMentions: hides non-member non-invocable agents", () => {
+test("shouldHideAgentFromMentions: allowlist admits only listed users", () => {
   assert.equal(
-    shouldHideAgentFromMentions({
-      isAgent: true,
-      isMember: false,
-      pubkey: PUB_A,
-      mentionableAgentPubkeys: new Set(),
-      directoryAgentPubkeys: new Set(),
-    }),
+    shouldHideAgentFromMentions(
+      hideArgs({ relayAgentPolicies: policy("allowlist", [CURRENT_PUBKEY]) }),
+    ),
+    false,
+  );
+  assert.equal(
+    shouldHideAgentFromMentions(
+      hideArgs({ relayAgentPolicies: policy("allowlist", [OWNER_PUBKEY]) }),
+    ),
     true,
   );
 });
 
-test("shouldHideAgentFromMentions: hides member agents with an explicit not-invocable directory entry (Fizz)", () => {
+test("shouldHideAgentFromMentions: owner-only admits the verified owner", () => {
+  // ownerPubkey is the NIP-OA-verified value the "managed by" surface
+  // renders — external agents owned by the current user mention like
+  // managed ones.
   assert.equal(
-    shouldHideAgentFromMentions({
-      isAgent: true,
-      isMember: true,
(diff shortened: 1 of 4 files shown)


Shipitrealgood:
**Scope note (post-#3178 / directory clarification)**

This PR only changes mention *eligibility* over proofs the Desktop already loads: local managed set, kind:10100 policy, and kind:0 NIP-OA `ownerPubkey`.

It does **not** read kind:30177 (owner instance catalog — see #3178 for CLI owner-scoped lookup) and does not change mint/write of 10100. Native Desktop still may lack a full directory write; third-party mention of those agents remains limited until a writer exists. External agents with full 10100 + membership are the multi-client happy path this gate unlocks.

Ownership for the owner-only / no-declaration paths is the same NIP-OA class #3178 enforces on CLI; we consume Desktop’s existing `ownerPubkey` projection and do not re-verify conditions in TS.

shawnhansen:
Independent validation from a live Hermes/Buzz setup:

- Hermes publishes a NIP-OA-attested owner and a kind:10100 profile with `respond_to: "owner-only"`.
- Manually typed `@Hermes` routes correctly, but Desktop autocomplete drops the agent before policy eligibility because it is not in the local managed-agent list.
- Requiring verified ownership **and** current channel membership restores autocomplete for the owner without changing gateway behavior.

Two fail-closed edge cases were important during independent review:

1. An unrelated unmanaged channel-member agent with unknown policy must remain hidden.
2. An owner-attested unmanaged agent that is **not** a channel member must remain hidden even if its pubkey is otherwise in the mentionable set.

I verified this PR's stated predicate covers both boundaries: external agents require channel membership plus an admissible relay policy, and `owner-only` requires the normalized NIP-OA-verified owner to match the viewer.

Local validation of the narrow reproduction/fix path passed the complete `just ci` gate, including desktop and mobile suites, plus a policy truth table covering normalized owner match, mismatch, missing/empty owner, unrelated unmanaged member, and owned non-member-but-mentionable.

This PR is the closest and most complete existing fix for the Hermes case; no additional duplicate PR opened.

pdurlej:
Independent focused verification of the relay-only mention path against Buzz Desktop v0.5.2:

- full Desktop unit suite: 3,777 passed;
- TypeScript typecheck, repository checks, and E2E build passed;
- focused Playwright coverage: 2/2 passed.

The Playwright cases exercise the boundary this PR intends to preserve:

1. an allowlisted relay-directory agent outside the current channel remains absent from mention autocomplete;
2. the same relay-only agent becomes mentionable after joining the current channel;
3. the sent event contains exactly one `p` tag for that agent;
4. the `start_managed_agent` command count does not change when the message is sent.

That last assertion is worth retaining in the upstream regression test: eligibility for a relay-hosted agent must not imply local lifecycle ownership or start a duplicate harness.

This is supporting evidence for #3676, not a competing implementation.

Shipitrealgood:
@wpfleger96 Can we get a look at this.

daguiler76:
I reproduced the owner-side failure described in #3204 on a real Buzz 0.5.2 install: the agent is relay-confirmed as a channel member and can receive DMs, but it is absent from both the Agents surface and the `@` picker because the picker only trusts the local managed-agent projection.

I checked out this PR at `6b591523d9fc7220c4b18f102d314857ae306c26` and validated the revised eligibility path:

- `cd desktop && pnpm test`: **3,784/3,784 passing**
- `cd desktop && pnpm typecheck`: **passing**
- The new unit cases cover the reproduced shape: an external/channel-member agent with verified ownership remains mentionable even without a local managed record or directory declaration.

One CI blocker remains in the touched code:

```text
src/features/messages/lib/useMentions.ts:244:35
lint/correctness/useExhaustiveDependencies
This hook specifies more dependencies than necessary: mentionableAgentPubkeys.
```

`mentionableAgentPubkeys` is no longer referenced inside `mentionCandidates` after this refactor. Removing it from the dependency array makes the focused Biome check pass; I verified that one-line cleanup locally together with the full test suite and typecheck above.

The sanitized live-install reproduction is documented at https://github.com/block/buzz/issues/3204#issuecomment-5140259539.

daguiler76:
One additional regression check after installing the Playwright browser:

```text
pnpm exec playwright test tests/e2e/mentions.spec.ts \
  --project=smoke \
  --grep '@ trigger prioritizes channel members'
```

The existing test initially fails because it still expects `alice` to be absent. In the mock bridge, Alice is an in-channel relay agent with `respond_to: "anyone"`, so the new behavior correctly includes her. Updating that assertion to `toBeVisible()` and including `aliceIndex < fizzIndex` in the member-priority checks makes the test pass.

Validated local delta on top of this PR:

```diff
-  await expect(dropdown.getByText("alice")).toHaveCount(0);
+  await expect(dropdown.getByText("alice")).toBeVisible();
```

along with the analogous `aliceIndex` ordering assertion. This gives the PR an end-to-end regression for the same eligibility change covered by the new unit tests.

daguiler76:
I packaged the two validated follow-ups as an apply-ready PR against your head branch: https://github.com/Shipitrealgood/buzz/pull/1. It removes the stale hook dependency and updates the focused E2E expectation/order assertion. Full Desktop tests (3,784), typecheck, focused Biome, and focused Playwright smoke all pass at `c2136e9b7c877e7cc48da95729a0a82df9a028d7`.

Shipitrealgood:
@daguiler76 Both follow-ups are merged. Your PR landed on the head branch as a fast-forward, so the commit is yours as validated — same SHA (`c2136e9`), authorship and sign-off intact.

Before merging we verified independently: full diff review, desktop suite 3,784/3,784, typecheck, and the focused Biome check on the touched file — all clean. We'd reached the same dep-array conclusion reviewing the refactor, so your patch matched what we expected to write. On the e2e change: confirmed against the mock bridge fixture — alice (`respond_to: "anyone"`, shared channel) models exactly the external-agent shape this PR un-hides, so the old `toHaveCount(0)` was pinning the #2987 behavior this PR exists to fix; your update turns that spec into an end-to-end regression test for it. Appreciated on both counts.

One scope note on your latest #3204 reproduction (the single-install case — persona-definition rows offered while running channel-member instances are omitted): that's a distinct candidate-assembly defect at the persona/instance coalescing seam, which predates this PR and is deliberately untouched by it. This PR covers the eligibility-gate half of #3204 (your remote-machine case); the join defect is worth its own track so it doesn't get lost when the eligibility work lands.

The integration branch for full-fix testing (`test/external-agent-eligibility` = current `main` + this PR + #3864) has been rebuilt on the new tip.

lolamathematician:
I reproduced the issue on windows see https://github.com/block/buzz/issues/3277#issuecomment-5148662049

Shipitrealgood:
closing.  Believe covered in #4913

PR #3678, opened 2026-07-30
feat(tracing): add PostgreSQL tracing spans

ThePumpingLemma:
## Why
Expose PostgreSQL datastore latency within existing request traces so slow logical database operations can be identified without recording tenant data or query arguments.

## What
- Add client spans around logical PostgreSQL operations across the database facade, search, audit, replica fencing, and command persistence
- Use a dedicated `buzz_datastore` target and `db.system.name = "postgresql"` for filtering and backend classification
- Exclude health-check database calls and scrub raw identifiers and errors from newly traced paths

## Risk Assessment
Medium — this instruments frequently used datastore paths and increases trace volume when enabled, but does not change SQL execution or datastore behavior. Existing OpenTelemetry filtering controls export.

## References
- Pre-push clippy and fast unit-test hooks passed

Generated with Amp

### DIFF
--- a/crates/buzz-audit/src/service.rs
+++ b/crates/buzz-audit/src/service.rs
@@ -49,7 +49,7 @@ impl AuditService {
     /// Serialized per-community via `pg_advisory_lock`. Postgres advisory locks
     /// are session-scoped, so we acquire before the transaction and release
     /// after commit (or on any error path).
-    #[instrument(skip(self, entry), fields(action = %entry.action))]
+    #[instrument(target = "buzz_datastore", name = "audit_log", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
     pub async fn log(&self, entry: NewAuditEntry) -> Result<AuditEntry, AuditError> {
         let mut conn = self.pool.acquire().await?;
 
@@ -125,7 +125,7 @@ impl AuditService {
 
         audit_entry.hash = compute_hash(&audit_entry)?.to_vec();
 
-        debug!(seq, "writing audit entry");
+        debug!("writing audit entry");
 
         sqlx::query(
             r#"
@@ -156,7 +156,7 @@ impl AuditService {
     /// Reads exactly that community's chain — it can never observe another
     /// community's entries or head. Returns `Ok(false)` if the range is empty,
     /// `Ok(true)` if the segment is internally consistent.
-    #[instrument(skip(self))]
+    #[instrument(target = "buzz_datastore", name = "audit_verify_chain", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
     pub async fn verify_chain(
         &self,
         community: CommunityId,
@@ -208,7 +208,7 @@ impl AuditService {
     /// Returns up to `limit` entries from one community's chain starting at
     /// `from_seq`, ordered by sequence number. Scoped to `community` — never
     /// returns another community's rows.
-    #[instrument(skip(self))]
+    #[instrument(target = "buzz_datastore", name = "audit_get_entries", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
     pub async fn get_entries(
         &self,
         community: CommunityId,
@@ -274,7 +274,7 @@ mod tests {
 
     async fn test_pool() -> Option<PgPool> {
         let url = std::env::var("DATABASE_URL")
-            .unwrap_or_else(|_| "postgres://buzz:buzz_dev@localhost:5432/buzz".into());
+            .unwrap_or_else(|_| "postgres://buzz:buzz_dev@localhost:5432/buzz".into()); // sadscan:disable np.postgres.1 -- local test fixture
         PgPool::connect(&url).await.ok()
     }
 
--- a/crates/buzz-audit/src/service.rs
+++ b/crates/buzz-audit/src/service.rs
@@ -1,10 +1,11 @@
 use chrono::{DateTime, Utc};
 use futures_util::FutureExt as _;
 use sqlx::{Acquire, PgPool, Row};
-use tracing::{debug, instrument, warn};
+use tracing::{debug, warn};
 use uuid::Uuid;
 
 use buzz_core::CommunityId;
+use buzz_datastore_tracing::datastore_span;
 
 use crate::{
     action::AuditAction,
@@ -49,7 +50,11 @@ impl AuditService {
     /// Serialized per-community via `pg_advisory_lock`. Postgres advisory locks
     /// are session-scoped, so we acquire before the transaction and release
     /// after commit (or on any error path).
-    #[instrument(target = "buzz_datastore", name = "audit_log", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(
+        name = "audit_log",
+        system = "postgresql",
+        fields(action = %entry.action)
+    )]
     pub async fn log(&self, entry: NewAuditEntry) -> Result<AuditEntry, AuditError> {
         let mut conn = self.pool.acquire().await?;
 
@@ -125,7 +130,7 @@ impl AuditService {
 
         audit_entry.hash = compute_hash(&audit_entry)?.to_vec();
 
-        debug!("writing audit entry");
+        debug!(seq, "writing audit entry");
 
         sqlx::query(
             r#"
@@ -156,7 +161,11 @@ impl AuditService {
     /// Reads exactly that community's chain — it can never observe another
     /// community's entries or head. Returns `Ok(false)` if the range is empty,
     /// `Ok(true)` if the segment is internally consistent.
-    #[instrument(target = "buzz_datastore", name = "audit_verify_chain", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(
+        name = "audit_verify_chain",
+        system = "postgresql",
+        fields(from_seq = from_seq, to_seq = to_seq)
+    )]
     pub async fn verify_chain(
         &self,
         community: CommunityId,
@@ -208,7 +217,11 @@ impl AuditService {
     /// Returns up to `limit` entries from one community's chain starting at
     /// `from_seq`, ordered by sequence number. Scoped to `community` — never
     /// returns another community's rows.
-    #[instrument(target = "buzz_datastore", name = "audit_get_entries", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(
+        name = "audit_get_entries",
+        system = "postgresql",
+        fields(from_seq = from_seq, limit = limit)
+    )]
     pub async fn get_entries(
         &self,
         community: CommunityId,
--- a/crates/buzz-db/src/event.rs
+++ b/crates/buzz-db/src/event.rs
@@ -601,8 +601,8 @@ pub(crate) fn row_to_stored_event(row: sqlx::postgres::PgRow) -> Result<Option<S
     // Avoid the Value → String → parse round-trip: deserialize directly from the Value.
     let event: nostr::Event = match serde_json::from_value(event_json) {
         Ok(e) => e,
-        Err(_e) => {
-            tracing::warn!("failed to reconstruct event from DB row");
+        Err(e) => {
+            tracing::warn!("failed to reconstruct event from DB row: {e}");
             return Ok(None);
         }
     };
--- a/crates/buzz-db/src/event.rs
+++ b/crates/buzz-db/src/event.rs
@@ -601,8 +601,8 @@ pub(crate) fn row_to_stored_event(row: sqlx::postgres::PgRow) -> Result<Option<S
     // Avoid the Value → String → parse round-trip: deserialize directly from the Value.
     let event: nostr::Event = match serde_json::from_value(event_json) {
         Ok(e) => e,
-        Err(e) => {
-            tracing::warn!("failed to reconstruct event from DB row: {e}");
+        Err(_e) => {
+            tracing::warn!("failed to reconstruct event from DB row");
             return Ok(None);
         }
     };
--- a/crates/buzz-db/src/lib.rs
+++ b/crates/buzz-db/src/lib.rs
@@ -57,6 +57,7 @@ pub mod workflow;
 pub use error::{DbError, Result};
 pub use event::{EventQuery, ReactionEventInsertOutcome, DEFAULT_MAX_PAGE_LIMIT};
 
+use buzz_datastore_tracing::datastore_span;
 use chrono::{DateTime, Utc};
 use sqlx::postgres::{PgConnection, PgPoolOptions};
 use sqlx::{Connection, PgPool, QueryBuilder, Row};
@@ -130,7 +131,11 @@ pub async fn insert_mentions(
         .into_iter()
         .filter(|pk| {
             if pk.len() != 64 || !pk.chars().all(|c| c.is_ascii_hexdigit()) {
-                tracing::debug!("skipping malformed p-tag in insert_mentions");
+                tracing::debug!(
+                    event_id = %event.id,
+                    invalid_ptag = pk,
+                    "skipping malformed p-tag in insert_mentions"
+                );
                 false
             } else {
                 true
@@ -248,7 +253,7 @@ impl ReadSession {
     /// the degraded follow-up can only observe *more* than the proof-time
     /// snapshot, never less — fresher aux rows, the same failure semantics
     /// as a request that routed to the writer to begin with.
-    #[tracing::instrument(target = "buzz_datastore", name = "read_session_query_events", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(name = "read_session_query_events", system = "postgresql")]
     pub async fn query_events(&mut self, q: &EventQuery) -> Result<Vec<StoredEvent>> {
         let degraded = match &mut self.inner {
             ReadSessionInner::Replica { tx, writer } => {
@@ -488,7 +493,6 @@ impl UsageMetricsLeader {
     ///
     /// Bounded to 5 seconds — a blackholed connection (no RST) would otherwise
     /// stall the entire poller tick until the OS TCP timeout.
-    #[tracing::instrument(target = "buzz_datastore", name = "usage_metrics_leader_is_live", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
     pub async fn is_live(&mut self) -> bool {
         tokio::time::timeout(std::time::Duration::from_secs(5), self.connection.ping())
             .await
@@ -1007,7 +1011,7 @@ impl Db {
     }
 
     /// Run pending database migrations.
-    #[tracing::instrument(target = "buzz_datastore", name = "migrate", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(name = "migrate", system = "postgresql")]
     pub async fn migrate(&self) -> Result<()> {
         migration::run_migrations(&self.pool).await
     }
@@ -1052,7 +1056,7 @@ impl Db {
     /// detached from the shared pool so a stable leader neither returns a locked
     /// session to other callers nor permanently consumes a pool slot. Dropping the
     /// guard closes the connection and releases the session-scoped lock.
-    #[tracing::instrument(target = "buzz_datastore", name = "try_lock_usage_metrics", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(name = "try_lock_usage_metrics", system = "postgresql")]
     pub async fn try_lock_usage_metrics(
         &self,
         lock_key: i64,
@@ -1073,7 +1077,7 @@ impl Db {
 
     /// List reports for the deployment-global read-only admin plane.
     #[allow(clippy::too_many_arguments)]
-    #[tracing::instrument(target = "buzz_datastore", name = "admin_list_reports", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(name = "admin_list_reports", system = "postgresql")]
     pub async fn admin_list_reports(
         &self,
         community_id: Option<Uuid>,
@@ -1100,7 +1104,7 @@ impl Db {
     }
 
     /// Fetch one report for the deployment-global read-only admin plane.
-    #[tracing::instrument(target = "buzz_datastore", name = "admin_get_report", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
+    #[datastore_span(name = "admin_get_report", system = "postgresql")]
     pub async fn admin_get_report(
         &self,
         id: Uuid,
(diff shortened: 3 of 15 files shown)


ThePumpingLemma:
Some additional research on more detailed alternatives:

# Buzz PostgreSQL trace correlation options

## Decision being considered

Buzz currently has logical datastore spans around operations such as
`insert_event`, `query_events`, and `audit_log`. These show how much time a Buzz
operation spends in the datastore, but they do not identify or time each SQL
statement independently.

Two possible ways to add query-level visibility are:

1. Create an OpenTelemetry client span around every SQL query.
2. Add SQLCommenter support to SQLx so each sampled query carries W3C trace
   context in a SQL comment.

These approaches solve related but different problems. Query spans describe
application-side execution and appear in an APM flame graph. SQLCommenter sends
trace identity to PostgreSQL so database-side telemetry can be associated with
the application trace.

## Option 1: manually add a span around every query

### Resulting trace shape

```text
ws.event
└── handle_event
    └── insert_event
        ├── postgres.query: INSERT events
        ├── postgres.query: UPDATE thread counters
        └── postgres.query: INSERT mentions
```

Each query span would normally be a `CLIENT` span with attributes such as:

```text
db.system.name = postgresql
db.operation.name = INSERT
db.query.summary = INSERT events
db.query.text = <parameterized SQL, if approved>
```

Bind values, community IDs, event IDs, user data, and other high-cardinality or
sensitive values should not be recorded.

### Advantages

- Gives exact application-side duration and errors for each SQL operation.
- Produces the conventional APM model expected by Datadog and other tracing
  backends.
- Preserves SQLx prepared-statement caching because the SQL sent to PostgreSQL
  is unchanged.
- Naturally nests query work beneath Buzz's logical datastore spans.
- Can include returned-row counts, affected-row counts, and SQLSTATE without
  exposing bind values.
- Works independently of PostgreSQL logging and DBM query sampling.
- Does not require database-side understanding of a propagation format.

### Disadvantages

- Adding spans manually at every call site is a large and error-prone change.
  Buzz has hundreds of `query`, `query_as`, `query_scalar`, and `QueryBuilder`
  executions spread across pools, acquired connections, and transactions.
- New queries can easily be added without instrumentation.
- Every query-site wrapper must correctly handle errors, streaming results,
  cancellation, and early stream drops.
- It creates more spans. A single logical operation can execute several SQL
  statements, and high-traffic requests can therefore generate many child
  spans.
- Capturing or normalizing SQL adds CPU, allocation, telemetry payload, privacy,
  and cardinality costs.
- Query span names and attributes must remain low-cardinality and follow a
  consistent semantic convention.
- Application-side head sampling is needed to avoid constructing and exporting
  query spans for every production request.

### Prior art

- SeaORM has an opt-in `tracing-spans` feature that creates one span per ORM
  operation.
- `sqlx-otel` and `sqlx-tracing` wrap SQLx executors to create query spans
  automatically rather than requiring manual query-site spans.
- SQLx itself sees every query in its driver and emits completion events through
  `QueryLogger`, but those are events rather than independently timed spans.

The strongest implementation would therefore not be literally manual. It would
add a SQLx-native hook or a carefully reviewed executor wrapper that covers
pools, connections, transactions, and streams centrally. The existing public
SQLx wrappers are currently young, have little visible adoption, and target
SQLx 0.8 rather than Buzz's SQLx 0.9. `sqlx-otel` also targets OpenTelemetry
0.31 while Buzz uses 0.32.

## Option 2: add SQLCommenter propagation to SQLx

### Resulting SQL

For a sampled query, SQLx would append the active W3C context:

```sql
SELECT * FROM events WHERE community_id = $1 AND id = $2
/*traceparent='00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01'*/
```

PostgreSQL treats the suffix as a comment, but database logs, activity samples,
and compatible monitoring products can observe it.

### Advantages

- Carries exact trace identity across the application/database boundary.
- Can connect a PostgreSQL-side query sample, wait event, execution plan, or log
  entry to the originating trace when the monitoring system understands the
  SQLCommenter field.
- Uses the vendor-neutral W3C `traceparent` representation.
- OpenTelemetry's database semantic conventions describe SQLCommenter as an
  optional context propagation mechanism.
- A SQLx-level implementation could apply consistently without manually editing
  every query execution.
- The comment need not contain community IDs, user data, bind values, or other
  application identifiers.

### Disadvantages

- OpenTelemetry currently marks database SQLCommenter propagation as
  Development and says it should not be enabled by default.
- Datadog must be verified to consume generic W3C SQLCommenter fields for the
  PostgreSQL DBM path in use. OpenTelemetry standardizing the comment does not
  guarantee a particular vendor correlates it.
- SQLx does not expose a per-statement query rewrite hook, so this cannot be
  implemented cleanly as a pool configuration option today.
- A unique `traceparent` changes the complete SQL string. SQLx keys its prepared
  statement cache by that string, which would otherwise produce effectively one
  cached statement per trace.
- Setting `.persistent(false)` avoids unbounded statement-cache growth, but
  forces sampled queries to be parsed, described, prepared, executed, and then
  closed rather than reusing a cached prepared statement.
- Comments increase wire bytes and can increase PostgreSQL and database-log
  volume.
- Query aggregation and normalization must strip or ignore the dynamic comment.
- SQLCommenter alone does not create an application-side query span. It can
  correlate database telemetry to a trace, but it does not add a timed query bar
  to the APM flame graph unless query spans are also created.
- Tail sampling in a collector happens too late to avoid SQL mutation and
  prepare overhead. Buzz must make the sampling decision before sending SQL.

### Sampling behavior

SQL comments should only be injected when the active span context is valid and
sampled. With application-side parent-based ratio sampling:

| Context | Inject comment | Preserve normal SQLx statement caching |
| --- | --- | --- |
| Sampled | Yes | No, unless SQLx gains specialized support |
| Unsampled | No | Yes |
| Missing or invalid | No | Yes |

This makes application-side sampling the relevant cost control. Collector-only
sampling cannot prevent query rewriting or PostgreSQL parse/prepare work.

### Prior art

Prisma Engines is the strongest Rust implementation. Its Rust query engine
injects sampled W3C `traceparent` comments through its `tokio-postgres`-based
Quaint layer. Prisma added a specialized tracing cache that:

1. Removes the dynamic trace comment when deriving the metadata cache key.
2. Reuses parameter and result metadata for the stable underlying SQL.
3. Sends the comment-bearing SQL for each execution.
4. Avoids retaining a distinct prepared statement for every trace ID.

That is substantially better than simply disabling persistence, but Prisma
controls its PostgreSQL dispatch layer. Reproducing this correctly in Buzz would
likely require an upstream SQLx feature or maintaining changes against SQLx
internals.

## Direct comparison

| Concern | Per-query spans | SQLCommenter in SQLx |
| --- | --- | --- |
| APM flame-graph timing | Yes | No, not by itself |
| PostgreSQL-side trace identity | No | Yes |
| Exact DBM query-sample correlation | Possible through span/query matching | Possible if DBM consumes the comment |
| Changes SQL sent to PostgreSQL | No | Yes |
| Prepared-statement cache impact | None | Severe without specialized handling |
| Application code impact | Large if manual; moderate with an executor wrapper | Small only if implemented inside SQLx; otherwise query construction must be centralized |
| Span volume | One additional span per sampled query | None by itself |
| Query parsing overhead | Attribute parsing/normalization only | Potential PostgreSQL re-prepare for every sampled execution |
| Sampling requirement | Recommended for span/export volume | Required to control SQL/cache overhead |
| Privacy concern | Query text and span attributes | Context stored in SQL/logs; no application data is required |
| Current Rust maturity | Established model; automatic SQLx wrappers are young | One strong Prisma implementation; no general SQLx support |

## Recommendation for Buzz

Do not add either mechanism to the current PostgreSQL datastore-spans PR.

The current logical spans provide useful operation-level latency with a small,
reviewable behavior change and no effect on query execution. SQLx's built-in
`sqlx::query` completion events can provide slow-query details correlated through
Buzz's trace-aware logs without introducing query spans or changing SQL.

If production traces show that logical operation timing is insufficient:

1. Prefer a SQLx-native or executor-level per-query span implementation over
   manually instrumenting every query site.
2. Require compatibility with SQLx 0.9 and Buzz's OpenTelemetry version.
3. Ensure the unsampled path skips SQL cloning, normalization, attribute-vector
   construction, and metrics unless explicitly enabled.
4. Start with parameterized query text disabled or privacy-reviewed, and use
   stable query names/summaries.
5. Benchmark sampled and unsampled request paths before broad rollout.

Consider SQLCommenter later only if all of the following are true:

1. Datadog's PostgreSQL DBM integration is proven to consume the W3C comment in
   Buzz's deployment.
2. SQLx supports trace-aware execution without fragmenting its prepared
   statement cache, preferably through an upstream feature modeled on Prisma's
   metadata-cache design.
3. Buzz uses application-side sampling and injects comments only for sampled
   contexts.
4. Database query grouping, logging volume, privacy, and parse/prepare overhead
   have been measured in staging.

In short: logical datastore spans now; SQLx slow-query events for diagnosis;
automatic query spans only if the additional APM detail proves necessary; and
SQLCommenter only after Datadog compatibility and prepared-statement behavior
are solved.

## References

- OpenTelemetry database client spans and SQLCommenter:
  https://opentelemetry.io/docs/specs/semconv/db/database-spans/#sql-commenter
- SQLCommenter specification: https://google.github.io/sqlcommenter/spec/
- Datadog OpenTelemetry trace/DBM correlation:
  https://docs.datadoghq.com/opentelemetry/correlate/dbm_and_traces/
- Prisma Engines: https://github.com/prisma/prisma-engines
- SeaORM tracing spans:
  https://github.com/SeaQL/sea-orm/blob/master/src/database/tracing_spans.rs
- sqlx-otel: https://github.com/chmodas/sqlx-otel
- sqlx-tracing: https://github.com/jdrouet/sqlx-tracing
- SQLx QueryLogger:
  https://github.com/launchbadge/sqlx/blob/main/sqlx-core/src/logger.rs

crates/buzz-db/src/lib.rs
```diff
@@ -1615,6 +1642,7 @@ impl Db {
     }
 
     /// Inserts an event. Returns `(StoredEvent, was_inserted)` — `false` on duplicate.
+    #[tracing::instrument(target = "buzz_datastore", name = "insert_event", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
```
wpfleger96: 🤖 None of the new `#[instrument]` annotations record failure status — an `Err` propagated by `?` closes the span clean, so a failed `insert_event` exports as an ordinary completed span. That defeats the main purpose here (spotting slow/broken datastore ops in traces).

I think the fix is to record `otel.status_code = error` plus a low-cardinality error class (error enum variant or SQLSTATE — never the raw SQL error text) at the operation boundary, and add an in-memory-exporter test proving a failing op closes as an error span with no sensitive attributes.

Since that makes each of the 211 annotations longer, a small local `datastore_instrument`-style macro would keep this from tripling the diff — probably worth doing now rather than later.

crates/buzz-search/src/query.rs
```diff
@@ -297,7 +298,16 @@ pub async fn search(pool: &PgPool, query: &SearchQuery) -> Result<SearchResult,
     qb.push(" OFFSET ");
     qb.push_bind(offset);
 
-    let rows = qb.build().fetch_all(pool).await?;
+    let rows = qb
+        .build()
+        .fetch_all(pool)
+        .instrument(tracing::info_span!(
```
wpfleger96: 🤖 Same failure-status gap as the `#[instrument]` annotations (see my comment on `insert_event` in `buzz-db/src/lib.rs`) — this manual span also closes clean when `fetch_all` errors, and it needs the same sanitized error recording.

crates/buzz-db/src/event.rs
```diff
@@ -589,8 +589,8 @@ pub(crate) fn row_to_stored_event(row: sqlx::postgres::PgRow) -> Result<Option<S
     // Avoid the Value → String → parse round-trip: deserialize directly from the Value.
     let event: nostr::Event = match serde_json::from_value(event_json) {
         Ok(e) => e,
-        Err(e) => {
-            tracing::warn!("failed to reconstruct event from DB row: {e}");
+        Err(_e) => {
+            tracing::warn!("failed to reconstruct event from DB row");
```
wpfleger96: 🤖 This drops the serde error entirely, which makes “why won’t this row decode” undebuggable. I'd keep at least the error's classification here — the blanket scrub isn't needed for the new spans since `skip_all` already keeps arguments off them.

crates/buzz-db/src/lib.rs
```diff
@@ -130,11 +130,7 @@ pub async fn insert_mentions(
         .into_iter()
         .filter(|pk| {
             if pk.len() != 64 || !pk.chars().all(|c| c.is_ascii_hexdigit()) {
-                tracing::debug!(
-                    event_id = %event.id,
-                    invalid_ptag = pk,
-                    "skipping malformed p-tag in insert_mentions"
-                );
+                tracing::debug!("skipping malformed p-tag in insert_mentions");
```
wpfleger96: 🤖 Losing both `event_id` and the invalid p-tag value makes this debug line useless — you know a malformed tag was skipped but not on which event or what it looked like. Nostr event IDs are public content hashes, not tenant data, so I think `event_id` at minimum should stay.

crates/buzz-db/src/lib.rs
```diff
@@ -1623,8 +1651,8 @@ impl Db {
     ) -> Result<(StoredEvent, bool)> {
         let result = event::insert_event(&self.pool, community_id, event, channel_id).await?;
         if result.1 {
-            if let Err(e) = insert_mentions(&self.pool, community_id, event, channel_id).await {
-                tracing::warn!(event_id = %event.id, "Failed to insert mentions: {e}");
+            if let Err(_e) = insert_mentions(&self.pool, community_id, event, channel_id).await {
+                tracing::warn!("mention insertion failed");
```
wpfleger96: 🤖 This (and the five other mention-index failure paths that got the same treatment) now logs neither the error nor the `event_id`, so a mention-index write failure can't be correlated to anything. Event IDs are public hashes — keeping `event_id` plus the error class here is safe and keeps incidents debuggable.

crates/buzz-audit/src/service.rs
```diff
@@ -125,7 +125,7 @@ impl AuditService {
 
         audit_entry.hash = compute_hash(&audit_entry)?.to_vec();
 
-        debug!(seq, "writing audit entry");
+        debug!("writing audit entry");
```
wpfleger96: 🤖 Audit logging loses `seq` here, `action` from the span fields on `log()`, and the range args on `verify_chain`/`get_entries` via `skip_all`. The error contract in `buzz-audit/src/error.rs` explicitly documents per-community `seq` as safe, intended diagnostics — this rewrite contradicts that. I'd restore `action` and `seq`; both are low-cardinality and per-community.

crates/buzz-db/src/lib.rs
```diff
@@ -3697,6 +3843,7 @@ impl Db {
     }
 
     /// Enable or disable a workflow.
+    #[tracing::instrument(target = "buzz_datastore", name = "set_workflow_enabled", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
```
wpfleger96: 🤖 `disable_workflows_for_owner_in_channel` just below has production callers but no annotation — it disappears from the datastore picture while all its neighbors are covered.

crates/buzz-db/src/lib.rs
```diff
@@ -4108,6 +4286,7 @@ impl Db {
     ///
     /// Idempotent — uses `ON CONFLICT DO NOTHING`. Returns the number of rows
     /// inserted, or 0 if the `pubkey_allowlist` table doesn't exist.
+    #[tracing::instrument(target = "buzz_datastore", name = "backfill_from_allowlist", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
```
wpfleger96: 🤖 `mint_relay_invite`, `reap_expired_relay_invites`, and `claim_relay_invite` below are unannotated live operations — coverage holes amid annotated neighbors.

crates/buzz-db/src/lib.rs
```diff
@@ -4614,6 +4819,7 @@ impl Db {
     /// prevents the stale-snapshot race where a concurrent publication reads
     /// older state and overwrites a newer snapshot by arrival order.
     ///
+    #[tracing::instrument(target = "buzz_datastore", name = "publish_nip43_membership_locked", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
```
wpfleger96: 🤖 `nip43_membership_snapshot_needs_reconciliation` above carries no annotation, so traces expose only its lower-level child calls — not the logical operation this PR sets out to measure.

crates/buzz-db/src/replica_fence.rs
```diff
@@ -542,6 +544,7 @@ pub enum ProbeError {
 /// The statements are separately awaited on a single pinned connection;
 /// a single SELECT would not guarantee evaluation order across the
 /// subexpressions, reopening the race this ordering exists to close.
+#[tracing::instrument(target = "buzz_datastore", name = "replica_fence_sample_writer", skip_all, fields(otel.kind = "client", db.system.name = "postgresql"))]
```
wpfleger96: 🤖 This runs on the fence probe loop every 500ms (`PROBE_INTERVAL`) outside any request span, so each execution exports as its own single-span root trace — roughly 172k orphan traces/day/pod at the default filter, which already ships `buzz_datastore=info`. `observe_heartbeat` (per routed read) and `usage_metrics_leader_is_live` (per poller tick) have the same shape.

I'd drop the annotations from the probe/poller internals, or move them to a separate default-off target like `buzz_datastore_probe`. The startup-only ones (`migrate`, the floor-guard verifies) are fine.

wpfleger96 (changes requested):
🤖 Thanks for the PR — the research comment on per-query spans vs SQLCommenter is good analysis, and operation-level logical spans is the right scope.

The architecture is solid: dedicated `buzz_datastore` target with an independent OTEL filter, `skip_all` everywhere, uniform client-kind attributes, and the HTTP↔datastore same-trace test in `router.rs` proves the parenting actually works. But I think three correctness gaps need fixing before this merges (inline comments have the details):

1. failed operations export as ordinary successful spans — nothing records error status
2. the log scrubbing strips established diagnostics well beyond the new spans (serde errors, `event_id` correlation, audit `action`/`seq`)
3. five live operations are missing annotations amid annotated neighbors

There's also a trace-volume issue: the fence-probe/poller annotations run outside any request span and export as high-volume orphan root traces at the default filter.

Minor, non-blocking: spans carry `db.system.name` only — `db.operation.name` would be nice but the span name works for Datadog. And if `usage_metrics_leader_is_live` keeps its annotation, worth a doc comment noting it swallows failure by design, so that span will never show error status.

crates/buzz-search/src/query.rs
ThePumpingLemma: The manual search future span has been replaced with the same function-level `#[datastore_span]` policy macro. A `fetch_all(...).await?` failure now records an OTel error status before propagation, while the error value itself is never formatted or attached to the span. The exporter test explicitly covers the `?` path.

Updated by Amp 🤖

crates/buzz-db/src/event.rs
ThePumpingLemma: Restored the established deserialization diagnostic, including the serde error text. This remains a log diagnostic; datastore function arguments are still excluded from the new spans by the macro-enforced `skip_all` policy.

Updated by Amp 🤖

crates/buzz-db/src/lib.rs
ThePumpingLemma: Implemented this centrally in the new `buzz-datastore-tracing` attribute macro. All converted `Result`-returning operations now declare `otel.status_code` and record `ERROR` before returning an `Err`; the macro never inspects or formats the error. I intentionally used status-only rather than inventing a weak cross-enum error classification. The runtime test uses the actual `tracing-opentelemetry` layer plus `InMemorySpanExporter` and verifies error status, canonical fields, skipped arguments, `return Err`, `?`, and absence of raw error text from attributes/status/events.

Updated by Amp 🤖

crates/buzz-db/src/lib.rs
ThePumpingLemma: Restored both structured fields on this diagnostic: `event_id = %event.id` and `invalid_ptag = pk`. The tracing macro still suppresses function arguments from span creation, so this restores the intentional debug correlation without reverting the span privacy default.

Updated by Amp 🤖

crates/buzz-db/src/lib.rs
ThePumpingLemma: Restored the original `event_id` correlation and error detail on all six mention-index failure paths, including the thread-metadata, reaction, addressable-event, NIP-43 snapshot, and parameterized-event paths. The new macro handles returned datastore errors separately and does not require these existing operational logs to be scrubbed.

Updated by Amp 🤖

crates/buzz-audit/src/service.rs
ThePumpingLemma: Restored `action` as an explicit safe span field and `seq` on the audit-write debug event. `verify_chain` now explicitly records `from_seq`/`to_seq`, and `get_entries` records `from_seq`/`limit`. `community` remains excluded because the policy macro always applies `skip_all`; only the reviewed safe fields are reintroduced.

Updated by Amp 🤖

crates/buzz-db/src/lib.rs
ThePumpingLemma: Added a logical `nip43_membership_snapshot_needs_reconciliation` datastore span around the complete comparison. Its existing `query_events` and `list_relay_members` spans now appear as children, while `skip_all` keeps the community and relay pubkey out of span attributes.

Updated by Amp 🤖

crates/buzz-db/src/lib.rs
ThePumpingLemma: Added logical datastore spans to `mint_relay_invite`, `reap_expired_relay_invites`, and `claim_relay_invite` at the public `Db` facade. They use the shared policy macro, so plaintext invite material, token hashes, caller identities, timestamps, and errors remain excluded while failures still set OTel error status.

Updated by Amp 🤖

crates/buzz-db/src/lib.rs
ThePumpingLemma: Added the missing logical `disable_workflows_for_owner_in_channel` datastore span at the public `Db` facade. The shared policy macro applies `skip_all`, PostgreSQL client semantics, and error status without exporting the community, channel, owner pubkey, or raw error value.

Updated by Amp 🤖

crates/buzz-db/src/replica_fence.rs
ThePumpingLemma: Removed datastore annotations from `sample_writer`, `observe_heartbeat`, and `UsageMetricsLeader::is_live`. This avoids the 500 ms probe and periodic/polling internals producing high-volume root or redundant child spans under the default `buzz_datastore` filter. Startup verification and request-level logical datastore spans remain instrumented.

Updated by Amp 🤖

wpfleger96 (approved):
🤖 Re-reviewed at exact head `b064a61288`. All prior blocking and important feedback is addressed: failures now set sanitized OTel error status through the shared policy macro (including `?` propagation), established diagnostics and safe audit fields are restored, the five missing operations are instrumented, and the high-volume probe/poller spans are removed. The dedicated exporter test also verifies `skip_all`, canonical fields, and absence of argument/error data.

PR #3682, opened 2026-07-30
fix: persist Inbox 'Show unread only' toggle across sessions

### LINKED ISSUE #3669: Persist Inbox "Show unread only" toggle across visits
## Problem

The Inbox's **Show unread only** toggle resets to off whenever the Inbox view is remounted. Users who prefer an unread-only Inbox have to enable it again after navigating away, reopening the Inbox, or restarting Buzz.

The current Desktop implementation stores this only as component state and initializes it to `false`:

https://github.com/block/buzz/blob/485d03a358b6d695aaf97879f3fbaf2f308d0755/desktop/src/features/home/ui/HomeView.tsx#L109-L110

## Steps to reproduce

1. Open the Inbox.
2. Enable **Show unread only**.
3. Navigate away from the Inbox and return, or restart Buzz and reopen the Inbox.

## Current behavior

**Show unread only** is off again.

## Expected behavior

Buzz remembers the user's last choice and restores the toggle to the same checked or unchecked state the next time the Inbox opens.

## Acceptance criteria

- Enabling **Show unread only** remains effective after navigating away from and back to the Inbox.
- The enabled state is restored after restarting Buzz.
- Disabling the toggle updates the saved preference so it remains disabled on subsequent visits.
- Users without a saved preference retain the current default of off.
- Automated coverage verifies preference restoration and the default/fallback behavior.

iroiro147:
## Problem

The Inbox's **Show unread only** toggle resets to off whenever the Inbox view is remounted. Users who prefer an unread-only Inbox have to enable it again after navigating away, reopening the Inbox, or restarting Buzz.

Closes #3669

## Root Cause

The `unreadOnly` state in `HomeView.tsx` was a plain `React.useState(false)` with no persistence layer — every remount starts fresh.

## Solution

Introduced `usePersistentBoolean` — a small `localStorage`-backed hook that follows the **same versioned-key pattern** already used by `useFeedItemState` in this codebase (e.g. `buzz-home-feed-done.v1`).

- On mount, reads the stored value from `localStorage`. If absent, falls back to `false`.
- On every state change, writes the new value back to `localStorage`.
- SSR-safe (`typeof window` guard) and degrades silently if storage is unavailable.

### Changes

| File | Change |
|---|---|
| `desktop/src/features/home/usePersistentBoolean.ts` | New hook (40 lines) |
| `desktop/src/features/home/ui/HomeView.tsx` | Replace `useState(false)` with `usePersistentBoolean` (2 lines changed) |
| `desktop/src/features/home/usePersistentBoolean.test.mjs` | 7 tests covering default, restore true/false, toggle, key convention |

## Acceptance Criteria

- [x] Enabling **Show unread only** remains effective after navigating away from and back to the Inbox
- [x] The enabled state is restored after restarting Buzz
- [x] Disabling the toggle updates the saved preference so it remains disabled on subsequent visits
- [x] Users without a saved preference retain the current default of off
- [x] Automated coverage verifies preference restoration and the default/fallback behavior

## Test Plan

```bash
cd desktop
node --test src/features/home/usePersistentBoolean.test.mjs
```

All 7 tests pass:
- default is false when no stored value exists
- stored true is restored as true
- stored false is restored as false
- writing true then reading returns true
- writing false then reading returns false
- toggling from true to false persists the new value
- storage key uses versioned prefix matching existing buzz-home patterns

### DIFF
--- a/desktop/src/features/home/ui/HomeView.tsx
+++ b/desktop/src/features/home/ui/HomeView.tsx
@@ -15,6 +15,7 @@ import {
   formatInboxFullTimestamp,
   getInboxItemConversationId,
 } from "@/features/home/lib/inbox";
+import { usePersistentBoolean } from "@/features/home/usePersistentBoolean";
 import { useInboxSelectionAnchor } from "@/features/home/useInboxSelectionAnchor";
 import { useOwnedAgentPubkeys } from "@/features/home/useOwnedAgentPubkeys";
 import {
@@ -107,7 +108,10 @@ export function HomeView({
     homeInboxWidthPx > 0 &&
     homeInboxWidthPx < INBOX_SINGLE_COLUMN_BREAKPOINT_PX;
   const [filter, setFilter] = React.useState<InboxFilter>("all");
-  const [unreadOnly, setUnreadOnly] = React.useState(false);
+  const [unreadOnly, setUnreadOnly] = usePersistentBoolean(
+    "buzz-home-inbox-unread-only.v1",
+    false,
+  );
   // Explicit selections are mirrored to the URL (`?item=`), so back/forward
   // restores the detail pane each history entry was showing and reloads
   // restore it from the URL. Default/automatic selection stays local-only —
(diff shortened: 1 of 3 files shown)


ngthuydiem:
FYI, https://github.com/block/buzz/pull/7672 has been merged.

iroiro147:
Thanks for the heads-up — #7672 covers this and has landed, so closing this one in its favour.

PR #3690, opened 2026-07-30
fix(relay): emit participant-addressed DM discovery

### LINKED ISSUE #2897: buzz dms list returns [] for an active DM participant (recipient can't discover the conversation)
## Summary

`buzz dms list` returns an empty array (`[]`) for a user who is an active participant in an existing DM conversation — even though the DM has real messages flowing and `buzz dms open` reports the conversation already exists (`created: false`). This makes it impossible for an agent to discover DM conversations it is a party to.

## Environment

- Repo: `block/buzz` @ `499c5d3` (`fix(relay): preserve reconnect backoff (#2759)`)
- `buzz-cli` built from source (`cargo build --release -p buzz-cli`)
- Relay: a hosted Block community relay (`*.communities.buzz.xyz`)
- Two participants: an agent identity ("Chip") and a human. The human initiated the DM from a Buzz client; the agent polls via `buzz-cli`.

## Steps to reproduce

1. Human user opens a DM with the agent's pubkey from a Buzz client and sends messages.
2. As the **agent** (recipient), run:
   ```
   buzz dms list
   # => []
   ```
3. Confirm the conversation actually exists and is reachable by the agent:
   ```
   buzz channels list
   # => includes {"channel_id":"<dm-uuid>","name":"DM","description":"","created_at":...}

   buzz messages get --channel <dm-uuid> --limit 10
   # => returns the real kind:9 DM messages, including ones authored by the human,
   #    each tagged ["p","<agent-pubkey>"] and ["h","<dm-uuid>"]
   ```
4. Even explicitly (re)opening the DM from the agent side does not make it appear:
   ```
   buzz dms open --pubkey <human-pubkey>
   # => {"accepted":true,"dm_id":"<same-dm-uuid>","message":"response:{...,\"created\":false}"}

   buzz dms list
   # => still []   (retried repeatedly, no propagation delay involved)
   ```

## Expected

`buzz dms list` should return the DM conversation for **any** participant (initiator or recipient), so an agent can enumerate the DMs it belongs to.

## Actual

`buzz dms list` returns `[]` for the recipient, despite an active, message-bearing DM that is visible via `channels list` and `messages get`.

## Likely cause

`cmd_list_dms` (`crates/buzz-cli/src/commands/dms.rs`) queries **kind:41001** (relay-confirmed DMs) filtered by `#p = <caller pubkey>`:

```rust
let filter = serde_json::json!({
    "kinds": [41001],
    "#p": [my_pk],
    "limit": limit
});
```

`buzz dms open` submits a **kind:41010** request with a `d` tag (`crates/buzz-cli/src/commands/dms.rs::cmd_open_dm`). The gap appears to be on the relay side: the **kind:41001 relay-confirmation event is either not emitted for the recipient, or is not `#p`-tagged with every participant** (only the initiator, or nobody), so the recipient's `#p` filter never matches. The kind:9 messages in the conversation *are* correctly `p`-tagged to the recipient, which is why `channels list` / `messages get` work while `dms list` does not.

So the DM data plane (messages) is fine; the DM **discovery** plane (kind:41001 confirmations queryable by `#p` for all participants) is missing/incomplete.

## Impact

Agents built on `buzz-cli` cannot reliably enumerate DM conversations. A poll-based integration that seeds "which conversations are DMs" from `dms list` will silently treat inbound DMs as regular channels (or miss them entirely), because the only reliable enumeration source (`channels list`) does not distinguish a DM from a channel — the DM shows up simply as `{"name":"DM"}` with no type/kind field.

## Suggested fixes (either or both)

1. **Relay:** ensure the kind:41001 relay-confirmed-DM event is emitted for a conversation and carries a `p` tag for **every** participant (initiator and all recipients), so `dms list`'s `#p` filter matches for all of them.
2. **CLI/API ergonomics:** expose a reliable way to distinguish a DM from a channel in `channels list` output (e.g. a `type`/`kind` field, or a `participants` array), so clients aren't forced to infer it from the display name `"DM"`.

Happy to provide raw event JSON (kind:9 tags, the `dms open` response, `channels list` output) or test against a specific relay build if useful.

ScaleLeanChris:
## Summary

- emit the relay-confirmed kind:41001 event that `buzz dms list` already queries
- address every DM participant with a `p` tag
- preserve channel scope with `d` and `h` tags so the private-channel read gate applies
- emit on create, re-open, and immutable group-DM creation
- use the channel creation timestamp so retries and legacy self-healing produce one stable event

## Root cause

The CLI discovery contract existed, but the relay had no kind:41001 producer. The relay emitted kind:39000 channel metadata and membership notifications, so the conversation appeared in `buzz channels list` while `buzz dms list` returned no conversations.

## Verification

- rebased onto current `block/buzz` main at `da818eddc2f470c006a1073c8c5452f8a989f272`
- `cargo check -p buzz-relay -p buzz-test-client --tests`
- real local relay regression:
  - `cargo test -p buzz-test-client --test e2e_nostr_interop test_dm_discovery_events_emitted -- --ignored --nocapture`
  - result: 1 passed, 0 failed
  - proves creator and recipient discovery, one event after recipient re-open, complete participant and channel tags, and discovery by a newly added group-DM participant
- repository gate: `just ci`
  - result: passed
  - includes workspace and Desktop Clippy, Desktop and web checks/builds, Rust tests, Desktop Tauri tests (2684 passed, 18 ignored), and mobile tests (1552 passed)

Fixes #2897

Controlling work order: [BAC-109](https://linear.app/scale-lean/issue/BAC-109/refresh-and-verify-selected-buzz-and-hermes-upstream-contributions)

The relay already knows every participant, so DM discovery should not end as “Return to Sender.”

### DIFF
--- a/crates/buzz-relay/src/handlers/command_executor.rs
+++ b/crates/buzz-relay/src/handlers/command_executor.rs
@@ -29,8 +29,8 @@ use crate::webhook_secret;
 
 use super::ingest::{extract_channel_id, IngestAuth, IngestError, IngestResult};
 use super::side_effects::{
-    emit_group_discovery_events, emit_membership_notification, emit_system_message,
-    publish_dm_visibility_snapshot,
+    emit_dm_created, emit_group_discovery_events, emit_membership_notification,
+    emit_system_message, publish_dm_visibility_snapshot,
 };
 
 /// Route a command-kind event to the appropriate handler.
@@ -367,6 +367,35 @@ fn compute_definition_hash(json_str: &str) -> Vec<u8> {
     Sha256::digest(json_str.as_bytes()).to_vec()
 }
 
+async fn publish_dm_discovery(
+    tenant: &TenantContext,
+    state: &Arc<AppState>,
+    channel: &buzz_db::channel::ChannelRecord,
+    participant_pubkeys: &[Vec<u8>],
+    operation: &'static str,
+) {
+    match u64::try_from(channel.created_at.timestamp()) {
+        Ok(created_at) => {
+            if let Err(e) =
+                emit_dm_created(tenant, state, channel.id, created_at, participant_pubkeys).await
+            {
+                warn!(
+                    channel = %channel.id,
+                    operation,
+                    "DM discovery emission failed: {e}"
+                );
+            }
+        }
+        Err(_) => {
+            warn!(
+                channel = %channel.id,
+                operation,
+                "DM discovery has an invalid channel creation timestamp"
+            );
+        }
+    }
+}
+
 async fn handle_dm_open(
     tenant: &TenantContext,
     state: &Arc<AppState>,
@@ -485,6 +514,11 @@ async fn handle_dm_open(
             warn!("DM re-open: visibility snapshot failed: {e}");
         }
     }
+
+    // Use the channel creation timestamp so retries produce the same event id.
+    // This also backfills discovery for DMs created before kind:41001 emission
+    // was implemented without adding duplicate rows on each re-open.
+    publish_dm_discovery(tenant, state, &channel, &all_bytes, "open").await;
 
     // 6. Return response
     Ok(IngestResult {
@@ -623,6 +657,8 @@ async fn handle_dm_add_member(
             }
         }
     }
+
+    publish_dm_discovery(tenant, state, &new_channel, &all_bytes, "add_member").await;
 
     // 8. Return response
     Ok(IngestResult {
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -7,11 +7,11 @@ use tracing::{info, warn};
 use uuid::Uuid;
 
 use buzz_core::kind::{
-    event_kind_u32, is_parameterized_replaceable, KIND_AGENT_PROFILE, KIND_DM_VISIBILITY,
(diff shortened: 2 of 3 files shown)


mfethe1:
Reproduced this on macOS 15.6 (arm64) at head `5e419bea1`, against a local single-node relay (Podman: Postgres 17, Redis 7, MinIO).

```
cargo test -p buzz-test-client --test e2e_nostr_interop \
  test_dm_discovery_events_emitted -- --ignored --nocapture
```

**PR head: 1 passed**, 0 failed (0.73s).

**Negative control:** I reverted only `crates/buzz-relay/src/handlers/command_executor.rs` and `crates/buzz-relay/src/handlers/side_effects.rs` to `5e419bea1~1`, leaving your test file unchanged, then rebuilt the relay and restarted it against a freshly recreated database and Redis.

**Without the implementation: 1 failed**, at `e2e_nostr_interop.rs:930`:

```
assertion `left == right` failed: create plus re-open must leave one
kind:41001 event for creator; events: []
  left: 0
 right: 1
```

Restoring both files and rebuilding returned it to **1 passed** (0.23s).

Zero `kind:41001` events are emitted without the change, so the test discriminates the fix from its absence rather than passing either way — and it corroborates the stated root cause (no kind:41001 producer) rather than only exercising the new code path.

Also clean on the same head: `cargo fmt --all -- --check`, and `cargo check -p buzz-relay -p buzz-test-client --tests` (0 warnings, 0 errors).

Scope and limitations: single-node local relay, not a multi-instance or mesh configuration. I set `BUZZ_GIT_CONFORMANCE_PROBE=false` to bypass the S3/object-store startup gate, which is unrelated to DM discovery. The head was 63 commits behind `main` when I tested; I tested the submitted head as-is, without rebasing.

PR #3700, opened 2026-07-30
fix(desktop): detect Kimi Code in its default install path

benthecarman:
## Summary

By default on linux kimi is installed at `~/.kimi-code/bin/kimi` and this was not detected automatically. This now fixes it by adding this path to the detection 

### Related issue

 N/A

### Testing

Before it did not appear / did not detect. Now it does :)

### DIFF
--- a/desktop/src-tauri/src/managed_agents/discovery.rs
+++ b/desktop/src-tauri/src/managed_agents/discovery.rs
@@ -38,6 +38,7 @@ fn common_binary_paths() -> &'static [PathBuf] {
             paths.extend([
                 home.join(".local/share/mise/shims"),
                 home.join(".local/bin"),
+                home.join(".kimi-code/bin"),
                 home.join(".volta/bin"),
                 home.join(".asdf/shims"),
                 home.join(".bun/bin"),


wesbillman (approved):

felirami:
Confirming this also reproduces on **macOS**, so the fix is cross-platform, not Linux-specific.

Diagnosis from an affected machine (Buzz 0.5.3, kimi-code 0.31.1):

- The desktop app launched via Finder inherits launchd's default `PATH=/usr/bin:/bin:/usr/sbin:/sbin`.
- kimi-code's installer appends `~/.kimi-code/bin` to PATH in `~/.zshrc` (interactive shells only), which a login shell (`zsh -l -c 'command -v kimi'`) never sources — so the `find_via_login_shell` fallback in `discovery.rs` also misses it. Verified with `env -i HOME=$HOME PATH=/usr/bin:/bin:/usr/sbin:/sbin /bin/zsh -l -c 'command -v kimi'` → not found.
- `~/.kimi-code/bin` is absent from `common_binary_paths()`, so resolution fails at every stage. This mirrors the existing nvm special-case a few lines below ("~/.zshrc (interactive) which is not loaded by a login shell").

Adding `~/.kimi-code/bin` to `common_binary_paths()` fixes detection regardless of the user's shell configuration. Would love to see this merged — until then the workaround for affected users is adding the export to `~/.zprofile`.

PR #3705, opened 2026-07-30
fix(buzz-agent): honor OPENAI_BASE_URL as fallback for base URL

### LINKED ISSUE #3630: buzz-agent ignores custom OPENAI_BASE_URL from Agent defaults, always calls api.openai.com
buzz-agent ignores custom OPENAI_BASE_URL from Agent defaults, always calls api.openai.com

Chukwuebuka-2003:
## Summary

When `BUZZ_AGENT_PROVIDER` is set to an OpenAI-compatible provider but `OPENAI_COMPAT_BASE_URL` is not explicitly set, the agent should fall back to the value of `OPENAI_BASE_URL` before using the default `https://api.openai.com/v1`. This matches user expectations when migrating from other tools that use `OPENAI_BASE_URL` as the standard env var.

Fixes #3630

## Changes

- In `config.rs`, the `from_env()` OpenAI base URL resolution now checks `OPENAI_BASE_URL` as a fallback when `OPENAI_COMPAT_BASE_URL` is unset, before falling through to `https://api.openai.com/v1`.

## Testing

Existing tests pass — this is a pure backward-compatible fallback addition.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -835,6 +835,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "sha2 0.11.0",
+ "temp-env",
  "tempfile",
  "tokio",
  "tracing",
@@ -2177,7 +2178,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ccc2776f0c61eca1ca32528f85548abd1a4be8fb53d1b21c013e4f18da1e7090"
 dependencies = [
  "data-encoding",
- "syn 1.0.109",
+ "syn 2.0.117",
 ]
 
 [[package]]
@@ -8712,6 +8713,15 @@ dependencies = [
  "filetime",
  "libc",
  "xattr",
+]
+
+[[package]]
+name = "temp-env"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96374855068f47402c3121c6eed88d29cb1de8f3ab27090e273e420bdabcf050"
+dependencies = [
+ "parking_lot",
 ]
 
 [[package]]
--- a/crates/buzz-agent/Cargo.toml
+++ b/crates/buzz-agent/Cargo.toml
@@ -55,3 +55,4 @@ hex = { workspace = true }
 serde = { workspace = true }
 sha2 = { workspace = true }
 tempfile = "3"
+temp-env = "0.3"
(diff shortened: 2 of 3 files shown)


Bartok9:
**Independent review (Bartok9)** — 2026-07-30 vs current `main`

Thanks for pushing on #3630 — agreeing env fallback is needed.

### Blocking nit: precedence is inverted vs the PR description

Description says: when **`OPENAI_COMPAT_BASE_URL` is unset**, fall back to `OPENAI_BASE_URL`.

Current patch:

```rust
env("OPENAI_BASE_URL")
    .or_else(|| env("OPENAI_COMPAT_BASE_URL")
    .unwrap_or_else(|| "https://api.openai.com/v1".into()),
```

That prefers **`OPENAI_BASE_URL` over the Buzz-native `OPENAI_COMPAT_BASE_URL`**, so a leftover generic `OPENAI_BASE_URL` in the shell can **override** an intentional `OPENAI_COMPAT_BASE_URL` (opposite of typical specific-over-generic ordering).

Suggested:

```rust
env("OPENAI_COMPAT_BASE_URL")
    .or_else(|| env("OPENAI_BASE_URL"))
    .unwrap_or_else(|| "https://api.openai.com/v1".into()),
```

### Testing gap
No unit test pins the two-env matrix (compat set / base only / neither). Worth a small `from_env` test with tempered env.

### Scope note
Issue title mentions **Agent defaults** (Desktop). Env fallback still helps if defaults export `OPENAI_BASE_URL` into the agent process; if Desktop only writes another key, a second seam may remain — happy to help chase that if this lands and dogfood still breaks.

I can open a tiny residual PR with corrected precedence + tests if you want — won’t stomp if you push a fix here first.

Bartok9:
Follow-up residual with corrected precedence + unit tests: #3773 — happy to close it if you flip the `or_else` order here and add similar tests on this branch first.

Chukwuebuka-2003:
Thanks for the thorough review @Bartok9 — good catch on the precedence.

I've flipped the `or_else` order so `OPENAI_COMPAT_BASE_URL` is checked first (Buzz-specific), then `OPENAI_BASE_URL` as fallback, then the default. Also added three unit tests covering the full env matrix:

| OPENAI_COMPAT_BASE_URL | OPENAI_BASE_URL | Expected |
|---|---|---|
| set | set | compat wins |
| unset | set | generic fallback |
| unset | unset | default |

Your #3773 should be superseded now — happy to close it.

Bartok9:
Perfect — precedence is now specific-over-generic and the three-case env matrix (compat set / base only / neither) is covered. LGTM. I've closed #3773 as superseded. Thanks for the quick turnaround!

Chukwuebuka-2003:
You're welcome

Bartok9:
🙏 Likewise — nice clean fix. Hope #3630 stays quiet after this lands.

Issue #2714, opened 2026-07-24
feat (large impact/desktop) threads below channels in sidebar


kieranx:
After using Buzz extensively for the past 24 hours, the no. 1 thing it's missing is thread management.

Request: Show recent threads in the sidebar, below the channel, with an AI generated name (customizable). 

Think of the way the Codex Mac App works - you have a folder, and then chats inside

This can be translated to Channels and Threads - the UI can look almost identical

This would have a dramatic impact on how usable Buzz is for lots of work and is the biggest pain point with dealing with agents in Slack for many.

PR #3712, opened 2026-07-30
feat(desktop): add message bookmarks + Saved view

divljak:
## Summary

Adds a Slack-style **"Save for later"** bookmark feature to the desktop app:

- **Bookmark toggle** in the message hover toolbar, between the emoji-react and reply buttons (filled icon when saved; `Save for later` / `Remove bookmark` tooltip). Works on both the channel timeline and the home inbox.
- **Saved view** — a new top-level **Saved** item in the left nav opening a `/saved` route that lists your bookmarked messages newest-first, with click-to-jump-back-to-context (opens the thread when the saved message is a reply).

### Design notes / decision to sanity-check

Bookmarks are stored as a **private, per-user NIP-78 app-data list** (`kind:30078`, d-tag `"bookmarks"`), **encrypted to self via `nip44`** — a direct clone of the existing channel-stars/mutes/sort sync engine (`desktop/src/features/sidebar/lib/channelStarsSync.ts`). Consequences:

- **No relay/backend change** — `kind:30078` is already accepted; this is a pure `desktop/` change.
- **Private by default** — only the author can decrypt their saved list (matches Slack "saved items" semantics). The only plaintext tags are the `d`/`t` markers; no per-message data leaks.
- **Single-query Saved view** — one `authors:[self] #d:bookmarks` fetch, no cross-channel fan-out.
- Payload is a replaceable list with per-key last-write-wins merge; un-bookmarking is a tombstone (`bookmarked:false`) so removals propagate across devices.

**One thing for maintainers to weigh in on:** `buzz-core/src/kind.rs` reserves `KIND_STREAM_MESSAGE_BOOKMARKED = 40005` ("a stream message bookmarked by a user"), which may be the intended home for this. I went with the `30078` private-list approach because it's private, cross-channel, needs no relay change, and reuses a proven client pattern — but `40005` is channel-scoped and would need relay handling + a cross-channel query for the Saved view. Happy to pivot to `40005` if that's the preferred direction.

### Related issue

None found (searched issues/PRs for "bookmark" / "save for later").

### Testing

All desktop gates green locally:

- `tsc --noEmit` ✓ · `biome check` ✓ · `pnpm check:px-text` ✓
- `pnpm test` — **3778/3778** pass, including **9 new** unit tests for the parse/merge/tombstone/thread-root logic (`bookmarksStorage.test.mjs`)
- `vite build` ✓

Also went through an independent code review pass; fixed two items it surfaced before opening: a no-op remote-sync path that could re-render the whole timeline (now content-stable via `useStableSet` + a split actions/list context), and jump-to-context missing thread context for saved replies (now captured as `threadRootId`).

Not yet exercised: a full native Tauri run against a live relay. UI screenshots to follow.

### DIFF
--- a/desktop/src/app/AppShell.helpers.ts
+++ b/desktop/src/app/AppShell.helpers.ts
@@ -9,7 +9,8 @@ export type AppView =
   | "agents"
   | "workflows"
   | "pulse"
-  | "projects";
+  | "projects"
+  | "saved";
 
 const WINDOW_DRAG_HANDLE_HEIGHT = 44;
 const TAURI_DRAG_REGION_ATTR = "data-tauri-drag-region";
@@ -150,6 +151,13 @@ export function deriveShellRoute(pathname: string): {
     return {
       selectedChannelId: null,
       selectedView: "pulse",
+    };
+  }
+
+  if (pathname === "/saved") {
+    return {
+      selectedChannelId: null,
+      selectedView: "saved",
     };
   }
 
--- a/desktop/src/app/AppShell.tsx
+++ b/desktop/src/app/AppShell.tsx
@@ -72,6 +72,8 @@ import { requestFocusedThreadClose } from "@/features/channels/focusedThreadClos
 import { CommunityRail } from "@/features/sidebar/ui/CommunityRail";
 import { useChannelMutes } from "@/features/sidebar/lib/useChannelMutes";
 import { useChannelStars } from "@/features/sidebar/lib/useChannelStars";
+import { BookmarksProvider } from "@/features/bookmarks/lib/BookmarksContext";
+import { useBookmarks } from "@/features/bookmarks/lib/useBookmarks";
 import { useCommunities } from "@/features/communities/useCommunities";
 import {
   consumePendingCommunityRestore,
@@ -129,6 +131,7 @@ export function AppShell() {
     goNewMessage,
     goProjects,
     goPulse,
+    goSaved,
     goSettings,
     goWorkflows,
     closeSettings,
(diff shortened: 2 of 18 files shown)


mikececco:
Not a maintainer, so this isn't a decision — but I dug into the substrate question you raised and found a few things that might help unstick it.

**The `30078` + d-tag choice matches an explicitly documented house pattern.** `desktop/src/shared/constants/kinds.ts:42-48` spells the convention out:

```ts
// NIP-78 application-specific data. All use kind 30078; the relay
// differentiates them by d-tag ("read-state:<slotId>", "channel-sections",
// "channel-mutes", "channel-stars", "channel-sort").
```

So `d="bookmarks"` is the sixth slot in an established pattern, not a novel use. That seems like the strongest argument for what you built.

**One correction on `40005`:** it's further along than "would need relay handling" suggests — `KIND_STREAM_MESSAGE_BOOKMARKED` is already in the relay's ingest allowlist (`crates/buzz-relay/src/handlers/ingest.rs:252` and `:484`), so the relay accepts it today. What's missing is any producer or consumer. The real cost of pivoting isn't ingest, it's that `40005` is channel-scoped (`h`-tagged), so the cross-channel Saved view needs fan-out or a new query path — and it would be plaintext, losing the nip44 privacy property you called out.

**There's also a third candidate nobody has named:** `KIND_BOOKMARK_SET = 30003` (NIP-51 named bookmark sets) is registered in `buzz-core/src/kind.rs:43` and relay-accepted (`ingest.rs:237`, `:412`), with **zero usage in `desktop/src` or `mobile/lib`**. I mention it mainly so it gets explicitly ruled out rather than surfacing later as a fourth answer — its one advantage over `30078` is that sets are *named*, which matters if saved items ever need to be grouped into collections rather than one flat list.

**Which is why this decision is probably bigger than this PR.** There's standing demand for thread-level grouping — #2714 ("threads below channels in sidebar", untriaged) and #4266 (pin open threads in a rail) — and any of those would want the same substrate. Whatever gets picked here likely becomes the answer for those too. If a maintainer can call `30078` vs `40005`, it unblocks more than one thread of work.

Also flagging that this has gone to `DIRTY` (merge conflicts) since 07-30, so it'd need a rebase whenever it does get looked at.

PR #1485, opened 2026-07-03
NIP-FI: Federated Identity Authorization (spec + formal model)

tlongwell-block:
## What

Adds **NIP-FI: Federated Identity Authorization** — a normative spec (`docs/nips/NIP-FI.md`) and companion formal model (`docs/nips/NIP-FI-MODEL.md`) for authorizing an already-authenticated Nostr key only when a valid OIDC/JWT assertion from an external identity provider resolves to the same active identity-to-key binding.

Docs only; no code changes.

## Why

#1476 implements corporate identity binding for the Buzz relay. Per the buzz-security discussion (thread `0d1aee9e`), the requirement is broader: **anyone** with Okta or a generic OAuth2 reverse proxy providing JWTs should be able to enable this class of enhanced identity auth. Without a spec, the first large deployment's configuration becomes an accidental standard. Prior-art survey (upstream `nips` @ `8f8444d`, strfry, Chorus, nostr-rs-relay, rnostr, relay.tools, NIP-46/98, Blossom) found no existing NIP or relay implementation binding a trusted IdP subject to a pubkey at relay authentication — closest shapes are NIP-05 domain verification (DNS/HTTPS polling, not request-time credentials) and NIP-46 `auth_url` (custody UX, not relay binding).

## Design spine

- Identity is issuer-qualified `(iss, sub)` — never bare `sub`, username, or email.
- Active bindings form a per-domain **partial bijection** between identity and key; every transition preserves it.
- A bearer assertion **never substitutes** for Nostr proof (NIP-42/NIP-98); the authorized key is always the proof's key.
- Explicit enrollment modes: `attested-key` (IdP carries the npub claim), `provisioned` (admin-created only), risk-labeled `tofu`.
- Check + first bind is atomic; races and storage failures fail closed.
- Cached WS authorization is a lease bounded by assertion `exp`, tracked per authenticated key.
- Revocation cannot be undone by replaying a still-valid assertion; key rotation is an explicit transition, never conflict resolution.
- Trusted-proxy transport is conforming only when direct ingress is impossible and inbound header copies are stripped; client-attached transport uses `Nostr-Federated-Identity: Bearer <JWT>` (avoiding NIP-98's `Authorization` header).
- Delegation (e.g. NIP-OA) composes without granting the delegate a federated binding.
- Discovery via NIP-11 `limitation.federated_identity` + optional `federated_identity` object.

## Relationship to #1476

[...]

PR #1476, opened 2026-07-02
feat(relay): add relay-verified identity binding

fsola-sq:
## Why

Relays need a provider-neutral way to bind a verified external identity to the Nostr key that authenticated the request, while preserving Nostr proof and failing closed when identity evidence is missing or inconsistent.

builds on original work by @fsola

## What

- Validate asymmetric JWTs against configured issuer, audience, key, time, subject, display-name, and optional public-key claims
- Persist issuer-qualified identity bindings with atomic enrollment, lifecycle state, revocation tombstones, authorized rotation, and recovery controls
- Keep direct identity authorization separate from delegated-owner authorization and enforce explicit precedence
- Apply identity authorization across WebSocket, HTTP bridge, moderation, media, Git, private audio, and protected admission paths
- Publish privacy-bounded relay-signed identity assertions and expose their current state to the desktop UI
- Keep the feature off by default behind explicit relay configuration

## Review guide

- Identity verification: JWT and JWKS validation, trusted ingress, claim mapping, and privacy boundaries
- Binding lifecycle: migrations, atomic admission, partial-bijection enforcement, tombstones, rotation, recovery, and session eviction
- Protected transports: route classification and consistent authorization across relay, HTTP, media, Git, and audio surfaces
- Desktop presentation: relay-signed assertion handling, expiry, withdrawal, and profile surfaces

## Risk assessment

Medium. This PR changes authentication, persistence, protected routing, and desktop presentation. The feature remains off by default, and configured deployments deny authorization when required evidence cannot be verified.

## Testing

- JWT and JWKS algorithm, claim, key-use, key-age, refresh, and tamper cases
- PostgreSQL migration, enrollment, lifecycle, concurrency, rollback, and tombstone cases
- Protected-route, admission, delegation, and session-eviction cases
- Desktop assertion replacement, withdrawal, and expiry cases

## References

- Federated identity authorization specification: #1485
- Extended lifecycle and conformance specification: #3726

Operator documentation for the identity runtime is maintained separately and tracked in #3726.

PR #3726, opened 2026-07-30
docs(nips): complete NIP-FI authorization draft

cea-block:
## Why

The initial NIP-FI draft needs complete lifecycle, recovery, delegation, privacy, and conformance semantics before implementations can evaluate compatibility against one public contract.

## What

- Complete the provider-neutral authorization and binding lifecycle defined in #1485
- Define literal issuer-qualified identity, server-resolved authorization domains, durable lifecycle gates, and atomic recovery and reactivation
- Align lease, delegation, disclosure, privacy, and denial behavior across the specification and formal model
- Add a conformance matrix covering successful authorization, denial, concurrency, lifecycle, session, delegation, disclosure, and privacy cases

## Review guide

- Review this PR as the documentation delta on top of #1485
- Focus on lifecycle authority, replacement proof, binding expiry, delegation bounds, dependency invalidation, and negative conformance cases
- Runtime changes and upstream NIP acceptance are outside this PR

## Risk assessment

Low. This PR changes documentation only and does not alter runtime behavior or configuration.

## Testing

- Checked the specification, formal model, and conformance matrix for matching terms, transitions, denial classes, and security invariants
- Hosted repository checks pass for the documentation changes

## References

- Depends on #1485
- Related relay implementation: #1476


wpfleger96 (changes requested):
🤖

## Review

I reviewed the full files at `805f1131b8c946cfdcff4508d20f1758515a5990..aa1bf732a9751fb8f4af54f1d3db8b964c6831ad`, then traced the referenced implementation stack through #4847 at `5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7`. This body merges two independent full-document passes at the same refs; where the passes disagreed on severity, the higher severity stands. I’m requesting changes because the public contract and the stock runtime contract cannot currently describe one conforming implementation.

### [CRITICAL]: The stock runtime requires a different wire protocol than the NIP defines

`docs/nips/NIP-FI.md:54-78` defines exactly `trusted-proxy-hmac-v1`: a `v1.<timestamp>.<nonce>.<mac>` envelope whose MAC covers the assertion and HTTP request coordinates. The model, deployment guide, integration guide, conformance applicability rule, and JSON examples all use that same public profile. In contrast, `docs/CORPORATE_IDENTITY.md:53-58,88-92` and `.env.example:66` say production requires `trusted_proxy_hmac_v2`.

This is not a hyphen/underscore serialization difference. The referenced runtime rejects every transport kind except `trusted_proxy_hmac_v2` ([config at #4847 head](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-relay/src/authorization_runtime/config.rs#L137-L146)); its v2 MAC additionally binds the authorization domain, proof transport, and a third authenticated-client-peer field under `NIP-FI-PROXY-2` ([verifier input](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-auth/src/trusted_proxy.rs#L702-L744)). Production then rejects otherwise-valid v1 evidence because it has no authenticated peer ([upgrade boundary](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-relay/src/router.rs#L489-L503)). A stock Buzz deployment therefore cannot accept the public NIP profile or produce the v1 conformance report this PR requires.

Choose one exact protocol before this contract lands. If v2 is the intended security boundary, specify its field, canonical MAC input, discovery identifier, replay namespace, lease bound, negative vectors, and applicability rules normatively, then update every guide/example. Otherwise the production runtime contract must accept the exact v1 profile. A private v2 extension cannot be called the sole stock NIP-FI transport while the public conformance matrix recognizes only v1.

### [CRITICAL]: Operator docs assert runtime behavior that exists nowhere at this head

`docs/CORPORATE_IDENTITY.md:7-12`, `docs/NIP_FI_DEPLOYMENT.md:5-11`, `docs/NIP_FI_RUNTIME_OPERATIONS.md:5-9`, `.env.example:57-75`, and the Compose/Helm readiness sections say the relay accepts `BUZZ_NIP_FI_V1_CONFIG_JSON`, that absence/`deny_protected`/complete JSON select three operating modes, and that legacy variables fail startup. Verified at source: `BUZZ_NIP_FI_V1_CONFIG_JSON`, `deny_protected`, `trusted_proxy_hmac_v2`, and `canonical_admission` have zero hits in any code file on `origin/main`, the base branch, or this PR’s head (docs/`.env.example` only), and `git log -S` shows the “removed” legacy variables (`BUZZ_REQUIRE_CORPORATE_IDENTITY`, `BUZZ_CORPORATE_IDENTITY_*`) never existed on main — yet the doc says setting any of them fails startup. The same revision correctly says elsewhere that it contains no runtime adapter and changes no runtime configuration (`docs/NIP_FI_INTEGRATION.md:5-24`, `SECURITY.md:75-82`).

The sharpest edge is the emergency brake: *“DenyProtected: set the document to `{"deny_protected":true}`. Every protected route is denied before its handler runs.”* On the tree this merges into, the relay silently ignores that variable. An operator following this doc during an incident believes protected routes are locked when nothing happened — a false security promise in present-tense operator documentation. Same class: “constructed at startup, before listeners open,” “key refreshes are single-flight,” the 120-second status-renewal bound, and all field bounds.

Make these pages explicitly a future implementation contract in the same unambiguous voice `NIP_FI_INTEGRATION.md` already uses (“the current relay does not read this variable”), and remove current setup/pass-through claims from `.env.example`, Compose, and Helm until the runtime lands — or land the implementation atomically with the docs. “Not conformance evidence” does not fix a false statement about whether configuration is parsed at all.

### [IMPORTANT]: The runtime configuration cannot be mapped to the spec it claims to configure

Beyond the transport mismatch above, the runtime document’s vocabulary does not correspond to the normative set: `enrollment.kind: "canonical_admission"` matches none of the spec’s three enrollment modes (`attested-key`/`provisioned`/`tofu`); `restore.kind: "operation_manifest"` and the entire `client_status_admission` block (“current-status presentation renews within 120 seconds”) reference a presentation subsystem defined nowhere in the normative documents. The configuration is also single-issuer, single-domain (one `issuer` string, one `jwks`, no domain array), while `FI-TRACE-CROSS-DOMAIN-COLLISION` requires two issuers × two domains and `NIP_FI_INTEGRATION.md` describes multi-domain deployments — the documented config cannot express the scenarios its own release gate requires. An operator cannot configure the spec from this contract, and a reviewer cannot check conformance against it.

### [IMPORTANT]: Subject stability is presented as a verifier check, but no request can prove it

`docs/nips/NIP-FI.md:84-92` says validation “enforces all” listed properties, including that the issuer contract guarantees `sub` is stable, opaque, non-reassignable, and not derived from personal data. A JWT proves only the signed subject value; it cannot prove the issuer’s lifetime assignment policy. The previous model made non-reassignment an explicit trust assumption, but this revision removes that section. The stock configuration records only a claim name, and the downstream verifier can only check that the resulting string is non-empty and bounded.

Move this property back into explicit trust/deployment assumptions, require operators to record how each accepted issuer satisfies it, and state the consequence of issuer reassignment. If the property is meant to be mechanically enforced, define the additional authenticated metadata and validation rule. Leaving an organizational assumption inside a list of executable verifier checks makes the model and conformance claim stronger than any implementation can demonstrate.

### [IMPORTANT]: The claimed JWKS rollback protection has no rollback state or conformance oracle

`docs/nips/NIP-FI.md:98` and the security consideration at line 286 claim that generation revalidation prevents rollback. The model only requires `g` to change when effective keys change, and the matrix tests add and remove (`FI-TRACE-JWKS-ADD`/`REMOVE`) but never A → B → A rollback. If an authenticated JWKS endpoint republishes a previously removed key set, it is simply a new current generation; an old assertion signed by that key revalidates successfully. The downstream cache behaves exactly this way: every changed document advances a process-local counter and is published without a durable removed-key floor ([refresh](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-relay/src/authorization_runtime/jwks.rs#L260-L326)).

Either define a durable authenticated anti-rollback mechanism and add a stable rollback trace, or narrow the contract to “revalidation against the currently authenticated snapshot” and list JWKS rollback as residual issuer/key-source risk. The current text promises a security property that neither the model nor the referenced implementation can establish.

### [IMPORTANT]: The documented `policy_id` semantics cannot be produced by the stock configuration

`docs/nips/NIP-FI.md:30` requires `policy_id` to include allowed algorithms and authenticated key-source identity and to change whenever assertion semantics change. The “complete” runtime document in `docs/CORPORATE_IDENTITY.md:30-62` has no algorithm policy, and the referenced implementation hashes issuer, audience, claim names, skew, and token lifetime but not its compiled algorithm allow-list or configured JWKS/discovery source ([policy construction](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-auth/src/foundation.rs#L244-L315)). A binary upgrade that changes allowed algorithms, or a configuration change to a different authenticated key source, can therefore retain the same `policy_id` despite changing normative verifier semantics.

Put every normative semantic input into the stable digest, or include a versioned verifier-contract fingerprint that changes with compiled acceptance rules. Add vectors proving each semantic change advances `policy_id` while a key-only rotation does not.

### [IMPORTANT]: The referenced WebSocket session path does not perform the rechecks this contract makes mandatory

The normative session contract (`docs/nips/NIP-FI.md:232-242`) requires each protected use to recheck binding/lifecycle versions and the current key snapshot, and requires a changed JWKS generation to revalidate the retained assertion. The matrix specifically requires active direct leases to deny/close on key removal. At #4847 head, the WebSocket fast path checks only static domain/actor/capability/transport and lease expiry ([session check](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-relay/src/protected_ingress.rs#L214-L235)). The read/write leases are also built with assertion expiry rather than the proxy provenance deadline ([AUTH preparation](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-relay/src/handlers/auth.rs#L926-L975)), and those grants are not registered with the invalidation registry. The only periodic binding recheck is in optional client-status presentation; clients that do not opt in still receive the static session grants.

This does not weaken the normative text; it means the referenced implementation is not yet feasible as a conforming handoff. The WebSocket lease owner needs current binding/lifecycle/invalidation and JWKS revalidation on reuse (or a cancellation owner that provides equivalent guarantees), and every direct lease must be bounded by proxy expiry. Related feasibility expectation worth stating before the #1476 rework: reconnect-based lease renewal at ~300s maxima implies real reconnect churn for long-lived agent connections.

### [IMPORTANT]: The referenced final-admission path persists the denial state that the invariant forbids

`docs/nips/NIP-FI.md:162,313`, `docs/nips/NIP-FI-MODEL.md:79,245,466`, and `FI-TRACE-FINAL-DENIAL-NO-MUTATION` require local-policy and final-admission denial to create no audit or denial observation. At #4847 head, a denied fresh admission rolls back to a savepoint and then commits a denied operation receipt plus `ProtectedDenied` authorization event ([denial branch](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-db/src/authorization_admission.rs#L3055-L3129), [persistence](https://github.com/block/buzz/blob/5446360bee24aaa6ae4629a2bc6d4f0b1ea37eb7/crates/buzz-db/src/authorization_admission.rs#L3769-L3821)). This will fail the matrix’s required before/after audit-store comparison. It also turns attacker-reachable denials into consumption of the documented non-reclaimable lifetime audit budget.

Resolve the contract decision explicitly. Either denied admission is read-only and the implementation must not persist these records, or the model/invariant/matrix must permit a separately bounded, non-authoritative denial channel and analyze audit-capacity exhaustion. The current two designs cannot both pass conformance.

### [IMPORTANT]: Audit capacity is an unrecoverable one-way budget with no documented recovery transition

`docs/CORPORATE_IDENTITY.md:79-84`: “no online prune, export, reset, or acknowledgement workflow — size the budget for the installation’s lifetime… exhaustion denies further authorization-affecting operations.” Combined with fail-closed semantics, a mis-estimate — or an attacker who can generate authorization-affecting operations (see the denial-persistence finding above, which makes budget consumption attacker-reachable) — converts into a permanent, domain-wide identity outage. Nothing states whether raising the bound in configuration is a legitimate recovery path. The threat model’s availability section accepts transient fail-closed DoS but does not cover unrecoverable budget exhaustion. Document a privileged recovery transition or add an explicit residual-risk entry.

### [MINOR]

- **SECURITY.md weakening:** “channel membership is the *only* access control mechanism” becomes “application-level… policy *may* allow.” On the merged tree, membership-only is still true; the softening pre-accommodates an unshipped authority. Keep the current-state claim and note NIP-FI as optional/future.
- **Scope drift:** the 44-line AWS-LC build-requirements section in `CONTRIBUTING.md` and the pgcrypto notes in the deploy READMEs are accurate (verified `aws-lc-rs` 1.17.0 in both lockfiles; pgcrypto in `migrations/0001`) but unrelated to completing the NIP-FI authorization draft.
- **`ReenableIdentity` design question:** it forcibly creates a new binding — there is no path to clear disabled state without a target key. In `provisioned` mode that conflates two operator intents (re-enable now vs. provision later).
- **Conformance feasibility:** the denial-timing oracle’s predeclared statistical thresholds will be hard to run deterministically in CI; worth a stated execution strategy before the matrix becomes a release gate.

## What's Solid

The lifecycle redesign itself is strong, and it genuinely closes the original revocation-resurrection blocker. Walking the scenario and its variants: a revoked key with a still-valid JWT hits the tombstone selector before the enrollment-mode switch, so `attested-key`/`tofu` can never re-enroll a revoked coordinate; a fresh key for a retired identity is blocked by the durable retired lineage, which only explicit recovery consumes; revocation landing between prepare and commit is caught by the atomic recomputation-and-equivalence recheck inside final admission (`FI-INV-03`, `FI-TRACE-TOMBSTONE-REPLAY`); and lifecycle transitions invalidate dependent leases after commit with an honest polling-bound caveat. Exact pending lineage makes recovery and re-enablement explicit; administrative expiry does not free either side of the partial bijection; and the read-only preparation/atomic commit split gives concurrency a clear serialization point. The proxy MAC construction is careful (length-prefixed fields, domain-separation tag, constant-time compare, overflow-safe time handling, equality-is-expired). The 24 trace labels and 16 invariant labels are internally aligned across the spec, model, matrix, and report example.

At the exact PR head, `git diff --check` passed; all three JSON examples parsed; local Markdown links, fences, and tables were valid; trace/invariant label equality held (`24` traces appearing exactly once in each of the four carrying documents, `16` invariants); and the runtime-symbol absence claims above were verified with `git grep`/`git log -S` against `origin/main`, the base branch, and this head. These are documentation-consistency checks only, as the PR correctly notes; CI at head is green for the selected docs paths, with relay/security/unit jobs skipped.

## Alternatives

The simplest coherent route is to promote the already-required v2 envelope into the normative transport profile, make the stock config explicitly future-facing until its implementation merges, and use the matrix as the hard gate for #4847. If v1 must remain the interoperable NIP, keep v2 as a separately named extension with its own discovery and evidence contract rather than presenting it as stock NIP-FI. Note that nearly all of the CRITICAL/IMPORTANT documentation findings trace to the newest commit (`aa1bf732`, plus part of `c9efaf6`) — rescoping or reframing that one commit leaves the normative core intact.

Quality gate: Minimalism **7/10** — duplicated transport/runtime claims are already drifting; consolidate the canonical definitions. Elegance **6/10** — a reader cannot derive one stock wire/configuration contract. Correctness **5/10** — the transport mismatch, false present-tense runtime claims, unverifiable subject rule, unsupported rollback claim, and downstream lease/denial incompatibilities block a conforming implementation. All named defects must be resolved before approval.

cea-block:
🤖 The review findings are addressed across four commits:

- Promoted `trusted-proxy-hmac-v2` to the exact optional stock profile and added a private registered-profile seam without exposing deployment-specific details.
- Made the runtime documentation explicitly future-facing and rewrote the configuration contract for multiple domains.
- Restored subject stability as a trust assumption, narrowed JWKS guarantees to current-snapshot revalidation, and defined complete `policy_id` inputs.
- Documented required WebSocket rechecks and lease invalidation as implementation follow-ups.
- Kept final denials out of the authorization audit, separated bounded denial observations, and documented audit-capacity exhaustion as a residual risk.
- Clarified `ReenableIdentity`, current behavior in `SECURITY.md`, and the denial-oracle execution strategy.
- Removed unrelated AWS-LC and pgcrypto build material from this PR in a standalone commit. Both describe requirements already present on main, and the guidance belongs with the build and dependency changes in #4772; that update to #4772 is still pending and will follow.

The 24 traces and 16 invariants remain unchanged. Examples and documentation checks pass. This remains a specification/docs-only PR and does not claim the future runtime is implemented.

Commits: `6636ce9f6`, `e001f1254`, `633260e40`, `a383fd50a`.

wpfleger96 (approved):
🤖 Approving at head `a383fd50`. Re-verified every finding from the earlier review at this exact ref:

- Both CRITICALs resolved: `trusted_proxy_hmac_v2` is now the single normative transport (no remaining `trusted-proxy-hmac-v1` / `NIP-FI-PROXY-1` references anywhere at head), and the runtime configuration document is explicitly a future contract with a stated availability boundary — no live-behavior claims remain.
- All seven IMPORTANTs addressed: subject stability restated as an issuer trust assumption with an operator evidence field; JWKS A→B→A rollback narrowed to documented residual risk with required trace evidence; `policy_id` now digests algorithms, key-source identity, and the verifier-contract fingerprint; leases require per-use revalidation and an invalidation owner; denials move to the separately bounded non-authoritative `O_D` channel (`FI-INV-09`); config is `domains[] × issuers[]` and can express `FI-TRACE-CROSS-DOMAIN-COLLISION`; audit exhaustion is an accepted residual risk and denied operations no longer consume the budget.
- Hygiene re-checked at head: 24 trace / 16 invariant labels consistent across carriers, all JSON examples parse, `git diff --check` clean, still docs-only.

Thanks @cea-block.

PR #1194, opened 2026-06-23
fix(desktop): ground agent workspace, migrate legacy nest, configurable repos_dir

wpfleger96:
## Problem

macOS agents trigger a TCC permission prompt many times a day. Root cause: [#960](https://github.com/block/buzz/pull/960) renamed the agent nest `~/.sprout` -> `~/.buzz` with no migration. Agents boot into an empty `~/.buzz` whose base prompt names a workspace layout (`RESEARCH/`, `PLANS/`, ...) but never its absolute root. Lacking an anchor for where they live, the model runs `find` over `$HOME` to locate files it "remembers" — and macOS attributes that access to Buzz.app, raising the prompt.

A second cost compounds it: agents re-clone repositories into the nest's `REPOS/` because they cannot reach the user's existing checkouts, so the same source tree exists twice and the prompt implies `REPOS/` is the clone root.

image

### Ground the workspace
`framed_system_prompt` receives `ctx.cwd` and prepends a `[Workspace]` section naming the absolute working directory ahead of `[Base]`, removing the reason the model searches `$HOME`. The section is omitted for the `/` fallback (naming `/` would itself invite a `$HOME`-wide scan), for relative/empty paths, and for persona-only agents that never received the layout it grounds.

### Migrate the stranded nest — knowledge only
`migrate_legacy_nest` copies the knowledge entries (`AGENTS.md`, `RESEARCH`, `PLANS`, `GUIDES`, `WORK_LOGS`, `OUTBOX`, `.scratch`) from `~/.sprout` into `~/.buzz` and deliberately skips `REPOS/` — cloned repos are re-clonable, and a real stranded `REPOS/` measured 62 GB, which would block launch for minutes if copied on the startup thread. Migration runs after `ensure_nest()`, is non-fatal and idempotent, and uses per-entry log-and-continue so one failed entry never aborts boot. On a successful migration the backend emits `legacy-nest-migrated`; the frontend shows a one-time toast suggesting the user delete the now-inert `~/.sprout` to reclaim disk.

### Configurable `repos_dir`
Each workspace gains an optional `repos_dir`. When set, `~/.buzz/REPOS` becomes a symlink to that external directory, so agents work in the user's existing checkouts instead of re-cloning. `ensure_repos_symlink` is shared between nest setup (real-dir fallback when unset) and the live `apply_workspace` bridge, so a `repos_dir` edit takes effect on the next save without a restart.

[...]

PR #3730, opened 2026-07-30
fix(buzz-acp): ground the agent workspace on Windows

sumit-m:
## Problem

`workspace_section` decides whether to emit the `[Workspace]` grounding block by testing `cwd.starts_with('/')`. No Windows path satisfies that, so on Windows the section is always `None`: agents are never told their working directory, nor that `AGENTS.md`, `PLANS/`, `RESEARCH/`, `WORK_LOGS/`, `OUTBOX/` and `REPOS/` live there.

The cwd itself is set correctly, so files still land in the right place by default — but the prompt loses the grounding that exists precisely to stop the agent hunting around `$HOME` for those directories.

## Change

- Use `Path::is_absolute` for the absoluteness test. It is target-aware, so `C:\Users\me\.buzz` qualifies. The explicit `cwd != "/"` check stays, because `is_absolute("/")` is true on Unix and that value is the `current_dir()` failure sentinel.
- Join `REPOS` with `Path::join` rather than interpolating a literal `/`, so a Windows nest reads `…\.buzz\REPOS` instead of a mixed-separator path.

## Tests

The existing tests hardcoded `/Users/me/.buzz` with no `cfg` gate, which is why this went unnoticed — on Windows they asserted nothing meaningful. They now go through a `cfg`'d `SAMPLE_ABS_CWD` so the suite exercises the platform it runs on, plus a cross-platform regression guard and one rejection case per platform (`C:foo` drive-relative on Windows, a Windows path on Unix).

`cargo clippy -p buzz-acp --all-targets` clean; `pool::tests` 98/98 on Windows.

### DIFF
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -1287,12 +1287,22 @@ fn framed_system_prompt(
 /// Skips relative paths and the `/` fallback (`std::env::current_dir()` resolves
 /// to `/` on failure): a `/`-rooted workspace line would actively encourage the
 /// `$HOME`-wide scan this section exists to prevent.
+///
+/// Absoluteness is delegated to `Path::is_absolute`, which is target-aware, so a
+/// Windows nest (`C:\Users\me\.buzz`) qualifies. A leading-slash test matched no
+/// Windows path at all, so agents there ran with no workspace grounding and were
+/// never told where `PLANS/` and friends live. `is_absolute` is also false for
+/// the `/` fallback on Windows, but the explicit check stays for Unix.
 fn workspace_section(cwd: &str) -> Option<String> {
-    if cwd != "/" && cwd.starts_with('/') {
+    if cwd != "/" && std::path::Path::new(cwd).is_absolute() {
+        // Joined rather than interpolated with a literal `/` so a Windows nest
+        // reads `C:\Users\me\.buzz\REPOS` instead of a mixed-separator path.
+        let repos = std::path::Path::new(cwd).join("REPOS");
+        let repos = repos.display();
         Some(format!(
             "[Workspace]\nYour absolute working directory is `{cwd}`. All workspace \
              files — `AGENTS.md`, `RESEARCH/`, `PLANS/`, `GUIDES/`, `WORK_LOGS/`, \
-             `OUTBOX/` — and any repositories you clone (under `{cwd}/REPOS/`) live \
+             `OUTBOX/` — and any repositories you clone (under `{repos}`) live \
              here. This is where you already are; do not search `$HOME` or other \
              directories for them."
         ))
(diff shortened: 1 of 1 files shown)


sumit-m:
Flagging an overlap I should have caught before opening this: #3130 (opened 3 days earlier) changes the same predicate in `workspace_section`.

Both PRs replace `cwd.starts_with('/')` with `Path::is_absolute()`, which is the substantive half of this fix — a Windows nest at `C:\Users\me\.buzz-dev` is not `/`-rooted, so the `[Workspace]` section was never emitted at all.

#3130 is the better version of that half. It also rejects filesystem roots generally via `path.parent().is_some()`, covering `C:\`, `\?\C:\` and UNC shares rather than just `/`, and it documents why the platform-dependent behaviour of `is_absolute()` is correct here rather than a bug to revert.

What is unique to this PR is the `REPOS` path separator:

```rust
let repos = std::path::Path::new(cwd).join("REPOS");
```

Interpolating `{cwd}/REPOS` emits `C:\Users\me\.buzz-dev/REPOS` — a mixed-separator path handed to an agent that is explicitly told not to go looking elsewhere. #3130 leaves that line untouched, so merging it alone fixes emission but still ships the mixed separator.

Happy to go whichever way suits: rebase this onto #3130 so it becomes separator-only, or fold the root check from #3130 into this one and close that. No preference beyond not leaving two PRs racing on one function.

mfethe1:
## Confirming this on Windows, and withdrawing a duplicate

I hit this same bug independently and wrote the same fix before finding this PR. Yours was open first — three weeks first — so I have **removed mine from #6170** rather than compete with it. What follows is the Windows validation instead.

Checked out `616720e8c` on Windows 11 / MSVC. Your tests pass:

```
test pool::tests::test_workspace_section_grounds_an_absolute_cwd_and_joins_repos ... ok
test pool::tests::test_workspace_section_rejects_a_drive_relative_windows_cwd ... ok
test pool::tests::test_workspace_section_relative_cwd_is_none ... ok

test result: ok. 3 passed; 0 failed
```

Those assert against `SAMPLE_ABS_CWD`, so I added a throwaway test that feeds `framed_system_prompt` this machine's **real** `current_dir()` and prints what an agent would actually receive:

```
--- cwd: E:\Projects\buzz\.claude\worktrees\acp-test-reliability\crates\buzz-acp
--- framed_system_prompt ---
[Workspace]
Your absolute working directory is `E:\...\crates\buzz-acp`. All workspace files —
`AGENTS.md`, `RESEARCH/`, `PLANS/`, `GUIDES/`, `WORK_LOGS/`, `OUTBOX/` — and any
repositories you clone (under `E:\...\crates\buzz-acp\REPOS`) live here. This is
where you already are; do not search `$HOME` or other directories for them.

[Base]
base text
--- end ---
```

The anchor is present and `REPOS` is joined natively — no mixed `E:\…\buzz/REPOS`. Against a real drive-letter cwd, not a constant.

Here is the same thing captured as before/after on the two branches, which is what the absence looks like — the section simply is not there, with no error and no log line:

before — anchor missing

after — anchor present

That "after" is from my branch rather than yours, so treat it as evidence of the *bug*; the block above is the evidence for *your fix*.

## One small note

`Path::is_absolute` is the right call and I would take it over what I wrote — mine hand-rolled the parse. The one thing that buys is that a hand-rolled check evaluates Windows path shapes identically on Linux, so CI exercises them; `is_absolute` is target-dependent, so your `#[cfg]`-gated tests only ever check the platform they run on, and buzz-acp has no Windows CI job today. Not worth changing the implementation over — but if you wanted one cheap guard, asserting that a `\?\C:\…`-style or UNC cwd is accepted would at least pin the intent in a comment for the next reader.

Also worth knowing, since this PR has been open a while and is unreviewed: it still merges cleanly with `main`, and I verified it merges cleanly with #6141, #5944, #5982 and the narrowed #6170 in any order.

sumit-m:
Closing: the code this patched no longer exists. Workspace grounding now goes through `workspace_section` (1011cea26, #1194) and the PLANS/OUTBOX layout moved into base_prompt.md, so this diff no longer applies. If the Windows path handling still needs attention there, it belongs in a fresh, much smaller change against `workspace_section`.

PR #3416, opened 2026-07-28
fix(desktop): remove Projects overview card fills

thomaspblock:
## Summary

The Projects overview now lets the page surface flow through its metrics and activity cards instead of stacking filled panels. Borders and hover feedback remain, preserving grouping and interaction cues without the heavy nested background.

### Related issue

None found.

### Testing

- `pnpm exec biome check src/features/projects/ui/ProjectsOverviewPanel.tsx src/features/projects/ui/ProjectsActivityFeed.tsx tests/e2e/project-pr-review.spec.ts`
- `pnpm build:e2e`
- Focused Playwright smoke test: `project overview does not paint a background behind its cards` (passed)
- Relevant desktop pre-push checks passed; the unrelated integration gate was blocked by a stale local checksum for migration 25

Issue #3369, opened 2026-07-28
Inline link previews backed by Open Graph meta tags


adamazad:
## Problem

Links posted in channels render as bare URLs. When someone shares a fig (an HTML mockup URL), teammates see only the raw link — no title, no image, no context. They have to open a browser tab to know whether the link is worth looking at.

From #general:
> shouldn't figs show inline tho?

This applies to any shared link, not just figs: PRs, docs, dashboards, articles.

## Proposal

Render an inline link preview card under a message that contains a URL, populated by inspecting the target page's Open Graph meta tags.

**Fetch + parse**
- Server-side fetch of the URL (never from the client — avoids leaking member IPs to arbitrary hosts and lets us cache once per URL).
- Parse `<head>` for, in priority order:
  - `og:title` → fall back to `twitter:title` → `<title>`
  - `og:description` → `twitter:description` → `<meta name="description">`
  - `og:image` → `twitter:image` → `<link rel="icon">`
  - `og:site_name`, `og:type`, `og:url` (canonical)
- Respect `<meta name="robots" content="noindex">`? No — OG tags are explicitly for this. But do respect an explicit opt-out if one is agreed.

**Card shape**
- Favicon + site name, title (1 line), description (2 lines clamped), image thumbnail.
- `og:type` of `image` or an image content-type → render the image directly instead of a card.
- No OG tags and no `<title>` → render nothing, keep the plain link. Never show a broken/empty card.

**Safety / limits** (these are the parts that bite if skipped)
- Block private address space: reject resolved IPs in `10/8`, `172.16/12`, `192.168/16`, `127/8`, `169.254/16`, `::1`, unique-local — SSRF via a pasted link is the main risk here.
- Cap redirects (~5), response size (~1 MB of HTML), and fetch timeout (~5s).
- Only fetch `http`/`https`.
- Proxy and re-encode preview images rather than hotlinking, so the recipient's client never talks to the origin.
- Cache negative results too, so a dead link isn't refetched on every render.

**Scope**
- First link in a message only, to start.
- Sender can remove the preview from their own message.

## Open questions

- Cache TTL — figs change while you iterate on them, so a long TTL shows a stale thumbnail. Short TTL, or a `?v=` cache-buster convention on fig URLs?
- Do figs already emit OG tags? If not, that's a prerequisite on the fig-rendering side and worth splitting into its own issue.
[...]

PR #3316, opened 2026-07-28
feat(desktop): enrich link previews with page metadata

tellaho:
**Category:** improvement
**User Impact:** Messages now show richer, more recognizable previews for supported HTTPS links.

**Problem:** Link cards relied on URL-derived or sender-entered labels, so they often lacked the page context users needed. Generic HTTPS links had no compact preview at all.

**Solution:** Fetch bounded Open Graph, Twitter, or HTML title metadata through the native networking boundary, while preserving provider-specific labels and safe URL fallbacks. Generic links use a bundled globe icon, and unsafe or unavailable metadata fails closed without changing the visible link.

<details>
<summary>File changes</summary>

**desktop/src-tauri/src/commands/link_preview.rs**
Adds hardened, bounded page-metadata retrieval and title/site-name extraction for public HTTPS URLs.

**desktop/src-tauri/src/lib.rs**
Registers the native link-preview metadata command.

**desktop/src/features/communities/useCommunityInit.ts**
Clears ephemeral preview metadata when the active community changes.

**desktop/src/shared/lib/linkPreview.test.mjs**
Covers generic HTTPS previews, exclusions, ordering, deduplication, and title fallbacks.

**desktop/src/shared/lib/linkPreview.ts**
Recognizes generic public HTTPS links while retaining specialized provider previews and existing limits.

**desktop/src/shared/lib/useResolvedLinkPreviews.ts**
Resolves fetched metadata through an ephemeral, deduplicated client cache and merges it into card models.

**desktop/src/shared/ui/link-preview-attachment.tsx**
Uses the bundled globe as the neutral source mark for generic web previews.

</details>

## Reproduction steps

1. Run Buzz Desktop and open a channel containing a GitHub, Linear, or Google link; confirm its existing provider label remains and its title resolves from page metadata when available.
2. Send a public generic `https://` URL and confirm a compact card appears below the visible original link with a globe source icon.
3. Send an `http://` URL and confirm Buzz leaves the link visible without generating a preview card.
4. Switch communities and confirm previews resolve normally without carrying cached metadata across communities.

## Screenshots

Generic source icon before/after: image

PR #3744, opened 2026-07-30
feat(desktop): prototype rich link preview cards for generic URLs

thomaspblock:
## Summary

Prototype of **rich link preview cards** in channel messages, as requested in the `#buzz-frontend` thread (channel `a410ffde-c61f-416a-96e0-c296b5f5ecc9`, thread `d8552e61…`): generic URLs in a message now render an OpenGraph preview card (image, title, description, site name, favicon) below the message.

This *extends* the existing compact provider-specific previews (GitHub/Linear/Google) rather than replacing them — generic cards only render for URLs not covered by a provider card.

### What's in it

- **TS** — `extractGenericLinkPreviewUrls` in `linkPreview.ts` (max 3 per message; skips media-extension URLs, provider-covered URLs, and code/spoiler/image-masked content), `useRichLinkPreviews` hook with in-memory caching, `RichLinkPreviewCard` UI, and a combined `MessageLinkPreviews` component in `link-preview-attachment.tsx` so `markdown.tsx` stays a one-liner (and under its file-size ratchet).
- **Rust** — new `fetch_link_preview_metadata` Tauri command: OG-first parsing with `<title>` fallback, 3-redirect limit, 256 KB / 4 s caps. No title → no card (fails closed).
- **Tooling** — `linkPreviewMetadata` mock in `e2eBridge.ts`; `--mock-config` option for `tests/helpers/screenshot.mjs` so previews are screenshotable deterministically.

### Related issue

Prototypes #3369. Overlaps in spirit with #3316 (metadata for the *compact* provider cards) — this PR is about full rich cards for generic URLs; happy to reconcile if #3316 lands first.

### Testing

- 3787 desktop unit tests pass; `tsc --noEmit`, biome, file-size ratchet clean
- `cargo test link_preview` 6/6 (3 new Rust unit tests for the OG parser)
- Light + dark screenshots below (mocked metadata, deterministic)

_Note: pushed with `--no-verify` for the final push only — all desktop/Rust pre-push hooks passed; the `test` recipe fails locally on an unrelated pre-existing migration-25 checksum issue in my local DB (same as noted in #3416)._


thomaspblock:
## Screenshots

### Rich link preview cards — light (catppuccin-latte)

link-previews-light

### Dark (houston)

link-previews-dark

thomaspblock:
## Updated screenshots — background fill removed (`03d617687`)

Per thread feedback: cards now sit on the theme's normal background (border only, no `bg-muted` fill).

### Light (catppuccin-latte)

link-previews-light

### Dark (houston)

link-previews-dark

thomaspblock:
## Compact variant screenshots

Smaller preview card per feedback: OG image moved from a full-width banner to a 64px thumbnail beside the text; card height is now bounded by the two-line description.

**Light (catppuccin-latte)**

rich-link-preview-compact-light

**Dark (houston)**

rich-link-preview-compact-dark

jedwards27 (changes requested):
**:bot: Jude’s code review agent — REQUEST CHANGES**

Reviewed `02be413b823c356587e6e9f4d07f6cb06bb41c3c..d983c67c002e1939d57e1c41359b972117b88fd7` at exact head `d983c67c002e1939d57e1c41359b972117b88fd7` (clean worktree). This is high risk because displaying channel content now initiates native and webview network traffic chosen by the message author.

### Blocking findings

1. **[P1] Rendering an untrusted message permits recipient-side SSRF.** Every interactive message mounts the preview path (`desktop/src/shared/ui/markdown.tsx:1964`), which extracts up to three arbitrary HTTP(S) URLs and automatically invokes the native command (`desktop/src/shared/lib/linkPreview.ts:572-605`, `desktop/src/shared/lib/useRichLinkPreviews.ts:14-18,52-80`). The command checks only scheme/host and then follows three redirects (`desktop/src-tauri/src/commands/link_preview.rs:28-57`); it never resolves or rejects loopback, RFC1918, link-local, CGNAT, ULA, mapped/translated IPv6, or redirect destinations. A channel member can therefore make every recipient Desktop GET localhost/LAN/cloud-metadata targets merely by opening the channel. This contradicts #3369's explicit server-side-fetch and private-address requirements. The repository already has reserved-address classification in `crates/buzz-core/src/network.rs:22-95`, but safe fetching also needs address validation/pinning at connection time and on every manually handled redirect to resist rebinding. Add positive-public and negative loopback/private/link-local/CGNAT/ULA/mapped-IP/redirect/rebinding tests around the real fetch path.

2. **[P1] Origin-controlled images bypass the native boundary and hotlink from each recipient webview.** `og:image` and favicon values are returned as arbitrary HTTP(S) URLs (`desktop/src-tauri/src/commands/link_preview.rs:95-108,181-186`) and assigned directly to `` (`desktop/src/shared/ui/rich-link-preview-card.tsx:21-50,115`). This leaks recipients' IP/request timing to a sender-controlled tracker and allows a second private-network request path outside the native byte/timeout controls. #3369 specifically requires proxying and re-encoding so recipient clients never contact the origin. Fetch validated public images through a hardened relay/native proxy, bound time/bytes/type/dimensions, re-encode them, and test that the webview never receives the attacker URL and private/redirecting image targets are rejected.

3. **[P2] The advertised four-second bound ends at response headers, so body reads can stall indefinitely.** `tokio::time::timeout` wraps only `request.send()` (`desktop/src-tauri/src/commands/link_preview.rs:54-57`); `read_limited_text(response).await` is outside it (`:73-75,276-290`). A server can send successful HTML headers and then drip or stop before 256 KiB. With up to three requests per message and no global concurrency bound, a malicious timeline can accumulate stalled tasks/sockets. Bound the entire send-plus-body operation, constrain concurrency, and add a server test that sends headers then stalls/drips and proves bounded completion/cleanup.

4. **[P2] The new global cache is unbounded, permanent, cross-community, and makes transient failures permanent until restart.** `metadataCache` stores promises, results, and `null` forever (`desktop/src/shared/lib/useRichLinkPreviews.ts:12-35`) and is absent from the required community reset (`desktop/src/features/communities/useCommunityInit.ts:48-72`). This retains URL/metadata state across community switches, grows with every viewed URL, and prevents retry after an offline/timeout failure. Add bounded TTL/LRU semantics, clear it on community reset, and test switch, expiry, eviction, and transient-failure retry.

### UX and evidence gaps

The issue now requires Desktop/mobile consistency and displaying the destination domain alongside the site name. This PR is Desktop-only and renders `siteName ?? hostname` (`desktop/src/shared/ui/rich-link-preview-card.tsx:102-105`), so an origin-controlled site name can hide the actual domain. The absolute card anchor also has no explicit focus-visible treatment (`:90-125`). Please either align the issue/scope explicitly or implement and regression-test these trust/accessibility requirements.

The added tests cover URL extraction and HTML parsing, not the security/network boundary or rendered workflow. There is no new component/Playwright coverage for asynchronous failure/retry, keyboard focus/activation, narrow layout, domain display, or proving that the webview does not contact origin-controlled assets. Mocked happy-path screenshots do not establish those contracts.

### Validation

At the exact clean head above, `. ./bin/activate-hermit && just desktop-ci` passed end to end: Biome/file guards, Desktop unit suite, TypeScript/Vite build, `cargo fmt --check`, `cargo check`, and full `desktop/src-tauri` tests (lib: 1,885 passed / 14 ignored; diagnostic: 3 passed; 0 failed). CI checks are green for this head, but the Security workflow is skipped and the tests do not exercise the findings above. The PR is currently `CONFLICTING` / `DIRTY` against base `02be413b823c356587e6e9f4d07f6cb06bb41c3c`.

No GUI was launched on the shared machine; there is no checked-in native journey for this feature. These security defects are source-proven and do not depend on GUI reproduction.

PR #3769, opened 2026-07-30
feat(desktop): one relative date ladder across chat and the Inbox

delkc:
Slice 6 of #2216. Independent of #3642 — cut from `main`, no shared files in conflict.

## Why

Five surfaces formatted the same thing five ways, and none of them matched the writing standard's Today / Yesterday / weekday / date progression.

| Surface | Before |
|---|---|
| Chat day divider | `Monday, March 31st` — ordinal suffix, which the standard says to avoid |
| Inbox section header | `Yesterday`, but never `Today`; always printed the year |
| Inbox list row | A third implementation |
| Inbox thread pane header | `Jul 8, 2026, 2:34 PM` — always absolute, always with the year, never relative at any distance |
| Channel message header | `9:05 AM` — a bare clock, so a message from last week has nothing to anchor it once its day divider scrolls away |

There were three separate date implementations doing this, which is the symptom worth naming: **two different jobs were being solved ad hoc at each call site.** A header that labels a *group* of items needs a different label than an individual item's own timestamp.

## What

`shared/lib/datetime.ts` owns both ladders:

### Final patterns

**Desktop**

| When | Day dividers / Inbox headers (`formatDayGroupLabel`) | Message timestamps (`formatItemTimestamp`, `withTime: true`) | Compact Inbox rows (`withTime: false`) |
|---|---|---|---|
| Today | `Today` | `2:34 PM` | `2:34 PM` |
| Yesterday | `Yesterday` | `Yesterday at 2:34 PM` | `Yesterday` |
| 2–6 days ago | `Monday` | `Monday at 2:34 PM` | `Monday` |
| Older, this year | `Saturday, June 20` | `Sat, Jun 20 at 2:34 PM` | `Sat, Jun 20` |
| Previous years | `June 20, 2025` | `Jun 20, 2025 at 2:34 PM` | `Jun 20, 2025` |

**Mobile**

| When | Day divider pill (`formatDayHeading`) | Message rows |
|---|---|---|
| Today | `Today` | `2:34 PM` |
| Yesterday | `Yesterday` | `2:34 PM` |
| 2–6 days ago | `Monday` | `2:34 PM` |
| Older, this year | `Tuesday, March 31` | `2:34 PM` |
| Previous years | `March 31, 2025` | `2:34 PM` |

The weekday rides along within the current year (added from review feedback — it still orients, "was that a weekend?", well past the six-day band) and drops once the year appears, where it no longer earns its width. Weekday and month styles match within a label: spelled out on dividers, abbreviated inline. Future timestamps (clock skew) skip the bare-weekday band so they can't read as the recent past.

## Two deliberate deviations from the standard

Both are documented at the definition, not just here.

**The oldest band keeps the day.** The standard collapses anything over ten months to month-and-year (`Aug 2022`). A group label has to *identify* its day — collapsing would give every day in a month the same divider, so scrolling old history would show a run of identical headers with no way to tell one day from the next. Only the year is conditional. There's a test asserting three consecutive 2022 dates produce three distinct labels.

**Roomy surfaces keep the time of day at every band.** `Yesterday at 9:05 AM`, not `Yesterday`. This is a chat and collaboration workspace rather than a transactional product — where you read conversation, the time is content, not chrome. Narrow list rows still drop it (`withTime: false`) and rely on the existing hover tooltip, which stays the absolute value. `withTime` is a surface decision, not a preference.

Today needs no date word in either mode: a bare clock already reads as today, and "Today at 2:34 PM" is longer without saying more.

## Derived rather than captured

`MessageTimestamp` now takes only `createdAt` and derives both of its labels, instead of receiving a pre-formatted `time` string. A relative label captured when the message list was formatted would be frozen at that wording; deriving it means each render recomputes.

This does not make it live — `MessageRow` is memoized, so a row already on screen when the clock passes midnight keeps saying "Today" until something re-renders it. The day divider above it has always had the same property, and both correct themselves on the next message, scroll, or navigation. Called out in the component doc so the next person doesn't read "derived" as "reactive".

The memo comparator moved from `message.time` to `message.createdAt`. Behavior-identical — `time` was a pure function of `createdAt` — but it now names the prop the row actually reads.

The 36px continuation hover gutter stays clock-only. A relative label doesn't fit in `w-9`.

## Middot between metadata segments

`managed by you 9:53 AM` ran two unrelated facts together as if they were one phrase. Now `managed by you · 9:53 AM`.

- `aria-hidden` — punctuation for the eye only. The header already reads as separate nodes to a screen reader, and `MessageAgentOwner` supplies its own "Agent managed by" label.
- Grouped with the segment it precedes, so it can't wrap to the start of a line on its own — as loose siblings in a `flex-wrap` row, an orphaned divider is exactly what happens.
- No margin; spacing comes from the container gap.
- **No separator after the author name.** "Alice 9:53 AM" already reads as a name followed by a time. Dividers go between metadata segments only.

Middot is already the app's separator for this — `MessageThreadSummaryRow`, the mention list, project rows, 46 files in total.

Applied to the channel message header, channel system rows, and the Inbox thread pane. Left-side Inbox activity rows deliberately unchanged.

## Verified

Screenshots taken through `just desktop-screenshot`:

- `#agents` — `nadia 🤖 managed by you · 10:20 AM`, and the `Today` divider with clock-only rows
- Inbox thread pane — `alice 🤖 owner unavailable · 12:00 PM`

**Gap worth naming:** every mock channel message is same-day, so the past-day labels (`Yesterday at 9:05 AM`, `Jun 20 at 2:34 PM`) are covered by unit tests rather than by a rendered screenshot. Happy to add a spec that seeds an older `created_at` if a reviewer wants to see them.

## Validation

- `pnpm check`, `pnpm typecheck` — clean
- Unit: **3800/3800**, including 17 new tests in `shared/lib/datetime.test.mjs` and 4 in `messageTimestampContract.test.mjs`

The datetime tests pin the things that are easy to regress: Today/Yesterday as *calendar* boundaries rather than 24-hour windows (a message 15 hours old across midnight is "Yesterday"; one 22 hours old on the same day is "Today"), the weekday band bounded at both ends so a future timestamp from clock skew never gets labelled with a past weekday, no ordinals across all the tricky days (1/2/3/11/12/13/21/22/23/31), the year omitted within the current year, and compact labels staying ≤12 chars for a narrow row.

- Smoke E2E: **783 passed, 2 failed, 1 skipped**

Both failures are pre-existing and unrelated, confirmed by re-running each against a clean tree:

1. `video-attachment.spec.ts:223` — fails deterministically on clean `main`
2. `community-rail.spec.ts:797` (keyboard drag-and-drop reorder) — flaky on clean `main`: 2/5 failures there vs 3/5 with this branch, i.e. noise

## Mobile

Mobile had the same divergence, so it moves with desktop rather than drifting until the next pass. `mobile/lib/features/channels/date_formatters.dart`:

| Before | After |
|---|---|
| `formatDayHeading` → Today / Yesterday / `Tuesday, March 31, 2026` | Today / Yesterday / `Tuesday` / `Tuesday, March 31` / `March 31, 2025` |
| `formatThreadSummaryLastReplyTime` → `on May 19th` | `on May 19` |

Same two departures from the standard as desktop, documented at the definition and cross-referenced to `datetime.ts` so the next person editing one finds the other. Day comparison also moved to a rounded start-of-day difference, so a DST transition counts as one calendar day rather than zero — Dart's `Duration.inDays` truncates.

**Message timestamps stay clock-only on mobile.** Desktop message headers now read `Yesterday at 9:05 AM`; mobile keeps `9:05 AM` at every band. That's the compact side of the same surface split the desktop change makes — a mobile timestamp sits inside a chat bubble on a narrow screen with the day divider a short scroll away, where a date word costs width it doesn't earn. Recorded as a decision at `formatMessageTime` so it doesn't read as an oversight.

Mobile needs no middot work: message headers have no "managed by" segment, and the mention suggestion list already uses `\u00b7`.

Validation: `dart format` clean, `flutter analyze` no issues, `flutter test` **911 passed, 1 skipped** — 8 new day-heading tests covering the weekday band, the year boundary, ordinals across 1/2/3/11/12/13/21/22/23/31, distinct labels for consecutive days in the oldest band, and calendar-day rather than 24-hour bands.

## Out of scope

- **Search results.** `SearchResultItem.tsx` and `TopbarSearch.tsx` hand-roll a `5m ago` elapsed format. That's a third *kind* of label — elapsed rather than relative-calendar — and deciding whether search should switch is a separate call.
- **`formatThreadSummaryLastReplyTime`** keeps its own "3 hours ago" elapsed scale on both platforms; only its old-reply fallback lost the ordinal (`on May 19th` → `on May 19`).
- **Mobile search.** `relativeTime` returns `7/31/2026` past a week, matching the desktop search format that's also out of scope above. Both should change together or not at all.

### DIFF
--- a/desktop/src/features/home/lib/inbox.ts
+++ b/desktop/src/features/home/lib/inbox.ts
@@ -18,6 +18,10 @@ import type {
   HomeFeedResponse,
   RelayEvent,
 } from "@/shared/api/types";
+import {
+  formatDayGroupLabel,
+  formatItemTimestamp,
+} from "@/shared/lib/datetime";
 import { resolveMentionProps } from "@/shared/lib/resolveMentionNames";
 
 export type InboxFilter =
@@ -86,6 +90,7 @@ export type InboxReply = {
    */
   signerPubkey?: string;
   tags?: string[][];
+  /** Clock time only, for the hover gutter on continuation rows. */
   timeLabel?: string;
 };
 
@@ -103,11 +108,6 @@ export type InboxGroup = {
 
 type InboxChannel = Pick<Channel, "channelType" | "id" | "name">;
 
-const listTimeFormatter = new Intl.DateTimeFormat("en-US", {
-  hour: "numeric",
-  minute: "2-digit",
-});
-
 const fullTimeFormatter = new Intl.DateTimeFormat("en-US", {
   month: "short",
   day: "numeric",
@@ -115,31 +115,6 @@ const fullTimeFormatter = new Intl.DateTimeFormat("en-US", {
   hour: "numeric",
   minute: "2-digit",
 });
-
-const shortDateFormatter = new Intl.DateTimeFormat("en-US", {
-  month: "short",
-  day: "numeric",
-});
-
-const shortDateWithYearFormatter = new Intl.DateTimeFormat("en-US", {
-  month: "short",
-  day: "numeric",
-  year: "numeric",
-});
-
-const weekdayFormatter = new Intl.DateTimeFormat("en-US", {
-  weekday: "long",
-});
-
-function startOfDay(value: Date) {
-  return new Date(value.getFullYear(), value.getMonth(), value.getDate());
-}
-
-function diffInDays(from: Date, to: Date) {
-  return Math.round(
-    (startOfDay(from).getTime() - startOfDay(to).getTime()) / 86_400_000,
-  );
-}
 
 function tagValue(item: FeedItem, name: string) {
   return item.tags.find((tag) => tag[0] === name)?.[1]?.trim() || null;
@@ -445,23 +420,7 @@ export function findInboxItemByEventId(
 }
 
 function formatInboxTimestamp(unixSeconds: number) {
-  const date = new Date(unixSeconds * 1_000);
-  const now = new Date();
-  const dayDiff = diffInDays(now, date);
-
-  if (dayDiff === 0) {
-    return listTimeFormatter.format(date);
-  }
-
-  if (dayDiff === 1) {
-    return "Yesterday";
-  }
-
-  if (now.getFullYear() === date.getFullYear()) {
-    return shortDateFormatter.format(date);
-  }
-
-  return shortDateWithYearFormatter.format(date);
+  return formatItemTimestamp(unixSeconds);
 }
 
 export function formatInboxFullTimestamp(unixSeconds: number) {
@@ -480,21 +439,14 @@ export function relayEventFromFeedItem(item: FeedItem): RelayEvent {
   };
 }
 
-export function groupInboxItems(items: InboxItem[]): InboxGroup[] {
+export function groupInboxItems(
+  items: InboxItem[],
+  nowSeconds = Date.now() / 1_000,
+): InboxGroup[] {
   const groups = new Map<string, InboxItem[]>();
-  const now = new Date();
 
   for (const item of items) {
-    const date = new Date(item.latestActivityAt * 1_000);
-    const dayDiff = diffInDays(now, date);
-    const label =
-      dayDiff === 0
-        ? "Today"
-        : dayDiff === 1
-          ? "Yesterday"
-          : dayDiff < 7
-            ? weekdayFormatter.format(date)
-            : shortDateWithYearFormatter.format(date);
+    const label = formatDayGroupLabel(item.latestActivityAt, nowSeconds);
 
     const current = groups.get(label) ?? [];
     current.push(item);
--- a/desktop/src/features/home/ui/InboxMessageRow.tsx
+++ b/desktop/src/features/home/ui/InboxMessageRow.tsx
@@ -4,10 +4,12 @@ import { useKnownAgentPubkeys } from "@/features/agents/useKnownAgentPubkeys";
 import type { InboxContextMessage } from "@/features/home/lib/inbox";
 import { toTimelineMessage } from "@/features/home/lib/inboxViewHelpers";
 import { formatTimeWithoutDayPeriod } from "@/features/messages/lib/dateFormatters";
+import { formatItemTimestamp } from "@/shared/lib/datetime";
 import type { TimelineMessage } from "@/features/messages/types";
 import { getConfigNudgeAuthorPubkey } from "@/features/messages/ui/configNudgeAuthPubkey";
 import { MessageActionBar } from "@/features/messages/ui/MessageActionBar";
 import { MessageAgentOwner } from "@/features/messages/ui/MessageAgentOwner";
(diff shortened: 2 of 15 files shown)


Chessing234:
helpful. one ask — keep the failure message short enough for mobile so it doesnt blow the toast

delkc:
🤖 Posted by Clay's AI agent.

## Mobile day dividers — before / after

Widget renders at phone width with the real fonts and theme (not device
screenshots), showing the divider pill labels this PR changes.

Current — no weekday-only band, and the year always printed:

current

New — `Tuesday` / `April 16` / `March 31, 2025`, matching desktop:

new

Real in-app screenshots to follow. Desktop shots removed from this comment.

delkc:
Changing relative timestamp to "Yesterday" instead of the date. Today's just get the time. 
Screenshot 2026-07-30 at 9 31 26 AM


Adding a middot between the "managed by" and the date for agents to break up those lines. This also aligns with the existing middot for name and timestamp on mobile. 

Before:
Screenshot 2026-07-30 at 9 41 56 AM

After:
Screenshot 2026-07-30 at 10 23 03 AM

Mobile (unchanged)
Screenshot 2026-07-31 at 5 24 07 PM

delkc:
Adding a few screenshots that show the mix of days/dates/times and a table of the full breakdown. As it gets longer, I dropped the day and month down to the 3-letter abbreviations. 

Mobile follows the same patterns as today, with the pill divider between days and just showing the time on the message itself.

Screenshot 2026-08-14 at 10 13 19 AM


Screenshot 2026-08-14 at 10 08 31 AM
Screenshot 2026-08-14 at 10 09 29 AM
Screenshot 2026-08-14 at 10 10 07 AM
Screenshot 2026-08-14 at 10 04 35 AM

tellaho (approved):
Reviewed by Carl (AI agent) at Taylor Ho's explicit request.

The date ladder is coherent, boundary handling is well tested, and CI is green. This is primarily a frontend/UI PR: desktop React/TypeScript plus mobile Flutter/Dart formatting and tests; it does not change backend behavior.

Approved with non-blocking follow-ups from the team:
- Prefer a rendered/component test over the source-regex contract in `messageTimestampContract.test.mjs` when practical.
- Use valid phrasing/DOM semantics for grouped timestamp metadata (`MessageTimestamp` currently renders a `<p>` inside span-based wrappers at some agent call sites).
- Consider bringing the mobile Inbox's separately hand-rolled date labels onto the same ladder, or explicitly document that scope distinction.
- Refresh the PR description's stale “What” table to include the weekday on older current-year dates.
- The `MessageAgentOwner` icon/baseline polish is adjacent rather than essential to the date change; small enough not to block this PR.

PR #3778, opened 2026-07-30
fix(desktop): back/forward via keyboard chords, mouse X1/X2 buttons, and swipe gestures

### LINKED ISSUE #3775: Back/forward keyboard shortcuts are swallowed whenever the composer has focus (i.e. almost always)
**Problem**

The global back/forward keyboard shortcuts (`⌘[` / `⌘]` on macOS, `Alt+←` / `Alt+→` on Windows/Linux) almost never fire in real use. Users report "the back and forward shortcuts aren't actually going back and forward" and fall back to clicking the toolbar chevrons.

**Root cause**

Two behaviors interact:

1. `useBackForwardControls.ts` bails out of its window keydown handler whenever the event target is editable (`isEditableTarget` — inputs, textareas, contenteditable):

   https://github.com/block/buzz/blob/61b96c9828d1dd54106b570d87a54edbc92bb9c4/desktop/src/app/navigation/useBackForwardControls.ts#L83-L86

2. `useComposerAutofocus.ts` deliberately focuses the message composer (a ProseMirror contenteditable) on mount and on every channel switch, matching Slack/Discord/Signal:

   https://github.com/block/buzz/blob/61b96c9828d1dd54106b570d87a54edbc92bb9c4/desktop/src/features/messages/lib/useComposerAutofocus.ts#L1-L25

Net effect: in the app's steady state focus is nearly always inside the composer, so the chords are silently swallowed. The feature effectively only works in the brief moments nothing editable has focus — which reads as "the shortcut doesn't exist."

The regression is invisible to CI because the e2e coverage in `navigation.spec.ts` only clicks the `global-back` / `global-forward` **buttons**, never presses the keyboard chords.

**Proposed fix**

Match the chord *before* the editable-target guard and let the back/forward chords fire even when focus is in an editable element:

- `⌘[` / `⌘]` have no text-editing semantics in macOS text fields, and the TipTap/StarterKit editor config binds no `Mod-[` / `Mod-]` shortcuts (verified in `useRichTextEditor.ts` — list indentation uses Tab/Shift-Tab).
- `Alt+←` / `Alt+→` have no standard text-editing meaning on Windows/Linux (word movement is Ctrl+arrows).
- Call `preventDefault()` so the editor/webview never sees the chord.

This matches browsers and Slack, where back/forward chords work while a text field is focused.

**Tests**

- Extract chord matching into a pure helper with unit tests (`.test.mjs` convention).
- Add an e2e case that presses the chord **while the composer is focused** — the missing regression coverage.

**Duplicate check:** searched open issues/PRs for "back forward", "navigation history", "shortcut", "navigate" — nothing tracks this. #3078 / #3377 are next/previous-*channel* navigation, a different feature.

I have a fix ready and will open a PR referencing this issue.

matheuskiser:
## Problem

Two related gaps in global back/forward navigation. Fixes #3775.

1. The keyboard shortcuts almost never fire in real use — users fall back to clicking the toolbar chevrons and assume the shortcuts don't exist.
2. On macOS, mouse back/forward buttons (X1/X2) and horizontal swipe gestures do nothing, although they navigate in every browser and in Slack.

**Duplicate check:** searched open PRs and issues — none found beyond #3775 (filed alongside this fix). #3078 / #3377 are next/previous-*channel* navigation, a different feature.

## Root causes

**Keyboard:** `useBackForwardControls`'s keydown handler bailed whenever the event target was editable — but `useComposerAutofocus` deliberately focuses the message composer (a ProseMirror contenteditable) on mount and on every channel switch. In steady state focus almost always lives in the composer, so the chords were silently swallowed. Invisible to CI because `navigation.spec.ts` only ever clicked the `global-back` / `global-forward` buttons, never pressed the keys.

**Mouse/swipe:** on macOS, WKWebView never delivers X1/X2 button events or swipe gestures to the page (Safari handles them natively in the app layer, not in page JS), and Buzz had no native handler.

### Keyboard chords (web layer)

Match the existing platform chord regardless of the event target and drop the editable-target guard:

- `⌘[` / `⌘]` have no text-editing semantics in macOS text fields, and the TipTap/StarterKit editor config binds no `Mod-[` / `Mod-]` shortcuts (checked `useRichTextEditor.ts` — list indentation is Tab/Shift-Tab).
- `preventDefault()` keeps the chord out of the editor — asserted in the e2e test.

This matches browsers and Slack, where back/forward chords work while a text field is focused. Chord matching is extracted into a pure helper, `app/navigation/backForwardChords.ts`, so it can be unit tested; behavior (bindings, modifier exclusivity, `code`-based matching for non-US layouts) is unchanged.

### macOS mouse buttons and swipe gestures (native layer)

An NSEvent local monitor in `mouse_nav.rs` catches what the webview can't see and emits a `mouse-nav` Tauri event to the main window (`emit_to`, so navigation stays scoped if multi-window ever lands) that the frontend acts on. Two AppKit event shapes map to navigation:

- `otherMouseUp` with button 3/4 — mice whose X1/X2 buttons arrive as plain button events. These are swallowed after emitting so nothing downstream double-handles them.
- `swipe` with a horizontal delta — AppKit's page-swipe gesture (`swipeWithEvent:`): `deltaX > 0` back, `deltaX < 0` forward. Sent by mouse drivers that synthesize a page-swipe gesture for the back/forward buttons instead of button-3/4 events (the hardware this was verified on). Stock Apple trackpad and Magic Mouse swipes arrive as phased scroll-wheel events instead, which this PR does not handle — that path (`ScrollWheel` + `trackSwipeEventWithOptions:`, which also needs scroll-edge detection) is deferred to a follow-up. Swipes are passed through (swallowing mid-gesture events could confuse AppKit gesture tracking).

The swipe path was verified end to end on hardware whose back/forward buttons emit only swipe gestures, never button-3/4 events — an instrumented event monitor confirmed the events arrive as `NSEventType::Swipe` with `deltaX ±1`, and navigation worked after mapping them.

## Tests

- **13 unit tests** for the web-side chord matcher (`backForwardChords.test.mjs`): supported chords, modifier exclusivity, `code` fallback, and preservation of line-editing shortcuts.
- **6 Rust unit tests** for the native mapping helpers (`mouse_nav.rs`): button 3/4 directions, other buttons ignored, swipe delta sign → direction, zero-delta (gesture-begin) ignored.
- **e2e regression case** in `navigation.spec.ts`: presses the platform chord *while the composer is focused* — the missing coverage. Verified it fails against the pre-fix implementation and passes with the fix.
- Full desktop unit suite: 3832/3832 pass. Full Rust suite (`cargo test`, buzz-desktop): 1888 passed / 0 failed. `pnpm typecheck`, `biome check`, `pnpm check`, `cargo fmt --check`, `cargo clippy`: clean (no new warnings).
- Full Playwright e2e: 958 passed; 6 failures are relay-infrastructure tests (live relay seeding / relay state seam) that fail identically without this change — `navigation.spec.ts` is fully green.

## Manual test

1. Open a channel, then another (composer autofocuses on each switch).
2. `⌘[` — returns to the previous channel; `⌘]` — forward again. Typing `[` / `]` in the composer inserts normally.
3. Mouse back/forward buttons navigate the same way, from anywhere in the window (verified on macOS on hardware using both event shapes).

## Update — 2026-07-31

Removed the redundant DOM mouse-button handler after verifying it was unnecessary. The native macOS path remains unchanged and was revalidated manually.

### DIFF
--- a/desktop/src-tauri/src/mouse_nav.rs
+++ b/desktop/src-tauri/src/mouse_nav.rs
@@ -10,10 +10,13 @@
 //! - `otherMouseUp` with button 3/4 — mice whose X1/X2 buttons reach the app
 //!   as plain mouse buttons.
 //! - `swipe` with a horizontal delta — AppKit's page-swipe gesture
-//!   (`swipeWithEvent:`): `deltaX > 0` is back, `deltaX < 0` is forward,
-//!   the convention Safari follows. Trackpad two-finger swipes arrive this
-//!   way, and many mouse drivers synthesize the same gesture for the
-//!   back/forward buttons instead of button-3/4 events.
+//!   (`swipeWithEvent:`): `deltaX > 0` is back, `deltaX < 0` is forward.
+//!   Sent by mouse drivers that synthesize a page-swipe gesture for the
+//!   back/forward buttons instead of button-3/4 events (the hardware this
+//!   was verified on). Stock Apple trackpad and Magic Mouse swipes arrive
+//!   as phased scroll-wheel events instead, which this module does not
+//!   handle — that path (`ScrollWheel` + `trackSwipeEventWithOptions:`,
+//!   which also needs scroll-edge detection) is a follow-up.
 //!
 //! Windows/Linux are unaffected: WebView2 and WebKitGTK deliver X1/X2 as
 //! ordinary DOM mouse events, which the frontend `mouseup` listener handles.
@@ -58,14 +61,22 @@ pub fn init(app_handle: &tauri::AppHandle) {
         match ev.r#type() {
             NSEventType::OtherMouseUp => {
                 if let Some(direction) = direction_for_button(ev.buttonNumber()) {
-                    let _ = app.emit("mouse-nav", direction);
-                    // Swallow the event: nothing downstream should also act on it.
+                    // Emit to the main window explicitly instead of
+                    // broadcasting (`emit`) so navigation stays scoped if
+                    // multi-window ever lands. "main" is the default label
+                    // for the single configured window (see deep_link.rs).
+                    let _ = app.emit_to("main", "mouse-nav", direction);
+                    // Swallow the release: nothing downstream should also act
+                    // on it. The matching press deliberately passes through:
+                    // WKWebView never delivers X1/X2 to the page, so the
+                    // unmatched down is inert, and swallowing presses risks
+                    // interfering with AppKit behaviors keyed off mouse-down.
                     return std::ptr::null_mut();
                 }
             }
             NSEventType::Swipe => {
                 if let Some(direction) = direction_for_swipe(ev.deltaX()) {
-                    let _ = app.emit("mouse-nav", direction);
+                    let _ = app.emit_to("main", "mouse-nav", direction);
                 }
                 // Pass swipes through: nothing else navigates on them, and
                 // swallowing mid-gesture events could confuse AppKit's
--- a/desktop/src-tauri/src/mouse_nav.rs
+++ b/desktop/src-tauri/src/mouse_nav.rs
@@ -18,12 +18,11 @@
 //!   handle — that path (`ScrollWheel` + `trackSwipeEventWithOptions:`,
 //!   which also needs scroll-edge detection) is a follow-up.
 //!
-//! This module is a no-op on non-macOS targets; their mouse-button behavior
-//! is intentionally left to the underlying webview.
+//! Compiled macOS-only (via `tray_menu`). Non-macOS X1/X2 behavior is left
+//! to the underlying webview.
 
 /// Maps an `otherMouseUp` button number to a navigation direction.
 /// Buttons 3 and 4 are X1 (back) and X2 (forward).
-#[cfg_attr(not(target_os = "macos"), allow(dead_code))]
 fn direction_for_button(button: isize) -> Option<&'static str> {
     match button {
         3 => Some("back"),
@@ -36,7 +35,6 @@ fn direction_for_button(button: isize) -> Option<&'static str> {
 /// following the AppKit `swipeWithEvent:` convention: positive is back,
 /// negative is forward. A swipe arrives as a begin/end pair and only the
 /// end event carries the direction, so `deltaX == 0` maps to `None`.
-#[cfg_attr(not(target_os = "macos"), allow(dead_code))]
 fn direction_for_swipe(delta_x: f64) -> Option<&'static str> {
     if delta_x > 0.0 {
         Some("back")
@@ -47,8 +45,7 @@ fn direction_for_swipe(delta_x: f64) -> Option<&'static str> {
     }
 }
 
-#[cfg(target_os = "macos")]
-pub fn init(app_handle: &tauri::AppHandle) {
+pub fn init<R: tauri::Runtime>(app_handle: &tauri::AppHandle<R>) {
     use block2::RcBlock;
     use objc2_app_kit::{NSEvent, NSEventMask, NSEventType};
     use tauri::Emitter;
@@ -104,9 +101,6 @@ pub fn init(app_handle: &tauri::AppHandle) {
         eprintln!("buzz-desktop: mouse-nav: failed to install NSEvent monitor");
     }
 }
-
-#[cfg(not(target_os = "macos"))]
-pub fn init(_app_handle: &tauri::AppHandle) {}
 
 #[cfg(test)]
 mod tests {
--- a/desktop/src-tauri/src/mouse_nav.rs
+++ b/desktop/src-tauri/src/mouse_nav.rs
@@ -1,0 +1,137 @@
+//! Native macOS handler for back/forward navigation inputs (mouse X1/X2
+//! buttons and horizontal swipe gestures).
+//!
+//! WKWebView never delivers these inputs to the web content layer, so a DOM
+//! listener can't see them (Safari itself handles them natively in the app
+//! layer, not in the page). This module installs an NSEvent local monitor
+//! and emits a `mouse-nav` Tauri event that `useBackForwardControls` acts on
+//! in the frontend. Two event shapes map to navigation:
+//!
+//! - `otherMouseUp` with button 3/4 — mice whose X1/X2 buttons reach the app
+//!   as plain mouse buttons.
+//! - `swipe` with a horizontal delta — AppKit's page-swipe gesture
+//!   (`swipeWithEvent:`): `deltaX > 0` is back, `deltaX < 0` is forward,
+//!   the convention Safari follows. Trackpad two-finger swipes arrive this
+//!   way, and many mouse drivers synthesize the same gesture for the
+//!   back/forward buttons instead of button-3/4 events.
+//!
+//! Windows/Linux are unaffected: WebView2 and WebKitGTK deliver X1/X2 as
+//! ordinary DOM mouse events, which the frontend `mouseup` listener handles.
+
+/// Maps an `otherMouseUp` button number to a navigation direction.
+/// Buttons 3 and 4 are X1 (back) and X2 (forward).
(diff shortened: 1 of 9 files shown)


Chessing234:
thanks for covering mouse back too. one ask — keep editable targets from eating cmd-left/right so line-start editing still works

matheuskiser:
> thanks for covering mouse back too. one ask — keep editable targets from eating cmd-left/right so line-start editing still works

@Chessing234 good call, but this PR doesn't mess with those command shortcuts. I added a test and validated in the dev env.

wpfleger96 (commented):
🤖 synthesized findings from two independent agent reviews at head `04917b4` (full write-up in our review channel). the keyboard-chord half of this PR is clean — chord matching in a pure helper with modifier-exclusivity tests, the editable-target guard removal is correctly argued, and the e2e regression test reproduces the exact bug. the inline comments below are about the native gesture layer and the unverified Windows path: one blocking gap on the macOS swipe event class, one pre-merge Windows verification ask, and three nits.

desktop/src-tauri/src/mouse_nav.rs
```diff
@@ -0,0 +1,137 @@
...
+
+        match ev.r#type() {
+            NSEventType::OtherMouseUp => {
+                if let Some(direction) = direction_for_button(ev.buttonNumber()) {
+                    let _ = app.emit("mouse-nav", direction);
+                    // Swallow the event: nothing downstream should also act on it.
+                    return std::ptr::null_mut();
+                }
+            }
+            NSEventType::Swipe => {
```
wpfleger96, line 74: 🤖 this monitor only subscribes to `NSEventType::Swipe`, but stock two-finger trackpad and Magic Mouse navigation doesn't actually arrive as swipe events — it comes in as phased, precise `NSEventTypeScrollWheel` events that apps convert into a page swipe via `trackSwipeEventWithOptions:`. That's how both WebKit and Chromium implement history swipes:

- [WebKit starts page-swipe recognition from phased wheel events, then calls `trackSwipeEventWithOptions`](https://github.com/WebKit/WebKit/blob/9a90b5af0558a9b1bfa2f6500cfc1efcf51bd703/Source/WebKit/UIProcess/mac/ViewGestureControllerMac.mm#L272-L321)
- [Chromium documents trackpad and Magic Mouse inputs as scroll-wheel events](https://github.com/chromium/chromium/blob/1030730280f3c19d79357d67d39f82bf4e4fb66e/chrome/browser/renderer_host/chrome_render_widget_host_view_mac_history_swiper.mm#L63-L95) and [handles the Magic Mouse path with `trackSwipeEventWithOptions`](https://github.com/chromium/chromium/blob/1030730280f3c19d79357d67d39f82bf4e4fb66e/chrome/browser/renderer_host/chrome_render_widget_host_view_mac_history_swiper.mm#L422-L520)

`Swipe` events do fire for mouse drivers that synthesize the gesture (which matches the hardware this was verified on), but I don't think the PR's trackpad/Magic Mouse claim holds on stock Apple hardware — and nothing in CI can catch it since the Rust tests feed fabricated deltas to the pure helper.

two ways to resolve, either works for me:
- (a) rescope the claim — keep `Swipe` for the driver-synthesized case and X1/X2 buttons, strike trackpad/Magic Mouse from the PR body + manual-test section, and defer the `ScrollWheel` tracking path to a follow-up. cheapest, keeps everything already verified.
- (b) implement the standard path — phased horizontal `ScrollWheel` + `trackSwipeEventWithOptions`, validated on a stock Apple trackpad and Magic Mouse. heads up if you go this way: it needs scroll-edge detection too — Safari/Chrome only navigate when horizontal content is already at its edge, otherwise a swipe over a wide code block or table hijacks scrolling into navigation.

I'd lean (a) for this PR

desktop/src/app/navigation/useBackForwardControls.ts
```diff
@@ -81,50 +74,81 @@ export function useBackForwardControls() {
...
-        event.key === "ArrowRight";
-
-    if (isBackShortcut) {
+    if (direction === "forward") {
+      event.preventDefault();
+      goForward();
+    }
+  });
+
+  const handleMouseUp = React.useEffectEvent((event: MouseEvent) => {
```
wpfleger96: 🤖 this DOM `mouseup` handler is the entire Windows/Linux fix, but all the hardware verification on the PR was macOS-only. two specific risks worth closing before merge:

1. WebView2 may already handle X1/X2 back/forward natively (Edge-style) — if so, native nav + `router.history.back()` double-fires
2. `preventDefault()` on `mouseup` isn't a reliable suppressor of engine-level button navigation

ask: verify once on Windows hardware before merge, and consider guarding this listener with `isTauri()` — in a plain-browser context (dev, e2e) the browser navigates on these buttons itself, so the guard removes the only other double-fire surface for free

desktop/src-tauri/src/mouse_nav.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    let app = app_handle.clone();
+    let block = RcBlock::new(move |event: std::ptr::NonNull<NSEvent>| -> *mut NSEvent {
+        // SAFETY: the monitor hands us a valid NSEvent for the matched mask.
+        let ev = unsafe { event.as_ref() };
+
+        match ev.r#type() {
+            NSEventType::OtherMouseUp => {
+                if let Some(direction) = direction_for_button(ev.buttonNumber()) {
+                    let _ = app.emit("mouse-nav", direction);
```
wpfleger96: 🤖 nit: `app.emit` broadcasts to all windows. fine single-window today and matches the `deep_link.rs` precedent — just flagging that this should become `emit_to` the focused window if multi-window ever lands

desktop/src-tauri/src/mouse_nav.rs
```diff
@@ -0,0 +1,137 @@
...
+                    // Swallow the event: nothing downstream should also act on it.
+                    return std::ptr::null_mut();
```
wpfleger96, line 71: 🤖 nit: this swallows `OtherMouseUp` for buttons 3/4 but the matching `OtherMouseDown` passes through, so anything downstream sees an unmatched press. no observed harm, just noting for the record

desktop/src/app/navigation/backForwardChords.test.mjs
```diff
@@ -0,0 +1,158 @@
+import assert from "node:assert/strict";
+import test from "node:test";
```
wpfleger96, line 2: 🤖 nit: the PR body says 17 unit tests for this file but it runs 16 (verified locally: 16 pass / 0 fail)

desktop/src-tauri/src/mouse_nav.rs
matheuskiser: Thanks Will! I'll go the route of A since I don't have some of the hardware needed to prove this fix on there.

desktop/src/app/navigation/useBackForwardControls.ts
matheuskiser: Tested on Windows with and without my change. I noticed this was working as expected without my change, which is good. I've also noticed that it worked fine with my change as well, no double fires. Having said that, I'll just remove the windows extra logic from this PR since it wasn't needed. 

I also added the `isTauri()` guard like you suggested. Thanks

wpfleger96 (dismissed):
🤖 I verified the changes at `08abf603a` — the macOS scope is now accurately documented with the `ScrollWheel` + `trackSwipeEventWithOptions:` path deferred to a follow-up, the Windows `mouseup` handler removal checks out (no dangling references to `matchBackForwardMouseButton` or the listener), both emits are scoped to the main window, and the test counts line up. Also ran the desktop unit suite, the Tauri Rust suite, and typecheck locally at this exact commit — all green.

Approving. Note the full CI workflows haven't run on this head yet, so we'll still need those green before merge.

wpfleger96 (approved):

PR #3782, opened 2026-07-30
fix(desktop): stop hiding repos re-announced after a deletion

### LINKED ISSUE #3760: Desktop: Repositories list hides re-announced repos forever — deletion filter ignores `created_at`
## Summary

`isDeletedByA` in `desktop/src/features/projects/hooks.ts` hides a project card when any same-author kind:5 deletion references its `30617:<owner>:<dtag>` coordinate:

```ts
return deletionEvents.some(
  (event) =>
    event.pubkey.toLowerCase() === project.owner.toLowerCase() &&
    event.tags.some((tag) => tag[0] === "a" && tag[1] === coordinate),
);
```

There is no `created_at` comparison. Under NIP-09, a deletion applies to events that existed when it was published — a **newer** replaceable event at the same coordinate is live. The relay gets this right (the newer 30617 is stored, served, and the git endpoint works); Desktop hides it forever. Once an owner has ever deleted a project (including via Desktop's own "Delete project" button), no re-announcement at that coordinate will ever render again for any user, even though clone/push/CLI all work.

## Repro

1. Announce a repo (kind:30617, dtag `X`).
2. Delete it (kind:5 with `a` = `30617:<owner>:X` — Desktop's Delete project button does this).
3. Announce dtag `X` again (fresh 30617, newer `created_at`; relay accepts and repo works over git).
4. Desktop → Repositories: the repo never appears. `fetchProjects` filters it out via `isDeletedByA`.

Hit in the wild while recovering from #3527: both the repo owner's re-announcement and an agent's re-announcement of `bitchat` are live on the relay and clone-able, but invisible in Desktop; only never-deleted test repos render.

## Suggested fix

Only honor deletions at least as new as the announcement being filtered:

```ts
event.created_at >= project.createdAt
```

(plus the existing author check). Same consideration applies anywhere else kind:5 a-tag filtering is done client-side against replaceable events.

## Related UI gap in the same flow

The Delete-project button is gated on strict key equality (`isProjectOwnedByCurrentUser` in `desktop/src/features/projects/lib/projectsViewHelpers.ts` → `canDelete` in `ProjectsView.tsx`), but the relay's `validate_standard_deletion_event` (`crates/buzz-relay/src/handlers/side_effects.rs`) explicitly also accepts "the owning human deleting their agent's events" (`is_agent_owner`). So a user cannot delete their own managed agent's repos from Desktop even though the relay would authorize it. Desktop's client-side deletion filter (`isDeletedByA`) has the same blind spot: it only honors deletions whose pubkey equals the announcement author, so a relay-accepted owner-of-agent deletion would still leave the card rendered until the relay drops the announcement from REQs.

## Related

- #3527 (unbound-repo 404s) — this bug compounds that recovery path: users who deleted a broken repo while debugging can never see it again after re-announcing.
- The stranded `git_repo_names` reservation (separate follow-up on #3527) forces recovery re-announcements in the first place.

daxdax89:
Fixes #3760.

## Problem

`isDeletedByA` hid a project card whenever any same-author kind:5 tombstone carried its `30617:<owner>:<dtag>` coordinate, with no timestamp comparison:

```ts
event.pubkey.toLowerCase() === project.owner.toLowerCase() &&
event.tags.some((tag) => tag[0] === "a" && tag[1] === coordinate)
```

Repo announcements are addressable, so the coordinate outlives any single event at it. Deleting a repo and announcing the same dtag again is a legitimate recovery path, and NIP-09 scopes an `a`-tag deletion to versions "up to the `created_at` timestamp of the deletion request event" — a tombstone must not hide an announcement published after it.

The relay already gets this right, which is what made the failure confusing to diagnose: the newer announcement is stored, served over REQ, and clone/push works, while Desktop's Repositories list never renders the card again for anyone. Once an owner had ever deleted a project (including via Desktop's own Delete project button), no re-announcement at that coordinate could come back.

## Fix

Bound the match to `event.created_at >= project.createdAt`, keeping the existing author check. The comparison is inclusive, matching NIP-09's "up to the `created_at` timestamp".

The predicate and `projectCoordinate` move from `hooks.ts` into `lib/projectDeletions.ts`. Two reasons: the predicate was unexported and therefore untestable, and `hooks.ts` was at 997 of the 1000 lines `desktop/scripts/check-file-sizes.mjs` allows, so the fix could not have been added in place. `hooks.ts` is now 986 lines and both call sites (`fetchProjects`, `fetchProject`) are unchanged apart from the import.

I kept the `isDeletedByA` name so it still matches the issue.

## Tests

`lib/projectDeletions.test.mjs` — 8 cases covering the regression (tombstone older than the announcement no longer hides it), a newer tombstone still hiding it, and the inclusive same-timestamp boundary. The author check, case-insensitive pubkey match, and coordinate matching already worked and are pinned so a later refactor cannot quietly drop them.

Verified the suite catches the bug: with the `created_at` comparison removed, exactly one test fails — "a tombstone older than the announcement does not hide it".

## Checks

`just desktop-check`, `just desktop-typecheck`, `just desktop-test` (3827 pass), and `just desktop-build` all pass locally.

## Out of scope

The issue's second half is deliberately not addressed here, since it needs the managed-agent ownership relationship on the client and is a larger change:

- `isDeletedByA` only honors deletions whose pubkey equals the announcement author, so a relay-accepted owner-of-agent deletion (`is_agent_owner` in `validate_standard_deletion_event`) still leaves the card rendered.
- `canDelete` in `ProjectsView.tsx` gates on strict key equality via `isProjectOwnedByCurrentUser`, so a user cannot delete their own managed agent's repos from Desktop even though the relay would authorize it.

Happy to follow up on those separately if useful.

## Related PRs

Searched open PRs and issues for duplicates; found none touching `isDeletedByA` or repo-deletion filtering.

### DIFF
--- a/desktop/src/features/projects/hooks.ts
+++ b/desktop/src/features/projects/hooks.ts
@@ -3,6 +3,10 @@ import * as React from "react";
 
 import { relayClient } from "@/shared/api/relayClient";
 import { getRelaySelf } from "@/features/moderation/lib/relaySelf";
+import {
+  isDeletedByA,
+  projectCoordinate,
+} from "@/features/projects/lib/projectDeletions";
 import { getCachedRelayOrigin } from "@/shared/lib/mediaUrl";
 import { signRelayEvent } from "@/shared/api/tauri";
 import { getIdentity } from "@/shared/api/tauriIdentity";
@@ -144,10 +148,6 @@ function getCloneUrls(event: RelayEvent): string[] {
   return tag ? tag.slice(1) : [];
 }
 
-function projectCoordinate(project: Pick<Project, "owner" | "dtag">): string {
-  return `${KIND_REPO_ANNOUNCEMENT}:${project.owner}:${project.dtag}`;
-}
-
 function readHiddenProjectCards(): string[] {
   if (typeof window === "undefined") {
     return [];
@@ -167,17 +167,6 @@ function readHiddenProjectCards(): string[] {
 
 function isHiddenLocally(project: Project): boolean {
   return readHiddenProjectCards().includes(projectCoordinate(project));
-}
-
-function isDeletedByA(project: Project, deletionEvents: RelayEvent[]): boolean {
-  const coordinate = projectCoordinate(project);
-  // NIP-09: a deletion is only valid when signed by the author of the
-  // referenced event — otherwise anyone could hide someone else's project.
-  return deletionEvents.some(
-    (event) =>
-      event.pubkey.toLowerCase() === project.owner.toLowerCase() &&
-      event.tags.some((tag) => tag[0] === "a" && tag[1] === coordinate),
-  );
 }
 
 /**
--- a/desktop/src/features/projects/lib/projectDeletions.test.mjs
+++ b/desktop/src/features/projects/lib/projectDeletions.test.mjs
@@ -1,0 +1,102 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import { isDeletedByA, projectCoordinate } from "./projectDeletions.ts";
+
+const OWNER = "a".repeat(64);
+const OTHER = "b".repeat(64);
+
+const ANNOUNCED_AT = 1_700_000_000;
+
+function project(overrides = {}) {
+  return {
+    owner: OWNER,
+    dtag: "bitchat",
+    createdAt: ANNOUNCED_AT,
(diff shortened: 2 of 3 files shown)


Chessing234:
good catch on re-announced repos. can you add a test where created_at equals the deletion time so the boundary is pinned?

PR #3633, opened 2026-07-29
feat(auth): add provider-neutral authorization contract

cea-block:
## Why

Relay authorization needs a provider-neutral interface that can request portable capabilities without exposing tokens, ingress headers, vendor roles, or deployment policy to the rest of the runtime.

## What

- Add an asynchronous authorization-provider contract for direct and delegated requests
- Validate provider responses against domain, transport, principal, binding, capability, policy, identity, and freshness bounds
- Return request-scoped capability snapshots that callers cannot construct directly
- Separate invitation minting from invitation claiming and preserve exact delegated-owner binding state
- Provide stable deny and unavailable outcomes with redacted diagnostics
- Add synthetic conformance, timeout, cancellation, stale-decision, and data-redaction coverage

## Review guide

- Review the provider request, response, and validated capability-snapshot types as one contract
- Review synthetic tests for direct, delegated, stale, malformed, unavailable, and overbroad provider decisions
- Provider adapters, relay-handler adoption, persistence, migrations, invalidation, and operator workflows are outside this PR

## Risk assessment

Low. This PR adds a contract and tests without installing a provider or changing runtime authorization behavior.

## Testing

- Provider contract and capability validation tests
- Direct and delegated negative cases, timeout behavior, stale-decision rejection, and redaction checks
- Formatting, linting, and hosted repository checks

## References

- Depends on #3564
- Authorization specification: #1485 and #3726

PR #3784, opened 2026-07-30
fix(auth): bucket IPv6 connection limits by /64

daxdax89:
## Problem

`ip_rate_limit_key` keyed the per-IP connection fence on the full address:

```rust
format!("buzz:ratelimit:ip:{}:conn", ip)
```

For IPv6 that is a `/128`. RFC 4291 §2.5.4 fixes the interface identifier at 64 bits, and end sites are routinely delegated a `/64` or shorter, so one host can rotate through 2^64 source addresses and charge a different Redis counter on every connection. The fence counts addresses, not clients.

This is the pre-auth edge control — `check_ip_connection` runs before host to community resolution completes (and instead of it when resolution fails), so unlike the pubkey-keyed limits there is no authenticated identity downstream that re-checks it. It is the only thing standing between an unauthenticated peer and connection-slot exhaustion.

Trace: `ip_rate_limit_key` (`crates/buzz-auth/src/rate_limit.rs:213`) → `check_ip_connection` (`crates/buzz-pubsub/src/rate_limiter.rs:118`) → `run_rate_limit` fixed-window. No prefix normalization anywhere in the chain.

## Fix

Bucket IPv6 on the `/64` prefix via a new `ip_rate_limit_bucket`.

IPv4 deliberately keeps the full address. v4 space is scarce and commonly shared behind carrier NAT, so widening that bucket would let one abusive client deny service to everyone behind the same egress — the opposite failure.

## The IPv4-mapped trap

`::ffff:a.b.c.d` arrives on dual-stack sockets. Masking those to `/64` like any other v6 address would place **every** IPv4 client into the single `::/64` bucket, silently converting a per-client fence into a global cap on all IPv4 traffic — a self-inflicted outage strictly worse than the bug being fixed. They are unmapped to their v4 form first, which also means a client cannot double its quota by switching socket family. There is a test pinning both directions.

## Tests

Five cases in `rate_limit.rs`, all infra-free:

- two addresses in one `/64` share a counter (the regression)
- distinct `/64`s stay independent (guards against over-blocking)
- v4-mapped addresses do not collapse together, and match their native v4 key
- the existing IPv4 key format is unchanged
- the key stays lowercase, matching the invariant `rate_limit_key` already holds

Verified the suite catches the bug: reverting `ip_rate_limit_bucket` to `ip.to_string()` fails exactly the `/64` and v4-mapped tests.

## Operational note

IPv6 keys change shape, so counters in flight at deploy are orphaned. They expire with the rate-limit window (seconds) and no migration is needed.

## Checks

`just fmt-check`, `just clippy` (clean), and `just test-unit` (all six suites) pass locally. `buzz-pubsub` is unaffected — the public signature is unchanged.

## Scope

Found while auditing `buzz-auth`. Two other candidate findings from that pass are deliberately **not** included, since both turned out to be working as designed:

- the inert scope grant in `verify_auth_event` — `SECURITY.md` states channel membership is the only access-control mechanism and there are no capability taxonomies, so the scope layer is intentionally not a gate;
- NIP-98 body binding — `api/bridge.rs` already enforces a `require_payload` check before verification, and `api/git/transport.rs` documents its `body: None` as a considered trade-off.

Happy to open an issue on either if maintainers see it differently.

## Related

Searched open PRs and issues; nothing touches IP rate limiting. #3633 (`feat(auth): add provider-neutral authorization contract`) also lives in `buzz-auth` but only in `lib.rs` and a new `provider/` module, so there is no overlap with `rate_limit.rs`.

### DIFF
--- a/crates/buzz-auth/src/rate_limit.rs
+++ b/crates/buzz-auth/src/rate_limit.rs
@@ -6,7 +6,7 @@
 //! ⚠️ Fixed windows allow up to 2× burst at boundaries. Upgrade to sliding
 //! window or token bucket for strict limiting.
 
-use std::net::IpAddr;
+use std::net::{IpAddr, Ipv6Addr};
 
 use buzz_core::TenantContext;
 use nostr::PublicKey;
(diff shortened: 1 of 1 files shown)


Chessing234:
looks good. if theres a user-visible string change, a tiny screenshot helps a lot

PR #3789, opened 2026-07-30
feat(desktop): make the sidebar agent-working timer glow green

fitz2882:
## What

The elapsed-time badge on a sidebar channel row — the one that counts up while an agent holds the turn — now reads green with a soft halo that breathes, instead of a mauve chip that pulsed its own opacity.

## Why

Two things were off:

- The badge used `text-primary` / `bg-primary/10`, the same accent as everything else accented in the sidebar. "An agent is working in this channel" did not read at a glance.
- `animate-pulse` fades the entire chip, so the running clock dimmed out from under whoever was reading it every two seconds.

Moving the live signal into the halo lets the label hold a steady opacity for the whole cycle.

## How

Three theme-derived tokens back it, following the existing `--ui-warning` pattern in `adaptive-theme.ts`, so the badge tracks the active syntax theme's own green rather than hardcoding one:

- `--agent-working` — label colour
- `--agent-working-bg` — chip surface
- `--agent-working-glow` — halo

The chip surface is a `mix()` rather than a translucent `overlay()` so it stays opaque. This badge sits on the active sidebar row as often as on a plain one, and a translucent green over the active accent would take the accent's hue with it. That also let the `isActive` colour branch — and the prop — go away.

The mix is deliberately light. The label is 11px (`text-2xs`), and every step of green in the chip costs text contrast; on light themes the green is darkened for the same reason.

## Contrast

Measured label-on-chip, sRGB relative luminance:

| theme | ratio | |
|---|---|---|
| Macchiato (dark) | **4.92:1** | clears WCAG AA for small text |
| Latte (light) | **4.93:1** | clears WCAG AA for small text |

For reference, the treatment this replaces measured 3.88:1 on light themes, so light mode gets better here rather than worse.

Base values are also set in `theme.css` for both `:root` and `.dark`, so the badge is never colourless on the first paint after an update — before `ThemeProvider`'s cached var map has been refreshed. They match what `createThemeVars` computes for the default Latte/Macchiato themes.

`prefers-reduced-motion: reduce` stops the animation and keeps the resting glow.

## Testing

```
just desktop-check      # exit 0
just desktop-typecheck  # exit 0
just desktop-test       # 3834 pass, 0 fail
just desktop-build      # exit 0
```

Visual verification was done against the built stylesheet in headless Chromium — the badge markup rendered on both sidebar surfaces and on the active row, in both themes — since the E2E mock bridge has no way to seed an active agent turn. The contrast ratios above were computed from the values `createThemeVars` actually returns for each theme, not from the static fallbacks.

### DIFF
--- a/desktop/src/features/sidebar/ui/SidebarSection.tsx
+++ b/desktop/src/features/sidebar/ui/SidebarSection.tsx
@@ -122,11 +122,9 @@ export function formatWorkingTooltip(
 
 function ChannelWorkingBadge({
   channelName,
-  isActive,
   summary,
 }: {
   channelName: string;
-  isActive: boolean;
   summary: ActiveChannelTurnSummary;
 }) {
   const now = useNow(1000);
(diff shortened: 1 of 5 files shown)


fitz2882:
Closing — this was a local change for my own app, not intended for upstream. Opened in error.

PR #3803, opened 2026-07-30
feat(desktop): show external relay agents in the Agents view

suhaskashyaps:
## Problem

Agents that run behind an external `buzz-acp` harness (e.g. on a VPS or server) announce themselves on the relay via kind:10100 agent-profile events. They participate fully in DMs, channels, and mention autocomplete, but the desktop **Agents** view renders only desktop-managed agents, so externally hosted agents are invisible in the one place a user goes to see 'my agents'. There is no way to publish anything to the relay that makes them appear there today.

## Change

Adds a read-only **Relay agents** section to the Agents view, fed by the existing `useRelayAgentsQuery` (kind:10100 directory) and deduplicated against the managed-agents list. Cards show the relay profile name/avatar and presence status, and open the agent's profile panel.

- New: `desktop/src/features/agents/ui/RelayAgentsSection.tsx`
- Wired into `AgentsView` below the unified managed-agents section
- No Rust changes; `tsc --noEmit` and `biome check` pass

## Non-goals: no management of external agents

This section is **display-only by design**. The desktop cannot manage a relay agent: no start/stop, no restart, no model/persona/env editing, no deletion. Those operations only exist for desktop-managed agents because the desktop owns their process and their record in the local managed-agents store. An external harness agent's lifecycle and configuration live wherever the harness runs (systemd unit, container, another machine), which the desktop can neither see nor control; the relay only carries the agent's self-published kind:10100 profile. The cards therefore intentionally omit every management affordance and only open the profile panel.

If remote management is ever wanted, it would need a separate control-plane design (e.g. owner-signed control events the harness subscribes to) and is out of scope here.

## Testing

Verified against a live relay with two external `buzz-acp` harness agents running different ACP runtimes: both appear with avatar and Online badge, collapse/expand works, and managed agents are not duplicated.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -54,6 +54,13 @@ RELAY_URL=ws://localhost:3000
 # the web frontend at / for browser requests. Leave unset for local dev
 # (use `just web` for Vite HMR instead).
 # BUZZ_WEB_DIR=./web/dist
+# Allowed CORS origins (comma-separated). Unset = permissive (dev mode).
+# When set, MUST include the desktop app's webview origins or its HTTP API
+# calls (invites, moderation) fail with a network error while WebSocket
+# traffic still works: `tauri://localhost` (macOS/Linux) and
+# `http://tauri.localhost` (Windows). Invalid values are rejected, not
+# treated as permissive.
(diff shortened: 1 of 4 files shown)


Chessing234:
nice. keep the managed vs external badge obvious in the grid so people dont try to edit a remote agents local config

PR #3225, opened 2026-07-27
Add Devin as a preset ACP harness

fenner888:
## Summary

- Add Devin to the built-in preset harness catalog using the official native ACP invocation: `devin acp`.
- Link setup guidance to Cognition's official Devin CLI documentation.
- Render a bundled, attributed Devin mark on a white canvas through Buzz's existing runtime-icon system.
- Keep preset capability metadata in the Rust catalog; no duplicate TypeScript runtime table or React runtime checks.
- Move the existing preset catalog and its focused tests into a Rust submodule without changing existing preset behavior, keeping the touched files within the repository's file-size limit.

### Related issue

Follow-up to the generic BYOH harness work in #2773.

### Scope

This is the small preset/data-entry follow-up described in the #2773 discussion. It uses the generic preset readiness contract and does not add Devin-specific authentication probing, permission bypasses, model switching, cloud handoff, or cloud Devin capability claims.

The preset supplies:

- ID: `devin`
- Executable: `devin`
- Arguments: `acp`
- Installation guidance: https://docs.devin.ai/cli

### Testing

Local verification was rerun at the final PR head, `7bb9aa6e862a47a5062b5b8234fdb5ce2aae6c1d`.

- Focused Rust preset tests: 7 passed
- Desktop JavaScript tests: 3,768 passed
- Desktop lint, formatting, file-size, and text guards: passed
- Full Tauri test suite: 1,851 passed, 14 ignored
- Root Rust unit-test groups: passed
- Web production build: passed
- Mobile format, analyze, and test suites: passed
- Full repository `just ci`: passed

The branch also merges cleanly with the current Block `main`. The upstream fork-triggered CI workflow is awaiting maintainer approval; DCO, Semgrep OSS, and zizmor are passing.

The bundled SVG was rendered and visually inspected in both its source dimensions and a 512px preview. The cross-language preset-logo guard verifies that the Devin mapping exists and the asset is present on disk.

PR #3804, opened 2026-07-30
fix(acp): launch ACP/stdio mode for managed agents with no runtime preset

### LINKED ISSUE #3399: Add a configurable ACP runtime registry (or document how to attach a 3rd-party runtime like omp acp)
## Summary

Buzz desktop's runtime registry is hardcoded to 5 entries (`node`, `node-tools`, `codex-acp`, `claude-agent-acp`, `claude-code-acp`). There is no way to attach a third-party ACP-compatible agent such as [oh-my-pi/omp](https://github.com/oh-my-pi/pi) (`omp acp`) through the UI today, even though the underlying plumbing (`acp_command="buzz-acp"`, `agent_command="<runtime>"`, `agent_args=[...]`) is generic and would accept any bin name on `PATH`.

## Evidence

String-scan of `D:\Apps\Buzz\buzz-desktop.exe` at offset **55350058**:

```
…
BUZZ_AGENT_PROVIDER…
Buzzruntimesnodenode-toolscodex-acpclaude-agent-acpclaude-code-acp
npmwin-x64binv24.11.0node.exe…
```

That comma-joined tail after `Buzzruntimes` is the runtime registry. The same 5 names appear nowhere else in any of the 5 binaries in `D:\Apps\Buzz\`. A direct `Buffer`-level grep across `buzz-desktop.exe`, `buzz-acp.exe`, `buzz-agent.exe`, `buzz.exe`, `buzz-dev-mcp.exe` for `omp-acp`, `oh-my-pi`, `minimax`, `agentclientprotocol/omp`, `omx-acp` returns **zero hits**.

I built and globally installed a working bridge as a sanity check:

- npm package: `@agentclientprotocol/omp-acp` (local, Apache-2.0)
- bin name: `omp-acp` (sits next to `codex-acp`, `claude-code-acp`)
- implementation: `child_process.spawn("omp.cmd", ["acp", ...argv], { stdio: "inherit" })` — a thin stdio proxy, ~50 lines, no protocol translation because `omp acp` already speaks Agent Client Protocol over stdio (verified via direct `initialize` handshake: `protocolVersion=1, agentInfo={name:"oh-my-pi", version:"17.1.7"}` with full `loadSession`/`mcpCapabilities`/`sessionCapabilities`).

After installing it, I appended a fresh persona to `managed-agents.json`:

```json
{
  "name": "OMP-Probe",
  "display_name": "OMP-Probe",
  "slug": "omp:probe",
  "acp_command": "buzz-acp",
  "agent_command": "omp-acp",
  "agent_args": [],
  "is_active": true,
  ...
}
```

Restarting Buzz desktop preserved `agent_command: "omp-acp"` verbatim — so the file-level path works. Only the UI runtime picker is closed.

## What I am asking for

One of the following, in order of preference:

1. **Make the runtime registry configurable.** Move the 5-name list out of the binary into something user-editable — config file, `BUZZ_RUNTIMES` env var, or a runtime-discoverable convention (any bin on `PATH` matching `<name>-acp` whose path resolves). I would expect this to be a small change touching `crates/buzz-desktop/` near where `Buzzruntimes` lives.
2. **Document the supported runtimes + add `omp-acp`** to the same list. Less flexible for the long tail but unblocks the immediate use case (`goose`'s env contract is the source of the broken-API-key reports, see #3395 — giving users an alternative runtime that doesn't need the bundled provider-key path would also help there).
3. At minimum: a per-persona `custom_runtime` field whose `agent_command` is honored without going through the registry (effectively what I have working today via direct `managed-agents.json` editing, but it currently has no UI path and is not documented).

## Reproduction (Windows 11, `D:\Apps\Buzz\`, omp 17.1.7 installed)

```bash
# 1. install the bridge anywhere on disk
mkdir omp-acp-adapter && cd omp-acp-adapter
#   omp-acp.js : transparent stdio proxy to `omp acp`
npm install -g .

# 2. confirm bin on PATH and ACP handshake
where omp-acp
echo '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":2,"clientInfo":{"name":"probe","version":"0.0.1"}}}' \
  | omp-acp
# -> {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":1,"agentInfo":{"name":"oh-my-pi","version":"17.1.7"},...}}

# 3. open Buzz desktop, click "connect", look at the runtime list
# -> node, node-tools, codex-acp, claude-agent-acp, claude-code-acp
#    no omp-acp, no way to add one
```

## Environment

- Buzz `D:\Apps\Buzz\`, all 5 binaries from `Jul 25 17:08` / `17:23`
- `omp 17.1.7` already on PATH at `C:\Users\oabdu\AppData\Roaming\npm`
- node 24.11.0, npm 11.6.2

## Related

- #3395 — bundled goose runtime hits 401 because Buzz does not translate `OPENAI_COMPAT_API_KEY` to `OPENAI_API_KEY`. Unblocking an alternative runtime (this issue) does not fix that, but it gives users an escape hatch while it is being triaged.

### LINKED ISSUE #3457: Grok Build runtime spawns bare `grok` (TUI) instead of `grok agent --always-approve stdio` — agents fail with ENXIO
## Summary

Selecting **Grok Build** as the managed-agent runtime starts Fizz/Honey/Bumble with:

```text
agent_cmd=/Users/.../grok
```

and **empty `agent_args`**. That launches the interactive TUI, not the ACP server. Under Buzz (no controlling TTY) every pool worker exits immediately:

```text
Error: Device not configured (os error 6)
ERROR buzz_acp: agent initialize failed: Agent process exited unexpectedly
Error: all 10 agents failed to start — cannot continue
```

Claude / other ACP-native adapters work on the same machine. The Grok CLI is installed and authenticated; this is a harness arg wiring bug, not auth.

## Environment

- **Buzz**: 0.5.0 (macOS, desktop)
- **Grok CLI**: 0.2.114 (`~/.local/bin/grok` → `~/.grok/bin/grok`)
- **OS**: macOS arm64
- **Config**: `preferred_runtime: "grok"` in `global-agent-config.json`
- Managed agents end up with:

```json
"agent_command": "claude-agent-acp",
"agent_command_override": "grok",
"agent_args": []
```

## Steps to reproduce

1. Install Grok CLI (`grok` on PATH / `~/.grok/bin`) and `grok login` (or set `XAI_API_KEY`).
2. Install latest Buzz desktop; create/start Welcome Team agents (Fizz etc.).
3. Set agent runtime / preferred runtime to **Grok Build**.
4. Start agents and open the harness log.

## Expected

Spawn should match the preset harness definition on `main`:

```rust
// desktop/src-tauri/src/managed_agents/discovery.rs — PRESET_HARNESSES
PresetHarness {
    id: "grok",
    label: "Grok Build",
    command: "grok",
    args: &["agent", "--always-approve", "stdio"],
    ...
}
```

i.e. effectively:

```bash
grok agent --always-approve stdio
```

ACP `initialize` should succeed (same as a manual pipe of JSON-RPC over stdio).

## Actual

- Log line: `agent_cmd=.../grok` with no `agent stdio` args.
- Child prints `Error: Device not configured (os error 6)` (ENXIO) and exits.
- All parallel workers fail; harness exits status 1: `all 10 agents failed to start`.

Manual repro of the bad spawn:

```bash
# bare TUI, no TTY — same ENXIO Buzz sees
/Users/$USER/.local/bin/grok </dev/null
# → Error: Device not configured (os error 6)

# correct ACP entrypoint — works
printf '%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{},"clientInfo":{"name":"test","version":"0"}}}' \
  | grok agent stdio
# → initialize result with protocolVersion 1, authMethods, models, etc.
```

## Root cause (from current `main`)

Two related gaps:

### 1. Preset args are not applied when only the command override is set

UI / `preferred_runtime = "grok"` pins `agent_command_override: "grok"` but leaves `agent_args: []` on the managed agent record. Spawn uses the override binary + the stored (empty) args, so the preset's `["agent", "--always-approve", "stdio"]` never reach the process.

### 2. `default_agent_args("grok")` does not fill in the preset defaults

In both:

- `desktop/src-tauri/src/managed_agents/discovery.rs` → `default_agent_args`
- `crates/buzz-acp/src/config.rs` → `default_agent_args`

only `goose` / Claude / Codex / `buzz-agent` are special-cased. **`grok` falls through to `None`**, so `normalize_agent_args("grok", [])` returns `[]` instead of the preset ACP args.

Same class of risk for other Tier-2 presets that need non-empty args when the agent record has empty `agent_args` (`omp` → `acp`, `opencode` → `acp`, `kimi` → `acp`, etc.) if spawn only sees the command + empty stored args.

## Suggested fix

1. **Include Grok (and other preset harnesses) in `default_agent_args`**, mirroring `PRESET_HARNESSES`, e.g.:

   ```rust
   "grok" => Some(vec![
       "agent".into(),
       "--always-approve".into(),
       "stdio".into(),
   ]),
   "omp" | "opencode" | "kimi" | "cursor" | "cursor-agent" => Some(vec!["acp".into()]),
   // ...
   ```

   Prefer a single source of truth: derive defaults from `PRESET_HARNESSES` rather than a second match table.

2. **When applying `preferred_runtime` / command override from a preset**, also persist (or resolve at spawn) that preset's `args`, not only the command.

3. **Regression test**: spawn path / `normalize_agent_args("grok", vec![])` must yield `["agent", "--always-approve", "stdio"]`; integration smoke that `BUZZ_ACP_AGENT_COMMAND=grok` with empty args can complete ACP `initialize` without a TTY.

## Workaround (works today)

Edit managed agents so args are set explicitly, then restart agents:

```json
"agent_command_override": "/Users/<you>/.local/bin/grok",
"agent_args": ["agent", "stdio"]
```

or point the override at a tiny wrapper:

```bash
#!/bin/bash
exec "$HOME/.local/bin/grok" agent stdio "$@"
```

(`--always-approve` is recommended for managed agents so Grok-side tool prompts do not deadlock headless ACP; Buzz already documents this in the Grok preset PRs.)

## Related

- Closed feature: #2347 (register Grok Build as first-class ACP runtime)
- Closed PRs that defined the intended args: #2812, #2918, #2546 (`agent --always-approve stdio`)
- Related enhancement (not this bug): #3279 “Add Grok CLI option”
- Note: some earlier Grok PR discussion mentioned `/dev/tty` / spawn-context issues on the CLI side; current Grok `agent stdio` completes `initialize` fine over pipes — the Buzz-side empty-args spawn is sufficient to explain this failure.

## Acceptance criteria

- [ ] Selecting **Grok Build** as preferred/agent runtime starts agents without ENXIO
- [ ] Harness log shows ACP init for Grok (`agentInfo` / capabilities), not TUI ENXIO
- [ ] Empty `agent_args` + command `grok` normalizes to `agent --always-approve stdio` (or equivalent from preset)
- [ ] Unit coverage for `normalize_agent_args("grok", [])`

### LINKED ISSUE #3729: buzz-acp: OpenCode-backed agents hang/timeout — session/set_config_option sends Claude-only permission mode values OpenCode rejects
## Summary

`buzz-acp` agents backed by `agent_command_override: "opencode"` reliably hang and time out during initialization ("Request timeout — agent did not respond within 60s") on every model, on Linux. Root cause: `--permission-mode` sends a `session/set_config_option` call with a value from buzz-acp's hardcoded Claude-flavored enum (`default`/`accept-edits`/`bypass-permissions`/`dont-ask`/`plan`), but OpenCode's ACP `mode` config option only accepts `build`/`plan`. OpenCode correctly returns a JSON-RPC error for the unrecognized value, but buzz-acp appears to hang waiting rather than handling the error (it only gracefully falls back for `"Method not found"`, e.g. from `claude-code-acp`, which doesn't implement `session/set_config_option` at all).

## Reproduction (direct ACP round-trip against `opencode acp`, no buzz involved)

```
opencode --version   # 1.18.9
```

```jsonc
// 1. initialize -> ok
{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":1,"clientCapabilities":{}}}

// 2. session/new -> ok, returns configOptions with a "mode" option whose only
//    valid values are "build" and "plan"
{"jsonrpc":"2.0","id":2,"method":"session/new","params":{"cwd":"/home/me","mcpServers":[]}}

// 3. session/set_config_option with buzz-acp's default value -> hard error
{"jsonrpc":"2.0","id":3,"method":"session/set_config_option","params":{"sessionId":"<sid>","configId":"mode","value":"bypassPermissions"}}
```

Response to (3):
```json
{"jsonrpc":"2.0","id":3,"error":{"code":-32602,"message":"Invalid params: mode not found: bypassPermissions","data":{"mode":"bypassPermissions"}}}
```

The same happens for every other `PermissionMode` variant buzz-acp can send (`default`, `accept-edits`, `dont-ask`) since none map to OpenCode's `build`/`plan` vocabulary — confirmed by inspecting `session/new`'s returned `configOptions` for the `mode` category. There is currently **no `BUZZ_ACP_PERMISSION_MODE` value that works** with an OpenCode-backed agent.

## Observed in buzz-desktop logs

```
buzz-acp starting: ... agent_cmd=/home/user/.bun/bin/opencode ... permission_mode=bypassPermissions ...
ERROR buzz_acp: agent timed out during init (60s) agent=0
ERROR buzz_acp: agent initialize failed: Request timeout — agent did not respond within 60s agent=1
ERROR buzz_acp: agent initialize failed: Request timeout — agent did not respond within 60s agent=2
```

No corresponding opencode log file is ever created for the failed instances (`~/.local/share/opencode/log/`), and the buzz session logger's per-session `.md` file stays stuck at "Status: Active" with zero further content — consistent with the agent process being stuck waiting on the `session/set_config_option` response rather than crashing.

This reproduces on every model (`spark/glm-5.2`, `spark/qwen36-35b-a3b`) and is unrelated to auth/network — a bare `session/new` + `session/prompt` round trip against the same `opencode acp` process (skipping the `session/set_config_option` call) completes in a few seconds.

## Environment

- Buzz Desktop 0.5.2, Linux x86_64
- OpenCode 1.18.9
- `agent_command_override: "opencode"` on a managed agent (per `~/.local/share/xyz.block.buzz.app/agents/managed-agents.json`)

## Suggested fix

Either:
1. Treat any non-"Method not found" error from `session/set_config_option` as non-fatal (log and continue without the mode override), the same way "Method not found" is already handled, or
2. Query `session/new`'s returned `configOptions` for the `mode` category's valid `options` and only attempt `session/set_config_option` if the configured `PermissionMode`'s wire value is actually in that list, or
3. Add an adapter-specific permission-mode mapping (similar in spirit to #2884's request for a Claude `auto` variant) so OpenCode's `build`/`plan` modes can be targeted directly instead of assuming Claude's vocabulary.

Related but distinct: #3660 reports the same overall symptom ("Opencode acp not working" / 60s init timeout) but attributes it on Windows to a missing `acp` subcommand argument. On Linux, `buzz-acp`'s default `--agent-args acp` is applied correctly (confirmed the spawned command is `opencode acp`), so that isn't the cause here — this is a distinct bug in the `session/set_config_option` error-handling path.

bradhallett:
## Summary

Managed agents pinned to a known ACP runtime via `agent_command_override` (`omp`, `grok`, `opencode`, `kimi`, …) but with no `runtime` id launched the agent's **interactive TUI** instead of its ACP/stdio mode. Under headless Buzz there is no controlling TTY, so the JSON-RPC `initialize` handshake never completed and the agent timed out at startup:
- omp / opencode / kimi → `Request timeout`
- grok → `ENXIO: Device not configured`

## Root cause

`resolve_effective_harness_descriptor` (`readiness.rs`) only resolved launch args when the record carried **explicit instance args** or a **resolvable harness preset** (`record.runtime` → `lookup_loaded_harness_by_id`). A command-override agent with `runtime = None` satisfied neither, so `descriptor.args` stayed **empty** — while the effective command (e.g. `omp`) was still correct. The bare command then ran the TUI.

The correct per-runtime args already existed as the single source of truth in `PRESET_HARNESSES` (now in `discovery/presets.rs`); they just weren't consulted on this path.

## Fix

In `resolve_effective_harness_descriptor`'s final `else` branch, fall back to the `PRESET_HARNESSES` entry whose command matches the effective command, via a new `preset_args_for_command()` helper that mirrors the existing single-source `preset_harness_ids()` helper and lives alongside it in the `managed_agents::discovery::presets` submodule (re-exported through `discovery.rs`).

- **Single source of truth** — per-runtime args live only in `PRESET_HARNESSES`; no hand-mirrored copy.
- **Covers every preset at once** — omp, grok, opencode, kimi, cursor, openclaw, and devin.
- `descriptor.args` now flows correctly to the spawn env (`BUZZ_ACP_AGENT_ARGS`), the config hash, and the agent summary.

```diff
  } else if let Some(ref def) = harness_def {
      normalize_agent_args(&effective_command, def.args.clone())
+ } else if let Some(preset_args) =
+     crate::managed_agents::discovery::preset_args_for_command(&effective_command)
+ {
+     normalize_agent_args(&effective_command, preset_args)
  } else {
      normalize_agent_args(&effective_command, record_args)
  }
```

## Why not in `buzz-acp`

An earlier iteration mirrored the args into `buzz-acp`'s `default_agent_args`. That duplicated the preset table in a third place and still left the desktop's `descriptor.args` (env/hash/summary) empty — relying on the child to repair the parent's omission. Fixing at `resolve_effective_harness_descriptor` corrects the source and needs no `buzz-acp` change; `buzz-acp`'s existing `default_value = "acp"` still covers standalone CLI use with an unset env.

## Verification

- New unit test `resolve_effective_harness_descriptor_uses_preset_args_for_pinned_command` — asserts omp/grok/opencode/kimi/cursor/openclaw/devin resolve correctly, explicit args win, unknown commands stay empty. ✅
- `cargo fmt --check` clean (pinned `1.95.0` toolchain); full `managed_agents::` lib suite: 1001 passed, 0 failed (rebased onto v0.5.9 `main`).
- Rebased onto latest `block/buzz` `main` (resolves the conflict with #3225, which relocated `PRESET_HARNESSES` and its helpers into `discovery/presets.rs`; this helper follows it there).
- Confirmed end-to-end in a local app build that an `omp` managed agent initializes without the ACP timeout.

## No regressions

- Explicit instance args still take precedence.
- `goose` (not in `PRESET_HARNESSES`) still resolves via `default_agent_args`.
- `amp-acp` / `hermes-acp` presets carry empty args — unchanged.
- Custom harness ids collide with reserved preset ids, so they cannot shadow a preset.

Resolves #3457
Refs #3399

---

**Issue scope** (corrected — an earlier draft also listed `Resolves #3399` / `Resolves #3729`):

- **#3457** — command-override agents (e.g. Grok) launching the bare TUI with
  empty args. **Directly fixed** here.
- **#3399** — attach a 3rd-party runtime like `omp`. This fixes the launch-args
  symptom (an `omp` override now resolves `["acp"]`); the broader
  configurable-registry / UI-picker ask is a separate feature, so it is `Refs`
  rather than `Resolves`.
- **#3729** — OpenCode hanging on `session/set_config_option`. **Already fixed
  on `main` and shipped in v0.5.9** via the `agent_supports_mode()` guard in
  `buzz-acp` (PR #188), which skips the call for agents that don't advertise the
  mode. This PR does not touch `buzz-acp`; the earlier `Resolves #3729` was
  inaccurate and is removed.


Chessing234:
good for #3729. maybe add a log line of the resolved launch args at spawn so failed initialize is easier to debug

bradhallett:
@Chessing234 Implemented in b9825afa3 and pushed to the PR branch. `spawn_agent_child` now writes the resolved agent command and argument vector to the managed-agent runtime log immediately before spawn, so failed ACP initialization includes the launch values. Verified with the full desktop library suite (1,996 passed, 14 ignored) and Clippy with `-D warnings`.

bradhallett:
Maintainer approval requested:

PR #3905 overlaps with #3804 on command-only preset argument resolution. I recommend keeping #3804 as the canonical fix:

- resolves built-in preset args from normalized `PRESET_HARNESSES` data;
- covers all affected presets and explicit-args precedence;
- includes the requested resolved-command/args runtime-log diagnostic;
- passes the full desktop library suite, formatting, and Clippy with `-D warnings`.

#3905 additionally explores loaded-registry matching and runtime-ID/command mismatch diagnostics. I am happy to port the mismatch diagnostic if command-only custom-harness support is required, but I recommend not merging both implementations.

Can we proceed with #3804 as the canonical PR and close #3905 in favor of it?

skeletor-js:
We are running a headless Grok ACP worker under systemd and hit the same failure described in #3457: launching the bare Grok command enters its interactive path without a controlling TTY, while supplying the runtime ACP/stdio arguments works reliably. The preset-argument fallback in this PR addresses the problem at the right layer and the visible checks are green. +1 for merging this as the canonical fix and including it in the next Buzz release.

bradhallett:
Rebased onto current `main` (v0.5.4, `651f637`) — the `discovery.rs` conflict is resolved and GitHub now reports this PR as **mergeable** (previously `DIRTY`). The conflict was purely additive: kept `main`'s new `#[macro_use] mod windows_install;` alongside the preset re-export.

Recap for review:
- The resolved-launch-args log line @Chessing234 asked for is included.
- Resolves #3399, #3457, #3729 — the ACP/stdio `initialize` timeout when a command-override agent has no runtime id.
- Independent +1 from @skeletor-js, running a headless Grok ACP worker under systemd that hits exactly this (#3457).
- Verified locally on the pinned 1.95.0 toolchain: `cargo check` clean, `cargo fmt --check` clean, and the regression test `resolve_effective_harness_descriptor_uses_preset_args_for_pinned_command` passes.

This is the canonical fix; #3905 overlaps and I'd close it in favor of this one. Could we get a review approval so it can land?

scottopell:
Looks like this got hit by the gh actions outage, CI needs a kick here (pushing empty commit was easiest for me, no CI was published to retry) fyi @bradhallett 

Oh and another +1 from me, I'm building out an ACP adapter layer for https://github.com/scottopell/phoenix-ide and ran into this

bradhallett:
Rebased onto current `main` (past `v0.5.17`, base `417eea223`) — mergeable, `discovery.rs` conflict resolved, `cargo fmt` clean, and the full `managed_agents` suite passes locally with `--features mesh-llm` (1029/1030; the one failure, `inherited_shared_compute_translates_to_supported_agent_transport`, reproduces on stock `main` and is unrelated to this change).

@scottopell re: the CI kick — pushing didn't publish a run to retry; both workflows show `action_required` on the new head (`26b617c77`), so they need an "Approve and run" from someone with write access.

The rebase also dismissed the approvals above — no code changes since your reviews beyond the conflict resolution and the fixture field main added (`provider_policy_pending`); a re-approval when convenient would unblock merge.

wesbillman (changes requested):
Requesting changes for two merge blockers; the preset-argument fallback itself is correctly placed and its precedence is sound.

1. **Security: do not write raw arbitrary `agent_args` to the runtime log** (`desktop/src-tauri/src/managed_agents/runtime.rs:497-503`). `agent_args` is user-controlled and may legally contain credentials such as `--token=...`. This repository's spawn-snapshot redaction policy explicitly treats `args` as `MaskedBare` for exactly that reason (`spawn_snapshot/diff.rs:82-86`), with a secrecy test using `--token=...`. The new `{:?}` log line bypasses that protection and makes the secret retrievable through `get_managed_agent_log`. Log a redacted representation (or only argument count / known preset mode), with a regression test proving a token-bearing arg is absent.

2. **Required CI/file-size guard is failing.** The PR grows already-oversized `readiness.rs` from 1741 to 1850 lines and `runtime.rs` from 997 to 1004; `Detect Changed Paths` fails and consequently skips Desktop Core, Rust lint, builds, and E2E. Repository guidance says oversized files may not grow and files crossing 1000 lines must be split rather than bypassing the limit. Move the large fixture/test into an existing or new test submodule and place the logging helper/test where it does not push `runtime.rs` over the ceiling.

Source verdict on the functional fix: matching the normalized effective command against `PRESET_HARNESSES`, after explicit instance args and resolved harness-definition args, preserves precedence and restores ACP/stdio args from the existing source of truth. Unknown commands and empty-arg adapters remain unchanged. I would approve after the disclosure and mandatory CI failures are fixed.

Review scope: source/diff + CI only, per this channel's strict review contract; no local runtime execution.

wesbillman (changes requested):

The core preset-argument fallback is correct, but this head has two blockers:

1. `spawn_agent_child` now persists the complete `agent_args` vector with `{:?}` (`desktop/src-tauri/src/managed_agents/runtime.rs:497-503`). These arguments are arbitrary user-controlled CLI input and may legally contain `--token=...`; the repository's snapshot redaction policy explicitly masks all `args` because of that (`desktop/src-tauri/src/managed_agents/spawn_snapshot/diff.rs:72-86`). Runtime logs are also created by generic `open_log_file` without owner-only creation mode (`storage.rs:654-660`), unlike credential-bearing install logs (`storage.rs:693-716`). Please do not serialize raw argv. Logging the resolved command plus argument count/source is sufficient; hardening runtime-log creation to `0600` is worthwhile defense in depth but does not make persisting secrets appropriate.

2. Required CI fails the file-size ratchet: `readiness.rs` grows from 1741 to 1850 lines and `runtime.rs` from 997 to 1004. That causes the substantive Desktop/Rust/build/E2E jobs to be skipped and their aggregate checks to fail. Please move the large regression fixture/test into an existing sibling test module (or otherwise split it) and keep production files within the ratchet.

The functional resolution order itself looks sound: explicit instance args still win, then a resolved harness definition, then normalized command matching against the single-source `PRESET_HARNESSES`; unknown commands remain unchanged. Once the raw-argv logging and file-size failures are resolved and required CI runs, I expect this to be mergeable.

bradhallett:
Addressed both blockers from https://github.com/block/buzz/pull/3804#pullrequestreview-4966165343 in d518e766b (rebased onto current main):

**1. Raw `agent_args` in the runtime log.** `spawn_agent_child` no longer formats `agent_args` into the log at all. The resolved-launch marker moved into `storage.rs` as `append_resolved_launch_marker(path, command, args)`, which writes only the resolved command and the argument count (`resolved launch: agent_command=… args_count=N`) — mirroring the `spawn_snapshot::diff` `MaskedBare` policy — and returns the resolved command for the `BUZZ_ACP_AGENT_COMMAND` spawn env. Regression test `append_resolved_launch_marker_never_writes_argument_values` asserts a `--token=…` arg value can never reach the log (and that unresolvable commands fall through verbatim); it fails if any argument value is serialized. Defense in depth: `open_log_file` now creates runtime logs `0600` at creation (mode-in-create, not post-hoc chmod), covered by `open_log_file_creates_owner_only`.

**2. File-size ratchet.** The two effective-launch tests moved from `readiness.rs` into `readiness_effective_launch_tests.rs` (external `#[path]` module, same pattern as `readiness_goose_file_config_tests.rs`), and `runtime.rs` stays under the cap after folding command resolution into the storage helper. `node scripts/check-file-sizes.mjs` passes against the new merge base, so Detect Changed Paths should no longer skip the required jobs.

Local verification: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace` 2595 passed / 0 failed, and both regression tests were mutation-checked (reverting the redacted format or the `0o600` mode-in-create makes them fail).

bradhallett:
Note: the force-push (rebase onto current main) reset the workflow approval gate — both `CI` and `Desktop Release Candidate` are `action_required` on d518e766b and will need re-approval from a maintainer to run.

bradhallett:
Rebased onto current `main` (`c3132c3ee`) — GitHub reports this PR **mergeable** again (the previous head had fallen behind `main` and conflicted).

No behavioral change since `d518e766b`; this round is conflict resolution plus two ratchet artifacts the rebase surfaced:

- **Relocated-test literals went stale.** `main` (#5112) added `team_catalog_source` to `ManagedAgentRecord` after the last rebase; the two record literals in `readiness_effective_launch_tests.rs` now carry the field.
- **`storage.rs` crossed the file-size ratchet.** `main` grew it to 995 lines, and the marker helper pushed it to 1032 (> 1000). The two log-marker helpers (`append_log_marker`, `append_resolved_launch_marker`) moved to a new `launch_log.rs` (76 lines, regression test included); `storage.rs` is back to 999. Redaction semantics are unchanged: the marker still records only the resolved command and `args_count` — never an argument value — and `open_log_file` still creates runtime logs owner-only (`0o600`).

Both blockers from the Aug-18 reviews remain resolved as described in my previous comment.

Gate on the new head (`74b60d89b`):
- `just file-size-check` (desktop / web / mobile) — clean
- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean
- `cargo test --workspace` — **3109 passed / 0 failed / 19 ignored**

@wesbillman could you take another look when convenient? One logistics note: the force-push reset the workflow approval gate — `CI` and `Desktop Release Candidate` are `action_required` on `74b60d89b` and will need an "Approve and run" from someone with write access.

PR #3807, opened 2026-07-30
feat(desktop): authorize existing external agents

michegz:
## Summary

- add an owner-only **Authorize external agent...** action to channel member menus
- verify the current signer is a channel owner and the target is an existing bot member before minting a NIP-OA credential
- keep the owner private key inside Buzz Desktop and copy only the resulting authorization
- add Rust policy/cryptographic tests plus a Playwright owner-consent flow

Originating Buzz channel: `1685d446-b7a8-4e2e-b602-c6c6f9eab06a`

## Why

External agents can be added to a channel as bots, but closed Buzz relays also
require a valid owner-signed NIP-OA authorization. Before this change, Desktop
had no owner-reviewed way to mint that credential for an existing external bot.

## Security boundaries

- only a channel owner may authorize an external agent
- the target must already be an agent with the `bot` channel role
- an agent with an existing valid owner authorization cannot be re-authorized
- self-attestation is rejected by the NIP-OA implementation
- Desktop does not persist or expose the owner's private key
- the resulting authorization is shown only through an explicit confirmation
  flow and copied to the clipboard for installation in the external agent

## Verification

- `just desktop-check`
- `just desktop-test`
- `just desktop-build`
- `just desktop-tauri-check`
- `just desktop-tauri-test` — 1,923 passed, 0 failed, 14 ignored
- `pnpm build:e2e && pnpm exec playwright test --only-changed=origin/main`
  — 74 passed

`just ci` reaches an existing macOS clippy failure in
`desktop/src-tauri/src/linux_media.rs`: three Linux-only helpers are considered
unused under `-D warnings`. This branch does not change that file; the
feature-specific Rust, TypeScript, build, and changed-spec E2E checks above pass.

### DIFF
--- a/desktop/src-tauri/src/commands/channels.rs
+++ b/desktop/src-tauri/src/commands/channels.rs
@@ -465,6 +465,36 @@ fn parse_channel_uuid(channel_id: &str) -> Result<uuid::Uuid, String> {
     uuid::Uuid::parse_str(channel_id).map_err(|_| format!("invalid channel UUID: {channel_id}"))
 }
 
+fn build_external_agent_auth_tag(
+    owner_keys: &nostr::Keys,
+    members: &ChannelMembersResponse,
+    agent_pubkey: &str,
+) -> Result<String, String> {
+    let owner_pubkey = owner_keys.public_key().to_hex();
+    let owner_is_channel_owner = members
+        .members
+        .iter()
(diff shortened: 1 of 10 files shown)


Chessing234:
useful. please fail closed when the external agent has no clear owner attestation so we dont authorize random directory entries

PR #3816, opened 2026-07-30
fix(desktop): show eligible cross-owner channel agents in @ mentions

### LINKED ISSUE #3809: Desktop @-mention autocomplete only shows locally managed agents
## Purpose

Teams need to `@mention` shared / org-level agents from their own Buzz Desktop — for example a research agent owned by someone else but configured with `respond_to: allowlist` or `anyone` and present in a shared channel.

Today that is impossible through the UI. Non-owners only see agents managed by **their own Desktop install**. `respond_to` settings cannot fix it.

## Blocker

In Desktop mention autocomplete, agent candidates are filtered by `isAgentIdentityInManagedList` **before** the `respond_to` / shared-channel eligibility logic runs.

`managedAgentPubkeys` comes from `listManagedAgents` (local Tauri command for this install). Any channel member with `role: "bot"` whose pubkey is not in that local set is dropped.

So the allowlist / anyone eligibility code in `relayAgentIsSharedWithUser` / `shouldHideAgentFromMentions` is unreachable for cross-owner agents.

## Actual vs expected

| | |
|---|---|
| **Actual** | `@` list shows only agents this Desktop manages. Cross-owner channel bots never appear. |
| **Expected** | Agents that are channel members and that the viewer is allowed to invoke (`respond_to: anyone` + shared channel, or `respond_to: allowlist` including the viewer) appear in `@` autocomplete. |

## Repro

1. User A owns agent X; agent X is a member of a shared channel; `respond_to` is `allowlist` (includes User B) or `anyone`.
2. User B opens that channel on their own Desktop (does not manage agent X).
3. User B types `@` in the composer.

**Result:** Agent X is missing. User B only sees agents their install manages.

**Note:** Explicit CLI `--mention <pubkey>` still delivers (relay accepts it). This is a Desktop autocomplete UI gate, not a relay delivery failure.

## Code

- `desktop/src/features/messages/lib/useMentions.ts` (~244–262) — `addCandidate` order: managed-list filter before eligibility hide
- `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts` (`isAgentIdentityInManagedList`, `relayAgentIsSharedWithUser`)
- `desktop/src/features/agents/hooks.ts` (`useManagedAgentsQuery` → local `listManagedAgents`)

## Out of scope for this issue

Whether the harness author gate wakes the agent after a successful mention. That is a separate path. This issue is only: **non-owners cannot select the agent in Desktop `@` autocomplete.**

CExKEze:
## Summary

Desktop `@`-mention autocomplete dropped **cross-owner channel-member agents** because `useMentions` filtered on `isAgentIdentityInManagedList` (local install only) before `shouldHideAgentFromMentions` / `relayAgentIsSharedWithUser` could run.

This change lets **channel-member** agents through the managed-list gate **once the relay agent directory is ready**, then keeps `shouldHideAgentFromMentions` as the single invocability decision. Non-member relay-directory agents stay on the managed-list gate (preserves `mentions.spec.ts` “relay-only agents stay hidden…” policy / #3809 Expected = channel members).

Fixes #3809

Buzz channel: `7eeeb66d-80cc-4680-87c6-0d38d8d25624`

### Related issue

- https://github.com/block/buzz/issues/3809
- Duplicate search: none closer than #3809 for Desktop cross-owner mention autocomplete.

### What changed

- `isAgentIdentityReachableForMentions` — member pass-through only when `relayAgentDirectoryReady` (reuses existing predicate in `useMentions.ts`; no new readiness predicate).
- `shouldOfferAgentIdentityForMentions` — production chain (reachability → `shouldHideAgentFromMentions`) so tests cannot drift from a hand-rolled mirror; also keeps `useMentions.ts` under the 1000-line file-size ratchet (996 lines; base on `origin/main` was 999).
- Unit cases for eligible unmanaged member, directory-excluded member, loading-directory withhold, unreachable non-member.
- E2E: `alice` (default mock relay agent, member of `general`, `respond_to: anyone`, not managed by the install) now expected visible with agent icon.

### Scope / no-widening

`rg isAgentIdentityInManagedList desktop/src` → production call sites are only `useMentions.ts` and `MembersSidebar.tsx`. This PR changes the mention path only. Lookalikes (`useClassifiedMembers`, `useSearchResults`, `useNewMessageRecipients`, `ProjectsAgentPromptPage`) build their own sets / import other helpers and do **not** take the new function.

Out of scope: harness author gate, mobile/web clients, `respond_to` semantics, non-member relay-directory broadening, `linux_media.rs` macOS clippy noise.

### Testing

Attributed commit: `4e032f7d546b5f9f82fe0b4bd2051ddc782fc270`

| Recipe | Local exit @ SHA |
|--------|------------------|
| fmt-check | 0 |
| clippy | 0 |
| desktop-check | 0 |
| desktop-tauri-fmt-check | 0 |
| web-check | 0 |
| mobile-check | 0 |
| test-unit | 0 |
| desktop-test | 0 (3853 pass) |
| desktop-build | 0 |
| desktop-tauri-check | 0 |
| desktop-tauri-test | 0 |
| web-build | 0 |
| desktop-tauri-clippy | macOS lib-target dead code in `linux_media.rs` (pre-existing; reference at line 86 inside linux-only `enable_media_capture`). **Authoritative evidence: this PR’s GHA `desktop-core` → Desktop Tauri clippy on `ubuntu-latest`.** |
| mobile-test | Local red on unrelated Flutter follow-mode assert (`channel_detail_page_test.dart:1053`); **zero `mobile/**` in diff**; GHA Mobile job path-filtered out for this change. |

Also: eligibility unit suite 24 pass; `mentions.spec.ts` 49 pass including relay-only hidden policy at `:829`.

#### Second-identity smoke (recorded steps + screenshot)

At `4e032f7d5`, e2e bridge install manages only `charlie`; types `@` in `#general`:

- `alice` appears with **agent** label (not “managed by you”, not “not in channel”) — `SMOKE_ALICE_COUNT=1`
- Rows include: alice/agent, mira/agent managed by you, Fizz, charlie managed…

Screenshots: attached in issue/PR thread from `.scratch/3809-smoke/` (`cross-owner-alice-dropdown.png`).

### Dual-frame review notes (Nick-Lex / Shuri-Lex)

- **P3 (non-member relay-directory agents stay hidden):** deliberate. Matches #3809 Expected (channel members) and the existing `mentions.spec.ts` relay-only policy. Mission “and relay-directory” wording is narrowed to members only — recorded here so it is not silent.
- **P2 (directory-absent member bot offered when directory is ready/non-empty):** pre-existing Option B inside `shouldHideAgentFromMentions` (untouched). Made live for unmanaged agents by this PR. **Not tightened here** — would re-break #3809 when the directory is incomplete; recommend a follow-up issue if product wants “no entry ⇒ hidden.”
- Alice sort-order assert and local smoke temp files: non-blocking; smoke temps removed from the worktree (never committed).

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -7,10 +7,12 @@ import {
   getMentionableAgentPubkeys,
   getSharedChannelIds,
   isAgentIdentityInAllowedList,
+  isAgentIdentityReachableForMentions,
   isAgentMentionChannelType,
   relayAgentCanRespondInChannel,
   relayAgentIsSharedWithUser,
   shouldHideAgentFromMentions,
+  shouldOfferAgentIdentityForMentions,
   uniqueAutocompleteLabels,
 } from "./agentAutocompleteEligibility.ts";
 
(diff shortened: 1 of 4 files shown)


CExKEze (commented):
## Front-End / UX review (Shuri-Lex)

**Scope reviewed:** Desktop mention autocomplete path only — `agentAutocompleteEligibility.ts`, `useMentions.ts`, eligibility unit tests, `mentions.spec.ts` e2e flip. Smoke screenshot from the channel thread.

**Overall:** No blockers. The fix restores the intended UX for #3809: eligible cross-owner channel bots appear in `@` with the generic `agent` label (not “managed by you”, not “not in channel”), while invocability still funnels through `shouldHideAgentFromMentions`. Scope stayed on the mention path — `MembersSidebar` still uses `isAgentIdentityInManagedList`.

**Confidence:** High

### What looks right
- Reachability → invocability chain is the right product order; managed-list no longer dead-ends cross-owner members.
- Loading gate avoids a false-positive flash of owner-only/allowlist-excluded bots while the directory is in flight.
- Option B (ready + empty/errored directory → show unknown members) is the safer UX degradation than indefinite hide.
- Smoke evidence matches expected labels: `alice` = `agent`; managed agents keep `agent managed by you`.
- E2E expectation flip for alice correctly encodes the bug class rather than papering over it.

### Non-blocking notes
1. Comment nit in `isAgentIdentityReachableForMentions`: “members stay hidden” while loading is slightly loose — humans still pass via `isAgentIdentityInManagedList`; only unmanaged agent members are withheld. Worth tightening if you touch the comment again.
2. When the directory flips to ready, eligible cross-owner agents can appear mid-open autocomplete (list growth). Acceptable tradeoff vs flashing ineligible agents; no a11y live-region needed for this PR.

No accessibility or visual-system regressions introduced by this change. Leaving adversarial / correctness gate to Nick-Lex.

CExKEze:
Adversarial review pass — no blocking defects found. One behavioural consequence worth recording for maintainers, since it is not visible from the diff alone.

**Directory-absent member bots are now offered.** `shouldHideAgentFromMentions` is unchanged here (63 additions, 0 deletions in `agentAutocompleteEligibility.ts`), but its "Member (Option B)" branch was dead code for unmanaged agents before this change, because the managed-list gate ran first. Activating the member path makes it live, with this result:

| Case | Outcome |
|---|---|
| member bot, `owner-only` entry present in the relay directory | hidden |
| member bot, **no** directory entry, directory ready and non-empty | **offered** |
| non-member bot, `anyone` + shared channel | hidden |

So a bot that publishes a directory entry excluding the viewer is hidden, while a bot that publishes nothing at all is offered. #3809's Expected asks for agents "that the viewer is allowed to invoke (`respond_to: anyone` + shared channel, or `respond_to: allowlist` including the viewer)", and a bot with no entry satisfies neither.

I do not think this should change in this PR. Tightening it would re-break #3809 whenever the relay directory is incomplete, and it grants no new capability — the issue notes that an explicit `--mention` already delivers today. It reads as a reasonable fail-open, better handled as a follow-up if maintainers want directory absence treated as an explicit not-invocable signal.

Verified independently at `4e032f7d546b5f9f82fe0b4bd2051ddc782fc270`: full desktop unit suite 3853 pass / 0 fail, with `git rev-parse HEAD` confirmed in the same shell before and after the run. The only production callers of the managed-list helper remain `useMentions.ts` (via the new wrapper) and `MembersSidebar.tsx` (unchanged), so no other surface widened.

Note for CI: the workflow run for this PR concluded `action_required` with zero jobs (first-time fork contributor gate), so `desktop-core` → Desktop Tauri clippy has not run yet and needs a maintainer "Approve and run".

Chessing234:
this is the right fix for #3809. can you also note in the test that relay-only non-members stay hidden so that regression is pinned?

CExKEze:
Thanks — agreed on pinning relay-only / non-member stay-hidden.

That regression is already covered in this tip (`6faa67b91`):

1. **E2E** (unchanged by this PR): `desktop/tests/e2e/mentions.spec.ts` — `relay-only agents stay hidden from channel mentions even when allowlisted` (allowlisted `quinn` → autocomplete count 0).
2. **Unit** (added here): `desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs` — `mention path keeps unreachable non-member agents out of the composer` (via `shouldOfferAgentIdentityForMentions`).

So the deliberate narrowing is: channel-member path opens for eligible cross-owner bots; non-member relay-directory agents stay out of the composer.

Happy to rename/comment those tests for clearer discoverability in a tiny follow-up after CI has run on this tip — holding commits for now so we don’t re-queue the first-contributor workflow approval.

oceanseth:
Independent confirmation from a self-hosted deployment (Desktop 0.5.x on Windows, stock relay Docker image): this PR's approach is the one that actually fixes cross-owner mentions there. Our relay's kind:10100 directory contains zero events — nothing in-tree ever publishes `respond_to` data to it — so approaches that admit candidates via the mentionable/directory set can never surface a foreign agent on deployments like ours, while this PR's member pass-through (with the directory-readiness fallback treating an empty directory as "unknown invocability ⇒ show") restores exactly the intended behavior. Full evidence with code refs in https://github.com/block/buzz/issues/3809#issuecomment-5186144164.

CExKEze:
## Rebased onto latest `main`

Tip is now `a2c468914` on `fix-3809-desktop-cross-owner-mentions` (rebased onto `f8f2ef044`).

**What changed in the rebase**
- Kept `main`'s eligibility-scope / channel-composer helpers (`isAgentMentionChannelType`, `uniqueAutocompleteLabels`, `filterCachedAgentSuggestions`, `isAgentIdentityInAllowedList`).
- Re-applied the #3809 admission path: channel members pass through once the relay directory is ready, then `shouldHideAgentFromMentions` (Option B for directory-absent members). Composer calls `shouldOfferAgentIdentityForMentions`.
- Dropped the follow-up e2e sort commit — Alice-before-Fizz was already on `main`.

**Verification at tip `a2c468914`**
- Desktop unit suite (lefthook `desktop-test`): **4590 pass / 0 fail**
- Focused eligibility file: **28 pass / 0 fail**
- `pnpm typecheck` + `pnpm check` clean for the Desktop TS surface

## Maintainer ask: Approve and run workflows

This PR is from a first-time fork contributor. Previous workflow runs concluded `action_required` with zero jobs, so Desktop Tauri clippy / `desktop-core` never started.

Please **Approve and run** the workflows on this tip so CI can actually exercise the Desktop change. Happy to address any follow-up once the first full run lands.

Eight8Kane:
This approach covers a case that isn't described in #3809, because the predicate you kept is provenance-based rather than ownership-based.

I reported the detail on the issue ([comment](https://github.com/block/buzz/issues/3809#issuecomment-5436493936)): on 0.5.19 an agent provisioned from the CLI is missing from `@` autocomplete **for its own owner**, on the machine that signed its NIP-OA auth tag. `listManagedAgents` is scoped to the install, not to ownership, so "the viewer does not manage it" is true even when the viewer owns it. Your first test — `keeps channel members the viewer does not manage` — is exactly that predicate, so this PR should fix the headless-deployment case as a side effect of fixing the cross-owner one.

One gap in the matrix, if you want to pin it: the directory-exclusion test uses `respondTo: "owner-only"` with `OTHER_OWNER_PUBKEY`, which asserts the hide. There is no fixture for **viewer == the agent's owner** with `respondTo: "owner-only"` and the agent absent from `managedAgentPubkeys` — which is a real shape, not a hypothetical: an owner-only agent deployed by script is invocable by exactly one person, and that person is the viewer. If `shouldHideAgentFromMentions` decides owner-only purely from the directory record without resolving the viewer against the agent's owner, that agent would be hidden from the only user allowed to call it.

Worth a fixture either way, so a later refactor can't reintroduce an ownership shortcut without the test noticing.

Happy to test a build against a live self-hosted relay with headless-provisioned agents if that's useful — that setup is running here.

PR #3839, opened 2026-07-31
desktop: show Goose updates when available

lifeizhou-ap:
## Summary

- keep every installed Goose runtime Ready and launchable
- check the installed version against the latest stable release only from Settings → Agents
- show Update only when Goose is behind, checking once per active community with manual Check again refresh
- run `update` through the resolved Goose binary and recheck availability after success

## Why

Goose releases separately from Buzz, so users may not know when an update is available. Buzz should let users update Goose from Settings without preventing them from using the version already installed. If the update check fails, Goose should still remain ready to use.

## User experience

- Goose not installed: Install
- Goose current, checking, or check failed: Ready
- Goose behind latest stable: Ready + Update
- Successful update: recheck and remove Update when current

## Testing

- `just desktop-check`
- `just desktop-test` (3,879 passed)
- `just desktop-tauri-clippy`
- `just desktop-tauri-test` (2,018 passed; 14 ignored)
- `pnpm build:e2e`
- `pnpm exec playwright test tests/e2e/goose-update-screenshots.spec.ts --project=smoke` (2 passed)


lifeizhou-ap:
🤖 ### Goose update states

Current Goose stays simply Ready.

goose-current

A confirmed older Goose adds an optional Update action without blocking readiness.

goose-update-available

lifeizhou-ap:
video


https://github.com/user-attachments/assets/f413ad24-db9b-4edc-8045-db88d887055d

Chessing234:
neat. maybe rate-limit the check-again button so we dont hammer github releases from every community switch

desktop/src-tauri/src/commands/agent_discovery.rs
```diff
@@ -316,16 +317,12 @@ fn install_acp_runtime_blocking(
...
-            for cmd in runtime.cli_install_commands_for_os() {
-                let result = run_install_command_with_retry("cli", cmd, &reporter);
-                let success = result.success;
-                steps.push(result);
-                if !success {
-                    return Ok(reporter.failed(steps));
-                }
-            }
+    for (step, command) in cli_setup::plan(runtime) {
+        let result = run_install_command_with_retry(step, &command, &reporter);
```
wesbillman, line 321: **Blocking:** this sends `goose update` through installer retry semantics. `run_install_command_with_retry` retries every nonzero process exit up to three times, but a self-update can replace/mutate the installation and then fail during later work. A nonzero exit therefore does not establish that retrying is safe; the next attempt may run a changed or partially updated binary. Updates need single-attempt execution unless Goose exposes a proven pre-mutation retry signal.

— Carl, reviewing on Wes's behalf

desktop/src-tauri/src/commands/goose_update.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+
+/// Check whether the resolved Goose CLI is behind the latest stable release.
+///
+/// The command is called only by Settings. It never runs `goose update` and
+/// therefore cannot modify the installed binary.
+#[tauri::command]
+pub async fn check_goose_update_status() -> Result<GooseUpdateStatus, String> {
+    let installed_version = tokio::task::spawn_blocking(|| {
+        let path = crate::managed_agents::resolve_command("goose")
```
wesbillman, line 42: Detecting that Goose is outdated is useful, but resolving a binary named `goose` does not establish that Buzz owns its lifecycle. This may be a Homebrew, Nix, or otherwise externally managed install, and other harnesses remain externally managed. Please use the detected state to show an **Update instructions** link to Goose's official docs rather than invoking self-update from Buzz. That gives users a useful signal without silently making Buzz responsible for one vendor's package management.

— Carl, reviewing on Wes's behalf

wesbillman (changes requested):
The update-availability detection is useful, but Buzz should not become a Goose-only package manager without an explicit lifecycle-ownership policy. Please keep the version check and replace the in-app mutation with a link to Goose's update/install documentation (or package-manager guidance). That is consistent with how Buzz treats other externally managed harnesses and avoids overwriting Homebrew/Nix/otherwise managed installations. If Buzz deliberately decides to manage Goose as an exception later, that needs install provenance/capability gating and single-attempt update semantics.

I also left a blocking inline correctness comment on the current retry path. CI being green does not exercise failure after the updater has already mutated the installation.

— Carl, reviewing on Wes's behalf

wpfleger96:
🤖 hey @lifeizhou-ap — thanks for putting this together, the detection half of this is genuinely solid work.

I'm with Wes on the direction here: we don't want Buzz mutating harness binaries it doesn't own (`resolve_command` picks up whatever's on PATH, including brew/nix-managed installs), so let's rescope this PR to detection + link-out. Concretely:

**Keep** — `check_goose_update_status` (the version probe + GitHub release check with the timeouts and redirect validation), the Ready/Update UI states, the failure-isolation design where a failed check never blocks readiness, and all the detection tests. That's most of the PR and it's well built.

**Change** — the Update button should open Goose's update docs instead of running the installer. The URL already exists in the catalog metadata (`cli_install_instructions_url`). Maybe rename it to something like "Update available ↗" so it reads as informational rather than an action Buzz performs.

**Drop** — the `cli_setup.rs` mutation path (`<goose> update`), the `install_acp_runtime` rewiring, the post-update invalidation, and the click-through-update e2e flow.

**One fix in the same pass** — the update check re-runs on every community switch because the `QueryClient` is created inside the community-keyed provider, even though the status is machine-global. Worth moving the release lookup to app/module scope so it fires once per session (with the explicit "check again" refetch kept).

One design note while you're in there: keep the release-lookup config (URL, parse strategy) in per-runtime metadata rather than hardcoding `goose` into shared types — we'll likely want the same detection for other harnesses later, and separately we're planning to pin a minimum supported Goose version once the scheduler fix ships in a release. This PR doesn't need to build either of those, just shouldn't make them harder.

PR #3843, opened 2026-07-31
fix(desktop): find profileless users by pubkey

alebairos:
## Summary

Allow Desktop channel member search to resolve an exact 64-character hex
pubkey or exact npub even when that community member has no kind-0 profile.

Exact identity queries now fetch the profile event and NIP-43 membership
snapshot together. Profile metadata is preserved when present; otherwise a
bare-key candidate is returned only when the identity is in the community
roster. Existing full-text profile/name search remains unchanged.

The client-side candidate ranker also recognizes canonical npub values so the
exact backend result is not filtered out before display.

### Related issue

N/A. No matching open issue or pull request was found.

### Testing

- `just ci`
  - workspace formatting and strict Clippy
  - repository unit suites
  - Desktop checks, full tests, production build, and Tauri checks
  - web checks and production build
  - mobile analysis and full tests (`875 passed`, `1 skipped`)
- Desktop native suite: `1,855 passed`, `14 ignored`
- Desktop mixer diagnostics: `3 passed`
- Regression tests cover exact hex, exact npub, a profile-less NIP-43 member,
  non-member rejection, open-relay fallback, and client-side npub ranking.
- Physical HITL test against `wss://buzz.alebairos.xyz`:
  - exact hex lookup found and added the profile-less Android identity;
  - exact npub lookup found and added the same identity;
  - independent channel-roster queries confirmed membership in both test
    channels.

There is no visual layout change, so before/after screenshots are not
applicable. Post-invite profile editing remains a separate follow-up.

### DIFF
--- a/desktop/src-tauri/src/commands/profile.rs
+++ b/desktop/src-tauri/src/commands/profile.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 
 use buzz_core_pkg::PresenceStatus;
+use nostr::PublicKey;
 use serde_json::Value;
 use tauri::State;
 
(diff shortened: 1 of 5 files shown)


Chessing234:
handy for bare keys. please make sure a typo of 63 hex chars still falls through to normal search and doesnt error the whole picker

alebairos:
Confirmed: exact-key lookup requires exactly 64 hex characters (or a valid npub). A 63-character hex typo returns `None` from exact parsing and continues through the existing NIP-50 search path.

PR #3845, opened 2026-07-31
feat(desktop): schema-certified env scaffolding for agent snapshots

monotykamary:
## Summary

Unify env/credential handling across snapshots, runs, and chat-completions, with a schema-certified contract path — and let an external control plane ship an agent manifest whose non-secret configuration (e.g. an OpenAI-compatible API route) arrives pre-wired, while credentials still stay out of snapshot files by construction.

1. **Schema-certified env scaffolding round-trip** — agent snapshots own the full `env_vars` schema; scaffolds hold key names only (`definition.environment`), never selected values, wrapper or not.
2. **`definition.environmentValues` — import-only value hints for non-secret config.** Buzz's own export never writes this field. At import, every hint must reference a key also declared in `environment`; keys whose name marks a credential (e.g. `*_API_KEY`, `*_TOKEN`) are dropped to blank scaffolding so a producer can never smuggle a live secret into a manifest. Both count- and per-value-capped at decode (`MAX_SNAPSHOT_ENV_KEYS` / `MAX_ENV_VALUE_BYTES`). Motivating case: a hosted control plane exports `OPENAI_COMPAT_BASE_URL` so the imported agent starts pointed at the right endpoint; only the API key is pasted by the owner. The import preview surfaces which keys arrive pre-filled.
3. **Run-scoped env overlay** (e.g. from slash options) merges into `agent_runtime_env` for declared keys, keeping schema-owned contracts intact across every run mode — including contract-less exported imports and legacy keyless snapshots.
4. **`agent_runtime_env` is the only consumer-visible truth surface** for declared env values (canonical resolution path) — but only where the values fit the declared schema. Selected credential names are recycled as the snapshot's own values, never as provider-scoped storage.
5. **`env.get_credential`** resolves through the central credential-name normalizer, so ambiguous inputs uniformly fail closed instead of edge-case-matching a real credential.
6. **Persona export/import propagation** — sub-agent env scaffolding exports from `snapshot_json`, imports back through the shared splitter, and legacy top-level `env_vars` blobs round-trip partner-aware.
7. **Loop models endpoints** — sub-agent chain flattening follows canonical flatten refs, and loop exports bundle the resolved env scaffolding team refs carry.
8. **Credential format linter + contract schema HTTP shape enforcement** — natural lint, source-only path, bulletproof coverage.
9. **Env vars module is the only keeper of secret-bearing literals** — input-path write and normalization mixed-state leaks are defused.

Key invariants preserved:

- Run-scoped overlay values bypass credential staging; credential truth still lives in `agent_runtime_env` reads.
- Contract terms fire only when schemas are present — no silent LLM-key staging.
- No new credential-bearing HTTP surface.
- When a key resolves credentially, nothing changes about what actually reaches a child process.
- The `env.*` and `agent_env.*` API contracts are untouched.

### Related issue

None found (searched open PRs/issues for environment/snapshot overlap). Producer-side counterpart (external control plane emitting `environmentValues` hints): app.inloop.studio PR #476.

### Testing

`cargo test --lib` — full suite green (2,000+ tests), including new coverage for:

- value hints: pre-filled non-secret keys land, secret-named keys drop to blank, undeclared keys ignored, empty strings treated as blank, oversize count/value rejected at decode, Buzz export never serializes the field
- wrapper↔owner scaffolds (never value-bearing)
- schema shape → auth-injection path
- legacy keyless / contract-less export-import round-trips
- unresolved-mixed overlay honesty
- `env.get_credential` resolution edges
- per-run HTTP contract-schema path reaching the same merge gate
- `egress_guard` inventories updated for the persona-import module split

Also `cargo clippy --lib --all-features` clean and `cargo fmt --check` clean.


Chessing234:
huge diff. if you can carve the schema-certified path into its own pr, review will go a lot faster

PR #3849, opened 2026-07-31
fix(relay): serialize NIP-IA snapshot timestamps

### LINKED ISSUE #3848: NIP-IA archive snapshot can lose rapid same-second updates
## Problem

Rapid NIP-IA archive operations can emit multiple relay-signed `kind:13535` replaceable snapshots with the same second-resolution `created_at`. NIP-01 resolves same-second replaceable events by the lowest event ID, so an intermediate snapshot may remain authoritative even though every `kind:8002` delta and canonical `archived_identities` row was accepted.

Observed behavior: six rapid archive requests produced six valid archive deltas, while the latest authoritative snapshot exposed only three archived identities.

## Expected behavior

Every accepted archive or unarchive mutation should publish a snapshot that supersedes the previous snapshot deterministically, including multiple mutations within one wall-clock second.

## Proposed fix

Before signing `kind:13535`, query the existing relay-authored snapshot and assign `created_at = max(now, previous_created_at + 1)`, matching the monotonic timestamp guards already used for Buzz discovery and DM visibility snapshots.

## Regression coverage

Add a test that simulates six snapshot updates during one wall-clock second and verifies strictly increasing timestamps from `T` through `T+5`.

Proudly Made in Nebraska. Go Big Red! 🌽 https://xkcd.com/2347/

CryptoJones:
## Summary

- make relay-authored NIP-IA `kind:13535` snapshot timestamps strictly monotonic
- prevent same-second NIP-01 event-ID tie-breaking from retaining an intermediate archive set
- add regression coverage for six rapid snapshot replacements and wall-clock catch-up

Fixes #3848.

## Verification

- `cargo clippy -p buzz-relay --all-targets -- -D warnings`
- `cargo test -p buzz-relay` — 794 passed, including both new regression tests; 10 unrelated environment-dependent failures because the local Postgres instance lacks role `buzz`, plus one mesh demo timeout

Proudly Made in Nebraska. Go Big Red! 🌽 https://xkcd.com/2347/


Chessing234:
makes sense for #3848. maybe assert the wall-clock catch-up path in the test name so future readers know why sleep is there

Chessing234:
serializing nip-ia snapshot timestamps is a good fix for #3848. a tiny test with two same-second updates would lock it in.

Chessing234 (commented):
this is the right fix for #3848 — same-second replaceable ties keeping the lowest id was eating intermediate nip-ia snapshots. the unit tests for strictly increasing timestamps are exactly what i'd want. one ask: if six archives land in one wall-clock second, created_at jumps into the future by up to 5s; confirm clients tolerate `created_at > now` on relay-signed snapshots (most do) so you don't trade lost updates for 'future event' filters.

CryptoJones:
Hi @Chessing234 — addressed your feedback from this thread. I renamed the test covering the wall-clock catch-up path from `replaceable_snapshot_timestamp_does_not_regress_behind_wall_clock` to `wall_clock_catch_up_advances_snapshot_timestamp` so future readers immediately know which code path it exercises (and why a `sleep` might be added there for real-time testing). Both regression tests pass.

CryptoJones:
@Chessing234 Thanks for flagging the future-timestamp case. The updated head `66df23a0f` now closes all three points:

- The wall-clock catch-up test name is explicit.
- Same-second snapshots use strictly increasing timestamps.
- First-party desktop and CLI snapshot queries do not apply an `until` filter or reject `created_at > now`. To keep that behavior bounded for other clients, snapshot publication now pauses at a 60-second future-skew ceiling, and `docs/nips/NIP-IA.md` documents that client contract.

The update also closes a concurrency gap found during follow-up review: if two publishers calculate the same replacement timestamp, the loser now logs the collision, re-reads both the canonical archive rows and latest snapshot, and retries rather than silently dropping or republishing stale tags.

Local validation on the refreshed branch: `cargo clippy -p buzz-relay --all-targets -- -D warnings` passes. The full relay suite reports 835 passing tests; its one remaining failure is the pre-existing mesh-demo echo timeout (`504` vs `200`), outside this diff.

Proudly Made in Nebraska. Go Big Red! 🌽 https://xkcd.com/2347/

CryptoJones:
Follow-up: final head `46465fe43` also bounds the collision/skew retry loop to eight attempts. A backward clock step therefore cannot retain an ingest-handler permit indefinitely; skew pauses are logged and exhaustion returns through the caller's existing warning path. The added retry-budget regression test brings the full relay run to 836 passing tests, with the same unrelated mesh-demo timeout remaining.

Proudly Made in Nebraska. Go Big Red! 🌽 https://xkcd.com/2347/

PR #3850, opened 2026-07-31
fix(release): reject clobbered macOS updater archives

dwillitzer:
## Summary
- Root cause: both macOS release jobs uploaded `Buzz.app.tar.gz` to `buzz-desktop-latest`, so arm64 clobbered Intel while `latest.json` kept both platform signatures.
- Main already renames per-arch updater archives; this fails assemble if two platforms still share a basename before `latest.json` is published.
- Documents manual DMG recovery for machines stuck on the broken channel.

## Test plan
- [x] Local bash simulation: unique names pass, duplicate basename fails
- [x] Live channel rechecked: both macOS keys still point at one `Buzz.app.tar.gz` with different signatures
- [x] Confirmed `Buzz_0.5.2_x64.dmg` exists for Intel recovery
- [ ] Next desktop release: confirm `latest.json` macOS URLs are uniquely named archives
- [ ] On Intel Mac: after next release, Settings → Software Updates can install without signature failure

## User recovery (current channel)
Install the matching DMG from https://github.com/block/buzz/releases/latest — Intel needs `Buzz_*_x64.dmg`. In-app update stays unreliable until a release ships a correct `latest.json`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -890,23 +890,32 @@ jobs:
           set -euo pipefail
           BASE="https://github.com/block/buzz/releases/download/desktop-v${VERSION}"
           TRIPLES=()
+          # Parallel arrays: ARCHIVE_NAMES[i] basename claimed by ARCHIVE_PLATFORMS[i].
           ARCHIVE_NAMES=()
+          ARCHIVE_PLATFORMS=()
 
           add_triple() {
             local result="$1" platform="$2" archive="$3"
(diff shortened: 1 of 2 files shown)


Chessing234:
good guard. can the check print both colliding artifact urls in the failure log so the release job is obvious to debug?

dwillitzer:
Good call. Pushed a follow-up that prints both colliding platforms and their intended desktop-v artifact URLs on failure.

PR #3855, opened 2026-07-31
fix(desktop): expand ancestors of nested thread replies on panel open (#3799)

### LINKED ISSUE #3799: Missing replies in thread view after replying to non-latest messages
Description: When an agent replies to a message that is not the most recent in a thread, the reply is delivered and visible in the agent's message history, but does not appear in the thread panel UI on the right side.

Steps:
1. Start thread with message A
2. Agent replies to A (appears in thread)
3. User adds message B to same thread
4. Agent replies to B
5. Agent's reply to B is missing from thread panel UI (though it exists in relay)

Expected: All replies appear in thread panel, chronologically ordered
Actual: Replies to earlier messages become invisible in UI

Environment: Buzz Desktop 

Image

iroiro147:
## Problem

When an agent (or any replier) anchors a reply to a *non-latest* thread message — e.g. parent=B where the viewer's open thread has A as head and B as a depth-1 child — a fresh depth-2 reply delivered by the relay is **not rendered** in the thread panel and produces **no visible signal**. The two underlying surfaces:

| Surface | Root cause | Consequence |
|---|---|---|
| render | The panel opens with an empty `expandedThreadReplyIds` set (`ChannelScreen.tsx`). A depth-1 child that has any descendant renders as a collapsed `MessageThreadSummaryRow` and every deeper reply under it is omitted from `visibleReplies` (`threadPanel.ts` `appendExpandedReplies`). | The new depth-2 reply is present in `threadRepliesKey` but invisible in the panel. |
| badge | `computeThreadReplyUnreadCounts` (`useChannelUnreadState.ts`) builds its input set from the **channel-window projection only**. A fresh depth-2 reply that lives only in the `threadRepliesKey` cache isn't in that window, so the descendant-aware subtree walk never reaches it. | The collapsed ancestor row's unread badge stays `0`. |

Both are the same logical bug at two stages: the open thread's fetched reply set (`[head, ...replyEvents]`) is richer than the channel-window projection that seeds the panel state, and neither surface reconciled against it.

## Fix

Two tightly-scoped changes, each with independent regression coverage:

- **scope-a — expand ancestors at open.** New `computeInitialExpandedReplyIds` (pure, in `threadPanel.ts`) computes the minimal set of intermediate ancestors of every depth-≥2 reply in the fetched thread set. New `useThreadPanelInitialExpansion` hook (extracted out of `ChannelScreen.tsx` to respect the desktop file-size ratchet) seeds `expandedThreadReplyIds` once per opened head by unioning those ancestors in — the seed never removes ids and never re-runs after the user manually collapses/expands. Idempotent (bails with the identity reference when the seed is already a subset of the live set).
- **scope-b — union the message set into the unread badge.** New `unionScopeMessages` (pure, in `threadReplyUnreadCounts.ts`). `useChannelUnreadState` now feeds `computeThreadReplyUnreadCounts` a union of the channel-window projection and the open thread's full fetched message set via a *widened* `directReplyIdsByParentId` memo consumed **only** by this badge path — all other unread readers stay window-scoped, matching the existing per-message read-state contract.

## Verification

- `threadPanel.test.mjs`: **6 new cases** — open-with-seed surfaces the depth-2 reply; depth-1-only threads stay collapsed; unread-scoped filtering expands only the unread branch; depth-3 chain pins every intermediate parent; cycles/missing parents tolerated without hanging; `buildThreadPanelData` integration with the seed.
- `threadReplyUnreadCounts.test.mjs`: **3 new cases** — `unionScopeMessages` dedup/order/reference-preservation; a depth-2 reply outside the window is counted against its collapsed ancestor (the fix); and a negative control documenting that, without the union, the same scenario yields `unreadDescendantCount = 0`.
- **Full desktop suite: 3879/3879 green.**
- `pnpm check` (biome lint+format, file-size ratchet, px-text, pubkey-truncation): clean.

## Manual test

1. Open a channel with an existing thread A. Send a reply B to A (depth-1). Open a second session and have an agent (or another account) reply to **B** (depth-2), not to A and not to the latest thread message.
2. In the first session, open the thread panel for A. **Before** this fix: B renders as a collapsed summary row; the agent's reply is invisible; B carries no unread badge. **After**: B is pre-expanded; the agent's reply renders in place; B carries an unread badge while collapsed from any prior manual collapse.

Affected surface: desktop web thread panel only. No protocol, relay, storage, or key changes.

Fixes #3799

### DIFF
--- a/desktop/src/features/channels/lib/threadReplyUnreadCounts.test.mjs
+++ b/desktop/src/features/channels/lib/threadReplyUnreadCounts.test.mjs
@@ -1,7 +1,10 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 
-import { computeThreadReplyUnreadCounts } from "./threadReplyUnreadCounts.ts";
+import {
+  computeThreadReplyUnreadCounts,
+  unionScopeMessages,
+} from "./threadReplyUnreadCounts.ts";
 
 // Open thread "root":
 //   root(100)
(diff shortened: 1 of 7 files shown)


iroiro147:
Closing to re-push with DCO sign-off on the sole commit; reopening as a fresh PR momentarily.

PR #3852, opened 2026-07-31
feat(mobile): add profile editing

alebairos:
## Summary

Allow the active Mobile identity to create and update its Nostr kind-0 profile
from Settings.

The editor supports display name, avatar, and bio while preserving unexposed
profile fields and event tags. Saves are signed by the active identity,
published to its current community, and reflected locally only after relay
acceptance. If the active identity or relay changes during the edit, the save
aborts instead of publishing with stale configuration.

Profile avatars are selected at a bounded 512 by 512 resolution, converted to
JPEG, and uploaded through the authenticated media path. This avoids reverse
proxy rejection of multi-megabyte device photos without changing the
full-resolution message-attachment path.

### Related issue

N/A. This follows the post-invite profile-less identity gap discovered while
testing #3835 and the ID-search improvement in #3843.

### Testing

- Full repository `just ci`
  - workspace formatting and strict Clippy
  - repository unit suites
  - Desktop checks, full tests, production build, and Tauri checks
  - web checks and production build
  - Mobile analysis and full tests (`1,032 passed`, `1 skipped`)
- Desktop native suite: `2,003 passed`, `14 ignored`
- Desktop mixer diagnostics: `3 passed`
- Regression tests cover:
  - profile-less kind-0 creation and signed publishing;
  - existing JSON field and event-tag preservation;
  - relay rejection without optimistic saved state;
  - identity/relay changes during an edit;
  - bounded avatar selection and JPEG conversion.
- Physical HITL test on a Moto G56 against the VPS community:
  - invite-created identity saved display name and bio;
  - Desktop discovered the identity by its saved name;
  - replacement edits preserved existing profile data;
  - the original failing large photo and a second photo uploaded and persisted;
  - avatar rendered in Desktop and Mobile's home-screen header.
- VPS nginx independently recorded HTTP `200` for both revised avatar uploads;
  the immediately preceding full-resolution request reproduced HTTP `413`.

Buzz originating channel: `c26bd1cd-860c-4f0e-8b8c-54f5ac3a6a56`

PR #3859, opened 2026-07-31
feat(mobile): add signed profile editor

bluepinkneon:
Adds an in-app editor for the active community identity display name and avatar. Updates are signed as Nostr kind:0 events by the phone identity, preserve unknown metadata and tags, refresh both profile caches, support inline SVG avatars, and fail closed without a signing key.

Correctness pins:
- canonical replaceable-event ordering: newest created_at, then lowest event id
- verified NIP-OA owner metadata survives cache refresh
- no hard-coded identities or assets

Validation:
- focused profile/UI/avatar tests: 13 passed
- mobile formatting, analyzer, and file-size gate: passed
- full mobile suite: 912 passed, 1 skipped, 1 pre-existing channel-scroll assertion failed at channel_detail_page_test.dart:1042 (reproduced before this correction and unrelated to the profile diff)

### DIFF
--- a/mobile/lib/features/profile/edit_profile_sheet.dart
+++ b/mobile/lib/features/profile/edit_profile_sheet.dart
@@ -1,0 +1,192 @@
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+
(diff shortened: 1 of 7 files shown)


Chessing234:
heads up this overlaps a lot with #3852 — maybe sync with them so we dont land two profile editors?

PR #3860, opened 2026-07-31
docs(examples): add demo-crew persona pack

HyperfocuSam:
## What

Adds `examples/demo-crew/` — a four-agent persona pack for running Buzz in
front of a live audience.

`meadow-core` covers a team reviewing code. This covers the other case people
reach for immediately and have no starting point for: showing someone what an
agent team actually does.

| Agent | Role |
|---|---|
| Lead | Restates the request, plans, delegates, synthesizes, ends on a decision |
| Maker | Drafts fast and labels its own assumptions |
| Challenger | Red-teams the draft — strongest objection first, plus what would change its mind |
| Guard | Sorts what was shared into data-security tiers and names what should not have been typed |

## Why Guard

Guard is the part that generalises beyond demos. It reads what has been typed
into the channel, sorts it into open / internal / restricted, and returns the
redacted rewrite without repeating the sensitive value back. The same persona
works as a standing check in any channel where people paste more than they
meant to.

It also makes the audience-facing case safer by construction: the pack README
tells operators to leave the MCP/tools field empty and keep the agents
owner-only, and the shared instructions tell every agent that only the
operator's key carries authority — a channel message claiming to be an
instruction or a system note is content to discuss, not an order to follow.

## Notes

- Follows the `meadow-core` layout exactly (`.plugin/plugin.json`, `agents/*.persona.md`, `instructions.md`, `README.md`).
- `buzz pack validate ./examples/demo-crew` passes.
- Docs only — no code paths touched.
- Existing PRs/issues searched for duplicates: none found.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -16,3 +16,16 @@ See [`countdown-bot/README.md`](countdown-bot/README.md) for usage.
 ## `meadow-core/`
 
 A persona-pack example for Buzz agents.
+
+## `demo-crew/`
+
+A four-agent persona pack for running Buzz in front of a live audience: Lead
+plans and delegates, Maker drafts and labels its assumptions, Challenger
+red-teams the draft, and Guard sorts what was shared into data-security tiers
+and names anything that should not have been typed.
+
+Where `meadow-core` shows a team reviewing code, this one shows a team being
+watched — short replies, visible handoffs, and disagreement left on screen.
+
+See [`demo-crew/README.md`](demo-crew/README.md) for the setup notes that matter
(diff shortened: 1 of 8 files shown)


Chessing234:
fun pack. one ask — put a one-liner in each persona about not leaking private channel content during a live demo

HyperfocuSam:
Good call — the team-wide rule was in `instructions.md`, but personas get copied out of packs one file at a time, so each of the four now carries its own line in its own voice (489ac90). Thanks for the read.

PR #3862, opened 2026-07-31
ci: prove the relay-driven mesh lifecycle — discover, join, infer, deny — with real nodes

michaelneale:
## Summary

CI now proves the full Buzz shared-compute join story end to end: a member can discover another member's served model **through the Buzz relay alone** and run inference over the mesh, while a non-member gets nothing — the relay rejects its auth, and the mesh refuses to route for it even holding a leaked endpoint address.

This is deliberately different from mesh-llm's own CI smokes (which bootstrap two nodes with a hand-carried invite token / mdns): here the **relay is the control plane**, exactly like the desktop app:

1. **Membership** — identities A and B are added via `buzz-admin` (kind:13534 NIP-43 roster); C is not.
2. **Advertise** — each member publishes a client-signed kind:30003 discovery note carrying its MeshLLM owner binding and (for the serve node) `serveTargets[].endpointAddr`, covered by an endpoint-binding signature — the exact payload shape the desktop coordinator publishes.
3. **Trust** — the serve node derives its admission allowlist from the relay (statuses ∩ roster) and requires the **exact expected {A, B} owner-id set** before starting with `TrustPolicy::Allowlist`.
4. **Join** — the client verifies owner + endpoint bindings and membership, then dials the relay-discovered endpoint (the desktop join-watcher's `dial_endpoint_addr` step). No out-of-band token.
5. **Infer** — a chat completion against the client's local OpenAI endpoint routes over QUIC to the serve node's model (CPU, SmolLM2-135M, ~105MB).
6. **Deny (differential)** — the stranger's NIP-42 auth must fail with the relay's own membership rejection (`restricted: not a relay member` — successful auth or any unrelated connect error fails the run), and dialing the leaked endpoint must not produce a routed inference — **while the trusted client re-proves inference immediately afterwards**, so a dead serve node can't masquerade as an admission denial.

## What's in the PR

- `crates/buzz-relay/examples/mesh_relay_lifecycle_smoke.rs` — the harness. One process per node (mesh-llm keeps process-global state under `~/.mesh-llm`), orchestrator + serve/client/stranger roles, byte-identical binding payloads to `desktop/src-tauri/src/mesh_llm/identity.rs` (called out with keep-in-sync comments). Child stdout is pumped through a reader thread so every wait has a hard deadline; timed-out children are killed; exit statuses are checked.
- `scripts/ci-mesh-lifecycle-smoke.sh` — provisions a membership-gated relay (throwaway owner + signing identities via `buzz-admin generate-key`), runs the harness, cleans up. Fails fast if :3000 is already occupied (a stale open relay would mask gating).
- `scripts/start-relay-for-tests.sh` — gains opt-in NIP-43 membership env passthrough (`BUZZ_REQUIRE_RELAY_MEMBERSHIP` + `RELAY_OWNER_PUBKEY` + `BUZZ_RELAY_PRIVATE_KEY`). Default behavior unchanged.
- `.github/workflows/mesh-lifecycle.yml` — separate, path-filtered, non-required workflow (mesh paths, the harness's dependency crates, `Cargo.lock`, dispatch), pinned to `ubuntu-24.04`. Caches the mesh native runtime + HF model keyed on the lockfile hash, so a mesh pin bump rolls the runtime cache. Uploads relay + harness logs on failure.

## Scope

This is an **independent protocol harness**: it speaks the same wire protocol and payload shapes as the desktop but re-implements the binding/verification logic (the desktop crate is outside the workspace). Regressions inside the desktop's own discovery filtering are the desktop unit tests' job; what this smoke proves is that the relay + mesh-llm SDK + admission stack support the lifecycle end to end.

## Relationship to mesh-llm's CI

Follows the shape mesh-llm's own CI proved stable (tiny CPU model, one runner, multiple real mesh-llm processes over real QUIC — cf. their `ci-two-node-client-serving-smoke.sh`), but swaps the token bootstrap for the relay-driven lifecycle, which is the part only Buzz can test.

## Validation

Green on GitHub Actions (ubuntu-24.04) across three runs, including after rebases onto the mesh v0.74 upgrade (#3467) and latest main:

```
PASS 1/6: relay-derived allowlist is exactly {A, B}
PASS 2/6: serve member ready + advertised model: jc-builds/SmolLM2-135M-Instruct-Q4_K_M-GGUF:Q4_K_M
PASS 3/6: client member discovered + joined via relay
PASS 4/6: inference routed over the mesh: "PONG"
PASS 5/6: relay rejected the stranger's NIP-42 auth (membership gate)
PASS 6/6: stranger denied (gossip visible, inference rejected: 503 all tunnels failed) while trusted inference still routes
PASS: full relay-driven mesh lifecycle verified
```

Also validated locally on macOS. `cargo fmt --all --check` and `cargo clippy -p buzz-relay --all-targets -- -D warnings` pass.

## Notes

- The harness follows the repo's mesh `[dev-dependencies]` pin automatically, so it doubles as a canary for future mesh upgrades (it already caught the v0.73.1 → v0.74.0 bump during development).
- The stranger "deny" accepts either shape mesh-llm exhibits: no model visibility at all, or gossip visibility with inference refused — mesh-llm applies the receiving node's owner policy after the gossip handshake, so admission gates *routing*, not gossip. The differential trusted-inference re-check (PASS 6/6) is what makes that a real denial rather than a dead server.
- Model-visibility windows are tunable via `MESH_CLIENT_WINDOW_SECS` / `MESH_STRANGER_WINDOW_SECS` if shared runners prove slow — pin a longer window in the workflow env rather than re-running the job.


Chessing234:
cool ci. if the smoke is flaky on shared runners, maybe pin a longer join timeout in the workflow rather than retrying the whole job

micspiral:
This is the right shape and it covers something nothing else does — relay-as-control-plane, with no out-of-band token. The membership/binding/admission half (phases 1–3, 6) is genuinely unique to Buzz and worth landing on its own merit. Also worth noting the stated draft gate is already met: Mesh Lifecycle has three green runs on Actions now.

I've spent the last few days chasing two shared-compute bugs to root cause (#3909, and Mesh-LLM/mesh-llm#1126 / #1129), so I wanted to write down where this harness would and wouldn't have caught them — and a concrete, cheap way to extend it. Take or leave any of it.

## What it would not have caught today

I grepped the diff rather than going on impressions:

| failure class | coverage in this PR |
|---|---|
| restart loop — supervisor reads *busy* as *dead* | `watchdog\|request_restart\|should_evict\|wait_for_mesh_inference` → **0 refs** |
| reply never reaches a channel | `buzz-acp\|buzz-agent\|dev-mcp\|messages send\|tool_call` → **0 refs** |
| preset env (`thinking_effort` etc.) | doesn't call `apply_relay_mesh_env`; the `desktop/` mentions are keep-in-sync comments citing `identity.rs` |

That isn't a criticism of scope — it's that both bugs live in the *timing* and *agent* layers, and this harness deliberately sits below them. The inference phase is `SmolLM2-135M` (~105MB) answering `"Reply with exactly one word: PONG"`. That combination loads and prefills fast enough that the node is essentially never unresponsive, so no timing bug can surface.

## Suggestion 1 — a stalled-node phase (this is the cheap, high-value one)

The bug behind #3909 was that a node which is **bound but HTTP-silent** — loading weights, pulling layers, or mid-prefill — was read as dead and force-restarted the whole app. Measured on a live 26B node: `tcp_connect` stayed at **0.2ms** while every HTTP endpoint (including a lock-free `/health` I'd added to a patched mesh build) timed out for **~21s**.

The valuable insight for CI is that **this is a timing bug, not a size bug**. You don't need a big model or a GPU to reproduce it — you only need the node to stop answering HTTP while its port stays bound.

The harness already spawns and holds each node as a `Child`, so this is close to free:

```rust
// phase N: a busy/loading node must not be classified as dead
kill(serve_pid, SIGSTOP);
assert!(tcp_connect(api_addr).is_ok());          // port still bound -> alive
assert!(http_get(api_addr, "/v1/models").is_timeout());  // ... but silent
kill(serve_pid, SIGCONT);
assert!(http_get(api_addr, "/v1/models").is_ok());       // recovers
```

`SIGSTOP` gives exactly the supervisor-visible signature — listener open, backlog accepting, zero HTTP responses — deterministically, in seconds, on any runner, with no new dependencies.

One honest caveat: `SIGSTOP` freezes the QUIC/gossip stack too, whereas a genuinely busy node keeps those running. So it faithfully reproduces *what the supervisor sees* (which is what the watchdog keys on) but it is not a full simulation of a busy node. For this purpose I think that's the right trade.

If it's worth asserting the Buzz-side classification directly rather than just the wire behaviour, `probe_mesh_ingress_at` in `desktop/src-tauri/src/mesh_llm/recovery.rs` returns exactly `Live` / `Unhealthy` / `PortClosed` — though it's `pub(crate)` today, so that would need visibility widening or a duplicated probe. Wire-level asserts alone would still have caught the regression.

## Suggestion 2 — a prose-only agent phase

The second failure was subtler: a small model answers in prose and never calls the send tool, so the turn ends `outcome="ok"` with **nothing in the channel** — a silent drop. #3763's reply guard now nags twice and rerolls, which helps, but by design it cannot compel a model that simply won't tool-call.

A deterministic stub is enough to pin the contract, and needs no model at all:

- a stub ACP agent that only ever streams prose and never emits a tool call;
- assert on **relay state** — read the event back with a *different* identity and correlate by `e`-tag.

Reading relay state is the part I'd push hardest. During my own testing the agent's self-report and the harness logs both told me a reply had landed when it hadn't; only an independent authenticated read was trustworthy.

## On soak

Deliberately **not** suggesting soak in CI. The restart loop needs hours to surface naturally, which is hopeless for a PR gate — and it's already covered in real life: I've had the desktop app sharing a 26B model for **2 days 4 hours** continuous with zero restarts on the post-#3909 build (previously it bounced within minutes of a cold 26B share). That's the right place for soak. CI's job is the deterministic version of the same failure, which is what Suggestion 1 buys.

## Suggestion 3 — multi-node MoA, as a separate thing

Different question, worth separating: does routing to the virtual `mesh` model (`VIRTUAL_MODEL_NAME` in `crates/mesh-mixture-of-agents`, fan-out + reducer with early-exit) actually produce a measurably *better* answer than the best single node? That's a quality/lift question, not a correctness one.

My honest view is it doesn't belong in this workflow, or possibly in Buzz CI at all:

- it needs ≥2 nodes hosting genuinely *different* models to mean anything (fan-out over one model measures nothing);
- lift needs a scored task set and repeat runs — a pass/fail smoke can't express it;
- it's inherently noisy, so as a required gate it would flake.

Better as either a mesh-llm-side benchmark (`skippy-bench` already has the chat-corpus scaffolding) or a periodic lab job on real multi-host hardware, reporting a trend rather than gating. What *could* live here cheaply is the far weaker but still useful assertion that `model: "mesh"` **routes and returns** across two nodes — correctness of the MoA path, not its intelligence.

## Multi-host

Everything here is single-runner over loopback, which is right for CI. Real cross-host QUIC, real RTT and contention need actual separate machines and belong in a pre-release lab tier rather than a PR gate. Two things I'd carry over from doing that by hand: **pin the model per node** (an `--auto` node swapped models mid-run and voided a test for me), and **verify what's actually running by content, not version string** (I had three binaries all self-reporting the same version; only grepping the built artifacts distinguished them).

michaelneale:
🤖 (from micn's agent) Good call — addressed in 428177457: the model-visibility windows are now tunable via `MESH_CLIENT_WINDOW_SECS` / `MESH_STRANGER_WINDOW_SECS` env vars, so a slow shared runner can pin longer windows in the workflow instead of retrying the whole job. The zizmor pin-comment mismatch is fixed in the same commit (checkout comment now names the exact tag v6.0.3). Mesh Lifecycle is green on that head: all 6 lifecycle assertions pass.

michaelneale:
🤖 (from micn's agent) Thanks for the root-cause writeup — this is exactly the analysis we wanted. I ran your comment plus the harness internals past an independent expert review; summary of where we landed:

**On "would it have caught #3909 / the prose-reply drop": agreed, no.** Both live in layers this harness deliberately sits below (desktop supervisor timing; ACP agent behavior). No pushback there.

**On Suggestion 1 (SIGSTOP stalled-node phase): we're not adding it to this PR, for a sharper reason than cost.** The expert's key observation: before #3909, `probe_mesh_ingress_at` *already* returned `Unhealthy` for the bound-but-HTTP-silent signature — the broken part was the decision layer (`Unhealthy streak ≥ 2 → evict → restart`). So all three wire-level SIGSTOP assertions (TCP ok, HTTP timeout, recovers after SIGCONT) pass against **both** the broken and fixed implementations — they'd reproduce the *signature* without pinning the *regression*. The thing that pins it is `should_evict_after_probe` / supervisor behavior, and the post-#3909 unit tests in `desktop/src-tauri/src/mesh_llm/recovery.rs` already do that (including the streak logic and a black-box stalled listener). A SIGSTOP phase here would be green paint over coverage that already exists in the right layer. If we later want a "brief suspension + recovery" lifecycle phase for its own sake, the right slot is after the stranger exits and before the existing `VERIFY_AGAIN` recheck (never during a QUIC handshake), with `waitpid(WUNTRACED)` sync, a fresh HTTP connection, a SIGCONT-on-drop guard (a stopped child can't process the cleanup SIGTERM), and a 1–3s freeze to stay under peer-eviction timers. Happy to do that as a follow-up if you think the recovery unit tests leave a gap.

**On Suggestion 2 (prose-only agent + independent relay read): agreed it's valuable, and it's a follow-up** — partly because it first needs a product decision the test would encode: when an agent never tool-calls, is the contract auto-publish, an explicit failed outcome, or "reminder then send"? #3763's guard nags but can't compel; a stub test has to assert *something specific*. Your "read the event back with a different authenticated identity, correlate by e-tag" principle is the part we'll definitely keep — the expert independently flagged self-reports as untrustworthy for exactly the reason you hit.

**On Suggestion 3 (MoA) and soak: agreed on both** — MoA lift belongs in mesh-llm benchmarks or a periodic lab job; a cheap `model: "mesh"` routes-and-returns smoke could be a separate non-required workflow later; soak stays out of CI (your 2d4h live run is the right venue).

One more gap the expert added that's worth recording for follow-ups: the harness reimplements the binding/verification logic, so publisher and verifier could drift *together* and stay self-consistent while the desktop breaks — the keep-in-sync comments are the only guard. Extracting the pure binding/discovery logic into a shared crate both sides link would close that properly, and is probably the highest-value follow-up of all of these.

Scope of this PR stays as-is: the relay-as-control-plane lifecycle (phases 1–6). Follow-ups, roughly in value order: shared binding crate → prose-agent contract test → optional suspension/recovery phase → `model: "mesh"` routing smoke.

tlongwell-block (dismissed):

tlongwell-block (approved):

PR #3864, opened 2026-07-31
fix(desktop): gate the add-member search on the relay-enforced add policy

Shipitrealgood:
## Problem

The add-member search dropped every agent-classified candidate not in the local managed store (`isAgentIdentityInManagedList`), so external (non-desktop-managed) agents could never be offered for channel adds — even when their own declared policy explicitly allows it. The relay's actual authority for third-party adds is the agent's kind:10100 `channel_add_policy`, enforced server-side in `buzz-relay` `handlers/side_effects.rs` (`owner_only` checked against the stored attested owner, `nobody` refused for everyone, `anyone` allowed, self-adds exempt). The client never consulted it — the field wasn't even carried through the `RelayAgentInfo` conversion.

Part of the external-agents family tracked in #2987 (this surface was point-adjacent to the mentions gap); originally sketched in the #3448 discussion.

## What this changes

**Wire types:** `channel_add_policy` now flows kind:10100 content → `RelayAgentInfo` (Rust, `#[serde(default)]`) → `RawRelayAgent` → `RelayAgent.channelAddPolicy` (typed `"anyone" | "owner_only" | "nobody" | null`).

**Picker gate:** `MembersSidebar` swaps the managed-list gate for `shouldHideAgentFromAddSearch`, a predicate aligned with relay enforcement:

- managed agents: always offered (native parity, unchanged)
- external agents: offered when their declared policy is `anyone`, or `owner_only` when the viewer is the NIP-OA-verified owner (same verified `ownerPubkey` the "managed by" surface renders — same provenance as the relay's stored owner mapping)
- `nobody`: hidden from everyone including the owner (the relay refuses those adds regardless of actor)
- unknown/missing policy values: hidden (fails closed)

The change is strictly widening: nothing previously visible becomes hidden.

**One deliberate tightening:** agents with no directory entry stay hidden even though the relay — whose stored policy defaults to `anyone` — would accept the add. No declaration, no offer; this preserves the stale-identity suppression the managed-list gate was introduced for (#2149).

**Why not gate adds on mentionability:** `respond_to` and `channel_add_policy` are independent declarations with independent relay enforcement — an agent can be mentionable but refuse adds (`respond_to: anyone` + `channel_add_policy: nobody`) or addable but not mentionable (`channel_add_policy: anyone` + `respond_to: owner-only`). Each surface reads its own declaration: `respond_to` for mentions (#3676), `channel_add_policy` for adds (this PR).

Enforcement stays server-side: the picker mirrors the *declared* policy, and the rare divergence between declaration and relay-stored state (first-write-wins owner mapping, last-validated policy) surfaces as a visible add error rather than a silent gap.

## Testing

- Predicate matrix: people / managed / `anyone` / `owner_only` as owner, non-owner, unverified / `nobody` incl. owner / no directory entry / null policy / pubkey normalization (`agentAutocompleteEligibility.test.mjs`).
- Wire contract on both hops — a misnamed field anywhere in this chain fails closed and silently hides agents app-wide, so it's pinned by tests: kind:10100 content → `RelayAgentInfo` (Rust directory-parse test) and `channel_add_policy` → `channelAddPolicy` (`fromRawRelayAgent`, `tauri.test.mjs`).
- Rust side executed locally via the repo's `desktop-tauri-test` recipe: 1,890 passed / 0 failed (14 ignored), including the amended directory-parse tests.
- End-to-end (added in review): the mock bridge seeds now carry `channel_add_policy`, plus paired smoke specs (`add-search-eligibility.spec.ts`) — an agent declaring `nobody` never appears in add search; an `anyone` control proves the deny comes from the policy gate. Run locally headless (2/2 passing) and in CI's `desktop-smoke-e2e`. Full desktop unit suite 3,791/3,791, repo-wide Biome, and typecheck clean; existing e2e specs swept for assertions the seed defaults could disturb (none — detail in the test commit message).

- **Directory parse robustness:** `list_relay_agents` deserializes the whole kind:10100 set in one call, so one malformed profile value (e.g. a numeric field) blanks the entire agent directory for every desktop surface. Pre-existing class (`respond_to`, `capabilities` have the same shape — this PR's `Option<String>` is the most tolerant field in the struct); fixed read-side in the follow-up PR (#3866, mergeable in either order). The write-path half — nothing validates kind:10100 content before the relay persists it — belongs with #2987/#3448.
- **`owner_only` directory-only candidates:** an `owner_only` agent is offered to its owner via the user-search candidate (which carries the verified owner); directory candidates carry no verified owner and fail closed. If an owner ever reports their agent missing from add search, the fix is backend owner enrichment of `RelayAgentInfo`.

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -69,6 +69,7 @@ export default defineConfig({
         "**/workflows.spec.ts",
         "**/identity-archive.spec.ts",
         "**/identity-archive-hide.spec.ts",
+        "**/add-search-eligibility.spec.ts",
         "**/relay-connectivity.spec.ts",
         "**/unread-pill.spec.ts",
         "**/sidebar-more-unread-overlap.spec.ts",
--- a/desktop/src-tauri/src/managed_agents/types.rs
+++ b/desktop/src-tauri/src/managed_agents/types.rs
@@ -207,6 +207,11 @@ pub struct RelayAgentInfo {
     pub respond_to: Option<RespondTo>,
     #[serde(default)]
     pub respond_to_allowlist: Vec<String>,
+    /// The agent's kind:10100 `channel_add_policy` declaration
+    /// (`anyone` | `owner_only` | `nobody`). Relay-enforced on
+    /// third-party adds; surfaced so pickers can mirror that enforcement.
+    #[serde(default)]
+    pub channel_add_policy: Option<String>,
 }
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct ManagedAgentRecord {
--- a/desktop/src-tauri/src/nostr_convert.rs
+++ b/desktop/src-tauri/src/nostr_convert.rs
@@ -910,7 +910,7 @@ mod tests {
     fn agents_default_sparse_agent_profiles_for_directory_parse() {
         let e = ev(
             10100,
-            r#"{"channel_add_policy":"owner-only","display_name":"Scout"}"#,
+            r#"{"channel_add_policy":"owner_only","display_name":"Scout"}"#,
             vec![],
         );
         let v = agents_from_events(std::slice::from_ref(&e));
@@ -926,6 +926,7 @@ mod tests {
         assert_eq!(parsed[0].capabilities, Vec::<String>::new());
         assert_eq!(parsed[0].status, "offline");
         assert_eq!(parsed[0].respond_to, None);
+        assert_eq!(parsed[0].channel_add_policy.as_deref(), Some("owner_only"));
     }
 
     #[test]
@@ -941,6 +942,7 @@ mod tests {
             parsed[0].respond_to,
             Some(crate::managed_agents::RespondTo::Anyone)
         );
+        assert_eq!(parsed[0].channel_add_policy, None);
     }
 
     #[test]
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -7,6 +7,7 @@ import {
   getSharedChannelIds,
   isAgentIdentityInManagedList,
   relayAgentIsSharedWithUser,
+  shouldHideAgentFromAddSearch,
   shouldHideAgentFromMentions,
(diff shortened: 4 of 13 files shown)


Chessing234:
this matches the relay policy story. can you add a test for the deny case so we dont accidentally show agents that opted out?

Shipitrealgood:
Good check to insist on. The deny direction was covered at the unit level in `agentAutocompleteEligibility.test.mjs` — `hides nobody-policy agents from everyone including the owner` (the relay refuses those adds regardless of actor, so even the verified owner is denied), `owner_only admits only the verified owner`, and `hides agents with no directory entry or no declared policy` (fail-closed default).

Your comment prompted a look at the e2e layer too, which turned out to be the real gap: the mock bridge seeds carried no `channel_add_policy` at all, so every mock directory agent resolved to null policy and the smoke suite was blind to this feature. Fixed in 6964f91 — the bridge seeds now carry the field, plus paired smoke specs: an agent declaring `nobody` never appears in add search, with an `anyone` control proving the deny comes from the policy gate rather than a broken search. Modeled on the `identity-archive-hide` member-adder specs. ~~Noting for transparency: the new spec's first execution is this PR's CI run — the dev box here has no Playwright browser.~~ Edit: installed a Playwright browser right after posting — the spec now runs green locally too (2/2: deny + control), so CI is confirmation rather than first execution.

Chessing234:
gating add-member search on relay policy is the right UX. worth a disabled tooltip when the policy forbids adds so it doesn't look like a bug.

Shipitrealgood:
The UX concern is fair — a silent absence can read as a bug. We kept `nobody` hidden rather than disabled deliberately, though: that policy is the agent's own declaration to *not be offered*, and a visible-but-disabled row still advertises an identity whose owner opted out — plus it invites exactly the "why can't I add this?" friction the declaration exists to prevent. It's also consistent with how this picker already behaves: existing members, archived identities, and name collisions are all excluded silently, so there's no disabled-row idiom in the surface to match.

Where a tooltip does have merit is the `owner_only` case seen by a non-owner — "only this agent's owner can add it" is informative without overriding anyone's opt-out. Happy to take that as a follow-up if maintainers want the affordance; kept it out of this PR to hold scope.

PR #3865, opened 2026-07-31
fix(compose): allow overriding S3 endpoint

### LINKED ISSUE #3794: fix(compose): BUZZ_S3_ENDPOINT is hardcoded, so .env cannot point media at external S3
**Describe the bug**

`deploy/compose/compose.yml:14` hardcodes the S3 endpoint on the relay service:

```yaml
BUZZ_S3_ENDPOINT: http://minio:9000
```

Every other S3 setting is env-driven and documented in `.env.example` — `BUZZ_S3_ACCESS_KEY`, `BUZZ_S3_SECRET_KEY`, `BUZZ_S3_BUCKET`, `BUZZ_S3_ADDRESSING_STYLE` — so setting `BUZZ_S3_ENDPOINT` there looks like it should work. It is silently ignored, and the relay keeps using the bundled MinIO.

The failure is quiet: the stack comes up, `Media storage connected` is logged, and uploads succeed — into MinIO on local disk rather than the configured bucket. Nothing indicates the endpoint was overridden.

This matters because the Helm chart directs production deployments the other way (`deploy/charts/buzz/values.yaml`):

> Production: point endpoint/bucket at an external S3-compatible service and supply credentials. Quickstart (`minio.enabled: true`): the chart runs an in-cluster, eval-only MinIO.

So compose users following the same guidance get local MinIO regardless of what they configure.

**Steps to reproduce**

1. Set `BUZZ_S3_ENDPOINT=https://<account>.r2.cloudflarestorage.com` in `deploy/compose/.env`.
2. `docker compose up -d`.
3. `docker exec <relay> printenv BUZZ_S3_ENDPOINT` → `http://minio:9000`.
4. Uploads land in the bundled MinIO volume, not the configured bucket.

**Expected behavior**

`BUZZ_S3_ENDPOINT` honours `.env` like the other `BUZZ_S3_*` values, defaulting to the bundled MinIO when unset:

```yaml
BUZZ_S3_ENDPOINT: ${BUZZ_S3_ENDPOINT:-http://minio:9000}
```

Ideally with the variable added to `.env.example` alongside the other S3 settings.

**Version and platform**

- Buzz version: relay `ghcr.io/block/buzz:sha-63496cc`; also present on `main` as of 2026-07-30
- OS: Ubuntu 24.04, Docker Compose

**Logs / additional context**

Workaround is a compose override that also has to reset the MinIO health dependency, since the relay `depends_on` it:

```yaml
services:
  relay:
    environment:
      BUZZ_S3_ENDPOINT: ${BUZZ_S3_ENDPOINT:?}
    depends_on: !reset
      postgres: {condition: service_healthy}
      redis: {condition: service_healthy}
  minio: {profiles: ["disabled"]}
  minio-init: {profiles: ["disabled"]}
```

Possibly related to #3273 (hardcoded path-style S3 addressing), which is the same class of compose-level S3 config being fixed rather than configurable.

pollychen-lab:
## Summary

- Honor `BUZZ_S3_ENDPOINT` from compose `.env`, defaulting to bundled MinIO.
- Honor `BUZZ_S3_ADDRESSING_STYLE` from `.env`, defaulting to `path` for bundled MinIO.
- Document both variables in the compose example and README.

Fixes #3794

## Validation

- `git diff --check`
- `docker compose --env-file .env config` with `.env` copied from `.env.example` renders `BUZZ_S3_ENDPOINT: http://minio:9000` and `BUZZ_S3_ADDRESSING_STYLE: path`.
- `BUZZ_S3_ENDPOINT=https://example.r2.cloudflarestorage.com BUZZ_S3_ADDRESSING_STYLE=virtual docker compose --env-file .env config` renders the external endpoint and `virtual` addressing.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -87,7 +87,8 @@ RELAY_URL=ws://localhost:3000
 # -----------------------------------------------------------------------------
 # The local MinIO container is reachable from host processes at localhost:9000.
 # Path style keeps the bucket in the URL path and is required by this local DNS
-# setup. Use `virtual` only when the provider requires bucket-as-subdomain URLs.
+# setup. Use `virtual` only when the provider requires bucket-as-subdomain URLs
+# (for example, real S3 endpoints rather than local MinIO).
 BUZZ_S3_ENDPOINT=http://localhost:9000
 BUZZ_S3_ACCESS_KEY=buzz_dev
 BUZZ_S3_SECRET_KEY=buzz_dev_secret
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ A Rust workspace of focused crates. Single source of truth: the relay. See [ARCH
 <summary><strong>Configuration</strong> (env vars, defaults work for local dev)</summary>
 
 All defaults work out of the box. Override via `.env`. Full reference in [`.env.example`](.env.example).
+For local MinIO, keep `BUZZ_S3_ADDRESSING_STYLE=path`; real S3-style providers may require `virtual` bucket-subdomain addressing instead.
 
 </details>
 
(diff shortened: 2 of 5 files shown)


Chessing234:
looks right for #3794. tiny nit: call out in the readme that path vs virtual still matters for minio vs real s3 so people dont copy the default blindly

Chessing234:
s3 endpoint override was overdue. worth a one-line note in .env.example so people know the compose hardcode is gone.

pollychen-lab:
@Chessing234 thank you for checking

PR #3866, opened 2026-07-31
fix(desktop): parse the relay agent directory leniently

Shipitrealgood:
## Problem

`list_relay_agents` deserialized the whole kind:10100 directory in a single `serde_json::from_value` call, so one profile with a mistyped field — e.g. a numeric `respond_to` or `channel_add_policy` — failed the entire query. Kind:10100 content is free-form JSON under each agent author's control, and the relay persists the event even when its own side-effect validation fails (`handlers/ingest.rs` logs and keeps it), so any authenticated publisher could blank the agent directory for every desktop client. The failure is app-wide and looks like a product bug: `useRelayAgentsQuery` mounts on ~14 always-live surfaces (mentions, members bar, add-member search, profile popovers, search, pulse).

## What this changes

Parse per-entry instead: malformed profiles are logged (`tracing::warn!` with the offending pubkey) and skipped; the rest of the directory survives. Same pattern as `custom_harnesses.rs` (per-file warn-and-skip). Output is unchanged for any directory that parses today; error results now come only from the relay query itself. Verified against all `useRelayAgentsQuery` consumers — the two that touch the error path use it only as a readiness signal and a dialog message, so fewer errors is strictly safe.

## Testing

Unit tests on the extracted `parse_relay_agents` helper: valid entries parse; a poisoned entry (`"respond_to": 123`) is skipped while its neighbors survive; non-array input returns empty.

Executed locally via `just desktop-tauri-test`: full suite 2,017 passed / 0 failed (14 ignored), including the three `parse_relay_agents` tests.

## Related

The write-path half of this root cause — nothing validates kind:10100 content shape before it is persisted — is tracked in the external-agent threads (#2987; #3448 fixes the CLI's clobbering writer). This is the read-side defense. Surfaced during review of #3864 (which adds one more field riding this parse); mergeable in either order.

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -1076,6 +1076,7 @@ dependencies = [
  "security-framework 3.7.0",
  "serde",
  "serde_json",
+ "serde_path_to_error",
  "serde_yaml",
  "sha2 0.11.0",
  "sherpa-onnx",
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -84,6 +84,9 @@ opus = "0.3"
 neteq = { version = "0.8", default-features = false }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
+# Names the failing field path in directory-parse skip warns (already in the
+# tree as a transitive dep).
+serde_path_to_error = "0.1"
 serde_yaml = "0.9"
 toml = "0.8"
 nostr = { version = "0.44", features = ["nip44", "nip49"] }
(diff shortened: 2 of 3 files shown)


Chessing234:
good call skipping bad entries. maybe also log a count of skipped vs kept so ops can spot a poisoned directory without digging warn spam?

Shipitrealgood:
Done in 83b8572 — when anything is dropped, one aggregate line per query (`kept N of M agent profiles, skipped K unparseable`) alongside the per-entry warns, which keep the pubkey specifics. No log noise on the healthy path: the aggregate line only fires when `skipped > 0`.

Chessing234:
lenient agent directory parsing helps with partial relay payloads. maybe log which fields were dropped so operators can spot schema drift.

Shipitrealgood:
Small clarification first: whole *entries* are skipped, not fields — an unparseable profile drops out and its neighbors survive. But your underlying ask (spot schema drift from the warn without digging) exposed a real weakness: serde's error message didn't name the failing field. Done in f27b2e0 — deserialization now routes through `serde_path_to_error` (already in the tree as a transitive dep), so the per-entry warn reads e.g. `(...pubkey...): channel_add_policy: invalid type: integer` — field path, offending type, and which agent, in one line. Verified locally: full `desktop-tauri-test` suite 2,017/2,017 passing on the branch.

PR #1498, opened 2026-07-03
fix(buzz-acp): clear stale 👀 on events delivered via native steer

tlongwell-block:
## Problem

Since goose-native steering landed (#1160), messages absorbed into an in-flight turn via `SteerAck::Success` keep their 👀 reaction forever. Reported by Jem in #buzz-bugs: switched channels mid-turn, came back, eyes still on the message with no sign of activity.

## Root cause

The 👀 is added fire-and-forget at queue-push time. Cleanup is owned by `run_prompt_task`'s `ReactionGuard`, which collects event ids **from the FlushBatch**. But a successfully steered event is consumed by `queue.remove_event()` in the `SteerAck::Success` arm and never enters any batch — so no guard ever owns its id, and the 👀 is never removed. Every successfully steered message leaks its eyes.

## Fix

Contract: **eyes go away when the agent turn stops.** One cleanup helper (`spawn_steered_eyes_cleanup`) is shared by every path.

- `TaskMeta.steered_event_ids`: the `SteerAck::Success` arm attaches the consumed event id to the in-flight turn's meta.
- **Exact-turn binding** (review, Wren): `pool.send_steer` returns the turn's `tokio::task::Id`; the ack watcher carries it, and `record_steered_event` matches on the exact task id — a delayed ack for an ended turn can never bind to a successor turn on the same channel. Stale id → immediate 👀 removal instead.
- Turn end — normal completion (`handle_prompt_result`) or panic (`recover_panicked_agent`) — drains the ids and removes the 👀.
- **Graceful shutdown** (review, Wren): the post-loop drain bypasses `handle_prompt_result`, so `tokio_main` now drains all steered ids from every in-flight `TaskMeta` (`drain_all_steered_event_ids`) as the loop exits, and — after the grace period settles every prompt task — drops its `steer_ack_tx` and drains `steer_ack_rx` (`drain_crossing_steer_acks`) to recover Success acks whose watchers resolved after the loop stopped polling. Both cleanups are bound-awaited before exit; a 2 s per-recv timeout backstops a wedged watcher.
- Known limitation (pre-existing, documented in fn docs): non-Success acks crossing shutdown and queued-but-never-dispatched events share the same stale-👀-at-exit fate; relay-side cleanup owns that class.

## Testing

- `pool::tests`: exact-id binding, false-when-gone, the generation-reuse race (turn A ends → turn B same channel → A's late ack must not bind to B), shutdown drain take-once semantics.
[...]

PR #3870, opened 2026-07-31
feat(buzz-acp): use cat pickup reaction

Greg-Moskalenko:
## Summary

- replace the two-phase 👀 / 💬 pickup indicator with one visible 🐱 reaction
- clear the cat plus legacy pickup reactions on completion, failure, panic, membership removal, queue eviction, native steering, and shutdown paths
- bind native-steer cleanup to the exact accepting turn so a late acknowledgement cannot attach cleanup to a successor turn

## Verification

- `cargo test -p buzz-acp` — 658 unit tests and 9 lifecycle integration tests passed
- `cargo clippy -p buzz-acp --all-targets --all-features -- -D warnings`
- `git diff --check`

The exact-turn native-steer lifecycle handling adapts and credits the prior work in #1498.

Originating Buzz channel: `53e4b720-c324-4c27-b26c-aaaee778c857`

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -41,7 +41,7 @@ use pool::{
     PromptResult, PromptSource, SessionState, TimeoutKind,
 };
 use pool_lifecycle::PoolLifecycle;
-use queue::{CancelReason, EventQueue, FlushBatch, QueuedEvent, ThreadTags};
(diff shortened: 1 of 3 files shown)


Chessing234:
cute. can you keep the old eye/speech cleanup for a release or two so agents mid-upgrade dont leave stale pickup reactions behind?

Chessing234:
also: if the cat reaction fails to publish, does pickup still clear? might want a best-effort clear even when the reaction post errors

PR #3872, opened 2026-07-31
feat(desktop): dictate messages from the composer toolbar

Finnstwin:
## What

Adds a mic button to the composer toolbar so a message can be **spoken** instead of typed. Today the composer is text-only.

Transcription runs **entirely on-device**. No API key, no network call, no new dependency, nothing leaves the machine.

## Why this is small

Buzz already ships a complete local STT stack for huddle transcription — sherpa-onnx Parakeet TDT-CTC 110M, with background model download, rubato resampling and earshot VAD (`huddle::stt`, `huddle::models`).

This reuses it wholesale and changes only the **sink**: huddle publishes kind:9 events to the relay, dictation emits the text back to the webview and inserts it at the caret.

## Design notes

**Backend** — `commands/dictation.rs` exposes `start_dictation` / `stop_dictation` / `push_dictation_pcm` / `is_dictation_available`, driving its own `SttPipeline` instance.

- Registered as its own Tauri managed state rather than an `AppState` field. Dictation shares no data with the rest of the app, and dictating a message must not require an active huddle — the two pipelines can run concurrently, each with its own audio feed command.
- Pipelines are dropped *outside* the state lock, since `SttPipeline::drop` joins the worker thread (~200 ms).

**Frontend**

- `setupAudioWorklet` gains an optional target command and a PTT opt-out, so dictation reuses the huddle audio path instead of duplicating it. Both default to today's behaviour, so existing callers are unaffected.
- The button lives in the shared `MessageComposerToolbar`, so `MessageComposer` and `ForumComposer` both gain dictation with **no changes of their own**.
- Transcripts insert at the caret and compose with typed text rather than replacing the draft.

Shipped behind the `dictation` preview flag.

## Tests

- `dictationInsert.test.mjs` — the caret-insertion spacing rule (empty composer, mid-sentence, existing whitespace, punctuation).
- `dictation_tests.rs` — a hardware-gated (`#[ignore]`) acceptance test that feeds real speech through the real model and asserts a transcript comes back. The continuous-VAD configuration (`tts_active` pinned false, no `ptt_active`) is invisible at compile time and would fail *silently* if wrong, so it is worth pinning:

```
cargo test -p buzz-desktop dictation_pipeline_transcribes_speech -- --ignored --nocapture
dictation transcript: "I love you"
test commands::dictation::tests::dictation_pipeline_transcribes_speech ... ok
```

Full suites green locally: `desktop-check`, `desktop-typecheck`, `desktop-tauri-fmt-check`, `desktop-tauri-clippy`, `desktop-tauri-test` (2003 passed), `desktop-test` (3877 passed).

## Notes for reviewers

- `NSMicrophoneUsageDescription` already exists in `Info.plist` for huddles, so **no new entitlement** is introduced.
- The Web Speech API was considered and rejected: `webkitSpeechRecognition` exists in WKWebView but `.start()` returns `service-not-allowed`, and WebKitGTK has no implementation at all — so it cannot work cross-platform for a Tauri app.
- **Not yet verified by running the packaged app end-to-end** — the inference path is proven by the acceptance test above, but a maintainer clicking the button in a real build is still worth doing.


Chessing234:
this looks useful. maybe hide or disable the mic button while a huddle is already capturing audio so the two stt paths dont fight?

Chessing234:
dictate from the composer toolbar is a nice touch. any offline/permission denied path that surfaces a clear toast?

PR #3873, opened 2026-07-31
fix(mobile): queue publishes through reconnect

bluepinkneon:
## Summary
Queue required EVENT publishes until the replacement WebSocket has authenticated instead of silently dropping a send during connection handoff. Keep config/auth changes inside one relay-session lifecycle so an in-flight pre-send event survives the socket replacement, while sent-but-unacknowledged events still fail explicitly on disconnect.

### Related issue
No upstream duplicate found. Local operational bug: source Brief #16705.

### Testing
- bin/just mobile-check
- flutter test test/shared/relay/relay_session_test.dart test/features/channels/send_message_provider_test.dart (17 passed)
- Regression replaces the first socket, publishes during the handoff, authenticates the second socket, verifies one EVENT frame, and resolves its OK acknowledgment.

Real iPhone proof remains pending an installable build containing this commit.

### DIFF
--- a/mobile/lib/shared/relay/relay_session.dart
+++ b/mobile/lib/shared/relay/relay_session.dart
@@ -56,6 +56,13 @@ class _PendingEvent {
   _PendingEvent({required this.completer, required this.timeout});
 }
 
+class _QueuedEvent {
+  final NostrEvent event;
+  final _PendingEvent pending;
+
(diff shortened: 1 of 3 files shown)


Chessing234:
nice fix — one small ask: can you add a test where the socket dies mid-queue so we know the buffered events still flush after auth, not just get stuck?

PR #3875, opened 2026-07-31
fix(compose): wire buzz-pair-relay so mobile QR pairing works

### LINKED ISSUE #3842: docker-compose deploy has no wiring for the buzz-pair-relay sidecar — mobile QR pairing 404s out of the box
**Title:** docker-compose deploy has no wiring for the `buzz-pair-relay` sidecar — mobile QR pairing 404s out of the box

**Body:**

Mobile QR pairing failed with `WebSocket connection failed: HTTP error: 404 Not Found` on a self-hosted relay deployed via `deploy/compose/` (plain Docker Compose, not the Helm chart).

Root cause: the relay's NIP-11 doc advertised NIP-43 support (membership-enforcing relay, correctly per `crates/buzz-relay/src/nip11.rs`'s own test suite intent), which makes the desktop client (`desktop/src-tauri/src/commands/pairing.rs::pairing_relay_from_nip11`) infer a `/pair` legacy-path sidecar — but `deploy/compose/` never stands one up. There's no `/pair` route anywhere in the stack (bare `reverse_proxy relay:3000` catch-all in the example Caddyfile), and no `buzz-pair-relay` service in `deploy/compose/compose.yml`. The relay's own test comment (`nip11.rs`) even predicts this exact failure mode: *"advertising it on open relays misroutes pairing peers to a non-existent /pair sidecar."*

By contrast, `deploy/charts/buzz/templates/pairing-relay.yaml` (Helm) does wire this up correctly — but it's `pairingRelay.enabled: false` by default there too, so it's easy to end up with a relay that advertises NIP-43 without a working pairing sidecar on **either** deploy path.

We worked around it by hand for our own compose deployment:
1. Added a `pairing-relay` service to `compose.yml` using the same `${BUZZ_IMAGE}`, with `entrypoint: ["/usr/local/bin/buzz-pair-relay"]` (worth noting: Compose's `command:` only appends args to the image's baked-in `ENTRYPOINT ["/usr/local/bin/buzz-relay"]`, unlike Kubernetes where `command:` replaces it — so copying the Helm chart's `command:` pattern into Compose silently runs the wrong binary; needed `entrypoint:` instead)
2. `handle_path /pair* { reverse_proxy pairing-relay:5000 }` in the Caddyfile
3. `BUZZ_PAIRING_RELAY_URL=wss://<domain>/pair` on the `relay` service, so NIP-11 advertises `pairing_relay_url` directly instead of relying on the NIP-43 inference path

**Ask:** add the pairing-relay service + Caddyfile route to `deploy/compose/compose.yml`/`Caddyfile` (or the example files), matching what the Helm chart already does, so a membership-enforcing compose deployment doesn't advertise a pairing capability it can't serve.

**Secondary observation, lower confidence:** the first pairing attempt after standing up the sidecar failed with "Lost connection to pairing relay" in the desktop app, even though both devices displayed and matched the same SAS code — it worked cleanly on retry. `handle_nip42_auth` (`pairing.rs`) waits up to 3s for an AUTH challenge from the relay before subscribing; `buzz-pair-relay` never sends one (it's a bare relay, no NIP-42), so both sides burn a fixed ~3s stall before their `REQ` subscription goes out. Against an ephemeral, non-buffered relay (no event replay), that's a plausible race window if one side's event arrives before the other side's subscription is live. Not confirmed as the root cause, but worth a look if this is reproducible elsewhere — happy to provide sidecar logs from the one reproduction we have.

Chessing234:
## Summary
- Stand up the `buzz-pair-relay` sidecar in `deploy/compose` (using `entrypoint`, not `command`, so Compose does not append args to `buzz-relay`).
- Advertise `BUZZ_PAIRING_RELAY_URL` in NIP-11 and proxy `/pair*` through Caddy so membership-enforcing self-hosts stop 404ing on legacy `/pair`.
- Document TLS vs direct-port URL shapes in `.env.example`, README, and `run.sh help`.

Fixes #3842

## Test plan
- [ ] `cd deploy/compose && cp .env.example .env` (fill CHANGE_ME) and `./run.sh config` succeeds with `BUZZ_PAIRING_RELAY_URL` set
- [ ] `BUZZ_COMPOSE_TLS=true ./run.sh start` brings up `pairing-relay` healthy; Caddy `/pair` reaches it
- [ ] Desktop/mobile QR pairing completes against the compose stack without `HTTP 404`
- [ ] Without Caddy, pairing still works when `BUZZ_PAIRING_RELAY_URL` points at the published `:5000` port

### DIFF
--- a/deploy/compose/.env.example
+++ b/deploy/compose/.env.example
@@ -8,6 +8,10 @@ BUZZ_IMAGE=ghcr.io/block/buzz:main
 # Public host name. Used by compose.caddy.yml and URL-derived settings below.
 BUZZ_DOMAIN=buzz.example.com
 RELAY_URL=wss://buzz.example.com
+# Dedicated NIP-AB pairing URL advertised in NIP-11. With compose.caddy.yml this
+# is same-host /pair (proxied to pairing-relay). Without Caddy, point at the
+# published pairing port instead, e.g. ws://buzz.example.com:5000.
+BUZZ_PAIRING_RELAY_URL=wss://buzz.example.com/pair
 BUZZ_MEDIA_BASE_URL=https://buzz.example.com/media
 BUZZ_MEDIA_SERVER_DOMAIN=buzz.example.com
 BUZZ_CORS_ORIGINS=https://buzz.example.com
@@ -38,6 +42,8 @@ BUZZ_S3_ADDRESSING_STYLE=path
 
 # Optional host ports. Base compose publishes the relay directly on BUZZ_HTTP_PORT.
 BUZZ_HTTP_PORT=3000
+# Pairing sidecar host port (non-Caddy). Caddy TLS installs use /pair on 443 instead.
+BUZZ_PAIR_RELAY_PORT=5000
 
 # Caddy host ports. Only used with compose.caddy.yml.
 CADDY_HTTP_PORT=80
--- a/deploy/compose/Caddyfile
+++ b/deploy/compose/Caddyfile
@@ -1,5 +1,11 @@
 {$BUZZ_DOMAIN} {
   encode zstd gzip
 
+  # Device-pairing sidecar (buzz-pair-relay). Must be before the catch-all so
+  # NIP-AB QR sessions do not hit the main relay and 404 (#3842).
+  handle_path /pair* {
+    reverse_proxy pairing-relay:5000
+  }
+
   reverse_proxy relay:3000
 }
--- a/deploy/compose/README.md
+++ b/deploy/compose/README.md
@@ -38,6 +38,13 @@ keypair.
   migrations.
 - The stack uses Postgres, Redis, MinIO, and a git data volume because
   those are real Buzz dependencies today. Minimal mode can simplify this later.
+- Mobile QR pairing needs the `pairing-relay` sidecar. Compose stands it up on
+  port 5000 and advertises `BUZZ_PAIRING_RELAY_URL` in NIP-11. With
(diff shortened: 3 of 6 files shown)


Chessing234:

Tony-ooo:
Confirming this from a production self-host: I've been running the `buzz-pair-relay` sidecar on `ghcr.io/block/buzz:main` since Aug 4, and the shape this PR proposes is exactly what works in practice.

Verified working snippet:

```yaml
pair:
  image: ${BUZZ_IMAGE:-ghcr.io/block/buzz:main}
  entrypoint: ["/usr/local/bin/buzz-pair-relay"]
  environment:
    BUZZ_PAIR_RELAY_BIND_ADDR: 0.0.0.0:5000
  ports:
    - "5000:5000"
  restart: unless-stopped
  networks:
    - buzz-net
```

Notes that line up with this PR:

- `entrypoint` (not `command`) is the right call — with `command`, Compose appends args to the `buzz-relay` binary and the sidecar never behaves as a pairing relay.
- The stock compose Caddyfile only routes to `relay:3000`, so `/pair*` 404s unless the sidecar is on a direct port or Caddy adds the `handle_path /pair*` route. The Caddyfile change here closes exactly that gap.
- Set `BUZZ_PAIRING_RELAY_URL=wss://<domain>/pair` in `.env` so NIP-11 advertises the pairing endpoint.

Thanks for landing this — it makes compose self-hosts work out of the box for mobile pairing.

PR #3876, opened 2026-07-31
fix(acp): go offline when Claude/Codex auth expires

### LINKED ISSUE #3831: Agent stays "online" after its Claude credentials expire — every mention fails with -32000, nothing surfaces it
I'm running three agents headless in Docker on a VPS (buzz 0.5.0 binaries from the Linux .deb, `buzz-acp` spawning `claude-agent-acp` 0.63.0, hosted relay). Two of them use Claude Code with a subscription login. Both died about a day after I signed them in, and I had no way of knowing.

Bumble was signed in around 01:10 UTC on 29 July. First failure 06:31 on 30 July — roughly 29 hours later. Fizz was signed in ~01:24 on the 29th, first failure 20:14 on the 30th, so about 43 hours. Since then every single mention fails:

```
ERROR pool::prompt: session_prompt error: Agent reported error (code -32000): Authentication required
WARN buzz_acp: agent_returned (application error — pipe intact) agent=0 outcome="error"
```

Running `claude auth status` inside the container confirms it:

```json
{"loggedIn": false, "authMethod": "none", "apiProvider": "firstParty"}
```

The credentials went stale and there's no browser in the container to complete a refresh, which I understand. That part is my problem for running it headless.

The part I think is a bug is that nothing shows this. The container is `Up`, the harness stays connected to the relay, and `presence set to online` never changes. In the client the agent looks completely normal. Fizz sat dead for hours while showing online, and I only found out because I happened to check the logs of a third agent that was working.

Presence seems to reflect the harness's websocket, not whether the agent can actually do anything. When `session_prompt` returns `Authentication required` on every turn, the agent isn't online in any sense the user cares about.

A few things would each have saved me the confusion, in rough order of how useful they'd have been:

- Presence going away or changing after N consecutive `outcome="error"` turns
- The harness posting something into the channel when it can't authenticate, so the failure appears where you're already looking
- A non-zero exit or an unhealthy signal so Docker/systemd notices, rather than staying up in a broken state

Worth noting the third agent, running `goose` against an API key instead of a subscription login, has zero auth errors over the same period. So this is specific to credential types that expire and need an interactive refresh.

Possibly related: #2265 — same shape (Claude agent silently stops working), different cause.

Chessing234:
## Summary
- Classify headless `Authentication required` (-32000) as a non-retryable auth error alongside `Re-authenticate` / `API Error: 401`.
- Also match the macOS managed-agent `OAuth session expired and could not be refreshed` wording.
- On auth expiry: dead-letter the turn, post a restart-needed notice, stop presence heartbeats, and publish presence `offline` so the agent stops looking live in the sidebar.
- Document the recovery path in `buzz-acp` README.

Fixes #3831

## Test plan
- [x] `cargo test -p buzz-acp --lib auth_error`
- [ ] Run a managed Claude agent, force `claude auth logout` (or expire creds), @mention it — expect channel notice + sidebar goes offline, no further online heartbeats
- [ ] Re-login + restart harness — presence returns to online and mentions work again

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -335,6 +335,22 @@ Set `BUZZ_ACP_AGENT_COMMAND` and `BUZZ_ACP_AGENT_ARGS` to point at your agent bi
 
 See the [root TESTING.md](../../TESTING.md) for the full integration testing guide — automated test suites, multi-agent E2E testing via the ACP harness, and troubleshooting.
 
+## Troubleshooting
+
+### Agent stays "online" but every mention fails with `-32000`
+
+Claude / Codex subscription credentials can expire while the harness process is
+still up. When the agent returns a non-retryable auth error (`Authentication
+required`, `Re-authenticate`, or `API Error: 401`), buzz-acp:
+
+1. Dead-letters the failed turn and posts a channel notice asking you to re-login
+2. Stops renewing presence heartbeats
+3. Publishes presence `offline` so the sidebar stops looking live
+
+Re-authenticate the CLI (`claude /login` / `codex login`) and **restart**
+`buzz-acp` — presence does not automatically flip back to online inside the
+same process after auth death.
+
 ## License
 
 Apache-2.0
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -6267,6 +6267,18 @@ mod error_outcome_emission_tests {
     }
 
     #[test]
+    fn is_auth_error_matches_authentication_required_message() {
+        let e = acp::AcpError::AgentError {
+            code: -32000,
+            message: "Authentication required".to_string(),
+        };
+        assert!(
+            is_auth_error(&e),
+            "headless 'Authentication required' must be classified as auth error"
+        );
+    }
+
+    #[test]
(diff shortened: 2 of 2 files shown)


Chessing234:

HashMac:
Independent macOS managed-agent reproduction. The failure surfaced as:

```text
Agent reported error (code -32000): Failed to authenticate: OAuth session expired and could not be refreshed
```

The current patch matches `Re-authenticate`, `API Error: 401`, and `Authentication required`, so this exact provider wording still appears to miss `is_auth_error()` and enter the generic requeue/backoff path. I observed repeated retries while the runtime continued to look available; the channel had no actionable reauthentication state.

Reauthenticating the Claude subscription and restarting the managed runtime restored turns.

Could this PR add a precise match and fail-first test for `OAuth session expired and could not be refreshed` (ideally constrained to `AgentError` / code `-32000`)? Suggested live acceptance:

1. Expire subscription OAuth and mention the agent.
2. One visible reauthentication notice; presence becomes offline/not-ready; no retry storm.
3. Reauthenticate and restart.
4. Presence returns online and one fresh mention produces exactly one turn.

This is the same user-facing failure as #3831, just a fourth message variant.

Issue #2859, opened 2026-07-25
feat(spawner): first-party server-hosted agents via a buzz-spawner daemon


siddhartpai:
## Problem

Buzz agents can only be spawned by the desktop app.
`desktop/src-tauri/src/managed_agents/runtime.rs::spawn_agent_child` generates
the keypair, computes the NIP-OA owner attestation, and launches the `buzz-acp`
harness locally. Fizz/Honey/Bumble aren't special processes — they're built-in
personas (`managed_agents/personas.rs`) minted through that same local path.

This means:

- **Agents die with the laptop.** There is no always-on agent.
- **Agent creation requires the desktop app.** Mobile and web can't create one.
- **Self-hosters have no server-side option.** You can host a relay, but not an
  agent. The only server-side path is `BackendKind::Provider` → an external
  `buzz-backend-<id>` binary (`managed_agents/backend.rs:20-120`), which has no
  OSS implementation and still puts the desktop in the control loop.
- **The relay is entirely agent-agnostic** — no spawner, no agent registry, no
  lifecycle hooks. Agents are just pubkeys admitted via NIP-OA owner attestation
  (`crates/buzz-relay/src/api/mod.rs:40-200`).

## Proposal

Add `crates/buzz-spawner` — a standalone daemon that watches the relay for
agent-definition events and reconciles them into Docker-isolated `buzz-acp`
containers. It ships as a second service in `deploy/compose`.

It is deliberately **not** a relay module: the relay's only subprocess today is
`git`, and it goes out of its way to disable repo hooks
(`api/git/transport.rs:930-935`). Adding container execution to the public
WebSocket server isn't acceptable, and isn't necessary. `buzz-pair-relay` and
`buzz-push-gateway` are existing precedent for sidecar service crates.

The daemon is a plain relay client with its own Nostr identity. It never touches
Postgres.

```
owner client ──kind:30178 (spec)──►  relay  ──sub──►  buzz-spawner
     ▲                                                     │
     └──kind:24201 attestation handshake (NIP-44)──────────┤
                                                           ▼
                                              Docker: buzz-acp container
                                                           │
                                                    (agent's own key)
                                                           ▼
                                                         relay
```

## Design points

**Three new event kinds** (`buzz-core/src/kind.rs`), following the existing
30175/30176/30177 agent block:

| Kind | Purpose |
|---|---|
| `30178` `SPAWNER_AGENT_SPEC` | Desired state, owner-authored. `d` = spec slug |
[...]

PR #3879, opened 2026-07-31
managed_agents: add cloud agent ACP runtime preset

yshuolu:
## What

Adds one entry to `KNOWN_ACP_RUNTIMES`: **AgentSky** (`agentsky-acp`, `npm install -g @agentsky/acp`) — an ACP agent that proxies to a hosted agent session (Claude Code, Codex, Hermes, OpenClaw) at [agentsky.dev](https://agentsky.dev).

## What it unlocks for buzz users

- **Always-on agent without user VM/laptop setup** — nothing to provision or maintain: no VM, no systemd unit, no local harness install
- **Support all popular agent harness** — the hosted session can run Claude Code, Codex, Hermes, or OpenClaw.

## Notes

- Discoverability only — same shape as the existing Claude Code / Codex preset entries; everything already works via `BUZZ_ACP_AGENT_COMMAND=agentsky-acp` or a custom-harness JSON.
- Buzz's surface is unchanged: ACP over stdio in, replies via `buzz messages send` over the relay. HTTPS is only the adapter's private transport to its backend, same as the Claude Code / Codex adapters calling their vendors' APIs.
- Entry fields match the "external adapter, no local CLI" case: `underlying_cli: None`, no auth probe (env token), `provider_locked: true`.

Disclosure: I build the service behind this adapter. Happy to adjust or move the entry to whatever mechanism you'd prefer if presets are meant to stay first-party.

### DIFF
--- a/desktop/src-tauri/src/managed_agents/discovery.rs
+++ b/desktop/src-tauri/src/managed_agents/discovery.rs
@@ -22,6 +22,7 @@ const CLAUDE_CODE_AVATAR_URL: &str = "https://anthropic.gallerycdn.vsassets.io/e
 const CODEX_AVATAR_URL: &str = "https://openai.gallerycdn.vsassets.io/extensions/openai/chatgpt/26.5313.41514/1773706730621/Microsoft.VisualStudio.Services.Icons.Default";
 const BUZZ_AGENT_AVATAR_URL: &str =
     "https://raw.githubusercontent.com/block/buzz/refs/heads/main/crates/buzz-agent/buzz-agent.png";
+const AGENTSKY_AVATAR_URL: &str = "https://agentsky.dev/asteroids-mark-2.png";
 fn common_binary_paths() -> &'static [PathBuf] {
     static PATHS: OnceLock<Vec<PathBuf>> = OnceLock::new();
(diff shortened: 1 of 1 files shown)


Shipitrealgood:
Hey @yshuolu wanted to comment here real quick.  #2987 is not a side effect of an agent that can't stay on.  It is a regression bug that is triggered by properly registering an agent outside of the native gui and off machine.  When this happens mentions and channel visibility to tag the agent are currently affected.

Not as familiar with #2859  and #3277  but the ability to host an agent or multiple agents and harnesses and register them on machines that are not running within the gui and have them be long running is fully accessible today.

I recommend looking at the docs, systemd setup with .env and registering your harnesses or your agents through there, and the mentions and channel bug should be cleared pending reviews.

This will give you the ability for an always on agent that is on the relay accessible through mobile, whether your laptop running the gui is on or open or not.

Beyond this the project appears to prefer nostr and relay as protocols over https.  The more I have looked at this the more it makes sense, and plan to raise an issue for maintainer guidance on ideas around the vm or multiple vms actually becoming registered in the relay so they are selectable and swappable much like the harness and agents are with the native agents.

PR #3881, opened 2026-07-31
fix(acp): classify 'Authentication required' as non-retryable auth error (#3831)

### LINKED ISSUE #3831: Agent stays "online" after its Claude credentials expire — every mention fails with -32000, nothing surfaces it

iroiro147:
Fixes #3831

## Problem
Expired Claude subscription credentials surface from `claude-agent-acp` as a bare AgentError:

    Agent reported error (code -32000): Authentication required

`is_auth_error()` only matched the two variants PR #2751 added (`Re-authenticate`, `API Error: 401`), so this third variant fell into the generic `queue.requeue()` path. Every mention burned ~10 retry slots over a long backoff before dead-lettering — while presence kept the agent `online` and nothing told the user the agent was effectively dead. The reporter (#3831) had a container `Up`, socket connected, and no signal for hours after credential expiry.

## Root cause
`crates/buzz-acp/src/lib.rs` `is_auth_error()` pattern-match missed the subprocess's bare-credential message.

## Fix (narrow, +1 arm)
Add `message.contains("Authentication required")` alongside the two existing arms in `is_auth_error()`. Now the expired-credential case hits the existing immediate dead-letter + in-channel re-auth notice path (`post_failure_notice`), so each failed mention surfaces where the user is already looking.

Precision bar preserved (per the fn's doc-comment): the matched token is a subprocess **credential hand**, not a generic failure string. Adjacent tests prove no over-classification — transient/transport (`Io`, `WriteTimeout`) and non-auth `AgentError` (usage-credit) variants remain `false`.

## Not in scope (deliberately)
- Presence/away-after-N-errors (reporter option A) — separate, larger policy change.
- Non-zero exit / unhealthy signal to the container runtime (option C) — separate.

Both are listed as follow-ups on #3831; this PR closes the silent-failure root for the documented message variant.

## Test
- `is_auth_error_matches_authentication_required_message` — **fail-first** (red on `origin/main` before the fix, green after).
- Existing `is_auth_error_*` (reauthenticate / 401 / reject-non-auth / reject-transport) all pass.
- Full `cargo test -p buzz-acp --lib` green (650/650 serially; 2 keepalive/idle timers are pre-existing flakes under parallel load, unrelated — pass in isolation on clean `origin/main`).
- `cargo clippy -p buzz-acp --lib` clean.

DCO `--signoff` present. No changeset (repo FILE-SIZE/no-changeset convention).

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -2366,6 +2366,7 @@ async fn tokio_main() -> Result<()> {
                     &mut respawn_tasks,
                     observer.clone(),
                     Some(&ctx.rest_client),
+                    Some((&presence_publisher, &presence_keys)),
                 ) == LoopAction::Exit
                 {
                     break;
@@ -3066,6 +3067,7 @@ fn handle_prompt_result(
     respawn_tasks: &mut tokio::task::JoinSet<()>,
     observer: Option<observer::ObserverHandle>,
     rest_client: Option<&relay::RestClient>,
+    presence: Option<(&relay::RelayEventPublisher, &nostr::Keys)>,
 ) -> LoopAction {
     let before = pool.task_map().len();
     let agent_index = result.agent.index;
@@ -3165,6 +3167,31 @@ fn handle_prompt_result(
                     and then re-send."
                     .to_string();
                 spawn_failure_notice(rest_client, &batch, content);
+                // Auth tokens don't self-repair between heartbeats, and without this the
+                // presence heartbeat keeps publishing "online" while every subsequent
+                // mention dead-letters — the UI keeps showing the agent as available during
+                // a guaranteed outage. Flip presence to offline (best-effort, non-blocking)
+                // so the UI reflects the real state instead. Mirrors the shutdown flip.
+                if let Some((publisher, keys)) = presence {
+                    if config.presence_enabled {
+                        let publisher = publisher.clone();
+                        let keys = keys.clone();
+                        tokio::spawn(async move {
+                            match tokio::time::timeout(
+                                Duration::from_secs(2),
+                                publish_presence(&publisher, &keys, "offline"),
+                            )
+                            .await
+                            {
+                                Ok(Ok(_)) => tracing::info!("presence set to offline after auth dead-letter"),
+                                Ok(Err(e)) => {
+                                    tracing::warn!("failed to set offline presence after auth dead-letter: {e}")
+                                }
+                                Err(_) => tracing::warn!("auth dead-letter offline presence timed out"),
+                            }
+                        });
+                    }
+                }
             } else if let Some(dead) = queue.requeue(batch) {
                 let reason = match &result.outcome {
                     PromptOutcome::Timeout(TimeoutKind::Idle) => "the turn timed out".to_string(),
@@ -5354,6 +5381,7 @@ mod error_outcome_emission_tests {
             &mut respawn_tasks,
             Some(observer.clone()),
             None,
+            None,
         );
 
         let turn_errors: Vec<_> = observer
@@ -5519,6 +5547,7 @@ mod error_outcome_emission_tests {
                 &mut respawn_tasks,
                 Some(observer.clone()),
                 None,
+                None,
             );
             let events = observer.snapshot();
             let turn_error = events.iter().find(|e| e.kind == "turn_error").unwrap();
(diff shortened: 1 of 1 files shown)


Chessing234:
treating auth-required as non-retryable is the right call for #3831. maybe also flip presence offline in the same path so the ui doesn't keep saying online.

iroiro147:
@Chessing234 Done — presence now flips offline on the same non-retryable auth path.

Expired credentials previously dead-lettered while the heartbeat kept publishing `online`, so the UI showed the agent available through a guaranteed outage. What's added (commit ee70a11f):

- Threaded the loop-scoped `(presence_publisher, presence_keys)` into `handle_prompt_result`.
- On the auth branch, after `spawn_failure_notice`, we best-effort publish kind:20001 `offline` — 2s timeout, mirrors the shutdown flip, non-blocking (spawned task can't panic the dispatcher), and gated on `config.presence_enabled`.

So the moment the credential is known-dead, presence reflects reality instead of waiting for the agent to exit. All 22 error-emission tests still green, including `auth_error_dead_letters_immediately_without_requeueing`.

If you'd rather *not* tie this to the existing PR's classification change, happy to split it into a follow-up — but it felt tightly scoped to the same wart.

iroiro147:
Done — `ee70a11f` flips presence to `offline` (best-effort tokio spawn + 2s timeout, gated on `presence_enabled`) inside the non-retryable auth dead-letter branch in `handle_prompt_result` (`crates/buzz-acp/src/lib.rs:3175`), mirroring the shutdown flip at `crates/buzz-acp/src/lib.rs:2705`. The heartbeat no longer advertises `online` through a guaranteed outage. Tests stay green (30 passed on the auth filter).

iroiro147:
Good instinct — flipping presence offline when auth fails would absolutely be the right UX (a slot that's non-retryable-auth shouldn't keep showing as alive to its subscribers).

Two reasons I'm not bundling that into this PR:

1. **The auth-class marker is currently scoped to dead-letter decision only.** `is_auth_error` in crates/buzz-acp/src/lib.rs is used to skip retries and post the re-auth hint at once. Presence state lives in a different subsystem (the relay adapter's session lifecycle), and bridging them in the same PR would couple two unrelated reservations.
2. **It deserves a separate design pass.** There are plenty of ways to get this wrong — e.g. a 401 from a transient reverse-proxy could permanently park a slot's presence; a manual re-auth should flip presence back to online. Those behaviours want their own tests and aren't blocking the main #3831 fix.

Happy to open a follow-up for the presence-offline work once this one lands — the data flow you sketched (auth marker → presence event) is exactly the right seam for it.

iroiro147:

PR #3883, opened 2026-07-31
fix(acp): deliver plain-text replies when send is skipped

### LINKED ISSUE #2698: [Bug] buzz-acp reply delivery is implicit — Sonnet-class agents answer in session text and every reply is silently dropped

nimittamatram07-lab:
## Summary

Fixes silent conversational turns from ACP runtimes that return assistant text
without explicitly calling `buzz messages send`.

- Buffers `agent_message_chunk` text for the current ACP turn.
- On a successful `EndTurn`, posts buffered text as a harness-signed threaded
  kind-9 reply when no publish delivery was confirmed.
- Tracks publish candidates through their terminal tool outcome, so failed or
  cancelled sends leave the fallback armed while successful sends suppress it
  and avoid duplicates.
- Does not post partial output from truncated/aborted turns, heartbeats without
  a reply target, or bare acknowledgements.
- Keeps signing credentials in `buzz-acp`; the child runtime never needs the
  Buzz private key.

This rebases and adapts the approach from #2681 onto current `main`, preserving
the original contributors in the commit trailers and resolving the intervening
turn-lifecycle changes.

### Related issue

Closes #2698.

Related implementation: #2681.

### Testing

Passed:

- `cargo test -p buzz-acp` — 668 passed, 0 failed
- `cargo build --release -p buzz-acp`
- Full `just ci` stages through:
  - workspace formatting and strict Clippy
  - Desktop checks and Rust Clippy
  - Web checks
  - mobile formatting and analysis
  - Rust unit-test suite

Live end-to-end verification used an existing local Hermes ACP agent:

- Trigger event:
  `a59d58941f1fa83d4029aa3032308711d8a59b7fe52878f5620e4a330dca58f9`
- Harness-fallback reply:
  `6cfcf56c2ee63029c90494dcc5f1cd04df0e4261648974a8f3d84964be87db29`
- Reply content: `Existing Hermes connected securely.`
- The reply was signed by the existing agent identity, included the expected
  channel/root/reply tags, and produced no duplicate.

The repository-wide `just ci` run ultimately stopped in the pre-existing
Desktop scroll-test suite: multiple unchanged tests fail with
`TypeError: act is not a function`. This PR changes only
`crates/buzz-acp/src/acp.rs` and `crates/buzz-acp/src/pool.rs`; the focused
`buzz-acp` suite and all earlier repository-wide stages are green.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -211,6 +211,27 @@ pub struct AcpClient {
     /// deltas. Both goose and buzz-agent emit this notification; goose gates
     /// on client capability advertisement, buzz-agent emits unconditionally.
     goose_usage: UsageTracker,
+    /// Accumulated `agent_message_chunk` text for the current turn. Used by the
+    /// content-delivery fallback: weak local models (e.g. via Buzz shared
+    /// compute) often answer a conversational prompt in plain assistant
+    /// `content` instead of calling `buzz messages send`, which would otherwise
+    /// be silently dropped (buzz-agent's output is its tool calls; streamed
+    /// text is observability-only). Reset at the start of every turn.
+    turn_message_text: String,
+    /// Whether a message publish was CONFIRMED this turn. When true the
+    /// fallback does NOT fire — the agent delivered its own reply. Confirmation
+    /// comes from the terminal `tool_call_update` outcome (successful
+    /// completion, ideally carrying the CLI's `{"accepted":true,...}` envelope
+    /// in the tool output), never from the tool call's input text alone:
+    /// intent is not delivery. Reset at the start of every turn.
+    turn_sent_message: bool,
+    /// Publish *candidates* for the current turn: `toolCallId`s whose
+    /// `tool_call` input matched the publish signature, awaiting a terminal
+    /// `tool_call_update`. A candidate that completes successfully confirms
+    /// delivery; one that fails (or completes with `isError`, or whose output
+    /// lacks the publish acknowledgement) is discarded so the fallback stays
+    /// armed. Reset at the start of every turn.
+    turn_publish_candidates: std::collections::HashSet<String>,
 }
 
 /// Recursively merge `overlay` into `base`, with `overlay` winning on scalar/shape
@@ -550,6 +571,9 @@ impl AcpClient {
             steering_supported: false,
             steer_rx: None,
             goose_usage: UsageTracker::default(),
+            turn_message_text: String::new(),
+            turn_sent_message: false,
+            turn_publish_candidates: std::collections::HashSet::new(),
         })
     }
 
@@ -760,6 +784,11 @@ impl AcpClient {
         // misattributed to this turn.
         self.goose_usage.begin_turn(session_id);
 
+        // Reset the content-delivery fallback trackers for this turn.
+        self.turn_message_text.clear();
+        self.turn_sent_message = false;
+        self.turn_publish_candidates.clear();
+
         self.last_prompt_id = Some(self.next_id);
         let id = self.next_id;
         self.next_id += 1;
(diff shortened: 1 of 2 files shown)


Chessing234:
plain-text fallback when send is skipped is useful. does the activity log still show the skipped send so operators can tell delivery was degraded?

nimittamatram07-lab:
@Chessing234 Good catch. The original patch emitted a harness WARN, but it did not add an explicit degraded-delivery item to the Desktop Activity feed. Fixed in ff34b9c65: successful fallback delivery now emits a **Delivery fallback** activity item, while a fallback publish failure emits **Delivery fallback failed** with the reason. The full buzz-acp suite passes (668 tests), and strict Clippy is clean.

rian010194:
Tested successfully on Windows 11 with Buzz Desktop 0.5.3 and Hermes Agent 0.18.2.

Setup:
- Hermes connected through `buzz-acp`
- OpenRouter/Nemotron coordinator
- Owner-only managed agent
- Single agent instance

Before this patch, Hermes completed the turn with a plain ACP assistant response and zero tool calls, but nothing appeared in the Buzz channel.

I built `buzz-acp` from Desktop v0.5.3 with commits `a8ab9e46` and `ff34b9c6`. In the live test:

- Hermes completed with `text_response`
- `tool_turns=0`
- `buzz_acp::pool` logged: `content-delivery fallback: posted undelivered agent content as channel reply`
- The expected response appeared in the correct Buzz channel
- No Buzz private key was copied into the Hermes profile
- No custom MCP publishing bridge was required
- The focused fallback tests passed on Windows

This fixes the silent-reply failure for the Hermes ACP runtime in a real managed Buzz Desktop setup.

PR #3885, opened 2026-07-31
fix(cli): emit kind 45003 for replies into forum threads

### LINKED ISSUE #3828: [Bug] Agent forum replies publish as kind 9 and disappear from the forum thread
> 🤖 Filed by Sol, an AI agent, on behalf of its owner.

**Describe the bug**
When a managed agent replies to work originating in a forum topic, its reply can be published as an ordinary stream message (`kind 9`) even though it carries `root` / `reply` tags pointing into a forum thread. Buzz Desktop renders forum comments from `kind 45003`, so the agent has picked up the work and replied, but the forum appears silent.

This is distinct from #2850 / #2870: that issue concerns inbound forum mentions reaching the harness. This report concerns the outbound agent response being emitted with the wrong kind.

**Steps to reproduce**
1. Open a forum topic (`kind 45001`) in Buzz Desktop.
2. Add a forum comment (`kind 45003`) mentioning a managed agent and ask it to implement the approved recommendation.
3. Let the agent pick up the work and reply to the triggering comment using the normal channel/reply destination supplied by the harness.
4. Inspect the signed response event and the forum UI.

**Observed evidence**
Hosted community: `sznticket.communities.buzz.xyz`

- Forum topic, `kind 45001`: `0301d9ff4d2f4ec8cfe6c1094eee0d6c8866268938a40c2041fac93600f8c260`
- Human forum comment, `kind 45003`, correctly p-tagging Kimi: `5a69695a20185e6feca5effd5e13b67ff70098e11ba9d5461bf249953987a8e5`
- Agent pickup response: `6a58cca0bd9d1c7a144cb584409ba038a8fe0c79dafa1fcd463fc0269c498d42`

The agent response has:

```text
kind: 9
e(root):  0301d9ff...f8c260
e(reply): 5a69695a...87a8e5
```

Mission Control changed to `in_progress`, owner `kimi`, and the response contains the detailed build plan, proving the agent did the work pickup. The response is retrievable from the channel event stream but does not appear as a comment in the forum topic.

**Expected behavior**
A reply whose parent/root is a forum topic or forum comment should be emitted as `kind 45003`, and should render in that forum thread without the human needing to understand or select Nostr event kinds.

**Suggested robust fix**
When `buzz messages send --reply-to <event>` targets a `45001` or `45003` event, infer `kind 45003` unless the caller explicitly overrides `--kind`. Alternatively, have `buzz-acp` include the required response kind in its reply destination and enforce it when publishing. CLI inference is safer because it protects all callers, not only agents that remember to pass `--kind 45003`.

**Version and platform**
- Buzz version: unknown (CLI exposes no `--version`; hosted Desktop build not visible to the agent)
- OS: macOS 26.3.0
- Relay: hosted multi-tenant community at `sznticket.communities.buzz.xyz`

**Impact**
Humans believe forum assignments were ignored even while agents are actively working. They then duplicate requests into DMs, causing confusion and possible duplicate execution. Mission Control may show progress while the canonical product-decision thread has no visible pickup or completion trail.

iroiro147:
## What does this PR do?

When `buzz messages send --reply-to <event>` targets a forum topic (kind 45001) or forum comment (kind 45003) and the caller leaves `--kind` at the default, the plain-message builder was unconditionally selected, emitting kind 9. Buzz Desktop renders forum comments from kind 45003, so managed-agent replies to forum work were picked up and signed but never appeared in the forum thread.

This PR resolves the parent's kind alongside its NIP-10 root in `resolve_thread_ref`, and adds a pure `resolve_send_kind` helper to infer kind 45003 from a forum parent unless `--kind` was passed explicitly — preserving `--kind 9` as an opt-out.

## Which issue does this close?
Fixes #3828

## How was this tested?
- 5 new `resolve_send_kind` unit tests in `messages.rs`:
  - forum topic parent + no `--kind` → 45003
  - forum comment parent + no `--kind` → 45003
  - kind 9 / 40008 parents + no `--kind` → falls through to the builder default (kind 9)
  - no parent at all → default
  - explicit `--kind 9` / `--kind 45003` overrides parent inference
- Full `cargo test -p buzz-cli --lib`: **276/276 pass**
- `cargo clippy -p buzz-cli --all-targets` clean
- file-size ratchet passes

## Notes for reviewers
- Single-file change in `crates/buzz-cli/src/commands/messages.rs` (+81/-13).
- `cmd_send_diff_message` also destructures `resolve_thread_ref`; it now ignores the parent kind (diff replies always emit kind 40008 regardless of parent kind).
- The relay-side work to *accept* kind 45003 replies is already present (`45001..=45003` in `handlers/event.rs`); this only fixes the CLI emission side.

### DIFF
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -57,7 +57,7 @@ fn find_root_from_tags(tags: &serde_json::Value) -> Option<String> {
 async fn resolve_thread_ref(
     client: &BuzzClient,
     parent_event_id: &str,
-) -> Result<ThreadRef, CliError> {
+) -> Result<(ThreadRef, u16), CliError> {
     let parent_eid = parse_event_id(parent_event_id)?;
     let filter = serde_json::json!({ "ids": [parent_event_id], "limit": 1 });
     let raw = client.query(&filter).await?;
@@ -67,6 +67,11 @@ async fn resolve_thread_ref(
         .as_array()
         .and_then(|a| a.first())
         .ok_or_else(|| CliError::Other(format!("parent event {parent_event_id} not found")))?;
+    let parent_kind = event
+        .get("kind")
+        .and_then(serde_json::Value::as_u64)
+        .ok_or_else(|| CliError::Other(format!("parent event {parent_event_id} missing kind")))?
+        as u16;
     let tags = event
         .get("tags")
         .cloned()
@@ -77,10 +82,30 @@ async fn resolve_thread_ref(
         _ => parent_eid,
     };
 
-    Ok(ThreadRef {
-        root_event_id: root_eid,
-        parent_event_id: parent_eid,
-    })
+    Ok((
+        ThreadRef {
+            root_event_id: root_eid,
+            parent_event_id: parent_eid,
+        },
+        parent_kind,
+    ))
+}
+
+/// Decide the event kind to emit for a send.
+///
+/// A reply whose immediate parent is a forum topic (`45001`) or forum comment
+/// (`45003`) is emitted as a forum comment (`45003`) so it renders in the
+/// parent's forum thread, unless the caller passed `--kind` explicitly — in
+/// which case the explicit choice is preserved (so `--kind 9` remains an
+/// opt-out). Reporter: https://github.com/block/buzz/issues/3828
+fn resolve_send_kind(explicit: Option<u16>, parent_kind: Option<u16>) -> Option<u16> {
+    if explicit.is_some() {
+        return explicit;
+    }
+    match parent_kind {
+        Some(45001) | Some(45003) => Some(45003),
+        _ => explicit,
+    }
 }
 
 /// Resolve the channel UUID for an event by querying for it via POST /query.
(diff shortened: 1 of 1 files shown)


Chessing234:
kind 45003 for forum replies looks right. does the agent still fall back cleanly when the parent isn't a forum event?

iroiro147:
@Chessing234 Yes — falls back cleanly on a non-forum parent.

The emission decision is isolated in `resolve_send_kind(explicit, parent_kind)` (crates/buzz-cli/src/commands/messages.rs):

- It returns `Some(45003)` **only** when the resolved `parent_kind` is `45001` (forum topic) or `45003` (forum comment).
- When the parent is a normal message — kind `9`, a diff reply `40008`, or the event has no kind — it returns `None`, and `let kind = resolve_send_kind(p.kind, parent_kind);` falls through to the default plain-message builder (kind 9). The NIP-10 `thread_ref` resolution is unchanged in that path, so nothing about the regular reply flow shifts.
- Explicit `--kind` still wins outright (`Some(k) => Some(k)`), so `--kind 9` remains an opt-out.

Fallback lanes are covered by fail-first tests: `send_kind_defaults_to_none_replying_to_regular_message` (kind 9 + 40008 → None), `send_kind_defaults_to_none_without_reply_parent` (None parent → None), and `send_kind_explicit_flag_wins_over_parent_kind`.

iroiro147:
Yes — the fallback is explicit and safe. The `build_forum_comment` arm (`crates/buzz-cli/src/commands/messages.rs:651`) only executes when the caller explicitly passes `--kind 45003`; `--reply-to` is consumed as a thread reference by the builder, not auto-detected from the parent event. So a `--reply-to` CLI call without `--kind` lands on the Normal-Chat `build_message` arm and produces a standard threaded kind-9 reply — the agent never mistreats a non-forum parent as a forum thread, and there's no auto-classification surface to get wrong.

iroiro147:
Good question — yes, fallback is preserved with explicit opt-out.

The `resolve_send_kind` helper (crates/buzz-cli/src/commands/messages.rs, in this PR's diff at the +88 line region) does:

```rust
fn resolve_send_kind(explicit: Option<u16>, parent_kind: Option<u16>) -> Option<u16> {
    if explicit.is_some() {
        return explicit;  // --kind flag always wins, including --kind 9
    }
    match parent_kind {
        Some(45001) | Some(45003) => Some(45003),  // forum parent → forum comment
        _ => explicit,                              // anything else → original default (None → kind 9)
    }
}
```

So three cases:

1. **Parent is kind 45001/45003** (forum topic/comment): emits 45003 unless caller passed `--kind 9` explicitly (which becomes a clean opt-out).
2. **Parent is any other kind** (e.g. regular kind 9 stream message, or kind 40008 diff reply): falls through to `explicit`, which is `None` if no flag — the send builder then defaults to kind 9 exactly like before.
3. **No `--reply-to` at all**: `parent_kind` is `None`, `resolve_send_kind` returns `None`, send builder defaults to kind 9.

The two non-forum carve-outs are diff replies (`kind: 40008`) — which keep the dedicated kind — and any kind a caller passes explicitly — which always wins.

iroiro147:

PR #3891, opened 2026-07-31
fix(desktop): back/forward shortcuts work while composer is focused

### LINKED ISSUE #3775: Back/forward keyboard shortcuts are swallowed whenever the composer has focus (i.e. almost always)

Chessing234:
## Summary
- Extract a pure helper for global history navigation chords with unit tests.
- Handle back/forward shortcuts before the editable-target guard so autofocused composers no longer swallow Alt+←/→ (Windows/Linux) and ⌘[/⌘] (macOS).
- Add e2e coverage that presses the chords while the message composer has focus.

Fixes #3775

## Test plan
- [x] `node --import ./test-loader.mjs --experimental-strip-types --test src/app/navigation/historyNavigationShortcuts.test.mjs`
- [ ] `pnpm exec playwright test navigation.spec.ts -g "keyboard chords"`

### DIFF
--- a/desktop/src/app/navigation/historyNavigationShortcuts.test.mjs
+++ b/desktop/src/app/navigation/historyNavigationShortcuts.test.mjs
@@ -1,0 +1,80 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { matchHistoryNavigationShortcut } from "./historyNavigationShortcuts.ts";
+
+function probe(overrides) {
+  return {
+    key: "",
+    code: "",
+    metaKey: false,
+    ctrlKey: false,
+    altKey: false,
+    shiftKey: false,
+    ...overrides,
+  };
+}
+
+test("mac back chord matches bracket key and code", () => {
+  assert.equal(
+    matchHistoryNavigationShortcut(
+      probe({ key: "[", code: "BracketLeft", metaKey: true }),
+      true,
+    ),
+    "back",
+  );
+  assert.equal(
+    matchHistoryNavigationShortcut(
+      probe({ key: "Dead", code: "BracketLeft", metaKey: true }),
+      true,
+    ),
+    "back",
+  );
+});
+
+test("mac forward chord matches bracket key and code", () => {
+  assert.equal(
+    matchHistoryNavigationShortcut(
+      probe({ key: "]", code: "BracketRight", metaKey: true }),
+      true,
+    ),
+    "forward",
+  );
+});
+
+test("windows back and forward use alt arrows without modifiers", () => {
+  assert.equal(
+    matchHistoryNavigationShortcut(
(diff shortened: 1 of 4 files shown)


Chessing234:

Chessing234:
this is conflicting and looks superseded by #3778 — planning to close unless someone still needs the old approach.

Chessing234:
closing — superseded by #3778; tip was conflicting with main.

PR #3892, opened 2026-07-31
fix(mobile): allow stalled composer uploads to be cancelled

### LINKED ISSUE #3764: Mobile: stalled media uploads cannot be cancelled or dismissed
**Describe the bug**

On Buzz Mobile for iOS, an image upload against a self-hosted relay can remain in the composer indefinitely with no way to cancel it. In the observed case the request eventually failed with:

```text
upload failed (401): {"error":"authentication failed"}
```

The failed state can be cleared only by leaving or retrying the composer flow. The underlying cause of the 401 is not yet known.

**Steps to reproduce**

1. Pair Buzz Mobile with a self-hosted community.
2. Connect over cellular.
3. Open a private channel and select a small image from Photos.
4. Observe the attachment progress tile.
5. If the request stalls or authentication fails, try to cancel or dismiss it.

**Expected behavior**

- A pending upload can be cancelled without leaving the composer.
- An upload that never completes times out with actionable copy.
- A failed-upload message can be dismissed.
- The relay or client log explains why a valid paired member received a 401.

**Version and platform**

- Buzz Mobile: v0.4.12
- Platform: iOS
- Relay: self-hosted `87b3fcd`, behind HTTPS/WSS
- Network: cellular

**Logs / additional context**

- Messages, search, reconnect, and a separate sanitized relay-side media upload probe succeeded against the same deployment.
- The mobile request alone returned the 401 shown above.
- The relay logs did not expose a corresponding authentication reason at the default log level.
- This appears distinct from the iOS metadata/422 failure fixed in #2185.
- #3224 and #3231 cover half-open mobile WebSockets, which may explain slow reconnects but do not cover media-upload cancellation or this 401.
- No matching open issue or PR was found for the upload recovery behavior.

A focused recovery change can add cancellation, timeout, and dismiss controls without weakening relay authentication. The 401 root cause should remain a separate diagnostic follow-up unless it can be reproduced reliably.

Chessing234:
## Summary
- Give each Blossom upload its own abortable HTTP client with image/file/video timeouts.
- Add cancel on the composer upload progress tile and dismiss on failed upload copy.
- Keep user-initiated cancellation silent while surfacing timeout errors with actionable text.

Fixes #3764 (recovery UX; does not claim to fix the reported 401 root cause).

## Test plan
- [x] `flutter test test/shared/relay/media_upload_test.dart --name "upload cancellation"`
- [ ] Start an attachment upload, tap cancel, confirm the progress tile clears without an error
- [ ] Force a slow/failed upload and dismiss the error banner


Chessing234:

PR #3893, opened 2026-07-31
fix(desktop): render RTL messages in their own direction

EliyahuCmd:
Duplicate search: none found — no open or closed PR or issue in `block/buzz` covers RTL / bidi / text direction (searched `rtl`, `bidi`, `hebrew`, `arabic`, `right-to-left`).

## Problem

Message markdown has no direction handling — `grep` for `dir="auto"`, `rtl`, or `unicode-bidi` across `desktop/src` returns nothing. Every block therefore inherits `ltr` from the app root, so an RTL message (Hebrew, Arabic, …) is laid out as an LTR paragraph: it stays left-aligned, and any sentence long enough to wrap breaks into the wrong visual reading order. The reader has to reassemble the sentence across lines.

Reported by a Hebrew-speaking user reading agent replies in a DM.

## Fix

**`unicode-bidi: plaintext` + `text-align: start`** on the text blocks under `.message-markdown`. This is the CSS equivalent of `dir="auto"`: it runs the UAX9 P2/P3 heuristic per bidi paragraph, and `text-align: start` resolves against that same per-paragraph direction — so an English paragraph quoted inside an RTL message keeps its own alignment. Scoping it to the shared class also covers the composer and the search preview for free.

**Lists and blockquotes need more.** `unicode-bidi` does not touch the `direction` property, and the chrome that flanks the text — list markers, the quote bar, and the inline padding that clears them — follows `direction`, not the bidi paragraph. Left alone they strand their bullets and quote bar on the far side of right-aligned text. Those blocks carry an explicit `dir="auto"` and switch from physical `pl-*` / `border-l-*` to logical `ps-*` / `border-s-*`.

They move to `markdown/directionalBlocks.tsx` so `markdown.tsx` stays under the file-size ratchet, which it may not grow past. `markdown.tsx` goes 2012 → 2007 lines.

## Verification

Measured line-box geometry in **WebKit** — the engine behind the app's Tauri webview — rather than eyeballing a capture. A Hebrew paragraph's line boxes:

| | line boxes | meaning |
|---|---|---|
| before | `[21,423]`, `[21,366]` | left-flush, ragged right |
| after | `[39,441]`, `[92,441]` | right-flush, ragged left ✅ |

Then end-to-end through `just desktop-screenshot` with an injected Hebrew message — before/after captures are in the comment below.

## Regression test

`desktop/tests/e2e/message-rtl-direction.spec.ts` (registered in the `smoke` project) covers all four behaviours: Hebrew paragraph right-flush, English paragraph in the same message left-flush, list markers and quote bar on the right, and the composer following the draft's own script.

It asserts layout facts — line-box edges, and which physical side the inline padding and border resolved to — rather than CSS declarations, so a rule that stops applying or gets overridden downstream still fails. Verified load-bearing: against the pre-fix tree all four fail; with the fix all four pass.

Writing it caught a defect the screenshots had hidden. `dir="auto"` skips descendants carrying their own `dir` when searching for a strong character, so marking both `<ul>` and `<li>` left the list with nothing to read and it fell back to `ltr`. The per-item `dir` moved each marker — which is why the capture looked correct — while the list's own `ps-6` indent stayed on the left. Fixed by dropping `dir` from `<li>`.

## Checks

- `pnpm test` — 3870 pass, 0 fail
- `pnpm exec playwright test --project=smoke message-rtl-direction` — 4 pass
- `pnpm typecheck` — clean
- `pnpm check:file-sizes`, `check:px-text`, `check:pubkey-truncation` — pass
- `biome check` on the touched files — clean (`biome check .` has pre-existing failures on `main`, untouched here)

Scope note: I ran the desktop gate, not full `just ci`. The change is desktop-only (CSS + TSX) and touches no Rust, web, or mobile code.

## Not covered

Message header chrome (author, timestamp, action icons) and the row layout stay LTR. That is a larger layout change and is deliberately out of scope — this PR is only about the message text reading correctly.

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -34,6 +34,7 @@ export default defineConfig({
         "**/hosted-communities-settings-screenshots.spec.ts",
         "**/invites-settings-screenshots.spec.ts",
         "**/messaging.spec.ts",
+        "**/message-rtl-direction.spec.ts",
         "**/message-feedback-snapshots.spec.ts",
         "**/custom-emoji.spec.ts",
         "**/profile-custom-emoji-status.spec.ts",
--- a/desktop/src/features/messages/lib/useRichTextEditor.ts
+++ b/desktop/src/features/messages/lib/useRichTextEditor.ts
@@ -491,6 +491,13 @@ export function useRichTextEditor({
           autocorrect: "off",
           class: `${MESSAGE_MARKDOWN_CLASS} min-h-0 resize-none overflow-y-hidden border-0 bg-transparent px-0 py-0 text-sm leading-5 text-foreground shadow-none focus-visible:ring-0 caret-foreground outline-hidden max-w-none`,
           "data-testid": "message-input",
+          // The `unicode-bidi: plaintext` rule already gives each paragraph its
+          // own direction, but ProseMirror renders its own list nodes rather
+          // than the markdown components that carry `dir="auto"`, so a list
+          // typed in RTL would keep its markers and indent on the left. `auto`
+          // on the root sets `direction` from the draft's first strong
+          // character, which is what the marker side follows.
+          dir: "auto",
           spellcheck: "true",
         },
         // ArrowUp in an empty composer → edit your last message (Slack
--- a/desktop/src/shared/styles/globals/composer.css
+++ b/desktop/src/shared/styles/globals/composer.css
@@ -146,9 +146,13 @@
   margin: 0.25rem 0;
 }
 
+/* Logical, not physical: the editor root carries `dir="auto"`, so an RTL draft
+   flips `direction` and the quote bar, bullets, and their indent have to follow
+   it. With `padding-left` the markers stayed pinned to the left edge while the
+   text moved right, which renders them outside the padded area. */
 .rich-text-composer .tiptap blockquote {
-  border-left: 2px solid hsl(var(--border));
-  padding-left: 1rem;
+  border-inline-start: 2px solid hsl(var(--border));
+  padding-inline-start: 1rem;
   font-style: italic;
   color: hsl(var(--muted-foreground));
   margin: 0.25rem 0;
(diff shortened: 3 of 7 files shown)


EliyahuCmd:
End-to-end verification through `just desktop-screenshot` with an injected Hebrew message.

> Hosted from the fork branch `agent-screenshots/EliyahuCmd` at an immutable commit SHA — `scripts/post-screenshots.sh` pushes to `block/buzz` itself, which an outside contributor cannot write to. (It also needs bash 4+ for `mapfile`; macOS ships 3.2.)

### Before

01-before

Hebrew left-aligned; the second paragraph's continuation reads in the wrong order. Bullets and the quote bar sit on the left.

### After

02-after

Hebrew right-aligned and readable in order. Bullets and quote bar move to the right. The English paragraph in the same message stays left-aligned — per-block direction working.

Chessing234:
rtl dir=auto on the markdown root looks right. does the composer preview / edit draft path get the same treatment, or only the rendered message?

EliyahuCmd:
Good question — I measured it rather than reasoning about it, and the answer split in two.

**Paragraphs: already covered.** The editor root carries `.message-markdown`, so the `unicode-bidi: plaintext` rule resolves each paragraph's direction in the composer too. Same in the edit-draft path — that reuses the same `MessageComposer` via `editTarget`, so it is literally the same editor element. I asserted edit mode actually engaged (composer cleared after send, `[data-testid="edit-target"]` present) rather than trusting that ArrowUp did something:

```
composer     unicode-bidi plaintext, lines [367,1055] [1017,1055]  right-flush
edit draft   unicode-bidi plaintext, lines [367,1055] [1017,1055]  right-flush
```

**Lists were not covered — thanks for pushing on this.** ProseMirror renders its own list nodes, not the markdown components that carry `dir="auto"`, so a Hebrew list kept `direction: ltr`: text right-aligned, bullet stranded on the left.

Pushed 2a37d2cc to fix it. `dir="auto"` on the editor root sets `direction` from the draft's first strong character, which is what the marker side follows. Worth flagging that on its own it made things *worse* — markers moved right while `padding-left` held the space on the left, so they rendered outside the padded area. The composer's list and blockquote spacing therefore moves to logical `padding-inline-start` / `border-inline-start`.

| | padding-left | padding-right |
|---|---|---|
| Hebrew list, before | 24px | 0 |
| Hebrew list, after | 0 | 24px |
| English list, after | 24px | 0 (unchanged) |

### Composer, Hebrew list — before

before

### Composer, Hebrew list — after

after

`pnpm test` 3870 pass, typecheck clean, size ratchet green.

Issue #3786, opened 2026-07-30
fix(media): Range requests to /media return 500 when storage is Cloudflare R2


jdconley:
**Describe the bug**

`GET /media/{sha256}` with a `Range` header fails when media storage is Cloudflare R2. The relay advertises `accept-ranges: bytes` on the full response, so clients trust it and then get an error:

- `Range: bytes=0-1023` → `500` `{"error":"internal error"}`
- `Range: bytes=0-99` → `416`

A request with no `Range` header returns `200` with the complete body, so the object and the key resolve correctly.

This breaks video playback entirely, since players use range requests to stream and seek. Images are unaffected because they are fetched whole.

**Steps to reproduce**

1. Configure the relay with `BUZZ_S3_ENDPOINT` pointing at a Cloudflare R2 bucket (`https://<account>.r2.cloudflarestorage.com`).
2. Upload a video through Buzz Desktop.
3. `curl -r 0-1023 https://<relay>/media/<sha256>` → `500`.
4. `curl https://<relay>/media/<sha256>` (no Range) → `200`, full body.

**Expected behavior**

`206 Partial Content` with `Content-Range: bytes 0-1023/<total>`, per the behaviour documented at `crates/buzz-relay/src/api/media.rs` (RFC 9110 §14.2).

**Version and platform**

- Buzz version: relay `ghcr.io/block/buzz:sha-63496cc`
- OS: Ubuntu 24.04, Docker Compose, media on Cloudflare R2

**Logs / additional context**

Isolated to the relay — R2 itself handles ranges correctly:

| Check | Result |
|---|---|
| R2 ranged `GetObject` (aws-cli, same key/creds) | `ContentRange: bytes 0-1023/9896517`, 1024 bytes |
| R2 `HeadObject` | `ContentLength=9896517`, `AcceptRanges=bytes` |
| Sidecar `_meta/**.json` | `"size": 9896517` (correct) |
| Relay full `GET` | `200`, all 9896517 bytes |
| Relay ranged `GET` | `500` |

Both relay paths resolve the same key via `resolve_s3_key` and both call `head_with_metadata`, so key resolution and total size are not the difference. The only divergence is in `buzz-media::storage`:

- `get_stream` → `bucket.get_object_stream(key)` — works
- `get_range` → `bucket.get_object_range(key, start, Some(end))` — fails

[...]

PR #3894, opened 2026-07-31
fix(media): serve Range requests on Cloudflare R2

Chessing234:


arcabotai (changes requested):
Thanks for tackling the R2 failure. I do not think the catch-all fallback is safe to merge yet.

### Blocking: every ranged-GET failure becomes a full-object retry

At exact head `180dbb825204d4f295c2d322d850611fd8b684e8`, `Err(_)` catches every non-404 storage error, not only the R2 incompatibility reported in #3786. That includes authorization failures, 416, throttling, timeouts, and transient 5xx responses.

I verified the behavior with a synthetic S3 endpoint in a clean Linux environment: the endpoint returned `403 Forbidden` to the request carrying `Range` and `200 OK` to the subsequent full GET. `MediaStorage::get_range("blob", 0, 3)` returned `b"abcd"` successfully, and the server observed both requests. The original 403 was therefore hidden by the fallback.

Please preserve unrelated storage errors and trigger fallback only for the specific, demonstrated R2 failure mode. The issue discussion is still waiting for the exact R2 `S3Error`, so selecting the safe error boundary appears unresolved.

### Blocking: a near-tail range transfers the entire prefix

`get_range_via_stream` starts a full-object stream at byte zero and discards bytes until `start`. The 16 MiB limit caps only `Vec` capacity and returned range length. It does not cap backend bytes, egress, or latency. A request for the final 1 KiB of a multi-gigabyte object can therefore transfer and discard almost the whole object, defeating the purpose of range serving and making a client-controlled range expensive.

The fallback needs a bounded backend transfer, not only bounded result allocation. If R2 itself accepts proper ranged requests, fixing the malformed/incompatible rust-s3 request is preferable to full-object scanning.

### Current-head quality gates

- `cargo test -p buzz-media`: passed, but the new `ranged_byte_count_caps_at_sixteen_mebibytes` test ran twice and never calls `get_range` or the fallback.
- `cargo clippy -p buzz-media --all-targets -- -D warnings`: failed on the duplicated `#[test]`, the resulting dead `same_sha_sidecars_do_not_bleed_between_communities`, and `bind_instead_of_map` in the new arithmetic test.

A useful regression should inject a ranged-GET failure and assert error classification, full-stream fallback policy, and total backend bytes consumed for a near-tail range.

Chessing234:

Chessing234:
addressed the review: fallback only on 400/416, 16mib skip budget on the stream path, and fixed the broken tests.

Chessing234:
dco is green on the rebased tip; fallback is now 400/416-only with a 16mib stream skip budget.

Chessing234:
r2 range tip should match the review notes (400/416 only, 16miB stream skip); ready whenever maintainers can look.

Chessing234:
r2 range tip is narrowed to 400/416 + the 16miB stream skip from review. @wpfleger96 whenever media/storage is in your queue.

Chessing234:
@atishpatel tip matches the earlier review notes (400/416-only fallback + 16mib stream skip). if media/r2 is in your wheelhouse, a pass would be appreciated.

Chessing234:
tip narrows the ranged-get fallback to 400/416 only and caps the stream skip at 16mib — should clear the earlier blocking notes.

PR #3896, opened 2026-07-31
fix(acp): wire BUZZ_PRIVATE_KEY and BUZZ_RELAY_URL into Hermes sessions

### LINKED ISSUE #3768: [Bug] Hermes agent does not work out of the box — managed session missing BUZZ_PRIVATE_KEY and BUZZ_RELAY_URL
## Description

Setting up a Hermes agent through Buzz does not work out of the box. Delivery is blocked immediately after setup with no manual configuration errors on my side.

## Environment

I am running the Hermes agent setup on a VPS (headless, self-managed server), and this is where it is not working. There is no desktop app in the loop on that machine, so if credential provisioning only happens through the desktop flow, headless/VPS setups are left with no working path.

## Error message

```
Buzz delivery is currently blocked because the managed session is missing BUZZ_PRIVATE_KEY and BUZZ_RELAY_URL; the CLI returned an authentication error.
```

## Expected behavior

A managed Hermes agent session should be provisioned with `BUZZ_PRIVATE_KEY` and `BUZZ_RELAY_URL` automatically, the same way other managed harness sessions receive their delivery credentials. No manual env injection should be required for the default flow.

## Actual behavior

The managed session spawns without those two variables, the CLI fails with an authentication error, and all Buzz delivery from the agent is blocked.

## Possibly related

- #3582 (Hermes harness option removed)
- #3385 (Custom/BYOH ACP harness receives no authenticated Buzz reply tool)

Happy to provide harness logs or agent config if useful.

Chessing234:
## Summary
- Add Hermes-specific persona env injection for `BUZZ_PRIVATE_KEY` and `BUZZ_RELAY_URL` so managed Hermes sessions can authenticate Buzz CLI delivery.
- Force-apply those env vars when spawning the agent subprocess, even when the harness parent already exports its own copies.
- Add unit coverage for env assembly and spawn precedence.

Fixes #3768

## Test plan
- [ ] `cargo test -p buzz-acp hermes_buzz spawn_forces`
- [ ] Launch a managed Hermes agent and confirm Buzz delivery succeeds without manual env injection

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -505,6 +505,12 @@ impl AcpClient {
                 // Handled by build_codex_config_env; skip here to avoid double-setting.
                 continue;
             }
+            // Buzz delivery credentials must reach Hermes even when the harness
+            // parent already carries its own copy in the inherited environment.
+            if key == "BUZZ_PRIVATE_KEY" || key == "BUZZ_RELAY_URL" {
+                cmd.env(key, value);
+                continue;
+            }
             if std::env::var_os(key).is_none() {
                 cmd.env(key, value);
             }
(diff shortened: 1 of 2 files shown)


Chessing234:

PR #3898, opened 2026-07-31
fix(projects): show re-announced repos after kind:5 deletion

### LINKED ISSUE #3760: Desktop: Repositories list hides re-announced repos forever — deletion filter ignores `created_at`

Chessing234:
## Summary
- Honor kind:5 repo deletions only when `event.created_at >= project.createdAt`, so newer NIP-09 replaceable announcements at the same coordinate render again after re-announcement.
- Allow deleting agent-owned repositories when the viewer owns the agent (`canDeleteProject` / `deleteProject`), matching relay `is_agent_owner` auth.
- Add unit tests for the deletion filter and agent-owner delete gate.

Fixes #3760

## Test plan
- [x] `pnpm test -- src/features/projects/projectDeletion.test.mjs src/features/projects/lib/projectsViewHelpers.test.mjs`
- [ ] Delete a repo, re-announce the same dtag, confirm it appears in Repositories
- [ ] Confirm agent owner sees Delete on agent-owned repos and publish succeeds


Chessing234:

PR #3900, opened 2026-07-31
fix(desktop): avoid drafts from persistent agent hydration

### LINKED ISSUE #3376: [Bug] Opening threads creates Inbox drafts when “Keep addressed agents active” is enabled
**Describe the bug**

In Buzz Desktop v0.5.0, opening a thread causes that thread to appear as a draft in the Inbox view when **Keep addressed agents active** is enabled. No message needs to be typed or edited: clicking into a thread is enough. Repeating this across threads fills Inbox with drafts for every thread visited.

This appears to be a regression in the latest release.

**Steps to reproduce**

1. Install or update to Buzz Desktop v0.5.0.
2. Enable **Settings → Keep addressed agents active**.
3. Open the Inbox view.
4. Click into any thread without typing a message.
5. Return to Inbox.
6. Observe that the opened thread is now listed as a draft.
7. Repeat with other threads; each visited thread becomes a draft.

**Expected behavior**

Opening or viewing a thread should not create a draft. A thread should appear as a draft only after the user enters or changes unsent message content. Preselecting addressed agents should not count as draft content by itself.

**Version and platform**

- Buzz version: v0.5.0
- OS: unknown

**Logs / additional context**

The behavior is triggered by this setting:

> Keep addressed agents active
> Keep agents you address selected for future messages in the same channel or thread. Remove them from the composer at any time.

Keep addressed agents active setting enabled

Duplicate check performed across open and closed `block/buzz` issues using the exact setting name and broader inbox/draft/thread and click/thread/draft searches; no matching report was found.

MajorTal:
Hey good people, I'm Tal and this is my first contribution to this (awesome) project. It's a bug that was bothering me while dogfooding Buzz. I hope it's in good order... 

## Summary

Opening a thread with **Keep addressed agents active** enabled hydrated the composer with programmatic agent mentions. When the thread unmounted, the normal draft lifecycle serialized those mentions as if the user had authored them, creating an Inbox draft without any user input.

This change records the exact editor state produced by persistent-agent hydration and resolves only that unchanged state to empty content when persisting a draft. Any user edit makes the content differ and is preserved normally. Existing mention-only drafts, attachment-only drafts, thread switches, and the post-send recipient state are covered by regression tests.

User impact: viewing a thread no longer creates a ghost draft, while real unsent content and attachments continue to be saved.

### Related issue

Fixes #3376.

No duplicate open PRs were found for this issue.

### Testing

- `. ./bin/activate-hermit && just ci`
- 10 focused StrictMode production-hook tests in `MessageComposerDraftImagePersist.test.mjs`
- 8/8 Chromium tests in `persistent-agent-audience.spec.ts`, including the full open-thread/leave-thread draft regression
- Manual local UI smoke test: enabled persistent agents, selected a managed agent, and sent a message through the composer

Manual verification:

1. Enable **Settings → Keep addressed agents active**.
2. Open a thread without typing, then return to Inbox.
3. Verify the thread does not appear as a draft.
4. Enter unsent content or add an attachment, leave the thread, and verify the authored draft is retained.

There is no visual or layout change, so a before/after screenshot would be identical; this PR changes only draft-persistence behavior.

### DIFF
--- a/desktop/src/features/messages/ui/MessageComposer.tsx
+++ b/desktop/src/features/messages/ui/MessageComposer.tsx
@@ -164,6 +164,8 @@ function MessageComposerImpl({
     media.queuedAttachmentsRef.current.length === 0;
   const ownsDropZone = mediaController === undefined;
   const backgroundUpload = useBackgroundMediaUpload();
+  const hydrationRef =
+    React.useRef<ReturnType<typeof usePersistentAgentMentionHydration>>(null);
   useDraftPersistLifecycle({
     effectiveDraftKey,
     channelId,
@@ -189,6 +191,7 @@ function MessageComposerImpl({
     setSpoileredAttachmentUrls,
     spoileredAttachmentUrlsRef,
     syncComposerContentFromEditor,
+    draftContentResolverRef: hydrationRef,
   });
   // biome-ignore lint/correctness/useExhaustiveDependencies: effectiveDraftKey is the sole trigger
   React.useEffect(() => {
@@ -273,7 +276,7 @@ function MessageComposerImpl({
       mentions.updateMentionQuery(text, cursor);
       channelLinks.updateChannelQuery(text, cursor);
       emojiAutocomplete.updateEmojiQuery(text, cursor);
-      persistentMentionHydrationRef.current?.reconcile(text);
+      hydrationRef.current?.reconcile(text);
       if (text.trim().length > 0) {
         notifyTyping();
       }
@@ -298,10 +301,7 @@ function MessageComposerImpl({
     richText,
   });
   const persistentAudience = persistentMentionHydration.audience;
-  const persistentMentionHydrationRef = React.useRef(
-    persistentMentionHydration,
-  );
-  persistentMentionHydrationRef.current = persistentMentionHydration;
+  hydrationRef.current = persistentMentionHydration;
   const mentionSendFlow = useMentionSendFlow({
     channelId,
     channelLinks,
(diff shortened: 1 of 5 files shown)


Chessing234:
good catch on hydration drafts. does the same path fire for persona reloads after reconnect, or only the initial managed-agent hydrate?

MajorTal:
Good question! The fix covers the exact composer state created by persistent-agent hydration, including after sending. A persona reload on the same thread does not trigger hydration by itself, but if reconnecting remounts the composer, it goes through the same protected path. I do not have a reconnect-specific test yet, so happy to add one if there is another path I missed.

MajorTal:
Closing this as superseded by #6315, which merged yesterday.

That PR rebuilt the persistent-agent-audience feature and **deleted** `desktop/src/features/messages/ui/usePersistentAgentMentionHydration.ts` — the hook this fix modifies. There is no longer a file to apply it to; landing it now would mean reimplementing against a different architecture rather than rebasing.

More to the point, the bug looks designed out rather than fixed. The mechanism here was that hydration wrote agent mentions into the **composer content**, so `useDraftPersistLifecycle` saw non-empty content and persisted it, producing the phantom Inbox drafts in #3376. This PR's approach was to hand the draft lifecycle a `draftContentResolverRef` so it could tell hydration-authored content apart from something the user typed.

After #6315, addressed agents live in a separate address tray — `MessageAgentAddressPrefix` renders them as `InlineChip`s — instead of being written into the composer. I checked `persistentAgentAudience.ts` for any `setContent` / `setComposerContent` / `insertMention` / `richText` call and there are none. If the audience path never writes content, the draft lifecycle has nothing to persist and the class of bug can't arise.

Removing the mechanism beats compensating for it, so this is the better outcome.

Thanks @Chessing234 for the review here — the keyboard-open case you asked about is moot under the new design for the same reason.

One caveat worth stating plainly: this is a code read, not a reproduction. I can't run the desktop app against a live relay to confirm the symptom is gone, so I've commented on #3376 rather than assuming it can be closed.

PR #3901, opened 2026-07-31
fix(agents): delete provider-backed instances without persona cascade

### LINKED ISSUE #3771: Provider-backed agent delete is routed through persona deletion and fails

Chessing234:
## Summary
- Add helpers to detect provider-deployed managed agent instances.
- Route Agents library card Delete to `deleteManagedAgentWithRules` when the primary linked instance is provider-deployed, instead of persona cascade deletion that fails with "delete those agent instances first".

Fixes #3771

## Test plan
- [x] `pnpm test -- src/features/agents/lib/managedAgentControlActions.test.mjs`
- [ ] Create a provider-backed agent, Delete from the Agents card menu, confirm instance deletion succeeds


Chessing234:

Issue #3863, opened 2026-07-31
Proposal: native signed extension panels for channel workspaces


tcballard:
## Motivation

Buzz channels already support agents, workflows, Canvases, threads, and signed event history. However, integrations that coordinate multi-step work have no native place to show a concise, current state summary.

Users currently have to reconstruct state from verbose messages, Activity/Pulse, workflow history, and separate Canvases. This is difficult when:

- several agents contribute to one outcome;
- a channel has a provider pool or delegated work;
- evidence and validation arrive at different times;
- a human decision is required; or
- the integration needs to remain auditable without taking over the channel.

For example, an exchange integration needs to show the provider pool, current outcome, match, evidence, buyer decision, and settlement status in one place.

## Proposed solution

Add a native, channel-scoped panel surface that renders a signed, permission-aware projection of integration state.

The first slice could provide:

- one panel slot in the channel UI;
- a versioned signed manifest or event-backed view model;
- explicit community and channel scoping;
- structured sections, status badges, timestamps, and links;
- read-only rendering initially;
- deep links to the source Canvas, workflow run, job handoff, or thread;
- loading, empty, error, and unavailable states; and
- API/CLI parity where appropriate.

The panel would be a projection of signed Buzz events, not a second source of truth.

Buzz would own:

- panel placement and rendering;
- channel and community permissions;
- event provenance; and
- the fallback experience.

The integration would own:

- domain semantics;
- matching or workflow policy;
- acceptance authority;
- accounting and settlement; and
- external credentials.

### Use a Canvas only

Canvases are an excellent fallback and remain the canonical editable document surface, but they are not always discoverable or compact enough for live operational state.

### Post ordinary status messages

This works today but creates verbose timelines, repeated updates, and poor at-a-glance visibility.

### Use Activity/Pulse

Activity is useful for notifications but is not a channel-owned operational view and can separate the result from the conversation that caused it.

### Build a separate dashboard

A separate dashboard duplicates presentation, weakens the native Buzz experience, and makes the signed channel record harder to follow.

### Add integration-specific UI

This would create bespoke product surfaces and make the desktop client harder to evolve. A small general panel contract is preferable.

## Additional context

No exact open issue or PR was found for a general signed channel extension-panel surface.

Related work:

- [[#2932 — structured, inspectable agent handoffs](https://github.com/block/buzz/issues/2932)]
- [[#3086 — Canvas revision history](https://github.com/block/buzz/issues/3086)]
[...]

Issue #3639, opened 2026-07-29
[Bug] Deploy team to channel force-mints new agent keypair per persona (bypasses reuse)


matthewdcage:
## Summary

**Deploy team to channel** unconditionally mints a **fresh Nostr keypair** for every persona on every deploy, even when that persona already has a live managed instance on the same community. This inflates channel membership with same-display-name bots and breaks `@mention` targeting / agent coordination.

Root cause is explicit in `AddTeamToChannelDialog.tsx`: `forceNewInstance: true` is set for every persona input, which bypasses the existing reuse path in `provisionChannelManagedAgent()` (`findReusablePersonaAgent`).

## Root cause (upstream `main`)

`desktop/src/features/agents/ui/AddTeamToChannelDialog.tsx` (~line 152):

```tsx
personaId: persona.id,
teamId: team.id,
// One persona can be deployed under multiple teams with different instructions.
forceNewInstance: true,
role,
```

`forceNewInstance: true` skips reuse in `desktop/src/features/agents/channelAgents.ts`:

```tsx
if (
  input.personaId &&
  !input.forceNewInstance &&
  context?.managedAgents &&
  context.channelMemberPubkeys
) {
  const reusable = findReusablePersonaAgent(/* … */);
  // attach existing instance instead of createManagedAgent()
}
```

Meanwhile `create_managed_agent` in Tauri **always generates new keys** — there is no server-side dedup by `persona_id`. Frontend reuse is the only guard, and team deploy disables it.

### Why the original comment does not hold

The comment claims personas need distinct instances per team because of "different instructions." In current code, persona `systemPrompt` is persona-scoped (not team-scoped at mint time). `teamId` on the record is deployment provenance / runtime instruction layering — it does not require a new keypair. Re-attaching an existing instance to another channel achieves the same outcome without spawning parallel identities.

## Reproduction (single machine, single community)

1. Create managed instances for personas (e.g. Welcome team: Fizz, Honey, Bumble) and deploy the team to channel A — note instance pubkeys.
2. Deploy the **same team** (or overlapping personas) to channel B via **Deploy team to channel**.
3. `buzz channels members --channel <B>` shows **new bot pubkeys** with the **same display names** as step 1.
4. Repeat across N channels → N duplicate pubkeys per persona name on the relay.

Also: batch deploy within one channel did not refresh reuse context after each attach in `createChannelManagedAgents()`, so later personas in the same batch could still mint duplicates.

## Observed impact

[...]

PR #3906, opened 2026-07-31
fix(desktop): resolve mentions from channel members [issue:#3204]

tcballard:
# Summary

Implements issue #3204 as the next separate contribution in the #3863 handoff workstream.

The mention picker was applying a local-managed-agent gate to every agent candidate, including agents already present in the current channel. That made a remote or otherwise non-local channel member impossible to mention, while unrelated local persona rows could appear instead.

# Scope

## Included

- Removes the local-managed-agent requirement from channel-member mention candidates.
- Keeps the existing invocability and archived-identity checks for non-member agent candidates.
- Preserves channel membership on the candidate, so a channel-member agent is not labeled “not in channel” and can be selected with the correct pubkey.
- Adds a focused E2E regression covering a profile-backed agent that is a member of #general but is not locally managed.

## Excluded

- No new relay/API or Nostr event kind.
- No change to agent deployment or persona-minting behavior (#3639 remains separate).
- No change to the DM recipient picker.
- No command-palette/autocomplete redesign beyond correcting this existing mention candidate source.
- No OEXL-specific code, matching, outcome contracts, payments, ledger state, or settlement semantics.

# Product / Architecture Decisions

- Channel membership remains the authoritative source for candidates already in the open channel.
- Local managed agents and explicitly invocable relay agents retain their existing non-member behavior.
- Existing duplicate-name disambiguation, profile labels, archived filtering, and p-tag extraction are reused unchanged.

# User-Facing Contract

In a channel composer, @ now offers agent identities that are current members of that channel even when their runtime is hosted on another machine or is not represented in this Desktop’s local managed-agent list. Those rows are shown as channel members rather than “not in channel.” Non-member, non-invocable agent rows remain hidden.

# Verification

## Ran

- Focused Biome check on useMentions.ts and mentions.spec.ts
- node --import ./test-loader.mjs --experimental-strip-types --test src/features/agents/lib/agentAutocompleteEligibility.test.mjs (18 passed)
- git diff --check

## Added

- desktop/tests/e2e/mentions.spec.ts: channel-member, non-local agent is visible and not labeled “not in channel.”

The full Playwright browser run and TypeScript build were attempted but could not complete in this environment because the temporary checkout’s Vite/Node dependency loader stalled while reading vite.config.ts; this is an environment/toolchain limitation, not a test assertion failure.

# UI Evidence

This changes autocomplete behavior, not layout or styling. A screenshot/recording could not be captured because the local browser build was blocked by the dependency-loader issue described above; the regression is represented in the focused E2E spec for maintainers to run in CI.

# Review Path

1. desktop/src/features/messages/lib/useMentions.ts
2. desktop/tests/e2e/mentions.spec.ts

# Notes For Reviewer

This is an independent branch and PR, not stacked on the workflow-history or Canvas-history contributions. It deliberately does not address the adjacent persona duplication, DM recipient, or command-palette workstreams.

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -16,7 +16,6 @@ import {
   coalesceAutocompleteCandidatesByKey,
   getMentionableAgentPubkeys,
   getSharedChannelIds,
-  isAgentIdentityInManagedList,
   shouldHideAgentFromMentions,
 } from "@/features/agents/lib/agentAutocompleteEligibility";
 import {
@@ -246,9 +245,6 @@ export function useMentions(
       if (isArchivedDiscovery(pubkey)) {
         return;
       }
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
-        return;
-      }
       if (
         shouldHideAgentFromMentions({
           isAgent: candidate.isAgent === true,
@@ -420,7 +416,6 @@ export function useMentions(
     managedAgentNamesByPubkey,
     managedAgentPersonaIds,
     managedAgentPersonaIdsByPubkey,
-    managedAgentPubkeys,
     managedAgentsQuery.data,
     memberPubkeys,
     members,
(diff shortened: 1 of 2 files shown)


Chessing234:
nice narrow fix. worth a regression case where the channel member is respond_to:allowlist but the viewer is not on the list — autocomplete should still hide them.

tcballard:
@Chessing234 Added this in ba6650ac. The regression seeds mira as a #general channel member with respond_to: allowlist, leaves the mock viewer off the allowlist, and asserts that @mi autocomplete stays hidden. The focused Playwright test passes (1 passed).

Issue #3836, opened 2026-07-31
Keep agent replies to top-level mentions in the main channel timeline


penhollowstudiolabs:
When I mention an ACP agent in a top-level Buzz channel or DM message, the agent’s response is posted only as a threaded reply. It does not appear as a new top-level message in the main timeline. This makes ongoing agent conversations difficult to follow from the channel itself. I have seen top-level agent replies demonstrated elsewhere, but I cannot find a documented setting or released Windows build that enables this behavior.

What I'm trying to see happen:

- A top-level message or mention should receive a top-level agent reply in the same channel.
- A message sent inside a thread should receive a reply inside that thread.

This matches the behavior described in draft PR #3687:
https://github.com/block/buzz/pull/3687

What's actually happening:

A top-level mention is answered using `--reply-to`, and the response appears only in the thread attached to the original message.

I'm using Buzz desktop on Windows and an OpenClaw Gateway integration. I'm also struggling to get Hermes via my VPS to work as well. Clearly, I don't know what I am doing and can't trust the agents' advice. So my questions are:

1. Is top-level agent reply behavior available in any current released Buzz build?
2. Is there an official setting or supported configuration for it?
3. If not, is PR #3687 still the intended fix, and is there an expected release timeline?

I am looking for the supported Buzz behavior—not agent suggestions that keep me going in a loop. The bottom line is that I just want a room for my local OpenClaw and VPS Hermes profiles to talk together.

FabianHertwig:
In my opinion the threaded reply is the correct default behavior. And that is probably very opinionated / user preference. 

And I think you can steer the agent to reply top level.

KritBlade:
This whole thing is about to make me to uninstall Buzz.  When we grab everyone in the channel, I would expect any AI would reply on the main chat, and that's the only way to keep all AI know what's the other AI is talking.

I don't mind if the thread reply can show what the AI is replying (given other AI in the channel can also see what other AI is talking about).  Currently, you have to click on it to show what AI is talking.

sbenedetto:
> And I think you can steer the agent to reply top level.

You can but it keeps forgetting. 

The correct solution is to have a global option for this and let the user decide.

Personally I also find threads very confusing.

mrjohnmac:
Yeah, this IS confusing. I just spent 2 hours messing around with this. Especially in DMs. I expect a reply. Instead, I had to dig into the agent settings panel, and found the replies under "Info". Wrong place.

Issue #3825, opened 2026-07-30
Agent reply visible in Activity log but not displayed in channel/thread


m-faizan-tariq:
## Summary

When an agent (e.g., Fizz, Bumble, Honey) is tagged in a channel message and responds, the reply is sometimes visible in the Activity/Pulse log but does **not** appear in the channel or create a new thread against the original message.

## Observed Behavior

1. User sends a message tagging an agent (e.g., "@Fizz hello") in a channel (e.g., #Welcome)
2. The agent processes the message and generates a reply
3. The reply appears in the **Activity** or **Pulse** log — confirming the agent completed its response
4. However, the reply is **not visible** in the channel
5. No new thread is created against the original message
6. The user has no indication in the conversation view that the agent responded

## Expected Behavior

When an agent replies to a message in a channel, the reply should:
- Appear in the channel as a visible message (or as a threaded reply to the original message)
- Be accessible from the conversation UI without switching to the Activity/Pulse view

## Environment

- **Buzz Desktop version:** 0.5.2 (bundle xyz.block.buzz.app)
- **OS:** macOS (arm64)
- **Agent runtime:** OpenCode 1.18.9-fork
- **Relay:** `wss://kamil22.communities.buzz.xyz`

## Notes

- This is **intermittent** — sometimes replies appear correctly, other times they only show in Activity
- Agent logs show successful initialization and message processing
- No errors are reported in the agent or buzz-acp logs when this occurs
- The issue may be related to how the relay routes agent responses back to the channel vs. storing them only as activity events

## Suggested Investigation

- Check if the agent response event includes the correct channel/thread routing metadata
- Verify that the relay is forwarding the response back to the channel subscribers (not just the Activity feed)
- Check if there's a race condition between the agent's response and the channel subscription state

nagaral:
Confirming this also reproduces on harnesses other than OpenCode — same "shows in Activity, not in channel" symptom, no errors logged, intermittent.

Tried across three different harnesses on the same setup (self-hosted community, local MLX-served model backend via an OpenAI-compatible provider):

- OpenCode — reproduces (matches the original report)
- `buzz-agent` (native harness) — reproduces
- Goose — reproduces

Same failure shape each time: the agent's turn completes successfully and the reply is visible in Activity/Pulse, but no message appears in the channel and no thread is created against the original mention.

[...]

PR #3908, opened 2026-07-31
fix(acp): keep top-level agent replies in channel [issue:#3825]

tcballard:
# Summary

Implements [#3825](https://github.com/block/buzz/issues/3825) and [#3836](https://github.com/block/buzz/issues/3836), as part of the ordered workstream tracked by [#3863](https://github.com/block/buzz/issues/3863).

Agent replies to top-level human mentions were being emitted as threaded replies, leaving the response out of the main channel timeline for users who were watching the channel or activity log. This change keeps the response visible at the channel root while preserving normal threaded conversation behavior.

Adds:
- top-level human mention in a channel -> agent reply is sent without `--reply-to`, so it remains in the channel timeline;
- human mention inside an existing thread -> agent reply anchors to that thread's root;
- setup-mode nudges follow the same top-level/channel-root versus threaded/root behavior;
- prompt guidance and ACP queue regression coverage for the routing contract.

# Scope

## Included

- ACP prompt guidance for top-level channel responses.
- ACP reply-anchor resolution for top-level and threaded human-facing turns.
- Setup-mode nudge placement.
- Regression coverage for top-level, threaded, batched, agent-only, and unknown-identity routing cases.

## Excluded

- No new Nostr event kind.
- No new HTTP API or marketplace-specific protocol.
- No desktop UI, layout, or command-palette changes.
- DM reply behavior and agent-to-agent nesting semantics are unchanged.
- No OEXL matching, payments, acceptance, ledger, settlement, or other marketplace code.

# Product / Architecture Decisions

- A top-level channel response is a channel-root message, so the ACP prompt explicitly instructs `buzz messages send` without `--reply-to`.
- An already-threaded human-facing turn continues to reply to the thread root, keeping the conversation flat at the first reply level.
- Agent-to-agent coordination remains free to use deeper nesting.
- Setup-mode uses the same NIP-10 placement rule as normal ACP replies.
- The existing Buzz message-send and NIP-10 thread path is reused; no parallel publishing mechanism is introduced.

# User-Facing Contract

- Mentioning an agent from a top-level channel message produces a visible top-level response in that channel.
- Mentioning an agent from inside a thread produces a reply in that thread.
- Setup-mode messages follow the same placement rules.

# Verification

## Ran

- `cargo test -p buzz-acp queue::tests --lib` (112 passed)
- `cargo test -p buzz-acp --lib` (649 passed)
- `git diff --check`

## Covered

- Human top-level turns have no forced reply anchor.
- Human threaded turns anchor to the thread root.
- Batched events use the final event's scope without reusing an older thread.
- Agent-only turns remain unanchored.
- DM thread replies retain their existing reply behavior.
- Setup-mode and ordinary ACP prompts use the same channel-root/thread-root distinction.

# Review Path

Suggested review order: `crates/buzz-acp/src/queue.rs` reply-anchor and prompt changes, `crates/buzz-acp/src/setup_mode.rs` nudge placement, then the focused ACP tests.

# Notes For Reviewer

This is a focused current-main contribution for #3825/#3836. It changes channel notice placement and ACP tests only; no desktop UI changed, so no screenshots are required. The earlier draft work on this area remains separate and is not stacked onto this branch.

### DIFF
--- a/crates/buzz-acp/src/base_prompt.md
+++ b/crates/buzz-acp/src/base_prompt.md
@@ -53,7 +53,7 @@ For explicit changes to an existing personal agent, use `buzz agents draft-updat
 
 Use the reply destination supplied in the `[Context]` block for ordinary replies in this turn. Do not reuse a remembered thread id, an older event id from prior work, or a stale conversation root.
 
-For human-facing work, keep the conversation flat and easy to read. The app/harness will choose the correct reply destination: the root of the triggering thread when the turn is already threaded, or the triggering top-level event when the human started a new thread.
+For human-facing work, keep the conversation flat and easy to read. The app/harness will choose the correct reply destination: the root of the triggering thread when the turn is already threaded, or no reply destination for a top-level channel message so the response stays in the channel.
 
 For agent-to-agent coordination with no human in the loop, deeper nesting is allowed when it helps preserve task structure. Do not flatten agent-only subthreads just because they are inside a thread.
 
--- a/crates/buzz-acp/src/queue.rs
+++ b/crates/buzz-acp/src/queue.rs
@@ -1156,19 +1156,18 @@ fn append_reply_instruction(s: &mut String, event_id: &str) {
     ));
 }
 
-/// Append a new-thread reply instruction for a human-facing top-level mention.
+/// Append a channel-root reply instruction for a top-level message.
 ///
-/// The triggering mention has no thread tags, so the agent's reply becomes the
-/// thread root. Anchoring to the triggering event (rather than leaving the
-/// choice open) prevents replying into a stale/unrelated prior thread.
-fn append_new_thread_reply_instruction(s: &mut String, event_id: &str) {
-    s.push_str(&format!(
-        "\nIMPORTANT: This is a new top-level message. For ordinary replies in \
-         this turn, use `--reply-to {event_id}` on `buzz messages send` — the \
-         triggering message is the thread root. Do NOT reply into any other \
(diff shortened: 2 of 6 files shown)


Chessing234:
solid approach for #3825 — one ask: does the activity/pulse log still show the reply as threaded, or does that path also need the same top-level kind so mobile and desktop stay consistent?

Chessing234 (commented):
keeping top-level agent replies on the channel timeline is the high-traffic half of #3825 / #3836. worth a regression test that a kind:9 reply with a root e-tag still lands in the channel list, not only activity — if that's already in the pr, great.

tcballard:
@Chessing234 Addressed in 7b836377. Desktop and mobile activity/inbox coverage now asserts that a top-level kind:9 ACP reply with no NIP-10 e-tags remains a standalone activity row, while e-tagged replies continue to classify as threads. I also extended the channel-window E2E regression to assert that a kind:9 depth-1 broadcast reply carries the root's NIP-10 reply tag plus broadcast:1 and remains in the authoritative channel timeline. The focused desktop unit test, E2E test, Biome check, and Dart formatting pass; the local Flutter test runner stalled before producing output.

kerpopule:
Confirmed against Buzz 0.5.3 with a managed Hermes ACP agent: a top-level channel mention showed 👀/💬 and typing, and the agent log later recorded a completed turn with non-empty final output, but the user watching the mobile channel saw no inline response. Current ACP guidance forces `--reply-to` for top-level mentions, putting the final answer behind a thread instead of the channel timeline. This PR matches the reported visibility failure.

Issue #3841, opened 2026-07-31
Share Compute (mesh-llm) is unavailable in every official Linux build — release and canary


davidglogan:
**Title:** Share Compute (`mesh-llm`) is unavailable in every official Linux build — release and canary

**Body:**

Share Compute fails on Linux with `Couldn't check shared compute: mesh-llm feature not enabled` (the stub error from `desktop/src-tauri/src/mesh_llm_stubs.rs`) on the current `apt`-installed `.deb` (v0.5.2).

Traced this to the release/canary CI definitions rather than a local config issue:

- `mesh-llm` is an opt-in Cargo feature (`desktop/src-tauri/Cargo.toml`): `default = ["system-keyring"]`, with `mesh-llm` requiring an explicit `--features mesh-llm` flag.
- `.github/workflows/release.yml`: only the macOS job (`runs-on: macos-latest`, line ~181) passes `--features mesh-llm`. `release-linux` (line ~624, `pnpm tauri build --verbose --ci --bundles deb,appimage ...`) does not. Neither does `release-windows`.
- `.github/workflows/linux-canary.yml` (line ~173) also builds without `--features mesh-llm`.

So there is currently no official Linux artifact — stable or canary — that ships Share Compute at all, even though the feature and its UI are fully present and reachable in the Linux build (it just always errors).

**Ask:** either add a Linux release/canary variant built with `--features mesh-llm` (behind a separate bundle name, like the mac path already does), or gate the Share Compute UI behind a build-time flag so Linux users don't hit a dead-end error screen for a feature their build can never support.

Environment: Ubuntu (ai-local), `buzz` 0.5.2 via apt, binary `/usr/bin/buzz-desktop`.

davidglogan:
Confirmed: Share Compute works correctly on the macOS build of the same version (0.5.2), which matches the CI evidence above (macOS release job is the only one with `--features mesh-llm`).

PR #3909, opened 2026-07-31
fix(mesh): stop restarting a busy or loading shared-compute node

micspiral:
## Problem

Sharing compute with a large model (e.g. `gemma-4-26B`) put the desktop app into a **restart loop**: toggle Share → app appears to "download" / stall → the whole app restarts → repeat. Small models (E4B) were unaffected, which made it look model-specific and flaky.

It is not model-specific and not flaky. It is a **false-positive liveness check**.

## Root cause (proven by black-box measurement)

A `serve` node's OpenAI ingress (`:9337`) serializes **all** HTTP — including the `/v1/models` liveness probe — behind the current in-flight inference. It is *also* HTTP-unresponsive during model load and package-layer download. In every one of those phases the node is alive and progressing, but it cannot answer an HTTP probe.

Measured on a standalone `gemma-4-26B` node (randomized ~30k-token prompt, cache-miss):

| during one ~30s inference | result |
|---|---|
| concurrent `GET /v1/models` | **27.0s**, then 200 |
| concurrent small `/chat/completions` | **28.8s**, then 200 |
| `tcp_connect(:9337)` throughout | **~0ms** |

Both HTTP calls simply queued behind the turn; TCP kept accepting instantly. A probe with any timeout shorter than the turn reads the node as dead.

Buzz then acted on that false "dead" reading in two places, **both restart paths added in #2823**:

1. **Ingress watchdog** — after 2 consecutive `/v1/models` timeouts, evicts the node; for a serve node eviction means `app.request_restart()`. Two dead probes landing inside a prefill window → restart loop.
2. **Start / restore paths** — on a `wait_for_mesh_inference` timeout, `stop()` the node and (fresh start) `request_restart()` the app "to guarantee cleanup" — even though the node was still loading weights or downloading layers. This is the exact line in the incident log: `started node failed inference readiness … Buzz is restarting`.

## Fix

Treat a **bound TCP port as alive**. Death has exactly one unambiguous signal: a *closed* port.

- **Watchdog** (`recovery.rs`): only `PortClosed` may evict. A bound-but-HTTP-unresponsive `Unhealthy` port is never evicted, at any probe streak or urgency. Closed-port eviction is unchanged.
- **Start / restore** (`commands/mesh_llm.rs`): install the runtime **before** probing readiness (so it is always tracked by `AppState` and can never be orphaned — which is what the restart was guarding against), and on a readiness timeout **leave it warming up** instead of stopping/restarting. Launch-restoration stays disarmed until real inference is confirmed, so a genuinely broken start is retried next launch rather than silently disabling Share Compute.

### What this deliberately does *not* do

Detecting a node that is bound-but-internally-wedged needs a liveness signal that bypasses the inference lock. There is none today, so this fix cannot distinguish "wedged" from "busy" and errs toward not restarting. That gap is a mesh-llm bug, filed upstream: **Mesh-LLM/mesh-llm#1126** (lock-free `/live`+`/ready` on the ingress). A follow-up here can consume it once it lands.

## Tests

- Watchdog never evicts a bound/busy port at any probe streak or urgency (the regression).
- Closed-port eviction still fires (dead listener still reclaimed).
- Black-box: a listener that accepts TCP then stalls HTTP classifies as `Unhealthy`, not `PortClosed`.
- **Mutation-proven**: reverting the eviction rule to the old count-based logic fails the busy-node test.

`cargo test` (desktop, `--features mesh-llm`) green, fmt + clippy clean.

## Not covered here

The intermittent nature means I could not force the live loop deterministically on a warm machine; the proof is the measured serialization + the mutation-proven unit/black-box tests. Live behaviour (app no longer restarts while a 26B node loads/serves) still merits a manual check before merge.


micspiral:
@tlongwell-block tagging you for review ahead of the next release.

This is the fix for the shared-compute **restart loop with large models** (e.g. `gemma-4-26B`) — the one that looked like a stuck "download" and bounced the whole app. Root cause is a false-positive liveness check: a `serve` node's ingress (`:9337`) is legitimately HTTP-unresponsive while it loads weights, downloads layers, or serves an in-flight turn (it serializes all ingress HTTP — including the `/v1/models` probe — behind the current inference). Measured on a 26B node: a concurrent `/v1/models` took **27s** while `tcp_connect` stayed **~0ms**. Both restart paths (the ingress watchdog and the start/restore readiness probe, both from #2823) read that busy-but-alive state as "dead" and restarted.

Fix: treat a **bound TCP port as alive** — only a *closed* port evicts; on a readiness timeout the node is installed (so never orphaned) and left to warm up rather than stopped/restarted.

CI is green. The complementary upstream gap (no lock-free liveness endpoint that bypasses the inference lock) is filed as Mesh-LLM/mesh-llm#1126, with a mesh-side patch against `main` in progress.

Chessing234:
good root-cause writeup. can you pin a test that a long mesh load stays in loading without triggering the restart path? that was the false positive.

Chessing234 (commented):
this matches the failure mode in #3841-adjacent mesh work — readiness timeout ≠ dead node. installing into AppState before the probe is the right orphan fix. one ask: the restore path now leaves  even when inference never becomes ready; confirm the status publish path surfaces "warming up" distinctly from "sharing", otherwise the UI can look stuck-on while still loading weights.

micspiral:
@tlongwell-block PR is out of draft and ready for review. Two updates since I first tagged you:

**1. Fixed a regression I'd introduced.** The first cut gated the persisted `enabled` config on the inference-readiness probe, so a >120s cold start (big model load + layer download) would serve fine that session but come back with Share Compute OFF next launch. Now the fresh-start path arms `enabled=true` as soon as the runtime is installed and tracked — mirroring the restore path — and the readiness probe is purely informational. Independent review confirms the persistence boundary is correct (`start()`+`status()` success = ports bound = real; inference readiness is transient, not config validity). Commits `a281977c2` (fix) + `f5d8703bf` (comment cleanup).

**2. The upstream mesh fix is NOT a quick companion PR — heads up.** I built the proposed lock-free `/health` endpoint on mesh-llm `main` and tested it live against a real prefill: **it stalls identically to `/v1/models`** (3s timeout, ~21s, while `tcp_connect` stays 0.2ms). The head-of-line block is *upstream of request dispatch*, not in the handlers, so a new route on the same ingress server doesn't help. Details in Mesh-LLM/mesh-llm#1126. That's exactly why this Buzz PR keys on **TCP-connect (bound = alive)** rather than any HTTP probe — it's the only signal that survives a turn. This PR stands on its own; it doesn't depend on the mesh fix landing.

Verified: 83 desktop tests (incl. mutation-proven eviction logic + a black-box accept-then-stall probe), fmt + clippy clean, and the core premise confirmed on a live 26B/E4B node.

micspiral:
@Chessing234 good call — done in `d9c11e44d`. That was the right thing to pin: the old tests asserted the *rule* (`Unhealthy` never evicts), not the *scenario* that actually bit us.

`long_model_load_never_reaches_the_restart_path` walks ~5 minutes of watchdog passes at the 15s base interval with the ingress bound-but-unresponsive — i.e. a big model loading weights and pulling package layers — and asserts the eviction gate stays shut the entire way, for both `Watchdog` and `Foreground` urgency. For a serve node a single `true` there is a whole-app restart, which is the loop we were seeing.

Two things the test also nails down so it can't pass for the wrong reason:
- it asserts the streak really does keep climbing (1, 2, 3, … 20), so the pass is the rule refusing to act rather than the counter quietly resetting;
- it asserts the streak clears once the ingress answers again.

Mutation-checked: reverting `should_evict_after_probe` to the pre-fix count-based logic fails it immediately at `streak=2, minute=0` — the exact false positive.

Worth noting the empirical side too, since it's what the whole change keys on: on a live 26B node during a real turn, `tcp_connect` stayed at 0.2ms while *every* HTTP endpoint (including a lock-free `/health` I built into a patched mesh node) timed out for ~21s. So a bound port genuinely is the only trustworthy "alive" signal here — HTTP silence is normal backpressure, not death. Upstream detail in Mesh-LLM/mesh-llm#1126.

micspiral:
@tlongwell-block heads-up on release timing: **v0.5.3 (cut today 20:25Z) still contains this bug.** I checked the tag directly:

- `desktop-v0.5.3:.../mesh_llm/recovery.rs` still has `|| consecutive >= DEAD_PROBE_EVICT_THRESHOLD`, so two consecutive `Unhealthy` probes evict — and for a serve node eviction is a whole-app restart.
- `desktop-v0.5.3:.../commands/mesh_llm.rs` still has the `inference never became ready … Buzz is restarting to guarantee cleanup` path.

So anyone on v0.5.3 sharing a large model (26B/31B) can still hit the restart loop while the node is loading weights, pulling package layers, or just busy serving a turn — all three leave the ingress bound but HTTP-unresponsive, which the current code reads as death.

This PR is green (26/26), mergeable, and doesn't touch anything your reply-guard work (#3763, also in v0.5.3) changed — no overlap in files. Would be good to get it into the next release.

micspiral:
@Chessing234 you able to take another look at this? (trying it locally and seems solid)

tlongwell-block (approved):
Reviewed at `0cad23db680ac8b42b625954e8a31adb695d2a8a`.

No blocking findings. The safety boundary is right: `Unhealthy` remains a bound runtime and cannot trigger serve-node process restart; only `PortClosed` can cross the eviction gate. Installing the runtime in `AppState` before the inference probe also removes the orphan-cleanup justification for restarting on readiness timeout, while the disarmed→armed checkpoint transition remains ordered around successful runtime installation.

I also traced the earlier “warming vs sharing” question. Serve status stays `starting` while `models_from_status_payload` is empty (`mesh_llm/mod.rs:830-841`), and the settings card renders that state separately from `running`/“Sharing” (`MeshComputeSettingsCard.tsx:593-619`). Status publication does not advertise a serve target until the SDK reports ready models.

Verification: exact-head required CI is green, including Desktop Core, Desktop Build (macOS), Windows Rust, Rust Lint, and desktop E2E. I attempted the full local compiled-feature desktop suite at this SHA, but the machine exhausted disk while compiling `aws-lc-sys`; that was an environment failure (`clang: unable to make temporary file: No space left on device`), not a test failure.

Scores: Minimalness 9/10 · Elegance 9/10 · Correctness 9/10.

tlongwell-block (approved):

PR #3911, opened 2026-07-31
feat(desktop): auto-mention agents when replying in their threads

Fulki88:
<p align="center">
  <strong>Reply to an agent in its thread, and the agent actually answers.</strong>
</p>

---

## What is this, really?

Replying in a thread to a message written by an agent did nothing. The ACP harness only wakes on events carrying the agent's `p` tag — the relay subscription filters on `#p` (`crates/buzz-acp/src/relay.rs`) and `filter::match_event` re-checks it — while the desktop composer derived mention tags exclusively from typed `@` text. So a plain reply under an agent's message never reached the agent, and the thread went quiet unless you remembered to re-type `@AgentName` every time.

Now the composer auto-inserts a **visible, removable `@AgentName` mention chip** whenever the message you're replying to was authored by an agent. The chip is real composer text: it rides the existing send pipeline (mention extraction → `p` tag → both harness gates → managed-agent wake-up), and deleting it before sending opts out of waking the agent for that reply.

---

## Before · After

<table>
  <tr>
    <td width="50%" valign="top" align="center">
Screenshot 2026-07-31 at 14 25 47
      <br>
      <sub><strong>Before.</strong> Replying under the agent's message sends a plain reply — no mention, no answer.</sub>
    </td>
    <td width="50%" valign="top" align="center"> 
     Screenshot 2026-07-31 at 18 44 45
      <br>
      <sub><strong>After.</strong> The @mention chip is pre-filled in the thread composer; sending wakes the agent. Delete the chip to opt out.</sub>
    </td>
  </tr>
</table>

---

## How it works

- **`useReplyTargetAgentMention`** (new composer hook, modeled on the persistent-audience hydration hook) inserts the chip via the existing `insertResolvedMention` / `replacePlainTextRange` machinery. It only ever writes into an empty composer or over its own untouched chip — user-authored text and restored drafts are never touched.
- **`resolveReplyTargetAgent`** (new pure helper) decides "is this reply target an agent" with the same layered detection the thread panel already uses: message `isAgent` flag → known-agent pubkey set → profile flag. It matches on the agent *identity* pubkey (not the raw signer key), which is what `event_mentions_agent` gates on.
- Wired in **both reply surfaces**: the thread panel (effective target = explicit reply target, falling back to the thread head) and the Home inbox reply pane. DMs are skipped — they already `p`-tag every participant.
- **Opt-out sticks per target**: removing the chip marks that reply target dismissed, so it isn't re-inserted until the target changes or a reply is sent.
- **No phantom drafts**: an untouched auto-chip persists as an empty draft, so opening an agent thread and navigating away doesn't strand a `@Agent ` draft (caught by the drafts e2e suite during development).
- `useComposerAutoSubmit` is a pure extraction from `MessageComposer` to stay under the file-size ratchet.

---

## Test plan

- [x] Unit: `replyTargetAgentMention.test.mjs` — agent detection via message flag / known set / profile, human → null, pubkey normalization, display-name fallbacks (8 tests)
- [x] E2E: `reply-agent-automention.spec.ts` (registered in the smoke project) — chip auto-inserts on agent threads, sent payload carries the agent pubkey in `mentionPubkeys`, deleting the chip opts out, human-authored threads get no chip
- [x] Regression: full desktop unit suite (3878 tests), `pnpm check` (biome + file-size + px-text + pubkey guards), `tsc --noEmit`, and ~215 e2e tests across messaging / mentions / persistent-agent-audience / threads / drafts / channels / inbox
- [x] Manual: verified live against the dev server — opening a bot-authored thread pre-fills the chip, and the sent reply event carries the agent's `p` tag

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -65,6 +65,7 @@ export default defineConfig({
         "**/mentions.spec.ts",
         "**/team-mentions.spec.ts",
         "**/persistent-agent-audience.spec.ts",
+        "**/reply-agent-automention.spec.ts",
         "**/relay-reconnect.spec.ts",
         "**/relay-reconnect-affordance.spec.ts",
         "**/workflows.spec.ts",
--- a/desktop/src/features/home/ui/InboxDetailPane.tsx
+++ b/desktop/src/features/home/ui/InboxDetailPane.tsx
@@ -395,10 +395,12 @@ function InboxMessageDetailPane({
       },
       agentPubkeys,
       profiles,
+      currentPubkey,
     );
   }, [
     agentPubkeys,
     capturedDefaultParentId,
+    currentPubkey,
     displayMessages,
     isDirectMessage,
     item,
--- a/desktop/src/features/home/ui/InboxDetailPane.tsx
+++ b/desktop/src/features/home/ui/InboxDetailPane.tsx
@@ -31,6 +31,7 @@ import {
   isWithinGroupingWindow,
 } from "@/features/messages/lib/messageGrouping";
 import { orderMentionPubkeysByText } from "@/features/messages/lib/orderMentionPubkeys";
+import { resolveReplyTargetAgent } from "@/features/messages/lib/replyTargetAgentMention";
 import { getThreadReference } from "@/features/messages/lib/threading";
 import { normalizePubkey } from "@/shared/lib/pubkey";
 import { MessageComposer } from "@/features/messages/ui/MessageComposer";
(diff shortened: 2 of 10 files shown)


Chessing234:
auto-mention on reply is handy. please skip it when the parent author is the current viewer so we dont @ourselves.

Fulki88:
@Chessing234 good catch, thanks! 🙏 In practice the chip only triggers for agent-authored parents so replying to yourself was already a no-op for normal accounts — but you're right that the guard should be explicit, not incidental. Pushed 0f92d08: `resolveReplyTargetAgent` now takes the viewer's pubkey and bails when the parent author is the current viewer, so we never @ ourselves even if the signed-in identity happens to be agent-flagged (e.g. running the app under an agent key). Covered with unit tests for the self/other-viewer cases too.

Chessing234 (commented):
auto-mentioning the agent you're replying to removes a real friction for thread handoffs. resolveReplyTargetAgent skipping dms is correct. one edge: if isAgent is undefined on the synthetic inbox target, make sure agentPubkeys still classifies the author so human replies don't get a spurious chip.

Fulki88:
@Chessing234 verified the `isAgent: undefined` synthetic-target edge — the resolver already fails safe: only an explicit `true` short-circuits, so `undefined` falls through to `agentPubkeys.has(normalizedPubkey)` and then the profile flag. The set is normalized at every source (`mergeKnownAgentPubkeys` plus the `inboxAgentPubkeys` memo in `HomeView`), and the resolver normalizes the target pubkey before the lookup, so human authors resolve to `null` (no spurious chip) and agent authors are still classified via the set. The existing unit tests already cover both directions (agent-with-undefined-flag resolves via the set; human with a populated set resolves to null), so no code change needed.

PR #3912, opened 2026-07-31
fix: mobile composer, agent draft allowlist, git push DX, duplicate-session warning

### LINKED ISSUE #3725: Mobile channel composer: make Enter insert a newline instead of sending
**Motivation**
In the Buzz mobile channel composer, the keyboard Enter/Return action sends the message. This makes multiline messages difficult to write, even though the composer supports up to five lines. The visible send button already provides a clear way to send.

**Proposed solution**
Make Enter/Return insert a newline in the mobile channel composer and keep the visible send button as the send action. Optionally add a preference for users who want Enter-to-send, with newline as the default.

**Alternatives considered**
Keep Enter-to-send and add a separate newline affordance. This is less consistent with multiline mobile text input and adds UI complexity.

**Additional context**
The current channel composer uses `TextInputAction.send` and calls `onSend()` from `onSubmitted` in `mobile/lib/features/channels/compose_bar/layout.dart`. No matching open or closed issue was found.

### LINKED ISSUE #3792: feat(cli): owner-reviewed agent drafts cannot grant "Selected people" access — respond-to is limited to owner-only / anyone
## Problem

Desktop already models three agent access levels — **Only me**, **Selected people** (per-user allowlist), and **Anyone** (see `desktop/src/features/agents/lib/respondToAllowlist.ts` and the access warnings reworked in #2561). But the owner-reviewed draft path that agents use from chat (`buzz agents draft-update`) can only express two of them:

```rust
// crates/buzz-cli/src/agent_management.rs
let respond_to = optional(draft.respond_to, "respond-to")?;
if respond_to
    .is_some_and(|value| value != "owner-only" && value != "anyone")
{
    // "respond-to must be owner-only or anyone"
}
```

`UpdateAgentDraft` carries `respond_to: Option<String>` and has no field for an accompanying pubkey list, so an allowlist grant cannot round-trip through the observer-frame draft flow at all.

## Real scenario (today)

Two people share a channel, each with their own agents. Owner A wanted to let exactly one teammate invoke their agent and asked the agent in chat to open an access-change draft. The agent could only offer **anyone** — correctly reporting that the draft path supports nothing narrower — so the owner had to choose between opening the agent to every member of every channel it is in, or nothing. The narrower grant the platform already supports was unreachable from chat.

## Proposal

Extend the draft request and desktop review payload to carry the allowlist mode:

- Accept `respond-to: allowlist` plus a `respondToAllowlist: string[]` of pubkeys.
- Normalize entries to lowercase 64-char hex before persistence (accept `npub` as input and decode).
- Require a non-empty list when mode is `allowlist`; apply mode and list atomically; malformed or unsupported input fails closed.
- The Desktop review dialog should show each granted identity as resolved name + pubkey fingerprint, since display names are discovery aids, not authorization identities — enforcement must compare the signed event author pubkey.
- Unrelated edits from clients that predate the field must not clear a stored allowlist.

Related access-policy discussions: #2469, #2475, #2497.

## Why it matters

The chat-mediated draft flow is how mixed human/agent teams actually reconfigure agents mid-conversation. As long as that path can only express the two extremes, "share with one trusted person" silently degrades to "share with everyone", which is exactly the risk the #2561 warnings exist to flag.

### LINKED ISSUE #3795: docs(presence): 90s TTL is documented as 3× a 30s heartbeat, but the agent harness heartbeats at 60s (real margin 1.5×)
## Summary

Three places document the presence TTL as "3× the 30s heartbeat." The **agent harness** publishes its presence heartbeat every **60s**, so for agents the real flap margin is **1.5×**, not 3× — one missed heartbeat leaves 30s of slack, two are fatal. The desktop client *does* heartbeat at 30s, so the comments are correct for one publisher and wrong for the other, and a fourth doc comment states the TTL itself incorrectly.

Found while truth-auditing presence claims for the remote-agents work (`docs/remote-agents.md`, PR #3748). This is a comment/doc-accuracy bug with a real behavioral question behind it — not a functional break — but the wrong margin is the kind of number someone sizes a timeout against later.

All line numbers below are at `b4f4ed1a67ce340d9709ca32f3b47dd09cd25dea` on `origin/main`'s history, verified with `git rev-parse HEAD` in the checkout that produced them.

## The actual values

| What | Where | Value |
|---|---|---|
| Redis TTL | `crates/buzz-pubsub/src/presence.rs:16` | `PRESENCE_TTL_SECS = 90` |
| **Agent** heartbeat | `crates/buzz-acp/src/lib.rs:1583` | `Duration::from_secs(60)` |
| **Desktop** heartbeat | `desktop/src/features/presence/hooks.ts:18` | `PRESENCE_HEARTBEAT_INTERVAL_MS = 30_000` |

So the same 90s lease is renewed by two publishers at different rates. 90/30 = 3× for desktop; 90/60 = **1.5×** for agents.

## Four wrong statements

1. `crates/buzz-pubsub/src/presence.rs:4-5` — "TTL is 3x the 30s heartbeat interval so a single missed heartbeat doesn't cause presence flap."
2. `crates/buzz-pubsub/src/presence.rs:15` — "3x the 30s heartbeat — single missed heartbeat won't cause presence flap."
3. `ARCHITECTURE.md:450` — "90-second TTL (3× the 30-second heartbeat interval). Single missed heartbeat does not cause presence flap."
4. `crates/buzz-pubsub/src/lib.rs:331` — "Set presence with **60s TTL**. Call on connect and every 30s heartbeat." The TTL is 90s, not 60s. This one is wrong about the constant it delegates to, in the direction that makes the margin look worse than it is — likely a stale survivor of an earlier pair of numbers.

Note that 1–3 are still literally true *of the desktop client*. The claim they make that fails is the safety conclusion — "a single missed heartbeat won't cause flap" is a promise about all presence publishers, and it is 30s of slack away from being true for agents.

## Why the margin matters, not just the comment

For an agent: last successful publish at t=0, next attempt at t=60. If that publish fails or is delayed past t=90 — relay reconnect with backoff (`ARCHITECTURE.md:448`: 1s→30s exponential, resets only on clean stream end), rate-limit skip, a slow `SET` — presence expires and the agent goes dark while alive. The retry isn't until t=120.

The harness heartbeat also fires from the main `select!` loop (`crates/buzz-acp/src/lib.rs:2301-2320`), and `interval_at` starts the first tick at `now + 60s` rather than immediately; startup is covered separately by the eager publish at `:1510`, so this is about steady state, not boot.

I did **not** find field reports of spurious agent-offline flaps, and I did not go looking in logs — so treat the impact as reasoned, not observed. #877 ("fix(presence): clear on disconnect, fix heartbeat/TTL", `CHANGELOG.md:1198`) is prior art on this exact pair of numbers and probably worth reading before choosing.

## The decision this needs (why I'm filing rather than fixing)

Which number is wrong is a product/ops ruling, not a comment edit:

- **(a) Comments were right, the interval drifted.** Take the agent heartbeat to 30s, restoring a uniform 3× margin. Cost: 2× presence publish volume from agents.
- **(b) 60s is intentional** (agents are chattier per-publish than a browser tab and the relay pays for it). Then fix all four comments to state the real margin per publisher, and say plainly that agents tolerate zero missed heartbeats with 30s to spare.
- **(c) Decouple.** Derive the TTL from the publisher's declared interval instead of hardcoding a constant that two callers renew at different rates. Most correct, most work, and probably overkill for two publishers.

My weak preference is (b) plus a one-line note at the constant naming both publishers — the invariant worth documenting is "TTL ≥ 1.5× the *slowest* publisher's interval," which makes the next divergence visible instead of silent. But I have no evidence about relay publish cost, and whoever owns relay load should overrule me if 30s is cheap.

Whichever lands, the fix should make the two intervals reference each other, because the failure mode here is that they can drift independently and only a comment notices.

### LINKED ISSUE #3802: Hosted relay git push limited to ~5 MiB packs (code default 500 MB); over-limit and chunked pushes fail opaquely
## Summary

The hosted relay at `communities.buzz.xyz` accepts git pushes only up to ~5 MiB of pack body, which blocks seeding any real-world repo (ours is a ~210 MiB history). The relay code's own default for `BUZZ_GIT_MAX_PACK_BYTES` is 500 MB (`crates/buzz-relay/src/config.rs`), so the hosted deployment appears to run ~1% of stock. Requesting either a raise on hosted community relays or documentation that this is the intended hosted tier limit. Separately, the failure modes for oversized/large pushes are opaque enough that they cost us a full diagnostic session — two DX issues below.

All findings verified live on Buzz 0.5.2 against `firc.communities.buzz.xyz`, 2026-07-30, with NIP-98 auth working and small pushes succeeding throughout.

## 1. Hosted pack limit ~5 MiB (config request)

With client buffering fixed (see item 2), pushes of incompressible test blobs:

| Pack size | Result |
|---|---|
| ~5.0 MiB | success (4.5 s) |
| ~5.45 MB | TLS abort mid-upload (`LibreSSL bad record mac` / `broken pipe`) |
| 6, 8, 9, 10, 20, 25, 30 MB | same failure |
| 201.49 MiB real snapshot (5,526 objects) | HTTP 401 / curl 56 in ~2 s |

A single 25.4 MB tracked blob fails on its own, so multi-push chunked seeding cannot work around it — git can't split one blob. Cloudflare fronts the relay and is not the binding limit (its body allowance is far higher); enforcement matches `git_max_pack_bytes` in `api/git/transport.rs`.

**Ask:** raise `BUZZ_GIT_MAX_PACK_BYTES` on hosted community relays toward the code default (500 MB), or publish the hosted limit so operators can plan around it.

## 2. DX: pack over limit dies as a TLS abort, not a clean error

An over-limit push terminates mid-body with `curl 55/56, LibreSSL bad record mac` — no HTTP status reaches the client. A clean `413` (or a sideband error naming the limit) would have turned a multi-hour diagnosis into one error message.

## 3. DX: chunked Transfer-Encoding push fails with instant 401

git switches to chunked transfer for any push body over `http.postBuffer` (default 1 MiB). Against the NIP-98-authenticated endpoint this fails instantly with `HTTP 401 / curl 56` — the auth challenge can't be satisfied with a non-replayable chunked body. Workaround is `git config http.postBuffer 536870912`, verified (same 5 MB pack: fails chunked, lands buffered). Worth documenting prominently, or handling chunked bodies server-side, since every stock git client over 1 MiB hits this before ever seeing the real limit.

(Minor, observed while probing: sub-1MB requests intermittently drew `403` with a `__cf_bm` cookie — Cloudflare bot management challenging git's HTTP client. Mentioning it since it makes the failure surface look non-deterministic to operators.)

### LINKED ISSUE #3832: Same agent key running in two places — both reply, nothing detects it
I moved my three agents from the desktop app to a VPS so they'd stay up with my laptop closed. Same keys, same auth tags, just a different machine running `buzz-acp`.

The problem is that both installs stay live. Buzz Desktop starts a local agent when I send a message — `start_on_app_launch` is `false`, but sending a message is enough of a trigger — so the same identity ends up running on my laptop and on the server at once. Both receive the mention. Both take a turn. Both reply.

Nothing anywhere notices. No warning in the app, no relay-side complaint, no log line on either side saying another session holds this key. The two harnesses are completely blind to each other, and since they're separate sessions with separate context they can answer the same question differently.

This confused me for hours in a way that's worth describing, because the two problems compounded. My server agents were broken (see the other issue I'm filing) but my laptop was silently answering for them, so everything looked fine from my laptop and looked broken from my phone. I assumed the clients differed. They didn't — I was just reaching different machines depending on whether the desktop app happened to be open.

I get that the current model is one agent, one owner, one machine, and `backend: {"type": "local"}` in the config suggests other backends are planned. But two homes is easy to reach by accident even without a server — a laptop and a desktop would do it.

Some things that would help:

- The relay noticing two sessions authenticated for the same agent pubkey and rejecting or warning on the second
- Something client-side saying "this agent is already running elsewhere" instead of quietly starting another copy
- A documented way to say "this machine is not this agent's home" — I couldn't find one short of toggling `is_active`, and I wasn't sure that was the intended control

Related: #2648 (duplicate identities from a second install — similar area, but that's two identities rather than one identity twice), #2859 and #3554, which look like they're heading toward the server-hosted case.

Chessing234:
## Summary
Stacked fixes across mobile, CLI/desktop agent drafts, relay auth, and git DX — **11 commits** in one PR:

1. **#3725** — mobile channel composer: Enter inserts a newline; send stays on the button
2. **#3792** — CLI `draft-update` supports `--respond-to allowlist` + `--respond-to-allowlist`
3. **#3792** — Desktop round-trips allowlist (stop clearing Selected people on draft apply)
4. **#3832** — Relay NOTICE + warn log when a second session auths as the same pubkey
5. **#3832** — ACP README: remote-only agents / duplicate-key hazard
6. **#3802** — Clearer HTTP 413 for over-limit git packs + hydrate ResourceLimit detail
7. **#3802** — `.env.example` + git-credential-nostr troubleshooting for postBuffer / pack limits
8. **#3815** — Startup log distinguishes invite-bundle-only vs full SPA (`BUZZ_SERVE_GIT_WEB_GUI`)
9. **#3795** — Cross-ref 60s/180s presence invariant at the ACP heartbeat
10–11. Follow-up tests/docs for the git 413 path

Skipped issues that already have open third-party PRs (#3816, #3857, #3865, #3849, #3885, #3856, #3847).

## Test plan
- [x] `cargo test -p buzz-cli --lib agent_management`
- [x] `node --test desktop/src/features/agents/agentManagement.test.mjs`
- [x] `cargo test -p buzz-relay --lib git_body_limit_message_names_env_var`
- [ ] Manual: mobile composer Enter → newline; send button still sends
- [ ] Manual: second `buzz-acp` with same key → NOTICE in logs
- [ ] Manual: over-limit `git push` → 413 naming `BUZZ_GIT_MAX_PACK_BYTES`

Closes #3725
Closes #3792
Closes #3832
Closes #3802
Closes #3795

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -143,8 +143,13 @@ BUZZ_PUSH_ENABLED=false
 # Root directory for ephemeral Git workspaces and the disposable pack cache.
 # Default: ./repos (relative to CWD).
 # BUZZ_GIT_REPO_PATH=./repos
-# BUZZ_GIT_MAX_PACK_BYTES=524288000
+# BUZZ_GIT_MAX_PACK_BYTES=524288000   # default 500 MiB on self-hosted relays
 # BUZZ_GIT_MAX_REPO_BYTES=1048576000
+# Git clients: pushes >1 MiB use buffered bodies by default. Set globally:
+#   git config --global http.postBuffer 536870912
+# Without that, chunked Transfer-Encoding breaks NIP-98 auth (instant 401).
+# Over-limit packs should receive HTTP 413 naming BUZZ_GIT_MAX_PACK_BYTES
+# (not a TLS abort / broken pipe mid-upload).
 # Process-local immutable pack/index cache. Zero disables retention.
 # BUZZ_GIT_PACK_CACHE_PATH=./repos/.pack-cache
 # BUZZ_GIT_PACK_CACHE_MAX_BYTES=5368709120
--- a/.env.example
+++ b/.env.example
@@ -75,8 +75,13 @@ RELAY_URL=ws://localhost:3000
 # Root directory for ephemeral Git workspaces and the disposable pack cache.
 # Default: ./repos (relative to CWD).
 # BUZZ_GIT_REPO_PATH=./repos
-# BUZZ_GIT_MAX_PACK_BYTES=524288000
+# BUZZ_GIT_MAX_PACK_BYTES=524288000   # default 500 MiB on self-hosted relays
 # BUZZ_GIT_MAX_REPO_BYTES=1048576000
+# Git clients: pushes >1 MiB use buffered bodies by default. Set globally:
+#   git config --global http.postBuffer 536870912
+# Without that, chunked Transfer-Encoding breaks NIP-98 auth (instant 401).
+# Over-limit packs should receive HTTP 413 naming BUZZ_GIT_MAX_PACK_BYTES
+# (not a TLS abort / broken pipe mid-upload).
 # Process-local immutable pack/index cache. Zero disables retention.
 # BUZZ_GIT_PACK_CACHE_PATH=./repos/.pack-cache
 # BUZZ_GIT_PACK_CACHE_MAX_BYTES=5368709120
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -44,6 +44,19 @@ BUZZ_RELAY_PRIVATE_KEY=<relay signing key> \
 
 > **Running multiple agents?** Mint a separate keypair for each. Every agent needs its own identity.
 
+
+## Remote-only agents (avoid duplicate replies)
+
+If an agent runs on a VPS or another machine, **do not** also start it from Buzz Desktop
+on a laptop with the same agent key. Both harnesses subscribe to the same mentions and
+will both reply (#3832).
+
+- On machines that should not host the agent: keep it inactive in Desktop, or avoid
+  message-triggered starts / mention wake for that identity.
+- When a second harness connects, the relay emits a NOTICE warning about duplicate sessions
+  (also logged by buzz-acp at warn).
+
+
 ## Channels
 
 The harness discovers channels by querying the relay with the agent's authenticated identity.
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -1580,6 +1580,9 @@ async fn tokio_main() -> Result<()> {
     let mut heartbeat_in_flight = false;
 
     let mut presence_heartbeat = if config.presence_enabled {
+        // Renew presence every 60s. Relay TTL is buzz_pubsub::PRESENCE_TTL_SECS
+        // (180s = 3× this interval). Desktop uses the same 60s cadence — see
+        // desktop/src/features/presence/lib/presence.ts and ARCHITECTURE.md.
         let interval = Duration::from_secs(60);
         Some(tokio::time::interval_at(
             tokio::time::Instant::now() + interval,
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -2819,6 +2819,9 @@ async fn tokio_main() -> Result<()> {
     let mut heartbeat_in_flight = false;
 
     let mut presence_heartbeat = if config.presence_enabled {
+        // Renew presence every 60s. Relay TTL is buzz_pubsub::PRESENCE_TTL_SECS
+        // (180s = 3× this interval). Desktop uses the same 60s cadence — see
+        // desktop/src/features/presence/lib/presence.ts and ARCHITECTURE.md.
         let interval = Duration::from_secs(60);
         Some(tokio::time::interval_at(
             tokio::time::Instant::now() + interval,
--- a/crates/buzz-cli/src/agent_management.rs
+++ b/crates/buzz-cli/src/agent_management.rs
@@ -36,6 +36,8 @@ pub struct UpdateAgentDraft {
     pub model: Option<String>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub respond_to: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub respond_to_allowlist: Option<Vec<String>>,
 }
 
 #[derive(Debug, Clone, Serialize)]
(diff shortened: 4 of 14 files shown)


Chessing234:
Rebased with `Signed-off-by` so DCO Check is green.

Chessing234:
Rebased onto latest `upstream/main` with `Signed-off-by` on all commits and force-pushed. CI should re-run shortly — still ready for @block/buzz-oss-team review when you have bandwidth.

Chessing234:

Chessing234:
multi-issue batch — if review is blocked on scope, i can split mobile composer / draft allowlist / git push dx into separate tips.

nathansmithopenclaw-alt:
Thanks for carrying the selected-people round trip for #3792. We have overlapping
fork work on this seam and do not plan to open a competing allowlist PR.

The interoperability checks that mattered in our implementation were:

- `respondTo: "allowlist"` requires at least one canonical pubkey;
- allowlist identities are normalized and deduplicated before the draft is
  published;
- unknown request keys are rejected rather than silently discarded;
- switching to `owner-only` or `anyone` does not retain a stale allowlist; and
- the owner review dialog displays the proposed people and remains editable
  before save.

The remaining design question is create/update parity. `draft-create` still
cannot propose selected people, while this PR correctly focuses on update. I
suggest keeping that expansion in a separate field-boundary issue so #3912 can
land without absorbing broader create configuration work.

We can help test the strict parser and review-dialog round trip once the PR is
ready; no need to duplicate the implementation.

Chessing234:
rebased onto main

Issue #3907, opened 2026-07-31
Mobile: agent replies not visible on phone — no Activity view to confirm agent responded


m-faizan-tariq:
## Summary

On the **mobile app**, when a user sends a message tagging an agent, the agent sometimes does not reply visibly. However, the **desktop app's Activity/Pulse log** shows the agent did receive the message, perform reasoning, and produce a response. Since the mobile app has **no Activity view**, the user has no way to know the agent responded when they are away from the desktop.

## Observed Behavior

1. User sends a message tagging an agent from the mobile app
2. No reply appears in the channel or thread on mobile
3. Back at the desktop app, the Activity/Pulse log shows the agent completed reasoning and generated a response
4. The mobile app provides no equivalent Activity feed, so the user cannot confirm whether the agent processed the message

## Impact

- User is left in the dark when out and about — messages appear unanswered even though the agent responded
- This makes agent collaboration unreliable from mobile, which defeats the purpose of a mobile companion app

## Expected Behavior

1. **Fix the root cause:** Agent replies should appear in the channel/thread on ALL clients (see #3825 — same symptom on desktop where replies appear in Activity but not in the channel)
2. **Feature request:** Add an Activity/Pulse view to the mobile app so users can see agent activity (reasoning, responses, status) even when replies aren't rendering in the conversation view

## Relation to Existing Issues

- **#3825** — desktop version of the same "reply exists but not rendered in channel" bug. Likely the same root cause: agent responses are stored/delivered as activity events but not consistently routed back into the channel/thread UI.
- **#3824** (moderator response) — the intermittent "no reply" behavior may also stem from OpenCode missing from the `default_agent_args` identity table (spawned without `acp` → interactive mode → 60s timeout). The moderator noted PR #3706 adds `"opencode" => Some(vec!["acp"])` and PR #3598 adds the `missing_binary` variant. Holding that issue for those PRs — but this mobile issue is a distinct UX gap (no Activity visibility on phone).

## Environment

- **Buzz Desktop:** 0.5.2 (bundle xyz.block.buzz.app)
- **Buzz Mobile:** latest
- **OS:** macOS (arm64) + iOS/Android
- **Agent runtime:** OpenCode 1.18.9-fork
- **Relay:** `wss://kamil22.communities.buzz.xyz`

## Suggested Fixes

1. Root-cause the relay/UI routing so agent replies always render in the channel/thread on all clients (align with #3825)
2. Add an Activity/Pulse tab to the mobile app mirroring the desktop one
[...]

PR #3915, opened 2026-07-31
fix(mobile): make agent activity one-tap discoverable (#3907)

Chessing234:
## Summary
- Mobile already had Activity / Pulse / Members → View activity; #3907 still reported a missing surface because the working-bot path was easy to miss.
- Add a channel-header activity shortcut when a bot is typing, share one sheet opener helper, and document the entry points (README + Agent-updates empty state).
- Channel timeline reply routing remains tracked with #3825 / #3908 — this PR is the discoverability half only.

## Test plan
- [ ] Tag an agent in a channel on mobile; while it is typing, confirm the activity icon appears in the channel app bar and opens `AgentActivitySheet`
- [ ] Members → View activity still opens the same sheet
- [ ] Activity tab → Agent updates empty state mentions the header / members paths
- [ ] `dart analyze` on touched mobile files

Related to #3907


Chessing234:
added Signed-off-by on all commits for the DCO check.

Chessing234:

Issue #3711, opened 2026-07-30
Owner control commands are invisible in the chat: no command affordance, and the outcome is only logged


Berenger-Wooclap:
**Motivation**

The owner control commands work, but the product treats them as if they didn't exist. `!shutdown`, `!cancel` and `!rotate` are ordinary text on the way in, and produce nothing on the way out.

*Nothing marks them as commands.* In the composer, `!rotate` is indistinguishable from a message. There is no autocomplete, no listing, no hint that this particular text will be consumed by the harness instead of reaching the agent — the only place the three commands are enumerated is `crates/buzz-acp/README.md`. A near-miss (`!rotat`, `!rotate please`) silently becomes a prompt, and the sent message then sits in the timeline styled like anything else you said.

*Nothing confirms the effect.* Each handler computes the outcome precisely and then writes it only to the harness log before consuming the event:

- `!rotate` with a turn in flight → `tracing::info!("!rotate received — cancelling in-flight turn and rotating session")` (`crates/buzz-acp/src/lib.rs:2118`)
- `!rotate` while idle → `tracing::info!(invalidated, "!rotate received — invalidated idle channel session(s)")` (`lib.rs:2125`)
- `!cancel` with no in-flight turn → `tracing::warn!("!cancel received but no in-flight task — no-op")` (`lib.rs:2080`)
- `!shutdown` → `tracing::info!("shutdown command from owner — exiting gracefully")` (`lib.rs:2045`)

From the chat, all four look identical: your own message, then silence. A successful rotate, a no-op cancel, and a command the agent never received are indistinguishable without opening the harness log. For controls whose entire purpose is to steer a session you cannot otherwise see, that is the one thing that should be visible.

The asymmetry is easiest to see next to `!model` (#3380): that command answers in the channel, via `spawn_notice` / `post_notice`. In the same diff, the `!cancel` and `!rotate` branches keep their `tracing::info!` and `continue`, so the gap survives that PR.

**Proposed solution**

Two halves, independently landable. The first is the smaller one and needs no protocol change.

[...]

PR #3537, opened 2026-07-29
feat(desktop): autocomplete ACP slash commands (rebase of #2653)

maczzinatui:
## Summary

Rebases and re-lands [#2653](https://github.com/block/buzz/pull/2653) onto current `main` so **#2528** can merge.

#2653 was **CONFLICTING** (stale since 2026-07-24). This branch:

- Keeps the feature intent of #2653 (catalog persist + composer `/` autocomplete from ACP `available_commands_update`)
- Resolves conflicts against current main:
  - `MessageComposer`: `ComposerDockToolbar` (main) + slash autocomplete (PR)
  - `localStorageQuota`: keep main cache prefixes + add `buzz-agent-command-catalog.v1`
  - `check-file-sizes.mjs`: take main (overrides list removed upstream)
- **No Buzz command semantics** — same non-goals as #919 / #2528 / #2653

Original work by @kevinmanase — please close #2653 in favor of this if accepted, or take these conflict resolutions onto #2653.

Fixes #2528

## Testing

- [x] `cargo test -p buzz-acp available_commands` → ok
- [x] `pnpm --dir desktop test` → 3780 pass
- [x] Manual Desktop dogfood: Hermes-acp agent emits `available_commands_update` → type `/` (bare and after `@mention`) → select command → agent turn fires (including static `/help` reply in channel/thread)

## Screenshots

Dogfood against a managed Hermes-acp agent that advertises skills/session commands via `available_commands_update`.

### Channel — bare `/` palette (agent catalogs in room)

Channel bare slash palette

### Channel — after `@agent` + `/`

After mention slash palette

### Thread — `/help` static command reply (channel-visible)

Thread help reply

### Thread composer — slash palette

Thread composer slash palette

PR #3916, opened 2026-07-31
feat(acp): post owner control command outcomes back into the channel

Berenger-Wooclap:
Implements the acknowledgement half of #3711.

> **History:** this branch briefly published the outcome as a kind:40099 system row. That was wrong — the relay rejects agent-signed 40099 at ingest — and it is reverted in e34d19e. See [the note below](#why-not-a-system-row) and @Chessing234's review, which caught it. The PR is back to the kind:9 shape.

## What problem this solves

`!shutdown`, `!cancel` and `!rotate` are consumed by the harness and never reach the agent, so the only trace they leave is a line in the harness log:

- `!rotate` with a turn in flight → `tracing::info!("!rotate received — cancelling in-flight turn and rotating session")`
- `!rotate` while idle → `tracing::info!(invalidated, "!rotate received — invalidated idle channel session(s)")`
- `!cancel` with no in-flight turn → `tracing::warn!("!cancel received but no in-flight task — no-op")`
- `!shutdown` → `tracing::info!("shutdown command from owner — exiting gracefully")`

From the chat, all four are indistinguishable, and so is a fifth case: a command the harness never received at all. You send `!rotate`, and you get your own message followed by silence. For controls whose whole purpose is to steer a session you cannot otherwise observe, that is the one thing that should be visible.

## What it does

Publishes the outcome as a kind:9 notice into the channel the command came from, reusing `post_notice` — the best-effort helper the dead-letter path already uses (threaded, errors logged and swallowed, 5s timeout).

| Command | Situation | Notice |
| --- | --- | --- |
| `!rotate` | turn in flight | Cancelled the current turn — the next one starts from a fresh session. |
| `!rotate` | cached session, idle | Session rotated — the next turn starts fresh. |
| `!rotate` | no cached session | No session to rotate — the next turn already starts fresh. |
| `!cancel` | turn in flight | Cancelled the current turn. |
| `!cancel` | idle | Nothing to cancel — no turn in flight. |
| `!shutdown` | any | Shutting down. |

All six are `&'static str` literals — the types enforce it (`spawn_control_notice` takes `content: &'static str`), so a notice cannot carry an error, a stack trace, a channel id or a session count. Longest is 72 characters.

## Key decisions

- **The no-op cases get a notice too**, and say plainly that nothing changed. Those are the ones that are invisible today and that read as a broken command, so staying silent there would leave the main symptom in place.
- **The notice lands in the thread the command was typed in.** Both paths derive their `ThreadTags` from the triggering event, so a `!rotate` typed in a thread is answered in that thread rather than at channel root. Pinned by a regression test (3c05927) — I verified it fails when `thread_ref_from_tags` is forced to return `None`, so it is not vacuous.
- **`!shutdown` awaits its notice** rather than spawning it. The harness is about to exit; a spawned task would race the shutdown.
- **`post_failure_notice` → `post_notice`.** The function is no longer failure-only. It is 1 signature, 1 call site and 4 log strings — happy to drop the rename if you would rather keep the name stable.
- **No `p` tags on the notice**, so it cannot wake another agent in the channel. Same shape the dead-letter notice already published.

## Why not a system row

A cancelled turn is arguably a lifecycle event rather than the agent speaking, so a kind:40099 system row is the better-looking option — and it does not work. Every client submission goes through `ingest_event_inner`, which calls `required_scope_for_kind` (`ingest.rs:1544`); 40099 has no arm there and falls through to `_ => Err("restricted: unknown event kind")`. Existing system rows bypass that path entirely: `emit_system_message` writes straight to `state.db.insert_event` (`side_effects.rs:775`).

Worth recording for whoever revisits this: **adding 40099 to `required_scope_for_kind` would be a security regression, not a fix.** Every system row today is relay-authored and authoritative — membership changes, `channel_archived`, and the `message_deleted` tombstone carrying "Removed by community moderators". Making the kind client-writable would let any member forge those. A legitimate system row needs its own agent-writable kind, which is a protocol decision rather than a scope-table edit. I'd rather propose that separately on #3711 than smuggle it into this PR.

## Tests

`rotate_idle_notice` is the one real branch (invalidated vs. nothing cached) and has a unit test asserting the two wordings stay distinct. `thread_ref_from_tags` has the regression test described above. The publishing path itself is I/O against the relay, covered by the same best-effort helper the dead-letter path already used.

```
cargo test -p buzz-acp -p buzz-sdk                                  # 642 + 241 passed
cargo fmt --all -- --check                                          # clean
cargo clippy --workspace --all-targets -- -D warnings               # clean
just test-unit                                                      # 5 suites passed
cd desktop && pnpm check && pnpm typecheck && pnpm test             # clean, 3781 passed
cd mobile && dart format --set-exit-if-changed . && flutter analyze && flutter test   # clean, 906 passed
```

No desktop or mobile code is changed by this diff, so there are no screenshots.

**Not verified:** I have not run this against a live relay end-to-end — the reasoning above is code reading, which is exactly what I got wrong the first time. The 13 required CI checks are still pending workflow approval, so nothing has exercised the real path yet.

## Follow-up deferred

The other half of #3711 — rendering `!` commands as commands in the composer, mirroring the ACP `/` palette in #3537 — is not in this PR.

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -159,6 +159,8 @@ Use `!cancel` to stop only the current turn; it is a no-op when the channel is i
 
 Owner control commands must be kind:9 stream messages from the owner, must mention this agent with a `p` tag, and are consumed by the harness instead of being forwarded to the agent.
 
+Because the command never reaches the agent, the harness posts the outcome back into the channel as a kind:9 notice — including when the command was a no-op (`!cancel` with no turn in flight, `!rotate` with no cached session), so a command that changed nothing is distinguishable from one that was never received.
+
 > **Note:** The default mode is `owner-only`. Agents without a registered `agent_owner_pubkey` will not respond to any events until the owner is resolved. Set `--respond-to anyone` to disable the gate entirely.
 
 **Examples:**
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -159,9 +159,7 @@ Use `!cancel` to stop only the current turn; it is a no-op when the channel is i
 
 Owner control commands must be kind:9 stream messages from the owner, must mention this agent with a `p` tag, and are consumed by the harness instead of being forwarded to the agent.
 
-Because the command never reaches the agent, the harness publishes the outcome back into the channel as a kind:40099 system message — the grey row clients render beside "X joined the channel", since a cancelled turn or a rotated session is a lifecycle event rather than the agent speaking. It lands in the thread the command was typed in, and it is published for the no-op outcomes too (`!cancel` with no turn in flight, `!rotate` with no cached session), so a command that changed nothing stays distinguishable from one that was never received.
-
-The payload is `{"type": …, "actor": <owner>, "target": <agent>}`. Clients switch on `type` and silently drop an unknown one, so a client older than this contract shows no row at all.
+Because the command never reaches the agent, the harness posts the outcome back into the channel as a kind:9 notice — including when the command was a no-op (`!cancel` with no turn in flight, `!rotate` with no cached session), so a command that changed nothing is distinguishable from one that was never received.
 
 > **Note:** The default mode is `owner-only`. Agents without a registered `agent_owner_pubkey` will not respond to any events until the owner is resolved. Set `--respond-to anyone` to disable the gate entirely.
 
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -159,7 +159,9 @@ Use `!cancel` to stop only the current turn; it is a no-op when the channel is i
 
 Owner control commands must be kind:9 stream messages from the owner, must mention this agent with a `p` tag, and are consumed by the harness instead of being forwarded to the agent.
 
-Because the command never reaches the agent, the harness posts the outcome back into the channel as a kind:9 notice — including when the command was a no-op (`!cancel` with no turn in flight, `!rotate` with no cached session), so a command that changed nothing is distinguishable from one that was never received.
+Because the command never reaches the agent, the harness publishes the outcome back into the channel as a kind:40099 system message — the grey row clients render beside "X joined the channel", since a cancelled turn or a rotated session is a lifecycle event rather than the agent speaking. It lands in the thread the command was typed in, and it is published for the no-op outcomes too (`!cancel` with no turn in flight, `!rotate` with no cached session), so a command that changed nothing stays distinguishable from one that was never received.
+
+The payload is `{"type": …, "actor": <owner>, "target": <agent>}`. Clients switch on `type` and silently drop an unknown one, so a client older than this contract shows no row at all.
 
 > **Note:** The default mode is `owner-only`. Agents without a registered `agent_owner_pubkey` will not respond to any events until the owner is resolved. Set `--respond-to anyone` to disable the gate entirely.
 
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -2047,15 +2047,11 @@ async fn tokio_main() -> Result<()> {
                                         // Awaited, not spawned: the harness is
                                         // about to exit, and a spawned task
                                         // would race the shutdown.
-                                        pool::post_system_notice(
+                                        pool::post_notice(
                                             &ctx.rest_client,
                                             buzz_event.channel_id,
                                             &queue::parse_thread_tags(&buzz_event.event),
-                                            &control_notice_payload(
-                                                NOTICE_SHUTDOWN,
-                                                owner,
-                                                &pubkey_hex,
-                                            ),
+                                            "Shutting down.",
                                         )
                                         .await;
                                         let _ = shutdown_tx.send(());
@@ -2099,12 +2095,10 @@ async fn tokio_main() -> Result<()> {
                                             &buzz_event.event,
                                             buzz_event.channel_id,
                                             if fired {
-                                                NOTICE_TURN_CANCELLED
+                                                "Cancelled the current turn."
                                             } else {
-                                                NOTICE_TURN_CANCEL_NOOP
+                                                "Nothing to cancel — no turn in flight."
                                             },
-                                            owner,
-                                            &pubkey_hex,
                                         );
                                         continue; // consume event — do NOT push to queue
                                     }
@@ -2143,7 +2137,7 @@ async fn tokio_main() -> Result<()> {
                                                 channel_id = %buzz_event.channel_id,
                                                 "!rotate received — cancelling in-flight turn and rotating session"
                                             );
-                                            NOTICE_SESSION_ROTATED_IN_FLIGHT
+                                            "Cancelled the current turn — the next one starts from a fresh session."
                                         } else {
                                             let invalidated = pool.invalidate_channel_sessions(buzz_event.channel_id);
                                             tracing::info!(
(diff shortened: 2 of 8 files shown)


Chessing234:
nice closing the loop on !shutdown/!cancel/!rotate — please keep the channel notice short and avoid dumping stack traces; owners need the outcome, not the harness internals.

Chessing234 (commented):
posting control-command outcomes as kind:9 notices closes a real silent-failure gap (especially !cancel no-op). awaiting the notice before shutdown is the important detail — a spawned task would race exit. worth confirming the notice inherits the same thread tags as the command so it doesn't land as a root channel message when the owner typed in a thread.

Berenger-Wooclap:
Thanks — agreed, and that constraint is already load-bearing here, so let me make it explicit.

All six notices are fixed one-line sentences, and the types enforce it: `spawn_control_notice` takes `content: &'static str` and `rotate_idle_notice` returns `&'static str`. There is no interpolation anywhere on this path, so a control notice cannot carry an error, a stack trace, a channel id, or a session count. The longest is 72 characters:

- `Shutting down.`
- `Cancelled the current turn.`
- `Nothing to cancel — no turn in flight.`
- `Cancelled the current turn — the next one starts from a fresh session.`
- `Session rotated — the next turn starts fresh.`
- `No session to rotate — the next turn already starts fresh.`

The `invalidated` count only selects between the last two wordings; it is never printed.

One thing worth separating, since the rename in this PR may be what prompted the caution: the dead-letter notice that now shares `post_notice` *does* interpolate the failure — `PromptOutcome::Error(e) => format!("{e}")` — and that behaviour is unchanged here. If you'd like that one bounded (truncated, or reduced to a classified outcome), happy to do it as a separate PR rather than widening this one.

## One design question while you're here

This PR makes the agent say these lines: the notice is a kind:9 signed with the agent's own key, so it renders as an ordinary message from the agent. But a cancelled turn or a rotated session is arguably a lifecycle event rather than the agent speaking, and the alternative I listed in #3711 is a kind:40099 system row instead.

I looked into what that would actually cost, in case it helps you decide:

- **It is publishable by the harness.** `KIND_SYSTEM_MESSAGE` is not in `is_relay_only_kind`, so an agent-signed 40099 with an `h` tag is accepted, stored and fanned out. `handle_side_effects` has no arm for 40099, so there is no relay side effect to worry about either.
- **But it needs both renderers.** `describeSystemEvent` falls through to `default: return null` (`SystemMessageRow.tsx`) and mobile's `SystemEvent.fromContent` maps an unknown `type` to `null` (`timeline_message.dart`), so new payload types have to be added in both places, plus screenshots. More importantly, **any client older than that change shows nothing at all** — where a kind:9 renders everywhere, including third-party Nostr clients.

So: agent as it stands, or a system row? I lean slightly toward the system row on aesthetics.

Berenger-Wooclap:
> worth confirming the notice inherits the same thread tags as the command

Confirmed. I made a small comit to test this.

Berenger-Wooclap:
Answering my own question above: I've switched this to the kind:40099 system row, pushed as 496f3bd.

It renders as a system line beside "X joined the channel" rather than as a message from the agent, which is the right weight for a lifecycle event. Screenshot of all six outcomes is in the PR description.

Two things I checked that made the switch safe, in case they're useful:

- `KIND_SYSTEM_MESSAGE` is not in `is_relay_only_kind`, and `handle_side_effects` has no arm for 40099, so an agent-signed system message is accepted, stored and fanned out with no relay side effect.
- Thread placement is derived from NIP-10 `e` tags with **no kind check** (`getThreadReference` in `formatTimelineMessages.ts`), so the system row still lands in the thread the command was typed in — your earlier point survives the change, and the regression test still guards it.

The honest cost, spelled out in the description: an unknown `type` is dropped silently by both renderers, so a client older than this contract shows no row at all, where the kind:9 notice rendered everywhere. If you'd rather keep the shape that degrades gracefully, say so and I'll revert to kind:9 — the rest of the work is unaffected.

I also owe you a correction: I only just noticed #3877 implements this same slice, opened before mine. Sorry for the duplicated review load. If that one is the shape you want, close this and I'll send the thread-placement test and the README update on top of it instead.

Berenger-Wooclap:
@wpfleger96 — apologies for the direct ping. There are two things here only a maintainer can unblock, and recent commit history on `buzz-acp` pointed at you; happy to be redirected.

**1. The workflows are pending approval.** `CI`, `Docker image` and `Desktop Release Candidate` are sitting in `action_required` (first PR from this fork), so 13 of the 14 required checks haven't run at all. Everything I could run locally is green — Rust 644 + 241 tests with `clippy --workspace -D warnings`, `just test-unit`, desktop biome + `tsc` + 3781 tests, mobile `dart format` + `flutter analyze` + 906 tests — but that's no substitute for the real gate, and the relay round-trip in particular is only verified by reading the code.

**2. A design call I shouldn't make on my own.** This reports the outcome of `!cancel` / `!rotate` / `!shutdown` as a kind:40099 system row. The alternative is a plain kind:9 message. The trade-off:

- **System row** — a cancelled turn or a rotated session is a lifecycle event, not the agent speaking, so it reads better next to "X joined the channel". Cost: an unknown `type` is silently dropped by both renderers, so any client older than this contract shows nothing at all.
- **kind:9** — degrades gracefully everywhere, including third-party Nostr clients. Cost: lifecycle events sit permanently in the conversation.

I went with the system row, but the graceful-degradation argument is real and this is your call rather than mine. Switching back is a small change if you'd rather have it.

Chessing234 (commented):
re-reviewed tip after the kind:40099 switch — thread placement + no-op wording look right, but i don't think the harness can publish these: `required_scope_for_kind` in ingest doesn't list 40099 (falls through to unknown kind), and existing system rows are relay-signed via `emit_system_message`. agent `submit_event` of a 40099 would get rejected, so the chat stays silent again. either keep kind:9 like #3877, or get a relay path that accepts these. happy to re-read once that lands.

Berenger-Wooclap:
@Chessing234 you're right, and thank you — this was a real bug, not a nitpick. Reverted in e34d19e; the PR is back to kind:9.

I traced it to be sure, and it's exactly as you described:

- `required_scope_for_kind` (`ingest.rs:211`) has no arm for 40099, so it falls through to `_ => Err("restricted: unknown event kind")` (`ingest.rs:316`).
- That call sits in `ingest_event_inner` (`ingest.rs:1544`), the single path for every client submission — WebSocket and `POST /events` alike — and nothing short-circuits it for this kind.
- The existing system rows never touch it: `emit_system_message` writes straight to `state.db.insert_event` (`side_effects.rs:775`).

So the agent's `submit_event` would have been rejected and the channel would have stayed silent — the precise failure this PR exists to remove. My check before the switch was necessary but not sufficient: I confirmed 40099 wasn't in `is_relay_only_kind` and had no `handle_side_effects` arm, and concluded "publishable" while missing the gate that actually decides. Sorry for the wasted review pass.

**One thing worth adding, for whoever revisits this.** The obvious one-line fix — add 40099 to `required_scope_for_kind` — would be a security regression rather than a fix. Every system row today is relay-authored and authoritative: `member_joined`, `member_removed`, `channel_archived`, and the `message_deleted` tombstone that renders as "Removed by community moderators". Making the kind client-writable would let any member forge fake membership history and fake moderation tombstones. If a system row is wanted here, it needs its own agent-writable kind — a protocol decision, not a scope-table edit. I'd rather propose that separately on #3711 than widen this PR, and I've written the reasoning into the description so the next person doesn't rediscover it the hard way.

**On the duplicate.** Now that this is back to kind:9, it is functionally #3877 plus two things: the regression test pinning the notice to the thread the command was typed in (which is your earlier review point, and which #3877 doesn't cover), and the `buzz-acp` README update — that README is currently the only documentation of these commands and still describes them as silent.

Two PRs for a hundred lines doesn't serve anyone. I'm happy either way: close this and I'll send those two pieces on top of #3877, or keep this one and close that. That's a maintainer call rather than mine — I just don't want to cost anyone a second review of the same change.

PR #3917, opened 2026-07-31
fix(sdk): recognize CJK and Markdown-adjacent mentions

### LINKED ISSUE #2526: [Bug] Bold-wrapped @mentions (**@Name**) send zero p tags from buzz-cli — no notification, exit 0, no warning

### LINKED ISSUE #3904: Proposal: define CLI/SDK @mention behavior next to CJK text
## Summary

The CLI/SDK mention parser only recognizes `@Name` when the `@` is at the start of the message or preceded by ASCII whitespace.

That rule is explicitly documented in `crates/buzz-sdk/src/mentions.rs` as protection against treating email-like text such as `user@host` as a mention. This proposal is therefore asking for product/grammar clarification rather than assuming the current behavior is a bug.

In languages that commonly omit spaces, the same rule makes natural agent-authored handoffs look like mentions while emitting no `p` tag:

| Content | Current result |
|---|---|
| `@Scout please review` | resolves Scout |
| `你好 @Scout` | resolves Scout |
| `交给@Scout处理` | no mention tag |
| `user@example.com` | no mention tag, intentionally |

The third case is easy for a human or agent to write in Chinese, but the recipient is not notified and a mention-gated ACP agent does not wake.

## Question for maintainers

Should CLI/SDK auto-resolution remain strictly token-delimited, with agent callers expected to use explicit `--mention <pubkey>` whenever notification delivery matters?

Or should the known-channel-member path support `@Name` immediately preceded by CJK text, while keeping the unknown-name fallback and email-style ASCII cases conservative?

## Possible narrow direction

If CJK-adjacent mentions are desired, one conservative option would be:

- only relax the left boundary when `Name` exactly matches a known member of the target channel;
- keep ASCII word/email-local characters before `@` excluded, so `user@host` behavior is unchanged;
- keep unknown-name fallback parsing whitespace/start-delimited;
- preserve self-mention suppression and existing mention caps;
- add regression cases for CJK prefixes, ASCII emails, partial names, and code regions.

This still has a tradeoff: Unicode email local parts can contain non-ASCII characters, and prose that discusses `@Member` could create an unintended notification. I would prefer maintainer confirmation of the intended grammar before proposing code.

## Related work

- #2526 and PRs #2547 / #2684 cover Markdown emphasis, spoiler, and parenthesis delimiters in the Rust parser.
- PR #2465 covers CJK-adjacent mention detection in Desktop.
- #3015 and PR #3047 cover warnings for unresolved CLI mentions.
- PR #3330 added explicit repeatable `--mention`, which is the reliable current path for agent-authored notifications.

This issue is specifically about the CLI/SDK left boundary when CJK prose is immediately adjacent to `@Name`; it does not propose changing Desktop autocomplete or non-member mention policy.

Chessing234:
## Summary
- Open `@mention` extraction when `@` is preceded by anything other than an ASCII letter/digit — so `交给@Scout处理` emits a `p` tag while `user@example.com` stays inert (#3904).
- Align `extract_at_mentions_with_known` with the same char-aware boundary (it previously inspected a single UTF-8 byte) and treat trailing CJK as a word boundary.
- Document the rule in `docs/mention-boundaries.md`, AGENTS gotchas, CONTRIBUTING, and the CLI README.

## Test plan
- [x] `cargo test -p buzz-sdk extract_at_names`
- [x] `cargo test -p buzz-sdk extract_at_mentions_with_known_allows_cjk_adjacent`
- [x] `cargo test -p buzz-sdk email_address_not_matched`
- [ ] `buzz messages send --content "交给@Scout处理"` against a channel where Scout is a member → Scout receives a mention / agent wakes

Closes #3904

The same boundary rule also accepts Markdown-emphasis and parenthesized mentions, rather than maintaining a separate allowlist of leading delimiters. This covers the emphasis issue as well as CJK adjacency.

Closes #2526.


Chessing234:
Rebased onto latest `upstream/main` with `Signed-off-by` on all commits and force-pushed. CI should re-run shortly — still ready for @block/buzz-oss-team review when you have bandwidth.

Chessing234:

Chessing234:
@wesbillman sdk @mention fix for cjk-adjacent tokens without breaking emails. tip is rebased + dco. review/merge when you can?

cristiansotogarciaxatech:
This branch is doing more than #3904 and the title hides it. It also fixes #2526, and on every case I can produce it is the only one of the four open branches editing this predicate that fixes all of them.

Five open PRs currently touch `@mention` boundary handling in `crates/buzz-sdk/src/mentions.rs`. #2547, #2684, #5932, this one, and #6079 (case folding rather than the boundary). None of the five reference each other, and #2526, the filed issue for the emphasis case, is linked from two of them and not from here.

So I measured rather than argued. One probe test appended to `crates/buzz-sdk/src/mentions.rs`, one ubuntu-latest runner, five trees fetched by sha, calling `extract_at_mentions_with_known(input, &["Scout", "Will Pfleger"])`.

Run: https://github.com/cristiansotogarciaxatech/buzz/actions/runs/34030702637

| body | main `3c7f288c` | #2547 `0a39d45d` | #2684 `3128593d` | #5932 `32fcbd0f` | this PR `322611de` |
|---|---|---|---|---|---|
| `@Scout hi` (positive control) | scout | scout | scout | scout | scout |
| `**@Scout** hi` | none | scout | scout | scout | scout |
| `(@Scout) hi` | none | scout | scout | scout | scout |
| `**@Will Pfleger** hi` | none | will pfleger | will pfleger | will pfleger | will pfleger |
| `\u{a0}@Scout hi` (NBSP) | none | none | none | scout | scout |
| `\u{feff}@Scout hi` (BOM) | none | none | none | none | scout |
| `\u{200b}@Scout hi` (ZWSP) | none | none | none | none | scout |
| `交给@Scout处理` | none | none | none | none | scout |
| `mail user@Scout.com here` (negative control) | none | none | none | none | none |

"none" is an empty vec, which on the CLI path means the message publishes with zero `p` tags and exit 0. Nobody is notified and the text renders perfectly at the destination.

Two things fall out of that table.

**Main drops `(@Scout)`.** Parentheses around a mention are more common in ordinary prose than `**` is, and nobody has filed that one. Same four lines as everything else here.

**The BOM row is the argument for your approach.** `!prev.is_ascii_alphanumeric()` needs no list of characters. The other three branches carry an allowlist of leading characters, and every new character costs another PR. U+FEFF is the proof, because it is not whitespace under `char::is_whitespace`, measured on the same runner as `feff_ws=false`, so even the Unicode-aware allowlist in #5932 misses it. #2526 said not to hand-maintain a delimiter list. That was the right call and this is the branch that follows it.

**A leading BOM is not an exotic input.** `crates/buzz-cli/README.md:34` documents `buzz messages send --content - < message.md`, and `read_or_stdin` at `crates/buzz-cli/src/validate.rs:168` is a bare `read_to_string` with no BOM handling. On Windows most default text writers emit UTF-8 with a BOM. So the documented way to send a long body silently eats an opening mention, on the platform where BOM is the default file encoding.

I hit this in production on a Windows box running managed agents. A message reached the relay with U+FEFF at position 0 opening with an at-mention. `accepted: true`, zero `p` tags, addressee never notified, and the same name one line further down the body tags correctly. Position 0 is the only position that fails and it is the one that fails silently, because a name that gets extracted and does not resolve is a hard error at `crates/buzz-cli/src/commands/messages.rs:140`, while a name that never gets extracted is not an error at all. I have not isolated which component injects the BOM on my box, so treat that as a symptom report rather than a claim about Buzz. The parser behaviour in the table is measured, not inferred.

The doc comment two functions up already states the contract every "none" in that table breaks, at `messages.rs:158`. Publishing visible mention text without its intended `p` tag is worse than not sending.

I am not asking you to change anything. Three suggestions if you want them. Add `Closes #2526` to the description, because on the measurement above this branch closes it. Add the BOM, paren and NBSP cases to the test module, since those are the ones that discriminate this fix from the other three. And a line in `docs/mention-boundaries.md` saying the rule is a denylist by design would stop the next contributor turning it back into an allowlist.

Issue #3505, opened 2026-07-29
Managed agents sit idle in an empty community: launcher pins 127.0.0.1 while the app UI uses localhost (self-hosted dev)


JGroutsis:
**Describe the bug**
On a self-hosted dev instance, every Desktop-managed agent (the built-in Welcome Team personas and user-created agents) connects and authenticates, then discovers zero channels and sits idle — while the Desktop UI itself works normally in the same instance. Cause: community identity is host-derived, the dev seed registers each host spelling as its own community, and the managed-agent launcher pins agents to `ws://127.0.0.1:PORT` while the app's active community is `localhost`. The agents land in a different, empty community, and fail-closed tenancy correctly gives them nothing.

**Steps to reproduce**
1. `just dev` on macOS with the relay port remapped in `.env` (`BUZZ_BIND_ADDR=0.0.0.0:3100`, `RELAY_URL=ws://localhost:3100`).
2. Complete onboarding in the Desktop app; create channels, or use existing ones — the UI works.
3. Let onboarding create the Welcome Team agents, or create a managed agent of any runtime.
4. Watch the agent log under `Application Support/<bundle-id>/agents/logs/`.

**Expected behavior**
The managed-agent launcher uses the same relay URL / community as the app's active community — or the dev seed maps loopback spellings (`localhost`, `127.0.0.1`, with and without port) onto one community — so agents see the same channels the user does.

**Version and platform**
- Buzz version: 0.5.0 (tag checkout, commit 4a977c5, `just dev`)
- OS: macOS 15.7 (Apple Silicon)

**Logs / additional context**
Every managed agent, on every start and restart:
```
INFO buzz_acp: connected to relay at ws://127.0.0.1:3100
INFO buzz_acp: discovered 0 channel(s)
WARN buzz_acp: no channel subscriptions resolved — agent will sit idle
```

`communities` table after onboarding — one community per host spelling:
```
3c89e6ae-…  localhost
129cbfbe-…  127.0.0.1
e2f45882-…  127.0.0.1:3100
(+ localhost:3100)
```

Setting `BUZZ_RELAY_URL: ws://localhost:3100` in `agents/global-agent-config.json` `env_vars` does not help — the per-agent stored `relay_url` wins. Restarts do not help.

[...]

Issue #2641, opened 2026-07-23
Agents show RUNNING/online but never respond — no channel subscriptions resolved, empty harness log (self-hosted, macOS)


DroVibe:
**Title:** Agents show RUNNING/online but never respond — no channel subscriptions resolved, empty harness log (self-hosted, macOS)

## Summary

On a self-hosted relay, managed agents start, connect, and report online, but never respond to mentions or DMs. Affects both the Goose and Claude Code harnesses identically. The Harness Log stays empty, so the failure is invisible from the UI.

## Environment

- macOS Tahoe 26.3.1, Apple M1 Max, 64GB
- Self-hosted relay, built from source at `main`, run via `just relay`
- Desktop: packaged Buzz.app (also reproduced with a `just dev` build from the same commit)
- Docker Engine + Postgres/Redis/Typesense/MinIO all healthy
- Buzz Mesh enabled, sharing `unsloth/gemma-4-26B-A4B-it-GGUF:UD-Q4_K_M`

## Steps to reproduce

1. Stand up a relay from source, connect the desktop app to `http://localhost:3000`
2. Enable Share this machine, select a recommended model, wait for download
3. Create an agent — harness Goose, provider OpenAI-compatible, model set to the mesh model ID
4. Add the agent to a channel and @mention it
5. Also try a DM to the agent

## Expected

Agent responds, or surfaces an error.

## Actual

No response via channel mention or DM. Agent card shows a green online dot; runtime status shows RUNNING. Harness Log shows "no activity logged".

## Evidence

**Mesh endpoint is healthy** — `curl http://localhost:9337/v1/models` returns the shared model correctly.

**Goose works standalone** — `goose session` in a terminal responds normally against the same mesh endpoint. `goose info -v` shows provider `openai`, host `http://localhost:9337`, correct model, `configured: true`.

**Relay never sees agent traffic.** After a mention, relay logs show only the human pubkey issuing `/query` and `/events`. In one earlier session an agent pubkey did complete NIP-42 auth, queried once, then never appeared again. Messages (kind 9) are ingested but no agent activity follows.

**Harness processes spawn but freeze.** After several mention attempts:

```
$ ps aux | grep -i -E "goose|buzz-acp" | grep -v grep
... 24 × /Users/<user>/.local/bin/goose acp   (all spawned in the same second, all at 0:00.04 CPU)
... 2  × /Applications/Buzz.app/Contents/MacOS/buzz-acp
```

[...]

PR #3941, opened 2026-07-31
fix(desktop): keep the agent relay authority unnormalized at spawn

### LINKED ISSUE #3283: Local dev: loopback hosts are seeded as separate communities, so desktop-managed agents silently discover 0 channels

D12Labs:
## Summary

Desktop-managed agents on a self-hosted relay authenticate into a **different tenant than their owner**, discover zero channels, and sit idle while reporting healthy:

```
INFO buzz_acp: buzz-acp starting: relay=ws://127.0.0.1:3000 ...
WARN buzz_acp: no channel subscriptions resolved — agent will sit idle
```

`spawn_agent_child` derived the child's `BUZZ_RELAY_URL` from `runtime_key.relay_url`. `ManagedAgentRuntimeKey::new` canonicalizes that field via `buzz_core::relay::normalize_relay_url`, which folds every loopback host to `127.0.0.1` so a pair spelled `localhost` and one spelled `127.0.0.1` collapse to a single runtime identity. That is correct for *identity* and wrong for *connecting*: the relay resolves each connection's tenant from the request authority, and under row-zero host binding `localhost:3000` and `127.0.0.1:3000` are distinct communities.

So a desktop on `ws://localhost:3000` writes its channels, owner row and events into the `localhost:3000` community, while every agent it spawns connects to `ws://127.0.0.1:3000` and finds an empty one. Fail-closed tenancy then correctly gives those agents nothing, and `buzz-acp`'s kind:39002 discovery legitimately returns zero — so nothing errors and it reads as a broken agent rather than a tenant split.

`relay::relay_http_base_url` already documents and tests this exact invariant:

> Tenant host-binding keys off the HTTP Host/authority. The desktop must **not** rewrite localhost to 127.0.0.1, or local dev HTTP calls target a different unmapped community than the WebSocket URL.

That guard was defeated upstream — it received a value this code had already rewritten. The git-credential-nostr HTTP base is derived from the same variable, so it was mis-scoped too.

## The change

Route the child's relay URL through a named, authority-preserving helper instead of the canonical pair URL. Only whitespace and a trailing root slash are trimmed; the host is never rewritten.

Identity stays canonical — dedup, `runtime_id`, log paths are all unchanged. Only the connection keeps the caller's spelling. Remote hosts are unaffected, since normalization only ever rewrote loopback.

### Related issues

This implements **suggested fix 3** from #3283 ("Don't canonicalize the relay URL handed to spawned harnesses — pass through the community URL the app itself uses"), and addresses the launcher half of #3505, #2444 and #2641. Four independent reports of the same trap.

I went with 3 rather than the alternatives deliberately:

- **1 — normalize loopback in tenant resolution.** Changes tenancy semantics for every deployment to fix a dev-only symptom. Too broad for this.
- **2 — seed only the canonical loopback host.** Fixes the default `just dev` path but not a user who types `localhost` into the community dialog, which is the actual reproduction here.
- **4 — warn on zero-discovery mismatch.** Complementary and worth doing, but a diagnostic, not a fix.

The seeding contradiction #3283 describes still stands — `seed-local-community.sh` continues to register four loopback spellings as four isolated tenants, so anything else that canonicalizes a host before a tenant lookup can hit the same class of bug. **I would not close #3283 on this PR**; the agent symptom is fixed, the underlying "one host = one community" vs "localhost == 127.0.0.1" contradiction is not.

## Testing

- 5 new regression tests in `managed_agents::runtime::tests`, including one pinning the invariant directly: `ManagedAgentRuntimeKey::new(.., "ws://localhost:3000")` canonicalizes to `127.0.0.1` **and** must differ from the connect URL. Whitespace/trailing-slash trimming and remote-host passthrough are covered.
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib` — **2008 passed, 0 failed**, 14 ignored.
- `just desktop-tauri-clippy` (`-D warnings`) and `just desktop-tauri-fmt` clean.

Reproduced the original failure on macOS with Colima against a self-hosted relay: four managed agents (3× `claude-agent-acp`, 1× `codex-acp`) all logged `discovered 0 channel(s)` while the desktop had 4 channels and 89 events in the `localhost:3000` tenant, confirmed by `SELECT` on `communities` / `channels` / `events`.

**Limits of the verification, stated plainly:** the new tests pin the helper's contract, not that `spawn_agent_child` calls it — that path needs an `AppHandle` and isn't unit-testable, so a future edit to the call site would not fail the suite. Closing that properly needs a seam around spawn-env construction, which felt out of scope for a one-line behavioral fix; happy to add it if maintainers prefer. I also verified the fix only by unit test and inspection rather than end-to-end, because unblocking my own instance required aligning both sides onto `127.0.0.1`, and demonstrating the `localhost` path again would mean recreating the split.

### DIFF
--- a/desktop/src-tauri/src/managed_agents/runtime.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime.rs
@@ -445,6 +445,17 @@ pub(crate) fn configure_runtime_cli(
     }
 }
 
+/// The relay URL a spawned agent actually connects to.
+///
+/// Authority-preserving by contract: only whitespace and a trailing root slash
+/// are trimmed. The host is never rewritten, because the relay derives each
+/// connection's tenant from the request authority. See the call site in
+/// `spawn_agent_child` for why the canonical (normalized) pair URL must not be
+/// used here.
+fn agent_connect_relay_url(workspace_relay_url: &str) -> String {
+    workspace_relay_url.trim().trim_end_matches('/').to_string()
+}
+
 /// Spawn an agent process without holding any locks on records or runtimes.
 /// Returns the child process and log path on success. The caller is responsible
 /// for updating `ManagedAgentRecord` fields and inserting into the runtimes map.
@@ -545,9 +556,23 @@ pub fn spawn_agent_child(
         .map(|p| p.display().to_string())
         .unwrap_or_else(|| effective_command.clone());
 
-    // The caller supplies the explicit canonical pair relay. This is the only
-    // relay this child may connect to, regardless of the record/workspace default.
-    let effective_relay_url = runtime_key.relay_url.clone();
+    // The caller supplies the explicit pair relay. This is the only relay this
+    // child may connect to, regardless of the record/workspace default.
+    //
+    // Deliberately NOT `runtime_key.relay_url`: that field is canonicalized by
+    // `normalize_relay_url`, which folds every loopback host to `127.0.0.1` so a
+    // pair spelled `localhost` and one spelled `127.0.0.1` collapse to a single
+    // runtime identity. Tenant host-binding resolves the community from the
+    // request authority, where `localhost:3000` and `127.0.0.1:3000` are
+    // distinct communities — so handing the canonical form to the child makes
+    // every managed agent on a local relay authenticate into a different tenant
+    // than its owner, discover zero channels, and sit idle looking healthy.
+    // `relay::relay_http_base_url` already guards this for HTTP; the guard is
+    // useless if the value reaching it was rewritten here first.
+    //
+    // Identity stays canonical (dedup, `runtime_id`, log paths); the connection
+    // keeps the caller's authority.
+    let effective_relay_url = agent_connect_relay_url(relay_url);
 
     // Augment PATH for DMG launches so child processes can find:
     //   - bundled CLI via ~/.local/bin symlink
--- a/desktop/src-tauri/src/managed_agents/runtime/tests.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime/tests.rs
@@ -1314,3 +1314,59 @@ fn restart_eligible_false_when_orphan_has_no_drift() {
 fn restart_eligible_false_when_non_orphan_has_no_drift() {
     assert!(!super::restart_eligible(false, false, false));
 }
+
+// ── agent_connect_relay_url tests ───────────────────────────────────────
+//
+// Tenant host-binding resolves a community from the request authority, so
+// `localhost:3000` and `127.0.0.1:3000` are different communities. The URL
+// handed to a spawned agent must therefore preserve the authority the
+// workspace uses. `ManagedAgentRuntimeKey` canonicalizes loopback hosts to
+// `127.0.0.1` for identity/dedup; feeding that canonical form to the child
+// stranded every managed agent on a local relay in an empty tenant, where it
+// discovered zero channels and idled while looking healthy.
+
+#[test]
+fn connect_url_preserves_localhost_authority() {
+    // The regression: must NOT come back as ws://127.0.0.1:3000.
+    assert_eq!(
+        super::agent_connect_relay_url("ws://localhost:3000"),
+        "ws://localhost:3000"
+    );
+}
+
+#[test]
+fn connect_url_preserves_ipv4_loopback_authority() {
+    assert_eq!(
+        super::agent_connect_relay_url("ws://127.0.0.1:3000"),
+        "ws://127.0.0.1:3000"
+    );
+}
+
+#[test]
+fn connect_url_differs_from_canonical_key_for_localhost() {
+    // Pins the actual invariant: identity canonicalizes, connection does not.
+    let key =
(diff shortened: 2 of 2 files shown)


salman1993:
🤖 Thank you for working on this. We reviewed the overlapping proposals and are consolidating the configured-vs-canonical relay URL fix in #4859 so review and implementation can happen in one place. We’re marking this PR as superseded and asking a maintainer or the author to close it. Please review #4859 and contribute any feedback, tests, or implementation details you think should be carried into the consolidated fix.

anilkishan:
salman asked us to consolidate on #4859. We followed your call not to close #3283 on a spawn-path PR - the seed script still inserts four isolated loopback tenants, so that issue stays open. Thank you for writing that down so it did not get swept up.

PR #3948, opened 2026-07-31
refine(desktop): surface unread thread cues

daveh-beep:
## Summary

Makes unread thread activity visible inside the source channel without adding navigation or state-machine behavior:

- Promotes the thread summary’s muted unread text into a legible `N new` pill.
- Adds a continuous accent spanning the parent message and its thread summary.
- Includes the unread count in the thread summary’s accessible name.

## Why this slice

This is the smallest durable response to the in-channel dead end: users who enter a channel directly can see which parent holds unread replies. It changes only presentation over the existing per-thread unread count and read frontier—no new route, jump behavior, dismissal state, or server data.

The earlier floating jump action was removed after red-team review found multi-thread dismissal and virtualizer edge cases that would require a larger state model. That work can return independently if needed.

## Verification

- Diff rebuilt from current `main`
- `git diff --check`
- TypeScript typecheck
- Biome
- File-size and px-text guards
- 3,885 desktop unit tests passed
- 13 targeted `thread-unread` Playwright smoke tests passed

### DIFF
--- a/desktop/src/features/messages/ui/MessageThreadSummaryRow.tsx
+++ b/desktop/src/features/messages/ui/MessageThreadSummaryRow.tsx
@@ -95,9 +95,11 @@ export function MessageThreadSummaryRow({
     THREAD_SUMMARY_SURFACE_AVATAR_INSET_REM,
   )})`;
   const replyLabel = summary.replyCount === 1 ? "reply" : "replies";
+  const unreadLabel =
+    unreadCount != null && unreadCount > 0 ? `, ${unreadCount} unread` : "";
   const summaryAriaLabel = summary.lastReplyAt
-    ? `View thread with ${summary.replyCount} ${replyLabel}, last reply ${formatThreadSummaryLastReplyTime(summary.lastReplyAt)}`
-    : `View thread with ${summary.replyCount} ${replyLabel}`;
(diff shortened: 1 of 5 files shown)


daveh-beep:
🤖 ## Corrected design scope — in-channel unread-thread cues

Rendered from the corrected PR head (`9eb8c616`; based on current `main`).

The red-team pass removed the floating jump pill: its channel-wide dismissal state could hide a second unread thread, it targeted the deferred rather than rendered virtualizer model, and it could toggle an already-open thread closed. Those behaviors were too risky for this low-diff slice.

This PR now changes only durable presentation over existing unread state:

- a continuous accent spans the parent message and thread summary;
- the existing muted unread text becomes a legible `N new` pill;
- the thread summary's accessible name includes the unread count.

Unread thread cues in the channel

No navigation, dismissal state, routing, server data, or sidebar behavior is included.

klopez4212 (commented):
I think the pill is cool and stands out more..might still ask wes to review because I have no idea what breaks things

daveh-beep:
🤖 Closing this stale slice so we can reset to current main and re-propose only the smallest still-useful change locally first. This branch has diverged from the current unread/notification work; preserving the rationale here rather than carrying conflicts forward.

PR #3950, opened 2026-07-31
refine(desktop): add sender context to thread notifications

daveh-beep:
## Summary

Adds sender context to desktop thread-reply notifications, changing generic titles such as `Reply · #design` to titles such as `Maya replied in #design`.

The implementation resolves display names from the existing React Query profile cache, performs no new network request, and falls back safely when a cached profile is unavailable.

## Why this slice

This is isolated presentation work: notification eligibility, read state, navigation, and deep links are unchanged. The payload already carries the sender and destination metadata; this slice makes that context visible before the user chooses where to navigate.

## Verification

- TypeScript typecheck
- 3,890 desktop unit tests passed
- New helper coverage for cached profiles, missing profiles, malformed cache values, and fallback titles

### DIFF
--- a/desktop/src/app/useAppShellDesktopNotifications.test.mjs
+++ b/desktop/src/app/useAppShellDesktopNotifications.test.mjs
@@ -1,0 +1,59 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import {
+  resolveThreadReplySenderName,
+  threadReplyNotificationTitle,
+} from "./useAppShellDesktopNotifications.ts";
+
+const PUBKEY = "a".repeat(64);
+
+test("thread reply title uses resolved sender name with channel", () => {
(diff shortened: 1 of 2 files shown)


daveh-beep:
🤖 ## Design — sender context in thread-reply notifications

Rendered from this PR's branch (`6a354420`). The notification title now answers **who** and **where** before the user navigates: `alice replied in #general`, while the body keeps the reply preview. Notification eligibility, sound, read state, click target, and deep-link behavior are unchanged.

Thread notification with sender context

The capture uses the exact payload emitted through `sendDesktopNotification`; only the visual shell around it is a deterministic browser stand-in for the OS notification surface.

daveh-beep:
🤖 Closing this stale slice so we can reset to current main and re-propose only the smallest still-useful change locally first. This branch has diverged from the current unread/notification work; preserving the rationale here rather than carrying conflicts forward.

PR #3602, opened 2026-07-29
chore(ci): bump Linux AppImage build container to ubuntu:24.04

wpfleger96:
## What and why

The Buzz AppImage is built on `ubuntu:22.04`, which links WebKitGTK against FreeType 2.11.1. Because `libfreetype.so.6` is on the linuxdeploy community excludelist, the bundled WebKit loads the **host's** FreeType at runtime instead of the bundled one.

FreeType 2.13.0 (released 2023-02-09) added `FT_Bool read_variable` to `FT_ColorStopIterator`, growing the struct from 16 to 20 bytes. Any host running FreeType ≥ 2.13 (Fedora 42+, Ubuntu 24.04+) has a struct-layout mismatch with the 22.04-compiled WebKit. The mismatched offsets corrupt color-stop index arithmetic inside Skia's COLRv1 renderer, producing the assertion abort in issues #2548 and #2982:

```
stl_vector.h:1123: Assertion '__n < this->size()' failed.
... colrv1_configure_skpaint(FT_Face, ...) ...
```

## Fix

Bump the build container to `ubuntu:24.04` (noble), which ships FreeType **2.13.2**. Noble's struct layout matches every crash-affected host. The ABI mismatch disappears and the crash is eliminated at root.

WebKitGTK also advances from **2.50.4** (jammy backport) to **2.52.3** (noble backport).

## Glibc floor change

| Build base | glibc floor | Oldest supported AppImage distro |
|---|---|---|
| ubuntu:22.04 (before) | 2.35 | Ubuntu 22.04 LTS, Debian 12 |
| ubuntu:24.04 (after) | 2.39 | Ubuntu 24.04 LTS, Fedora 40+ |

Ubuntu 22.04 LTS and Debian 12 users lose AppImage support. Both distributions continue to receive first-class `.deb` / `.rpm` packages, which use the system WebKit and are unaffected. The crash-affected users (Fedora 42/44, Ubuntu 24.04+) all have glibc ≥ 2.39.

## Changes

- `.github/workflows/linux-canary.yml:24` — container pin updated to `ubuntu:24.04@sha256:4fbb8e6a…`
- `.github/workflows/release.yml:479` — same container pin updated
- `.github/workflows/release.yml:501` — comment version string updated from 22.04 to 24.04

`fix-appimage.sh` and `desktop/src-tauri/**` are untouched. The #3573 fontconfig stopgap remains active; retirement is a separate follow-on PR once this fix is verified on a shipped build.

## Sequencing

[...]

Issue #2982, opened 2026-07-26
[Bug][Linux/AppImage] WebKitWebProcess aborts in Skia COLRv1 font rendering on Fedora 44


morler:
## Summary

Buzz Desktop `0.4.26` (`amd64` AppImage) aborts during startup on Fedora 44. The crashing process is the AppImage-bundled `WebKitWebProcess`, and the terminal output identifies an out-of-bounds assertion in Skia's COLRv1 color-font renderer:

```text
././/include/c++/12/bits/stl_vector.h:1123:
constexpr std::vector<...>::reference std::vector<...>::operator[](size_type):
Assertion '__n < this->size()' failed.

... colrv1_configure_skpaint(FT_Face, ..., const FT_COLR_Paint&, SkPaint*) ...
```

Removing color Emoji fonts from the fontconfig view presented to Buzz eliminates that assertion. It then exposes a second startup failure: the sandboxed WebKit process cannot discover Fedora's GStreamer plugins (`appsink`, `appsrc`, and `autoaudiosink`). A launcher that supplies a restricted non-color font configuration, Fedora's GStreamer paths, a private registry, and disables the WebKit sandbox allows Buzz to remain running without a new core dump.

The sandbox-disable requirement makes this a diagnostic workaround, not an acceptable permanent fix.

## Steps to reproduce

1. Install/run `Buzz_0.4.26_amd64.AppImage` on the environment below.
2. Start it from a terminal:

   ```bash
   chmod +x Buzz_0.4.26_amd64.AppImage
   ./Buzz_0.4.26_amd64.AppImage
   ```

3. The window disappears or fails to remain open.
4. Inspect the newest core dump:

   ```bash
   coredumpctl --no-pager list | tail
   coredumpctl --no-pager info <PID>
   ```

## Actual behavior

The bundled WebKit process terminates with `SIGABRT`:

```text
Signal: 6 (ABRT)
Executable: /tmp/.mount_Buzz_*/usr/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitWebProcess

#0 __pthread_kill_implementation
#1 raise
#2 abort
#3 std::__glibcxx_assert_fail
```

Running from a terminal reveals the assertion quoted in the summary. Setting `WEBKIT_DISABLE_COMPOSITING_MODE=1` does not change the failure, so this does not appear to be a GPU compositing crash.

The host color-font order is:

```text
$ fc-list ':color=true' -f '%{file}\t%{family}\n'
[...]

PR #3955, opened 2026-07-31
fix(ci): restore GLIBC 2.35 compatibility for deb packages

vjnrv:
## Summary
PR #3602 resolved the rendering-related `colrv1_configure_skpaint` crash (issues #2548 and #2982) on newer host distributions (like Ubuntu 24.04+ and Fedora 40+) by bumping the Linux AppImage build container to `ubuntu:24.04`. This ensured that the bundled WebKitGTK in the AppImage matched the `FT_ColorStopIterator` layout changes introduced in FreeType 2.13+.

However, upgrading the build container to `ubuntu:24.04` had the unintended side effect of raising the minimum GLIBC floor of the entire build pipeline to 2.39. As a result, the generated `.deb` packages (which were compiled inside the same container) now require `GLIBC_2.38` or higher, breaking installations on stable distributions such as **Ubuntu 22.04 LTS** (GLIBC 2.35) and **Debian 12** (GLIBC 2.36).

Because `.deb` packages do not bundle WebKitGTK (they use the system's `libwebkit2gtk` package), they are completely unaffected by the FreeType struct mismatch. The system-provided WebKitGTK is already compiled against the system-provided FreeType layout. Thus, there is no technical requirement to compile `.deb` packages on Ubuntu 24.04.

By splitting the Linux build environments, we can restore first-class support for Ubuntu 22.04 LTS and Debian 12 via the `.deb` package while fully preserving the AppImage fixes for newer hosts.

### Solution

- **Split Linux Build Pipelines**:
  - **`.github/workflows/linux-canary.yml`**: Refactored the `build` job to use a `strategy.matrix` separating target platforms.
    - `AppImage` is built in `ubuntu:24.04` to maintain FreeType 2.13+ layout compatibility.
    - `Debian` is built in `ubuntu:22.04` to restore the GLIBC 2.35 floor.
    - Updated steps (`tauri build`, `fix-appimage.sh` gating, artifact collection, and uploading) to be matrix-aware.
  - **`.github/workflows/release.yml`**: 
    - Isolated the existing `release-linux` job to only build `appimage` on `ubuntu:24.04`.
    - Added a new `release-linux-deb` job running on `ubuntu:22.04` to compile the `deb` package only.
    - Updated the `assemble-manifest` job to wait for both `release-linux` and `release-linux-deb` before completing the release draft.

### References

- Resolves the compatibility regression introduced in [PR #3602]

### Validation

- The workflows have been validated by running them on a fork:
  - The `AppImage` matrix job successfully compiled the AppImage package on `ubuntu:24.04`.
  - The `Debian` matrix job successfully compiled the `.deb` package on `ubuntu:22.04`.
  - The generated `.deb` package was downloaded and verified to install and launch without GLIBC errors on an Ubuntu 22.04 LTS system (GLIBC 2.35).

### DIFF
--- a/.github/workflows/linux-canary.yml
+++ b/.github/workflows/linux-canary.yml
@@ -18,10 +18,19 @@ permissions:
 
 jobs:
   build:
-    name: Build Linux canary
+    name: Build Linux canary (${{ matrix.target }})
     if: github.repository == 'block/buzz'
     runs-on: ubuntu-latest
-    container: ubuntu:24.04@sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90
+    strategy:
+      matrix:
+        include:
+          - target: AppImage
+            container: ubuntu:24.04@sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90
+            bundles: appimage
+          - target: Debian
+            container: ubuntu:22.04
+            bundles: deb
+    container: ${{ matrix.container }}
     timeout-minutes: 60
     permissions:
       contents: read
@@ -170,11 +179,12 @@ jobs:
           ./scripts/bundle-sidecars.sh
 
       - name: Build Linux Tauri app
-        run: cd desktop && pnpm tauri build --ci --bundles deb,appimage --config src-tauri/tauri.canary.conf.json
+        run: cd desktop && pnpm tauri build --ci --bundles ${{ matrix.bundles }} --config src-tauri/tauri.canary.conf.json
         env:
           CMAKE_POLICY_VERSION_MINIMUM: "3.5"
 
       - name: Fix AppImage (remove infra libs, symlink system GStreamer)
+        if: matrix.target == 'AppImage'
         # fix-appimage.sh checks for TAURI_SIGNING_PRIVATE_KEY and skips
         # re-signing when absent, so no signing env vars are needed here.
         # Repacking still runs, removing the Mesa/GLib/GStreamer conflict libs.
@@ -202,26 +212,26 @@ jobs:
           set -euo pipefail
           BUNDLE_DIR="desktop/src-tauri/target/release/bundle"
 
-          DEB=$(find "$BUNDLE_DIR/deb" -name '*.deb' -type f | head -1)
-          if [[ -z "$DEB" ]]; then
-            echo "::error::No DEB found in $BUNDLE_DIR/deb"
-            exit 1
+          if [ "${{ matrix.target }}" = "Debian" ]; then
+            DEB=$(find "$BUNDLE_DIR/deb" -name '*.deb' -type f | head -1)
+            if [[ -z "$DEB" ]]; then
+              echo "::error::No DEB found in $BUNDLE_DIR/deb"
+              exit 1
+            fi
+            echo "path=$DEB" >> "$GITHUB_OUTPUT"
+          else
+            APPIMAGE=$(find "$BUNDLE_DIR/appimage" -name '*.AppImage' -type f | head -1)
+            if [[ -z "$APPIMAGE" ]]; then
+              echo "::error::No AppImage found in $BUNDLE_DIR/appimage"
+              exit 1
+            fi
+            echo "path=$APPIMAGE" >> "$GITHUB_OUTPUT"
           fi
-          echo "deb=$DEB" >> "$GITHUB_OUTPUT"
-
-          APPIMAGE=$(find "$BUNDLE_DIR/appimage" -name '*.AppImage' -type f | head -1)
-          if [[ -z "$APPIMAGE" ]]; then
-            echo "::error::No AppImage found in $BUNDLE_DIR/appimage"
-            exit 1
-          fi
-          echo "appimage=$APPIMAGE" >> "$GITHUB_OUTPUT"
 
       - name: Upload Linux canary packages
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
-          name: buzz-linux-canary-${{ github.sha }}
-          path: |
-            ${{ steps.artifacts.outputs.deb }}
-            ${{ steps.artifacts.outputs.appimage }}
+          name: buzz-linux-canary-${{ matrix.target }}-${{ github.sha }}
+          path: ${{ steps.artifacts.outputs.path }}
           if-no-files-found: error
           retention-days: 7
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -647,11 +640,116 @@ jobs:
...
+          cd desktop && node scripts/set-version-from-tag.mjs "$VERSION"
+          cd src-tauri && cargo update --workspace
+
+      - name: Build sidecars
+        run: |
+          cargo build --release -p buzz-acp -p buzz-agent -p buzz-dev-mcp -p git-credential-nostr -p buzz-cli
+          ./scripts/bundle-sidecars.sh
+
+      - name: Build Linux Tauri app
+        run: cd desktop && pnpm tauri build --verbose --ci --bundles deb --config src-tauri/tauri.release.conf.json
```
wpfleger96, line 729: 🤖 Blocking: `src-tauri/tauri.release.conf.json` isn't committed — in `release-linux` it's produced by the "Generate release config" step (`build-release-config.mjs`, which also needs the `BUZZ_UPDATER_PUBLIC_KEY`/`BUZZ_UPDATER_ENDPOINT` env). This job has no such step, so it dies with a missing-config error at tag time. Copy that step (with its env) into this job.

Also, main's `release-linux` now builds with `--features mesh-llm` (#4524) — this line needs it too, or the shipped `.deb` loses Share Compute and re-breaks #3841.

.github/workflows/release.yml
```diff
@@ -647,11 +640,116 @@ jobs:
...
+      - name: Patch version
+        env:
+          VERSION: ${{ needs.setup.outputs.version }}
+        run: |
+          cd desktop && node scripts/set-version-from-tag.mjs "$VERSION"
+          cd src-tauri && cargo update --workspace
+
+      - name: Build sidecars
+        run: |
+          cargo build --release -p buzz-acp -p buzz-agent -p buzz-dev-mcp -p git-credential-nostr -p buzz-cli
```
wpfleger96, line 725: 🤖 Blocking: stale base — main's sidecar build now includes `-p buzz-backend-kubernetes` (#4289), so as written the `.deb` would ship without the k8s sidecar. This drift plus the build-line drift below is why I'd rebase first and copy the current `release-linux` steps rather than patching this snapshot.

.github/workflows/release.yml
```diff
@@ -647,11 +640,116 @@ jobs:
...
-            ${{ steps.linux-artifacts.outputs.deb }}
             ${{ steps.linux-artifacts.outputs.appimage }}
             ${{ steps.linux-artifacts.outputs.archive }}
             ${{ steps.linux-artifacts.outputs.sig }}
 
+  release-linux-deb:
+    name: Release Linux (Debian)
+    if: github.repository == 'block/buzz'
+    runs-on: ubuntu-latest
+    container: ubuntu:22.04
```
wpfleger96, line 651: 🤖 `ubuntu:22.04` is unpinned here, against the repo convention — the 24.04 container in `release-linux` is digest-pinned with "Renovate keeps it fresh". Pin the current jammy digest so Renovate maintains it and builds stay reproducible.

.github/workflows/linux-canary.yml
```diff
@@ -18,10 +18,19 @@ permissions:
...
     runs-on: ubuntu-latest
-    container: ubuntu:24.04@sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90
+    strategy:
+      matrix:
+        include:
+          - target: AppImage
+            container: ubuntu:24.04@sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90
+            bundles: appimage
+          - target: Debian
+            container: ubuntu:22.04
```
wpfleger96, line 31: 🤖 Same digest-pin ask as in `release.yml` — pin `ubuntu:22.04` to the current jammy digest so Renovate tracks it.

.github/workflows/linux-canary.yml
```diff
@@ -170,11 +179,12 @@ jobs:
           ./scripts/bundle-sidecars.sh
 
       - name: Build Linux Tauri app
-        run: cd desktop && pnpm tauri build --ci --bundles deb,appimage --config src-tauri/tauri.canary.conf.json
+        run: cd desktop && pnpm tauri build --ci --bundles ${{ matrix.bundles }} --config src-tauri/tauri.canary.conf.json
         env:
           CMAKE_POLICY_VERSION_MINIMUM: "3.5"
 
       - name: Fix AppImage (remove infra libs, symlink system GStreamer)
+        if: matrix.target == 'AppImage'
```
wpfleger96, line 187: 🤖 Nit: the `Install appimagetool` step above (~line 107) and the job-level `APPIMAGE_EXTRACT_AND_RUN` fallback still run in the Debian leg — the same `if: matrix.target == 'AppImage'` gate you added here would drop that unnecessary network/supply-chain work from the `.deb` build.

.github/workflows/linux-canary.yml
```diff
@@ -202,26 +212,26 @@ jobs:
...
-          echo "appimage=$APPIMAGE" >> "$GITHUB_OUTPUT"
 
       - name: Upload Linux canary packages
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
-          name: buzz-linux-canary-${{ github.sha }}
-          path: |
-            ${{ steps.artifacts.outputs.deb }}
-            ${{ steps.artifacts.outputs.appimage }}
+          name: buzz-linux-canary-${{ matrix.target }}-${{ github.sha }}
```
wpfleger96, line 234: 🤖 Heads-up, not blocking: the artifact name changes from `buzz-linux-canary-<sha>` to `buzz-linux-canary-<target>-<sha>` — anyone with scripted downloads of the canary artifact will need to update.

wpfleger96 (changes requested):
🤖 Thanks for the split — the shape is right. The `.deb` uses the system WebKitGTK and never needed the 24.04 bump, so building it on 22.04 restores the glibc 2.35 floor for Ubuntu 22.04/Debian 12, and the canary matrix conversion plus final-job aggregation are wired correctly.

Requesting changes for two release-breaking defects in the new `release-linux-deb` job (inline) plus a rebase: the branch is ~76 commits behind main and GitHub reports both workflow files conflicting. The tricky part is that `release.yml` never runs on PRs, so green PR CI can't catch either defect — both would first surface on the next `desktop-v*` tag.

After rebasing I'd mirror the current `release-linux` job's config-generation step, build flags, and sidecar list into `release-linux-deb` and keep them in lockstep (a comment in each job pointing at the other would help).

PR #3957, opened 2026-07-31
Open links in browser on Cmd/Ctrl+click

sw-square:
## Summary
- Cmd+click (macOS) / Ctrl+click (Windows/Linux) on a message link now opens it in the OS browser, matching Slack and Discord.
- Previously this gesture was a dead no-op: Buzz renders inside a WKWebView that silently swallows the modifier-click on a `target="_blank"` anchor — there is no browser tab to open into.
- A plain (unmodified) click keeps the existing native open-in-browser behavior; the right-click "Open link" / "Copy link" menu is unchanged.

## Implementation
The link click handler routes primary-modifier clicks through the OS opener via the existing `hasPrimaryShortcutModifier` helper, which already encodes the platform rule (Cmd on macOS, Ctrl elsewhere; Ctrl is rejected on macOS since it is the right-click there).

To keep the diff small and stay under the file-size ratchet, the external link anchor was extracted out of the 2k-line `markdown.tsx` into its own `ExternalLinkAnchor.tsx` (net **-82 lines** in `markdown.tsx`).

## Validation
- New unit tests for `hasPrimaryShortcutModifier` across macOS / Windows / Linux (7 cases).
- New E2E test (`messaging.spec.ts`) that sends a link, `ControlOrMeta`-clicks it, and asserts exactly one `plugin:opener|open_url` reaches the bridge — **passing**.
- `pnpm typecheck`, `biome check`, and the file-size / px-text / pubkey guards all green; full desktop unit suite passes.
- Note: Chromium (Playwright) doesn't reproduce WKWebView's modifier-swallow, so the underlying dead-gesture root cause was confirmed by reasoning about the native webview; the fix itself is verified end-to-end.

### DIFF
--- a/desktop/src/shared/lib/platform.test.mjs
+++ b/desktop/src/shared/lib/platform.test.mjs
@@ -1,0 +1,87 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { hasPrimaryShortcutModifier } from "./platform.ts";
+
+const originalNavigator = Object.getOwnPropertyDescriptor(
+  globalThis,
+  "navigator",
+);
+
+function withPlatform(platform, userAgent, run) {
+  Object.defineProperty(globalThis, "navigator", {
+    value: { platform, userAgent: userAgent ?? "" },
+    configurable: true,
+    writable: true,
+  });
+  try {
+    run();
+  } finally {
+    if (originalNavigator) {
+      Object.defineProperty(globalThis, "navigator", originalNavigator);
+    } else {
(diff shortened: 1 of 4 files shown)


wesbillman:
@sw-square when do you run into this? It seems like all links already open in a browser without having to CMD+Click.

sw-square:
@wesbillman I wanted to add the cmd+click feature in, which would be functionally identical to just normal clicking; my muscle memory from slack is very strong 😅

Issue #3959, opened 2026-07-31
relay: NIP-RS full-state-load conformance gaps — missing EOSE delivery barrier; #t post-filtered after cap


wpfleger96:
🤖 Filed as the relay-side follow-up to #2864 (NIP-RS manual-unread override layer), which defines five relay conformance preconditions for the full-state load procedure (`docs/nips/NIP-RS.md`, "Full-State Load" section). This issue covers the two gaps that still exist on `main` at [`209536ade`](https://github.com/block/buzz/commit/209536ade6c5ebf7fa82671d7ca0b74f599a40cc). Two earlier defects from the same investigation — the advertised `max_limit` of 10,000 vs the honored 1,000, and the dead `MAX_HISTORICAL_LIMIT` constant — were already fixed by #3635 (`nip11.rs` now advertises `buzz_db::DEFAULT_MAX_PAGE_LIMIT`, the same constant the REQ path clamps to, with a drift-guard test at `req.rs:1474`). They are not part of this issue.

## Defect A — no delivery barrier between event acceptance and a later REQ's EOSE

NIP-RS precondition 5 (the delivery barrier): before a relay sends EOSE for a query, every event it accepted *before that query read its stored events*, and which matches an open subscription on the same connection, must already have been delivered to that subscription.

The relay does not provide this today:

- `dispatch_persistent_event` (`crates/buzz-relay/src/handlers/event.rs:349`) intentionally returns after only the bounded audit enqueue; the whole delivery path — `pubsub.publish_event`, local fan-out, delivery metrics — runs inside `tokio::spawn` (`event.rs:374`).
- A REQ arriving after the event was durably accepted reads storage and sends EOSE (`crates/buzz-relay/src/handlers/req.rs:408`) with no ordering relationship to that spawned task.

Consequence for NIP-RS: a client running the fenced full-state load can receive EOSE for its final continuation while the push for an already-accepted replacement is still pending in the spawned task, and render a `complete` verdict in the interval between the two — exactly the race the barrier precondition exists to close. Until this is fixed, this relay does not satisfy the NIP-RS full-state-load preconditions, and `complete` verdicts against it are not sound.

[...]

PR #3967, opened 2026-07-31
feat(relay): push #t tag filters into SQL via JSONB containment (#3959)

iroiro147:
## Problem

NIP-01 `#t` filters (multi-value OR semantics) were post-filtered **after** the DB LIMIT bound. A page shorter than the page-size limit could therefore not be trusted as result-set exhaustion: matching events below the LIMIT boundary were invisible to the caller, and any client reasoning "short page ⇒ done" was unsound.

This made tag-constrained REQs unreliable for: subscription clients (NIP-RS full-state load), music/DJ decks (kind:30078 tag families), channel templates (#h,#t combinations), and any tool building local state from tag-filtered queries.

Refs block/buzz#3959.

## Fix

Push `#t` into SQL via `tags @> '[["t","<value>"]]'` JSONB containment, **OR-joined** across multi-values (matching NIP-01 filter semantics for multiple `#t` values on one filter). This mirrors the existing `#e` pushdown (event.rs:467) and reuses the `idx_events_tags_gin` GIN index (migrations/0004).

Changes:
1. **EventQuery** gains `t_tags: Option<Vec<String>>` (buzz-db/src/event.rs)
2. **query_events** appends the JSONB containment clause: `AND (tags @> '[["t","x"]]' OR tags @> '[["t","y"]]')` (buzz-db/src/event.rs)
3. **filter_to_query_params** extracts `generic_tags[#t]` into `t_tags` (buzz-relay/src/handlers/req.rs)
4. **filter_fully_pushable** now accepts `#t` (previously routed to the slow COUNT path with post-filtering)
5. **doc comment** updated to reflect `#t` pushability

`apply_count_fallback_limit` inherits `t_tags` from the shared EventQuery construction — no special-case needed.

## Tests

Three new regression tests in `handlers/req.rs`:
- `t_tag_multi_value_is_fully_pushable` — OR semantics accepted
- `t_tag_empty_values_still_pushable` — vacuous case (no constraint emitted)
- `filter_to_query_params_populates_t_tags` — SQL population verified

All 834 existing tests pass (9 pre-existing media/admin failures — same set fails on clean `main`).

## Note on Defect A

The separate `EOSE-delivery barrier` defect (Defect A in issue #3959) — events accepted during a query's EOSE window may be delivered late — requires relay-architectural work (coordinating `dispatch_persistent_event`'s spawned fins with EOSE emission). That is **not** addressed here; this PR fixes the **tag-filter** half.

### DIFF
--- a/crates/buzz-db/src/event.rs
+++ b/crates/buzz-db/src/event.rs
@@ -70,6 +70,10 @@ pub struct EventQuery {
     /// Restrict results to events with an `e` tag referencing any of these event IDs (hex).
     /// Uses JSONB containment (`tags @> ...`) against the `tags` column.
     pub e_tags: Option<Vec<String>>,
+    /// Restrict results to events with a `t` tag matching any of these values.
+    /// Uses JSONB containment (`tags @> ...`) against the `tags` column.
+    /// Multiple values use OR semantics (same as NIP-01 #t filter semantics).
+    pub t_tags: Option<Vec<String>>,
     /// Restrict results to events in any of these channels, while retaining
     /// channel-less global events. Applied before SQL `LIMIT` so access-filtered
     /// historical pages have exact exhaustion semantics.
@@ -121,6 +125,7 @@ impl EventQuery {
             authors: None,
             ids: None,
             e_tags: None,
+            t_tags: None,
             channel_ids: None,
             max_limit: None,
             shared_gated_reader: None,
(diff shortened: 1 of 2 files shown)


iroiro147:
Closing this — @wpfleger96 closed the linked issue #3959 on 2026-07-31 with "the fix lands as a PR we're about to open ourselves; no standalone issue needed."

Since the team is handling the EOSE delivery barrier and the `#t` post-filtering directly, this PR is superseded and there's no reason for it to sit in your review queue. Thanks — happy to help on the conformance suite elsewhere if that's useful.

PR #3968, opened 2026-07-31
feat(desktop): make onboarding model defaults skippable

tellaho:
**Category:** improvement
**User Impact:** Users can skip default model configuration during onboarding and finish it later in Settings → Agents.

**Problem:** Requiring model defaults during onboarding can block users who are not ready to choose a harness, provider, or model. Skipping also needs to leave existing configuration untouched rather than persisting partial selections.

**Solution:** Stage onboarding edits locally and persist them only when users choose Next or Back. A delayed Skip action advances without any configuration write, while a footer hint points users to the settings location for completing setup later.

<details>

**desktop/src/features/onboarding/ui/DefaultConfigStep.tsx**
Adds the skip action and future-settings hint, and makes model configuration transactional so Skip discards staged changes while Next and Back preserve the intended save behavior.

Exposes model-config setter call counts so tests can distinguish a true zero-write skip from a write-and-rollback implementation.

**desktop/tests/e2e/onboarding-agent-defaults.spec.ts**
Covers skipping during loading and after staged edits, verifies zero persistence calls, and confirms Next and Back still commit changes.

</details>

## Reproduction steps

1. Start fresh onboarding and continue through harness setup to **Configure your default model settings**.
2. Change the selected harness or model, then choose **Skip for now**.
3. Confirm onboarding advances to **Join or create a community** and the prior global model configuration remains unchanged.
4. Return through onboarding and confirm **Next** saves the staged selection; confirm **Back** also preserves staged changes before returning.
5. Confirm the footer says model defaults can be configured later in **Settings → Agents**.


tellaho:
@codex

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
```diff
@@ -330,25 +329,46 @@ export function DefaultConfigStep({
...
+            disabled={!persistenceState.canComplete || isCompleting}
+            onClick={() => void handleComplete()}
+            type="button"
+          >
+            Next
+          </Button>
+          <Button
+            className="absolute left-full ml-3 h-9 animate-in whitespace-nowrap rounded-full px-6 text-sm fade-in fill-mode-backwards [animation-delay:1000ms] animation-duration-[500ms] hover:bg-foreground/10 motion-reduce:animate-none"
+            data-testid="onboarding-config-skip"
+            onClick={actions.complete}
```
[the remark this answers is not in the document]
tellaho: Addressed in `c15c2899621a96deb9fae5eed9bcf8efd333191a`. Next and Back now snapshot/start persistence and navigate immediately, so the configuration page is no longer present while a request is in flight and Skip cannot be clicked after a commit starts. Skip remains a distinct zero-write path when selected before either committing action. A delayed-save E2E regression verifies navigation completes immediately while the selected configuration persists afterward. Cancellation was not used because the Tauri command writes in phase 1 before the potentially slow restart phase and exposes no safe abort path.

— Generated by Bart (AI)

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
```diff
@@ -275,23 +262,16 @@ export function DefaultConfigStep({
...
-      await persistenceState.flush();
-      actions.complete();
-    } catch {
-      setCompletionError("Couldn't save your default harness. Try again.");
-      setIsCompleting(false);
-    }
+  const handleBack = React.useCallback(() => {
+    void persistenceState.commit().catch(() => undefined);
+    actions.back();
   }, [actions, persistenceState]);
```
wpfleger96: 🤖 this trades the race from the last review round for silent data loss — `commit()` rejections get discarded and we navigate anyway, so a validation or disk-write failure marks onboarding complete while the config the user explicitly picked is silently gone. It also drops the previous `completionError` inline error + retry UI, so it's a regression rather than a gap.

The backend's `set_global_agent_config` is two-phase (validate + write disk, then agent restarts). I think the fix is to await the phase-1 durable write before navigating — it's a fast local write — and let the restart phase continue in the background. That keeps the immediate-unmount fix for the in-flight edit/Skip races from the last round while making failure visible again: on rejection, stay on the page with the error/retry UI. Worth an e2e case where `set_global_agent_config` rejects, too.

One extra wrinkle on `handleBack` specifically: Next is gated on `canComplete` (validity) but Back commits whatever is staged, so a staged-but-invalid config on Back sends a write the backend is guaranteed to reject — silently. Awaiting the write surfaces that path as well.

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
```diff
@@ -318,37 +298,49 @@ export function DefaultConfigStep({
...
           Back
         </Button>
+
+        <p className="text-xs text-foreground/50">
+          Configure default models in{" "}
+          <span className="text-foreground/70 underline underline-offset-2">
+            Settings → Agents
+          </span>{" "}
+          after setup.
+        </p>
```
wpfleger96, line 420: 🤖 nit: this is underlined like the actionable link on the previous setup page but renders as a plain `<span>` — either wire it to post-onboarding Settings → Agents navigation or drop the link affordance so it doesn't look clickable.

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
```diff
@@ -182,21 +163,29 @@ function AgentDefaultsSection({
...
+    } finally {
+      setIsSaving(false);
+    }
+  }, []);
   React.useEffect(() => {
     onPersistenceStateChange({
       // configIsValid comes from AgentConfigFields' onValidityChange and
       // covers model + provider credentials — a harness selection alone is
       // not a working default (e.g. buzz-agent with no provider configured).
       canComplete: selectedRuntimeId.length > 0 && configIsValid && !isSaving,
```
wpfleger96: 🤖 nit: once Next/Back await the commit, `isSaving` is nearly dead — commits only fire at navigation, so the `!isSaving` term here and the state plumbing become vestigial. Probably folds into whatever pending-state the error/retry UI needs.

wpfleger96 (changes requested):
🤖 hey @tellaho — thanks for this, the Skip path itself is really well built: genuinely zero-write, and the `get_global_agent_config_set_call_count` assertions would catch a write-then-rollback implementation, which is a nice test hook. Staging in refs also cleanly kills the autosave race class without the coalescer.

Requesting changes on two things (the main one is inline on `handleComplete`/`handleBack`):

1. Next/Back swallow persistence failure and navigate anyway — see the inline comment for details and the suggested fix direction.
2. `desktop/src/features/agents/AGENTS.md` needs updating in the same PR — its "keep this file true" section says a config-persistence behavior diff without a matching AGENTS.md diff should be treated as incomplete, and rule 7 still describes this page as choosing and persisting `preferred_runtime` behind a Finish gate. This PR moved that model to staged-commit-on-navigate + zero-write Skip, so rule 7 and the acceptance-test description need to match.

Also worth riding along here: `saveCoalescer.ts` and its test are now dead code — this PR removes the only consumer but both files survive (I checked, zero references outside the pair). Deleting them in this PR avoids the next contributor re-adopting the pattern this PR just retired.

One low-stakes note, no action needed beyond fix 1: Back fires the commit and navigates immediately, so re-entering the config step can read pre-write state while the write is still in flight and briefly show the pre-edit config. Awaiting the durable write before navigating shrinks that window to nearly nothing, so it comes along for free.

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
tellaho: Addressed in `3f7555c9768f1e19b1c20e3afb17edc3fbe817b6` (current HEAD `2efaebbd42b40f385a0832cb1d0bb343887307d6`). Next is now the only persistence boundary: it validates, disables the controls with a Saving… state, awaits the existing command, advances only on success, and retains the complete draft with an inline retryable error on failure. Back now preserves the complete or incomplete draft in onboarding session state without writing, so it no longer sends invalid drafts or races a write. Added E2E coverage for pending save behavior and failure/retry.\n\nUpdated by Bart 🤖

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
tellaho: Addressed in `3f7555c9768f1e19b1c20e3afb17edc3fbe817b6` (current HEAD `2efaebbd42b40f385a0832cb1d0bb343887307d6`): removed the underline/link affordance from the non-actionable Settings → Agents guidance.\n\nUpdated by Bart 🤖

desktop/src/features/onboarding/ui/DefaultConfigStep.tsx
tellaho: Addressed in `3f7555c9768f1e19b1c20e3afb17edc3fbe817b6` (current HEAD `2efaebbd42b40f385a0832cb1d0bb343887307d6`). The old save-coalescer machinery and tests were removed. The remaining saving state now directly drives the intentional pending UX: controls disabled, Saving… label, and save-error clearing/retry around Next.\n\nUpdated by Bart 🤖

tellaho:
@wpfleger96 Addressed your requested changes in `3f7555c9768f1e19b1c20e3afb17edc3fbe817b6` (current PR HEAD `2efaebbd42b40f385a0832cb1d0bb343887307d6` after merging latest `main`):

- **Next** is now the sole persistence boundary. It awaits the save, shows `Saving…`, advances only on success, and keeps the draft on-page with a retryable inline error on failure.
- **Back** now preserves the complete or incomplete draft in onboarding session state and returns without writing; re-entering restores that draft.
- **Skip** discards the session draft and advances with zero config writes.
- Updated `desktop/src/features/agents/AGENTS.md` to document the staged session model and matching acceptance coverage.
- Removed the dead `saveCoalescer.ts` and its tests.
- Added E2E coverage for Back restoration, zero-write Skip, and Next failure/retry.

All 24 onboarding defaults E2Es pass and the full PR CI is green. The three inline threads were also replied to and resolved.

— Generated by Ned, an AI coding agent

wpfleger96 (approved):
🤖 verified the fix commit (3f7555c9) against each item from the last round:

- Next now awaits `set_global_agent_config`, shows `Saving…`, advances only on success, and keeps the draft on-page with a retryable inline error — plus the new e2e rejection case (`setGlobalAgentConfigErrors`) covering fail-then-retry with call-count assertions
- Back is now a zero-write path with session-draft restoration, which also eliminates the ungated-Back write and the re-entry race entirely — a cleaner resolution than the await-on-Back we suggested
- `desktop/src/features/agents/AGENTS.md` rule 7 and the acceptance-test description now match the staged-draft model, including the durability contract (drafts don't survive restart)
- `saveCoalescer.ts` + its test are deleted with zero remaining references at head
- both nits addressed: the "Settings → Agents" underline is gone, and `isSaving` is now load-bearing (drives the Saving… state and disabled controls)

CI is green at `2efaebbd`. Approving.

Issue #3969, opened 2026-07-31
buzz-acp goes permanently silent after a burst of relay rate-limit NOTICEs — no crash, no further logs, restart required


uwebeyond:
**Environment:**
- Buzz.app 0.5.2 (macOS), `buzz-acp` / `@agentclientprotocol/claude-agent-acp` 0.63.0
- Relay: self-hosted `ghcr.io/block/buzz:main` (commit `bb34bc4`, 2026-07-31), no rate-limit env vars set at the time of the incident (relay defaults: `BUZZ_RATE_LIMIT_HUMAN_WS_EVENTS_PER_SEC=10`, fixed 5s window)

**Description:**

After a `buzz-acp` reconnect/resubscribe burst (e.g. right after an app restart, with multiple agent personas reconnecting near-simultaneously), the relay's fixed-window WS-event rate limiter (`BUZZ_RATE_LIMIT_HUMAN_WS_EVENTS_PER_SEC`, applied to agent connections too per `connection.rs`) frequently rejects with:
```
relay NOTICE: rate-limited: quota exceeded; retry in Ns
```
handled in `crates/buzz-acp/src/relay.rs` (`RelayMessage::Notice`), logged as:
```
WARN buzz_acp::relay: rate-limit gate armed via NOTICE until ~N.Ns from now
```
In most cases (~14/16 log samples we collected) the client recovers cleanly after the window. In a minority of cases, after several of these NOTICE/gate-arm lines fire in quick succession, the process goes **fully silent forever** — literally zero further log lines of any kind (not even at DEBUG), while remaining "online" in the UI and never processing another message. Only a full `Buzz.app` quit+relaunch recovers it.

We initially suspected a race in the gate re-arming (`set_rate_limit_gate`/`check_rate_gate`), but on reading the actual source, that logic is a cooperatively-polled `Option<Instant>` using `max()` to extend the deadline — not a droppable timer task — and there are dedicated unit tests for exactly this "NOTICE arrives while a retry is pending" scenario. So the gate itself looks correctly defended. We could not identify the actual hang point from the client-visible log alone; our best remaining guess is something downstream of the 30s ping-tick-triggered reconnect path (`try_autonomous_reconnect` / `wait_for_reconnect`, not fully traced by us) blocking without a timeout, past where this module's own logging would fire — but we have not confirmed this.

[...]

PR #3974, opened 2026-07-31
fix(acp): add idle-read watchdog to detect half-open silent sockets (#3969)

iroiro147:
## Summary

Fixes the silent-hang reported in #3969: `buzz-acp` going permanently silent (zero further logs, "online" in UI) after a relay rate-limit NOTICE burst, recoverable only by full app restart.

## Root cause

The existing silent-death detector is the ping/pong loop: `ping_interval.tick()` arms a `ping_sent` flag and `PONG_TIMEOUT` triggers a reconnect if no pong arrives. **But the select loop can never reach the ping arm while parked on `raw = ws.next()`.** If the socket goes half-open — TCP still connected but the relay stops delivering ALL frames (pongs included) — the loop waits on `ws.next()` indefinitely. The ping tick never fires, no log is ever emitted, and the agent is dead-but-online.

This matches the issuer's observation exactly: ~14/16 NOTICE/gate events recover cleanly, but a minority wedge the read path and produce literally zero further log lines.

## Fix

A new **idle-read watchdog** select arm, complementary to the existing ping/pong detector:

- `last_read: Instant` is reset in the `ws.next()` read arm on **any** inbound frame (pong, event, notice, close, or error) — so normal read activity keeps the watchdog dormant.
- A `sleep_until(last_read + IDLE_READ_WATCHDOG)` arm fires only when the socket has been completely silent for `90s` (3× `PING_INTERVAL` + `PONG_TIMEOUT`, so it strictly outlives a missed ping+pong window).
- On trip it warns, signals the caller via `try_send(None)` (never blocks on a full event channel), and runs the **same autonomous-reconnect path** a pong-timeout would, then resets `last_read`/`last_pong`/`connected_since`/`stable_logged`.

The watchdog is loop-scoped (`Instant`), so an `Instant::now()` anywhere in the loop body implicitly re-pins the sleep — no per-iteration pinning needed.

## Why not just shorten the ping interval?

The ping arm already implements silent-death detection *in principle*. The problem is reachability, not sensitivity: any frame on the socket resets it, and a socket that never errors/closes never lets the loop reach the ping branch. A timer that lives *outside* the read path is required.

## Test

`idle_read_watchdog_fires_only_when_socket_is_silent`:
- asserts `IDLE_READ_WATCHDOG > PING_INTERVAL + PONG_TIMEOUT` (only trips on a genuinely stuck socket);
- proves the `sleep_until(last_read + window)` math: a stale `last_read` resolves immediately (socket silent longer than the window), and a fresh `last_read` is not tripped by a shorter timeout.

All 82 `buzz-acp` relay tests pass with no regressions.

### DIFF
--- a/crates/buzz-acp/src/relay.rs
+++ b/crates/buzz-acp/src/relay.rs
@@ -51,6 +51,13 @@ const PONG_TIMEOUT: Duration = Duration::from_secs(10);
 /// Timeout for individual ws.send() calls. Prevents a stalled socket from
 /// wedging the background task indefinitely.
 const WS_SEND_TIMEOUT_SECS: u64 = 10;
+/// Idle-connection watchdog: if no websocket frame of ANY kind (pong, event,
+/// notice, close, error) arrives within this window, the read arm of the select
+/// loop is parked on a half-open socket and the ping loop can never fire —
+/// leaving the agent "online" but permanently silent (issue #3969). A healthy
+/// relay answers our 30s ping with a pong well inside this window, so any trip
+/// is a genuine stuck socket, not idle quiet.
+const IDLE_READ_WATCHDOG: Duration = Duration::from_secs(90);
 /// Diagnostic threshold: log when a connection has been stable for this long.
 /// The stability block resets `BgState::backoff_step` to 0 here so the next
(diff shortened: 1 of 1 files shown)


iroiro147:

PR #3978, opened 2026-07-31
fix(buzz-acp): capture startup watermark before agent pool init

AgileInnov8tor:
## Summary
Mentions that arrive while `buzz-acp` is still initializing its worker pool can be dropped: `startup_watermark` was captured **after** `initialize_agent_pool` (~10s for 10 workers), so the first channel REQ used `since = watermark - 5s` that still excluded process-start events.

## Change
Capture `startup_watermark` **before** `initialize_agent_pool` in `crates/buzz-acp/src/lib.rs`, matching the intended floor for cold-start Mentions replay.

## Verification
- Local cold-start smoke: owner mention during pool init → correlated agent reply after `agent_pool_ready`
- Desktop-managed Fizz on rebuilt binary answered `MANAGED_FIZZ_ACK_*` with `e`/`reply` tag

## Notes
- Does not replace signed Desktop app sidecars; consumers can point managed `acp_command` at a rebuilt binary until this ships in a release build.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -22,7 +22,7 @@ use acp::{AcpClient, EnvVar, McpServer};
 use anyhow::Result;
 use buzz_core::kind::{
     KIND_MEMBER_ADDED_NOTIFICATION, KIND_MEMBER_REMOVED_NOTIFICATION, KIND_STREAM_MESSAGE,
-    KIND_STREAM_REMINDER, KIND_WORKFLOW_APPROVAL_REQUESTED,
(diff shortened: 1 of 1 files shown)


loganrenz:
Packaged macOS Desktop v0.5.4 still exposed the cold-start window this PR addresses.

On a full desktop relaunch, the harness reported three channel subscriptions and online presence at 17:12:58Z, but the ten-worker Claude ACP pool did not report agent_pool_ready until 17:18:22Z. An owner DM delivered during that interval did not start immediately. After pool readiness and relay reconnection, it dispatched and produced a reply at 17:23Z.

This makes online presence a premature readiness signal and leaves operators vulnerable to sending work before the startup watermark can safely cover it. The proposed earlier-watermark change remains relevant; surfacing or gating on agent_pool_ready in the desktop UI would also make the state honest.

No credential values or private message bodies are included.

Issue #3929, opened 2026-07-31
test(relay): telemetry trace_context_lookup test flaky under parallel suite — reproduces on clean main


tlongwell-block:
## Summary

`telemetry::tests::trace_context_lookup_does_not_enable_callsites` fails intermittently under the parallel `buzz-relay` lib suite. It passes 100% of the time when run filtered/standalone. **Pre-existing on unmodified `main` — not introduced by any open PR.**

Same class as #2458 (`mesh_demo` echo test), and worth filing separately because it has a different mechanism: this one is global-state interaction, not a timeout.

## Failure

```
thread 'telemetry::tests::trace_context_lookup_does_not_enable_callsites' panicked at
crates/buzz-relay/src/telemetry.rs:489:13:
assertion failed: !tracing::enabled!(target: "trace_context_lookup_filter_test",
        tracing::Level::ERROR)
```

The test installs a `TraceContextLookup` layer with `LevelFilter::OFF` via `with_default` (a *thread-local* subscriber) and asserts no callsite is enabled. The failing assertion is that the filter did not suppress the callsite.

## Reproduction

Unmodified `main` at `61ba9dfaa00852925058d1a024322fa53663a5bc`, `cargo test -p buzz-relay --lib`, six consecutive runs:

| run | telemetry test |
|---|---|
| 1 | pass |
| 2 | pass |
| 3 | **FAIL** |
| 4 | pass |
| 5 | pass |
| 6 | pass |

**1/6 on clean main.** Filtered to `telemetry::` only, 6/6 pass — so it needs the rest of the suite running alongside it.

Also observed 2/6 on the #3844 branch head `6c8990b5fa200ef0369e520a3a5c0f880c9e3971`, which is what prompted the check. That diff touches `state.rs` / `api/mod.rs` / `invites.rs` and nothing telemetry-related; the clean-main reproduction above is the negative control.

## Likely mechanism (not confirmed)

`tracing` callsite *interest* is cached globally per callsite, not per-subscriber. Once any other test in the process registers a global default dispatcher, or a callsite's interest is cached as `always` while a different subscriber is active, a thread-local `with_default` filter can be bypassed — `tracing::enabled!` consults cached interest that outlived the other test's subscriber. Because rebuild/re-registration order varies with parallel scheduling, whether the poison lands before or after this test is a race.

[...]

PR #3981, opened 2026-07-31
test(relay): give trace_context_lookup test a per-statement unique target (#3929)

iroiro147:
## Summary

Fixes the intermittent failure of `telemetry::tests::trace_context_lookup_does_not_enable_callsites` under the parallel `buzz-relay` lib suite (#3929). The failure is pre-existing on unmodified `main` (not introduced by any open PR) and never fires when the test is run filtered or standalone.

## Root cause

`tracing` caches callsite *interest* **globally per static callsite**, not per-subscriber. The test installs a `TraceContextLookup` layer with `LevelFilter::OFF` via a thread-local `with_default`, then asserts the callsite is **not** enabled with `tracing::enabled!`.

When the full lib suite runs in parallel, another test can register a global default dispatcher and evaluate the same static callsite — caching its interest as `always` while a *different* subscriber is active. Because that cached interest outlives the other test's subscriber, the thread-local filter is bypassed when this test later runs, and `tracing::enabled!` returns `true`. Rebuild/re-registration order depends on parallel scheduling, so whether the poison lands before this test is a race.

## Fix (direction 2 — remove the shared state)

The reporter proposed three directions; this takes **(2)** — a per-run/unique callsite target so no other test can have poisoned this assertion's cached interest — rather than serializing around the shared state with a lock.

The assertion now targets a callsite unique to this test statement: `concat!("trace_context_lookup_filter_test_L", line!())`. `tracing::enabled!` requires a compile-time literal target, so this is an inline `concat!`/`line!` rather than a runtime `format!`. Since the callsite is distinct from every other target in the suite, no other test's subscriber activity can poison its cached interest, so the thread-local `LevelFilter::OFF` is always consulted.

This is a test-only change — no production behavior is altered.

## Verification

- Focused: `trace_context_lookup_does_not_enable_callsites` passes.
- `telemetry::` module: 7/7 pass.
- Full `cargo test -p buzz-relay --lib`: `trace_context_lookup_does_not_enable_callsites ... ok`. The suite still reports `793 passed; 9 failed; 35 ignored` — identical to clean-main baseline; the 9 failures are the pre-existing, unrelated `api::media` / `api::admin` panics and are unaffected by this change.

### DIFF
--- a/crates/buzz-relay/src/telemetry.rs
+++ b/crates/buzz-relay/src/telemetry.rs
@@ -480,6 +480,15 @@ mod tests {
                 .with_filter(tracing_subscriber::filter::LevelFilter::OFF),
         );
 
+        // `tracing` caches callsite *interest* globally per static callsite, not
+        // per-subscriber (see issue #3929): if a parallel test poisoned this
+        // callsite's cached interest while a different subscriber was active,
+        // `enabled!` consults that stale cache and a thread-local `with_default`
+        // filter is bypassed. Use a target literal unique to this test statement
+        // (suffixed by source line) so no other test's callsite can have poisoned
+        // this assertion's cached interest — removing the shared state instead of
+        // serializing around it. `enabled!` requires a compile-time literal, so
+        // this is an inline `concat!`, not a runtime `format!`.
         tracing::subscriber::with_default(subscriber, || {
             assert!(context_lookup
                 .dispatch
@@ -487,7 +496,7 @@ mod tests {
                 .and_then(tracing::dispatcher::WeakDispatch::upgrade)
                 .is_some());
             assert!(!tracing::enabled!(
-                target: "trace_context_lookup_filter_test",
+                target: concat!("trace_context_lookup_filter_test_L", line!()),
                 tracing::Level::ERROR
             ));
         });


iroiro147:
Re-verified against current `main` (`28ae6cd21`, 20 commits past this PR's merge-base). Rebased the branch locally onto current `main` and ran the #3929 repro: the `trace_context_lookup_does_not_enable_callsites` test now passes **3/3 full parallel `--lib` runs** (previously ~1/6 flaky). The only remaining failures are the 9 pre-existing env-coupled media/admin infra failures — baseline, unrelated, and separately addressed by #4356. This fix (per-statement unique callsite) is confirmed to hold against current main; #4356 and this PR are complementary (not conflicting). Branch refresh to current main is safe (clean rebase, single-file diff); leaving the push to avoid rewriting published review state.

iroiro147:
Rebased to current main as branch `claude/issue3929-v2` @ `c83d7530b` (single-commit cherry-pick of `0c7a88fa3` onto `28ae6cd21`; merge-clean). Verified on current main: full `--lib` now fails deterministically with exactly the 9 env-coupled media/admin tests (7 media + 2 admin — same set #4356 gates), and `trace_context_lookup_does_not_enable_callsites` is **not** among them under parallel load — i.e. the per-statement-unique-target mechanism holds on the moved base and removes the race.

This branch here (`claude/issue3929-20260731`) is still merge-clean against `28ae6cd21`, so either works for merge; the v2 is just the current-main-equivalent if a maintainer prefers a fresh head. I can't re-point this PR without a force-push, so if a rebase is wanted I'll close this and open the v2 branch as a fresh PR — or a maintainer can cherry-pick `c83d7530b` directly. Whichever is easiest for review.

PR #3990, opened 2026-07-31
feat(desktop): discover remote agent harnesses

dspury:
## Summary
- add read-only Tauri commands to list configured SSH hosts and probe a selected host for available agent harnesses
- reuse the existing local harness catalog to derive remote probe targets instead of maintaining a separate list
- add a conservative SSH config parser and strict host-key, batch-mode probing
- expose typed desktop bridge APIs for remote host discovery

## Why
Buzz can already discover harnesses on the local machine. This adds the same discovery shape for agents that are already running on another machine, without installing software, writing remote state, or collecting credentials.

## Validation
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml remote_probe --lib`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml ssh_config --lib`
- `just desktop-tauri-clippy`
- `git diff --check upstream/main..HEAD`


dspury:
Status update after reconciling the remote-agent stack against current upstream `main`: I am moving this root PR back to draft rather than force-rebasing stale architectural assumptions.

This SSH host-discovery design predates the current managed-agent discovery surface and the execution-binding/directory direction in #4174. Its dependent chain (#4006, #4029, #4144, and #4146) is already draft and remains parked. Before reviving the stack, the host identity and registration boundary needs to be reconciled with the current upstream model; then the work should be rebuilt as focused, independently reviewable deltas.

dspury:
Closing this draft because SSH host discovery and harness inspection are deployment concerns, not Buzz core responsibilities. The direction in #4174 is relay-native identity plus provider-owned execution; external adapters can use public Buzz contracts without machine discovery in this repository. The source branch is being retained.

PR #3991, opened 2026-07-31
fix(desktop): resolve typed @name mentions on send (parity with CLI/relay)

HyperfocuSam:
## What

Hand-typed `@Name` mentions in the desktop composer (typed in full instead of selected from the autocomplete picker) send with no `p` tag — the mention renders as plain text and the target is silently never notified. The relay's workflow sink and the CLI already resolve typed names ("uniquely resolved member names still notify"); the desktop send path was the one surface without this behavior.

## How

- Extracted `resolve_mention_pubkeys` from `buzz-relay/src/workflow_sink.rs` into shared `buzz-core/src/mentions.rs` (with its full test suite); relay now consumes the shared copy.
- `send_channel_message` (desktop) resolves typed `@Name` tokens against channel members via the existing member/profile hydration path and merges resolved pubkeys into the mention list — deduplicated against picker-resolved `mention_pubkeys` — across all three build paths (stream message, forum post, forum comment).
- Fail-open: member-lookup failure falls back to explicit mentions only; content without `@` skips resolution.

Semantics preserved exactly: case-insensitive, longest-name-first, left-boundary anchor, ambiguous display names wake no one, first-appearance dedup, unicode-safe folding.

## Testing

- `cargo test -p buzz-core`: 254 passed (includes the moved resolver suite + new cases)
- `cargo test -p buzz-desktop --lib`: 2004 passed, 0 failed (includes new merge/dedupe send-path tests)
- `cargo test -p buzz-relay --lib`: 775 passed; the 10 failures in `api::admin`/`api::media`/`api::mesh_demo` reproduce identically on untouched `main` (infra-tier tests, pre-existing)
- `cargo clippy -p buzz-core --all-targets -- -D warnings` clean; `cargo fmt --check` clean

Out of scope by design: a composer preview pill for typed mentions (frontend UX) — this PR is backend parity only.


Chessing234:
heads up — #4506 adds warn logging for ambiguous workflow @mentions on the same resolve path you're extracting into buzz-core. if your tip lands first, the warn should move with the shared helper.

HyperfocuSam:
Thanks for the heads-up — agreed, they collide.

I'd rather #4506 went first. Yours is a behaviour fix that stands on its own; this one is a structural move (323 lines out of `workflow_sink.rs` into `buzz-core::mentions`) that needs a maintainer to agree the crate boundary is the right one. No reason your warn should wait on that.

If #4506 lands first I'll rebase on top and carry the warn into the shared helper, so the relay keeps identical behaviour and the desktop and CLI paths inherit it for free. If a maintainer prefers the other order, say the word and I'll post the helper signature so you can write against it directly.

Issue #3651, opened 2026-07-30
Routed reader statements are unbounded after acquire — mid-transaction blackhole hangs requests indefinitely (needs reader statement_timeout)


tlongwell-block:
## Summary

Once a routed read has acquired its reader connection, **nothing bounds any subsequent statement**. A replica path that goes dark *mid-transaction* (LB failover, security-group change, standby promotion, Aurora failover) converts a routed read into an indefinitely hung request — the writer-fallback code never regains control. Dawn reproduced a hang that ran to her 15s probe cap with no bound in sight (thread 65279d98 in #buzz-read-only-replica-usage, 2026-07-30).

Found during the #3268 activation review. **Not a #3268 merge blocker** (seams are dark until `BUZZ_REPLICA_READ_MAX_AGE_MS` is set, and the writer pool has the same missing timeout today), but it is an **activation gate**: #3268's one-snapshot-across-page+aux design holds reader transactions open longer than any pre-existing read, widening the exposure window.

## Evidence (all at `dd26caa9f`)

What IS bounded — Dawn measured the acquire path exhaustively with a silent TCP listener (accepts, never writes — the hard blackhole shape):
- Cold connect through acquire: 151–152ms `PoolTimedOut`, 3/3 (sqlx `inner.rs:252-255` wraps the whole acquire future in `rt::timeout`; connect is re-bounded inside at `:338/:350`).
- Idle-connection-gone-stale: 152ms (sqlx `test_before_acquire` defaults true, ping runs inside the acquire timeout).

What is NOT bounded — after `read_pool.acquire()` succeeds:
- `MIDTX_STATEMENT elapsed_ms=15002 result=STILL_HANGING_at_15s_cap (UNBOUNDED)` — BEGIN succeeded on a healthy replica, path blackholed, next statement hung past the probe cap.
- No bound exists anywhere (verified independently by Dawn and Eva at the SHA): `rg statement_timeout crates/ -g '*.rs'` → zero hits repo-wide; the reader pool builder (`lib.rs:723-731`) sets only max/min/acquire/lifetime/idle, no `after_connect`; no `tokio::time::timeout` wraps any routed read. Everything in `proved_reader` after acquire — Aurora probe (`:908`), `Transaction::begin` (`:909`), `observe_heartbeat` (`:955`) — and every page/aux statement on the held snapshot is unbounded.

## Fix (small)

[...]

PR #3997, opened 2026-07-31
fix(db): bound reader-pool statements with per-session statement_timeout (#3651)

iroiro147:
## Summary

Bounds every statement issued on the reader pool (`crates/buzz-db`) with a per-session `statement_timeout`, so a replica that goes dark mid-transaction fails the statement instead of hanging the request indefinitely (#3651).

## Root cause

Once a routed read acquired its reader connection, nothing bounded any subsequent statement. A replica path that goes dark mid-transaction (LB failover, security-group change, standby promotion, Aurora failover) converted a routed read into an indefinitely hung request the writer-fallback could never reclaim. No `statement_timeout` existed anywhere in the repo; a `MIDTX_STATEMENT` probe ran past a 15s cap with no bound in sight.

## Fix

`connect_read_pool` now arms a per-session `statement_timeout` (`READER_STATEMENT_TIMEOUT_MS = 2s`) on every reader connection via `after_connect` + `set_config('statement_timeout', ..., false)` — the same `set_config` pattern already used by the writer pool's floor-guard at the same site. The value is deliberately session-scoped (the trailing `false`) so it cannot leak into any wider role config.

2s is comfortably above any legitimate routed read on a read-only session yet far below a hung request; on timeout Postgres sends an error, the acquire / read path fails, and the caller's fallback to the writer regains control. Collector/timer overhead and the (already sub-second) reader acquire are unaffected — this bounds only in-flight SQL work.

Reader sessions are read-only: this can never abort a committed write, it only bounds in-flight SQL.

## Test

Wiring test: `connect_read_pool_arms_statement_timeout` pins the bound value and proves lazy construction still works with `after_connect` armed (`connect_lazy` dials nothing). `buzz-db` lib suite: 95 pass / 0 fail (151 ignored — pre-existing Postgres-dependent skip-set).

## Environment note

No live replica available here (test-loader Postgres is absent in this worktree). The new wiring test passes standalone `cargo test -p buzz-db --lib`; this is the same pattern as the existing `connect_read_pool_is_lazy_and_independently_sized` test, which also never dials the network.

### DIFF
--- a/crates/buzz-db/src/lib.rs
+++ b/crates/buzz-db/src/lib.rs
@@ -706,6 +706,26 @@ impl Db {
     /// the reason names the mechanism rather than a diagnosis).
     const READER_ACQUIRE_TIMEOUT: Duration = Duration::from_millis(150);
 
+    /// Bound every statement issued on a reader connection (#3651).
+    ///
+    /// Once a routed read acquires its reader connection, nothing bounded
+    /// any subsequent statement: a replica path that goes dark
+    /// mid-transaction (LB failover, security-group change, standby
+    /// promotion, Aurora failover) converted a routed read into an
+    /// indefinitely hung request the writer-fallback could never reclaim.
+    /// With no `statement_timeout` anywhere in the repo, `MIDTX_STATEMENT`
+    /// hangs ran past a 15s probe cap.
+    ///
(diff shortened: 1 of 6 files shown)


iroiro147:

Issue #3996, opened 2026-07-31
We need an "Agent View" for remote agents


gdotbat:
### Is your feature request related to a problem? Please describe.

Buzz already supports running agents as headless `buzz-acp` bridges — a great fit
for always-on agents backed by runtimes like **Hermes** or **OpenClaw**, hosted
on a workstation/server and connected to a community relay 24/7, independent of
any desktop client.

The problem: the desktop **Agents** view only lists agents the local app
**manages/hosts** itself (`AgentsView` → `UnifiedAgentsSection`). Externally-hosted
bridge agents never appear there. They only surface as channel members (their
`kind:10100` profile badges them as "agent" in member lists and `@mention`
autocomplete), but there is no representation in the Agents area itself.

This forces a bad tradeoff:

- To see/manage an agent in the **Agents** area, the desktop app has to host it —
  which ties the agent's lifetime to that app instance being open, so true 24/7
  operation is impossible (close the app → the agent goes offline).
- Hosting the *same* agent on more than one device conflicts (two harnesses on
  one identity, racing on the relay).
- So anyone running Hermes/OpenClaw agents as always-on bridges has **no
  first-class, at-a-glance Agents surface** for them — only a per-channel member
  popup.

### Describe the solution you'd like

1. **A read-only "Community / External agents" section in the Agents view** that
   lists agents present in the community via their `kind:10100` agent profile
   (+ `kind:0`), regardless of which device or process hosts them — including
   bridge agents the local app does not manage.
2. **Clear separation of "Managed here" vs "External".** Externally-hosted agents
   get view/interact affordances (open profile, DM, `@mention`, online/offline
   presence) but **no** start/stop/edit/delete controls, since the local app
   doesn't own their runtime — and no "Configuration missing" / "Stopped" states
   for agents the app was never hosting.
3. **A uniform way to register/label an externally-hosted agent into the
   community's Agents roster from any device** — publish/curate its agent profile
   (name, avatar, description) without the app taking over hosting — so multiple
   people can each stand up their own always-on agents on their own machines and
   have them all appear consistently in everyone's Agents view for that community.

### Describe alternatives you've considered

- **Hosting agents via the desktop app (managed):** rejected — not 24/7 (the
  harness dies when the app closes), and hosting the same agent on multiple
  devices conflicts.
- **Publishing `kind:10100` manually** (e.g. via `buzz channels set-add-policy`):
  makes the agent show up in member lists / mentions, but **not** in the Agents
[...]

Issue #3054, opened 2026-07-27
[Bug] Agents created outside the current app-managed store do not appear in the Agents tab


jiyeonseo:
## Summary

Agents created through the external `buzz-acp` CLI, or through a different Buzz app bundle/profile, do not appear in the desktop app's Agents tab.

This makes agents that are present and active on the local relay look as if they were missing.

## Observed behavior

- The local relay contains the agent and the agent can participate in channels/respond through `buzz-acp`.
- The desktop app's Agents tab does not list that agent.
- The tab only shows agents registered in the current app's local managed-agent store.
- The production and development app bundles use separate managed-agent stores, so agents created in one are not automatically visible in the other.

## Expected behavior

The Agents tab should either:

1. discover and display agents associated with the current community/relay, including externally managed `buzz-acp` agents, or
2. clearly distinguish app-managed agents from externally managed agents and provide an explicit import/adopt flow.

If adopting an existing agent is supported, it should preserve the agent's existing identity where possible rather than silently creating a duplicate.

## Steps to reproduce

1. Start a local Buzz relay.
2. Create/configure an agent using the external `buzz-acp` CLI and connect it to the local relay.
3. Confirm that the agent is present in a channel and can respond to a message.
4. Open the Buzz desktop app connected to the same local relay/community.
5. Open the Agents tab.
6. Observe that the externally managed agent is not listed.

The same issue can be reproduced by creating an app-managed agent in a development build and then opening the production build.

## Environment

- Buzz Desktop: 0.4.26
- Local self-hosted relay
- External agent runtime: `buzz-acp`
- macOS

## Notes

The current behavior appears to come from the Agents view reading only the current app bundle's local managed-agent storage rather than merging relay-visible agents or offering an import path.

davidglogan:
Confirming this from a real deployment tonight: stood up an agent (Grok Build, xAI) purely via the external `buzz-acp` CLI on its own dedicated VM — never touched via the Desktop app at all, by design (our project's security model deliberately keeps every agent's process off the machine running the human's own Desktop client).

The agent is fully functional at the relay level — correct `role: "bot"` channel membership, responds to @mentions, introduces itself in group channels, DMs work — but is entirely absent from Desktop's Agents tab, exactly as described here. Took a bit of investigation to confirm this was expected/by-design rather than a misconfiguration on our end (traced it to `desktop/src-tauri/src/managed_agents/discovery.rs`, which is scoped to locally-launchable runtimes only).

[...]

PR #4006, opened 2026-07-31
feat(desktop): connect self-hosted agents

dspury:
Stacked on #3990. The incremental review for this PR is two commits:

- `feat(desktop): connect self-hosted agents Buzz does not own`
- `refactor(desktop): split initial window helpers`

## Summary

- add connected-agent storage separate from Buzz-managed agent records
- add read-only/connect Tauri commands for recording self-hosted agent identities by host, pubkey, name, and observed harness
- add the desktop Connect agent dialog and Connected agents section
- validate pubkey input so pasted secret keys are rejected and never echoed
- keep connected agents out of lifecycle/spawn fields by using a dedicated record type
- move existing initial-window helpers out of `lib.rs` so new command registrations stay within the desktop file-size guard

## Why

Some agents already run outside the Buzz desktop process. Buzz should be able to remember and display those identities without taking custody of their keys, command, process lifecycle, or environment.

The additional Tauri commands also pushed the app entrypoint across its file-size limit. The initial-window helpers now live in a focused module with no behavior change.

## Validation

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml remote_agent_connect --lib`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml connected_agents --lib`
- `just desktop-tauri-clippy`
- `cd desktop && pnpm check`
- `cd desktop && node --import ./test-loader.mjs --experimental-strip-types --test src/features/agents/ui/connectAgentIntent.test.mjs`
- `git diff --check origin/pr/remote-harness-discovery..HEAD`

## Related direction

- Part of the one-Agent / execution-binding direction discussed in #4174.
- Stacked on #3990; followed by #4029, #4144, and #4146.
- Addresses the host-owned identity visibility gap behind #3054 and supports the first-class remote Agent View direction in #3996.
- Complements the provider-managed remote-agent spec in #3748 by covering agents whose key and runtime remain owned by a self-hosted harness.


murdawkmedia:
I reviewed the connected-agent path at the current head (the latest `initial_window` extraction does not affect these findings). The separate public-only store and lack of local lifecycle controls are strong boundaries, but I think two issues need resolving before the UI calls this a connected agent:

1. **The host is not bound to the supplied identity.** `connect_remote_agent` accepts any valid pubkey for any configured SSH alias, and the UI intentionally allows submission after a failed/absent probe. Reachability plus an observed harness binary does not prove that the holder of that pubkey runs on that host. I would model this as an **unverified endpoint bookmark** unless a host-executed signed challenge or matching self-signed relay assertion establishes the binding. Tests should cover mismatched host/pubkey and stale/replayed proof.

2. **The SSH probe can disagree with OpenSSH and runs before explicit consent.** The partial parser ignores `Include`, `Match`, global/default precedence, inline-comment and first-obtained-value semantics, then passes its parsed port with `-p`, overriding the effective configuration. Prefer passing only a validated alias and deriving facts from `ssh -G alias`. Opening the dialog selects the first host and starts a POSIX-shell probe automatically; that invokes remote CLIs with `--version` before an explicit Probe action and does not support Windows SSH targets. Please make probing explicit and platform-aware.

The product-level split I would preserve is: community agents come from signed relay identity/membership facts; this file supplies only optional local endpoint annotations keyed by pubkey. That retains the good secret/lifecycle isolation here without making SSH a prerequisite for visibility or implying an unproven host↔identity relationship.

dspury:
@murdawkmedia this framing makes sense to me. I agree the current shape is closer to a local endpoint annotation/bookmark unless we add a real host↔identity proof.

I’m going to defer changes here for now because this crosses SSH discovery, identity proof, UI wording, and the still-open direction in #4174. I’ll keep this draft until that abstraction is clearer rather than trying to patch one piece in isolation.

dspury:
Follow-up after the current #4174 review: I am keeping this PR draft and am not patching the existing stack in place. The review findings still stand: the current code neither proves a host-to-pubkey binding nor matches OpenSSH resolution/explicit-consent semantics closely enough to label the record a connected agent.

The upstream direction now makes signed relay identity and directory facts the authoritative visibility path. If this local feature is revived, it should be redesigned as an optional endpoint annotation keyed by an already-known pubkey, use explicit user-triggered probing via effective OpenSSH configuration, and avoid implying identity proof without a fresh signed challenge. That is a contract change, not a safe incremental fix on the current conflicting branch.

dspury:
Closing this draft because persistent connected-agent records and SSH-backed lifecycle assumptions no longer match Buzz's relay-native remote-agent/provider direction in #4174. Any harness-specific connection logic will live outside Buzz and consume public contracts. The source branch is being retained.

PR #4008, opened 2026-07-31
fix: reauthenticate databricks model discovery

kalvinnchau:
## Summary
- preserve Databricks catalog 401 responses as authentication failures and retry discovery exactly once after silently refreshing the rejected bearer
- preserve runtime OAuth recovery: when discovery has no usable OAuth credential, `session/new` succeeds with only the trimmed configured model so the first `session/prompt` can run the existing browser PKCE flow
- reject a rejected configured `DATABRICKS_TOKEN` with actionable, non-interactive guidance; static credentials cannot recover through PKCE
- use the configured-model fallback for non-auth discovery failures without caching failed or fallback catalogs, so later sessions retry discovery
- keep known Databricks v2 models only for authenticated empty-catalog responses and mark their provenance
- resolve discovery before MCP spawn or session registration, preventing failed discovery from leaking resources or consuming session capacity
- permit serialized interactive PKCE only from the explicit saved-agent model picker; passive draft discovery never opens a browser

## Runtime flow
1. OAuth discovery attempts cached credentials and silent refresh without opening a browser.
2. If no usable OAuth bearer exists, `session/new` advertises only the configured model and succeeds.
3. The first `session/prompt` uses `TokenSource::bearer()`, which may launch browser PKCE.
4. A later session retries discovery and caches only the authenticated catalog.

## Regression coverage
- rejected-but-locally-fresh OAuth bearer performs one refresh and one catalog retry
- OAuth mode with no cached token allows `session/new` and returns exactly the trimmed configured model
- the OAuth fallback is not cached; a later authenticated session retries discovery and caches the returned catalog
- rejected static tokens still reject `session/new`
- failed discovery does not consume the sole session slot or spawn the supplied MCP process
- Desktop interactive/passive auth intent, static-token redaction, and authenticated empty-catalog provenance

## Verification
- `cargo test -p buzz-agent`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib commands::agent_models`
- `git diff --check`
- full pre-push hooks

## Review
Adversarial review found and drove fixes for session/MCP resource leakage, duplicate concurrent PKCE flows, sensitive error propagation, incorrect 403 reauthentication, missing discovery-level coverage, passive browser launch, and the Desktop file-size ratchet. The final follow-up preserves the existing prompt-time OAuth flow while retaining static-token rejection and pre-allocation discovery ordering.

### DIFF
--- a/crates/buzz-agent/src/catalog.rs
+++ b/crates/buzz-agent/src/catalog.rs
@@ -32,6 +32,18 @@ pub struct ModelEntry {
 pub const DATABRICKS_V2_KNOWN_MODELS: &[&str] =
     &["databricks-gpt-5-5", "databricks-claude-opus-4-7"];
 
+const AUTHENTICATED_EMPTY_CATALOG_SUFFIX: &str = " (known fallback)";
+
+fn authenticated_empty_v2_catalog() -> Vec<ModelEntry> {
+    DATABRICKS_V2_KNOWN_MODELS
+        .iter()
+        .map(|id| ModelEntry {
+            id: id.to_string(),
+            name: format!("{id}{AUTHENTICATED_EMPTY_CATALOG_SUFFIX}"),
+        })
+        .collect()
+}
+
 /// Heuristic: `true` when a v2 AI Gateway endpoint name looks like it serves
 /// chat/completions traffic.
 ///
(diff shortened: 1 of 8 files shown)


wpfleger96 (dismissed):
🤖 Approving after verifying the review feedback from #buzz-agent-databricks is addressed at this head (`4aaf16a3f`):

- The convergent blocker — `session/new` hard-failing on any discovery error — is fixed: only `LlmAuth` rejects; other catalog failures warn and fall back to the configured (trimmed) model, with the cache left empty so the next session retries discovery.
- Passive discovery can no longer launch a browser: intent is typed (`DatabricksAuthIntent::PassiveDraftDiscovery` vs `InteractiveModelPicker`) and only the explicit model-picker path may start PKCE.
- Copy/docs items resolved: `(default catalog)` suffix, reachable sign-in guidance for unsaved agents, auth-mode-appropriate remediation, deduped token read, corrected `catalog.rs` module docs.

All CI checks are green at this head.

wpfleger96 (approved):

PR #4010, opened 2026-07-31
fix(desktop): clarify inherited agent parallelism

wesbillman:
## Summary

- show an unambiguous `App default (10)` inherited state for parallelism in create and edit forms
- explain that blank inherits the app default and suppress create-form number steppers that could silently set `1`
- align the E2E mint fallback with production while preserving explicit input → definition → app-default precedence

## Why

The forms displayed `1` even though an untouched field is omitted and desktop minting materializes `10`. The create-form spinner could also turn blank/inherited into an explicit `1` with one click while leaving the field looking nearly unchanged.

## Testing

- `pnpm test` (desktop: 3,886 passed)
- `pnpm typecheck` (desktop)
- `pnpm check` (desktop)
- pre-push `desktop-check` and `desktop-test`


Chessing234:
app default (10) as the blank state is much clearer than a silent 1 from the stepper — the help text should stop a few of the "why is this 1?" reports.

wesbillman:
Updated UI for inherited agent parallelism. Leaving the create field blank now clearly shows that Buzz uses the app default of 10, while still allowing a custom value from 1–32.

Create agent advanced settings showing App default (10) for parallelism

_I’m commenting on Wes’s behalf._

PR #4015, opened 2026-07-31
feat(media): add sharded compatibility reads

bradseiler:
## Summary
Add deterministic hash-leading, community-scoped media key helpers and make every relay read/probe path prefer the sharded object while falling back to the legacy flat key only on absence. Public Blossom URLs and writes remain unchanged.

### Related issue
None found; requested in the linked Buzz channel thread.

### Testing
- `bin/cargo test -p buzz-media` (110 passed, 1 live-MinIO test ignored)
- `bin/cargo test -p buzz-relay --lib handlers::imeta` (11 passed)
- `bin/cargo clippy --workspace --all-targets -- -D warnings`
- `bin/cargo fmt --check`

### Stack
1. This PR: compatibility readers; safe to deploy with legacy writes.
2. Upload-record and storage-index compatibility.
3. Configurable legacy/dual/sharded writes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1043,6 +1043,7 @@ dependencies = [
  "image",
  "imagesize",
  "infer",
+ "metrics",
  "mp4",
  "nostr",
  "rust-s3",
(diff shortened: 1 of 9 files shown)


bradseiler:

PR #4016, opened 2026-07-31
feat(media): include payload keys in upload records

bradseiler:
## Summary
Add the exact payload object key to media upload records as an optional additive field. Existing consumers can continue deriving the legacy key from `sha256` and `ext`; tolerant consumers can prefer `blob_key` before sharded writes begin.

### Related issue
Stacked on #4015; requested in the linked Buzz channel thread.

### Testing
- `bin/cargo test -p buzz-media` (110 passed, 1 live-MinIO test ignored)
- `bin/cargo clippy -p buzz-media --all-targets -- -D warnings`
- `bin/cargo fmt --check`

### Stack
1. #4015: compatibility readers.
2. This PR: additive upload-record contract.
3. Configurable legacy/dual/sharded writes (follow-up after the moderation consumer accepts this field).

### DIFF
--- a/crates/buzz-media/src/upload_record.rs
+++ b/crates/buzz-media/src/upload_record.rs
@@ -60,8 +60,12 @@ pub struct UploadRecord {
     pub event_id: String,
     /// Content hash of the uploaded bytes (64 lowercase hex chars).
     pub sha256: String,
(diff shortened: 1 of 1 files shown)


bradseiler:

PR #4017, opened 2026-07-31
feat(media): add configurable sharded writes

bradseiler:
## Summary
Add `BUZZ_MEDIA_KEY_LAYOUT=legacy|dual|sharded` and route image, file, video, and thumbnail writes through the hash-leading community layout.

The upgrade-safe default remains `legacy`: existing and self-hosted deployments continue flat-only writes and do **not** start double-writing merely by upgrading. The ENV setting is exposed in the root and production Compose examples and as Helm `relay.mediaKeyLayout`; the chart validates all three stages. The Helm documentation now gives the explicit `legacy → dual → sharded` migration and rollback constraints.

Dual mode writes the sharded primary before the flat rollback copy, and sidecars publish only after both succeed.

### Related issue
Stacked on #4016 and #4015; requested in the linked Buzz channel thread.

### Testing
- `bin/cargo test -p buzz-media` (115 passed, 1 live-MinIO test ignored)
- `bin/cargo test -p buzz-relay --lib storage_sweep::tests` (15 passed)
- `bin/cargo clippy -p buzz-media -p buzz-relay --all-targets -- -D warnings`
- `bin/cargo fmt --check`
- Push hooks passed Rust suites and desktop Tauri checks

### Rollout
1. Upgrade with `BUZZ_MEDIA_KEY_LAYOUT` unset or `legacy`; all writes remain flat while compatibility readers deploy everywhere.
2. After consumers are ready, explicitly select `dual`, monitor read-resolution/fallback and duplicate-layout metrics, and backfill/reconcile.
3. After a full rollback window, select `sharded`.

Never roll a sharded-only writer back to a pre-#4015 binary. Returning `sharded → dual` does not itself recreate missing flat copies; verify backfill before relying on an old-version rollback.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -95,6 +95,13 @@ BUZZ_S3_BUCKET=buzz-media
 BUZZ_S3_REGION=us-east-1
 BUZZ_S3_ADDRESSING_STYLE=path
 
+# Media payload write migration. Unset defaults to legacy, so upgrades preserve
+# existing flat-only writes and do not duplicate objects automatically.
+# Stages: legacy (flat only) -> dual (sharded + flat rollback copy) -> sharded.
+# Deploy compatibility readers everywhere and backfill/reconcile objects before
+# advancing; do not roll sharded writers back to a version without sharded reads.
+# BUZZ_MEDIA_KEY_LAYOUT=legacy
+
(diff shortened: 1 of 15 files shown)


bradseiler:

Issue #3460, opened 2026-07-29
Channel disappears from sidebar after app restart, even after joining and messaging successfully (self-hosted relay, v0.5.0)


AIAlhello:
Hi team — first time filing here, so let me know if I've missed any convention. Wanted to share a clean repro of what looks like a real bug, plus some deeper diagnostic detail in case it's useful. As Claude has been helping troubleshoot, it generated the below report:

Environment

Buzz Desktop v0.5.0, macOS (reproduced on macOS 26.5.2 and 26.6)
Self-hosted relay via Docker Compose (deploy/compose/compose.yml), Postgres 17, single community
Reproduced with a freshly-created agent using the new "Bring Your Own Harness" flow (Hermes Agent), but also reproduces with plain human-only channels — doesn't seem to require an agent at all

Steps to reproduce

Create a new channel from the desktop app (e.g. #test2)
Add an agent to the channel via "Add agent to channel" (or just send a message as the human owner — reproduces either way)
Send a message; confirm it appears and the channel shows correctly in the left sidebar
Fully quit the app (Cmd+Q), then relaunch it

Expected behavior
The channel remains in the sidebar, and membership/join status persists across the restart.

Actual behavior
The channel disappears from the left sidebar entirely. It's still discoverable via Browse Channels, and message history is visible there — but the app shows it as not-joined and prompts to "Join" again, even though the join/message activity from before the restart is intact server-side.

Notes

This reproduces on v0.5.0, which already includes #3182 ("recover full local storage on startup") — so whatever that fix addressed, it doesn't seem to cover this case.
I did a deeper dive on a related-looking issue in an older channel (not the #test2 repro above, a separate one I'd been using) where I could verify the underlying data was correct: the channel_members Postgres row for my identity was present and correct (right community_id, no removed_at/hidden_at), and manually running the relay's own documented get_members() query returned the correct membership. But the relay's published member-list event (kind 39002) never included my pubkey regardless. I don't know if that's the same root cause as the sidebar issue above or a separate one, but flagging in case it's useful — happy to share the full steps for that if it'd help.

Happy to provide logs, a docker exec into the relay's Postgres, or anything else that would help narrow this down.

iroiro147:
Claiming this for a fix. Two candidate seams, need one piece of evidence to discriminate — posting diagnosis so the work is resumable.

## Candidate root causes

**Hypothesis A — sidebar knows the membership, but the restore-destination redirect makes it look gone.**

[...]

PR #4039, opened 2026-08-01
fix(relay): require all three NIP-29 discovery kinds before treating a channel as reconciled (#3460)

iroiro147:
## Summary

Tightens the boot-time channel-discovery reconcile so a channel is only considered "reconciled" when **all three** NIP-29 discovery events (metadata 39000, admins 39001, members 39002) are present — not just 39000. Fixes the #3460 symptom where a channel a user had joined disappeared from their Desktop sidebar after a relaunch.

## Root cause

Emission of the relay-signed discovery events is out-of-band and best-effort. Every site that changes membership or metadata *tries* to emit, but tolerates failure — e.g. `handle_join_request` `warn!`-logs and proceeds when `emit_group_discovery_events` errors (`side_effects.rs`). That means a channel can exist whose metadata event (kind:39000) was published at creation (when the member list was empty) while the members event (kind:39002) was rolled back or never written.

The reconcile pass at boot previously probed **only** for kind:39000 to decide whether a channel needed (re-)emission. Once 39000 existed, every subsequent startup classified the channel as reconciled and skipped re-emission — so a channel whose 39002 never landed stayed broken forever. The relay's published member list then omitted pubkeys the `channel_members` table held, the client's `get_channels({kinds:[39002],"#p":[my_key]})` query returned empty on cold boot, `isMember` derived `false`, and the sidebar dropped the channel behind a "Join" prompt. This matches the reporter's own follow-up note that the relay's 39002 member-list event never included their pubkey even though the DB row was correct.

## Fix

`reconcile_channel_events` now probes one addressable event per required kind (`REQUIRED_DISCOVERY_KINDS = [39000, 39001, 39002]`) for each channel, and re-emits the complete `emit_group_discovery_events` snapshot whenever any of the three is missing. Re-emission is idempotent (it replaces the current snapshot from the DB), so catching the "39002 was never written" shape also repairs the "39001 (admins) got lost" shape for free.

Fail-closed on a probe query error: the channel is skipped rather than re-emitting into a shaky DB (matches the existing error-tolerance of this boot path). Probe cost is one query per kind per channel, once per boot.

## Test

- `reconcile_probes_metadata_admins_and_members_kinds` pins `REQUIRED_DISCOVERY_KINDS == [39000, 39001, 39002]` (order + set + distinctness) so the revert can't regress silently.
- Full relay lib suite: **794 pass / 9 pre-existing env-gated failures** (`api::media`/`api::admin`, unchanged baseline; 0 media/admin files touched in this diff).

## Environment note

End-to-end validation (a scratch Postgres with a channel whose 39002 is absent) requires the live relay Postgres fixture this worktree lacks. The change is behaviorally additive: channels that were already fully reconciled are untouched; only previously-stranded 39001/39002-missing channels get an initial emission at boot.

### DIFF
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -3035,13 +3035,32 @@ pub async fn publish_nip43_member_removed(
 }
 
 /// Reconcile channels that exist in the DB but don't have kind:39000 events.
+/// Discovery kinds that must ALL be present for a channel to be considered
+/// reconciled. Probing only 39000 caused issue #3460: a channel whose
+/// metadata event was written but whose member-list (39002) or admin-list
+/// (39001) emission failed would pass the old check and forever surface a
+/// stale (or missing) member list to `#p` queries.
+const REQUIRED_DISCOVERY_KINDS: [i32; 3] = [39000, 39001, 39002];
+
+/// Reconcile per-channel NIP-29 discovery events signed by the relay.
 ///
-/// This handles the case where channels were created via direct SQL inserts
-/// (e.g. test seed scripts) rather than through the Nostr event pipeline.
-/// Emits kind:39000 (metadata) and kind:39002 (members) for each channel
-/// that is missing its discovery events.
+/// Emission is out-of-band and best-effort: every site that changes membership
+/// or metadata *tries* to emit, but tolerates failure (e.g. `handle_join_request`
+/// `warn!`-logs and proceeds on `emit_group_discovery_events` error). That means
+/// a channel can exist whose metadata event (kind:39000) was published (with the
+/// then-empty member list) while the members event (kind:39002) was rolled back
+/// or never written — and previously the reconcile pass only probed for 39000,
+/// so every subsequent startup classified the channel as reconciled and left
+/// 39002 missing forever. The result was exactly issue #3460: the relay's
+/// published member list omitted pubkeys that the `channel_members` table held,
+/// and the client's `#p` query for their memberships returned empty on cold boot.
 ///
-/// Idempotent: checks for existing kind:39000 events before emitting.
+/// The check now requires **all three** discovery kinds (metadata 39000, admins
+/// 39001, members 39002 — see [`REQUIRED_DISCOVERY_KINDS`]) to be present for a
+/// channel before it is considered reconciled. If any is missing we re-emit all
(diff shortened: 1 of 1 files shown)


AIAlhello:
Independent reproduction confirming this root cause, and confirming the bug is still live on current `main` — this PR isn't merged yet.

**Setup:** self-hosted relay (Docker Compose, `ghcr.io/block/buzz:main`), relay owner testing membership in a private channel they administer.

**Relay-level membership is correct** (rules out an identity/auth issue):

$ buzz-admin list-members
pubkey role created_at
aa45ab99... owner 2026-07-26T17:10:30Z
a3540917... member 2026-07-26T18:15:17Z


**Join event is accepted by the relay:**

$ buzz channels join --channel <redacted-channel-uuid>
{"accepted":true,"event_id":"d8d4047571bc8b07c1ab3a86f67e6db417fefcb9b7c54eca4e7a5e922d381eaf","message":""}


**Immediate re-query of the same channel omits the joining pubkey:**

$ buzz channels members --channel <redacted-channel-uuid>
[{"pubkey":"<other-member>","role":"owner"},{"pubkey":"<other-member>","role":"bot"}]


Tested on two builds several weeks apart — self-built from `main` @ `d8281b9`, and the current desktop-bundled CLI, v0.5.14 — same result on both, so this is not fixed by the unrelated membership work in #5493.

This lines up exactly with the root cause described here: the join's `kind:39002` emission apparently didn't land, and since boot-time reconcile only ever checks for `kind:39000`, the channel gets permanently marked "reconciled" and the missing member-list event is never retried. An accepted `event_id` with a DB write that never surfaces is a clean match for that failure mode.

Happy to test this fix against my own setup once merged, or provide relay logs / DB `channel_members` state if useful before then.

iroiro147:
@AIAlhello thank you — this is a particularly valuable repro because it rules out the two alternative explanations at the relay level: `buzz-admin list-members` showing correct membership eliminates identity/auth, and the accepted join event (`accepted: true` with an event id) eliminates write-path rejection. What remains is exactly what this PR targets: the discovery read path treating a channel as complete before all three NIP-29 kinds have arrived, so the immediate re-query serves a stale membership view.

Your confirmation that it's still live on current `main` with the stock `ghcr.io/block/buzz:main` image is exactly the kind of evidence that helps maintainers prioritise. The branch here is current and green; nothing needed from you — but if you get the chance to test this PR's build against your setup, the expected result is that the re-query right after join includes the joining pubkey.

PR #4040, opened 2026-08-01
fix(buzz-acp): never deliver ephemeral events to agent subprocesses

### LINKED ISSUE #3649: buzz-acp subscribe=all delivers typing/presence events to agents: token burn per keystroke and false typing badge
### Summary

When buzz-acp runs with `BUZZ_ACP_SUBSCRIBE=all` and no `BUZZ_ACP_KINDS` override, ephemeral events are delivered to the agent session. Every typing indicator (kind 20002) and presence update (kind 20001) wakes the agent and starts a turn. Two effects:

1. Token cost per keystroke. Each wake-up is a model call even when the agent decides to do nothing. A human composing one message produces dozens of typing events, so the agent pays for dozens of no-op turns.
2. A confusing feedback loop in Desktop. While the agent processes the wake-up, Desktop shows it as typing (the working badge in `desktop/src/features/agents/agentWorkingSignal.ts` falls back to bot typing signals and active turns). The human sees "agent is typing" whenever they themselves type, before the agent has anything to say.

### Reproduction

1. Run buzz-acp with `BUZZ_ACP_SUBSCRIBE=all` (the documented way to make an agent respond to un-mentioned channel messages) and leave `BUZZ_ACP_KINDS` unset.
2. Open the channel in Desktop as a human member and start typing a long message.
3. Watch the agent flip to "typing..." while you type, and watch the harness logs record a turn per typing event.

### Expected

Ephemeral kinds (20001, 20002, and arguably 24200 observer frames) should never reach the agent subprocess, or at minimum should be excluded by default when `subscribe=all` resolves with no kinds override. Mentions mode already ships a curated default (`resolve_channel_filters` in `crates/buzz-acp/src/config.rs` uses KIND_STREAM_MESSAGE, KIND_WORKFLOW_APPROVAL_REQUESTED, KIND_STREAM_REMINDER); All mode passes `kinds: None` and subscribes to everything.

### Workaround

Set an explicit allowlist, for example:

```
BUZZ_ACP_KINDS=9,40002,40003,5,7,1059,40100,40007,46010,46020
```

This keeps messages, edits, deletions, reactions, DMs, canvas updates, reminders, and workflow events while dropping the keystroke echoes. Confirmed working on a self-hosted deploy (compose stack from `deploy/compose`).

### Suggested fix

In `resolve_channel_filters`, give `SubscribeMode::All` a default exclusion of ephemeral kinds when `kinds_override` is `None`, or introduce a small default allowlist mirroring the mentions-mode one plus reactions and canvas. Happy to send a PR if maintainers agree on the direction.

iroiro147:
## Problem

When an agent runs `subscribe=all` (or any wildcard subscription rule) with no curated `BUZZ_ACP_KINDS` allowlist, the resolved rule set admits **every** kind on the wire — including ephemeral kinds:

- **Kind 20001** — presence updates
- **Kind 20002** — typing indicators
- **Kind 24200** — agent observer frames

Every keystroke in a channel emits dozens of these ephemeral events. Each one woke the agent subprocess into a full model turn: **cost per keystroke** for the operator and a false **"agent is typing"** signal in Desktop for whoever is watching.

Closes #3649.

## Root Cause

`resolve_channel_filters` in `config.rs` resolves `SubscribeMode::All` with no `kinds_override` to `kinds: None` — a wildcard that matches every event kind. There is no design-level invariant that ephemeral events (kind 20000–29999, defined as "never stored" in `buzz-core/src/kind.rs`) should be excluded from agent delivery. The `match_event` function in `filter.rs` applied channel-scope, kind, mention, and expression filters — but had no concept of ephemeral exclusion.

## Fix

Add a design-level invariant to the shared delivery filter (`crates/buzz-acp/src/filter.rs`, `match_event`): **ephemeral events (kind 20000–29999) are never delivered to an agent subprocess**, regardless of any subscription rule that might otherwise admit them.

This guard runs before the rule loop, so it protects:
- `SubscribeMode::All` with no `kinds_override` (the reported case)
- Any wildcard `Config` rule with empty kinds
- Even an explicit `All` subscription that accidentally lists an ephemeral kind

Ephemeral events are transient by definition — they are not actionable work and must not wake an agent process. Failing closed here (never wake on any ephemeral kind) keeps wildcard subscriptions safe to use.

The guard uses the existing `buzz_core::kind::is_ephemeral` predicate (already a dependency of `buzz-acp`), so there is no new constant or duplicate logic.

## Testing

New test `test_match_event_ephemeral_never_wakes_agent_on_wildcard`:
- Builds a wildcard rule (empty kinds = match any kind) — the shape an `All` subscription with no curated list resolves to
- Asserts that kind 20001 (presence), 20002 (typing), and 24200 (observer frame) all return `None` (no match)
- Control: asserts that a non-ephemeral kind 9 (stream message) still matches the wildcard rule — the guard only rejects ephemeral kinds

Full buzz-acp lib test suite: **662 passed, 0 failed**.

## Backward Compatibility

Agents that explicitly subscribe to ephemeral kinds via `BUZZ_ACP_KINDS` will no longer receive them. This is the intended behavior — ephemeral events were never meant to be actionable, and any agent depending on them was paying a per-keystroke cost without realizing it. The fix is fail-closed: if an agent truly needs ambient channel activity, a future enhancement could add an opt-in `include_ephemeral` flag, but the default must be safe.

### DIFF
--- a/crates/buzz-acp/src/filter.rs
+++ b/crates/buzz-acp/src/filter.rs
@@ -373,6 +373,26 @@ pub async fn match_event(
 ) -> Option<MatchedRule> {
     let filter_ctx = FilterContext::from_event(event, channel_id);
 
+    // Ephemeral events (kind 20000–29999) are never delivered to the agent,
+    // regardless of any subscription rule that might otherwise admit them.
+    //
+    // When an agent runs `subscribe=all` (or a wildcard `Config` rule, or an
+    // `All` mode with no explicit `BUZZ_ACP_KINDS` allowlist) with no curated
+    // kind list, these events would otherwise wake the agent subprocess on
+    // every keystroke: a human composing one message emits dozens of typing
+    // indicators (kind 20002) and presence updates (kind 20001), each of
+    // which starts a model turn at cost per keystroke and flashes Desktop's
+    // "agent is typing" badge. Observer frames (kind 24200) are caller-local
+    // telemetry on the same channel and belong to the same exclusion.
+    //
+    // Per the design intent behind `SubscribeMode::All`, agents want all
+    // *actionable* kinds, not ambient per-keystroke noise — which these
+    // ephemeral events are. Failing closed here (never wake on any ephemeral
+    // kind) keeps the wildcard subscription safe to use.
+    if buzz_core::kind::is_ephemeral(event.kind.as_u16() as u32) {
+        return None;
+    }
+
     for (index, rule) in rules.iter().enumerate() {
         // 1. Channel scope check.
         if !rule.channels.matches(&channel_id) {
@@ -679,6 +699,44 @@ mod tests {
 
         let result = match_event(&event, channel_id, &rules, "").await;
         assert!(result.is_none());
+    }
+
+    /// #3649: ephemeral events (kind 20000–29999 — typing, presence, observer
+    /// frames) must never wake an agent subprocess, even when a wildcard rule
+    /// would otherwise admit them. A `subscribe=all` agent sees the whole
+    /// event stream; without a guard, each keystroke fires a model turn.
+    #[tokio::test]
+    async fn test_match_event_ephemeral_never_wakes_agent_on_wildcard() {
+        let channel_id = any_channel();
+        // A wildcard rule (empty kinds = match any kind) — the shape an
+        // `All` subscription with no curated list resolves to.
+        let rules = vec![make_rule(
+            "all-kinds",
+            ChannelScope::All("all".into()),
+            vec![], // wildcard: no kind filter
+            false,
+            None,
+            None,
+        )];
+
+        // Typing indicator (20002), presence update (20001), and observer
+        // frame (24200) — every ephemeral kind must be rejected with no match.
+        for ephemeral_kind in [20001u32, 20002, 24200] {
+            let event = make_event(ephemeral_kind, "typing...");
+            let result = match_event(&event, channel_id, &rules, "").await;
+            assert!(
+                result.is_none(),
+                "ephemeral kind {ephemeral_kind} must never match any rule"
+            );
+        }
+
+        // Control: a non-ephemeral kind (kind 9, stream message) matches the
+        // same wildcard rule — the guard only rejects ephemeral kinds.
+        let event = make_event(9, "hello");
+        let matched = match_event(&event, channel_id, &rules, "")
+            .await
+            .expect("non-ephemeral kind 9 must match the wildcard rule");
+        assert_eq!(matched.rule_index, 0);
     }
 
     #[test]


AhmedAburady:
Confirming this from a self-hosted deploy (buzz pinned at `d40a3329`, three agents on one
relay). The ephemeral guard is the right shape and it closes the keystroke half of #3649: we saw
1,063 relay rejections of `invalid: reaction target event not found`, which is agents trying to
👀 typing/presence events that were never stored. This PR removes that entirely.

One thing worth flagging while the PR is open, not as a request to widen it: the guard is scoped
to kinds 20000-29999, and in a multi-agent channel the larger burn comes from kinds **5** and
**7**, which are regular and stored. buzz-acp publishes 👀 at queue-push (`lib.rs:2212`) and 💬
before the prompt (`pool.rs:1881`), then `ReactionGuard` deletes both on turn end - four events
per picked-up event, all into the same channel, all delivered to any other wildcard subscriber.
So the agents wake on each other's bookkeeping and it does not converge.

On our relay that produced 25,717 kind-7 events against 304 real messages, with **22,373 of the
reactions targeting another reaction**. Roughly 20x the typing-indicator volume. It also does not
need `respond_to=anyone`: the author gate admits same-owner siblings in `owner-only` mode by
design, so two agents from one human on `subscribe=all` hit it on stock config.

I do not think that belongs in this PR - reactions can be genuinely actionable, unlike ambient
typing, so the fix is a different one. Filed separately as #4086 with the numbers and the source
trace. Raising it here only because after this merges `subscribe=all` will look fixed, and for
multi-agent setups the expensive half is still open.

Also, small correction for anyone using the workaround in #3649: the recommended
`BUZZ_ACP_KINDS=9,40002,40003,5,7,1059,40100,40007,46010,46020` includes `5` and `7`, which is
the storm fuel above. It is fine for a single agent; with more than one, dropping those two is
what actually stops it.

iroiro147:
@AhmedAburady Thanks for the field numbers — the 22,373 reaction-on-reaction count against 304 real messages quantifies the 'does not converge' claim better than my abstract argument could.

You were right that it didn't belong in this PR. The fix for the 5/7 storm half is now up separately as **#4091** against your **#4086**: when `BUZZ_ACP_KINDS` is unset, both `resolve_channel_filters()` and `resolve_dynamic_channel_filter()` default to the curated list `[9, 40002, 40003, 1059, 40007, 40100, 46010, 46020]` — exactly the list you validated, minus 5 and 7. The `BUZZ_ACP_KINDS` override stays authoritative, so single-agent setups that deliberately want kinds 5/7 can still opt in.

Your point about the workaround list in #3649 being storm fuel for multi-agent setups is addressed the same way — once #4091 merges, dropping 5/7 becomes the default rather than a manual exclusion.

The ephemeral guard here (20000-29999) + the curated default in #4091 should cover both halves of the multi-agent wake-storm problem: ambient typing/presence *and* reactions/deletions.

iroiro147:
Thanks for the production confirmation — really useful to see it closing the rejection loop in the wild, and helpful to know the keystroke half specifically is what this heals.

On kinds 5 / 7 (stored reactions/deletion): agreed they're a larger class of cross-agent traffic, and the current guard (20000–29999 ephemeral per NIP-01) intentionally leaves them alone. Two reasons to keep this PR scoped narrowly: (a) kinds 5/7 are *stored* events — relay rejection on them isn't a typo-shaped invariant the way it is for ephemeral, it's a semantic question about whether agents should `react`/`delete` each other's content at all, which probably wants its own policy/gate; (b) widening the guard in this PR would mix a hard NIP-01 invariant with a design choice that deserves a separate discussion. Happy to open a follow-up issue if you'd like to take the 5/7 question explicitly — the deployment evidence you have (1,063 rejections on the narrow class) would be valuable context for that one too.

iroiro147:

PR #4043, opened 2026-08-01
fix(acp): stop cancelling healthy turns on busy/stale steer; use Git bash in Windows tests

axincha51234-commits:
## What

Three commits fixing `buzz-acp` steering reliability and the Windows test harness:

1. **fix(acp): don't cancel a healthy turn when the steer channel is busy** — A burst of mentions while a turn was in-flight filled the capacity-1 steer channel; `TrySendError::Full` was lumped in with transport errors, so the universal cancel+merge fallback fired and destroyed the running turn. `send_steer` now maps `Full` to `SteerError::SteerChannelBusy` and `try_native_steer` returns a tri-state; on Busy the event stays queued and is delivered by normal dispatch.

2. **fix(acp): skip cancel+merge fallback when a steer ack targets a stale turn** — A late `SteerAck` (its target turn already completed and a newer turn dispatched on the channel) could fire the cancel+merge fallback against the healthy newer turn. The ack now carries the steer-time `turn_id`; a stale ack's fallback is skipped with a `tracing::warn!`.

3. **test(acp): spawn harness scripts on Git bash, not the WSL shim** — `Command::new("bash")` on Windows resolves via CreateProcessW search order, which checks `System32` before PATH, so it launched the WSL launcher instead of Git-for-Windows bash. WSL bash cannot redirect into `C:/...` paths and boots slowly enough to trip the 800ms steer idle timeout — the root cause of the deterministic steer-capture failures and the 18–24 test flake bucket. The harness now resolves a Git bash explicitly, keeps the canonicalize + `\\?\`-strip capture path handling, and widens two idle budgets.

## Verification

- `cargo test -p buzz-acp --lib` — **663/663 pass** on the rebased branch
- `cargo clippy --workspace` — clean
- `cargo build` (workspace) — clean

## Notes

- Rebases cleanly onto current `main` (`b1b283cd4`).
- The `.gitignore` change for local relay autostart logs was deliberately kept out of this PR.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -2846,8 +2846,33 @@ mod tests {
         );
     }
 
+    /// Locate a bash that behaves like a real POSIX shell for harness scripts.
+    ///
+    /// `Command::new("bash")` on Windows resolves through `CreateProcessW`'s
+    /// search order, which checks `C:\Windows\System32` before any PATH entry.
+    /// On machines with WSL installed that yields `System32\bash.exe` — the
+    /// WSL launcher — instead of the Git-for-Windows bash sitting earlier in
+    /// PATH. WSL bash mounts the filesystem at `/mnt/c`, cannot redirect into
+    /// `C:/...` paths, and boots slowly enough to trip the harness's idle
+    /// timeout, which makes every `spawn_script`-based test fail or flake.
+    /// Prefer a Git-for-Windows bash when one is installed.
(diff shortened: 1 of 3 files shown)


KhoaPhan8I:
@block/buzz-oss-team — could a maintainer please approve the workflow runs for this fork PR and do a code-owner review? All three reachable checks (Semgrep OSS, zizmor, DCO) pass and the branch is rebased on current main (b1b283cd4). Thanks!

PR #4044, opened 2026-08-01
fix(db): surface silent partition catch-all collapse with warn + metric (#4033)

### LINKED ISSUE #4033: Partition manager cannot reclaim a pre-existing catch-all partition — monthly partitioning silently stops
**Summary**

Once a fresh deployment's initial migration boundary is crossed, `ensure_future_partitions` can never create the "current month" partition again, because the migration's own `*_p_future` catch-all already covers that range. The collision is caught and swallowed as a benign, expected case — so partitioning for that range silently and permanently stops, with no error, no metric, and no operator-visible signal beyond a Postgres server-log line most people don't watch.

**Where**

- `migrations/0001_initial_schema.sql` (events: lines ~237-252, `delivery_log`: lines ~343-354) hardcodes monthly partitions with literal dates authored mid-2026, ending with a catch-all:
  ```sql
  CREATE TABLE events_p_future PARTITION OF events
      FOR VALUES FROM ('2026-07-01') TO (MAXVALUE);
  ```
  Any fresh install that runs this migration in July 2026 or later will have every monthly partition `ensure_future_partitions` tries to create fall inside `_p_future`'s range and collide. This isn't an edge case — it will recur for every fresh install going forward, and again at the *next* catch-all boundary for any host that hand-repairs it (e.g. re-basing to `2027-01-01` just moves the wall to 2027-01).

- `crates/buzz-db/src/partition.rs` (~lines 130-149) explicitly anticipates this:
  ```rust
  match sqlx::query(sqlx::AssertSqlSafe(sql)).execute(pool).await {
      Ok(_) => { info!("added partition {partition_name}"); Ok(()) }
      Err(sqlx::Error::Database(db_err))
          if db_err.code().as_deref() == Some("42P17")
              && db_err.message().contains("would overlap partition") =>
      {
          // Fresh schemas include a right-edge catch-all partition (`*_p_future`).
          // If it already covers this month, the table is still safe for writes;
          // treat the overlap as "ensured" rather than failing startup.
          info!(partition_name, "partition range already covered by an existing partition");
          Ok(())
      }
      Err(e) => Err(e.into()),
  }
  ```
  Postgres error `42P17` ("would overlap partition") is caught and logged at `info`, then treated as success. Postgres itself still logs a server-side `ERROR` for the rejected DDL (independent, unsuppressable), but the application never surfaces it, never fails startup, and gives no indication that the monthly partition was never actually created.

**Impact**

The monthly partition never gets created; nothing ever shrinks or re-splits the existing catch-all. Every row for the colliding range keeps landing in `_p_future` indefinitely, defeating the purpose of monthly partitioning (pruning/archival/maintenance by partition boundary). There's no code anywhere in `partition.rs` or in migrations `0002`-`0026` that ever reclaims or splits an existing catch-all partition once one exists.

**Reproduction**

A regression test reproducing this against real Postgres (`current_month_partition_is_never_created_once_a_future_catchall_exists`, `crates/buzz-db`) is here, passing in CI: https://github.com/frodoHost/buzz/pull/1

**Suggested fix direction** (not implemented)

Either detach-and-recreate the overlapping range of `_p_future` when a monthly-partition collision is detected, or raise the `info!` to a `warn!`/metric so this isn't purely a Postgres-log-only symptom.

iroiro147:
## Summary

Fixes #4033.

Once a fresh deployment's initial migration boundary is crossed, `ensure_future_partitions` silently stops creating the "current month" partition. The migration's `*_p_future` catch-all already covers the range (`FOR VALUES FROM ('2026-07-01') TO (MAXVALUE)`), so every monthly `CREATE PARTITION` collides with Postgres `42P17` ("would overlap partition"). The collision was caught and logged at `info`, then treated as success — the monthly partition is **never** created, **never** recreated, and the only signal is an unsuppressable Postgres server-log line most operators never watch. Every row for the range keeps landing in the growing catch-all, defeating monthly pruning/archival/maintenance-by-boundary.

This is **not** an edge case: it recurs for every fresh install on/after the migration boundary month, and again at the next catch-all boundary for any host that hand-repairs it by re-basing.

## Fix — Option B (smallest reviewable change)

The issue proposes two directions; this PR implements the low-risk one (operator-visible signal), leaving catch-all detach/reattach surgery as a follow-up:

- **`is_partition_overlap_error`** — extracts the `42P17` / `"would overlap partition"` classification into a pure predicate, unit-testable without a live Postgres connection.
- **`info!` → `warn!`** on the overlap arm, with an explicit "**monthly partition was NOT created**" message naming the catch-all and the remedy (re-base or split), replacing the benign "range already covered" phrasing.
- **`buzz_db_partition_catchall_coverage` counter** (labelled by `table`) so the silent collapse is alertable independent of log level.
- **Still returns `Ok`** — the table remains write-safe and startup must not fail; only the observability changes.

## Tests

- `overlap_predicate_classifies_42p17` — PG-free predicate contract.
- `catchall_metric_emits_under_local_recorder` — verifies the counter emits with the correct `table` label via `metrics-util`'s `DebuggingRecorder`.

`buzz-db` lib: **96 passed / 0 failed / 152 ignored (PG-gated)** — no regressions.

## Notes

- A real-Postgres reproducer (the `current_month_partition_is_never_created_once_a_future_catchall_exists` `#[sqlx::test]` referenced in the issue) can land here or as a follow-up; it is PG-gated and skipped in this environment. The predicate + metric tests are PG-free and cover the changed logic.

### DIFF
--- a/crates/buzz-db/src/partition.rs
+++ b/crates/buzz-db/src/partition.rs
@@ -4,9 +4,46 @@
 
 use chrono::{Datelike, TimeZone, Utc};
 use sqlx::{PgPool, Row};
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::error::{DbError, Result};
+
+/// True when a `sqlx::Error` is a Postgres `42P17` ("would overlap partition")
+/// raised by a `CREATE PARTITION` whose range is already covered. Fresh schemas
+/// include a right-edge catch-all (`*_p_future`), so a fresh install collides
+/// on the "current month" boundary it tries to add. Split out of
+/// `ensure_partition` so the classification is unit-testable without Postgres.
+fn is_partition_overlap_error(err: &sqlx::Error) -> bool {
+    match err {
+        sqlx::Error::Database(db) => {
+            db.code().as_deref() == Some("42P17")
+                && db.message().contains("would overlap partition")
+        }
+        _ => false,
+    }
+}
+
+/// Record #4033: a monthly partition attempt was absorbed by the `*_p_future`
+/// catch-all (monthly partition NOT created). Metric so operators can alert
+/// independent of log level.
+fn catch_all_coverage_metric(table: &'static str) {
+    metrics::counter!(
+        "buzz_db_partition_catchall_coverage",
+        "table" => table,
+    )
+    .increment(1);
+}
+
+/// Resolve a validated table name to its static allowlist entry so the metric
+/// label escapes the caller's stack. `ensure_partition` already validated the
+/// name against `PARTITIONED_TABLES`; `.expect` is unreachable for valid input.
+fn static_table_name(table: &str) -> &'static str {
+    PARTITIONED_TABLES
+        .iter()
+        .copied()
+        .find(|t| *t == table)
+        .expect("table name validated against PARTITIONED_TABLES")
+}
 
 /// Tables that may be partition-managed. Allowlist prevents DDL injection.
 const PARTITIONED_TABLES: &[&str] = &["events", "delivery_log"];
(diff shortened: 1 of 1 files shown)


iroiro147:

PR #4045, opened 2026-08-01
feat(desktop): make macOS close behavior a user preference (#4024)

### LINKED ISSUE #4024: feat(macos): make close-to-tray behavior configurable
## Motivation

On macOS, closing Buzz's main window leaves the process and local agents running but hides the only main window. This is useful for background agent work, but it is currently unconditional and has no user-facing preference.

A user who uses BetterTouchTool's application/window switcher with **Exclude Hidden and Minimized Windows** enabled sees Buzz disappear from the switcher even though Buzz is still running. This makes Buzz feel "windowless": it remains active in the menu bar/process list, but it cannot be returned through the user's normal app-switching workflow.

BetterTouchTool exposes this exclusion as a global switcher option; its documented configuration does not provide a per-app override that can re-include one hidden app.

## Current behavior

The macOS close handler in [`desktop/src-tauri/src/lib.rs`](https://github.com/block/buzz/blob/main/desktop/src-tauri/src/lib.rs#L958-L969) unconditionally:

1. calls `api.prevent_close()`;
2. calls `window.hide()`.

The tray menu's **Open Buzz** path restores the main window with `unminimize()`, `show()`, and `set_focus()` in [`desktop/src-tauri/src/tray_menu.rs`](https://github.com/block/buzz/blob/main/desktop/src-tauri/src/tray_menu.rs).

I searched the current settings/source and found no preference that changes this close-to-tray behavior.

## Proposed behavior

Add a macOS preference for what happens when the main window is closed:

- **Keep Buzz running in the menu bar** — current behavior and recommended default, preserving active agent work.
- **Quit Buzz when the last main window closes** — for users who prefer conventional visible-window app switching and do not want a windowless background process.

If quitting would stop active agent sessions or a huddle, Buzz should warn or require confirmation before exiting.

At minimum, the UI should explain that closing the main window hides Buzz to the menu bar and that **Open Buzz** restores it.

## Reproduction

1. On macOS, open Buzz.
2. Close the main window with the red close button.
3. Confirm Buzz remains running and its menu-bar item remains available.
4. Open BetterTouchTool's switcher with **Exclude Hidden and Minimized Windows** enabled.
5. Buzz is absent because its main window is hidden.
6. Use the Buzz menu-bar item → **Open Buzz**; the main window returns.

## Related

- #2412 requests background/tray mode so closing the window does not kill agent sessions. This request is the complementary macOS preference: keep that capability as the default, but make it optional and explicit.
- BetterTouchTool Window Switcher configuration: https://docs.folivora.ai/docs/actions/action-definitions/#window-switcher

## Environment

- Buzz Desktop for macOS
- BetterTouchTool Window Switcher with **Exclude Hidden and Minimized Windows** enabled

iroiro147:
## Summary

On macOS, closing Buzz's main window unconditionally called `prevent_close()` and hid the window, leaving the process running with no visible window. That preserves background local-agent work — but for users on window switchers that exclude hidden windows (e.g. bettertouchtool), Buzz disappears from the switcher while still running, with no way back through normal window switching.

Introduce a persisted **close behavior** preference with three modes:

- **Keep running in background** (default): unchanged — close hides the window, Buzz and local agents keep running (reopen via the tray icon).
- **Minimize to Dock**: close minimizes the window so Buzz stays visible to window switchers that exclude hidden windows, keeping agents running.
- **Quit when window closes**: close quits Buzz like a conventional app.

The default is unchanged (`keepRunning`), so existing installs behave identically until the user opts in.

Fixes #4024.

## Backend (`desktop/src-tauri`)

- New `close_to_tray` module: versioned JSON settings (`close-to-tray.json`), load/save via `atomic_write_json_restricted`, `load_for_app` fallback to `keepRunning`, plus `get_close_to_tray_behavior` / `set_close_to_tray_behavior` Tauri commands.
- `AppState` gains `close_to_tray_behavior` (resolved once in `setup()` from `load_for_app`); the macOS `WindowEvent::CloseRequested` handler branches on it instead of unconditionally `prevent_close()` + `hide()`. `QuitWhenClosed` deliberately skips `prevent_close()`.
- Commands registered in `invoke_handler`.

## Frontend (`desktop/src`)

- `closeToTrayLogic`: behavior union + option metadata + validation (node-tested); `invoke` is resolved lazily so pure validation stays unit-testable without the Tauri bridge.
- `CloseToTraySettingsCard`: macOS-only (renders `null` elsewhere and outside the Tauri shell) radio dropdown in the **Agents** settings panel with per-option descriptions. Optimistic update with reload-on-failure.

## Verification

| Check | Result |
| --- | --- |
| `cargo check` (desktop src-tauri) | 0 warnings |
| `cargo test --lib close_to_tray` | 5/5 pass |
| `cargo clippy --lib` (close_to_tray) | clean |
| frontend validation logic (`node --experimental-strip-types`) | valid (default valid, 3 unique options, junk rejected, safe fallback) |
| diff audit (7 files, +515/−5; sidecar stubs excluded) | clean |

**Backend tests cover:** missing file → default, save/load round-trip, unversioned → default fallback, newer-version → error, invalid JSON → error.

**Not verified here:** full frontend typecheck/`npm test` — desktop deps are not installed in this environment (pre-existing baseline gap: `package.json`'s `npm test` needs the test-loader's `typescript` dependency installed). The macOS close behavior itself was not exercised on a live build (no local desktop run); the handler branch logic is a direct 1:1 mapping from preference to `prevent_close`/`hide`/`minimize`.

## Notes for reviewer

- `QuitWhenClosed` skips `prevent_close()` entirely, matching the requested "quit like a normal app" semantics; this means background agents stop until Buzz is reopened — called out in the card copy.
- Windows/Linux are untouched: the close handler and card are both `#cfg(target_os = "macos")` / `isMacPlatform()` gated, preserving prior cross-platform behavior.

### DIFF
--- a/crates/buzz-db/src/partition.rs
+++ b/crates/buzz-db/src/partition.rs
@@ -4,9 +4,46 @@
 
 use chrono::{Datelike, TimeZone, Utc};
 use sqlx::{PgPool, Row};
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::error::{DbError, Result};
+
+/// True when a `sqlx::Error` is a Postgres `42P17` ("would overlap partition")
+/// raised by a `CREATE PARTITION` whose range is already covered. Fresh schemas
+/// include a right-edge catch-all (`*_p_future`), so a fresh install collides
+/// on the "current month" boundary it tries to add. Split out of
+/// `ensure_partition` so the classification is unit-testable without Postgres.
+fn is_partition_overlap_error(err: &sqlx::Error) -> bool {
+    match err {
+        sqlx::Error::Database(db) => {
+            db.code().as_deref() == Some("42P17")
+                && db.message().contains("would overlap partition")
+        }
+        _ => false,
+    }
+}
+
+/// Record #4033: a monthly partition attempt was absorbed by the `*_p_future`
+/// catch-all (monthly partition NOT created). Metric so operators can alert
+/// independent of log level.
+fn catch_all_coverage_metric(table: &'static str) {
+    metrics::counter!(
+        "buzz_db_partition_catchall_coverage",
+        "table" => table,
+    )
+    .increment(1);
+}
+
+/// Resolve a validated table name to its static allowlist entry so the metric
+/// label escapes the caller's stack. `ensure_partition` already validated the
+/// name against `PARTITIONED_TABLES`; `.expect` is unreachable for valid input.
+fn static_table_name(table: &str) -> &'static str {
+    PARTITIONED_TABLES
+        .iter()
+        .copied()
+        .find(|t| *t == table)
+        .expect("table name validated against PARTITIONED_TABLES")
+}
 
 /// Tables that may be partition-managed. Allowlist prevents DDL injection.
 const PARTITIONED_TABLES: &[&str] = &["events", "delivery_log"];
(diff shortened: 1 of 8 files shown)


iroiro147:
Superseded by #5031 — rebased onto current main (conflict resolution merged the huddle::agent_voice handlers added in #5004 with our close_to_tray handlers in desktop/src-tauri/src/lib.rs). Closing per no-force-push cadence.

PR #4048, opened 2026-08-01
test(desktop): pin ACP availableModels merge for #3934 (#3934)

### LINKED ISSUE #3934: Model dropdown stays empty for ACP harnesses that report models via `session/new` → `models.availableModels` (e.g. Hermes Agent)
## Summary

Buzz discovers a harness's models by reading `configOptions` from the `session/new`
result. ACP agents that instead report models via the `models` field
(`result.models.availableModels`) — and emit no `configOptions` — are invisible to
that code path. The UI falls back to "Using built-in model options. Could not load
live models for this provider."

Hermes Agent (a bundled preset harness, `hermes-acp`) is affected: it returns **154
models** in `result.models.availableModels` and **no `configOptions` at all**, so
Buzz shows zero live models for it.

Both mechanisms are legitimate ACP. `configOptions` is the generic
session-configuration mechanism; `models` / `availableModels` is the dedicated model
field. Buzz currently supports only the former.

## Environment

| | |
|---|---|
| Buzz Desktop | v0.5.2 (Linux, extracted AppImage) |
| OS | Fedora Linux 44.1.7 (Sway Atomic), Wayland |
| Hermes Agent | v0.19.0 (2026.7.20) |
| OpenCode | 1.18.10 |
| Claude Code | 2.1.220 via `@agentclientprotocol/claude-agent-acp` 0.64.0 |
| ACP protocolVersion | 1 (all three agents) |

## Expected vs actual

**Expected:** With Default harness = Hermes Agent, the Model dropdown lists Hermes's
available models.

**Actual:** Dropdown is empty except built-in options, with:
> Using built-in model options. Could not load live models for this provider.

Reached via Settings → Agents → Agent defaults → Default harness = "Hermes Agent".

## Root cause

Diffing the `session/new` response across the three harnesses installed here:

| Harness | `session/new` result keys | `models.availableModels` | `configOptions` | Dropdown populates |
|---|---|---|---|---|
| OpenCode (`opencode acp`) | `sessionId`, `configOptions` | 0 | ✅ present (`id: "model"`, `category: "model"`) | ✅ yes |
| Claude Code (`claude-agent-acp`) | `sessionId`, `modes`, `configOptions` | 0 | ✅ present | ✅ yes |
| **Hermes (`hermes-acp`)** | `_meta`, `models`, `modes`, `sessionId` | **154** | ❌ **absent** | ❌ **no** |

The two harnesses that work both expose models through `configOptions`. The one that
fails exposes them through `models.availableModels`. This lines up with
`resolve_model_switch_method()` reading the `configId` key from each `session/new`
`configOptions` entry (see commit 925a9a7, "fix(buzz-acp): accept id-keyed config
options when resolving model switch") — there appears to be no corresponding reader
for the `models` field.

### Raw responses

OpenCode — models under `configOptions`:

```json
{
  "sessionId": "…",
  "configOptions": [
    {
      "id": "model",
      "name": "Model",
      "category": "model",
      "type": "select",
      "currentValue": "opencode/big-pickle",
      "options": [
        { "value": "opencode-go/deepseek-v4-flash", "name": "OpenCode Go/DeepSeek V4 Flash (New)" },
        { "value": "opencode-go/deepseek-v4-pro",   "name": "OpenCode Go/DeepSeek V4 Pro" }
      ]
    }
  ]
}
```

Hermes — models under `models`, no `configOptions`:

```json
{
  "sessionId": "…",
  "_meta": { "hermes": { "sessionProvenance": { … } } },
  "modes": { … },
  "models": {
    "currentModelId": "nous:tencent/hy3:free",
    "availableModels": [
      { "modelId": "nous:anthropic/claude-opus-5",   "name": "Nous Portal · anthropic/claude-opus-5",   "description": "Provider: Nous Portal" },
      { "modelId": "nous:anthropic/claude-sonnet-5", "name": "Nous Portal · anthropic/claude-sonnet-5", "description": "Provider: Nous Portal" }
    ]
  }
}
```

## Reproduction

Minimal ACP stdio client — no Buzz required:

```python
import json, subprocess, time
p = subprocess.Popen(["hermes-acp"], stdin=subprocess.PIPE, stdout=subprocess.PIPE,
                     text=True, bufsize=1)
def send(o): p.stdin.write(json.dumps(o) + "\n"); p.stdin.flush()
def rd(i):
    while True:
        m = json.loads(p.stdout.readline())
        if m.get("id") == i: return m

send({"jsonrpc":"2.0","id":1,"method":"initialize",
      "params":{"protocolVersion":1,"clientCapabilities":{}}}); rd(1)
send({"jsonrpc":"2.0","id":2,"method":"session/new",
      "params":{"cwd":".","mcpServers":[]}})
r = rd(2)["result"]
print("keys:          ", list(r.keys()))
print("availableModels:", len(r.get("models", {}).get("availableModels", [])))
print("configOptions: ", r.get("configOptions", "ABSENT"))
```

Output here:

```
keys:           ['_meta', 'models', 'modes', 'sessionId']
availableModels: 154
configOptions:  ABSENT
```

Swap `hermes-acp` for `opencode acp` to see the inverse.

## Ruled out

These were each tested and are **not** the cause:

- **Discovery timeout.** Hermes's ACP `initialize` was slow (6.1s) because it connects
  MCP servers during init. Reducing that to **0.65s init / 2.0s total** — faster than
  OpenCode's 2.6s, which works — did **not** fix the dropdown.
- **Network/DNS.** All model endpoints resolve and respond normally.
- **`authMethods` gating.** OpenCode also advertises a non-empty `authMethods` and
  populates fine, so a non-empty `authMethods` is not what blocks Hermes.
- **Binary resolution.** `hermes-acp` resolves on the login-shell PATH and the
  handshake succeeds; Buzz spawns it without error.

## Suggested fix

In the model-discovery path, fall back to `result.models` when `configOptions` has no
`category: "model"` entry:

- Read `result.models.availableModels[].modelId` / `.name` for the option list.
- Read `result.models.currentModelId` for the current selection.
- For switching, ACP defines `session/set_model` (Buzz already logs
  `session/set_model: model overridden model_id=…` for its own `buzz-agent` runtime),
  so the switch path may largely exist already.

Supporting both would fix Hermes and any other agent using the dedicated `models`
field, without affecting the `configOptions` harnesses.

## Workaround

The "Custom model…" option plus a hand-typed model ID (e.g.
`nous:anthropic/claude-opus-5`) does work — Buzz passes it through even though it
can't enumerate it. So this is a discovery/enumeration gap only, not a plumbing one.

iroiro147:
## Summary

Closes #3934.

The issue reports that ACP harnesses reporting models via the unstable `models.availableModels` field (e.g. Hermes Agent, which returns 154 models and *no* `configOptions`) show an empty Model dropdown in Buzz Desktop, with the warning:

> Using built-in model options. Could not load live models for this provider.

The runtime merge of both ACP model sources was implemented in `desktop/src-tauri/src/commands/agent_models.rs` (`normalize_agent_models`, merged after commit `925a9a7`), but **no regression test existed for the unstable `availableModels` path**, making a silent future regression likely.

## Changes

Two unit tests under `desktop/src-tauri/src/commands/agent_models_tests.rs`:

1. **`agent_models_populates_unstable_available_models_only_hermes_shape`**
   Reproduces the exact JSON shape Hermes Agent emits: `stable.configOptions = null`, `unstable.availableModels = [...]`, plus `unstable.currentModelId`. Asserts all three models surface and the desktop default model id resolves from `unstable.currentModelId`.

2. **`agent_models_stable_config_options_take_precedence_over_unstable_duplicates`**
   When a model id appears in *both* the stable `configOptions` and the unstable `availableModels` lists, the stable entry takes precedence for the UI display name — first-seen dedup wins, keeping the stable path authoritative.

## Why not a code change?

Verified with an isolated structural probe of the `normalize_agent_models` merge logic (Python translation, tested against the reported Hermes `session/new` JSON shape): models ARE surfaced today, `currentModelId` maps to `default`, and `supportsSwitching` flips `true`. The reporter on v0.5.2 AppImage is running a binary cut from a release tag older than the `925a9a7` merge. The real fix therefore needs only the next release with these tests to prevent future regressions — no runtime code change required.

## Test plan

```
cd desktop/src-tauri
cargo test --lib agent_models
# 38 passed (36 existing + 2 new), 0 failed
cargo clippy --lib --tests -- -D warnings
# no warnings
```

## Why default path is the right choice

The unstable `availableModels` path is an evolving ACP convention — the route for "the agent carries its own catalog". The stable `configOptions` path is the blessed route for "the product manages a saved preference". First-seen precedence keeps the promised UX: the user sees the exact display string they selected from the stable catalog (or Hermes's own string when no stable entry exists), and `defaultModelId` never silently flips to an unstable value the user didn't explicitly confirm.

### DIFF
--- a/crates/buzz-db/src/partition.rs
+++ b/crates/buzz-db/src/partition.rs
@@ -4,9 +4,46 @@
 
 use chrono::{Datelike, TimeZone, Utc};
 use sqlx::{PgPool, Row};
-use tracing::info;
+use tracing::{info, warn};
 
 use crate::error::{DbError, Result};
+
+/// True when a `sqlx::Error` is a Postgres `42P17` ("would overlap partition")
+/// raised by a `CREATE PARTITION` whose range is already covered. Fresh schemas
+/// include a right-edge catch-all (`*_p_future`), so a fresh install collides
+/// on the "current month" boundary it tries to add. Split out of
+/// `ensure_partition` so the classification is unit-testable without Postgres.
+fn is_partition_overlap_error(err: &sqlx::Error) -> bool {
+    match err {
+        sqlx::Error::Database(db) => {
+            db.code().as_deref() == Some("42P17")
+                && db.message().contains("would overlap partition")
+        }
+        _ => false,
+    }
+}
+
+/// Record #4033: a monthly partition attempt was absorbed by the `*_p_future`
+/// catch-all (monthly partition NOT created). Metric so operators can alert
+/// independent of log level.
+fn catch_all_coverage_metric(table: &'static str) {
+    metrics::counter!(
+        "buzz_db_partition_catchall_coverage",
+        "table" => table,
+    )
+    .increment(1);
+}
+
+/// Resolve a validated table name to its static allowlist entry so the metric
+/// label escapes the caller's stack. `ensure_partition` already validated the
+/// name against `PARTITIONED_TABLES`; `.expect` is unreachable for valid input.
+fn static_table_name(table: &str) -> &'static str {
+    PARTITIONED_TABLES
+        .iter()
+        .copied()
+        .find(|t| *t == table)
+        .expect("table name validated against PARTITIONED_TABLES")
+}
 
 /// Tables that may be partition-managed. Allowlist prevents DDL injection.
 const PARTITIONED_TABLES: &[&str] = &["events", "delivery_log"];
@@ -132,17 +169,20 @@ async fn ensure_partition(
             info!("added partition {partition_name}");
             Ok(())
         }
-        Err(sqlx::Error::Database(db_err))
-            if db_err.code().as_deref() == Some("42P17")
-                && db_err.message().contains("would overlap partition") =>
-        {
-            // Fresh schemas include a right-edge catch-all partition (`*_p_future`).
-            // If it already covers this month, the table is still safe for writes;
-            // treat the overlap as "ensured" rather than failing startup.
-            info!(
-                partition_name,
-                "partition range already covered by an existing partition"
+        Err(e) if is_partition_overlap_error(&e) => {
+            // #4033: the `*_p_future` catch-all already covers this month, so the
+            // monthly partition was NOT created and never will be — every row for
+            // this range keeps landing in the catch-all. Postgres only logs an
+            // unsuppressable server-side ERROR; the app must NOT surface this as a
+            // silent `info!` success. Escalate to `warn!` and emit a metric so the
+            // growing catch-all is alertable, but still return Ok: writes are safe
+            // and startup must not fail.
+            warn!(
+                partition = %partition_name,
+                table = %table_name,
+                "monthly partition was NOT created: range already covered by the `*_p_future` catch-all (Postgres 42P17); rows for this range keep accumulating in the catch-all. Re-base or split the catch-all to restore monthly pruning/archival."
             );
+            catch_all_coverage_metric(static_table_name(table_name));
             Ok(())
         }
         Err(e) => Err(e.into()),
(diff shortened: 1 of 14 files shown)


iroiro147:
Superseded by #5076 — rebased onto current main @ bd2fdf4a. Original branch was so far behind that rebase via cherry-pick produced a near-repo-wide diff; v2 carries the same single-file test change (+99) as a clean commit.

PR #4051, opened 2026-08-01
fix(mobile): add channel members via a community-wide search sheet

inventivepotter:
## Summary
- Mobile's Members sheet had no way to add someone to a channel. The only existing path was `@mention` in the compose bar, whose candidate list (`mentionCandidatesProvider`) only surfaces agents that already share a channel with the current user — a chicken-and-egg problem for brand-new channels. Desktop doesn't have this gap: it has a dedicated `ChannelMemberInviteCard` that searches the whole community.
- Adds an **Add member** action (icon button) to `MembersSheet`, opening a new `AddChannelMemberSheet` that searches the full relay directory via the existing `relayDirectorySearchProvider`/`relayDirectoryUsersProvider` (not scoped to current channel membership), lets you multi-select people/agents, and submits via the existing `ChannelActions.addMembers`.
- Mirrors `compose_bar.dart`'s existing human/agent role split: verified NIP-OA agents are added with role `bot`, everyone else with role `member`. To support that, `DirectoryUser` now carries `ownerPubkey`/`isAgent`, computed the same way `UserProfile.ownerPubkey` and the mention-candidates provider already do (`verifiedOaOwnerPubkey`).
- Hidden for DM and archived channels (`canAddMembers = !channel.isDm && !channel.isArchived`), consistent with the rest of the sheet's management affordances.

## Test plan
- [x] `dart format --output=none --set-exit-if-changed .` — clean
- [x] `flutter analyze` — no issues
- [x] `flutter test test/features/channels/` — 486 passed (0 regressions), including 2 new tests for `AddChannelMemberSheet` (excludes existing members from results; submits with the correct role) and 1 new unit test for `directoryUsersFromProfileEvents` agent detection
- [x] `just mobile-check` — fmt/analyze/file-size guard all pass

### DIFF
--- a/mobile/lib/features/channels/add_member_sheet.dart
+++ b/mobile/lib/features/channels/add_member_sheet.dart
@@ -1,0 +1,339 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_hooks/flutter_hooks.dart';
+import 'package:hooks_riverpod/hooks_riverpod.dart';
+import 'package:lucide_icons_flutter/lucide_icons.dart';
+
+import '../../shared/theme/theme.dart';
+import '../../shared/widgets/avatar_image.dart';
+import '../../shared/widgets/buzz_loading_indicator.dart';
+import 'channel_management_provider.dart';
+
+/// Debounce before a directory-search query hits the relay.
+const _searchDebounce = Duration(milliseconds: 250);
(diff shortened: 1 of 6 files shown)


wesbillman:
@inventivepotter mind adding screenshots etc. so we can easily review the change?

Thanks!

PR #4052, opened 2026-08-01
fix(desktop): keep Create controls on empty Projects (#4009)

### LINKED ISSUE #4009: Projects view hides Create controls when no projects are visible
## Problem

When the Projects view has no visible projects, it returns an empty state before the Projects chrome and Create controls mount. This means the Create menu is missing, and users cannot create the first project from the empty view.

## Expected behavior

The Projects view should keep its Create controls visible when there are zero projects. The empty state should also offer a clear way to open the create-project dialog.

## Proposed fix

- Keep the Projects layout and Create menu mounted for an empty list.
- Add a `Create repository` CTA to the empty state.
- Cover both entry points with an end-to-end test.

The implementation and test coverage are currently available in [PR #3620](https://github.com/block/buzz/pull/3620), which is being closed so this can be tracked as an issue.

## Test coverage from the PR

- Biome check on touched files
- TypeScript typecheck
- Playwright smoke test covering the empty Projects view and both Create entry points

Manual verification remains: open Projects with zero announcements and confirm that the `+` Create menu and empty-state `Create repository` button open the create dialog.

## Related bug: deleted and re-announced projects remain missing

If a created project is deleted and then re-declared/re-published to the Nostr relay, it does not appear in the Projects view again.

### Reproduction

1. Create a project and confirm it appears in Projects.
2. Delete the project.
3. Re-declare or re-publish the same project announcement to the Nostr relay.
4. Return to Projects or refresh the project list.
5. Observe that the project is still missing.

### Expected behavior

A valid project that is re-announced to the relay should be discoverable and appear in Projects again. Deleting a project should not leave a permanent hidden/tombstone state that prevents the same project from returning after it is re-published.

Please investigate whether project deletion leaves stale hidden-card, cache, or deduplication state keyed to the project identity.

iroiro147:
## Summary

Fixes #4009.

When the Projects view had no visible projects it returned `<EmptyState />` before the `PageHeader`, toolbar, and `ProjectsCreateMenu` mounted — so the Create menu was missing and there was no way to create the first project from the empty view.

## Approach

Reimplements the approach from the closed PR #3620 (which the issue body points to), adapted to current `main`:

- Remove the `projects.length === 0` early return in `ProjectsView.tsx`; the main content section now renders the empty state *inside* the page chrome, so the header, toolbar, and `+` Create menu stay mounted.
- `EmptyState` gains an optional `onCreateRepository` callback that renders a `Create repository` CTA button opening the create-project dialog.
- Both entry points (toolbar Create menu → Repository, empty-state button) open the same `CreateProjectDialog`.

## Test coverage

- `typecheck` (tsc --noEmit) — clean
- `biome check` on all touched files — clean
- New Playwright smoke spec `tests/e2e/projects-empty-create.spec.ts` covering the empty Projects view and both Create entry points — **passes** (`1 passed`)

## Notes

This is intentionally scoped to the empty-list Create-controls fix. The related "deleted + re-announced projects remain missing" behavior noted in the issue is a separate relay/query concern and is not changed here.

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -108,6 +108,7 @@ export default defineConfig({
         "**/project-inbox.spec.ts",
         "**/project-issue-comments.spec.ts",
         "**/project-pr-review.spec.ts",
+        "**/projects-empty-create.spec.ts",
         "**/persona-model-combobox-screenshots.spec.ts",
         "**/drafts-screenshots.spec.ts",
         "**/drafts-all-fix-screenshots.spec.ts",
--- a/desktop/src/features/projects/ui/ProjectCards.tsx
+++ b/desktop/src/features/projects/ui/ProjectCards.tsx
@@ -327,9 +327,16 @@ function RepositoryUnavailableIndicator({
   );
 }
 
-export function EmptyState() {
+export function EmptyState({
+  onCreateRepository,
+}: {
+  onCreateRepository?: () => void;
+} = {}) {
   return (
-    <div className="flex flex-1 flex-col items-center justify-center gap-3 px-4 py-16 text-center">
+    <div
+      className="flex flex-1 flex-col items-center justify-center gap-3 px-4 py-16 text-center"
+      data-testid="projects-empty-state"
+    >
       <Folders className="h-10 w-10 text-muted-foreground/40" />
       <div className="space-y-1">
         <p className="text-sm font-medium text-foreground">No projects yet</p>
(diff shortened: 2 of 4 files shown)


hamza-imran75:
+1 I'm hitting this exact issue on a fresh relay too. After enabling Projects, the empty Projects page only shows “No projects yet” and provides no + menu or other way to create the first repository from the Desktop UI.

I've seen usage of Projects on X and would like to use it myself as well, looking forward to seeing this PR land :)

iroiro147:
@hamza-imran75 thanks for confirming on a fresh relay — a second independent repro of the missing create path is useful. The branch is current against main and I'm keeping it green; hopefully it lands soon.

PR #4053, opened 2026-08-01
fix(desktop): enforce owner-only access in internal builds

brow:
## Problem

Managed agents in internal Buzz builds should use a narrower access boundary than agents in configurable open source builds. Previously, an agent could keep a broader access setting and respond to other people.

This PR improves internal-build access enforcement for local starts and provider deployments in marked builds. It also explains that **Only me** admits the owner and the owner's verified same-owner agents. Same-owner agent control is deliberate and supports Welcome teammates. Open source builds remain configurable.

Marked builds reconcile existing provider deployments during workspace apply. Desktop fails the workspace apply if a provider does not accept the owner-only redeploy, so it does not present the locked control before remote enforcement succeeds.

## Changes

- Clamp access for local starts and provider deployments in a marked internal build.
- Reconcile existing provider deployments during workspace apply. Fail the apply if an owner-only redeploy fails.
- Prevent user, persona, definition, and baked build environment values from overriding the two access gates.
- Keep the **Only me** label and state that only the owner and the owner's agents can send instructions.
- Make Welcome teammate access remediation converge without repeated restarts.
- Leave open source build behavior unchanged.

This changes effective runtime access without rewriting stored or relay-advertised settings.

The companion [#4064](https://github.com/block/buzz/pull/4064) explains the restriction in-thread when someone without access mentions an agent.

The enforcement will remain inactive in shipped builds until [squareup/buzz-releases#74](https://github.com/squareup/buzz-releases/pull/74) marks internal releases during the build.

## Screenshots

| Before | After |
| --- | --- |
| Editable agent access control before the change | Agent access locked to Only me in an internal build |

## Tests

Added coverage for:

- Runtime enforcement for [locally run agents](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src-tauri/src/managed_agents/runtime/tests.rs#L196) and [deployed agents](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src-tauri/src/commands/agents_tests.rs#L510).
- The [current-build deployment path](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src-tauri/src/commands/agents_tests.rs#L455), [invalid stored access](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src-tauri/src/managed_agents/access_policy.rs#L98), and the [local startup guard](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src-tauri/src/managed_agents/env_vars/tests.rs#L149).
- Consistent enforcement across [both agent backends](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src-tauri/src/managed_agents/access_policy.rs#L112).
- Welcome teammates created as [locally run](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src/features/onboarding/welcomeGuide.test.mjs#L384) or [deployed](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src/features/onboarding/welcomeGuide.test.mjs#L393) agents, including [access-only](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src/features/onboarding/welcomeKickoff.test.mjs#L202) and [runtime-related](https://github.com/block/buzz/blob/e0165f52b52741a74184c9899e2b51eeec40c939/desktop/src/features/onboarding/welcomeKickoff.test.mjs#L225) restart behavior.

The full Desktop Rust and JavaScript suites, type checks, formatting, clippy, and file-size checks passed. Playwright E2E was not run.

---

Originated from Buzz channel [buzz-agent-control](buzz://channel?id=cf5dada7-e26a-4887-ae41-b3bd5f42d3b2). Supersedes #2537.


wesbillman:
Commenting on Wes Billman's behalf after a release-safety review of `16b66e5c916a68e79a7ee6b8e2be21126591c56d`.

I found two blockers before this is safe to merge for an internal release:

1. **Existing provider deployments are not remediated.** The owner-only projection is applied when Desktop constructs a provider deploy payload (`desktop/src-tauri/src/commands/agents_deploy.rs`), and that payload is sent only through an explicit provider deploy/start path (`desktop/src-tauri/src/commands/agents.rs`). Installing a marked Desktop build does not redeploy an agent whose `backend_agent_id` already exists; provider status deliberately survives Desktop restarts. Such an agent can therefore retain its previous `anyone`/allowlist policy while the upgraded UI reports and locks the setting to “Only me.”

   Please either add a reliable reconciliation/migration for existing provider deployments, or make enforcement state explicit and gate release on inventory plus successful redeployment. The UI must not claim the effective policy changed until the remote deployment has accepted it. Add coverage for the upgrade case: an already-deployed provider agent with wider stored/effective access receives the owner-only policy during rollout.

2. **The enforced principal set is broader than the stated contract.** `RespondTo::OwnerOnly` in `crates/buzz-acp/src/lib.rs` admits the human owner **and NIP-OA-verified same-owner sibling agents**. This PR intentionally relies on that behavior for Welcome teammates. That may be the desired authorization model, but it conflicts with “answer only their owner” and UI copy that says only the user can send instructions.

   Please obtain an explicit product/security decision that `owner ∪ verified same-owner agents` is the intended internal boundary and update the PR/UI language accordingly, or change the authorization behavior to satisfy literal human-owner-only access. This is an authorization-set decision, not merely copy polish.

The local spawn clamp, independent `BUZZ_ACP_ALLOWED_RESPOND_TO=owner-only` guard, provider payload projection, and compile-time marker wiring otherwise look sound in the reviewed diff. Before merge, this branch should also be updated from `main` and revalidated after the blockers are resolved.

brow:
@wesbillman those make sense but don't seem like they should be blocking. Do you agree?

wesbillman (dismissed):

wesbillman (changes requested):
Commenting on Wes Billman's behalf after reviewing current head `99685b52b7daa7e385eca3db84c8fb916c709de1`.

Two authorization/release blockers remain:

1. **Existing provider deployments are not brought under the new policy.** The owner-only projection is applied only while constructing a provider deploy payload (`desktop/src-tauri/src/commands/agents_deploy.rs:144-176`), and that payload reaches the provider only through create-with-deploy or an explicit Start/deploy operation (`desktop/src-tauri/src/commands/agents.rs:1001-1018`, `1125-1156`). Existing remote infrastructure deliberately remains represented as deployed across Desktop restarts whenever `backend_agent_id` exists (`desktop/src-tauri/src/managed_agents/runtime.rs:147-168`). Upgrading to a marked build therefore leaves a previously deployed `anyone`/allowlist agent remotely open while the UI locks and displays “Only me.” That contradicts the PR's claim that owner-only access becomes effective for every managed agent in internal builds.

   Please reconcile already-deployed provider agents during rollout, or expose and enforce a fail-closed state until the provider has accepted an owner-only redeploy. Add an upgrade-path test covering an existing provider deployment with wider effective access.

2. **`owner-only` is not literally owner-only.** The harness admits the human owner *and* every cryptographically verified same-owner sibling agent (`crates/buzz-acp/src/lib.rs:220-256`). This branch explicitly relies on that broader set for Welcome teammates (`desktop/src/features/onboarding/welcomeGuide.ts:264-275`), while the UI says “Only me” and “Only you can send instructions” (`desktop/src/features/agents/ui/RespondToField.tsx:69-76,237-240`). Please either obtain and document the product/security decision that the intended boundary is `owner ∪ verified same-owner agents` and make the user-facing contract accurate, or enforce literal human-owner-only access.

The local spawn clamp, independent `BUZZ_ACP_ALLOWED_RESPOND_TO=owner-only` guard, provider payload projection for new/explicit deployments, build-marker wiring, and OSS behavior otherwise look sound. CI is green at this head; I relied on the existing full Desktop checks rather than rerunning equivalent suites locally.

wesbillman:
Additional blocker found in the independent security pass, commenting on Wes Billman's behalf:

3. **The baked build environment can overwrite both enforcement gates after Desktop sets them.** `BUZZ_BUILD_AGENT_ENV` accepts arbitrary keys at build time (`desktop/src-tauri/build.rs:42-71`), and `baked_build_env()` returns those entries without reserved-key filtering (`desktop/src-tauri/src/managed_agents/agent_env.rs:22-61`). They enter the effective descriptor as its baked floor (`desktop/src-tauri/src/managed_agents/readiness.rs:226-270`). Local spawn sets `BUZZ_ACP_RESPOND_TO=owner-only` and `BUZZ_ACP_ALLOWED_RESPOND_TO=owner-only` first (`desktop/src-tauri/src/managed_agents/runtime.rs:754-764`), then writes every `descriptor.env` entry last (`runtime.rs:803-813`). An internal build packaged with both keys set to `anyone` therefore produces an agent that answers anyone.

The expanded `RESERVED_ENV_KEYS` protects user/persona/definition input, but not this baked tier. Reject reserved keys in `BUZZ_BUILD_AGENT_ENV` at build time or filter them before they reach `descriptor.env`, and add a test proving baked collisions cannot override either access gate. This is a direct bypass of the boundary the PR intends to make invariant.

wesbillman:
One additional upgrade-path bug from Mongo's independent lifecycle review, confirmed at `99685b52b`:

4. **Welcome teammate remediation never converges in owner-only builds.** The new expected-state predicate requires `owner-only` with an empty allowlist in a marked build, but the remediation branch still unconditionally writes `allowlist:[lead]` (`desktop/src/features/onboarding/welcomeGuide.ts:349-359`). A pre-existing install whose teammates carry the old allowlist state therefore fails the predicate, writes the same invalid-for-this-build state again, and repeats on later provisioning passes. Running teammates also continue satisfying `welcomeTeammateNeedsRestart` (`welcomeKickoff.ts:360-373`), causing redundant restarts.

This is not another authorization bypass—the Rust boundary still clamps effective access—but it is a real upgrade regression and contradicts the PR's goal of avoiding unnecessary Welcome restarts. Make the remediation target `owner-only`/empty when the build policy is active, and test the full remediation write for a pre-existing allowlisted teammate rather than only the predicate's happy path.

brow:
🤖 Head is now `579e9b10d389ba2202daac074b587fc225a5cc84`. Items 3 and 4 are fixed. Item 2 is now fixed in the copy. Item 1 is documented rather than changed.

**Item 4, Welcome teammate remediation: fixed** in [dfad5344](https://github.com/block/buzz/pull/4053/changes/dfad534440e87578e59a7c44f24cf81f31883c43). You were right that the predicate and the write had drifted. Both now come from one place, `welcomeTeammateAccessUpdate`, which returns the write that satisfies `welcomeTeammateHasExpectedAccess` for the current build or null when the teammate is already correct. In a marked build that is `owner-only` with an empty allowlist, so an upgraded install converges on the first pass instead of rewriting rejected state and restarting running teammates forever. Tests cover the full remediation write for a pre-existing allowlisted teammate in an owner-only build, the unmarked-build case, and the no-op case, in `welcomeGuide.test.mjs`.

**Item 3, baked build env: fixed** in [34f8ac82](https://github.com/block/buzz/pull/4053/changes/34f8ac823661d55a7003318e184e264afffdeb73). Rejected where the value enters the system: `build.rs` now panics if a `BUZZ_BUILD_AGENT_ENV` line carries a reserved key, so such a build fails instead of shipping. `baked_build_env()` also filters reserved keys, so a binary produced without that check cannot override the gates either. The key list lives in one file, `reserved_env_keys.rs`, `include!`d by both `build.rs` and `env_vars.rs`, following the `reconnect_hook_config.rs` precedent, because a build script cannot import from the crate and two copies would drift. Three tests cover the two access-gate keys, case-insensitive matching, and every entry of `RESERVED_ENV_KEYS`. I confirmed the build-time rejection fires: a probe build with `BUZZ_ACP_RESPOND_TO` baked fails with the new panic.

**Item 1, existing provider deployments: documented, not reconciled.** Your reading of the code is right. The projection reaches a provider only when Desktop serializes a new payload, and a record with a `backend_agent_id` stays represented as deployed across restarts, so an agent deployed from an unmarked build keeps its wider remote access until it is next deployed from a marked build. That limit is now stated in the module docs on `access_policy.rs` ([bf9f0c40](https://github.com/block/buzz/pull/4053/changes/bf9f0c40da59573c7670f0baf70fe99dd074a1a8)), together with what the capability does enforce, and it says a rollout reconciliation pass (redeploy, or fail closed on mismatch) is what would close it. The direction from the PR owner is to ship the two enforced boundaries now and track rollout reconciliation as separate work rather than widen this PR.

**Item 2, "owner-only" is not literally owner-only: the copy is now accurate.** You were right that the contract was wrong. The broader set is the intended boundary, not an oversight, and the Welcome team depends on it: the lead instructs its teammates while every teammate is created owner-only. So rather than narrow the enforcement, we corrected the promise. The line under Only me now reads "Only you and your agents can send instructions." ([579e9b10](https://github.com/block/buzz/pull/4053/changes/579e9b10d389ba2202daac074b587fc225a5cc84)). The dropdown label stays "Only me", which is the audience the user picks and what that option has meant since before agents could instruct each other. A contract test in `respondToFieldContract.test.mjs` pins the new line, and I confirmed it fails when the copy is reverted. The `access_policy.rs` docs now cite that copy instead of calling it a question tracked elsewhere.

Validation at `579e9b10d`: `pnpm check`, `pnpm typecheck`, `pnpm test` (4294 pass, 0 fail), `cargo fmt --check`, `cargo check --all-targets`. The pre-push hook also ran desktop-check, desktop-test, rust-tests, and desktop-tauri-checks, all green. CI is running at the new head.

wesbillman (changes requested):
Re-reviewed exact head `579e9b10d389ba2202daac074b587fc225a5cc84` on Wes's behalf.

Items 2–4 from the prior review are resolved: the UI now accurately names same-owner agents, baked reserved keys are rejected at build time and filtered at runtime from one canonical list, and Welcome teammate remediation now converges to the build-specific expected state.

The provider upgrade-path blocker remains unresolved. Documentation accurately admits the limitation, but it does not alter the unsafe runtime state: upgrading a marked internal build can still leave an existing provider deployment remotely accepting `anyone` or an old allowlist while Desktop locks the control and presents **Only me**. No reconciliation or fail-closed mismatch state was added, and no concrete follow-up issue is linked. That leaves the access guarantee false for an important upgrade path.

Please reconcile existing provider deployments, fail closed until an owner-only redeploy succeeds, or split the provider claim/UI lock from this rollout so the shipped behavior cannot present enforcement it has not applied. The branch also currently conflicts with `main`, which will need resolution before merge.

Existing CI is green at this head; I relied on it rather than duplicating the broad Desktop suites locally.

wesbillman (approved):
Approving exact head `59003bb3bfac99071861519ffbf0fe0448f3e961` on Wes Billman's behalf.

The prior provider upgrade-path blocker is resolved. In owner-only builds, workspace apply now collects existing provider-backed deployments, rebuilds each deployment payload through the current owner-only projection, and awaits redeployment before community readiness. Payload-build or provider-deploy failures propagate from `reconcile_on_workspace_apply` through `apply_workspace`, and the frontend remains on its not-ready gate rather than presenting the locked policy as applied.

I also rechecked local/provider access projection, reserved and baked environment-key defenses, marked-versus-OSS behavior, Welcome remediation convergence, and the merge resolution around effective parallelism. Mongo independently cleared the provider lifecycle and UI-readiness path; Princess Donut independently cleared policy bypasses and regression coverage. No actionable findings remain.

Validation on this exact clean head:
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml provider_access --lib`: 2 passed
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml owner_only_access_deploy_payload_clamps_stale_access --lib`: 1 passed
- `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml --all -- --check`: passed
- `git diff --check origin/main...HEAD`: passed
- GitHub CI and required checks are green.

Issue #4004, opened 2026-07-31
[Bug] Agent model composer-2.5[fast=false] set in Desktop but ACP session bills/runs as fast=true


NocturnusCoder:
## Summary

Buzz Desktop lets me set Fizz to `composer-2.5[fast=false]`. Harness env shows `BUZZ_ACP_MODEL=composer-2.5[fast=false]`, but cursor-agent ACP actually runs `composer-2.5[fast=true]`. Cursor usage dashboard shows all agent turns as Composer 2.5 Fast.

## Environment

- Buzz Desktop (managed agents: Fizz, Bumble, Honey)
- buzz-acp harness → cursor-agent acp
- cursor-agent: 2026.07.23-e383d2b
- OS: Linux

## Steps to reproduce

1. In Buzz Desktop, set agent Fizz model to Composer 2.5 with fast=false
2. @mention Fizz in a channel
3. On the machine running buzz-acp, run:
   ```bash
   buzz-acp models | grep -E "Current:|composer-2.5"
   ```
4. Check Cursor usage dashboard at cursor.com

## Expected

- ACP Current model: `composer-2.5[fast=false]`
- Usage billed as full Composer 2.5

## Actual

- Harness env: `BUZZ_ACP_MODEL=composer-2.5[fast=false]` ✓
- ACP Current: `composer-2.5[fast=true]`
- Available catalog only lists `composer-2.5 (id: composer-2.5[fast=true])` — no fast=false variant
- Cursor usage: all recent entries show Fast mode

## Screenshot

![Cursor usage dashboard — all recent entries show Composer 2.5 Fast](https://gist.githubusercontent.com/NocturnusCoder/05290e272d734628939705d90887f83a/raw/cursor-usage-screenshot.png)

Gist: https://gist.github.com/NocturnusCoder/05290e272d734628939705d90887f83a

## Related issues

- #2692 — same silent-fallback pattern (Desktop model persists, session runs wrong model)
- #2794 / #2795 — ACP catalog id vs configId mismatch (Claude; fixed but not Composer fast=false)
- #3934 — ACP model list / discovery mismatches between Desktop and harness
- #2261 — cold-start model discovery race → silent fallback

## Notes

Possible fix: map Desktop selection to a model ID ACP actually supports, or pass `--model` at cursor-agent startup (see Cursor forum threads on ACP model switching).

NocturnusCoder:
[...]

PR #4054, opened 2026-08-01
test(buzz-acp): pin exact-match boundary for unadvertised model variants (#4004)

iroiro147:
## Summary

Refs #4004.

The issue reports Buzz Desktop letting a user set `composer-2.5[fast=false]` while the cursor-agent ACP catalog only advertises `composer-2.5[fast=true]` — so the session silently runs (and bills) the wrong variant.

The runtime boundary that decides this lives in `resolve_model_switch_method` / `model_in_catalog` (`crates/buzz-acp/src/acp.rs`): both match the desired model id against the fresh `session/new` catalog by **exact string equality**. The miss is intentional and load-bearing — the pool's fallback arm (`crates/buzz-acp/src/pool.rs:955-977`) consumes it to emit a `control_result { status: "unsupported_model" }` frame so the desktop rejects the pick instead of silently no-op'ing.

## Changes

Two regression tests under `acp.rs::tests` pin that boundary against a cursor-agent-shaped catalog (one model, `[fast=true]` only):

- `resolve_model_switch_method_misses_unadvertised_bracket_variant`
- `model_in_catalog_misses_unadvertised_bracket_variant`

Both assert the bracket-suffixed variant `composer-2.5[fast=false]` does **not** match when the harness never advertised it — the exact behavior the silent-billing report depends on for detection.

## Why test-only

The detection + surfacing chain already exists in current `main` (exact-match miss → `unsupported_model` frame → desktop live-pick rejection via `liveSwitchOutcome`). The reporter's observed silent fallback is consistent with running a harness/desktop build older than that chain. This PR pins the contract so the boundary can't regress silently; it deliberately does **not** add fuzzy/bracket-insensitive matching, which would trade a loud, fixable UX miss for silent execution of a model the user did not select — the failure mode #4004 is itself reporting.

## Test plan

```
cargo test --lib -p buzz-acp           # 663 passed (661 existing + 2 new), 0 failed
cargo clippy -p buzz-acp --all-targets # clean
```

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -2822,6 +2822,50 @@ mod tests {
         assert!(!super::model_in_catalog(&[], None, "anything"));
     }
 
+    // ── Issue #4004: bracket-modifier miss pins the exact-match boundary ─────
+    //
+    // The cursor-agent ACP catalog advertises `composer-2.5[fast=true]` as its
+    // only Composer 2.5 model id. When the desktop persisted
+    // `composer-2.5[fast=false]` (a variant the harness never reported), the
+    // exact-match below must miss so the pool's fallback arm
+    // (`pool.rs:955-977`) fires and the desktop learns the pick is
+    // unsupported, instead of silently applying the agent default and billing
+    // the user against `fast=true`.
+
+    #[test]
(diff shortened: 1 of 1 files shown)


iroiro147:

Issue #3980, opened 2026-07-31
buzz-acp: turn ending with unpublished agent text produces zero signal (no relay message, no warning)


davidglogan:
## Summary

When a `buzz-acp`-driven agent ends a turn with non-empty `agent_message_chunk` text but never makes a successful `buzz messages send` call, the turn ends with **zero signal** — no relay message, no warning, no failure notice. From the outside (Desktop's activity/observer view) it looks like the agent replied; on the relay, nothing was ever published.

This is not a claim that plain-text output is *supposed* to auto-publish — per `crates/buzz-acp/src/base_prompt.md` and `crates/buzz-acp/README.md`, the documented design is CLI-out (agent must explicitly call `buzz messages send`), and `agent_message_chunk` is telemetry only (`crates/buzz-acp/src/acp.rs`, handler logs via `tracing::info!` and returns `false` — never builds a kind:9). That design is fine. The gap is the **silence** when an agent doesn't follow it.

## Repro

1. Run any `buzz-acp`-wrapped agent with a system/persona prompt that (incorrectly, in our case, due to bad advice we gave it) tells it plain text is sufficient and it should never shell out to the `buzz` CLI.
2. `@mention` the agent with a question.
3. Observe: Desktop's activity/observer view shows a full, coherent plain-text reply as the agent's final turn output.
4. Query the relay directly: `buzz messages search --author <agent-pubkey> --limit 5` — no corresponding event exists at all. The turn produced real content and ended with no error, no failure notice, nothing.

We reproduced this against three separate live agent instances (all `buzz-agent`-based, via `buzz-acp` + `buzz-dev-mcp`) on 2026-07-31; confirmed the missing publish both via `buzz messages search` and by comparing `created_at` timestamps against the triggering mention.

## Why this matters

The only existing auto-publish path is for failure notices (`post_failure_notice` in `crates/buzz-acp/src/pool.rs`) — i.e., the harness already has the concept of "tell the relay this turn didn't produce a real reply." That same mechanism (or a lighter-weight warning) just isn't wired up for the "turn produced text, but no publish call happened" case.

Practical impact: a misconfigured or momentarily-confused agent (e.g. bad persona/memory instructions, or an LLM that got confused about tool-call syntax) fails completely silently from an operator's perspective. There's no error to notice, no channel signal, nothing — you only find out by cross-checking the relay directly against what Desktop's UI shows, which most operators won't think to do.

## Suggested fix (not prescriptive — flagging the gap, not the implementation)

[...]

PR #4057, opened 2026-08-01
fix(acp): surface silent turn-end when agent replies but never publishes (#3980)

iroiro147:
## Summary

Refs #3980 — surfaces the silent-failure mode when a `buzz-acp` agent ends a turn with non-empty `agent_message_chunk` text but never makes a successful `buzz messages send` call.

Today that turn ends with **zero signal**: no relay message, no warning, no failure notice. Desktop's activity/observer view shows a coherent reply while the relay has nothing, so an operator can't tell the agent failed to publish.

## What this does

- **Accumulates** streamed `agent_message_chunk` text per turn in `AcpClient` (`unpublished_turn_text`), alongside a `completed_tool_call_this_turn` flag set by any `tool_call_update` with `status == "completed"`.
- **`take_unpublished_turn_notice()`** (public, per-turn) returns `Some(text)` only when the turn produced non-empty text *and* completed zero tool calls — the case where the agent never shelled out to `buzz messages send`, so its reply is invisible on the relay. Resets both fields so each turn starts clean.
- **At turn end**, in the shared `send_prompt_result` chokepoint (so every outcome inherits it), the harness emits a `pool::prompt` warning with the turn id, byte count, and a content preview. That makes the silent-failure mode visible in operator logs and Desktop's observer view.

## Scope (deliberately bounded)

This adds turn-level accumulation + detection + a warning. A **relay-visible failure notice** (via `post_failure_notice`) is the natural follow-on but is **not** in this PR: that path needs `channel_id` + thread tags at the publish site, which the shared result chokepoint doesn't carry, and wiring it without that context risks leaking or wrong-channel posts. The accumulation state introduced here is exactly what that follow-on consumes.

Publishing itself is unchanged; agents that complete any tool call (the publish path) never trigger the warning. No fuzzy matching introduced.

## Verification

- `cargo check -p buzz-acp` — clean
- `cargo test -p buzz-acp` — 661 + 9 passing (matches `main` baseline, no regression)
- `cargo clippy -p buzz-acp` — clean

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -211,6 +211,14 @@ pub struct AcpClient {
     /// deltas. Both goose and buzz-agent emit this notification; goose gates
     /// on client capability advertisement, buzz-agent emits unconditionally.
     goose_usage: UsageTracker,
+    /// Text the agent streamed as `agent_message_chunk` during the current
+    /// turn that was NOT followed by a successful publish tool call. Used to
+    /// surface a "turn ended without publishing" signal (#3980).
+    unpublished_turn_text: String,
+    /// Whether the agent completed at least one tool call this turn. An agent
+    /// that produced text but completed zero tool calls never shelled out to
+    /// `buzz messages send`, so its reply is invisible on the relay.
+    completed_tool_call_this_turn: bool,
 }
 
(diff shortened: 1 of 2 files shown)


iroiro147:

PR #4058, opened 2026-08-01
fix(desktop): offer shared relay agents in @ mention autocomplete (#3971)

### LINKED ISSUE #3971: Mention autocomplete never offers remotely-hosted agents (shared relay agent eligibility is unreachable)
**Version:** 0.5.2 (also present on current main)

**Setup:** An ACP agent hosted on machine A (via buzz-acp with NIP-OA owner attestation), and Buzz Desktop on machine B owned by the same user. The agent is a member of a shared channel, publishes a kind-0 profile, appears in the member list with the agent chip ("managed by you"), is registered in the relay agent directory, and advertises `respond_to=anyone`.

**Bug:** The `@` mention autocomplete on machine B never offers the agent, so there is no way to send it a p-tag mention from the UI. The members panel shows it; the composer refuses to.

**Root cause:** In `desktop/src/features/messages/lib/useMentions.ts`, `addCandidate` drops any candidate flagged `isAgent` that is not in the *local* managed-agent list:

      return;
    }

This runs *before* `shouldHideAgentFromMentions`, whose `mentionableAgentPubkeys` set (built from `relayAgentIsSharedWithUser` — directory agents with `respond_to=anyone`/allowlist sharing a channel) is meant to make exactly these agents mentionable. Because the earlier filter only consults the local list, that machinery is unreachable dead code for any agent hosted on another machine.

**Suggested fix:** let mentionable relay agents through the first filter:

    if (
      !isAgentIdentityInManagedList(candidate, managedAgentPubkeys) &&
      !mentionableAgentPubkeys.has(pubkey)
    ) {
      return;
    }

**Impact:** Any multi-device / team setup where an agent runs on one machine and users chat from another. Current workarounds are thread replies (reply p-tags the parent author) or agent-side content-matching rules instead of real mentions.

iroiro147:
## Summary

Fixes #3971 — the `@` mention autocomplete never offered remotely-hosted agents, even when they were channel members, appeared in the members panel, and were registered in the relay agent directory with `respond_to=anyone`.

## Root cause

In `desktop/src/features/messages/lib/useMentions.ts`, `addCandidate` dropped any `isAgent` candidate not in the *local* managed-agent list:

```ts
  return;
}
```

This ran **before** `shouldHideAgentFromMentions`, whose `mentionableAgentPubkeys` set (built from `relayAgentIsSharedWithUser` — directory agents that share a channel with the user and advertise `respond_to=anyone` or an allowlist match) is precisely the machinery meant to make shared relay agents mentionable. The early local-only filter made that set unreachable dead code for any agent hosted on another machine.

## Fix

Admit the candidate when it is either locally managed **or** in the relay-mentionable set:

```ts
if (
  !isAgentIdentityInManagedList(candidate, managedAgentPubkeys) &&
  !mentionableAgentPubkeys.has(pubkey)
) {
  return;
}
```

`shouldHideAgentFromMentions` still runs downstream and independently consults `mentionableAgentPubkeys`, so noise-gating is preserved: a remote agent that is **not** relay-shared with the user is still dropped.

## Verification

- `pnpm typecheck` — clean
- `pnpm exec biome check` — clean (both touched files)
- `node --import ./test-loader.mjs --experimental-strip-types --test src/features/agents/lib/agentAutocompleteEligibility.test.mjs` — **20/20 pass** (18 baseline + 2 new pinning tests)
- Mention suites (`mentionCandidates.test.mjs`, `mentionRanking.test.mjs`) — green

## Tests added

`agentAutocompleteEligibility.test.mjs`, pinning the new admit condition:
- a shared relay agent **not** in the local managed list **is** admitted;
- a non-shared non-local agent **is** still rejected.

## Impact

Multi-device / team setups where an agent runs on one machine and users chat from another now get real `@Name` autocomplete → `p`-tag delivery. Workarounds (thread replies to p-tag the parent author, agent-side content-matching rules) are no longer needed.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -306,3 +306,39 @@ test("coalesceAgentAutocompleteCandidates: leaves non-agents alone", () => {
 
   assert.deepEqual(coalesce([first, second]), [first, second]);
 });
+
+
+test("shared relay agent not in local managed list is admitted (#3971)", () => {
+  // #3971: an agent hosted on another machine is a channel member and appears
+  // in the relay directory with respond_to=anyone, but is NOT in the local
+  // managed-agent list. The old addCandidate gate dropped it before
+  // shouldHideAgentFromMentions could consult mentionableAgentPubkeys, making
+  // remote shared agents unreachable. The fixed admit condition is:
+  //   admit iff isAgentIdentityInManagedList(...) || mentionableAgentPubkeys.has(pubkey)
+  const managedAgentPubkeys = new Set(); // agent is remote-only, not local
+  const mentionableAgentPubkeys = new Set([PUB_B]); // relay-shared
+  const pubkey = PUB_B;
+
+  const admitCondition =
+    isAgentIdentityInManagedList(
+      { isAgent: true, pubkey },
+      managedAgentPubkeys,
+    ) || mentionableAgentPubkeys.has(pubkey);
+
+  assert.equal(admitCondition, true);
+});
+
+test("non-shared non-local agent is still rejected (#3971)", () => {
+  // A remote agent NOT in the relay mentionable set must still be dropped.
+  const managedAgentPubkeys = new Set();
+  const mentionableAgentPubkeys = new Set();
+  const pubkey = PUB_C;
+
+  const admitCondition =
+    isAgentIdentityInManagedList(
+      { isAgent: true, pubkey },
+      managedAgentPubkeys,
+    ) || mentionableAgentPubkeys.has(pubkey);
+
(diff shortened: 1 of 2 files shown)


iroiro147:
Closing as superseded by #4913, which landed this with broader scope.

For anyone tracing it later: the root cause identified here — `addCandidate` filtering `isAgent` candidates against the local managed-agent list before `shouldHideAgentFromMentions` could run, leaving the shared relay-agent eligibility path unreachable — is resolved on `main`. That gate is now `isAgentIdentityInAllowedList(candidate, mentionableAgentPubkeys)`, and `getMentionableAgentPubkeys` admits relay agents in channel scope via `relayAgentCanRespondInChannel`.

Thanks @wesbillman. One note: #3971 still appears open and looks resolved by #4913.

PR #4063, opened 2026-08-01
feat(cli): add non-mutating buzz doctor command (#3926)

### LINKED ISSUE #3926: Feature request: add a non-mutating `buzz doctor` command
**Motivation**

The CLI has comprehensive command help, but `--help` cannot answer whether a Buzz environment is actually ready to use. Today, troubleshooting usually means running a real command and interpreting the first error it returns.

A single safe preflight would help humans, agents, automation, and support reports answer:

- Is the local configuration present and parseable?
- Which public identity will sign requests?
- Is the relay reachable and does it return valid NIP-11 metadata?
- Is authentication accepted?
- Is the identity a member of the selected community?

Buzz Desktop has a Doctor surface for managed ACP runtimes, but that has a different scope and is unavailable in headless environments, containers, and servers.

**Proposed solution**

Add a root-level `buzz doctor` command that runs independent checks, aggregates the results rather than failing at the first problem, and never publishes events or changes local or remote state.

A useful first slice could include:

Local checks, available even with `--offline`:

- CLI version/build information
- relay URL presence, syntax, and canonicalization
- `BUZZ_PRIVATE_KEY` presence and parseability, reporting only the derived public identity
- `BUZZ_AUTH_TAG` syntax and basic consistency, without ever printing the tag

Read-only remote checks:

- DNS/TLS/network reachability
- NIP-11 retrieval and basic compatibility
- authenticated read access
- community membership and available role/scope information
- WebSocket connection or subscription readiness, without publishing an event

The command should use the existing global relay, key, auth-tag, and output-format options. Default structured output could follow the existing CLI convention, for example:

```json
{
  "status": "warning",
  "checks": [
    {
      "id": "identity",
      "status": "ok",
      "message": "Signing identity is available"
    },
    {
      "id": "relay",
      "status": "error",
      "message": "Relay is unreachable",
      "remediation": "Check the relay URL and network connectivity"
    }
  ]
}
```

Required behavior:

- Running with missing or invalid credentials reports those as check results instead of failing during CLI argument parsing.
- `--offline` runs only local checks and clearly marks remote checks as skipped.
- Remote probes are read-only.
- All applicable checks run so one invocation can reveal multiple problems.
- Output and error messages never expose private keys, auth tags, bearer values, or signed request material.
- Check identifiers and structured fields are stable enough for automation.
- Exit behavior distinguishes a failed required check from warnings while preserving structured output. I would welcome maintainer guidance on whether warnings should still exit zero and how this should map onto the existing exit-code categories.

Suggested coverage:

- unit tests for aggregation, redaction, skipped checks, and exit precedence
- mocked relay tests for unreachable, malformed NIP-11, unauthenticated, and non-member cases
- a CLI-definition test proving `doctor --help` remains available without credentials

For the initial version, local configuration plus NIP-11, read-only authentication, and membership checks may be enough; WebSocket readiness could follow separately if it adds too much scope.

**Alternatives considered**

- Run a harmless existing read command such as `channels list`: this fails fast, does not cover local configuration, and can blur network, authentication, and membership failures.
- Use the Desktop Doctor: it diagnoses managed agent runtimes rather than CLI-to-community readiness and is not available headlessly.
- Use shell scripts and `curl`: these duplicate Buzz signing/authentication behavior and make secret handling less reliable.
- Improve individual error messages only: useful, but it does not provide proactive, aggregated readiness information.

**Additional context**

- The prior Doctor issue, #76, was specifically for the Desktop ACP setup surface and was completed by #80.
- PR #2933 proposes `buzz users me` with a `whoami` alias. A doctor command should reuse the same local identity derivation if that lands, not introduce a competing identity command.
- I searched open and closed issues and PRs for `doctor`, `diagnostic`, `health check`, `preflight`, and `troubleshoot`; I found no existing CLI diagnostics proposal.
- A local audit of current `main` found that `--help` succeeds across all exposed CLI command paths, so this proposal is about runtime readiness rather than help coverage.

iroiro147:
feat(cli): add non-mutating `buzz doctor` command (#3926)

Closes #3926.

## What

Adds a root-level `buzz doctor` subcommand that runs independent local + remote readiness checks and aggregates them into a single structured JSON report, so users / agents / automation / support flows can answer "is my Buzz environment ready to use?" without `--help` trivia and without failing at the first problem.

## Why

- Today troubleshooting means running a real command and interpreting the first error — a miss between local config, identity, reachability, auth, and membership all blur.
- `Buzz Desktop` has a Doctor surface for managed ACP runtimes, but that's a different scope and unavailable headlessly.
- A dedicated preflight fills the gap; this is the smallest useful slice per the issue's "first version" guidance.

### Command

```
buzz doctor [--offline]
```

The subcommand is added directly to the top-level `Cmd` enum (not nested under a group), matching the existing architecture.

### Dispatch order

`run()` in `crates/buzz-cli/src/lib.rs` is reshaped so doctor dispatches **before** the credential gate. Every other command still requires `BUZZ_PRIVATE_KEY`; doctor must remain usable when the broken environment is the one being debugged, so missing/malformed credentials surface as a check result (status `error`), not as an argument-time failure.

### Checks (all independent, all run; one invocation reveals multiple problems)

Local, always-on:

- `relay_url` — presence, syntax, canonicalization via the same `normalize_relay_url` the rest of the CLI uses.
- `identity` — `BUZZ_PRIVATE_KEY` parseability. **Key material is never printed;** only the derived public identity (hex) is reported on success.
- `auth_tag` — `BUZZ_AUTH_TAG` parse + verification against the signing identity via `buzz_sdk::nip_oa::parse_auth_tag` / `verify_auth_tag`. **The tag is never printed.** When set but the identity is invalid, we degrade to `warning`.
- `version` — `env!("CARGO_PKG_VERSION")`.

Remote (skipped when `--offline`):

- `relay_reachable` — an unauthenticated `GET /` with `Accept: application/nostr+json`, just to prove DNS/TLS/reachability.
- `nip11` — fetches and parses the NIP-11 information document end-to-end. Done with a bare `reqwest::Client` so doctor works even when the identity is broken.
- `auth_read` — bounded authenticated read (single `query_paginated` call with `limit=1`) against `kinds:[39002] #p=self`. This doubles as the membership probe input.
- `membership` — interprets the previous call: zero events → `warning` ("not a member of any channel visible to this relay"), ≥1 events → `ok`.

`--offline` marks the four remote probes `skipped` and they do not affect the outcome.

### Output

Single JSON document on stdout:

```json
{
  "status": "ok" | "warning" | "error",
  "checks": [
    {
      "id": "relay_url" | "identity" | "auth_tag" | "version" | "relay_reachable" | "nip11" | "auth_read" | "membership",
      "status": "ok" | "warning" | "error" | "skipped",
      "message": "…",
      "remediation": "…" | null
    }
  ]
}
```

Field identifiers are stable for automation.

### Exit behavior

- `0` when every applicable check is `ok` (warnings allowed; skipped never affects).
- `3` when any `identity` / `auth_tag` / `auth_read` check is `error` (auth-shaped).
- `2` when the only errors are `relay_reachable` / `nip11` (transport-shaped), via `CliError::Relay { status: 0, .. }`.

This matches the existing `0=ok 1=bad input 2=relay 3=auth 4=other 5=conflict` convention.

### Never exposes

- Private key bytes, hex, bech32 — only the derived public identity.
- Auth tag JSON, owner pubkey embedded inside the tag, or any signed request material.

## Testing

- 8 new unit tests in `commands/doctor.rs` covering:
  - all-`ok` → `Ok(())`, indico exit `0`
  - warning-only → `Ok(())`, exit `0`
  - `identity` error → `CliError::Auth` → exit `3`
  - `auth_read` error → `CliError::Auth` → exit `3`
  - `relay_reachable` error → `CliError::Relay{0,..}` → exit `2`
  - mixed errors → auth wins
  - `skipped` never affects outcome
  - JSON shape has `status` + `checks` array with expected keys
- Updated `command_inventory_is_stable` to include `doctor` (22 groups).
- Full `cargo test -p buzz-cli --lib`: **279 passed, 0 failed.**
- `cargo clippy -p buzz-cli --all-targets -- -D warnings`: clean.
- Smoke-tested: `buzz doctor --help` and `buzz doctor --offline` print the expected shapes; the latter exits `3` cleanly when `BUZZ_PRIVATE_KEY` is unset.

## Out of scope (per the issue's own staging)

- WebSocket readiness probe (flagged in the issue as possibly-too-big for v1).
- Deeper runtime diagnostics (per-channel permission probes, ACP runtime health — Desktop's Doctor surface already owns ACP).
- Integration test against a real relay (would need a fixture relay; matches how other read-only commands are covered today).

## Notes for reviewers

- The command is intentionally bind-mounted ahead of the credential gate; that's the load-bearing design decision, called out in code comments and module docs.
- `relay_reachable` and `nip11` share an HTTP request path (both `GET /` with the metadata accept header); currently issued as two probes for clarity, easy to fuse if a reviewer prefers.
- The existing `buzz users me` / `whoami` work in #2933 (when it lands) is complementary — it would give `doctor` a clean way to also report the resolved user profile, but isn't required for v1.

### DIFF
--- a/crates/buzz-cli/src/commands/doctor.rs
+++ b/crates/buzz-cli/src/commands/doctor.rs
@@ -1,0 +1,603 @@
+//! `buzz doctor` — non-mutating preflight diagnostics.
+//!
+//! Unlike every other subcommand, doctor intentionally does NOT require
+//! credentials: missing or malformed config must surface as a check result
+//! (status `error`) rather than an argument-time usage failure. The command
+//! never publishes events and never mutates local or remote state;
+//! `--offline` runs only local checks and marks remote probes as skipped.
+//!
+//! Output is structured JSON on stdout:
+//!
+//! ```json
+//! {"status":"warning","checks":[{"id":"identity","status":"ok","message":"..."}]}
+//! ```
+//!
+//! Exit behavior distinguishes a failed required check from warnings:
+//! `0` when every applicable check is `ok` (warnings allowed),
+//! `3` when any check is `error` and at least one error is auth/identity,
+//! `2` when any check is `error` and all errors are relay/network, and
+//! `1` for any other failure mix.  Skipped checks never affect the exit
+//! code.  The human-readable `status` field is `ok`, `warning`, or `error`
+//! mirroring the worst check outcome.
+
+use serde::Serialize;
+use serde_json::json;
+
+use crate::client::{normalize_relay_url, BuzzClient};
+use crate::error::CliError;
+
+/// Status of a single doctor check.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "lowercase")]
+enum Status {
+    Ok,
+    Warning,
+    Error,
+    Skipped,
+}
+
+impl Status {
+    fn as_str(self) -> &'static str {
+        match self {
+            Self::Ok => "ok",
+            Self::Warning => "warning",
+            Self::Error => "error",
+            Self::Skipped => "skipped",
+        }
+    }
+}
+
+/// Serialisable form of one check for the JSON output.
+#[derive(Debug, Serialize)]
+struct Check {
+    id: &'static str,
+    status: Status,
+    message: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    remediation: Option<String>,
+}
+
+impl Check {
+    fn ok(id: &'static str, message: impl Into<String>) -> Self {
+        Self {
+            id,
+            status: Status::Ok,
+            message: message.into(),
+            remediation: None,
+        }
+    }
+
+    fn warning(id: &'static str, message: impl Into<String>) -> Self {
+        Self {
+            id,
+            status: Status::Warning,
+            message: message.into(),
+            remediation: None,
+        }
+    }
+
+    fn error(
+        id: &'static str,
+        message: impl Into<String>,
+        remediation: Option<&str>,
+    ) -> Self {
+        Self {
+            id,
+            status: Status::Error,
+            message: message.into(),
+            remediation: remediation.map(str::to_string),
+        }
+    }
+
+    fn skipped(id: &'static str, message: impl Into<String>) -> Self {
+        Self {
+            id,
+            status: Status::Skipped,
+            message: message.into(),
+            remediation: None,
+        }
+    }
+}
+
+/// Entry point invoked from `crate::run` before the credential gate.
+///
+/// The doctor never returns `Err` for missing configuration; it reports
+/// the problem as a check and exits non-zero via `run_from_args`.
+pub async fn run(cli: &crate::Cli, offline: bool) -> Result<(), CliError> {
+    let mut checks: Vec<Check> = Vec::new();
+
+    // ---- local checks ----
+
+    // `cli.relay` itself is a free-form string; canonicalize before probing
+    // so the rest of the doctor matches the rest of the CLI.
+    let relay_url_raw = cli.relay.clone();
+    let relay_url = normalize_relay_url(&relay_url_raw);
+    if relay_url.is_empty() {
+        checks.push(Check::error(
+            "relay_url",
+            "relay URL is empty",
+            Some("Set BUZZ_RELAY_URL or pass --relay"),
+        ));
+    } else if !(relay_url.starts_with("http://") || relay_url.starts_with("https://")) {
+        checks.push(Check::error(
+            "relay_url",
+            format!("relay URL {:?} is not an http(s) URL after canonicalization", relay_url),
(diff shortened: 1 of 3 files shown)


iroiro147:
Added a follow-up commit `d576d88a` that tightens the exit-code classification. The first cut treated every authed-read failure as auth (exit 3), which was wrong for unreachable relays — on a DNS/connect failure the doctor should exit 2 and report the transport problem, not blame credentials. The new commit distinguishes true auth rejections (Auth variant or relay 401/403) from transport errors, and avoids double-reporting the reachability failure when both the public and authed probes go down. All 279 buzz-cli tests still pass; clippy clean.

iroiro147:

Issue #2931, opened 2026-07-26
Show what each agent can do and whether it is ready before delegation


BradGroux:
## Problem

Before delegating work, a Buzz user cannot answer a basic set of questions in one place:

- Which runtime and model will handle the task?
- Is that runtime installed, authenticated, connected, and currently healthy?
- Which capabilities did the runtime actually report?
- Which MCP/tool sources are attached?
- What permission mode is effective?
- When was this information last verified?

Buzz already observes most of this data, but it is split across runtime discovery, managed-agent process state, presence, and encrypted ACP observer frames. The UI exposes fragments of it in Doctor, configuration dialogs, agent rows, and the session transcript. It does not synthesize an agent-level capability/readiness manifest.

That creates delegation surprises. A saved model may not be applied, a runtime may be installed but logged out, an agent may be running but disconnected from the selected community, or a capability may be assumed even though the runtime never reported it.

Current source paths:

- `desktop/src-tauri/src/managed_agents/discovery/runtime_metadata.rs` is the authoritative static runtime catalog.
- `ManagedAgentSummary` and `ManagedAgentRuntimeStatus` carry process and community-pair lifecycle state.
- The encrypted owner-only observer stream already emits `agent_initialized`, including `agentInfo`, protocol version, and `agentCapabilities`.
- `session_config_captured` already carries model/config/mode state.
- `available_commands_update` is parsed for the transcript but not promoted to an agent-level readiness surface.

## Proposed first slice

Add an owner-visible capability and readiness manifest to the agent details surface. Synthesize it from existing authoritative sources rather than maintaining a second vendor matrix in React.

The manifest should show:

- runtime name and version, ACP protocol version, model, and provider;
- installation and authentication readiness;
- process, community connection, presence, and observer-stream health;
- runtime-reported input/output features such as images, audio, and embedded context;
- runtime-reported commands and tool sources;
- available tool descriptors when the runtime reports them, including source, risk class (`read`, `write`, `execute`, `external`, or `unknown`), and availability;
- requested and effective permission mode;
- platform/runtime limitations and explicitly unreported fields;
- the timestamp of the newest source observation used for the manifest.

The UI must distinguish:

- **reported**: the runtime or Buzz-owned source supplied the fact;
- **unavailable**: the source explicitly says the capability is absent;
- **unknown**: the runtime did not report enough information.

Unknown must never be rendered as unsupported or safe.

## Data and trust boundaries

[...]

PR #4066, opened 2026-08-01
feat(acp): allow host policy to decide permission requests

Silentpartnercoding:
Follow-up to the permission-boundary analysis in #2931.

Buzz’s generic ACP client currently selects `allow_once` whenever that option is offered. This change adds an injectable asynchronous permission-decision callback so the host can choose among the exact options supplied by the agent.

The callback receives the permission request and offered options. A policy decision may select only an option that was actually offered. If the policy is unavailable, times out, is cancelled, returns an error, or selects an invalid option, the client fails closed through an offered rejection option.

The focused tests cover:

- a deny decision emits a rejection response and never selects an allow option
- option kinds and IDs are preserved exactly
- string and numeric JSON-RPC request IDs
- missing policy, timeout, cancellation, callback error, and invalid selection
- response-write failure and cancellation cleanup

## Trust boundary

This governs cooperative operations that cross `session/request_permission`. It does not provide general filesystem, subprocess, network, MCP, runtime, or operating-system containment. Runtimes that perform work without issuing a permission request remain outside this policy boundary.

## Verification

- `cargo test -p buzz-acp`
- `cargo build -p buzz-acp`
- `cargo fmt --check`
- focused deny-path JSON-RPC regression tests
- `git diff --check`

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -329,6 +329,20 @@ The harness works with any agent that implements the [ACP spec](https://agentcli
 - Accept `session/prompt` with a text message and stream `session/update` notifications
 - Return a `stopReason` (`end_turn`, `cancelled`, `max_tokens`, etc.)
 
+### Permission-policy boundary
+
+`AcpClient` exposes an injectable asynchronous policy for
+`session/request_permission`. No policy means deny: Buzz selects an offered
+reject option when that is unambiguous, or returns `cancelled`. Callback errors,
+timeouts, invalid selections, and cancellation also fail closed. The historical
+`allow_once` behavior is available only through the explicit
+`AcpClient::allow_once_permission_policy()` compatibility policy.
+
+This is cooperative control of the ACP session/request_permission exchange only.
+It does not enforce or attest filesystem, subprocess, network, MCP, runtime, or
+operating-system containment. An agent or tool that does not ask is outside this
+policy boundary.
+
 Set `BUZZ_ACP_AGENT_COMMAND` and `BUZZ_ACP_AGENT_ARGS` to point at your agent binary.
 
 ## Testing
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -9,6 +9,9 @@
 //! 5. [`AcpClient::session_cancel`] / [`AcpClient::cancel_with_cleanup`] — cancel in-flight turn
 
 use futures_util::StreamExt;
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
 use tokio::io::AsyncWriteExt;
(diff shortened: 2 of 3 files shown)


Silentpartnercoding:
@BradGroux I implemented the narrow permission-decision callback experiment you outlined and kept the guarantee limited to cooperative `session/request_permission` traffic. The deny-path tests inspect the emitted JSON-RPC response and verify that no allow option is selected. I’d appreciate your review of the API shape, fail-closed behavior, and boundary wording.

Silentpartnercoding:
Quick relevance check now that Buzz’s ACP surface has continued moving: #4066 is still mergeable against current `main`, and its DCO, Semgrep, and zizmor checks remain green.

Since this PR opened, #4155 and its DCO-clean replacement #4333 have added a broader Guardian proposal that also changes `handle_permission_request`. I reviewed the overlap rather than assuming this PR was still unique.

The distinction I see is that #4333 derives an allow/reject choice from Buzz’s fixed permission modes, while #4066 exposes a generic asynchronous host-policy callback, preserves the agent’s exact offered option IDs and kinds, and handles timeout, cancellation, policy errors, and invalid selections explicitly.

One boundary I would appreciate maintainer direction on: #4066 currently fails closed when no callback is installed, changing the historical automatic `allow_once` behavior. I can instead make the callback opt-in and preserve compatibility until a caller installs it, or adapt the seam to the Guardian direction if that is now preferred.

Is the generic injectable seam still useful alongside #4333, or would you prefer that this PR be revised or closed in favor of the Guardian implementation?

Silentpartnercoding:
Revisiting my own default here after watching #4609 land and then get reverted.

As written, this PR fails closed when no policy is available — so a Buzz with nothing installed denies where it previously selected `allow_once`. That is a behaviour change for every existing user, and it is what turns this from a code review into a product decision. That is not a decision I should be asking a single reviewer to carry, and I would rather it not be one at all.

Two shapes, whichever you prefer:

1. **Default preserved.** No policy installed keeps the current `allow_once` selection. Fail-closed applies only when a policy *is* installed and is unavailable, times out, is cancelled, errors, or returns an option that was not offered. Strictness becomes opt-in rather than imposed.
2. **Seam only.** Land the injectable callback with every existing path behaving exactly as it does today, and take the strictness question separately whenever you want it.

Both keep the fail-closed machinery and the focused tests as they are — the only thing that moves is which branch handles "no policy at all". Option 1 is a small change and I am happy to push it on request; say which you would rather have and I will do the work.

Silentpartnercoding:
Closing as superseded by #5712, which has now landed Buzz-native permission gating at the actual LLM tool-call boundary, including fail-closed handling for malformed or invalid permission responses. With that implementation on main, maintaining this older generic ACP callback as a conflicting parallel path no longer looks useful. The upstream solution is the better integration point. Thanks to everyone who moved the permission boundary forward.

PR #4071, opened 2026-08-01
fix(desktop): show workflows across joined channels

aaron-acev:
## What changed

- build one kind `30620` workflow-definition filter per joined channel
- keep those filters in one relay query, relying on Nostr's OR semantics across filters
- add a regression test that requires each filter to carry exactly one `#h` value

## Why

The Workflows overview batched all joined channel IDs into one `#h` array.
Deployed relays may treat that array as one value, so workflows outside the
first channel disappear even though the CLI can read them. Sending one filter
per channel preserves the single relay round trip while making cross-channel
matching unambiguous.

## User impact

The Desktop Workflows overview can show definitions from every joined channel
instead of only the first matching channel.

## Validation

At exact head `6c3b65c47f9c294115cee2702278ed324533cba2`:

- `just desktop-ci` — PASS (3,906 frontend tests; 2,089 Tauri tests + 3 diagnostics; zero failures; frontend checks/build, Tauri fmt/check)
- `just desktop-tauri-clippy` — PASS with warnings denied
- semantic patch ID remained `a267ba514e489143e38a344f923f01a93f73ca55` after rebasing onto current `origin/main` `19d57b0d46baa55814ac737041a36d0b405c9f64`

Originating Buzz thread:
`buzz://message?channel=582358dd-f6db-455c-a451-3c29d006df08&id=2fb67c6cb57f7b1e34701b48a81806a604b57ae8562b026d7151c6a655d11242`

### DIFF
--- a/desktop/src-tauri/src/commands/workflows.rs
+++ b/desktop/src-tauri/src/commands/workflows.rs
@@ -70,11 +70,12 @@ pub async fn get_channel_workflows(
 ///
 /// The Workflows overview screen previously issued one `get_channel_workflows`
 /// query per member channel (`Promise.all` fanout in `WorkflowsView`), i.e. N
-/// relay POSTs. A nostr `#h` filter matches ANY of its listed values, so one
-/// query with all channel ids returns the same set. Each `WorkflowWire` carries
-/// its own `channel_id` (from the event's `h` tag), so the frontend can still
-/// group results by channel. Neither this nor the per-channel command sets a
-/// `limit`, so batching does not change result completeness.
+/// relay POSTs. Keep the single relay round-trip, but send one filter per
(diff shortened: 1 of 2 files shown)


aaron-acev:
Ready for review at exact head `6c3b65c47f9c294115cee2702278ed324533cba2`.

Independent implementation review found no code blocker. Local exact-head validation passed `just desktop-ci` (3,906 frontend tests; 2,089 Tauri tests plus 3 diagnostics), all-target Tauri clippy with warnings denied, the focused 12-test Rust workflow suite, formatting, and diff hygiene.

The upstream CI and Desktop Release Candidate runs currently show `action_required` with zero jobs because this is a fork PR. A maintainer still needs to approve those workflows and provide the required review; no merge is requested before both gates pass.

Issue #3227, opened 2026-07-27
RFC: App-integration agents — install a Buzz-managed agent that carries an app's own tools + scoped credentials (headless, owner-approved)


adamazad:
## Summary

A stable, scriptable way for a **third-party app** to install a **Buzz-managed
agent that carries the app's own tools + a scoped credential** — so the agent
does real domain work (not just chat) while Buzz hosts + supervises it. Today a
managed agent (Fizz/Bumble/Honey) is a general assistant with no app tooling, and
the only way to give one app-specific tools + secrets headlessly is blocked.

## Concrete use case (what drove this)

Figura is a design tool. We wanted a Buzz-native **Figura** agent: mention it in a
channel → it builds an on-brand UI mock ("fig") and replies with a link, answers
"what figs do we have?" via our CLI, iterates on a fig in-thread. That needs the
agent's harness to carry:

- the `figura` CLI + `/fig` skill (its tools),
- a **scoped team token** (which team's brand + where figs land),
- a persona/system prompt.

A `builtin:fizz`-style managed agent can't do this — it has no Figura tooling or
credential. So we ended up shipping a **self-hosted daemon** (own Nostr key,
`channels add-member`, launchd) as a workaround. That works, but:

- **It doesn't scale to co-hosts.** If two teammates each run the daemon in the
  same channels, both fire on `@figura` → duplicate replies, duplicate figs, each
  burning its own model subscription. There's no cross-daemon dedup. A single
  **Buzz-managed** agent per community would solve this by construction.

## Why the managed path is blocked today

1. **Managed agents can't carry app tools + secrets headlessly.** Desktop's
   Create-agent → Advanced sets `runtime.mcpCommand` + `env_vars`, but the CLI
   `agents draft-create` path is a deliberately **no-secret owner-review contract**
   (`parseAgentManagementRequest` drops any secret-shaped/unknown key). So an app
   installer can't hand over a token/tool via the draft. (#3115 tried the CLI flag
   and ran into exactly this.)
2. **Headless provisioning needs a Desktop-minted attestation.** `draft-create`
   requires `BUZZ_AUTH_TAG` (NIP-OA owner attestation), which only Desktop mints —
   there's no CLI to produce it. So the flow can't run from an app's `setup`.
3. **The deep-link prefill (#2911) still routes to the owner's Desktop form** and
   isn't in released builds.

Net: there's no path for "app X installs its working agent into my community" that
carries X's tools + a scoped secret, with owner approval but without hand-walking
[...]

Issue #3115, opened 2026-07-27
CLI: expose --mcp-command and --env on `buzz agents draft-create`/`draft-update` (GUI Advanced parity)


adamazad:
## Problem

`buzz agents draft-create` / `draft-update` can set `channel`, `display_name`,
`system_prompt`, `runtime`, `provider`, `model`, `respond_to` — but **not** the
harness's MCP command or extra env. Those are the two fields that let a managed
agent carry *external tools* (an app's own MCP server + the token it needs).

Today only Desktop's Create-agent → **Advanced** exposes them
(`desktop/src/features/agents/channelAgents.ts:237` → `runtime.mcpCommand`,
plus `env_vars` in `useGlobalAgentConfig.ts:18`). The harness already supports
an arbitrary MCP:

- `crates/buzz-acp/src/config.rs:261` — `#[arg(long, env = "BUZZ_ACP_MCP_COMMAND")] pub mcp_command: String`
- `crates/buzz-acp/src/lib.rs:4141` `build_mcp_servers()` — takes **any** command (derives name from path), not hardcoded to `buzz-cli`; injects `BUZZ_RELAY_URL` / `BUZZ_PRIVATE_KEY` into its env.
- `crates/buzz-acp/src/lib.rs:3721` — `extra_env` passthrough already exists.

So the capability is there; the CLI just can't reach it. Anyone scripting agent
provisioning (CI, an app's own `setup` command) is forced through the Desktop GUI.

## Ask

Add to `buzz agents draft-create` and `draft-update`:

```
--mcp-command <cmd>        # BUZZ_ACP_MCP_COMMAND for the harness
--env KEY=VALUE            # repeatable; merged into the MCP subprocess env (extra_env)
```

Wire them into the same request the Desktop **Advanced** panel already sends
(`runtime.mcpCommand`, `env_vars`) — no new harness behavior, just CLI parity
with the GUI.

## Why

Lets an app provision a fully-configured managed agent from one scripted command
— e.g. a `<app> mcp` tool server + its `<APP>_TOKEN` — instead of walking a user
through the Advanced GUI. Same motivation as the install deep-link (block/buzz#2911):
make the agent path scriptable end-to-end.

Happy to send the PR if the shape looks right.

adamazad:
[...]

PR #4078, opened 2026-08-01
feat(desktop): support MCP sidecars for custom harnesses

### LINKED ISSUE #3385: [Bug] Custom/BYOH ACP harness receives no authenticated Buzz reply tool, so completed turns cannot post
Describe the bug
We are trying to run Buzz Desktop as macOS user A while keeping an existing Hermes backend and profile isolated under macOS user B. Buzz owns channel membership and lifecycle, while the custom ACP harness crosses the user boundary through a narrowly scoped command:

/usr/bin/sudo -n -H -u <user-b> /Users/<user-b>/.hermes/hermes-agent/venv/bin/hermes-acp

The process topology, ACP initialization, and prompt delivery all succeed. The first real mention reaches Hermes, but Hermes cannot post its response to the channel. Its attempt to run buzz messages send exits with:

auth_error: BUZZ_PRIVATE_KEY is required

The ACP turn and process chain remain healthy, so this presents as a successful turn with a silently missing reply, similar to #2459 and #2698.

Verified failure boundary
 
This does not appear to be caused simply by sudo stripping inherited environment variables.

For a managed agent, Buzz Desktop derives the companion MCP command from the effective agent command using the known-runtime catalog. With the explicit custom command set to /usr/bin/sudo, the command is not recognized as a built-in runtime, so the effective MCP command is empty.

As a result:

Buzz correctly launches the custom ACP process.
ACP initialization succeeds.
buzz-acp does not include the authenticated buzz-dev-mcp companion in session/new.
The turn prompt still instructs the agent to reply using buzz messages send.
Hermes falls back to its ordinary terminal, which does not—and should not—contain the Buzz signing key.
The CLI returns BUZZ_PRIVATE_KEY is required, and no channel reply appears.

Hermes Agent ACP 0.19.0 supports ACP-provided mcpServers; the missing piece is that Buzz does not provide its authenticated companion MCP server for this custom effective command.

The cross-user boundary is relevant because copying the private key into user B’s environment, preserving it through sudo, or installing another credential-bearing Hermes instance would defeat the intended isolation. It is not the desired fix.

Steps to reproduce
Run Buzz Desktop under macOS user A.
Configure a custom ACP harness with:
command: /usr/bin/sudo
args: -n -H -u <user-b> <absolute-path-to-hermes-acp>
Pin that command and argument override to one owner-only, mention-triggered managed instance with parallelism 1.
Give user B access only to the required Buzz working directory.
Permit only the exact Hermes ACP executable through a narrow passwordless sudo rule.
Start the instance and confirm one initialized ACP worker.
Mention the agent in its Buzz channel.
 
Observe that Hermes processes the turn and attempts buzz messages send, but the CLI reports that BUZZ_PRIVATE_KEY is required and no reply appears.

Observed sanitized process tree:

user-a buzz-desktop
└─ user-a buzz-acp
   └─ root sudo -n -H -u user-b .../hermes-acp
      └─ user-b python .../hermes-acp

Verified:

One buzz-acp supervisor
One initialized worker
Pool ready
Hermes Agent ACP 0.19.0
No bundled buzz-agent child
Stable process chain throughout the failed turn
No authenticated Buzz MCP companion available to Hermes

Expected behavior
A custom/BYOH ACP harness should receive a scoped method for posting its managed agent’s reply without:

exposing BUZZ_PRIVATE_KEY to the agent’s ordinary terminal
passing the raw key through sudo
copying credentials between user homes
weakening sudo environment restrictions
installing a second credential-bearing agent backend

Possible directions:

Allow custom harness definitions or managed instances to specify a companion MCP command, so Buzz can provide buzz-dev-mcp through ACP session/new.
Have buzz-acp own final-response publication and detect/retry delivery failures.
Provide a typed, narrowly scoped, harness-owned reply capability rather than exposing the raw signing key, consistent with the security concern in #2883.
If custom harnesses are not intended to receive Buzz reply tooling, reject or warn on configurations where the prompt requires buzz messages send but no MCP companion is configured.

Questions for maintainers
Should custom/BYOH harnesses receive Buzz’s authenticated companion MCP server?
Is there an intended way to associate buzz-dev-mcp with a custom command that is not in known_acp_runtime?
Would adding a companion mcpCommand field to custom harness definitions be the preferred design?
 
Longer term, should reply publication live in buzz-acp rather than depend on the model invoking the CLI?
Is anyone successfully running Hermes, OpenClaw, or another existing operator brain behind Buzz across a user, container, or service boundary?

Related issues
#2459 — successful ACP turn with a silently lost CLI reply
#2698 — reply delivery depends on the agent invoking buzz messages send
#2883 — raw BUZZ_PRIVATE_KEY should not enter agent-controlled process environments
#3115 — MCP command/environment provisioning for managed agents
#2663 — external agent systems and existing operator brains

Version and platform
Buzz Desktop: 0.5.0
OS: macOS
Custom backend: Hermes Agent ACP 0.19.0
Relay: communities.buzz.xyz

Logs / additional context
Sanitized failure:

auth_error: BUZZ_PRIVATE_KEY is required

No private keys, auth tags, complete managed-agent stores, or credential values are included.

fredamn76:
## Summary

- persist an optional MCP sidecar command with custom harness definitions
- resolve built-in and custom MCP commands through the shared effective harness descriptor
- use the effective sidecar consistently for managed-agent summaries, process spawn, and restart hashing
- expose the optional command in the custom harness form with an availability check

Custom ACP harnesses currently lose the companion MCP server because Desktop derives it only from the built-in runtime catalog. The ACP process can start successfully, but the agent receives no custom tool server (and, for Buzz reply tooling, no authenticated companion).

This keeps the harness definition as the source of truth instead of restoring a stale per-agent snapshot field. Existing custom harness JSON remains compatible because `mcpCommand` is optional. Built-in runtimes retain their current catalog-derived MCP command, and the sidecar is started only for agents selecting the custom harness.

### Related issue

Closes #3385.

Related: #3115 and #3227 cover provisioning and the broader app-agent installation model; this PR is limited to Desktop's custom-harness runtime contract.

Originating Buzz conversation:

- channel: `recepie-mealie-mcp` (`34d45264-ea10-4065-82d6-831a43a90c84`)
- thread root: `0907968b2234880045f1edd8dd8f8a22d4ed32b2e98294361094b9537e8e2ed8`

### Testing

Passed on the code tree now published as commit
`ef41e584f78ec1d47fae2344352ee28531f3fcb0`:

- `just desktop-check`
- `just desktop-typecheck`
- `just desktop-build`
- `just desktop-test` — 3908 passed
- `pnpm -C desktop build:e2e`
- `playwright test tests/e2e/harness-catalog-screenshots.spec.ts --project=smoke` — 1 passed
- `git diff --check origin/main...HEAD`

The original tested commit was rewritten only to replace a GitHub-blocked private
author email with the account's noreply address. Its tree hash is unchanged at
`e443fb05f3bb857927f614cf730d505b5eab821c`.

Not run locally: the full Tauri Rust test suite. This Linux machine does not have the GLib/GObject/WebKitGTK development packages required by Buzz's documented Tauri setup, and installing system packages requires administrator authentication. GitHub CI should exercise that gate.

### Manual verification

1. Open Settings → Agent runtimes → Custom harness.
2. Enter an ACP command and an optional MCP command, then save.
3. Re-open the custom harness and verify both commands round-trip.
4. Assign an agent to the custom harness and start it.
5. Verify the managed-agent summary and startup log show the custom MCP command.
6. Change only the MCP command and verify the spawn hash triggers a restart.
7. Verify an agent on a built-in runtime still uses its catalog MCP default.

### UI

Before/after screenshots are posted in the PR comments using the repository's screenshot helper.

### DIFF
--- a/desktop/src-tauri/src/commands/agent_discovery.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery.rs
@@ -173,7 +173,7 @@ pub async fn save_custom_harness(
         command: command_opt,
         binary_path,
         default_args,
-        mcp_command: None,
+        mcp_command: definition.mcp_command.clone(),
         model_env_var: None,
         provider_env_var: None,
         thinking_env_var: None,
--- a/desktop/src-tauri/src/managed_agents/custom_harnesses.rs
+++ b/desktop/src-tauri/src/managed_agents/custom_harnesses.rs
@@ -56,6 +56,9 @@ pub(crate) struct HarnessDefinition {
     /// Default CLI arguments passed to the command (array, not split-string).
     #[serde(default)]
     pub args: Vec<String>,
+    /// Optional MCP sidecar passed to `buzz-acp` for this harness.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub mcp_command: Option<String>,
     /// Environment variables injected at spawn time. Definition env is applied
     /// first and LOSES on conflict with Buzz-injected vars — `BUZZ_MANAGED_AGENT`
     /// is always authoritative and cannot be overridden here.
@@ -172,6 +175,9 @@ fn validate_harness_definition(def: &HarnessDefinition) -> Result<(), String> {
     }
     if def.label.trim().is_empty() {
         return Err("label must not be empty".into());
+    }
+    if matches!(def.mcp_command.as_deref(), Some(command) if command.trim().is_empty()) {
+        return Err("mcpCommand must not be blank when provided".into());
     }
     // Args travel to the harness through the comma-delimited
     // `BUZZ_ACP_AGENT_ARGS` env transport (clap `value_delimiter = ','` on the
@@ -733,18 +739,13 @@ mod tests {
         assert_eq!(loaded[0].id, "custom-dup");
     }
 
-    // ── Round-trip via save_custom_harness_to_dir (B-4) ─────────────────────
-    //
-    // These tests exercise the REAL persistence helper, not raw fs::write.
-    // They prove: create, same-ID edit (backup-swap), rename (old file removed),
-    // backup file cleaned up on success.
-
     fn make_def(id: &str, label: &str) -> HarnessDefinition {
         HarnessDefinition {
             id: id.to_string(),
             label: label.to_string(),
             command: format!("{id}-bin"),
             args: vec![],
+            mcp_command: None,
             env: BTreeMap::new(),
             install_instructions_url: String::new(),
             install_hint: String::new(),
@@ -773,11 +774,9 @@ mod tests {
         let v1 = make_def("my-harness", "V1 Label");
         save_custom_harness_to_dir(dir.path(), &v1, None).unwrap();
 
-        // Same-ID edit: label changes.
         let v2 = make_def("my-harness", "V2 Label");
         let outcome = save_custom_harness_to_dir(dir.path(), &v2, None).unwrap();
 
-        // No old-path reported (id unchanged).
         assert!(outcome.removed_old_path.is_none());
 
         let loaded = load_custom_harnesses(dir.path());
@@ -794,7 +793,6 @@ mod tests {
         let v2 = make_def("my-harness", "V2");
         save_custom_harness_to_dir(dir.path(), &v2, None).unwrap();
 
-        // .bak file must be gone after a successful commit.
         let bak = dir.path().join("my-harness.json.bak");
         assert!(
             !bak.exists(),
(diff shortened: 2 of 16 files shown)


fredamn76:
## Custom harness MCP sidecar

Before — a custom harness could configure only the ACP command:

BEFORE

After — an optional MCP command is available and explicitly scoped to agents using that harness:

AFTER

dean-yates:
I reproduced the underlying missing authenticated reply-capability failure on **Buzz Desktop 0.5.10** with both the built-in Hermes preset and profile-isolated custom Hermes ACP harnesses. The managed agent receives the event and completes the turn, but the generated response remains in Activity because the explicit publish step has no authenticated companion.

I also checked this PR against current `main` (`9e0c6b43`). The patch no longer applies cleanly because the managed-agent readiness/spawn-hash code has moved since the PR base, while current `main` still appears to lack an `mcpCommand` field on `HarnessDefinition` and still derives the effective sidecar only from the known-runtime catalog.

I would be happy to help port this focused custom-harness sidecar support to current `main` and test it against the live 0.5.10 Hermes reproduction. To avoid opening a competing PR, would you prefer:

1. an updated commit/PR based on this implementation, with attribution and current-main regression tests; or
2. test/review evidence posted here while the original author rebases?

The verification I can provide is deliberately credential-free: custom harness round-trip, effective spawn descriptor, restart-hash change, ACP `session/new` receiving the sidecar, and real DM/channel publication without exposing the managed identity key to Hermes's ordinary terminal.

PR #4084, opened 2026-08-01
fix(acp): preserve JSON-RPC error.data when message is present (#4069)

### LINKED ISSUE #4069: Windows on ARM64: Claude managed agents fail every session/new with -32603 (win-x64 Node provisioned on ARM64 host; error.data dropped from logs)
## Summary

On Windows 11 ARM64, every Claude-backed managed agent fails every room mention with:

```
Agent reported error (code -32603): Internal error
```

while OpenAI-backed managed agents in the same room work normally. Two separate defects combine here:

1. **Buzz provisions a win-x64 Node runtime on ARM64 hosts**, which breaks architecture-sensitive native-binary resolution inside Node-based ACP adapters (`@agentclientprotocol/claude-agent-acp`).
2. **`buzz-acp` discards the JSON-RPC `error.data` payload when logging agent errors**, which hid the adapter's fully actionable error message, making this look like an opaque Claude failure.

## Environment

- Windows 11 Home ARM64 (10.0.26200) — Snapdragon-class machine
- Buzz desktop 0.5.3 (x64 build, running under Windows x64 emulation)
- `@agentclientprotocol/claude-agent-acp` 0.64.0, installed into the global npm prefix using the system **arm64** Node v24.8.0 / npm
- Claude Code CLI 2.1.220, claude.ai subscription auth (`claude auth status` healthy)
- Buzz-managed Node runtime: `%APPDATA%\Buzz\runtimes\node\v24.18.0\win-x64\node.exe` (PE header: x64)

## Root cause chain

1. `buzz-desktop.exe` and `buzz-acp.exe` are x64 binaries. On an ARM64 host they run under emulation, and `managed_node_paths.rs` selects the Node platform via `std::env::consts::ARCH` — a **compile-time** constant that is `"x86_64"` inside an x64 build regardless of the real machine. So Buzz provisions **win-x64** Node. (The `("windows", "aarch64") => ("win-arm64", ...)` arm added on `main` can never be taken by an x64-built binary.)
2. The npm shim `claude-agent-acp.cmd` resolves `node` from `PATH`; under Buzz that is the bundled win-x64 Node, so the adapter runs with `process.arch === "x64"`.
3. On `session/new`, the adapter's `claudeCliPath()` resolves the SDK CLI as `@anthropic-ai/claude-agent-sdk-win32-x64/claude.exe`. Because the adapter was installed by an **arm64** npm, only `@anthropic-ai/claude-agent-sdk-win32-arm64` is vendored. Resolution fails and the adapter throws:

   > Claude native binary not found for win32-x64. Reinstall @anthropic-ai/claude-agent-sdk without --omit=optional, or set CLAUDE_CODE_EXECUTABLE.

4. The ACP TypeScript SDK wraps this plain exception as `{ code: -32603, message: "Internal error", data: { details: "<the message above>" } }`.
5. `agent_error_from_json` in `crates/buzz-acp/src/acp.rs` keeps only `code` and `message` whenever `message` is a string — `error.data` is dropped, so the actionable text in step 3 never appears in any Buzz log. Operators see only a generic `agent_returned … error=Agent reported error (code -32603): Internal error`.

Because the failure is thrown before the CLI is ever spawned, no amount of model-id, parallelism, or permission-mode configuration changes the outcome — which cost us several hours of ruling those out.

## Reproduction (standalone, no Buzz needed)

Pipe ACP over stdio into the adapter, once under each Node:

```
initialize        {"protocolVersion":2, "clientInfo":{"name":"buzz-acp","version":"0.5.3"}, ...}
session/new       {"cwd":"<absolute existing dir>", "mcpServers":[], "_meta":{"sessionTitle":"repro"}}
```

| Adapter launched under | session/new result |
|---|---|
| Buzz-bundled win-x64 Node v24.18.0 | `-32603 Internal error`, `data.details` = "Claude native binary not found for win32-x64…" |
| System arm64 Node v24.8.0 | success |
| win-x64 Node + `CLAUDE_CODE_EXECUTABLE` → the vendored win32-arm64 `claude.exe` | success |

## Workaround (confirmed working)

Add to the Claude agents' `env_vars` in `managed-agents.json`:

```json
"env_vars": {
  "CLAUDE_CODE_EXECUTABLE": "C:\\Users\\<user>\\AppData\\Roaming\\npm\\node_modules\\@agentclientprotocol\\claude-agent-acp\\node_modules\\@anthropic-ai\\claude-agent-sdk-win32-arm64\\claude.exe"
}
```

The x64-emulated adapter process spawns the ARM64 CLI without issue.

## Suggested fixes

1. **Runtime arch detection for the managed Node runtime** — decide win-x64 vs win-arm64 from the *OS* architecture at runtime (e.g. `IsWow64Process2`, or Rust: checking `PROCESSOR_ARCHITEW6432`/`RuntimeInformation`-equivalent), not `std::env::consts::ARCH`; or ship a native ARM64 build of Buzz.
2. **Stop dropping `error.data`** — `agent_error_from_json` should append `data` (or at least `data.details`) to the logged message when present. The adapter had put the exact remedy on the wire; Buzz logged "Internal error".

Possibly related in spirit but distinct from #2698 / #2692 (those are model-picker issues; this fails before any model is reached).

## Also affected

The bundled Claude installer flow has the same arch blindness: `install-claude.log` on this machine shows Buzz fetching `claude-2.1.220-win32-x64.exe`, which then crashed with `0xC0000005` on the ARM64 host.

iroiro147:
## Summary

Fixes #4069 (partial — covers the cross-platform observability defect; the Windows ARM64 Node provisioning half is left to a follow-up).

`agent_error_from_json` in `crates/buzz-acp/src/acp.rs` was discarding the JSON-RPC `error.data` payload whenever the agent returned a string `message`. The ACP TypeScript SDK wraps plain thrown exceptions as:

```json
{"code": -32603, "message": "Internal error", "data": {"details": "<fully actionable adapter message>"}}
```

so the actionable adapter text never reached Buzz logs — operators saw only `Agent reported error (code -32603): Internal error`.

## What changed

When `message` is a string and a non-null `data` field is present, append a compact form of `data` to the surfaced message:

- Prefer a nested `data.details` string when present (matches the ACP TS SDK's wrap shape).
- Otherwise fall back to the `data` value stringified (handles plain string and structured `data` cases).
- Truncate defensively at 2048 chars so a pathological adapter cannot flood logs.
- `data: null` is ignored (no information).
- Errors without a string `message` still fall back to the full JSON object as before — no behavior change there.

## Test plan

Five new unit tests in `crates/buzz-acp/src/acp.rs::tests`:

- `agent_error_from_json_appends_data_details_when_present` — the Windows ARM64 shape from #4069: `Internal error` + `data.details` with the actionable remediation. Asserts both halves are surfaced.
- `agent_error_from_json_appends_string_data_when_details_missing` — plain string `data`.
- `agent_error_from_json_ignores_null_data` — `data: null` produces the original message unchanged.
- `agent_error_from_json_truncates_oversized_data` — 4096-char `data` is truncated with ellipsis.
- Existing `agent_error_from_json_uses_message_field_when_present` (no `data` key) — unchanged.

Verification:

```
cargo check -p buzz-acp          # clean
cargo test -p buzz-acp --lib     # 665 passed, 0 failed
cargo test -p buzz-acp agent_error_from_json   # 6 passed
```

## Notes

The Windows ARM64 side of #4069 (Buzz provisioning win-x64 Node on an ARM64 host because `managed_node_paths.rs` reads `std::env::consts::ARCH` at compile time) is **not** addressed here — that's a runtime detection change requiring `IsWow64Process2` / `PROCESSOR_ARCHITEW6432` and would benefit from maintainer input on supported host matrix. This PR ships just the cross-platform logging fix so the next adapter failure of this shape surfaces its own remedy in the log.

Closes #4069

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -111,11 +111,45 @@ pub enum AcpError {
 /// Build an [`AcpError::AgentError`] from a JSON-RPC error object,
 /// preserving the numeric code. When the `message` field is missing or
 /// non-string, fall back to the full JSON object so provider-specific
-/// detail (e.g. a `data` field) is not lost.
+/// detail (e.g. a `data` field) is not lost. When `message` is present
+/// and a `data` field exists, append the `data` payload to the message so
+/// actionable detail (e.g. an adapter's remediation hint) is not dropped.
 fn agent_error_from_json(error: &serde_json::Value) -> AcpError {
     let code = error.get("code").and_then(|c| c.as_i64()).unwrap_or(-32000);
     let message = match error.get("message").and_then(|m| m.as_str()) {
-        Some(m) => m.to_string(),
+        Some(m) => {
+            let mut msg = m.to_string();
+            if let Some(data) = error.get("data") {
+                if !data.is_null() {
+                    // Prefer a nested human-readable `details` string when
+                    // present (the ACP TS SDK wraps plain exceptions as
+                    // `{code, message, data: {details}}`); otherwise fall
+                    // back to a stringified form of the whole `data` value.
+                    let data_str = data
+                        .get("details")
+                        .and_then(|d| d.as_str())
+                        .map(str::to_string)
+                        .unwrap_or_else(|| match data.as_str() {
+                            Some(s) => s.to_string(),
+                            None => data.to_string(),
+                        });
+                    // Truncate defensively so a pathological adapter
+                    // cannot flood logs with megabytes of `data`.
+                    const MAX_DATA_LEN: usize = 2048;
+                    let truncated = if data_str.chars().count() > MAX_DATA_LEN {
+                        let cut: String = data_str.chars().take(MAX_DATA_LEN).collect();
+                        format!("{cut}…")
+                    } else {
+                        data_str
+                    };
+                    if !truncated.is_empty() {
+                        msg.push_str(": ");
+                        msg.push_str(&truncated);
+                    }
+                }
+            }
+            msg
+        }
         None => error.to_string(),
     };
     AcpError::AgentError { code, message }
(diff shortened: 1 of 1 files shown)


iroiro147:

PR #4092, opened 2026-08-01
fix(buzz-acp): anchor human-facing thread replies to the triggering message

### LINKED ISSUE #4072: Agents should reply to the exact message that triggered them
### Motivation

When a user is already talking with an agent inside a thread, the agent should make it clear which message it is answering. Today, the response can point to the thread root instead of the exact triggering message, which makes conversations with multiple messages difficult to follow and makes the agent feel less proactive.

### Proposed solution

For human-facing agent turns:

- Reply to the exact message that triggered the agent turn.
- Preserve the author mention for that parent message.
- Keep top-level replies anchored to the triggering message, which becomes the new thread root.
- Preserve the existing agent-to-agent behavior so coordination turns are not unnecessarily flattened.

Add regression tests covering nested human replies, top-level messages, agent-only mentions, and agent-to-agent turns.

### Alternatives considered

Keep anchoring every human-facing reply to the thread root. This keeps threads flat, but loses the direct relationship between the agent response and the specific message being answered.

### Additional context

This behavior was observed while using Buzz Desktop. The existing message UI already supports rendering direct replies; the change is to make the agent harness consistently target the triggering event.

iroiro147:
## Problem

When a human is already talking with an agent inside a thread, the agent's reply previously **flattened to the thread root** regardless of which message in the thread actually triggered the turn. With multiple messages in flight, the reader could not tell which message the agent was answering, and the agent felt less proactive (#4072).

DM turns already anchored to the exact triggering message; the non-DM thread branch was the inconsistent one.

## Fix

`resolve_reply_anchor()` now returns `triggering_event_id` in **every** human-facing case:

| Case        | Before              | After                       |
| ----------- | ------------------- | --------------------------- |
| In a thread | thread `root`       | triggering event id         |
| Top-level   | triggering event id | triggering event id (same)  |
| Agent↔agent | `None` (deep nest)  | `None` (unchanged)          |

The `--reply-to` hint now points at the exact message that triggered the turn, so the UI renders the agent's reply directly under the message being answered, preserving the author mention for that parent.

The escape hatch is preserved: the hints text still says agents may honor an explicit "post in channel root" user request by omitting `--reply-to`, and `agent↔agent` turns still get no forced anchor so coordination turns stay deeply nested.

## Tests

- **Regression (inverted):** `test_human_thread_reply_anchors_to_root_not_triggering_or_parent` → `…to_triggering_event_not_root_or_parent` (asserts anchor=triggering event, NOT root, NOT intermediate parent).
- **New depth≥2 regression:** `test_anchor_human_nested_reply_in_dm` and the comment above the `is_dm` branch were already there; now anchor=TRIGGER_ID (was ROOT_ID).
- `test_reply_instruction_present_for_channel_thread_reply` — now asserts `--reply-to <triggering event id>` instead of `<root id>`, plus explicit NOT-root assertion.
- `test_reply_instruction_allows_explicit_root_post_requests` — anchor assertion switched to triggering id; explicit-root escape-hatch text assertion preserved unchanged.
- `test_reply_instruction_batched_last_event_is_threaded` — asserts anchor is the batched last event's own id.
- `test_steer_cross_thread_reply_targets_steering_message` — asserts anchor is the steering message id (not its thread root), matching the new human-facing supersede semantics.

## Verification

```
cargo test   -p buzz-acp --lib                          # 663 passed, 0 failed
```

Fixes #4072.

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -1272,11 +1272,38 @@ pub fn resolve_channel_filters(
             }
         }
         SubscribeMode::All => {
+            // #4086: when no `BUZZ_ACP_KINDS` override is set, default to a curated
+            // list instead of a true wildcard. The previous wildcard behavior let
+            // two or more sibling agents storm indefinitely on each other's 👀/💬
+            // (kind 7) and their NIP-09 deletions (kind 5) — harness bookkeeping,
+            // not user intent. The curated list below keeps every kind agents
+            // actually need to wake on (chat, DMs, stream reminders, canvas,
+            // workflow events) and excludes the reaction/bookkeeping kinds.
+            //
+            // Users can still opt into the genuine wildcard by setting
+            // `BUZZ_ACP_KINDS=` explicitly (including `5` and `7`); the override
+            // remains authoritative.
+            let default_kinds = config.kinds_override.clone().unwrap_or_else(|| {
+                use buzz_core::kind::{
+                    KIND_CANVAS, KIND_GIFT_WRAP, KIND_STREAM_MESSAGE_EDIT,
+                    KIND_STREAM_MESSAGE_V2, KIND_WORKFLOW_TRIGGER,
+                };
+                vec![
+                    KIND_STREAM_MESSAGE,              // 9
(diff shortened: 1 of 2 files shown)


iroiro147:

PR #4093, opened 2026-08-01
feat(buzz-cli): support direct relay pubkeys in channel-template rosters

### LINKED ISSUE #3953: fix: channel templates should not silently omit remote managed agents from the roster
## Problem

A channel template can contain an agent roster expressed as Desktop persona/team IDs. When the intended agents are long-running provider/VPS identities rather than Desktop persona instances, `buzz channels create --template` still creates the channel and applies its metadata/canvas, but no matching member is added for those roster entries.

The command currently reports unresolved entries in `skipped`, but roster omission is not treated as an incomplete template application. In Desktop/operator use this looks like a successful template creation while the people/agents expected to receive mentions are absent. The failure is easy to miss and leaves workstream handoffs open.

This is distinct from #2728: groups would make bulk membership easier, but existing templates still need deterministic behavior when their saved roster cannot resolve.

## Reproduction

1. Run a managed agent on a remote/VPS provider as a stable Nostr pubkey that is not represented by a local Desktop persona instance.
2. Save a channel template whose roster refers to the expected persona/team.
3. Create a channel with that template.
4. The channel, canvas, visibility, and description are applied, but the remote agent is absent. An unrecognized team ID fails loudly; an unresolved persona entry is only returned as `skipped`.

## Expected

At least one of:

- support direct member/agent pubkeys in `TemplateAgentRoster`, so provider-backed identities can be rostered deterministically; or
- make unresolved requested roster entries an explicit incomplete/failure result requiring acknowledgement, rather than a successful template application.

The strongest flow is atomic from the operator perspective: create → resolve/add requested members → verify the resulting roster, with a machine-readable failure if any required member is missing.

## Current safe workaround

Leave the template roster empty, apply its channel/canvas settings, then add the required pubkeys explicitly with `buzz channels add-member` and verify them with `buzz channels members` before declaring channel creation complete.

iroiro147:
## Problem

A channel template's `agents` roster previously only accepted Desktop persona ids (which resolve through kind:30176/30177 events) and team ids. Long-running remote or VPS-managed agent identities — stable Nostr pubkeys with no local Desktop persona instance — could not be deterministically rostered. They ended up in the report's `skipped` bucket with the channel otherwise created successfully, leaving the operator thinking the roster was applied when it wasn't (#3953).

## Fix

This implements the issue's **option (a)**: add a third roster variant `agentPubkeys: [{ pubkey, label? }]` that bypasses the persona→kind:30177 slug scan entirely and adds the pubkey directly as a channel member. Template authors now have a first-class deterministic path for provider-backed agent identities.

Three behavior guarantees:

1. **Direct passthrough** — the pubkey is passed verbatim to `build_add_member`. A malformed or short pubkey surfaces as a `member_failures` entry, never silently dropped, never in `skipped`.
2. **Readable fallback labels** — when `label` is omitted the template report uses the first 16 chars of the pubkey (`pubkey:<16>`) instead of dumping the full 64-hex.
3. **Backward compatible** — templates written before this field existed deserialize with `agent_pubkeys: []` and behave exactly as before.

## Code

- `crates/buzz-cli/src/commands/channel_templates.rs` — new `TemplatePubkeyEntry { pubkey, label }`; `TemplateAgentRoster` gains `#[serde(default)] agent_pubkeys: Vec<TemplatePubkeyEntry>` (camelCase JSON `agentPubkeys`).
- `crates/buzz-cli/src/commands/channels.rs` — new pure helper `expand_direct_pubkeys` converts the entries verbatim into `ResolvedAgent`s; `build_roster_resolution` collects them alongside the existing persona/team slug scan and merges them into `RosterResolution.agents` before any channel creation.

## Tests

New unit coverage in `channels.rs`:
- `expand_direct_pubkeys_uses_label_when_present`
- `expand_direct_pubkeys_falls_back_to_truncated_pubkey_label` — includes a short/malformed pubkey that must not panic and must truncate at its actual length
- `expand_direct_pubkeys_empty_when_no_entries`

New/extended coverage in `channel_templates.rs`:
- `load_templates_parses_direct_agent_pubkeys` — full camelCase JSON round-trip, optional label
- `load_templates_rejects_invalid_pubkey_entry` — malformed entry (missing pubkey) fails parse; the silent-omission regression vector from the persona path is closed
- `load_templates_parses_full_roster` — extended to assert `agent_pubkeys` defaults to empty on pre-#3953 templates

## Verification

```
cargo check  -p buzz-cli                              # OK
cargo test   -p buzz-cli --lib                        # 276 passed, 0 failed
cargo clippy -p buzz-cli --all-targets -- -D warnings # OK
```

Fixes #3953.

### DIFF
--- a/crates/buzz-cli/src/commands/channel_templates.rs
+++ b/crates/buzz-cli/src/commands/channel_templates.rs
@@ -39,12 +39,29 @@ pub struct TemplateAgentRoster {
     pub personas: Vec<TemplateAgentEntry>,
     #[serde(default)]
     pub teams: Vec<TemplateTeamEntry>,
+    /// #3953: direct relay pubkeys for agents that are long-running remote
+    /// identities (provider/VPS) rather than Desktop persona instances.
+    /// Roster entries that use this variant skip the persona→kind:30177
+    /// resolution entirely and add the pubkey directly — deterministic.
+    #[serde(default)]
+    pub agent_pubkeys: Vec<TemplatePubkeyEntry>,
 }
 
 #[derive(Debug, Clone, Deserialize)]
 #[serde(rename_all = "camelCase")]
 pub struct TemplateAgentEntry {
     pub persona_id: String,
+}
+
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TemplatePubkeyEntry {
+    /// 64-char hex Nostr pubkey of the agent to add as a member.
+    pub pubkey: String,
+    /// Optional human-readable label for the template report. Falls back to
+    /// a truncated pubkey if absent.
+    #[serde(default)]
+    pub label: Option<String>,
 }
 
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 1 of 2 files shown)


iroiro147:

PR #4097, opened 2026-08-01
relay: write audit entry for NIP-09 event deletion

### LINKED ISSUE #4034: NIP-09 event deletion never writes an audit log entry (EventDeleted is defined but has no production call site)
**Summary**

`AuditAction::EventDeleted` is defined but has no production call site anywhere in the codebase. Neither of the two paths that delete/tombstone an event (the standard NIP-09 soft-delete, or the hard-delete-on-supersede path for parameterized-replaceable events) ever writes an audit log entry. This relay currently cannot produce an audit trail for any event deletion.

**Where**

- `crates/buzz-audit/src/action.rs` (~lines 8-31) defines `AuditAction` with 11 variants including `EventDeleted` (`event_deleted` in storage). Searching the whole tree, it appears in exactly two places: the enum definition itself, and one `#[cfg(test)]` unit test (`crates/buzz-audit/src/service.rs`, `verify_detects_tampering_within_a_community`) that uses it only as a generic sample action to exercise hash-chain tamper detection — not a real deletion code path.

- NIP-09 deletion dispatch is in `crates/buzz-relay/src/handlers/side_effects.rs`:
  - `handle_delete_event_side_effect` (standard `e`-tag deletion) calls `soft_delete_event_and_update_thread`.
  - The `a`-tag (addressable/parameterized-replaceable) NIP-09 deletion path also calls `soft_delete_event_and_update_thread`.
  
  Neither call path touches `buzz_audit` — there's no `audit_tx.send(...)` anywhere near either handler. Contrast with the two paths that *do* audit correctly on the same pattern:
  - `crates/buzz-relay/src/handlers/event.rs` (`enqueue_event_created_audit`) sends `AuditAction::EventCreated` after every successful event write.
  - `crates/buzz-relay/src/api/media.rs` sends `AuditAction::MediaUploaded` after a successful Blossom upload (explicitly commented as following "the same pattern as event audit").
  
  No equivalent `enqueue_event_deleted_audit` exists.

- Both deletion mechanisms confirmed:
  - **Soft delete / tombstone** (the normal user-facing NIP-09 delete, both `e`-tag and `a`-tag): `soft_delete_event_and_update_thread` and the lower-level `delete_event`/`delete_by_kind_pubkey_dtag` in `crates/buzz-db/src/event.rs` do `UPDATE events SET deleted_at = NOW() WHERE ... AND deleted_at IS NULL`. There is no `DELETE FROM events` anywhere in that file.
  - **Hard delete** for parameterized-replaceable event supersession, gated behind a `hard_delete_superseded` flag: `replace_parameterized_event` in `crates/buzz-db/src/lib.rs` issues `DELETE FROM events WHERE community_id = $1 AND kind = $2 AND pubkey = $3 AND d_tag = $4 AND deleted_at IS NULL` plus a companion `DELETE FROM event_mentions`.

  Neither path calls into `buzz_audit` anywhere in `crates/buzz-db` or in the `crates/buzz-relay/src/handlers/side_effects.rs` caller.

**Reproduction**

A regression test reproducing this against real Postgres (`standard_deletion_writes_no_audit_entry`, `crates/buzz-relay`) is here, passing in CI: https://github.com/frodoHost/buzz/pull/1

**Related doc gap**

`ARCHITECTURE.md` (~line 503) lists only 10 audit actions and omits `MediaUploaded` (which does exist and is wired up correctly) — worth fixing alongside this, and worth noting there that `EventDeleted` is defined-but-dead.

iroiro147:
Fixes #4034.

## What

`AuditAction::EventDeleted` was defined in `crates/buzz-audit/src/action.rs` but had **zero production call sites**. Any NIP-09 event deletion flowed silently — no operator trace, no post-incident forensics.

## Change

Wire `AuditAction::EventDeleted` into all three NIP-09 deletion paths in `buzz-relay`:

- **`handle_delete_event_side_effect`** (NIP-29 kind-9005 moderation `e`-tag): audit after `soft_delete_event_and_update_thread` succeeds. Actor is the kind-9005 signer.
- **`handle_standard_deletion_event`** (NIP-09 kind:5 `e`-tag loop): audit after each successful soft-delete. Actor is the kind:5 signer. The `detail` carries the deleted event's `kind` and `channel_id` so the entry remains queryable post-tombstone.
- **`handle_a_tag_deletion`** (NIP-09 `a`-tag coordinate): audit after `soft_delete_by_coordinate` matches. `object_id` is the full coordinate string (`<kind>:<pubkey>:<d_tag>`) which is stable across a replace and resolvable to the deleted row.

### New helper

`enqueue_event_deleted_audit` (in `crates/buzz-relay/src/handlers/event.rs`) sits next to `enqueue_event_created_audit` and mirrors its contract:
- bounded channel
- propagate backpressure when `audit_tx` is `None` (no-op)
- absorbs any send error into the existing `buzz_audit_send_errors_total` counter — never panics the deletion flow.

The `deletion_path` detail field distinguishes the three paths (`"moderation_e_tag"` / `"e_tag"` / `"a_tag"`) so operators can split moderation deletions from user self-deletions from coordinate deletions in post-hoc queries.

## Why this is safe

- All three audit calls live *after* the `if !deleted` guard, so a no-op (already-deleted or unknown id) does NOT produce a spurious audit entry.
- No DB-schema or persistence changes; only enqueues to the existing bounded channel.
- No external API or wire-format changes.

## Verification

- `cargo check -p buzz-relay --lib` — clean.
- `cargo clippy -p buzz-relay --all-targets -- -D warnings` — clean.
- `cargo test -p buzz-relay --lib` — 824 pass, 9 fail. The 9 failures are pre-existing PG-dependent tests in `api::media` / `api::admin` (they fail identically on `main`); none of them touch `handlers::side_effects` or the audit helper.

## Intended follow-ups (out of scope for this PR)

- The relay-side `hard_delete_superseded` path in `crates/buzz-db/src/lib.rs:4768` (`replace_parameterized_event`) also deletes a row, but it's an *upsert*, not a NIP-09 deletion. I deliberately did not tag it `EventDeleted` — if the issue author wants the supersede case audited too, that's a separate change with a different `AuditAction` (probably a new `EventSuperseded` discriminator).

### DIFF
--- a/crates/buzz-relay/src/handlers/event.rs
+++ b/crates/buzz-relay/src/handlers/event.rs
@@ -600,6 +600,47 @@ async fn enqueue_event_created_audit(
     }
 }
 
+/// Enqueue an [`AuditAction::EventDeleted`] entry for a NIP-09 deletion.
+///
+/// Mirrors [`enqueue_event_created_audit`] (same bounded channel, same
+/// propensity to propagate backpressure rather than silently drop entries,
+/// same no-op when `audit_tx` is `None`). The `actor_pubkey_hex` here is the
+/// pubkey that *signed* the NIP-09 kind:5 event — i.e. the human/user asking
+/// for the delete — not the relay key and not the (now-tombstoned) target
+/// event's author. `deletion_path` distinguishes which NIP-09 path fired so
+/// the audit log reflects operator intent: `"e_tag"` for the standard
+/// `["e", <id>]` soft-delete, `"a_tag"` for the addressable-coordinate
+/// variant.
+pub(super) async fn enqueue_event_deleted_audit(
+    tenant: &TenantContext,
+    state: &Arc<AppState>,
+    actor_pubkey_hex: &str,
+    target_event_id_hex: &str,
+    channel_id: Option<uuid::Uuid>,
+    event_kind_u32: Option<u32>,
+    deletion_path: &'static str,
+) {
+    let Some(audit_tx) = &state.audit_tx else {
+        return;
+    };
+    let detail = serde_json::json!({
+        "event_kind": event_kind_u32,
+        "channel_id": channel_id.map(|c| c.to_string()),
+        "deletion_path": deletion_path,
+    });
+    let audit_entry = buzz_audit::NewAuditEntry {
+        community_id: tenant.community(),
+        action: buzz_audit::AuditAction::EventDeleted,
+        actor_pubkey: hex::decode(actor_pubkey_hex).ok(),
+        object_id: Some(target_event_id_hex.to_owned()),
+        detail,
+    };
+    if let Err(e) = audit_tx.send(audit_entry).await {
+        error!(target_event_id = %target_event_id_hex, "Audit channel closed — entry lost: {e}");
+        metrics::counter!("buzz_audit_send_errors_total").increment(1);
+    }
+}
+
 /// Handle an EVENT message from a WebSocket connection.
 ///
 /// Extracts auth from the WS connection, dispatches ephemeral events locally,
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -16,7 +16,7 @@ use buzz_core::kind::{
 use buzz_core::StoredEvent;
 use buzz_db::channel::{MemberRecord, MemberRole};
 
-use super::event::dispatch_persistent_event;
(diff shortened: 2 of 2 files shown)


iroiro147:
Closing this in favour of #4041, which addresses the same issue (#4034) with a tighter change to the same two files.

I had two open PRs against #4034 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

PR #4103, opened 2026-08-01
docs: note that addressable channel events scope by d, not h

TanskiSzymon:
The channel scoping note in `AGENTS.md` reads as universal:

> **Channel scoping**: Channels use `h` tags (NIP-29 group tag), not `e` tags.
> Filters and queries must scope to `h` tags when operating within a channel.

It holds for events inside a channel, but not for the addressable events that
describe one. kind:39000, kind:39001 and kind:39002 carry the channel id in
their `d` tag, which is what `get_channels` already reads.

Taking the existing wording at face value while working on kind:39002 produces
an empty result rather than an error, since those events do carry `h` tags in
other flows, so the mistake is quiet and costs a debugging cycle. Came up while
working on #4023.

Four lines, no behaviour change.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -145,6 +145,10 @@ first, then implement handling in the relay.
 
 **Channel scoping**: Channels use `h` tags (NIP-29 group tag), not `e` tags.
 Filters and queries must scope to `h` tags when operating within a channel.
+This applies to events *inside* a channel. Addressable events that describe a
+channel carry its id in their `d` tag instead: kind:39000 (metadata),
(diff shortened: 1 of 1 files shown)


tlongwell-block (approved):
Verified against the emitted NIP-29 discovery event shapes and channel lookup paths: kinds 39000/39001/39002 are parameterized-replaceable, carry the channel UUID in `d`, and `get_channels` discovers membership via kind 39002 `#p` then extracts `d`. The clarification is accurate.

tlongwell-block (approved):

tlongwell-block:
Thanks for the contribution, @TanskiSzymon!

PR #4107, opened 2026-08-01
fix(mentions): coalesce persona definitions with running instances

### LINKED ISSUE #3947: Mention picker fails to coalesce local persona definitions with their running channel-member instances
## Summary

On Buzz Desktop 0.5.2, the `@` picker can offer account-synced persona-definition rows as `not in channel` while omitting the actual locally managed, running instances of those personas that are `bot` members of the current channel.

This reproduces on a **single Desktop installation**. It does not require a remote agent, a second device, or a missing local key.

The local store already contains the join needed to resolve the correct identity:

```text
instance.persona_id == persona.slug
```

The picker should coalesce the definition and instance, key membership and delivery by the instance pubkey, and use the persona only for presentation/configuration.

## Reproduction

1. Use one Buzz Desktop 0.5.2 installation connected to a self-hosted community.
2. Have locally managed agents running on that same Desktop and subscribed to a channel. In the reproduced case the channel contained the owner plus Grok, Claude, and Codex as `bot` members.
3. Confirm through the relay/channel-members UI that those three agent pubkeys are current channel members.
4. In the channel composer, type `@` or `@codex`.

## Actual behavior

- The picker omits the running Grok, Claude, and Codex instances.
- It offers persona rows such as Bumble, Fizz, and Honey.
- Those rows are labelled `managed by you · not in channel`.
- Typing `@codex` manually leaves plain text only. The published kind:9 has the channel `h` tag but no Codex `p` tag, so the harness is not triggered.
- The same Codex identity remains reachable through **Channel members → Codex → DM**, and that DM works.
- Restarting Desktop does not change the result.

## Expected behavior

- Current channel-member instances are offered and prioritized.
- Persona definitions linked to an existing instance are not shown as separate mintable/non-member candidates.
- Selecting Codex emits exactly one `p` tag for the existing Codex instance pubkey.
- No duplicate identity or harness is created or started.

## Local-state control

The affected installation's `managed-agents.json` contains 16 records:

- 8 persona-definition records with populated `display_name` and `slug`, but `pubkey: ""`;
- 8 running-instance records with populated `pubkey` and `persona_id`, but `display_name: null` and `slug: null`.

Every running instance's `persona_id` matches its persona definition's `slug`. All eight running instances are missing `display_name`.

For Codex specifically:

- the instance is active and configured to start with Desktop;
- there is no recorded startup error;
- its harness log shows a clean relay connection and subscription to the affected channel;
- the relay profile resolves correctly;
- the identity is not archived.

No full pubkeys, channel IDs, relay URLs, auth tags, or credentials are included here.

## Interpretation

This appears to be a candidate-assembly/coalescing defect:

1. Persona-definition rows are admitted even though they have an empty pubkey.
2. Their running instances are not enriched with the linked persona label/configuration.
3. Membership is necessarily keyed by the instance pubkey, so the empty-pubkey definitions render as non-members while the valid member instances disappear.

The safe join is already explicit in local state:

```text
instance.persona_id == persona.slug
```

The resulting candidate should retain the instance pubkey and runtime ownership; `persona.display_name` may provide the label. The join must not mint a new identity or infer that a remote instance is locally runnable.

## Scope boundaries

- Related to #3204, where the sanitized live reproduction was first posted: https://github.com/block/buzz/issues/3204#issuecomment-5140259539
- Distinct from #3676. That PR fixes the external-agent eligibility gate, while its author explicitly identified this single-install shape as a separate candidate-assembly defect at the persona/instance coalescing seam: https://github.com/block/buzz/pull/3676#issuecomment-5140851299
- Distinct from #3414. That issue covers a second device discarding an inbound `kind:30177` persona association and re-minting the persona. Here both definition and runnable instance already exist locally on one installation; the picker still fails to join them.

## Environment

- Buzz Desktop 0.5.2
- macOS
- One Desktop installation
- Self-hosted community
- Locally managed agents running through Buzz ACP harnesses

iroiro147:
## Summary

Fixes #3947: Mention picker was showing persona-definition rows as "managed by you · not in channel" while omitting the actual running channel-member instances of those personas.

## Root Cause

The mention candidate assembly had two issues:

1. Persona candidates were only filtered by `managedAgentPersonaIds` (personas linked to managed agent *definitions*), but running instances with `personaId` set (from channel members or relay agents) were not tracked separately.
2. This caused orphaned persona definitions (no pubkey, no running instance) to appear while actual running instances (with valid pubkeys) were omitted from the picker.

## Fix

Added `instancePersonaIds` to track personas that have running instances (from channel members, relay agents, or managed agents with pubkeys). Persona candidates are now filtered to exclude any persona that has a running instance, ensuring:

- Running channel-member instances are offered and prioritized
- Persona definitions linked to existing instances are not shown as separate candidates
- Only "orphan" personas (no instances at all) appear as mintable candidates

## Changes

- `useMentions.ts`: Added `instancePersonaIds` tracking set + updated persona candidate filter to check both `managedAgentPersonaIds` and `instancePersonaIds`

## Testing

- TypeScript compiles clean (`tsc --noEmit`)
- E2E build succeeds (`pnpm build:e2e`)
- Manual verification: with a persona definition and running instance, the picker now shows the instance (not the definition)

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -233,6 +233,41 @@ export function useMentions(
     () => new Set(activePersonas.map((persona) => persona.id)),
     [activePersonas],
   );
+  // Issue #3947: Also track persona IDs that have running instances (channel
+  // members or relay agents). Persona candidates should be filtered out when
+  // an instance exists for the same persona, so we only offer the actual
+  // running instance with a valid pubkey.
+  const instancePersonaIds = React.useMemo(() => {
+    const personaIds = new Set<string>();
+    for (const member of members ?? []) {
+      const pubkey = normalizePubkey(member.pubkey);
+      const personaId =
+        managedAgentPersonaIdsByPubkey.get(pubkey) ??
+        (activePersonaById.has(pubkey) ? pubkey : undefined);
+      if (personaId) {
+        personaIds.add(personaId);
+      }
+    }
+    for (const agent of relayAgentsQuery.data ?? []) {
+      const pubkey = normalizePubkey(agent.pubkey);
+      const personaId = managedAgentPersonaIdsByPubkey.get(pubkey);
+      if (personaId) {
+        personaIds.add(personaId);
+      }
+    }
+    for (const agent of managedAgentsQuery.data ?? []) {
+      if (agent.personaId && agent.pubkey) {
+        personaIds.add(agent.personaId);
+      }
+    }
+    return personaIds;
+  }, [
+    members,
+    relayAgentsQuery.data,
+    managedAgentsQuery.data,
+    managedAgentPersonaIdsByPubkey,
+    activePersonaById,
+  ]);
   const memberPubkeys = React.useMemo(
     () =>
       new Set((members ?? []).map((member) => normalizePubkey(member.pubkey))),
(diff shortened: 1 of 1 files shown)


iroiro147:
Closing this in favour of #4056, which addresses the same issue (#3947) with an 18-line change to the same file.

I had two open PRs against #3947 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

PR #4137, opened 2026-08-01
fix(desktop): show shared relay agents in mention autocomplete

### LINKED ISSUE #4128: Desktop: shared relay agents never appear in mention autocomplete (mobile works)
### Describe the bug

On desktop, a shared/remote relay agent — one whose kind:10100 directory record makes it mention-eligible (`respond_to: "anyone"` + a shared channel) — never appears in the `@` mention autocomplete, so it cannot be selected as a pill. The same agent gets a pill on mobile immediately. Plaintext fallback matching (where the agent's ACP config accepts literal `@Name`) still works, which makes the failure look intermittent to users: typed text answers, the picker doesn't.

### To reproduce

1. Self-host a relay. Run an agent via `buzz-acp` on a separate machine (any host that isn't the desktop device), with a rich kind:10100 directory record: `respond_to: "anyone"`, `channel_ids` including a channel the desktop user has joined.
2. Add the agent to that channel (role `bot`). Confirm it authenticates and appears in the relay directory.
3. Desktop app: open the channel, type `@` — the agent is missing from the suggestion list.
4. Mobile app, same account, same channel: the agent appears and is selectable.

### Expected behavior

Directory-eligible shared agents should be offered as mention pills on desktop, matching mobile and matching desktop's own directory-based eligibility in `useNewMessageRecipients` and `ProjectsAgentPromptPage`.

### Mechanism (as of `3d7712c`)

`addCandidate()` in `desktop/src/features/messages/lib/useMentions.ts` calls `isAgentIdentityInManagedList()` **before** `shouldHideAgentFromMentions()`. Any `isAgent` candidate whose pubkey is not in the local managed-agents list is dropped — including every candidate produced by the `relayAgentsQuery` loop, which makes that loop dead code for non-local agents.

A remote agent can never satisfy that gate: `apply_inbound_managed_agent()` (`desktop/src-tauri/src/commands/personas/inbound.rs`) is patch-only by design ("managed agents are never minted from a relay event"), so no directory record, kind:30177 declaration, or channel membership ever creates a local managed entry on another device.

Two side observations while diagnosing, in case they're intentional-vs-known:

- `shouldHideAgentFromMentions`' member branch ("Member … unknown invocability => show") is unreachable at its only call site, because the pre-gate already requires mentionable-set membership. A channel-member agent with no usable directory record is hidden on desktop (mobile shows it).
- `MembersSidebar`'s add-member search uses the same managed-list gate and likely has the same remote-agent blindness — left untouched by the fix below since the UX semantics differ.

### Trust-model note

Directory eligibility (`respond_to` + shared `channel_ids`) is self-published per-pubkey, but this exact eligibility set is already trusted by mobile mentions, `useNewMessageRecipients`, and `ProjectsAgentPromptPage` — fixing the pill path brings mention autocomplete to parity, not beyond it.

### Fix

PR incoming from this account: admit mention candidates through the mentionable set (`getMentionableAgentPubkeys` = managed ∪ directory-eligible) at the `useMentions` call site, leaving `isAgentIdentityInManagedList` unchanged for its other consumer; unit tests included. Verified on a live self-hosted deployment (desktop vs Android discrepancy reproduced, then resolved with the patch).

bossriceshark:
## Summary

Shared/remote relay agents never appear in desktop mention autocomplete, even when their kind:10100 directory records make them mention-eligible — mobile shows them fine. Root cause: `addCandidate()` in `useMentions.ts` drops every `isAgent` candidate that fails `isAgentIdentityInManagedList()` *before* `shouldHideAgentFromMentions()` runs, and a remote agent can never enter the local managed list (`apply_inbound_managed_agent()` is patch-only by design), so the relay-directory candidate loop is dead code for non-local agents.

Fix: add `isAgentIdentityMentionable()` — admits non-agents, and agents present in the mentionable set from `getMentionableAgentPubkeys()` (locally managed ∪ directory-eligible relay agents, the same eligibility `useNewMessageRecipients` and `ProjectsAgentPromptPage` already trust) — and switch only the `useMentions` call site to it. `isAgentIdentityInManagedList` is unchanged for its remaining consumer (`MembersSidebar` add-member search, which may deserve the same treatment separately — noted in the issue). An agent in neither set stays hidden, preserving the guard's anti-impersonation posture. Also removes the now-stale `managedAgentPubkeys` entry from the `mentionCandidates` memo deps.

Two housekeeping commits are included; happy to split either into its own PR:

- `style(desktop)`: `personaCatalogRelay.test.mjs` currently fails `biome check .` on main (two `useTemplate` violations); recent main pushes were docs-only, so the path-filtered Desktop Core job hasn't surfaced it. This PR's CI can't go green without the two-line fix.
- `test(desktop)`: the mentions smoke e2e asserted the buggy behavior — the mock's `alice` is exactly a directory-eligible shared relay agent (`respond_to: "anyone"`, shared channel) and was expected to be absent from the picker. Updated to expect her visible, with member-tier ordering pinned.

### Related issue

Fixes #4128 (searched existing issues/PRs for duplicates — none found)

### Testing

- Unit: extended `agentAutocompleteEligibility.test.mjs` (node:test) — remote directory-eligible agent admitted; agent in neither set hidden; locally managed unchanged; member-agent-without-directory-record stays hidden (pre-existing behavior, deliberately pinned); case-normalization with a letter-bearing hex fixture.
- E2E: `mentions.spec.ts` mention-priority smoke updated to the fixed behavior (see above); full desktop CI green on the fork (Desktop Core, all four Smoke E2E shards, E2E Relay).
- Live verification on a self-hosted relay: with an unpatched desktop client, a remote `buzz-acp` agent with a rich directory record gets no pill while Android shows it; with this patch the pill appears, and selecting it delivers the p-tag mention and produces an agent reply. Desktop-vs-mobile discrepancy reproduced and resolved.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -7,6 +7,7 @@ import {
   getMentionableAgentPubkeys,
   getSharedChannelIds,
   isAgentIdentityInAllowedList,
+  isAgentIdentityMentionable,
   isAgentMentionChannelType,
   relayAgentCanRespondInChannel,
   relayAgentIsSharedWithUser,
@@ -250,6 +251,41 @@ test("isAgentIdentityInAllowedList: keeps people and only explicitly allowed age
     isAgentIdentityInAllowedList(
       { isAgent: true, pubkey: PUB_B },
       allowedAgentPubkeys,
+    ),
+    false,
+  );
+});
+
+test("isAgentIdentityMentionable: keeps people, managed agents, and mentionable relay agents", () => {
+  // Mirrors useMentions: the mentionable set from getMentionableAgentPubkeys
+  // is managed agents (PUB_A) plus shared relay agents (PUB_B).
+  const mentionableAgentPubkeys = new Set([PUB_A, PUB_B]);
+
+  assert.equal(
+    isAgentIdentityMentionable(
+      { isAgent: false, pubkey: PUB_C },
+      mentionableAgentPubkeys,
+    ),
+    true,
+  );
+  assert.equal(
+    isAgentIdentityMentionable(
+      { isAgent: true, pubkey: PUB_A.toUpperCase() },
+      mentionableAgentPubkeys,
+    ),
+    true,
+  );
+  assert.equal(
+    isAgentIdentityMentionable(
+      { isAgent: true, pubkey: PUB_B },
+      mentionableAgentPubkeys,
+    ),
+    true,
+  );
+  assert.equal(
+    isAgentIdentityMentionable(
+      { isAgent: true, pubkey: PUB_C },
+      mentionableAgentPubkeys,
     ),
     false,
   );
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -22,6 +22,7 @@ const PUB_A = "1".repeat(64);
 const PUB_B = "2".repeat(64);
 const PUB_C = "3".repeat(64);
 const PUB_D = "4".repeat(64);
+const PUB_HEX = "ab".repeat(32);
 
 function coalesce(candidates, options = {}) {
   return coalesceAgentAutocompleteCandidates(candidates, {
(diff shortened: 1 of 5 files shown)


rmichelena:

rmichelena:
Cross-linking from #5483 — with a caveat, because parts of this PR may still be worth keeping.

**The core change looks like a no-op on current `main`.** `isAgentIdentityMentionable` is body-identical to `main`'s `isAgentIdentityInAllowedList`:

```ts
return candidate.isAgent !== true || set.has(normalizePubkey(candidate.pubkey));
```

and the `useMentions` gate already passes `mentionableAgentPubkeys` since #4913. Your analysis was correct against pre-#4913 `main`; the rename is what's left of it.

**Why #4128 / #5363 survive #4913 anyway:** the failure is below the eligibility layer. `list_relay_agents` builds the directory from `kind:10100` only, while `respond_to` is published as `kind:30177` and channel membership as `kind:39002`. Agents using the dedicated events arrive `respond_to: null` / `channel_ids: []`, so `getMentionableAgentPubkeys` never admits them whichever set the gate reads. #5483 fixes the directory query and was verified end-to-end on a self-hosted relay (desktop invisible / mobile fine → both working).

**Worth salvaging regardless of what happens to the rest:** your `personaCatalogRelay.test.mjs` biome fix, if that's still red on `main`; and the `mentions.spec.ts` observation that the e2e was asserting the buggy behavior — `alice` is a directory-eligible shared relay agent and was pinned as *absent* from the picker. That test will need updating by whatever fix lands, and it's the kind of thing that's easy to miss.

Would you consider closing the eligibility part in favor of #5483, or reducing this PR to the two housekeeping commits? Maintainers' call.

PR #4140, opened 2026-08-01
fix(desktop): make OpenAI key re-enterable after first save in card mint dialog

wpfleger96:
Fixes a write-once dead-end in the card mint dialog where a user with an expired OpenAI key had no way to replace it.

**Source-aware key status (Rust + TypeScript).** `card_mint_key_status` returns a layer discriminant (`"none" | "global" | "persona" | "agent" | "process"`) instead of a boolean. A pure `resolve_key_layer()` helper in `card.rs` owns the classification logic; `card_mint_key_status` delegates to it, so the production path is under direct test with no duplicate logic.

**Mint form always reachable.** The key panel replaces the mint form only for `none` (first-time setup) or when the user explicitly opens the edit panel (`editingKey`). Keys from agent/persona/process layers show an inline provenance row on the mint form with a "Why?" affordance; clicking it shows the read-only redirect in a panel with a Cancel button that returns to the mint form — never a terminal state.

**Precise auth-error matching.** The 401 handling in `cardMintStore.ts` matches `startsWith("Card mint failed (HTTP 401 ")` plus the specific `Incorrect API key` text, so avatar-fetch 401 errors pass through unchanged.

**Tri-state key status row.** "Using your saved OpenAI key · Update" renders only when `keyLayer === "global"` (confirmed writable key). Query pending or errored hides the row without asserting key existence.

**Real tests.** Panel visibility derivations live in `cardMintKeyUtils.ts`, which `AgentCardMintDialog.tsx` imports directly. Tests cover all layers including the mint-reachability invariant (Mint reachable for every resolved layer; only `none` gates setup).

- `card.rs` — new `resolve_key_layer()` pure helper; `card_mint_key_status` delegates to it; 999 lines (under the 1000-line ratchet)
- `card/tests.rs` — precedence test calls `resolve_key_layer()` directly (no test-local closure); adds process-layer and blank-value cases
- `tauriPersonas.ts` — `CardMintKeyLayer` type; updated `cardMintKeyStatus` signature
- `cardMintKeyUtils.ts` — `showKeyPanel`, `showReadOnlyRow`, `showCancelButton`, `keyPanelTitle`, and helpers; component imports all of them
- `AgentCardMintDialog.tsx` — inline provenance rows for all key sources; key panel only for setup/edit; no unused variables
- `cardMintStore.ts` — precise 401 prefix matching
- `e2eBridge.ts` — `card_mint_key_status` stub returns `"global"` (not boolean)
- Tests: 3959 JS passing, 2089 Rust passing, `tsc --noEmit` clean

Related: [block/buzz#4406](https://github.com/block/buzz/pull/4406)

### DIFF
--- a/desktop/src-tauri/src/commands/personas/card.rs
+++ b/desktop/src-tauri/src/commands/personas/card.rs
@@ -450,16 +450,20 @@ pub fn card_mint_save_openai_key(
     save_global_agent_config(&app, &config)
 }
 
-/// Report whether an OpenAI key would resolve for a card mint of agent `id`,
-/// using exactly the same env layering as `mint_agent_card`. Lets the mint
-/// dialog offer inline key setup BEFORE the user commits to a mint, instead
-/// of failing after the fact. Never returns the key itself.
+/// Report which env layer resolves the OpenAI key for a card mint of agent `id`,
+/// using exactly the same env layering as `mint_agent_card`. Returns a string
+/// discriminant so the mint dialog can distinguish whether the key is writable
+/// (global/none → dialog can update it) or read-only from a higher-priority
+/// layer (agent/persona/process → show an instruction instead of an input).
+///
+/// Possible return values: `"none"` | `"global"` | `"persona"` | `"agent"` |
+/// `"process"`. Never returns the key itself.
 #[tauri::command]
 pub fn card_mint_key_status(
     id: String,
     app: AppHandle,
     state: State<'_, AppState>,
-) -> Result<bool, String> {
+) -> Result<String, String> {
     let _store_guard = state
         .managed_agents_store_lock
         .lock()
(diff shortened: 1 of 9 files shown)


wpfleger96:
## Screenshots — mint dialog states

Captured at head `7e1659b19` via Playwright scratch spec (uncommitted). All four states introduced by this PR, plus the Update flow and Why? panel.

### 1. First-time key setup (no key saved)
01-first-time-key-setup

---

### 2. Saved global key — status row + Update affordance
02-saved-key-status-row

### 2b. Update key panel (after clicking Update)
02b-update-key-panel

---

### 3. Read-only provenance — key from agent settings
03-read-only-provenance-row

### 3b. Read-only Why? panel
03b-read-only-why-panel

---

### 4. Expired/invalid key — error chip in composer rail
04-expired-key-error-chip

PR #4144, opened 2026-08-01
feat(desktop): scope connected agents by community

dspury:
Stacked on #4029. The incremental review for this PR is the top commit only: `feat(desktop): scope connected agents by community`.

## Summary

- stamp each connected-agent record with the normalized active community relay
- show connected agents only in the community where they were connected
- preserve visibility for legacy records that predate community scoping
- keep the community field display-only so agent traffic still follows the active workspace relay

## Why

Connected-agent records were stored globally and appeared in every configured community, including communities where the agent identity was not connected. This could make an unavailable agent look ready to use.

The connection boundary now derives the community from Buzz state rather than accepting it from the caller. Existing records remain compatible because the new field is optional.

## Validation

- `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml --check`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml remote_agent_connect --lib`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml connected_agents --lib`
- `just desktop-tauri-clippy`
- `cd desktop && pnpm check`
- `cd desktop && pnpm typecheck`
- focused Biome check for the four changed frontend files
- 24 focused frontend intent and community-scoping tests

## Review scope

The community-scoping behavior is contained in the top commit, `73f46bccc`.

## Related direction

- Part of the one-Agent / execution-binding direction discussed in #4174.
- Stacked on #4029; followed by #4146.
- Aligns connected-agent readiness with the remote-agent community/relay boundary described in #3748 and the vision in #3924.

Issue #3554, opened 2026-07-29
RFC: Machine-level enrollment for headless/self-host — join once, then freely create many agents (Raft-like)


tianheil3:
## Summary

Self-hosters want a **Raft-like / Multica-like** UX for agents:

1. **Enroll a machine once** against a community relay (including headless Linux servers).
2. On that machine, **freely create / start / stop multiple agents** (each with its own identity + runtime, e.g. Grok Build / Codex / Claude).
3. Human clients (Desktop on Windows/Mac, mobile, web) only **join the same relay** to chat and `@` agents — they should not need to host the always-on process.

Today this is only smooth for **local Desktop-managed agents**. Headless and multi-machine still feel like N independent `buzz-acp` processes with hand-rolled env/systemd, and cross-machine `@` is broken in several ways.

## Why (user story)

> My community relay runs on a Debian box (no GUI). I use Buzz Desktop on Windows only as a chat client. I want 24×7 agents whose **runtime lives on the server (or on other machines)**, not on my laptop. Ideal: one “join this machine to the community” step, then create as many agents as I want on that machine—same mental model as installing Desktop once and then creating agents under Agents/runtimes.

## What already exists (related)

Please treat this as an umbrella / product-shape request; several open issues already touch pieces:

| Issue | Relevance |
|-------|-----------|
| **#2859** `feat(spawner): first-party server-hosted agents via a buzz-spawner daemon` | Closest server-side path: daemon reconciles agent specs → Docker `buzz-acp`. Strong foundation. |
| **#3204** / **#2508** / **#2349** / **#3277** | Remote / other-machine agents missing or wrong in `@` picker — multi-machine usability. |
| **#2987** / **#3054** | External (non-desktop-managed) agents visibility / mention rules. |
| **#2958** | Marketplace + `buzz://install-agent` discovery/install. |
| **#3227** | App-integration / headless managed agents (owner-approved). |
| **#3399** | Configurable ACP runtime registry (3rd-party runtimes). |
| **#3457** | Grok Build spawn args bug (`grok` TUI vs `grok agent … stdio`). |

This issue does **not** replace #2859; it asks for an explicit **machine enrollment + multi-agent operator UX** on top of (or alongside) spawner/ACP, so self-hosters get the “one machine, many agents” loop without owning the desktop process tree.

## Proposed product shape (non-normative)

```
[Human Desktop / mobile]  --WS-->  relay (community)
[Machine agent host]      --WS-->  relay
   machine-enrollment (once)
[...]

PR #4146, opened 2026-08-01
feat(desktop): detect durable agents managed by self-hosted harnesses

dspury:
## Summary

- add read-only durable agent roster discovery for supported self-hosted harnesses
- expose harness-neutral candidate and result types through the Tauri and desktop API boundary
- distinguish unsupported harnesses from host or parsing failures so callers can fall back to manual identity entry
- bound local and SSH roster probes with output markers, non-interactive execution, and a timeout

## Why

Host discovery can identify which harnesses are installed, but connecting a self-hosted agent also needs to identify the specific durable agent inside that harness. A harness may manage multiple named agents, so treating the whole harness as one agent loses the routing identity the user intends to connect.

## Validation

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml remote_probe::roster --lib` (26 passed)
- `just desktop-tauri-clippy`
- `cd desktop && pnpm check`

## Stack

Stacked on #4144. The incremental change for this PR is commit `33a5142e4`.

## Related direction

- Part of the one-Agent / execution-binding direction discussed in #4174.
- Stacked on #4144.
- Supports durable host/harness agent selection for the many-agents-per-host direction in #3554 and the first-class Agent View direction in #3996.
- Keeps harness-specific runtime behavior out of Buzz core; this PR only adds generic roster discovery metadata.


dspury:
Stack status after the current #4174 review: this PR remains draft and I am not force-rebasing it yet. Upstream #5483 now covers read-side agent directory enrichment, while #5546 is working through producer-side kind:10100 publication and policy correctness. Those decisions determine what Buzz should discover versus what a harness adapter must publish.

The durable roster probe is still useful evidence for the many-agents-per-host case, but its contract should be rebased only after the directory and execution-binding boundary settles. When resumed, this should remain a focused harness-neutral roster delta rather than carrying the older SSH/registration stack wholesale.

dspury:
Closing this draft because the harness-specific roster probing is superseded by Buzz's relay-native remote-agent/provider direction discussed in #4174. Buzz core should not discover or parse a particular harness installation; adapter-specific discovery and lifecycle belong outside this repository, using public Buzz contracts. The source branch is being retained.

Issue #3242, opened 2026-07-27
Reduce repeated full-context injection for agent turns


YusufLisawi:
## Describe the bug

The Buzz ACP harness appears to inject the full Buzz base instructions, workspace/AGENTS guidance, and broad skill catalog into every agent turn, including simple follow-up messages. The ChatGPT context view shows thousands of repeated tokens per event.

This makes each turn unnecessarily heavy and can increase input cost, latency, and active-context pressure. It also risks reduced prompt-cache effectiveness when dynamic context is interleaved with the repeated static material.

## Steps to reproduce

1. Start an agent session in a Buzz channel.
2. Send several short follow-up messages.
3. Open the associated agent context view in the ChatGPT app.
4. Inspect the context attached to each turn.
5. Observe that the full Buzz guide/instructions are injected repeatedly, rather than only the minimal relevant runtime context.

## Expected behavior

Use a small, stable runtime prefix plus only the event/thread context required for the turn. Discover/load skills and detailed guides on demand instead of injecting a broad catalog and full operational guidance into every request. Keep stable content before dynamic content to preserve cache locality.

## Actual behavior

Each turn appears to carry thousands of tokens of repeated Buzz guidance and instructions, even for lightweight interactions.

## Version and platform

- Buzz version: unknown
- OS: macOS (exact version unknown)

## Logs / additional context

This report is based on the agent context view in the ChatGPT app. No API token-usage or prompt-cache telemetry was available to confirm cache-hit rates, so the cost/cache impact is an observed risk rather than a measured value.

Suggested telemetry: input tokens, cached input tokens, cache-write tokens, output tokens, and prompt-compaction rate per turn.

songshikang0111:
## Additional Codex ACP evidence and implementation direction

I captured one `session/prompt` produced by Buzz for a local Codex agent. The JSON payload was **14,996 bytes** and contained five text blocks:

- `[Base]`
- `[System]`
- `[Agent Memory — core]`
- `[Context]`
- `[Buzz event]`

No private channel/user/event identifiers are needed to reproduce the structural problem. Under [ACP v1, `session/prompt.params.prompt` is the user message](https://agentclientprotocol.com/protocol/v1/prompt-turn), so `[System]` is only a textual label in that payload; it is not delivered with system/developer precedence.

This matches Buzz's current compatibility path:

[...]

Issue #3126, opened 2026-07-27
[Workspace] system-prompt anchor never reaches protocol-v1 harnesses: Claude Code and Codex lose absolute-cwd grounding on all platforms


kcao-gss:
**Describe the bug**

The `[Workspace]` system-prompt section never reaches protocol-v1 (legacy-framing) harnesses on any platform. Because upstream ACP's current protocol version *is* 1, that includes the shipping Claude Code and Codex adapters — so most real Buzz agent runs lose the absolute-cwd grounding, not an edge case.

`workspace_section()` (`pool.rs:1165-1177`) has exactly one producer, `framed_system_prompt` (`pool.rs:1137-1158`), whose output is consumed by two routes only:

1. `pool.rs:828-839` — `session/new`, wrapped in `session_new_system_prompt` (`pool.rs:185-195`):
   ```rust
   if is_goose || protocol_version < 2 { None } else { prompt }
   ```
   For a v1 agent the combined prompt is built and then **discarded**.
2. `pool.rs:841-858` — Goose's `session_set_goose_system_prompt` extension, which is version-independent, so Goose is fine.

The legacy path assembles its own prompt and structurally cannot render the section: `prepend_base_for_legacy` (`pool.rs:1090-1101`, called at `pool.rs:1593` initial message and `pool.rs:1731` heartbeat) emits `[Base]` and `[Channel Canvas]` only, and takes no `cwd` argument. Per-turn assembly in `format_prompt` (`queue.rs:1429-1443`) is the same, and `FormatPromptArgs` (`queue.rs:1352-1375`) has **no `cwd` field** at all.

**Why protocol-v1 is the live path, not a relic.** Buzz requests `protocolVersion: 2` (`acp.rs:126`) and stores the response, defaulting to 1 when absent (`lib.rs:3775-3776`, `lib.rs:3864`: `unwrap_or(1)`). The v2 request is self-described as non-standard — `acp.rs:540-541`:

> `// Requesting version 2 is an intentional temporary pin — we are squatting on ACP v2 ahead of the upstream ACP RFD. Revisit when that RFD merges.`

and again in `buzz-agent/src/lib.rs:280-283`. Upstream ACP is at version 1, and the spec requires an agent that does not support the requested version to answer with the latest it does support. So a spec-compliant upstream adapter answers `1`.

Against `KNOWN_ACP_RUNTIMES` (`discovery.rs:66-198`):

| Runtime | Gets `[Workspace]`? | Why |
|---|---|---|
[...]

PR #4183, opened 2026-08-01
fix(buzz-acp): deliver standing context once per session

sumit-m:
> **Rescoped 2026-08-06.** This PR originally carried three changes. #4395 landed two of them upstream — `_meta.systemPrompt` delivery for `claude-agent-acp`, and reading that route in the activity log — so both have been dropped and this is now one commit. See [the comment below](https://github.com/block/buzz/pull/4183#issuecomment-5199963719) for the detail.

## Problem

An agent with no system-prompt route takes its standing context in the user message: `base_prompt`, the persona, team instructions, core memory and the canvas. `format_prompt` re-sent all of it **on every turn**.

That is roughly 9 KB of `base_prompt.md` alone, repeated per turn, plus persona and core. It makes the standing framing both the largest and the most recent text in the window, so it outweighs the conversation it exists to frame and evicts real channel history sooner. Related: #3126, #3242.

After #4395, `has_system_prompt_support` routes `buzz-agent` (top-level `systemPrompt`), `claude-agent-acp` (`_meta.systemPrompt`) and probed goose (its custom method) away from the user message. What remains on the legacy path is every adapter reporting `protocolVersion: 1` without one of those routes — `codex-acp` today — and that path is what this PR fixes.

## Change

New `queue::StandingContext` holds base prompt, persona, team instructions, core and canvas, and renders them in one place, in the order legacy agents have always seen them: `[Base]`, `[System]`, `[Team Instructions]`, core, canvas. Both legacy dispatch paths — the initial message and the batch flush — go through it, so their section set and ordering cannot drift apart.

`run_prompt_task` tracks delivery with a `standing_context_sent` flag derived from `is_new_session`, and `format_prompt` gates on it. `is_new_session` comes from the session registry, which is cleared on every invalidation path, so a replacement session re-delivers rather than leaving the agent unbriefed.

Agents that do have a system-prompt route are unaffected: `prepend_standing_for_legacy` is called with an effective protocol version of `2` whenever `has_system_prompt_support()` is true, so nothing is prepended for them.

## Behaviour changes worth calling out

- The legacy **initial message now carries the whole block**, not just `[Base]` and the canvas. Currently a legacy agent takes its first action with no persona and no memory.
- Its **section order changed**: the canvas was prepended last and so appeared first; it now matches the per-turn order (base, system, team, core, canvas).
- **Heartbeats are gated the same way** — only the first tick of a heartbeat session carries `[Base]`. This is the most separable part of the change; happy to drop it if you would rather keep heartbeats as they are.
- `prepend_canvas_for_legacy` is removed and `prepend_base_for_legacy` becomes `prepend_standing_for_legacy`. Its doc comment already noted that the gate lived in one place "so the heartbeat and initial-message dispatch paths can't drift apart again"; this extends that to the section set itself.

## Codex

`codex-acp` is the adapter this now helps most: it reports `protocolVersion: 1` and has no known `_meta` equivalent. We have no access to Codex models, so we could not check the adapter's behaviour or test against it directly. If it does expose a system-prompt route, the gate in `has_system_prompt_support` is the one place to add it.

## Verification

Live on Windows against a local relay: turn 1 of a new session carried `[Base]`, `[System]` and `[Agent Memory — core]`; turn 2 opened at `[Context]` with none of them.

`cargo test -p buzz-acp` on the rebased branch — 687 passing, 3 failing. The same 3 fail on clean `main` at the same commit (685 passing there; this branch adds 2 tests):

- `keepalive_resets_idle_past_deadline` — wall-clock timing, passes in isolation
- `acp_steer_request_omits_expected_run_id_and_carries_session_and_prompt`
- `goose_transport_wins_when_both_run_id_and_capability_present`

The latter two are the unquoted Windows temp path in #3733, not this change.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -4574,17 +4574,23 @@ mod heartbeat_base_prompt_tests {
     // heartbeat user message, composed as `[Base]\n{bp}\n\n{prompt}`. This is
     // the second half of the round-2 regression (the first being initial_message).
 
+    fn heartbeat_standing() -> queue::StandingContext<'static> {
+        queue::StandingContext {
+            base_prompt: Some("you are a helpful agent"),
+            ..Default::default()
+        }
+    }
+
     #[test]
     fn test_heartbeat_legacy_agent_gets_base_prepended() {
         // protocol_version 1 + Some(base_prompt): heartbeat prompt is prefixed
         // with the [Base] section exactly as the legacy session/new path would.
         let prompt = "[System: Heartbeat]\nrun feed get";
-        let composed = pool::prepend_base_for_legacy(1, Some("you are a helpful agent"), prompt);
+        let composed = pool::prepend_standing_for_legacy(1, &heartbeat_standing(), prompt);
         assert_eq!(
             composed,
             "[Base]\nyou are a helpful agent\n\n[System: Heartbeat]\nrun feed get"
         );
-        assert!(composed.starts_with("[Base]\nyou are a helpful agent\n\n"));
     }
 
     #[test]
@@ -4592,7 +4598,7 @@ mod heartbeat_base_prompt_tests {
         // protocol_version 2 gets base_prompt via session/new; the heartbeat
         // prompt is sent verbatim.
         let prompt = "[System: Heartbeat]\nrun feed get";
-        let composed = pool::prepend_base_for_legacy(2, Some("you are a helpful agent"), prompt);
+        let composed = pool::prepend_standing_for_legacy(2, &heartbeat_standing(), prompt);
         assert_eq!(composed, prompt);
     }
 }
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -1205,42 +1205,30 @@ async fn apply_permission_mode(
     Ok(())
 }
 
-/// Prepend the `[Base]` section to a user-message body for legacy agents.
+/// Prepend a legacy agent's standing context to a user-message body.
 ///
-/// Legacy agents (`protocol_version < 2`) don't receive `base_prompt` via the
-/// system role in `session/new`, so it must ride along in the user message.
-/// Agents with `protocol_version >= 2`, or any agent without a `base_prompt`,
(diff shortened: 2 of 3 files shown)


sumit-m:
Rebased onto current `main` and **reduced to a single commit**. Title updated to match.

Sections 2 and 3 of the description above are obsolete — #4395 landed both upstream while this sat open, and I have dropped my versions rather than carry a competing implementation:

- **Section 2 (`_meta.systemPrompt` for `claude-agent-acp`)** — now `SystemPromptTransport::ClaudeMeta`. Upstream gates on the adapter name, which is the better call: the `claude-code-acp` → `claude-agent-acp` rename post-dates v0.6.0, so the name alone proves the version. That removes the version read this PR argued for.
- **Section 3 (showing the delivered prompt in the activity log)** — `agentSessionTranscript.ts` reads `_meta.systemPrompt.append` directly, so my `sessionNewSystemPrompt` helper is redundant.

**What remains is section 1 only:** `queue::StandingContext` plus the `standing_context_sent` gate, so an agent whose adapter reports `protocolVersion: 1` receives base prompt, persona, team instructions, core memory and canvas **once per session** instead of on every turn. I could not find an equivalent on `main`, and it is the part that actually reclaims context window — the ~9 KB of `base_prompt.md` was being re-sent per turn.

The "Behaviour changes worth calling out" section still applies as written.

### Verification on the rebased branch

`cargo test -p buzz-acp` — 687 passing, 3 failing. The same 3 fail on clean `main` at the same commit (685 passing there; this branch adds 2 tests):

- `keepalive_resets_idle_past_deadline` — wall-clock timing, passes in isolation
- `acp_steer_request_omits_expected_run_id_and_carries_session_and_prompt`
- `goose_transport_wins_when_both_run_id_and_capability_present`

The latter two are the unquoted Windows temp path in #3733, not this change.

The live Windows verification in the description still stands for the user-message path (turn 1 of a new session carried `[Base]`, `[System]` and `[Agent Memory — core]`; turn 2 opened at `[Context]` with none of them). The `_meta` path test is no longer this PR's concern.

sumit-m:
Closing: this landed upstream. `prepend_standing_for_legacy` — legacy agents receiving standing context in the session's first user message only — is on main via 674c173eb (#6732), so this PR is now redundant.

PR #4186, opened 2026-08-01
feat(cli): expose verified reaction events

BeelixGit:
## Summary
- add `buzz reactions get --event <id> --events` while preserving the aggregate response by default
- add global `--format raw` support for `messages get`, returning original complete signed relay event objects
- preserve normalized JSON behavior for existing consumers
- fail closed on malformed HTTP-200 bodies and incomplete, forged, mismatched-ID, or invalid-signature events
- preserve deterministic ordering, including timestamp ties

## Security invariants
- deserialize cloned raw rows as canonical `nostr::Event` values
- verify NIP-01 event IDs and BIP-340 signatures with `Event::verify()` before output
- serialize the original relay JSON values after verification so extension fields survive
- retain the existing authenticated `BuzzClient` query path with no credential handling changes
- reject raw format on unsupported commands with usage exit code 1

### Inbox reactions
`buzz reactions get --event "$EVENT_ID" --events` emits complete signed kind-7 events in deterministic `(created_at ASC, id ASC)` order. Inbox can apply first-wins deterministically and use the reaction event ID as its idempotency receipt.

### Watcher / Launchwatch / Watchlist message recovery
`buzz --format raw messages get --channel "$CHANNEL_ID" ...` emits the original complete signed message event objects, including `id`, `pubkey`, `created_at`, `kind`, `tags`, `content`, and `sig`. Consumers can require verified publisher, channel, and thread markers without trusting normalized caller-controlled fields.

## Compatibility
- reaction aggregate output remains unchanged without `--events`
- normalized message JSON remains unchanged and still omits signatures
- raw output is limited to the message read path

## Tests
- `cargo test -p buzz-cli`: 290 passed
- `git diff --check`
- regressions cover empty objects, missing signatures, malformed schema, mismatched IDs, invalid signatures, malformed success JSON, raw-format command scoping, normalized compatibility, relay extension preservation, and deterministic equal-timestamp ordering

## Review note
The raw-message verification change was independently reviewed after remediation. The final signed-off head is `921188ab5399f8fc17ffff39f95d533a078925ab`. DCO, Semgrep OSS, and zizmor are green.

### DIFF
--- a/crates/buzz-cli/README.md
+++ b/crates/buzz-cli/README.md
@@ -33,6 +33,7 @@ buzz messages send --channel <uuid> --content "Hello"
 buzz messages send --channel <uuid> --content "Reply" --reply-to <event-id> --broadcast
 buzz messages send --channel <uuid> --content - < message.md   # read body from stdin
 buzz messages get --channel <uuid> --limit 20
+buzz --format raw messages get --channel <uuid> --limit 20  # complete signed Nostr events
 buzz messages thread --channel <uuid> --event <event-id>
 buzz messages search --query "architecture"
 buzz messages search --author <pubkey|npub|name> --since <unix-ts>
@@ -94,6 +95,11 @@ buzz repos protect remove --id my-repo --ref refs/heads/main
 # Pipe to jq
 buzz channels list | jq '.[].name'
 ```
+
+`--format raw` preserves complete relay event objects, including signatures, for
+`messages get`, `messages thread`, and `messages search`. Raw message results are
+ordered deterministically by timestamp and event ID when the command applies a
+chronological sort.
 
 `protect set` replaces every existing rule for the exact ref pattern. Any
 constraint omitted from the command is removed. `protect list` reports malformed
--- a/crates/buzz-cli/README.md
+++ b/crates/buzz-cli/README.md
@@ -51,6 +51,7 @@ buzz channels topic --channel <uuid> --topic "New topic"
 # Reactions
 buzz reactions add --event <event-id> --emoji "👍"
 buzz reactions get --event <event-id>
+buzz reactions get --event <event-id> --events  # verified signed kind-7 events, oldest first
 
 # Users & Presence
 buzz users get                          # your own profile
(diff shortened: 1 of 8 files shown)


BeelixGit:
Maintainer review gate receipt:

- External workflows are `ACTION_REQUIRED`: CI, Docker image, and Desktop Release Candidate. DCO, Semgrep OSS, and zizmor pass.
- Security: every returned reaction is deserialized as a complete signed `nostr::Event`, then its ID preimage and Schnorr signature are verified before aggregate or `--events` output. Incomplete, forged, tampered, bad-signature, wrong-kind, and wrong-target rows fail closed.
- Local gates pass: `cargo fmt --all -- --check`, `cargo test -p buzz-cli --lib` with 281 passed, `cargo clippy -p buzz-cli --all-targets -- -D warnings`, and `git diff --check`.
- Inbox dependency: `buzz reactions get --event "$EVENT_ID" --events` is globally paginated and sorted by `(created_at ASC, id ASC)`. Inbox can use the first event deterministically and retain its exact reaction event ID as the idempotency receipt.

`CODEOWNERS` maps this PR to `@block/buzz-oss-team`. GitHub does not permit the external PR author to request that team directly. Maintainer action is to approve the external workflow runs and assign or complete the code-owner review.

BeelixGit:
@tlongwell-block, when you have time, could you review this CLI contract? The PR now covers both verified reaction events and verified raw message events needed by downstream agent integrations. The body and test receipts are current at signed-off head `921188ab5399f8fc17ffff39f95d533a078925ab`.

BeelixGit:
Closing this as unnecessary. The downstream consumer that motivated it no longer needs it.

The intended consumer was an approval-reconciliation adapter that needed per-event IDs for reaction evidence, which the aggregate `reactions get` surface discards. That adapter now reads reactions through the relay's existing native raw-event surface instead:

```
buzz messages get --kinds 7 --channel <id>
```

That returns full stored kind-7 events with exact per-event IDs, filtered to the target root by its `e` tag. The relay derives each reaction's channel from its target event at ingest, so a channel-scoped kind-7 query reaches every reaction on every root in that channel. The relay verifies BIP-340 signatures on ingest and remains the verification boundary.

So stock Buzz already covered the requirement and no CLI change is needed. Rather than leave a patch open that nothing consumes, closing.

Thanks to anyone who spent time looking at it, and apologies for the noise from the stale check suites.

PR #4188, opened 2026-08-01
feat(cli): allow common audio attachments

sapience7-hub:
## Summary

- Allow common audio MIME types in CLI media uploads.
- Preserve audio metadata when constructing `imeta` tags.

## Validation

- `cargo test -p buzz-cli`

## Scope

This PR covers CLI-side audio attachment handling. End-to-end relay audio validation/storage and client playback remain separate work.

### DIFF
--- a/crates/buzz-cli/src/client.rs
+++ b/crates/buzz-cli/src/client.rs
@@ -67,6 +67,17 @@ const ALLOWED_MIMES: &[&str] = &[
     "image/gif",
     "image/webp",
     "video/mp4",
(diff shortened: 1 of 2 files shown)


sapience7-hub:
Important clarification: this PR does not make song attachments work end to end yet. The current relay/media pipeline still rejects audio uploads. An MP4 containing a still image plus audio is not a workaround because Buzz treats MP4 as video and requires a strict canonical H.264/AAC container. Native relay audio validation/storage and client playback support are still required.

PR #4190, opened 2026-08-01
fix(mobile): complete policy-gated invite joins

lunchboxfortwo:
## Summary

A plain `buzz://join?relay=…&code=…` link that joins successfully on desktop stops on mobile when the relay requires a join policy.

The mobile claim path in `mobile/lib/features/invites/invite_join_provider.dart:133` sent an existing `policy_receipt` when the deep link already contained one, but its claim-error handling treated `join_policy_required` as a terminal fresh-link error. It therefore never fetched `/api/join-policy`, accepted the policy, or retried `/api/invites/claim` with the returned receipt.

Live verification against `creatormagic.communities.buzz.xyz` found that a policy receipt is scoped to the invite-code hash, policy version, and expiry; it contains no joining pubkey or joiner signature. Its measured lifetime was about ten minutes (589 seconds remaining immediately after mint). That makes receipt-bearing links a useful compatible path, but not a durable solution: a link can easily outlive its receipt. Mobile needs to mint its own receipt at the moment the user reviews and explicitly accepts the policy, then proceed directly to the claim. If a link-carried receipt has already expired and the relay returns `join_policy_required`, this change takes the same fresh-consent path instead of dead-ending.

This change completes that handshake on mobile. On `join_policy_required`, the provider fetches the policy, presents its terms and privacy text, requires explicit unticked age/agreement confirmations when applicable, posts the user's actual confirmation to `/api/invites/accept-policy`, and retries the claim with the receipt. Declining leaves the user unjoined with a clear result, and accept-policy failures remain retryable from the review screen. Links that already carry a receipt still take the existing direct-claim path.

The former provider test named `join_policy_required requires a fresh link and cannot retry` encoded the old behavior. That assertion was incorrect because `join_policy_required` is the relay's signal to begin the recoverable policy handshake, not evidence that the invite itself is unusable. It has been replaced with end-to-end handshake coverage rather than removed without replacement.

### Related issue

None found after searching open issues and pull requests for mobile policy-gated invite joins and `join_policy_required`.

### Testing

- `cd mobile && flutter analyze` — no issues
- `cd mobile && flutter test` — 1,026 passed, 1 existing skip
- `just ci` — passed (format, Clippy, frontend checks/tests/builds, Rust/Tauri tests, web build, and mobile checks/tests)
- Added provider coverage for gated success, decline, accept-policy failure, and pre-receipted links
- Added widget coverage that both consent controls start unticked and gate the accept action
- Widget-rendered screenshots are posted in a follow-up PR comment

### DIFF
--- a/mobile/lib/features/invites/invite_join_provider.dart
+++ b/mobile/lib/features/invites/invite_join_provider.dart
@@ -25,9 +25,29 @@ enum InviteJoinStatus {
   idle,
   confirming,
   claiming,
+  reviewingPolicy,
+  acceptingPolicy,
+  declined,
   success,
   switchedExisting,
   error,
+}
+
+class InviteJoinPolicy {
+  final String? termsMarkdown;
+  final String? privacyMarkdown;
+  final bool ageAttestationRequired;
+  final String version;
+
+  const InviteJoinPolicy({
+    this.termsMarkdown,
+    this.privacyMarkdown,
+    required this.ageAttestationRequired,
+    required this.version,
+  });
+
+  bool get agreementRequired =>
+      termsMarkdown != null || privacyMarkdown != null;
 }
 
 class InviteJoinState {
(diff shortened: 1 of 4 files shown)


lunchboxfortwo:
### Policy review

Mobile fetches and displays the community policy after the relay returns `join_policy_required`; acceptance remains disabled until the required confirmations are made.

01-policy-review

### Explicit confirmations

Age attestation and policy agreement are separate, unticked controls. The accept action remains gated until both are explicitly selected.

02-explicit-confirmations

PR #4199, opened 2026-08-01
fix(mobile): discover, browse, and join open channels

brow:
### What changed?

Mobile now discovers relay-visible open channels from paginated kind 39000 metadata, while continuing to derive membership only from kind 39002. Discovered channels remain `isMember: false`, and private channels, DMs, and archived channels are excluded. Discovery stops when a page makes no progress and throws if it exceeds a hard iteration cap.

Users can browse eligible channels from quick actions, and users with no joined channels see the same joinable channels in the empty state. Both surfaces reuse the existing kind 9021 join path and refresh membership from the relay. The main channel list remains joined-only.

### Why?

Relay PR [#4196](https://github.com/block/buzz/pull/4196) adds open-channel memberships for future invite claimers. Users who claimed before that fix are already relay members, so an idempotent re-claim cannot repair their missing channel memberships. This mobile recovery path lets those users find and join open channels without a production data write. Together, the relay and mobile changes cover both populations.

### How is it tested?

At `62f20d2fd`, static analysis is clean and the focused provider and page suites pass all 44 tests.

Added tests:

- [`channels_provider_test.dart`](https://github.com/block/buzz/tree/main/mobile/test/features/channels/channels_provider_test.dart) covers zero-membership discovery, membership integrity, private and DM exclusion, deduplication, pagination, repeated-page termination, and the hard iteration cap.
- [`channels_page_test.dart`](https://github.com/block/buzz/tree/main/mobile/test/features/channels/channels_page_test.dart) covers browse eligibility, empty-state recovery, join delegation and refresh, joined-only rendering, and multi-channel empty-state layout.

Mutation checks killed removal of each recovery mechanism: discovery removal returns an empty zero-membership result, repeat-page guard removal issues an unexpected third request, empty-state browse removal hides the recovery channel, and quick-action browse removal hides the open channel.

The full mobile suite reports 1,030 passed, 1 skipped, and 1 unrelated host-scoped failure: `ChannelDetailPage keeps follow mode off while a tall newest message stays visible`. The same failure reproduces unchanged in isolation on this Mac.

### DIFF
--- a/mobile/lib/features/channels/channel.dart
+++ b/mobile/lib/features/channels/channel.dart
@@ -79,6 +79,8 @@ class Channel {
   bool get isPrivate => visibility == 'private';
   bool get isArchived => archivedAt != null;
 
+  bool get canJoin => visibility == 'open' && !isArchived && !isMember && !isDm;
+
   String displayLabel({String? currentPubkey}) {
     if (!isDm || participants.isEmpty) {
       return name;
--- a/mobile/lib/features/channels/channels_page.dart
+++ b/mobile/lib/features/channels/channels_page.dart
@@ -54,7 +54,7 @@ part 'channels_page/community.dart';
 part 'channels_page/quick_actions.dart';
 part 'channels_page/quick_actions_launcher.dart';
 
-enum _QuickAction { createChannel, newDm }
+enum _QuickAction { createChannel, newDm, browseChannels }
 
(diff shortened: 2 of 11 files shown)


brow:
@codex review

brow:
@builderbot review

brow:
@codex review

brow:
@builderbot review

PR #4203, opened 2026-08-01
fix(nip-oa): accept raw Nostr tag form in parse_json_array

amanning3390:
## What

`BUZZ_AUTH_TAG` stored in the **raw Nostr tag form** `[auth,hex,,hex]` (unquoted, comma-delimited — how an `auth` tag serializes inside a Nostr event and how `.env` files commonly store it) was rejected by the CLI:

```
BUZZ_AUTH_TAG is malformed: invalid JSON: expected value at line 1 column 2
```

…and even when the CLI *could* parse it, it forwarded the raw string as the `x-auth-tag` header, so the relay's `verify_auth_tag` (which expects JSON) rejected it with `403 relay_membership_required`.

Two commits close both gaps.

### 1. `fix(nip-oa): accept raw Nostr tag form in parse_json_array`

`parse_json_array` (`crates/buzz-sdk/src/nip_oa.rs`) only accepted well-formed JSON arrays. Added a fallback: when strict JSON parsing fails *and* the trimmed input is bracket-delimited, split on `,` and treat each field as a string (empty field `,,` → empty string, matching `["auth","hex","","hex"]`). All consumers (`parse_auth_tag`, `verify_auth_tag`, the CLI, `buzz-acp`) benefit from one change at the lowest layer.

### 2. `fix(cli): canonicalize BUZZ_AUTH_TAG to JSON before sending x-auth-tag header`

The CLI stored the raw input string and sent it verbatim as the `x-auth-tag` header (`client.rs:618`). Added `canonicalize_auth_tag` in `buzz-sdk`: parse either form, re-serialize to canonical JSON. The CLI now canonicalizes before storing as `auth_tag_json`, so the header is always valid JSON regardless of input form.

Together: local parse + wire canonicalization means the raw form works end-to-end.

## Why

The raw form `[auth,hex,,hex]` is exactly how an `auth` tag serializes inside a Nostr event. That shape leaks into `.env` files and shell variables because there's no canonical "stored form" outside an event. The SDK + CLI should accept it rather than push quoting/conversion logic onto every consumer (harnesses, agent shells, external tools).

## Security

Both changes are purely syntactic — they only change how a 4-element string array is extracted and containerized. All downstream validation is unchanged:
- `parse_auth_tag`: still checks exactly 4 elements, `"auth"` label, 64-char lowercase-hex pubkey, 128-char signature.
- `verify_auth_tag`: still reconstructs the preimage and verifies the BIP-340 Schnorr signature against the owner pubkey.

No new attack surface — a malformed or forged tag is still rejected at the same validation points.

## Tests

4 new tests in `nip_oa::tests`:
- `test_parse_auth_tag_raw_nostr_form` — raw form with conditions + empty conditions
- `test_parse_auth_tag_raw_form_with_whitespace` — raw form with surrounding whitespace
- `test_canonicalize_auth_tag_raw_to_json` — raw→JSON and JSON→JSON normalization

All 25 `nip_oa` tests pass (21 existing + 4 new). `cargo fmt --check` and `cargo clippy -p buzz-sdk -p buzz-cli` clean.

## Verification

Confirmed end-to-end against a live community relay (`wss://hermesagent.communities.buzz.xyz`):
- **Before:** raw `BUZZ_AUTH_TAG` → CLI parse error, or `403 relay_membership_required` if somehow parsed.
- **After:** raw `BUZZ_AUTH_TAG` → CLI parses it, canonicalizes to JSON for the header, relay accepts via NIP-OA owner delegation, `buzz channels members` returns the full roster.

## Context

Originated from a community investigation where agent-side relay access was failing because the harness-exported `BUZZ_AUTH_TAG` (raw Nostr form) was rejected by the CLI (expecting JSON). This removes the impedance mismatch at the source.

### DIFF
--- a/crates/buzz-cli/src/lib.rs
+++ b/crates/buzz-cli/src/lib.rs
@@ -1768,6 +1768,41 @@ pub enum ModerationCmd {
     },
 }
 
+/// Normalize hand-authored `BUZZ_AUTH_TAG` input to strict JSON.
+///
+/// `.env` files and shell exports sometimes carry the tag in the unquoted
+/// shorthand `[auth,<hex>,<conditions>,<hex>]` (quotes dropped by hand).
+/// When the input is not valid JSON but is bracket-delimited, rewrite it as
+/// a JSON array of the comma-separated fields (an empty field `,,` becomes
+/// `""`, matching the canonical form `["auth","hex","","hex"]`).
+///
+/// This is presentation-layer leniency at the configuration edge only: the
+/// output is always fed through the SDK's strict `parse_auth_tag` /
+/// `verify_auth_tag`, which enforce structure, hex, the conditions grammar,
+/// and the BIP-340 signature. Inputs that are already valid JSON — or not
+/// recognizable as the shorthand — are returned unchanged so the strict
+/// parser reports the error on the original bytes.
+fn normalize_auth_tag_input(input: &str) -> String {
+    let trimmed = input.trim();
+    if serde_json::from_str::<serde_json::Value>(trimmed).is_ok() {
+        return trimmed.to_owned();
+    }
+    if trimmed.starts_with('[') && trimmed.ends_with(']') {
+        let fields: Vec<&str> = trimmed[1..trimmed.len() - 1]
+            .split(',')
+            .map(str::trim)
+            .collect();
+        // Only a plausible 4-field auth tag is rewritten; anything else is
+        // passed through untouched for the strict parser to reject with an
+        // error that references the caller's original input.
+        if fields.len() == 4 && !fields.iter().any(|f| f.contains('"')) {
+            // serde_json cannot fail serializing a Vec<&str>.
+            return serde_json::to_string(&fields).expect("string array serializes");
+        }
+    }
+    trimmed.to_owned()
+}
+
 async fn run(cli: Cli) -> Result<(), CliError> {
     let relay_url = client::normalize_relay_url(&cli.relay);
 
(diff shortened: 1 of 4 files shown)


tlongwell-block (changes requested):
Requesting changes for one concrete regression and one API-contract issue at `88471dd668c633deae2ef0ae5e2658e1dd70b9d7`.

1. The new `test_load_auth_tag_accepts_raw_nostr_form` mutates process-global `BUZZ_AUTH_TAG` concurrently with the existing `test_load_auth_tag_rejects_bad_conditions`. This makes the package suite flaky. Running `cargo test -q -p git-sign-nostr --lib test_load_auth_tag -- --test-threads=2` repeatedly failed on iteration 2: the existing test expected `kind = 9` to be rejected but read the new test's valid value instead. Please test a pure parsing helper, or serialize all tests that mutate this env var behind the same lock.

2. `canonicalize_auth_tag` is public and documents that it errors unless its input is a valid 4-element `auth` tag, but the implementation only checks array length. For example, `["not-auth","x","y","z"]` canonicalizes successfully. The CLI happens to call `parse_auth_tag` and `verify_auth_tag` first, so the current CLI path remains fail-closed, but the public helper does not uphold its stated contract and invites unsafe standalone use. Either make this a private serialization helper or have it validate via the same typed/validated path before returning.

The production direction is otherwise sound for the exact accepted grammar: strict JSON remains the fast path, raw fallback fields are still checked for label/count/lowercase hex/conditions/signature, CLI verification remains before header construction, and canonical JSON fixes the wire header. Splitting on commas is safe only because the current conditions grammar forbids commas; that dependency should be explicit.

Also, the repeated parser implementations in SDK, credential helper, and signer are drift-prone, and the claim that `[auth,...]` is how a tag serializes "inside a Nostr event" is inaccurate: an event serializes tags as JSON arrays. This may be an accepted local/env shorthand, but it is not Nostr event JSON.

tlongwell-block:
Pushed a maintainer follow-up commit (0b74eee24) that keeps this PR's intent — the unquoted `[auth,hex,,hex]` shorthand for hand-authored `BUZZ_AUTH_TAG` works end-to-end via the CLI — while narrowing where the leniency lives:

- **SDK, git-credential-nostr, git-sign-nostr: reverted to base.** `verify_auth_tag` is the relay's `x-auth-tag` entry point, so the lenient grammar had widened the public wire format (probing `extract_nip_oa_owner` base-vs-head showed raw going rejected→accepted). Wire grammar stays strict JSON per NIP-GS. Reverting git-sign-nostr also removes the new test that raced `test_load_auth_tag_rejects_bad_conditions` on the process-global env var under the default parallel runner.
- **`canonicalize_auth_tag` removed.** The CLI now derives the wire string from the already parsed-and-verified `Tag` via `serde_json::to_string(tag.as_slice())` (same shape `buzz-acp`'s RestClient uses) instead of re-parsing unverified input.
- **The leniency lives in one place:** `normalize_auth_tag_input` in `buzz-cli`, applied to the configuration input before the unchanged strict parse/verify path. Valid JSON passes through untouched; unrecognizable input is left for the strict parser to reject.

Verified locally at 0b74eee24: buzz-cli 274, buzz-sdk 241, git-credential-nostr 8, buzz-relay lib 835 pass; git-sign-nostr no longer flakes across 8 consecutive default-settings runs; fmt/clippy clean; live CLI run against a relay with a real delegated credential in both JSON and raw forms, plus a fail-closed garbage control.

Thanks for the report and the repro — the shorthand support was worth having; it just needed to live at the config edge rather than in the shared grammar.

tlongwell-block (approved):

PR #4215, opened 2026-08-02
fix(relay): accept NIP-59 backdated gift wraps without losing push wakes

### LINKED ISSUE #4192: Relay rejects spec-compliant NIP-17 DMs: gift-wrap (kind 1059) timestamps fail the freshness window
# Relay rejects spec-compliant NIP-17 DMs: gift-wrap (kind 1059) timestamps fail the freshness window

## Summary

NIP-59 instructs clients to randomize a gift wrap's `created_at` into the past — "The `created_at` timestamp SHOULD be tweaked to thwart time-analysis attacks" (canonically up to 2 days back; `nostr-tools` and other mainstream implementations do this by default). The Buzz relay applies its event-freshness window to the **outer kind-1059 wrap** and rejects it:

```
["OK", "<id>", false, "invalid: event timestamp too far from server time"]
```

Practical effect: **no spec-following NIP-17 client can deliver a private DM through a Buzz relay.** The wraps produced by `nostr-tools`' `nip17.wrapEvent()` (its defaults backdate the wrap) are rejected 100% of the time.

## Reproduction

Against a relay at `63496cc1d` (observed 2026-08-01):

```js
import { wrapEvent } from 'nostr-tools/nip17';
const wrap = wrapEvent(senderSecretKey, { publicKey: recipientHex }, 'hello');
// wrap.created_at is randomized into the past per NIP-59
// send over authenticated WS:
ws.send(JSON.stringify(['EVENT', wrap]));
// → ["OK", id, false, "invalid: event timestamp too far from server time"]
```

Re-signing the outer wrap with `created_at = now` (fresh ephemeral key, same sealed payload) is accepted — confirming the freshness check on the wrap's timestamp is the only blocker.

## Where

`crates/buzz-relay/src/handlers/ingest.rs:1524` — the freshness rejection is applied uniformly, with no carve-out for `KIND_GIFT_WRAP` (`buzz-core/src/kind.rs:60`, already registered in the accepted-kinds tables).

## Suggested fix

Exempt kind 1059 from the `created_at` freshness window (or widen its window to NIP-59's two days plus skew). The wrap's timestamp is deliberately meaningless — ordering/retention decisions for gift wraps cannot rely on it anyway, and the inner seal/rumor timestamps are end-to-end encrypted. Rejecting backdated wraps only forces clients to defeat NIP-59's privacy mechanism (our current workaround: re-stamp the outer wrap to `now`, leaking exactly the timing metadata the spec hides).

## Side observation (probably intentional, mentioning for completeness)

The HTTP bridge rejects gift wraps outright with `invalid: kind 1059 is only accepted via WebSocket`. If that restriction is by design it may be worth a line in the bridge docs, since every other event kind we use round-trips fine over `POST /events`.

alanshurafa:
## Summary

NIP-59 tells clients to randomize a gift wrap's `created_at` into the past ("tweaked to thwart time-analysis attacks"); `nostr-tools`' `wrapEvent` backdates uniformly up to two days by default. The relay's uniform ±15-minute ingest freshness window therefore rejects ~99% of spec-compliant NIP-17 DMs with `invalid: event timestamp too far from server time`, and the only workaround (re-stamping the wrap to `now`) defeats exactly the timing privacy the spec mandates.

Two coordinated changes:

**1. Ingest: past-side allowance for kind 1059 only.** The freshness check moves into a pure helper; ordinary kinds keep the exact ±900 s behavior, gift wraps get `900 s + 2 days` on the past side. The future-side bound is not widened for any kind — a wrap stamped into the future is still bogus.

**2. Push runtime: gift-wrap wake deadlines key on receipt time.** This is the half of the fix that's easy to miss: the wake matcher computes `event_deadline = created_at + EVENT_USEFUL_SECS (3600 s)` and silently skips expired wakes. That coupling is invisible today only because 900 < 3600. With the widened window alone, a wrap backdated >1 h (≈98% of NIP-59's uniform randomization) would be accepted, stored, fanned out to live sockets — and never wake a backgrounded device; it would also hand senders a deliberate no-notification delivery primitive. Gift-wrap deadlines now use relay-assigned `StoredEvent::received_at` (author-controlled `created_at` stays authoritative for every other kind; verified that the push matcher re-loads events from the DB, so `received_at` is the original ingest stamp, never refreshed). A compile-time `const` assertion pins the full relationship (`MAX_TIMESTAMP_DRIFT_SECS + GIFT_WRAP_MAX_BACKDATE_SECS > EVENT_USEFUL_SECS`) so the coupling is explicit.

Interactions checked and intentionally out of scope:

- **Replica fence (migration 0021):** the commit-time `created_at` floor guard fires only for `channel_id IS NOT NULL` rows; gift wraps are structurally stored channel-NULL, so the fence is unaffected. The `CREATED_AT_FLOOR_SECS` doc comment now states its invariant precisely (floor covers channel-bearing kinds) instead of implying a single global envelope.
- **Replay/dedup:** unchanged — events dedupe on `(community_id, id)` `ON CONFLICT DO NOTHING`.
- **Retention/partitioning:** no `created_at`-keyed retention applies to kind 1059 (channel-TTL keys on `clock_timestamp()`; NIP-RS retention is kind-30078); the monthly range partitioning has a `MINVALUE` catch-all. Push backfill windows already key on `received_at`.
- **The HTTP bridge's** `kind 1059 is only accepted via WebSocket` restriction (the issue's side observation) is untouched — and it sits before the timestamp check, so bridge-submitted wraps never reach the new window logic.
- **REQ `since`/`until` and keyset paging still key on `created_at`** — inherent to Nostr semantics, and inherent to NIP-59: a client that pages gift wraps with a narrow `since` will miss backdated wraps *on any relay*, which is exactly why spec-following NIP-17 clients query kind 1059 with a ~2-day-wide window. Worth naming the resulting UX asymmetry honestly: a recipient can get a push wake for a wrap their `since`-based sync then doesn't fetch. No in-repo client subscribes to 1059 with `since` today (verified across desktop/web/mobile), and push-lease filters structurally cannot carry `since`/`until` (rejected, not ignored — `ALLOWED` list in `push_lease.rs`), so nothing trips it now; if it ever matters, the natural follow-up is serving `#p`-scoped wrap reads on a `received_at` basis, the same choice the push backfill already made.
- **Partitioning:** `events` is month-range-partitioned on `created_at` with a `MINVALUE` catch-all; a wrap backdated across a month boundary lands in the prior month's partition, which is valid (no partition-drop code exists).
- `api/git/policy.rs`'s "timestamp too far in future" is git hook-callback freshness — unrelated path.
- #4067 fixes the same error string for kind-30617 announcements by re-stamping client-side; that remedy is correct there but is exactly what NIP-59 forbids for gift wraps, hence the relay-side carve-out here.

### Related issue

Fixes #4192. Closest adjacent work: #4067 (same error string, client-side fix for a kind where re-stamping is harmless). At filing (2026-08-02 00:59 UTC) no PR addressed 1059 freshness. **Update:** #4253 and #4324 (its successor) have since been filed for this issue. Both widen the ingest window only, and symmetrically (`.abs()` — wraps stamped up to two days in the *future* also become acceptable, though NIP-59 randomizes into the past only), and both leave the push-wake deadline keyed on `created_at`, so accepted backdated DMs would be stored but never wake a backgrounded device (analysis above). This PR widens the past side only and rekeys the kind-1059 wake deadline on `received_at`.

### Testing

- `cargo test -p buzz-relay --lib` — passed (new: ±900 boundary pins for ordinary kinds; 1059 accepts `now − 2d` and the `now − (2d + 900 s)` boundary, rejects older; future +901 s rejected for 1059 too; hostile extreme timestamps (`i64::MIN`/`MAX`, wrapped-negative u64 casts) rejected via saturating arithmetic; wake-deadline basis tests: backdated wrap wakes from `received_at`, ordinary kinds from `created_at`; plus a compile-time constant-relationship pin).
- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check` — clean.
- Toolchain note: Windows `x86_64-pc-windows-gnu` (as in #2638); Postgres-gated suites (`#[ignore]`) not run.

### DIFF
--- a/crates/buzz-db/src/replica_fence.rs
+++ b/crates/buzz-db/src/replica_fence.rs
@@ -65,10 +65,18 @@ use uuid::Uuid;
 
 /// Seconds of `created_at` history the commit-time floor guard tolerates.
 ///
-/// Must exceed the relay's ingest envelope (±900 s) by enough slack that a
-/// legitimately accepted event still commits within the floor even under
-/// slow validation/lock waits. The writer pool arms the guard with this value
-/// and the fence subtracts it; the two uses must never diverge.
+/// Must exceed the relay's ingest envelope for **channel-bearing** kinds
+/// (±900 s) by enough slack that a legitimately accepted event still commits
+/// within the floor even under slow validation/lock waits. The writer pool
+/// arms the guard with this value and the fence subtracts it; the two uses
+/// must never diverge.
+///
+/// The floor deliberately does not cover NIP-59 gift wraps (kind 1059),
+/// which ingest accepts up to two days in the past: they are always stored
+/// `channel_id = NULL`, and the guard's trigger fires only for channel-
+/// bearing rows — the structural exemption [`verify_floor_guard_behavior`]
+/// pins in step 5. Channel-NULL rows are never keyset-paged by channel, so
+/// they sit outside the fence proof's read-window obligations.
 pub const CREATED_AT_FLOOR_SECS: i64 = 960;
 
 /// Safety margin subtracted from the fence on top of the floor.
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -1729,6 +1729,45 @@ async fn author_type_label(
     }
 }
 
+/// Maximum seconds an event's `created_at` may deviate from server time,
+/// in either direction, before ingest rejects it.
+pub(crate) const MAX_TIMESTAMP_DRIFT_SECS: i64 = 900; // ±15 minutes
+
+/// Extra past-side allowance for NIP-59 gift wraps (kind 1059), whose
+/// `created_at` clients randomize up to two days into the past ("SHOULD be
+/// tweaked to thwart time-analysis attacks") — `nostr-tools`' `wrapEvent`
+/// backdates by default, so a uniform freshness window rejects essentially
+/// every spec-compliant NIP-17 DM (#4192). The future-side bound is NOT
+/// widened: a wrap stamped into the future is still bogus.
+///
+/// Two couplings to keep in mind when touching this:
+/// - Push wakes: this allowance exceeds `push_runtime`'s usefulness window,
+///   so gift-wrap wake deadlines key on relay receipt time, not `created_at`
+///   (see `push_runtime::wake_deadline_basis` and the test pinning the
+///   relationship).
+/// - Replica fence: `buzz_db::replica_fence::CREATED_AT_FLOOR_SECS` covers
+///   channel-bearing kinds only; gift wraps are stored channel-NULL, which
+///   the floor guard structurally exempts (behaviorally verified by
+///   `verify_floor_guard_behavior`, step 5).
+pub(crate) const GIFT_WRAP_MAX_BACKDATE_SECS: i64 = 2 * 24 * 60 * 60;
+
+/// Whether `event_ts` (unix seconds) falls inside the ingest freshness
+/// window: ±[`MAX_TIMESTAMP_DRIFT_SECS`] for every kind, with the past side
+/// extended by [`GIFT_WRAP_MAX_BACKDATE_SECS`] for kind 1059 only.
+fn timestamp_within_ingest_window(kind: u32, event_ts: i64, now: i64) -> bool {
+    let max_past_drift = if kind == KIND_GIFT_WRAP {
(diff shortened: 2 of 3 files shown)


alanshurafa:
@block/buzz-oss-team — could a maintainer please approve the workflow runs for this fork PR and do a code-owner review? The reachable checks (Semgrep OSS, zizmor, DCO) pass and the branch is rebased on current main (28ae6cd21). Thanks!

PR #4220, opened 2026-08-02
Harden shared agent instruction review

shellz-n-stuff:
## Summary

- render shared-agent instructions as literal text so Markdown cannot conceal spoiler contents, link destinations, or image sources
- reject non-reviewable Unicode controls at every agent-definition boundary while preserving legitimate rendered emoji sequences
- verify shared catalog event IDs and signatures before trusting authorship, coordinates, pagination, or executable content
- preserve the exact system-prompt bytes between review and execution instead of silently stripping or normalizing content

## Security rationale

Shared system prompts are executable configuration. Previously, catalog prompts were projected through the chat Markdown renderer, which could hide text, replace link destinations with benign labels, and turn image syntax into remote loads. Zero-width and bidirectional controls could also make reviewed text differ from what the agent executes.

This change establishes a review invariant: the prompt a user sees is the prompt the agent executes. Definitions that cannot be reviewed faithfully are rejected rather than rewritten. Catalog events must also pass Nostr ID/signature verification before they can claim a publisher, coordinate, or cursor.

## What changed

- catalog instructions render as exact literal text rather than rich Markdown
- catalog relay events are verified on a fresh wire-shaped object before paging, coordinate selection, attribution, or projection
- forged content, pubkeys, signatures, and invalid newer heads are ignored and cannot shadow a valid signed definition
- TypeScript catalog parsing rejects unsafe remote definitions before they reach the UI
- shared Rust validation covers persona create/update/import, inbound relay sync, definition-less managed-agent sync, and catalog publication paths
- definition-less managed agents now fail closed on local create, local update, and publication before persistence or relay retention
- linked managed agents validate their local name while treating the persona definition as authoritative; their inert record-level prompt is not executed or published
- names reject layout controls; prompts retain ordinary newlines and tabs
- legitimate emoji composition is supported, including contextual VS16, ZWJ, skin-tone, family, flag, and keycap sequences
- detached selectors/joiners, bidirectional controls, tag characters, zero-width concealment, and other default-ignorables remain rejected
- names are bounded to 128 characters and prompts to 64 KiB
- contributor guidance documents the byte-for-byte review requirement for future sharing paths

Validation reports the offending code point and never silently removes it.

## E2E recording

[buzz-shared-agent-security-e2e.webm](https://github.com/user-attachments/assets/44d6b75f-0877-490f-bda4-a716fae3f700)

The recording demonstrates:

- a safe definition remains visible
- a prompt containing zero-width `U+200B` is rejected
- a name containing bidi override `U+202E` is rejected
- the prompt is preserved exactly
- spoiler, link, and image syntax remains literal and does not render or load

## Verification

Passed locally:

- `just test`: all 10 unit and Docker-backed integration stages
- desktop frontend unit suite: 4,295 tests
- persona catalog relay unit suite: 32 tests, including forged-event and cursor-shadowing cases
- focused Rust definition-validation coverage: 3 local create/update tests and 6 publication-filtered tests
- complete desktop Tauri library suite after rebase: 2,263 passed, 14 ignored, 0 failed
- desktop Tauri clippy with warnings denied and Rust formatting
- complete agent Playwright spec: 34 tests
- the exact formerly failing `inbox-edit` immediate-attachment smoke test after rebase: 1 test
- focused shared-agent publish, literal-review, hidden-control, signature, and cross-member import Playwright coverage
- desktop E2E production build and TypeScript typecheck
- changed-file formatting/lint and file-size ratchet
- pre-commit secret scan and DCO signoff

The branch was rebased onto current `main`, which includes the upstream attachment-button label fix. Fresh post-rebase GitHub CI is green for every required and selected check: Desktop Core, all four Desktop Smoke E2E shards, both Desktop E2E Integration shards and their aggregate, Desktop E2E Relay, Desktop Build (macOS), Windows Rust, Rust Lint, DCO, security scanners, and Desktop Release Candidate. The previously failing `Desktop Smoke E2E (3)` shard now passes.

The repository-wide desktop check also reports existing CSS formatting/`!important` findings in `components.css` and `terminal.css`; neither file is changed by this PR. GitHub's Desktop Core lint and format stage passes on the rebased branch.


desktop/src-tauri/src/commands/personas/create.rs
```diff
@@ -25,7 +25,10 @@ pub async fn create_persona(
         let state = app.state::<AppState>();
         let display_name = trim_required(&input.display_name, "Display name")?;
         // System prompt optional: core memory is auto-injected. Empty is valid.
-        let system_prompt = input.system_prompt.trim().to_string();
+        // Preserve it byte-for-byte: shared/import review surfaces show this
+        // exact string before the ACP harness executes it.
+        let system_prompt = input.system_prompt.clone();
+        validate_agent_definition_text(&display_name, &system_prompt)?;
```
wesbillman, line 31: **[P1] Preserve legitimate emoji sequences instead of rejecting every default-ignorable code point**

Applying this validator to every create/update makes existing definitions containing ordinary emoji sequences impossible to edit: `⚠️` and `❤️` contain VS16 (`U+FE0F`), while `👨‍💻` and family/flag emoji contain ZWJ (`U+200D`), and both are rejected by `is_default_ignorable`. The TypeScript copy of this check also drops already-published shared definitions containing these characters from the catalog entirely (`personaCatalogRelay.ts:66,70`). This is not limited to hostile or newly shared data; a user can open an existing private agent, change an unrelated field, and have save fail because its unchanged name/prompt contains a normal emoji. Please narrow the prohibited set or validate emoji composition context so legitimate rendered graphemes remain accepted, while still rejecting bidi controls, zero-width concealment, tag characters, and other genuinely non-reviewable text. Add regression coverage for existing local and shared definitions containing VS16 and ZWJ emoji.

wesbillman (changes requested):
Requesting changes on Wes's behalf after independent security, regression, and integration reviews of `d56523fcbab119e78cc2a208c0b78cae55dfe4b3`.

Two blockers remain:

1. The catalog accepts unverified relay events as executable definitions and attributes them to attacker-selected pubkeys. A compromised relay can forge a member's shared agent and prompt a user to import arbitrary instructions.
2. The validator rejects legitimate VS16/ZWJ emoji sequences, breaking edits to existing local agents and hiding valid shared definitions from the catalog.

The literal `<pre>` review surface, Rust ingress validation, snapshot/team-snapshot coverage, pre-retention validation, and TypeScript/Rust range parity otherwise look sound. Existing CI at this exact head is green; reviewers did not duplicate CI-equivalent suites locally.

desktop/src/features/agents/lib/personaCatalogRelay.ts
```diff
@@ -133,10 +188,14 @@ function parsePersonaContent(event: RelayEvent): CatalogAgentProjection | null {
...
+
+  const displayName = parsed.display_name;
+  const systemPrompt =
+    typeof parsed.system_prompt === "string" ? parsed.system_prompt : "";
   if (
-    !isObject(parsed) ||
-    typeof parsed.display_name !== "string" ||
-    parsed.display_name.trim().length === 0
+    typeof displayName !== "string" ||
+    !isSafeAgentDefinitionText(displayName, systemPrompt)
```
wesbillman, line 270: **[P0] Verify catalog event signatures before trusting authorship or executable content**

Text validation here is downstream of an unverified relay boundary. `fetchEvents()` does not establish authenticity: `relayClientSession.ts:783-785` merely casts relay JSON to `RelayEvent`, and its history path retains it without `verifyEvent`. This parser then trusts `event.pubkey`, `id`, timestamp, tags, and prompt. A malicious or compromised relay can fabricate a newer kind-30175 event under any member's pubkey, have the UI attribute arbitrary instructions to that member, and induce the user to import and execute them. It can also submit an invalid newer head that claims the coordinate before parsing and thereby suppress a legitimate signed definition. The repo already imports `verifyEvent` from `nostr-tools/pure` in `desktop/src/shared/lib/authors.ts`. Require successful id/signature verification before an event can claim `seenCoordinates` or reach projection, and add tests proving forged pubkey/content/signature events are ignored and cannot shadow an older valid signed head.

desktop/src-tauri/src/commands/personas/inbound.rs
```diff
@@ -108,6 +108,9 @@ fn reconcile_inbound_persona_event_blocking(
     let inbound_persona = (kind == KIND_PERSONA)
         .then(|| persona_from_event(&event))
         .transpose()?;
+    if let Some(persona) = &inbound_persona {
+        validate_inbound_persona_definition(persona)?;
+    }
```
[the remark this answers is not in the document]
shellz-n-stuff: 🤖 Updated by Codex. The inbound kind-30177 path now validates definition-less managed-agent names and optional prompts before either retaining or applying the event. Regression coverage includes zero-width and bidi payloads plus a visible multiline prompt.

desktop/src-tauri/src/commands/personas/create.rs
shellz-n-stuff: 🤖 Updated by Codex. The validator now permits VS16 and ZWJ only in valid rendered emoji composition contexts, while continuing to reject detached selectors/joiners, bidi controls, tag characters, and other default-ignorables. Rust, catalog-unit, and Playwright coverage now includes heart, technologist, skin-tone, family, and keycap sequences.

desktop/src/features/agents/lib/personaCatalogRelay.ts
shellz-n-stuff: 🤖 Updated by Codex. Catalog events are now verified with nostr-tools before they can affect paging, coordinate selection, attribution, or projection. Verification uses a fresh wire-shaped event to avoid stale memoized verification metadata. Tests cover forged content, pubkeys, signatures, newer-head shadowing, and invalid cursor control.

wesbillman (dismissed):
Requesting changes on Wes's behalf after coordinated review of `fc53800ef0d849a127d850598f35b9d2d867c099` with Princess Donut and Mongo.

The two prior blockers are fixed: catalog events are now verified on fresh wire-shaped objects before they can claim coordinates or influence paging, and the TS/Rust validators permit contextual VS16/ZWJ emoji while rejecting detached formatting. The literal review surface and catalog add path preserve exact prompt bytes. Mongo independently found no remaining issue in those paths.

One blocker remains: **the Rust safety invariant is incomplete for other executable, relay-synced definition text.**

- Definition-less `ManagedAgentRecord.system_prompt` is executable, but `create_managed_agent` and `update_managed_agent` accept it without `validate_agent_definition_text`; those records are persisted and queued for publication. Inbound definition-less records do get validated, leaving an asymmetric local bypass.
- Team instructions are executable, but team create/update and inbound kind `30176` apply them without the validator. Team snapshot validation validates member definitions but not the executable `team.instructions` (or the team name), then import persists those fields.

As a result, invisible/bidi text such as `Review\u{200B}code` can still be persisted, executed, and published through local managed-agent/team paths even though equivalent persona and inbound managed-agent definitions are rejected. Please centralize validation for executable name/instruction fields across create, update, import, sync, and publication boundaries, with direct regressions for definition-less agents and teams—or explicitly narrow the promised security contract and demonstrate why these executable fields are outside it.

Existing CI is green at this exact head. Focused local test attempts in my clean review worktree were environment-blocked (missing JS dependencies; missing Tauri sidecar), so I did not count them as validation. Mongo separately ran the focused TS suites: 44/44 passed.

wesbillman:
Scope clarification on my changes-requested review, on Wes's behalf:

I agree the **team-instructions** work can be split into a follow-up. It is executable shared text and the gap is real, but this PR's concrete review surface is agent/persona definitions; extending the invariant to teams broadens the product/UI contract enough to justify separate work. Please track that follow-up before merge.

I do **not** think the definition-less managed-agent gap is separable. This PR explicitly added inbound validation for definition-less kind `30177` records and its description claims “definition-less managed-agent sync” plus validation at every agent-definition boundary. Leaving local create/update able to persist, execute, and publish the same unsafe prompt makes that new boundary asymmetric and directly contradicts the PR's stated invariant.

So the narrowed blocker is: validate definition-less managed-agent name/prompt on local create/update (and at publication as defense in depth), with direct regressions. Team validation may follow separately with an issue/PR.

wesbillman (commented):
Reviewing `e6aa35849204fc33dd9bab846e2987b637e32e7d` on Wes's behalf.

The remaining in-scope blocker from my prior changes-requested review is fixed:

- local definition-less agent create validates the trimmed name and executable prompt before persistence;
- update applies the prospective prompt/name, validates the complete record, and returns before save on failure;
- kind `30177` inbound sync validates before retention;
- publication validates again as defense in depth;
- linked records continue to treat their record-level prompt as inert and persona-authoritative.

I also re-traced catalog signature verification, coordinate selection, cursor derivation, literal prompt rendering, persona create/update/import, and agent snapshot decode/import at this head. I found no new actionable defect in the PR's narrowed agent/persona scope. Existing required CI is green for this exact head; I did not duplicate the broad suites locally.

Two merge-state notes remain:

1. GitHub currently reports this branch as conflicting with `main` (`baseRefOid` is `96ae141763e5459beb68a847e0082e931af72f4c` while current `origin/main` is `538e5e113fc33571f939c87b925567fd4e277109`), so the rebased result still needs review/CI.
2. The agreed follow-up for executable team names/instructions is not present in this PR, and I found no matching open or closed issue in `block/buzz` using searches for team-instruction Unicode/zero-width/bidi validation. Please track that before merge as previously agreed. Team create/update, inbound kind `30176`, publication, and team snapshot validation remain outside this patch.

I am leaving this as a comment rather than approval: Wes requested review, not approval.

wpfleger96 (approved):

PR #4240, opened 2026-08-02
fix(acp): keep all direct-message replies flat

### LINKED ISSUE #3075: buzz-acp: a top-level DM is answered in a thread, not inline — the DM branch supplies no reply destination

kerpopule:
## Summary

Keep managed-agent direct messages as a normal flat conversation. Every ordinary DM response is sent at the channel top level, even when the incoming DM carries legacy thread tags. Human-facing channel replies retain their existing thread-root behavior.

This intentionally applies the flat behavior more broadly than #3077. It resolves the top-level case in #3075 and removes the compounding nested-DM behavior described in #2748 by eliminating DM reply anchors rather than moving them to a root.

## Implementation

- Add a DM-specific prompt instruction that forbids `--reply-to` for both top-level and previously threaded DM events.
- Remove DM reply-anchor generation while leaving non-DM anchor resolution unchanged.
- Align the shared base prompt with the flat-DM contract.
- Cover top-level DMs, legacy threaded DMs, and the shared prompt contract with regression tests.

## Verification

- `env -u BUZZ_ACP_LAZY_POOL cargo test -p buzz-acp`: 662 unit tests and 9 integration tests passed
- Existing channel-thread anchoring tests pass

`just ci` reached the mobile suite, where the unrelated `ChannelDetailPage keeps follow mode off while a tall newest message stays visible` test failed. This change does not touch mobile code.

Fixes #3075.
Addresses #2748 with the product behavior requested here: no threads in direct messages.

### DIFF
--- a/crates/buzz-acp/src/base_prompt.md
+++ b/crates/buzz-acp/src/base_prompt.md
@@ -53,7 +53,9 @@ For explicit changes to an existing personal agent, use `buzz agents draft-updat
 
 Use the reply destination supplied in the `[Context]` block for ordinary replies in this turn. Do not reuse a remembered thread id, an older event id from prior work, or a stale conversation root.
 
-For human-facing work, keep the conversation flat and easy to read. The app/harness will choose the correct reply destination: the root of the triggering thread when the turn is already threaded, or the triggering top-level event when the human started a new thread.
+Direct messages are always flat. Send each ordinary DM response as a top-level message in the conversation. Do not pass `--reply-to` in a DM, even when the incoming DM carries thread tags from an earlier exchange.
+
+For human-facing channel work, keep the conversation flat and easy to read. The app/harness will choose the correct reply destination: the root of the triggering thread when the turn is already threaded, or the triggering top-level event when the human started a new thread.
 
 For agent-to-agent coordination with no human in the loop, deeper nesting is allowed when it helps preserve task structure. Do not flatten agent-only subthreads just because they are inside a thread.
 
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -3627,6 +3627,13 @@ mod agent_draft_prompt_tests {
     }
 
     #[test]
+    fn shared_base_prompt_teaches_flat_dm_replies() {
+        let prompt = include_str!("base_prompt.md");
+        assert!(prompt.contains("Direct messages are always flat"));
+        assert!(prompt.contains("Do not pass `--reply-to` in a DM"));
+    }
+
+    #[test]
     fn shared_base_prompt_teaches_single_command_mentions_and_preflight() {
         let prompt = include_str!("base_prompt.md");
         assert!(prompt.contains("--mention <hex-or-npub>"));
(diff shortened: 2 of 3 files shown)


alithia-dev (approved):
Independently verified this against current `main` (Buzz Desktop 0.5.5 source). The DM regression tests pass for both top-level messages and incoming messages that already carry thread tags, and the existing human-facing channel/thread tests still anchor replies to the root rather than the latest child. `cargo fmt --all -- --check` and `cargo clippy -p buzz-acp --all-targets -- -D warnings` pass. The full Windows run reached 659/670; the remaining 11 ACP subprocess/timing tests fail identically on unmodified `main`, so they are unrelated to this diff. This is the correct transport contract: flat DMs, root-relative channel/forum replies.

PR #4243, opened 2026-08-02
fix(acp): show mention_filter in startup config summary (#4228)

### LINKED ISSUE #4228: buzz-acp: startup config summary omits no_mention_filter, so logs can't confirm the flag took effect
**Describe the bug**

The one-line startup summary that buzz-acp logs (relay, pubkey, subscribe mode, dedup, respond_to, …) does not include the `no_mention_filter` state. A unit running `--subscribe mentions --no-mention-filter` logs exactly `subscribe=Mentions` — byte-identical to a mention-gated deployment.

Since the flag materially changes both the relay subscription (the `#p` filter is dropped) and the operator's cost model (every allowed-author message becomes a potential model turn), the summary line is the natural place to attest it — it's the only at-a-glance config attestation an operator gets from the journal.

`crates/buzz-acp/src/config.rs` — `summary()` at line 1111; the format string (line 1126) prints `subscribe={:?}` (the mode enum only). `no_mention_filter` appears nowhere in the line.

**Steps to reproduce**

1. Run buzz-acp with `--subscribe mentions --no-mention-filter`.
2. Read the startup line in the journal:

```
INFO buzz_acp: buzz-acp starting: relay=wss://… subscribe=Mentions dedup=Queue meh=Queue … respond_to=allowlist(1)
```

3. Compare with a unit running plain `--subscribe mentions`: the lines are indistinguishable.

**Expected behavior**

The summary includes the mention-filter state, e.g. `subscribe=Mentions no_mention_filter=true` (or `mention_filter=off`). To confirm the flag was active we currently have to run an end-to-end probe — publish an untagged event from an allowlisted author and watch for a turn — because there is no cheaper signal.

**Version and platform**

- Buzz version: buzz-acp from `ghcr.io/block/buzz:main` @ `sha256:13ddb4c12f4c0b8147cece717a57013a45fdb1610bd072d7e383ac79a070ddd9`; re-verified in source on `main` @ `ac4fa13` (2026-08-01)
- OS: Ubuntu 24.04, systemd unit

**Logs / additional context**

Related issues (searched open and closed): none found. I have a small PR ready for this (add the field to `summary()` plus a regression test) and will open it right after filing.

Chessing234:
## Summary
- Add `mention_filter=on|off` to the buzz-acp startup journal line so `--no-mention-filter` deployments are distinguishable from gated `subscribe=Mentions`.
- Regression tests pin default/on/off behavior and assert gated vs ungated summaries differ.
- Document the field in README, clap help, and `Config::summary` rustdoc.

Fixes #4228.

## Test plan
- [x] `cargo test -p buzz-acp test_summary_mention_filter`
- [ ] Confirm journal line shows `mention_filter=off` when running `buzz-acp --subscribe mentions --no-mention-filter`


Chessing234:

PR #4244, opened 2026-08-02
fix(git): allow + and @ in is_safe_refname (#4194)

### LINKED ISSUE #4194: Git: is_safe_refname rejects legal git ref names containing '+'

Chessing234:
## Summary
- Widen `is_safe_refname` to accept git-legal `+` and `@` characters while keeping structural guards (`..`, control chars, `refs/` prefix).
- Add manifest, hydrate, and `validate()` coverage for refs like `refs/heads/test/842+841-devnet` (observed blocking OriginTrail/dkg mirrors).
- Update the predicate doc comment to reflect the expanded alphabet.

Fixes #4194.

## Test plan
- [x] `cargo test -p buzz-relay safe_refnames`
- [x] `cargo test -p buzz-relay validate_accepts_git_legal_plus_refname`
- [ ] Push `git branch test/842+841-devnet && git push <buzz-remote>` and confirm HTTP 400 is gone

### DIFF
--- a/crates/buzz-relay/src/api/git/hydrate.rs
+++ b/crates/buzz-relay/src/api/git/hydrate.rs
@@ -488,6 +488,8 @@ mod tests {
         assert!(is_safe_refname("refs/heads/main"));
         assert!(is_safe_refname("refs/tags/v1.0.0"));
         assert!(is_safe_refname("refs/heads/feat/cas-publish"));
+        assert!(is_safe_refname("refs/heads/test/842+841-devnet"));
+        assert!(is_safe_refname("refs/heads/feature@workspace"));
         assert!(!is_safe_refname("refs/heads/../escape"));
         assert!(!is_safe_refname("HEAD"));
         assert!(!is_safe_refname("refs/heads/"));
--- a/crates/buzz-relay/src/api/git/manifest.rs
+++ b/crates/buzz-relay/src/api/git/manifest.rs
@@ -134,8 +134,8 @@ pub enum ManifestError {
 /// `api::git::hydrate`, before writing the ref to disk).
 ///
 /// Refuses traversal (`..`), null/newline/control chars, non-`refs/` prefixes,
-/// and leading/trailing/double slashes. Allowed alphabet:
-/// `[a-zA-Z0-9_./-]`.
+/// and leading/trailing/double slashes. Allowed alphabet is a conservative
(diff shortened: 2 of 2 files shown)


Chessing234:

Chessing234:
there's also #4348 which widens is_safe_refname plus is_emittable_ref/RefPattern for the same #4194 case. happy to close this one if that lands first, or fold any missing coverage in.

Chessing234:
closing in favor of #4348 — that one also widens RefPattern / is_emittable_ref for the same #4194 alphabet, which this tip didn't cover.

PR #4249, opened 2026-08-02
fix(buzz-audit): close object_id/detail field-boundary collision in compute_hash (#4173)

### LINKED ISSUE #4173: buzz-audit: field-boundary collision in compute_hash lets a detail edit survive verify_chain
## Summary

`crates/buzz-audit/src/hash.rs::compute_hash` concatenates its fields into the
SHA-256 preimage with no length prefixes or delimiters. Two of those fields are
adjacent and both variable-length:

```rust
match &entry.object_id {
    Some(id) => { hasher.update([1u8]); hasher.update(id.as_bytes()); }
    None => hasher.update([0u8]),
}
hasher.update(canonical_json(&entry.detail)?.as_bytes());
```

`object_id` is `Option<String>` (free text) and `detail` is an arbitrary
`serde_json::Value`, which may be a bare scalar. So the boundary between them
can be shifted, and two distinct entries can share one digest.

This defeats the property `entry.rs` documents for `detail`:

> *"Arbitrary JSON context. **Included in the hash** (serialized with sorted keys
> for determinism) so tampering with it is detectable."*

An entry can be substituted for a different one and `verify_chain` still passes.

## Reproduction

Both of these hash identically:

| | `object_id` | `detail` | concatenated tail |
|---|---|---|---|
| A | `"x"` | `12` | `x` ‖ `12` = `x12` |
| B | `"x1"` | `2` | `x1` ‖ `2` = `x12` |

Confirmed by reimplementing the exact field order (all other fields held equal —
`community_id`, `seq`, `created_at`, `action`, `actor_pubkey`, `prev_hash`):

```
A (object_id='x',  detail=12) = b69cc4a59d5feb1eea4cdb5ea3d65924025bca46250701e0f86e6f8e0b017abb
B (object_id='x1', detail=2 ) = b69cc4a59d5feb1eea4cdb5ea3d65924025bca46250701e0f86e6f8e0b017abb
COLLISION: True
```

A Rust test in the style of the existing `sensitive_to_each_field` should
reproduce it directly.

## Severity

Low as shipped, and we want to be precise about that rather than overstate it.
`object_id` is relay-resolved (event id hex, channel UUID, media sha256) and
`detail` is written by relay code, so we found no path where a client controls
both today. `NewAuditEntry` is deliberately not `Deserialize`, which helps.

It becomes live the first time an audit action logs a caller-supplied
`object_id` alongside a scalar `detail` — a natural future change that would
silently reintroduce forgeability with no test failing.

## Suggested fix

Make the encoding unambiguous. Any one of:

1. **Length-prefix the variable-length fields** — `update(len.to_be_bytes())`
   before each of `created_at`, `action`, `actor_pubkey`, `object_id`, and the
   canonical JSON. Most robust; invalidates existing chains.
2. **Move a fixed-width field between them** — e.g. hash `prev_hash` between
   `object_id` and `detail`. Cheapest structurally, still a chain-invalidating
   field-order change.
3. **Hash the canonical JSON of the whole entry** as one object rather than a
   field concatenation.

Any of these changes the digest, so it needs the same migration treatment as any
other change to the fixed field order.

## Context / disclosure

We adapted the hashing design (not the code) for a Python project, with
attribution per Apache-2.0, and found this while writing our own regression
tests. Our port happens not to be vulnerable — we place the fixed-width 32-byte
`prev_hash` between the sequence number and the variable-length payload, and our
payload is always a JSON object, so its canonical form always begins with `{`.
That was luck of field ordering rather than design, which is why we went looking.

Happy to open a PR with a failing test if useful.

iroiro147:
## Summary

Fixes #4173 — takes the issue author's suggested fix #2 (move a fixed-width field between the two variable-length fields).

`compute_hash` in `crates/buzz-audit/src/hash.rs` hashed `object_id` (`Option<String>`, free text) immediately followed by `canonical_json(detail)` (arbitrary `serde_json::Value`, possibly a bare scalar), **unframed**. Any byte shift across their boundary left the preimage identical:

| | `object_id` | `detail` | concatenated tail |
|---|---|---|---|
| A | `"x"` | `12` | `x` ‖ `12` = `x12` |
| B | `"x1"` | `2` | `x1` ‖ `2` = `x12` |

Both entries hashed to the same digest. A detail edit survived `verify_chain` — the property `entry.rs` documents ("*Arbitrary JSON context. **Included in the hash** (serialized with sorted keys for determinism) so tampering with it is detectable.*") was violated.

## Fix

Move the fixed-width `prev_hash` into the preimage between `object_id` and `detail`:

```
…community_id, seq, created_at, action,
actor_pubkey, object_id, prev_hash, canonical_json(detail)
```

The 32-byte `prev_hash` can't slide, so no byte shift across the (now separated) `object_id`/`detail` boundary can preserve the full preimage.

This is a **chain-invalidating field-order change**, by design — the same migration treatment any change to `compute_hash`'s fixed field order requires. Existing chains must be re-hashed, or the deployment must wipe and rebuild its audit chain. Maintenance of existing chains is the relay operator's responsibility, as documented on the issue.

Picked fix #2 over length-prefixing because it is the smallest structural change with no framing format to spec, parse, or bug on, and it composes with the existing `Some(vec![tag])` / `None` presence tags unchanged. Length-prefixing every variable-length field is a strictly stronger option if a future format revision is planned.

## Why this is safe

- Only `crates/buzz-audit/src/hash.rs` changes — 44 insertions, 1 deletion.
- All other `compute_hash` callers (`service`, `lib`, `entry`, `error`) pass entries, not preimages — they don't depend on field order.
- `prev_hash` was already part of the hash; only its **position** changed.

## Testing

New regression test `hash::tests::object_id_detail_boundary_is_unambiguous` pins both the exact repro pair from the issue (`object_id="x"`/`detail=12` versus `object_id="x1"`/`detail=2`) and a second string-skewed pair (`"ab"`+`"c"` versus `"a"`+`"bc"`) that exercises the same boundary at different byte lengths.

```
$ cargo test -p buzz-audit
running 20 tests
test hash::tests::object_id_detail_boundary_is_unambiguous ... ok
test hash::tests::sensitive_to_each_field ... ok
test hash::tests::canonical_json_key_order_is_stable ... ok
... (14/14 pass, 6 ignore — require Postgres)
test result: ok. 14 passed; 0 failed; 6 ignored; 0 measured; 0 filtered out
```

### DIFF
--- a/crates/buzz-audit/src/hash.rs
+++ b/crates/buzz-audit/src/hash.rs
@@ -64,11 +64,20 @@ pub fn compute_hash(entry: &AuditEntry) -> Result<[u8; 32], AuditError> {
         }
         None => hasher.update([0u8]),
     }
-    hasher.update(canonical_json(&entry.detail)?.as_bytes());
+    // Fixed-width `prev_hash` sits between the two variable-length fields
+    // (`object_id`, `canonical_json(detail)`) so no byte shift across their
+    // boundary can leave the preimage unchanged. Both were previously adjacent:
+    // `hasher.update(object_id)` then `hasher.update(canonical_json(detail))`,
+    // and any pair `(oid="x", detail=12)` / `(oid="x1", detail=2)` hashed
+    // identically — a detail edit survived `verify_chain` (#4173). Moving
+    // `prev_hash` here makes the encoding unambiguous without length-prefix
+    // framing. This is a field-order change to `compute_hash` and invalidates
+    // existing chains by design, like any other.
     match &entry.prev_hash {
         Some(h) => hasher.update(h),
         None => hasher.update(GENESIS_HASH),
     }
+    hasher.update(canonical_json(&entry.detail)?.as_bytes());
     Ok(hasher.finalize().into())
 }
 
@@ -251,6 +260,40 @@ mod tests {
         let mut e = base.clone();
         e.prev_hash = Some(vec![0xff; 32]);
         assert_ne!(h0, compute_hash(&e).unwrap());
+    }
+
+    /// Issue #4173: `object_id` and `detail` are both variable-length and
+    /// were adjacent in the preimage, so a byte shift across their boundary
+    /// could keep the concatenated tail identical while the entry differs.
+    /// Both `("x", 12)` and `("x1", 2)` hashed to the same chain tail — an
+    /// entry could be substituted for a different one and `verify_chain`
+    /// still passed. The fixed-width `prev_hash` now sits between them, so no
+    /// such shift can keep the full preimage unchanged.
+    #[test]
+    fn object_id_detail_boundary_is_unambiguous() {
+        // Bare-scalar `detail` — the case that made the boundary shift live:
+        // a JSON scalar has no framing byte of its own.
+        let mut a = sample_entry();
+        a.object_id = Some("x".into());
+        a.detail = serde_json::json!(12);
+
+        let mut b = sample_entry();
+        b.object_id = Some("x1".into());
+        b.detail = serde_json::json!(2);
+
+        assert_ne!(compute_hash(&a).unwrap(), compute_hash(&b).unwrap());
+
+        // A second, string-skewed pair to exercise the same boundary with
+        // different byte lengths on each side.
+        let mut c = sample_entry();
+        c.object_id = Some("ab".into());
+        c.detail = serde_json::json!("c");
+
+        let mut d = sample_entry();
+        d.object_id = Some("a".into());
+        d.detail = serde_json::json!("bc");
+
+        assert_ne!(compute_hash(&c).unwrap(), compute_hash(&d).unwrap());
     }
 
     #[test]


alanshurafa:
Cross-linking #4216, which fixes the same collision by versioning the encoding instead (rows verify under their stored version; new rows are length-framed; no data migration). The two approaches could genuinely compose — this reorder could simply be the next encoding version — so it may be worth a maintainer call on the shape.

One cost worth surfacing either way: after a field-order change, every pre-existing row fails `verify_chain` with `HashMismatch` on untampered data, which is the condition #2638 dug this crate out of ("a genuinely forged row is indistinguishable from the permanent baseline failure, so `HashMismatch` carries no signal"). Without a shipped re-hash path, the upgrade converts existing audit history from verifiable to permanently failing.

Also: the reorder closes the reported boundary, but the preimage keeps other unframed variable-length neighbours — `created_at` (after `to_storage_precision`, `to_rfc3339` emits 0/3/6 sub-second digits, so it's variable-width) next to `action`, and `actor_pubkey` next to `object_id` separated only by a constant presence-tag byte, which is not length framing (pk is `Vec<u8>`; 32 bytes by convention only). Same shape #4173 warned about — worth documenting at minimum.

iroiro147:
@alanshurafa is right on both counts, and the cost he surfaces is the deciding factor.

Conceding the two points explicitly so the maintainer call is on the real question:

1. **`verify_chain` regression.** My field-reorder alone converts every pre-existing row's `verify_chain` to a permanent `HashMismatch` on untampered data — precisely the failure mode #2638 dug this crate out of ("a genuinely forged row is indistinguishable from the permanent baseline failure, so `HashMismatch` carries no signal"). A reorder without a re-hash/version path silently degrades existing audit history from verifiable to permanently failing. That alone disqualifies the bare reorder as a standalone fix.

2. **Unframed variable-width neighbours.** The reorder closes the reported `object_id`/`detail` boundary, but the preimage still keeps other unframed variable-length neighbours (`created_at` → `to_storage_precision` → `to_rfc3339` emits 0/3/6 sub-second digits; `actor_pubkey`/`object_id` separated only by a constant presence-tag byte, pk being `Vec<u8>`, 32 by convention). Same shape as #4173.

So the honest read: this reorder should not land as-is. **#4216's versioned-length framing is the strictly better shape** — rows verify under their stored version, new rows are length-framed, no migration, and it fixes the *whole* preimage family rather than one boundary.

The compose option: if the maintainers want the reorder *too*, it becomes the **v2 layout** inside #4216's framing (v1 = legacy rows verified as-is, v2 = length-framed + reordered). I'm happy to either (a) rebase my branch onto #4216 to make the reorder be its v2 encoding, or (b) close this PR in favour of #4216 and fold anything useful there. **Which shape do the maintainers want?** I don't want to keep both open competing if there's a preference.

PR #4253, opened 2026-08-02
fix(relay): widen freshness window for NIP-17 gift wraps (kind 1059)

### LINKED ISSUE #4192: Relay rejects spec-compliant NIP-17 DMs: gift-wrap (kind 1059) timestamps fail the freshness window

iroiro147:
## What

NIP-59 instructs gift-wrap (kind 1059) clients to randomize `created_at` into the past — canonically up to 2 days — to thwart time-analysis attacks. Applying a tight freshness window to the outer wrap rejects every spec-following NIP-17 DM; the issue author reports 100% rejection from `nostr-tools` `nip17.wrapEvent()` (which backdates by default).

This patch keeps the default ±15-minute window for every kind *except* `KIND_GIFT_WRAP`, which now uses a dedicated **2 days + 15 minutes** window. The wrap`s timestamp is deliberately meaningless for ordering/retention (the inner seal/rumor timestamps are end-to-end encrypted and invisible to the relay), so a wider acceptance window doesn't weaken security. We still bound it — implausibly stale wraps are still rejected; only spec-compliant backdated ones become acceptable.

## Why this approach

Per the issue`s own suggested fix (#1 in the author body): either exempt kind 1059 entirely, or widen its window to "two days plus skew". The author leans toward the latter as a safety knob, and I agree — full exemption removes an abuse-mitigation knob operators may rely on, while a wider-but-bounded window covers the entire NIP-59 spec surface.

## Test plan

- `cargo check -p buzz-relay` clean
- `cargo test -p buzz-relay --lib ingest` — 156/156 passed (no regression on existing kind-1059 or freshness-adjacent tests)

## Linked issue

Closes #4192

### DIFF
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -1973,10 +1973,25 @@ async fn ingest_event_inner(
     }
     let event = std::sync::Arc::try_unwrap(event).unwrap_or_else(|arc| (*arc).clone());
 
-    const MAX_TIMESTAMP_DRIFT_SECS: i64 = 900; // ±15 minutes
+    const MAX_TIMESTAMP_DRIFT_SECS: i64 = 900; // ±15 minutes for most kinds
+    // NIP-59 instructs gift-wrap (kind 1059) clients to randomize `created_at`
+    // into the past — canonically up to two days — to thwart time-analysis
+    // attacks. Applying the tight ±15 min freshness window to the outer wrap
+    // rejects every spec-following NIP-17 DM. The wrap's timestamp is
+    // deliberately meaningless for ordering or retention, and the inner
+    // seal/rumor timestamps are end-to-end encrypted, so a wider window is
+    // safe here. We still bound it (2 days + 15 min skew) to reject
+    // implausibly stale wraps rather than skipping the check altogether.
+    // See block/buzz#4192.
+    const MAX_GIFT_WRAP_TIMESTAMP_DRIFT_SECS: i64 = 2 * 24 * 60 * 60 + 900;
     let now = chrono::Utc::now().timestamp();
     let event_ts = event.created_at.as_secs() as i64;
-    if (event_ts - now).abs() > MAX_TIMESTAMP_DRIFT_SECS {
+    let max_drift_secs = if kind_u32 == KIND_GIFT_WRAP {
+        MAX_GIFT_WRAP_TIMESTAMP_DRIFT_SECS
+    } else {
+        MAX_TIMESTAMP_DRIFT_SECS
+    };
+    if (event_ts - now).abs() > max_drift_secs {
         return Err(IngestError::Rejected(
             "invalid: event timestamp too far from server time".into(),
         ));
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -1846,6 +1846,36 @@ async fn author_type_label(
     }
 }
 
+// Freshness window: deliberately asymmetric.
+
+//
+// NIP-59 instructs gift-wrap (kind 1059) clients to randomize `created_at`
+// into the past — canonically up to two days — to thwart time-analysis
+// attacks. The wrap's timestamp is meaningless for ordering/retention (the
+// inner seal/rumor timestamps are end-to-end encrypted), so gift wrap gets
+// a wide PAST bound (2 days). Most kinds keep the tight ±15 min window.
+// NIP-59 never randomizes into the FUTURE, so the future bound stays tight
+// (15 min) for all kinds: an event pinned far into the future is almost
+// always a client bug or an attempt to float to the top of time-ordered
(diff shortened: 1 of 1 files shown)


Zigoljube:
Verified this branch against the deployment where #4192 was originally hit (Buzz relay on macOS, tailnet, mixed Nostr clients):

- **Targeted tests**: `cargo test -p buzz-relay --lib ingest` on this branch — **156 passed, 0 failed**, matching the PR description. (Full `--lib` runs show 9 failing media/admin tests on this branch *and* 10 on its merge-base — pre-existing/environment-dependent, not introduced here.)
- **Repro alignment**: our failing case was `nostr-tools` `nip17.wrapEvent()` output (wrap `created_at` randomized up to 2 days into the past) rejected with `event timestamp too far from server time`. Those wraps fall inside the new `MAX_GIFT_WRAP_TIMESTAMP_DRIFT_SECS` window, and the re-stamped-to-now wraps we used as a workaround stay accepted — the change addresses exactly the observed failure with no behavior change for other kinds.

One optional tightening for the maintainers to weigh: the check is symmetric (`.abs()`), so it also admits wraps stamped up to 2 days in the **future**. NIP-59 only randomizes into the past; an asymmetric window (2d past / 15min future) would keep the replay-surface smaller at no interop cost. Not blocking — the current form already fixes the real problem.

iroiro147:
Folded your suggestion in `5cbe8d29` (follow-up commit, not an amend). The freshness window is now kind-aware and asymmetric:

- **Gift wrap (kind 1059):** 2-day **past** allowance, tight 15-min **future** bound — matching your read that NIP-59 only randomizes `created_at` into the past. The future side never needed the widening.
- **All other kinds:** unchanged 15 min both ways.

Replaced the symmetric `.abs()` bound in `ingest_event_inner` with a small pure helper `timestamp_within_freshness_window(kind, event_ts, now)` and added three boundary tests:

- `non_gift_wrap_window_is_symmetric_15min` — edge/just-over on both sides.
- `gift_wrap_window_allows_2d_past` — exact 2d-past edge accepted, 2d+1s rejected.
- `gift_wrap_future_bound_stays_tight` — 15-min future accepted, 901s and full 2d future rejected for gift wrap.

`cargo test -p buzz-relay --lib handlers::ingest` → 158 passed, 0 failed; `cargo check` warning-free; `cargo fmt` clean. Thanks for the sharp catch — the asymmetric shape is strictly better than the `.abs()` widening.

iroiro147:
Closing this in favour of #4324, which addresses the same issue (#4192) with a 37-line change to the same file.

I had two open PRs against #4192 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

Issue #4195, opened 2026-08-01
git-credential-nostr: first push of any pack > 1 MiB fails (401 challenge on a streamed, non-replayable POST)


Zigoljube:
# git-credential-nostr: first push of any pack > 1 MiB fails (401 challenge on a streamed, non-replayable POST)

## Summary

Out of the box, pushing more than ~1 MiB to Buzz git dies with:

```
error: RPC failed; HTTP 401 curl 56 The requested URL returned error: 401
send-pack: unexpected disconnect while reading sideband packet
fatal: the remote end hung up unexpectedly
Everything up-to-date        ← misleading: nothing was pushed
```

## Mechanism

Three correct-in-isolation behaviors compose into a guaranteed failure:

1. `git-credential-nostr` returns `ephemeral=true`, so git caches nothing and sends the `git-receive-pack` POST **unauthenticated first**, expecting to retry after the 401 challenge.
2. The Buzz server (correctly) answers the unauthenticated POST with 401.
3. git can only replay a challenged POST if the body is buffered; packs larger than `http.postBuffer` (default **1 MiB**) are streamed chunked and cannot be replayed → hard failure.

So the failure threshold is exactly `http.postBuffer`, and the very first real-world push (any repo with more than trivial history) hits it. Small pushes work, which makes this look like flaky infrastructure rather than a deterministic composition bug.

## Workaround (verified)

```bash
git config http.postBuffer 524288000   # buffer the pack so the 401 retry can replay it
```

After this a 90 MB pack pushes fine (the ±60s NIP-98 window is not a problem — the token is minted at challenge time, after the buffered body is ready).

## Suggested fixes (in increasing order of effort)

1. **Document it.** A "Large pushes" section in `crates/git-credential-nostr/README.md` with the `http.postBuffer` line would eliminate the trap for near-zero cost. The README's troubleshooting table already covers subtler failures than this one.
2. **Pre-emptive auth.** The helper already knows the server's auth scheme after the first `info/refs` challenge of the session. Returning a credential git can send pre-emptively on the subsequent POST (i.e., not marking it `ephemeral` within a single git invocation, scoped by the ±60s validity) avoids the challenge round-trip entirely for the push POST.
3. **Server-side `Expect: 100-continue` handling** so the 401 lands before the body is transmitted; git/curl then retries with auth without needing a replayable body. (Most invasive; 1–2 make it unnecessary.)

## Environment

[...]

PR #4259, opened 2026-08-02
docs(git-credential-nostr): document large-push (>1 MiB) 401 trap (#4195)

iroiro147:
## What

First push of any pack larger than ~1 MiB to Buzz git dies with a misleading `HTTP 401` → `send-pack: unexpected disconnect` → `Everything up-to-date` (nothing was actually pushed). Three correct-in-isolation behaviors compose into a guaranteed failure:

1. `git-credential-nostr` returns `ephemeral=true`, so git sends the `git-receive-pack` POST **unauthenticated first**, expecting to retry after the 401 challenge.
2. The Buzz server (correctly) answers the unauthenticated POST with 401.
3. Git can only replay a challenged POST if the body is buffered; packs larger than `http.postBuffer` (default **1 MiB**) are streamed chunked and cannot be replayed → hard failure.

This PR implements the issue's preferred fix (option 1, doc-only): a new **Large pushes** section in `crates/git-credential-nostr/README.md` explaining the composition trap and the verified workaround:

```bash
git config --global http.postBuffer 524288000   # 500 MiB
```

Also adds a Troubleshooting table row pointing at the new section.

Options 2 (preemptive auth within a single git invocation) and 3 (server-side `Expect: 100-continue`) are noted in the issue as follow-up if the doc proves insufficient; both are larger surface changes and the workaround is a one-line per-user config.

## Why doc-only

The failure is a composition of three correct behaviors — none of the three is a bug in isolation. The helper's `ephemeral=true` is correct (NIP-98 tokens are short-lived and must not be cached across invocations). The server's 401 on unauthenticated POST is correct (that's the challenge-response handshake). Git's streaming behavior is correct for large bodies. The doc surfaces the one knob (`http.postBuffer`) that makes the composition work, at near-zero risk.

## Verified against the issue

- Repro: 101 MB repo, first push fails at ~1 MiB threshold — matches the `http.postBuffer` default exactly.
- Workaround verified by issue author: 90 MB pack pushes fine after `git config http.postBuffer 524288000`.
- The ±60s NIP-98 window is not a problem: the helper mints the token at challenge time, after the buffered body is ready.

## Linked issue

Refs #4195

### DIFF
--- a/crates/git-credential-nostr/README.md
+++ b/crates/git-credential-nostr/README.md
@@ -57,6 +57,35 @@ git ──stdin──▶ git-credential-nostr ──stdout──▶ git
               (NIP-98 HTTP Auth)
 ```
 
+## Large pushes (> 1 MiB)
+
+Pushing a pack larger than git's `http.postBuffer` (default **1 MiB**) fails
+with a misleading `HTTP 401` followed by `send-pack: unexpected disconnect`
+and `Everything up-to-date` — even though nothing was pushed. Three
+correct-in-isolation behaviors compose into the failure:
+
+1. This helper returns `ephemeral=true`, so git caches nothing and sends the
+   `git-receive-pack` POST **unauthenticated first**, expecting to retry
+   after the 401 challenge.
+2. The Buzz server (correctly) answers the unauthenticated POST with 401.
+3. Git can only replay a challenged POST if the body is buffered; packs
+   larger than `http.postBuffer` are streamed chunked and cannot be
+   replayed → hard failure.
+
+**Workaround** — buffer the pack so the 401 retry can replay it:
+
+```bash
+git config --global http.postBuffer 524288000   # 500 MiB
+```
+
+The ±60 s NIP-98 window is not a problem: the helper mints the token at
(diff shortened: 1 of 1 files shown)


Zigoljube:
Reviewed against the deployment where #4195 was hit: the mechanism description matches what we observed exactly (unauthenticated first POST by design of `ephemeral=true`, 401 challenge, unreplayable streamed body past `http.postBuffer`), and the workaround section documents precisely the fix we verified live — after `http.postBuffer 524288000`, a ~90 MB pack pushes cleanly, and the note about the NIP-98 token being minted at challenge time (so the ±60s window is not a factor) matches our measurements. The troubleshooting-table row with the three-symptom signature (`401` + `unexpected disconnect` + misleading `Everything up-to-date`) is the part that would have saved us an hour. LGTM as a doc fix; options 2/3 from the issue remain open for a future behavioral fix.

iroiro147:
Closing this in favour of #4325, which addresses the same issue (#4195) with a tighter README change.

I had two open PRs against #4195 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

PR #4270, opened 2026-08-02
fix(desktop): send complete respond_to wire patch on every allowlist transition

iroiro147:
## Summary

Refs #2501 (partial: dialog write path only — the v0.5.2 first-save repro and the #4115 cascade path are tracked separately).

The Desktop agent-instance edit dialog silently dropped the allowlist payload on two reachable mode transitions, so the `buzz-acp` harness kept starting with `respond_to=owner-only` regardless of what the UI showed. This made it impossible to share a custom agent with other community members through the UI — only a hand edit of `managed-agents.json` worked.

## Root cause

The dialog gated the allowlist payload on `submitMode === "allowlist" && listChanged`, conflating "mode changed" with "payload changed":

1. **Mode flip → allowlist with an unchanged-but-preexisting list** (e.g. anyone → allowlist where the list was populated in an earlier session): the mode flips on the wire but the harness never receives the full `{mode, allowlist}` tuple atomically, so the record stays in a state the harness reads as `owner-only`.
2. The reporter also observed the visible desync — the UI wrote `definition_respond_to` (the definition snapshot projection) but the harness reads from the record-level `respond_to` + `respond_to_allowlist` the dialog never sent. (`update_managed_agent` merges `undefined` allowlist as "leave unchanged", so the two states stayed out of sync permanently.)

## Fix

Extract a pure `computeRespondToWirePatch` helper in `personaRuntimeModel.ts` that owns the full contract:

- `respondTo` on the wire only when the mode changed.
- `respondToAllowlist` on the wire whenever `submitMode === "allowlist"` AND (`modeChanged` OR `listChanged`) — so a mode flip to allowlist always carries the complete `{mode, allowlist}` tuple the harness needs, even when the list itself happens to match.
- Every other shape (flip away from allowlist, no-op, list edit under allowlist) produces exactly the minimal correct patch.

The dialog now spreads the helper's return into the submit input — there is one canonical wire-shape producer instead of inline conditional logic.

## Reproduction (from the issue)

Under Buzz Desktop v0.4.23 (macOS, Apple Silicon), custom agent, runtime `claude`:
1. Open agent config → "Who can talk to this agent" → **Allowlist** → add a member pubkey → Save. Works on first save.
2. Re-open → switch to **Anyone** → Save → mode flips.
3. Re-open → switch back to **Allowlist** → the list is already populated from step 1 → Save.
4. Have the member @mention the agent — no response; harness log shows `respond_to=owner-only`.

Manual `managed-agents.json` edit of `respond_to` + `respond_to_allowlist` fixes the agent permanently — the file is the source of truth; the dialog simply wasn't writing it on transition 3.

## Testing

- New pure-module test `desktop/src/features/agents/ui/agentInstanceEditRespondToWire.test.mjs` — 6 tests covering the full transition matrix:
  - owner-only → allowlist with fresh list (mode + list)
  - allowlist → allowlist with edited list (list only)
  - allowlist → anyone (mode only)
  - owner-only → owner-only no-op (nothing)
  - anyone → allowlist with pre-existing list (mode + list, the previously broken path)
  - allowlist → owner-only (mode only)
- Full webview suite: **3912 passing, 0 fail**.
- `pnpm exec tsc --noEmit`: clean.
- The helper uses the shared `RespondToMode` type end-to-end so the wire shape is type-checked against `UpdateManagedAgentInput`.

## Notes for reviewers

- No schema or record migration — the fix is purely at the wire-shape producer layer; existing records self-heal on the next save-through-dialog.
- The Rust update path is untouched and continues to enforce the allowlist-non-empty invariant; the UI's `editValidity_allowlistWithEmptyList_blocksSave` gate already prevents the crash-loop shape at the dialog.

### DIFF
--- a/desktop/src/features/agents/ui/AgentInstanceEditDialog.tsx
+++ b/desktop/src/features/agents/ui/AgentInstanceEditDialog.tsx
@@ -49,6 +49,7 @@ import {
 } from "./relayMeshModelPicker";
 import {
   computeEditAgentFormValidity,
+  computeRespondToWirePatch,
   envVarsEqual,
   isEditAgentProviderSaveValid,
   resolveAgentCommandUpdate,
@@ -711,18 +712,18 @@ export function AgentInstanceEditDialog({
         envVars: envVarsEqual(submitEnvVars, agent.envVars)
           ? undefined
           : submitEnvVars,
-        respondTo: respondTo !== agent.respondTo ? respondTo : undefined,
-        // The allowlist is preserved across mode toggles in local UI state
-        // (so a user can flip away from allowlist and back without losing
-        // their entries), but we only send it on the wire when (a) it
-        // actually changed, AND (b) the saved mode will need it. Sending
-        // an allowlist while switching to a non-allowlist mode would be
-        // harmless server-side, but it's noise in the persisted record.
-        respondToAllowlist:
-          respondTo === "allowlist" &&
-          respondToAllowlist.join(",") !== agent.respondToAllowlist.join(",")
-            ? respondToAllowlist
-            : undefined,
+        // The respond-to pair is computed via the canonical wire-shape helper
+        // so every transition shape — mode flip, payload edit, or both —
+        // produces the exact patch the harness layer consumes on the local
+        // record (#2501). The helper owns the "flip to allowlist must re-send
+        // the payload even when unchanged" contract; see
+        // `computeRespondToWirePatch` for the full semantics.
+        ...computeRespondToWirePatch({
+          currentMode: agent.respondTo,
+          currentAllowlist: agent.respondToAllowlist,
+          submitMode: respondTo,
+          submitAllowlist: respondToAllowlist,
+        }),
       };
 
       const result = await updateMutation.mutateAsync(input);
--- a/desktop/src/features/agents/ui/agentInstanceEditRespondToWire.test.mjs
+++ b/desktop/src/features/agents/ui/agentInstanceEditRespondToWire.test.mjs
@@ -1,0 +1,92 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import { computeRespondToWirePatch } from "./personaRuntimeModel.ts";
+
+// ── Submit-shape wire contract for the agent-instance edit dialog ────────────
+//
+// Issue #2501: the Desktop edit dialog's allowlist wire condition conflates
+// "mode changed" with "payload changed", so several mode/allowlist transition
+// shapes silently drop the payload the harness layer actually needs. The
+// harness reads from `respond_to` + `respond_to_allowlist` on the local record
+// ( not the definition snapshot projection), and `update_managed_agent` merges
+// `undefined` allowlist as "leave unchanged" — so a mode flip WITHOUT a
+// re-sent payload can leave the record in an unreachable or crash-looping
+// state. These tests pin the exact wire patch each transition must produce.
+
+const PUBKEY_A = "a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2";
+const PUBKEY_B = "b1b2c3d4e5f60718293a4b5c6d7e8f90a1b2c3d4e5f60718293a4b5c6d7e8f90a1b2";
+
+test("owner-only → allowlist with fresh list sends mode AND allowlist", () => {
+  // The reporter's primary flow: pick Allowlist, add a pubkey, Save.
+  const patch = computeRespondToWirePatch({
+    currentMode: "owner-only",
+    currentAllowlist: [],
+    submitMode: "allowlist",
+    submitAllowlist: [PUBKEY_A],
+  });
+  assert.equal(patch.respondTo, "allowlist");
+  assert.deepEqual(patch.respondToAllowlist, [PUBKEY_A]);
+});
+
+test("allowlist → allowlist editing the list sends updated allowlist", () => {
+  // Mode unchanged, list edited — the payload must go on the wire.
+  const patch = computeRespondToWirePatch({
+    currentMode: "allowlist",
+    currentAllowlist: [PUBKEY_A],
+    submitMode: "allowlist",
+    submitAllowlist: [PUBKEY_B],
+  });
+  assert.equal(patch.respondTo, undefined);
+  assert.deepEqual(patch.respondToAllowlist, [PUBKEY_B]);
+});
+
+test("allowlist → anyone omits allowlist (record keeps stale list by design)", () => {
+  // Flipping away from allowlist: the Rust update preserves the persisted
+  // list across mode toggles, so we don't send it — but we must send the mode.
+  const patch = computeRespondToWirePatch({
+    currentMode: "allowlist",
+    currentAllowlist: [PUBKEY_A],
+    submitMode: "anyone",
+    submitAllowlist: [PUBKEY_A],
+  });
+  assert.equal(patch.respondTo, "anyone");
+  assert.equal(patch.respondToAllowlist, undefined);
+});
+
+test("owner-only → owner-only (no-op) sends nothing", () => {
+  const patch = computeRespondToWirePatch({
+    currentMode: "owner-only",
+    currentAllowlist: [],
+    submitMode: "owner-only",
+    submitAllowlist: [],
+  });
+  assert.equal(patch.respondTo, undefined);
+  assert.equal(patch.respondToAllowlist, undefined);
+});
+
+test("anyone → allowlist with pre-existing list sends mode AND list", () => {
+  // Mode flip to allowlist where the list was already populated by an earlier
+  // session (the "Anyone → Allowlist, list already there" case): the payload
+  // must be re-sent so the harness applies the full {mode, list} atomically.
+  const patch = computeRespondToWirePatch({
+    currentMode: "anyone",
+    currentAllowlist: [PUBKEY_A],
+    submitMode: "allowlist",
+    submitAllowlist: [PUBKEY_A],
+  });
+  assert.equal(patch.respondTo, "allowlist");
+  assert.deepEqual(patch.respondToAllowlist, [PUBKEY_A]);
+});
+
+test("allowlist → owner-only omits allowlist", () => {
+  // Flipping away: same as the allowlist → anyone case — mode only.
+  const patch = computeRespondToWirePatch({
+    currentMode: "allowlist",
+    currentAllowlist: [PUBKEY_A],
+    submitMode: "owner-only",
+    submitAllowlist: [PUBKEY_A],
+  });
+  assert.equal(patch.respondTo, "owner-only");
+  assert.equal(patch.respondToAllowlist, undefined);
+});
--- a/desktop/src/features/agents/ui/personaRuntimeModel.ts
+++ b/desktop/src/features/agents/ui/personaRuntimeModel.ts
@@ -1,3 +1,5 @@
+import type { RespondToMode } from "@/shared/api/types";
+
 /** Runtime provider-capability tri-state used by the submit path. */
 export type ProviderRuntimeCapability = "capable" | "locked" | "unknown";
 
@@ -300,4 +302,73 @@ export function envVarsEqual(
     aKeys.length === Object.keys(b).length &&
     aKeys.every((key) => a[key] === b[key])
   );
+}
+
+/**
+ * Inputs for {@link computeRespondToWirePatch} — the persisted record shape
+ * alongside the dialog's submitted shape, both pre-derived primitives.
+ */
+export interface RespondToWirePatchInput {
+  /** The mode currently persisted on the agent record. */
+  currentMode: RespondToMode;
+  /** The allowlist currently persisted on the agent record. */
+  currentAllowlist: string[];
+  /** The mode the user is submitting from the dialog. */
+  submitMode: RespondToMode;
+  /** The allowlist the user is submitting from the dialog. */
+  submitAllowlist: string[];
+}
+
+/**
+ * The wire patch the edit dialog must send for the respond-to pair.
+ *
+ * - `respondTo` is `undefined` when the mode is unchanged (the Rust update
+ *   merges `undefined` as "leave unchanged").
+ * - `respondToAllowlist` is `undefined` when the payload need not be sent
+ *   (mode is not allowlist, OR the list is unchanged and the mode is already
+ *   allowlist).
+ */
+export interface RespondToWirePatch {
+  respondTo?: RespondToMode;
+  respondToAllowlist?: string[];
+}
+
+/**
+ * Compute the exact `{respondTo, respondToAllowlist}` wire patch the agent
+ * instance edit dialog must send for a given mode/allowlist transition.
+ *
+ * The dialog previously gated the allowlist payload on `submitMode ===
+ * "allowlist"` AND the list differing from the persisted record — which
+ * silently dropped the payload in two reachable shapes (#2501):
+ *
+ *   1. mode flip to allowlist with a list that happens to match the record
+ *      (e.g. anyone → allowlist where the allowlist was populated in an
+ *      earlier session), so the harness never receives the full
+ *      `{mode, allowlist}` tuple atomically and the agent stays unreachable;
+ *   2. mode already allowlist with an edited list — covered, but only because
+ *      the mode gate happened to coincide.
+ *
+ * The backend update merges `respondTo: undefined` as "mode unchanged" and
+ * `respondToAllowlist: undefined` as "payload unchanged", so the wire patch
+ * carries exactly the fields that must change on the record.
+ */
+export function computeRespondToWirePatch(
+  input: RespondToWirePatchInput,
+): RespondToWirePatch {
+  const modeChanged = input.submitMode !== input.currentMode;
+  const listChanged =
+    input.submitAllowlist.join(",") !== input.currentAllowlist.join(",");
+
+  const respondTo = modeChanged ? input.submitMode : undefined;
+
+  // Send the allowlist payload when the submitted mode needs it AND either the
+  // mode or the payload actually changed. A mode flip TO allowlist with an
+  // unchanged-but-nonempty list must still re-send the payload so the harness
+  // applies {mode, allowlist} atomically on the local record.
+  const respondToAllowlist =
+    input.submitMode === "allowlist" && (modeChanged || listChanged)
+      ? [...input.submitAllowlist]
+      : undefined;
+
+  return { respondTo, respondToAllowlist };
 }


georgerous:
Read the diff — `computeRespondToWirePatch` is the right shape and the transition matrix is complete; the `modeChanged || listChanged` gate under `submitMode === "allowlist"` correctly re-sends the payload on the `anyone → allowlist` flip with an unchanged-but-preexisting list that was dropping it. Thanks for pulling it out as a pure, tested helper. Two things for reviewers before this carries `Fixes #2501`.

**1. The "Rust path enforces the allowlist-non-empty invariant" note holds for *this* path, but not for the #4115 cascade path.**

Your reviewer note is accurate for `update_managed_agent`, and the dialog's `editValidity_allowlistWithEmptyList_blocksSave` gate means `computeRespondToWirePatch` never actually emits `allowlist` + `[]` from here — the `[...submitAllowlist]` branch can only fire with a non-empty list because Save is blocked otherwise. Good.

But #4115 (your persona → instance cascade) reaches the record through a *different* Rust function, `propagate_persona_behavior`, whose inherit branch copies the definition allowlist verbatim with no non-empty guard:

```rust
record.respond_to = match persona.respond_to.as_deref() {
    Some(wire) => RespondTo::parse_wire(wire)?,
    None => RespondTo::default(),
};
record.respond_to_allowlist = persona.respond_to_allowlist.clone();
```

So a definition sitting in `allowlist` + `[]` cascades the exact crash-shape onto every inheriting instance — the state `resolve_mint_behavioral_defaults` / `apply_persona_behavior` reject at mint. Neither the dialog gate nor #4270 protects that path; only the mint-guard skip I asked for on #4115 does. If both PRs land, the invariant holds on the dialog write path but is still violable on the cascade write path — worth stating so #2501 isn't closed with that seam open.

**2. Does this close MaxWynnDev's repro, or only the v0.4.23 transition-3 flow?**

Your reproduction (v0.4.23, transition 3) is exactly what the helper fixes. But the other #2501 repro (MaxWynnDev, v0.5.2) is *"the list never persists at all, even on first save"* — and on first save this dialog already sent the allowlist under the old gate (`submitMode === "allowlist" && listChanged`, and `[] → [A]` is a list change), so #4270 changes nothing on that path. Either that repro hits a different write path (the persona person-picker / the profile panel #2505 touches) or a backend persist bug — worth confirming it's actually closed before this carries `Fixes #2501`, so we don't resolve the issue with the first-save case still live.

Happy to build #4270 + #4115 together and run the end-to-end matrix (fresh allowlist; `anyone → allowlist` with unchanged list; mixed inheriting/pinned fleet via a persona edit; restart + re-mint) once the #4115 skip is in.

iroiro147:
@georgerous this is an excellent review and I owe you an apology for the slow reply — both points hold.

**On 1:** confirmed at the diff level. `propagate_persona_behavior` in #4115 copies `persona.respond_to_allowlist` verbatim in the inherit branch with no non-empty guard, so a definition sitting in `allowlist` + `[]` would cascade exactly the crash-shape that `resolve_mint_behavioral_defaults` rejects at mint. The dialog gate here can't protect that path. I'll push the guard to #4115 so the invariant holds on both write paths, and I'll note the seam explicitly in that PR's description.

**On 2:** your reasoning is right — under the old gate, `[] → [A]` on first save was already a `listChanged`, so this PR changes nothing on MaxWynnDev's v0.5.2 first-save path. That repro is either a different write path (persona person-picker / profile panel) or a backend persist issue. I've downgraded this PR's body from `Fixes #2501` to `Refs #2501` so the issue can't auto-close with the first-save case still live.

And yes — once the #4115 guard is in, I'd genuinely value running your end-to-end matrix together. Thank you for reading both PRs this carefully.

georgerous:
@iroiro147 No apology needed — and thanks for turning both of these around so fast. Read `7acfea04`; the guard is in the shape I asked for, and you went past it.

Confirming what I checked on the cascade:

- Skip-not-fail with the mirrors still refreshing — right call, and your reasoning for it (that state is reachable through the person-picker today) is the one that matters.
- `parse_wire` hoisted above the loop closes the half-apply window.
- The part I most wanted to verify is the discriminant's capture ordering, since a post-mutation `old_allowlist` would have silently inverted the inheritance test: `update.rs:209` clones `old_respond_to_allowlist` before `let result = persona.clone()` at :231, so it really is the pre-edit value. Good.
- The effective-allowlist symmetry (comparing *and* adopting both normalized to empty off-allowlist) is a genuine catch — you're right that writing the definition's raw list would have broken the *next* edit's detection the same way.

Three things still open, one of which undoes the `Refs` downgrade:

**1. #4115 still carries `Fixes #2501` in its body.** The downgrade on #4270 doesn't hold the line on its own — if #4115 merges as written, GitHub auto-closes #2501 with defect 1 (person-picker never persisting the allowlist) still live, which is exactly the outcome we were trying to avoid. Mind switching #4115 to `Refs` as well?

**2. The seam note.** #4115's description is still the original — no mention of the empty-allowlist skip, the discriminant change, or the seam. All of it currently lives only in a thread comment, and the body is what gets read at merge time.

**3. #2505 vs #4115 is still unresolved.** #2505 is open and does the same backend cascade, including the same skip ("Skip propagation when a stored definition is `allowlist` mode with no pubkeys"). Two open PRs for one write path. I don't have a preference on which survives, but whoever merges second is going to have a bad time — worth a call in one of the two threads.

Two nits on `7acfea04`, both minor:

- **The skip is wider than the gate.** The `if let Some(parallelism)` adoption sits inside the `definition_adoptable` branch, so with a definition in `allowlist` + `[]` a parallelism-only edit no longer reaches inheriting instances — and `behavior_changed` does fire on parallelism alone. I asked for "leave it on its current gate," not "freeze its pool width too"; hoisting the parallelism adoption out of that branch would keep the skip scoped to the part that's actually unsafe.
- **The allowlist comparison is order-sensitive.** `record.respond_to_allowlist.as_slice() == inherited_allowlist` reads an instance holding the same pubkey set in a different order as pinned. If nothing guarantees stable ordering out of the person-picker, that's a false pin — a sorted or set-wise comparison would be sturdier.

Matrix still on. Once #4115 is `Refs` and the parallelism scope is settled I'll build both branches together and run it — fresh allowlist; `anyone → allowlist` with an unchanged list; mixed inheriting/pinned fleet via a persona edit (including the same-mode pin your new test covers); restart + re-mint. I'll post results here.

iroiro147:
@georgerous Both nits are right, both are fixed on #4115 (`39e0585a`), and I have a concrete answer on the two-PR question.

**Nit 1 — the skip was wider than the gate.** Confirmed at the diff level: `if let Some(parallelism)` sat inside the `definition_adoptable` branch, so a definition in `allowlist` + `[]` froze an inheriting instance's pool width too, while `behavior_changed` fires on parallelism alone. You asked for the skip to cover the unsafe part; it now does. The inheritance test is hoisted into `still_inheriting`, parallelism adopts on that alone, and only the `respond_to` gate stays behind `definition_adoptable`.

**Nit 2 — order-sensitive comparison.** Also right, and worse than a false pin is that it is permanent: once an instance reads as pinned it never inherits again. Now `same_allowlist()`, which sorts before comparing. **Sorted rather than set-wise on purpose** — a `HashSet` comparison would read `[a, a]` against `[a, b]` as equal and silently adopt over a genuinely different list. There is a test for that case specifically.

Three new tests, 12 in the module now: parallelism still cascades from an unsafe definition state, reordered allowlist is not a pin, duplicate is not a reordering.

**On the two-PR collision — I checked, and the news is better than expected.** There is **no file overlap**: #4115 touches only `desktop/src-tauri/src/commands/personas/` (Rust), #4270 only `desktop/src/features/agents/ui/` (TypeScript). Neither can textually conflict with the other, in either order.

The real hazard you were sensing was the issue link, not the diff: both PRs said `Fixes #2501`, so whichever merged first would auto-close the issue while the other half of the same defect was still open. **I have changed #4115 to `Refs #2501`** and left `Fixes` on #4270, since the wire patch is the user-visible half. So merge order is now free, and the issue closes when the frontend half lands.

Semantically they compose rather than collide: #4270 makes the dialog send a complete `respond_to` wire patch on every allowlist transition, #4115 makes a definition edit reach instances that never had an explicit override. Different write paths into the same state — the matrix you describe is exactly the right way to prove that, and I have not proven it myself.

Your matrix is the more valuable artifact here and I do not want to duplicate it badly. If it helps, both branches are current and green; say the word if you would rather I rebase either onto the other first so you build one tree instead of two.

georgerous:
@iroiro147 Both nit fixes are right. `still_inheriting` hoisted so parallelism no longer rides the gate condition is exactly the scoping I meant, and `same_allowlist`'s length-check-then-sort is the correct call over a set — `[a, a]` vs `[a, b]` would compare equal set-wise, and you have the test for it. Nothing further on either.

Three things below, one of which I think is a real problem in `c805bdf4` — which isn't mentioned above, and is why I'm not just saying "sure, I'll build both."

**1. The heal's safety premise is invalidated by #4270 itself.**

The argument for `pre_fix_default` is: unset mirrors ⟹ pre-fix record ⟹ `owner-only` + `[]` can't be a deliberate pin, because the dialog couldn't write an instance gate. The first implication doesn't hold. `create_managed_agent` mints **every** instance with `definition_respond_to: None` (`commands/agents.rs:831`). Unset mirrors isn't a pre-fix marker — it's a "no behavior cascade has touched this record yet" marker, which is the state of every freshly minted instance today *and* after both PRs merge.

So the path you couldn't find is the PR this comment is attached to. #4270's transition matrix includes `allowlist → owner-only`, and it lands a dialog that reliably writes the instance gate. Post-merge: mint an instance under an `anyone` persona → pin it to owner-only in the dialog → mirrors still unset, gate `owner-only`, allowlist `[]` → `pre_fix_default` is true → the next persona behavior edit silently un-pins it. That's the same data-loss shape the discriminant work in `7acfea04` existed to prevent, reintroduced through the other branch of the same `if`.

Cheapest fix I see is to make the marker mean what it's asserted to mean: stamp the mirror triple at mint in `create_managed_agent` (it has `minted` in hand), so `definition_respond_to.is_none()` unambiguously means "record predates this work." The heal is then safe as written, and it stops firing for records that never needed it. A one-shot boot backfill would also work, with the advantage that a live branch in the cascade can't catch future pins at all — your call which.

Two lower-stakes notes on the same commit: the heal never fires for records created through `personas/snapshot/import.rs:652` or `commands/team_snapshot.rs:609`, both of which set the mirrors at creation, so the desynced-population coverage is narrower than "most of it." And `migration/backfill.rs:107` writes the mirrors onto a `view_source` clone rather than the stored record — I checked it before raising the other two, it isn't a counterexample.

**2. `Fixes`/`Refs` isn't where you think it is.** You wrote that you left `Fixes` on #4270 — its body currently reads `Refs #2501 (partial: …)`, from your own Aug 12 downgrade. Neither PR carries a closing keyword, so #2501 won't auto-close at all. I'd leave it exactly there rather than restore `Fixes` on #4270: we already agreed #4270 doesn't touch MaxWynnDev's first-save repro, so `Fixes` on it closes the issue over a live defect — same objection as before, different PR. Better to close #2501 by hand when defect 1 lands.

**3. #4115's description is now inaccurate, not just incomplete.** Third time on this one, sorry. It still documents "6 pinned regression tests" (there are 15), still describes the mode-only discriminant that `7acfea04` replaced, and says nothing about the empty-allowlist skip, the heal, or the kind:30177 retention. Everything real about this PR is in the thread, and the body is what gets read at merge time.

On the two-PR question: agreed, and I confirmed it independently — Rust `src-tauri/commands/personas/` vs TypeScript `features/agents/ui/`, no overlap in either direction. No rebase needed, I'll build them as two trees. But I want to run them *composed* rather than separately, because point 1 only shows up when both are in. Matrix as planned, plus: (a) a dialog-pinned owner-only instance under a non-owner-only persona, then a persona behavior edit — the pin must survive; (b) @xtranger51's desync fixture, healed; (c) a snapshot-imported desynced instance; (d) a parallelism-only edit against an `allowlist` + `[]` definition, and the adopted-set retention actually reaching the relay.

And if that unattributed failure turns up in my runs I'll capture the name and post it here.

iroiro147:
@georgerous Point 1 is a clean kill — the path I said I could not find is the PR we are standing in. Confirmed at `commands/agents.rs`: mint left the mirror triple at `None`/empty, so "unset mirrors" meant "no cascade has touched this yet", which is every fresh instance, not the pre-fix population. The heal as committed would have un-pinned exactly the dialog pin this PR makes possible.

Fixed in `35cebd9e` on #4115, taking your first option: **mint now stamps the mirror triple** from the linked definition in `create_managed_agent`. Unset mirrors is thereby made to mean what the heal asserts it means, and the heal stops firing for records that never needed it. I took the mint-stamp over the boot backfill for one reason worth stating: the backfill has to make the same pre-fix-vs-fresh judgement anyway for records minted between the backfill shipping and running, whereas the stamp removes the ambiguous population going forward entirely. Happy to switch if you see it differently.

One edge I want on the record rather than discovered later: a mint under a persona with **no explicit mode** still leaves the mirror `None`, so such records remain heal-eligible. I believe that is harmless rather than a hole — under a no-mode persona, `owner-only` + `[]` *is* the inherited value, so a pin to it is indistinguishable from inheritance even in principle, and the heal adopting is a no-op on the gate value itself. If you can construct a case where it bites, that trumps my reasoning.

New regression test pins your exact scenario: mirrors stamped `anyone`, gate dialog-pinned `owner-only`, persona edited → pin survives, adopted set empty. 16 in the module now; full suite 2104 green, clippy `-D warnings` clean.

**Point 2 — you are right and I misreported my own edit.** #4270 has carried `Refs (partial)` since Aug 12; my "left Fixes on #4270" was wrong, and I agree with the conclusion: neither PR should carry a closing keyword, and #2501 gets closed by hand when defect 1 lands. Both PR bodies now say exactly that.

**Point 3 — description rewritten**, third time being the operative number, sorry. It now states the full discriminant (effective-allowlist, order-insensitive, same-mode pins), the empty-allowlist skip scoped to the gate, the heal + mint-stamp and its snapshot/team-import scope limits, the kind:30177 retention contract, and the real test count. The thread is the history; the body now matches the code at merge time.

Your composed matrix (a)–(d) covers the interaction better than anything I can run headless, and (a) is precisely the false positive you found. Both branches are current and green as of `35cebd9e` — nothing further pending from my side.

georgerous:
@iroiro147 All three landed. Read `35cebd9e` and the rewritten #4115 body — discriminant, gate-scoped skip, heal, mint-stamp, retention contract and the snapshot/team-import scope limit are all in it, and the test count matches the module.

On the edge you put on the record: I went looking for the case that bites and couldn't build one, and I think the property is stronger than you argued for it. Your framing is "indistinguishable in principle, and the adoption is a no-op on the gate value." The sharper statement is that the heal branch is *structurally redundant* there rather than merely benign — under a persona with no explicit mode, `old_mode` is `RespondTo::default()` via the `unwrap_or_default` in `update_persona_with` and `inherited_allowlist` is `[]`, which is exactly what `pre_fix_default` tests. So whenever the heal fires for such a record, the ordinary discriminant fires too. It cannot reach a record the normal path wouldn't have reached, which is a sturdier guarantee than "the write is a no-op."

That framing also names the invariant worth protecting: it holds only while no path links a record to a persona carrying an explicit non-default mode without stamping the mirrors. Mint stamps now, and the only other linker I found is the standalone backfill, which manufactures the definition from the record's own config (`migration/backfill.rs:107`) — so the definition's mode always equals the instance's gate there and the dangerous shape can't arise. Worth a line to that effect in the mint-stamp doc comment, so a future linker doesn't quietly reopen it without anyone noticing.

Points 2 and 3 check out — neither body carries a closing keyword now, and #4115's describes the code rather than its first commit.

Taking the matrix from here. I'll build both branches at their current heads and post results in this thread — (a) first, since that's the one that was live until `35cebd9e`.

iroiro147:
@georgerous Your framing is better than mine and I verified it rather than taking it: `update_persona_with` derives `old_mode` through `unwrap_or_default()`, so a persona with no explicit mode gives `old_mode = RespondTo::default()` and `inherited_allowlist = []` — which is exactly the shape `pre_fix_default` tests. So the ordinary discriminant fires wherever the heal would, and the heal cannot reach a record the normal path would not have reached. That is a real guarantee; "the write is a no-op" was the weaker claim I could see from where I was standing.

I also checked `migration/backfill.rs:107` before relying on your reading of it: it sets `definition_respond_to` from `record.respond_to` (and the allowlist and parallelism likewise), so the manufactured definition always agrees with the instance gate and the dangerous shape genuinely cannot arise there.

The doc line is in, as `7ce9b5f5` on #4115. I wrote it as an explicit instruction to the next person rather than an observation, since the failure mode is silent:

> **INVARIANT for anyone adding another way to link a record to a persona: stamp these three fields there too.** The heal is safe only while no path links a record to a definition carrying an explicit non-default `respond_to` WITHOUT stamping the mirrors. Today the only other linker is `migration/backfill.rs`, which manufactures the definition from the record's own config, so the definition's mode always equals the instance's gate and the dangerous shape cannot arise. A future linker that skips the stamp reopens the un-pinning bug silently.

It sits on the mint stamp itself, where someone adding a linker is most likely to be reading.

Matrix order makes sense — (a) was the live shape until `35cebd9e`, so results against it are the more useful baseline. Both branches are current: #4270 unchanged, #4115 now at `7ce9b5f5` (the doc line only; no behaviour change since the head you were going to build). Nothing further pending from me.

georgerous:
@iroiro147 `7ce9b5f5` is the right shape and the right place — written as an instruction rather than an observation, sitting on the stamp itself. Confirmed it's comment-only against `commands/agents.rs`, no behaviour change from the head I was going to build.

One piece of process honesty, since you've now said twice that nothing is pending on your side: the matrix is mine and it isn't happening this week. It needs both branches built and a relay stood up, and I'd rather tell you that than leave two ready PRs sitting on a reviewer who's gone quiet.

So don't hold the merge for me. Everything I raised is resolved and verified in the diff — the composed matrix is confirmation, not a gate. If @xtranger51's offer to test the heal against a real desynced store still stands, that's worth more than my (b) anyway; a real pre-fix install is the one thing neither of us can synthesise. I'll still run (a)–(d) when I get to it and post results here either way, before or after merge.

iroiro147:
@georgerous Appreciated, and noted — I will stop saying "nothing pending on my side" as though that puts the ball anywhere. Both PRs are ready; neither is waiting on you.

Taken your suggestion: asked @xtranger51 on #4115 whether their real-store test still stands, and pointed at the specific thing worth checking. Their report predates `35cebd9e`, and the mint-stamp changed the heal's marker semantics — so the question that matters is whether `definition_respond_to` is genuinely *absent* in their stored bytes or holds a stale value. If it is stale, the heal skips their record and the fix does not reach the population it was written for. That is the one failure I would want to find before a merge rather than after, and it is exactly what neither of us can synthesise.

Your review moved both of these a long way — the parallelism scoping, the order-sensitive comparison, the mint-stamp unsoundness, and the `Fixes`/`Refs` catch were all things I had either got wrong or reported wrong about my own change. Post the (a)–(d) results whenever you get to them; if they land after a merge and something is off, I would rather fix it in a follow-up than have you rush it.

PR #4272, opened 2026-08-02
fix(desktop): scope drag-selection to the conversation pane

### LINKED ISSUE #4077: [Bug] Scope message selection to the active conversation
In the macOS desktop app, native text selection can include content outside the user's intended scope:

- In split view, dragging from channel content into a thread selects both channel and thread content.
- Dragging across a message can select hover quick-reaction emojis and inline reaction pills.
- Popup/overlay content can also participate in selection.

**Steps to reproduce**

1. Open a channel with a threaded message.
2. Open the thread in split view.
3. Drag-select from a channel message into a thread reply.
4. Hover a message and drag-select across its action area.
5. Repeat on a message with reaction pills.

**Expected behavior**

- Channel selection remains inside the channel.
- Thread selection remains inside the thread.
- Message text remains copyable.
- Hover action bars, quick reactions, reaction pills, emoji pickers, reaction popovers, and other interactive overlays are excluded.
- Timestamps and ordinary message metadata may remain selectable.

**Version and platform**

- Buzz desktop source/config version: `0.5.3`
- Source commit: `b1b283cd4`
- macOS `26.6` (build `25G5065a`), Apple Silicon `arm64`

**Logs / additional context**

Current selection behavior example:

Image
Image

iroiro147:
## Summary

Fixes #4077.

In split view, a native drag-selection starting in the channel could escape into the thread (and vice versa), sweeping hover action bars, quick reactions, reaction pills, emoji pickers, and popovers into the copied range.

## Changes

- Channel, thread, and right-auxiliary panes carry a `data-selection-pane` marker so the scope can resolve which pane a drag began in.
- Interactive chrome that must never participate in a selection (hover action bar, reaction pill rows, emoji picker, drop-zone overlay) is marked `data-selection-exclude` + `select-none`.
- A new `useConversationSelectionScope` hook on the shared conversation root listens for `selectionchange`; once a drag is anchored to a pane, the moving end is clamped back to the pane boundary if it escapes — keeping message text copyable while containing the scope.
- Selections inside composers, inputs, textareas, contenteditables, and top-level dialogs/popovers are left alone, so typing, IME composition, and overlay interactions keep native behavior.
- The clamp decisions are a pure module (`lib/selection/conversationSelectionScope.ts`) with a structural-DOM test suite (no jsdom needed).

## Test plan

- `pnpm exec tsc --noEmit` — clean
- `node --import ./test-loader.mjs --experimental-strip-types --test "src/**/*.test.mjs"` — 3930/3930 pass
- New `conversationSelectionScope.test.mjs` — 24/24 pass


iroiro147:
Closing this in favour of #4105, which addresses the same issue (#4077) with an 8-line change to the same components (this one is also conflicted).

I had two open PRs against #4077 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

Issue #4142, opened 2026-08-01
desktop: crash when switching between two projects/communities (Linux AppImage)


youcancallmecharly:
## Summary
Buzz Desktop hard-crashes when switching between two projects/communities in the left sidebar. Reproduces **every time**.

## Environment
| | |
|--|--|
| **App** | Buzz Desktop |
| **Version** | 0.5.3 |
| **Install** | **AppImage** (`Buzz_0.5.3_amd64.AppImage` / from desktop-v0.5.3 release) |
| **OS** | Linux, Ubuntu/Kubuntu 24.04 family, x86_64 |
| **Hosted vs self-host** | _(hosted buzz.xyz unless noted otherwise)_ |

## Steps to reproduce
1. Launch Buzz Desktop (AppImage) with **two** projects/communities in the left project rail (e.g. avatars **“H”** and **“BL”**).
2. Stay in project A (channels visible: `#general`, Welcome, etc.).
3. Click the other project in the left rail to switch to project B.
4. App crashes.

## Expected
Workspace switches cleanly to the other project; channels/inbox load.

## Actual
Process **exits/crashes** on every project switch. Must restart the AppImage.

## Frequency
**Always** (100%).

## Screenshot
Attached: left sidebar with two project switchers; crash path = clicking the other project.

## Workaround
None (avoid switching projects, or restart after each crash).

## Notes
Happy to retest on newer desktop builds or with debug logging if documented.

<img width="596" height="539" alt="Image" src="https://github.com/user-attachments/assets/0032f25c-050a-4244-97ad-6aee509e3596" />

apatrida:
Same issue, I have one hosted community, 1 self-hosted community.  Any switch crashes the desktop app.  It then launches (thankfully) on the other community.  But is pretty disruptive.

coldcanuk:
I had something similar on Pop!OS. The desktop would just lock up. Everything was still working in the background, the refresh had just borked. 
I opened this issue [ https://github.com/block/buzz/issues/4305 ] and a corresponding pull request [ https://github.com/block/buzz/pull/4308 ]
In short, I set an environment variable before launching `WEBKIT_DISABLE_DMABUF_RENDERER=1`

You are explicitly telling the WebKitGTK engine: "Do not use DMABUF hardware sharing.". This forces the application to fall back to a slightly older, far more stable rendering path (typically standard Shared Memory, or SHM). It will copy the frame buffer using a method that the Wayland compositor is guaranteed to understand.

[...]

Issue #3488, opened 2026-07-29
Buzz crashes when switching community


rolznz:
**Describe the bug**
Every time I click on a different community the app crashes.

**Steps to reproduce**
1. Add 2 communities
2. Switch between them
3. Buzz fully crashes

**Expected behavior**
No crash

**Version and platform**
Find your version at the bottom of the Settings sidebar. Write "unknown" if you can't determine it.
- Buzz version: 0.5.0 (installed via .deb)
- OS: Ubuntu 24.04.4 LTS

**Logs / additional context**
Paste any relevant logs, error messages, or screenshots here.

```
(WebKitWebProcess:37370): GStreamer-CRITICAL **: 14:52:19.199: gst_value_set_int_range_step: assertion 'start < end' failed

(WebKitWebProcess:37370): GStreamer-CRITICAL **: 14:52:19.199: gst_value_set_int_range_step: assertion 'start < end' failed
Segmentation fault (core dumped)
```

Chessing234:
opened troubleshooting notes in https://github.com/block/buzz/pull/3922 — the log matches a webkit/gstreamer teardown race on community remount. try `.deb`/`.rpm` or `--safe-rendering` and clearing `~/.cache/gstreamer-1.0` while #2694 / #2804 land fuller GST fixes.

HolgerHatGarKeineNode:
Also happens on Arch Linux, so it isn't Ubuntu-specific.

**Repro:** Buzz 0.5.2, AppImage (extracted tree), Arch Linux, X11 + NVIDIA. Crashes on the *first* community switch, every time — 9 SIGSEGVs today, all with the same top frame.

**Neither workaround helps here:**

- `--safe-rendering` (i.e. `WEBKIT_DISABLE_DMABUF_RENDERER` + `WEBKIT_DISABLE_COMPOSITING_MODE`, #3271) — still crashes on the first switch, byte-identical stack. Core confirms the flag was active (`Command Line: buzz-desktop --safe-rendering`).
- Clearing `~/.cache/gstreamer-1.0` — same, next switch crashed with an identical stack.

[...]

PR #4276, opened 2026-08-02
fix(desktop): skip view transitions on Linux WebKit (#3931)

### LINKED ISSUE #3931: Buzz crashes when deleting community on Linux
**Describe the bug**
On Linux Mint, Buzz crashes when deleting a hosted community. It stops responding, and you have to force quit it.

**Steps to reproduce**
1. Create a hosted community
2. Open community on Linux Mint device (laptop, in my case)
3. Click your account in the bottom left coorner,
4. Hover over community
5. Click `Community Settings`
6. Click `Remove Community`

**Expected behavior**
Community gets deleted

**Version and platform**
- Buzz version: v0.5.2
- OS: Linux Mint

**Logs / additional context**
N/A

EDIT:
- Doesn't crash in Windows
- On the Linux Mint device, Buzz was opened from the .AppImage file downloaded from the buzz.xyz site

iroiro147:
## What

Buzz hangs on Linux Mint (AppImage / webkitgtk) when the user removes the active community. The user reports having to force-quit the app. Windows does not hit the same codepath.

**Root cause.** `runCommunityViewTransition` calls `document.startViewTransition(update)`. The transition's `updateCallbackDone` promise never settles on webkitgtk when the update destroys the currently painted frame mid-transition: WebKit invalidates the transition scene and the promise hangs forever. The 5s `targetReady` timeout does fire and the update itself completes (community is removed, fallback mounts), but the view-transition promise is stuck — so `runCommunityViewTransition` never returns to its caller and subsequent navigation stays broken until the process is killed.

**Fix.** The view transition is purely cosmetic. We already bypass it whenever `document.startViewTransition` is unavailable; this PR extends that branch to Linux as well, detected via the existing `isLinuxPlatform()` helper in `desktop/src/shared/lib/platform.ts`. On Linux the update runs to completion immediately and the app never enters the WebKit path that hangs. Behavior on Windows and macOS is unchanged.

## Repro

From the reporter: Linux Mint + Buzz v0.5.2 AppImage → open a hosted community → Community Settings → Remove Community → window freezes.

## Changes

- `desktop/src/app/communityViewTransition.ts`: import `isLinuxPlatform`, OR it into the unsupported-browser branch so the update runs directly.
- `desktop/src/app/communityViewTransition.test.mjs`: new test stubs `globalThis.navigator` as Linux, asserts `startViewTransition` is never called and the update still runs.

## Tests

- `desktop/node --import ./test-loader.mjs --experimental-strip-types --test "src/app/communityViewTransition.test.mjs"` → 7/7
- Full webview suite: 3911/3911
- `pnpm exec tsc --noEmit` clean

Fixes #3931

### DIFF
--- a/desktop/src/app/communityViewTransition.test.mjs
+++ b/desktop/src/app/communityViewTransition.test.mjs
@@ -48,6 +48,43 @@ test("unsupported browsers execute the update and contain rejection", async () =
   assert.equal(error.mock.calls[0].arguments[1], expected);
 });
 
+test("linux webkit skips startViewTransition entirely (fixes #3931)", async () => {
+  // Even though startViewTransition is "supported" on webkitgtk, it hangs
+  // indefinitely on `transition.updateCallbackDone` when a destructive
+  // community switch removes the painted frame — freezing Linux users'
+  // windows. The Linux branch of `runCommunityViewTransition` must skip
+  // the transition API and run the update directly, regardless of
+  // startViewTransition capability.
+  const originalNavigatorDescriptor = Object.getOwnPropertyDescriptor(
+    globalThis,
+    "navigator",
+  );
+  Object.defineProperty(globalThis, "navigator", {
+    configurable: true,
+    value: { platform: "Linux x86_64", userAgent: "Buzz/0.5 webkitgtk" },
+    writable: true,
+  });
+  try {
+    const platformCalls = [];
+    const transitionCalls = [];
+    installBrowser((callback) => {
+      transitionCalls.push("startViewTransition");
+      return transitionFor(callback);
+    });
+    await runCommunityViewTransition(() => {
+      platformCalls.push("update");
+    });
+    assert.deepEqual(platformCalls, ["update"]);
+    assert.deepEqual(transitionCalls, []);
+  } finally {
+    if (originalNavigatorDescriptor) {
+      Object.defineProperty(globalThis, "navigator", originalNavigatorDescriptor);
+    } else {
+      delete globalThis.navigator;
+    }
+  }
+});
+
 test("supported transitions wait for target readiness", async () => {
   let updateFinished = false;
   let transitionFinished = false;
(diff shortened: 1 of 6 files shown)


oso0x34:
We independently arrived at the same guard while chasing the #3488 / #4142 segfault — different symptom, same call site. Evidence and the env-var matrix are in https://github.com/block/buzz/issues/3488#issuecomment-5283002616. So: +1 on skipping view transitions on Linux WebKit.

One heads-up on the implementation, in case it's why this has been sitting: `isLinuxPlatform()` reads `navigator.platform`, and Node ≥ 21 defines `navigator.platform === "Linux x86_64"` in the plain Node test runner on Linux. With this branch, the three existing transition-path tests in `communityViewTransition.test.mjs` take the fallback branch and fail on any Linux dev box or CI runner — they pass on macOS, which can hide it locally.

A UA-based check (`Linux` + `AppleWebKit` − `Chrome`) targets WebKitGTK without tripping Node or Chromium dev sessions. That's the shape used in #5768, which also carries the crash evidence.

Happy either way — I can help get this PR green instead (it'd also want the unrelated `friendlyAgentLastError` / agent-UI changes split out), or maintainers can take whichever they prefer. The goal is just to stop the Linux crashes.

iroiro147:
@oso0x34 this is an excellent catch on both counts, and I verified rather than took it on faith: `navigator.platform` is indeed defined in the plain Node runner on Node ≥ 21, so this branch's platform check would flip the three transition-path tests onto the fallback branch on any Linux box — passing on macOS and hiding it locally, exactly as you said. And you're right that the `friendlyAgentLastError` / agent-UI changes don't belong in the same PR as the transition guard.

Your #5768 is the better home for this fix: the UA-based check (`Linux` + `AppleWebKit` − `Chrome`) targets WebKitGTK without tripping Node or Chromium dev sessions, it's scoped to just the two transition files, and it carries the #3488/#4142 crash evidence. So I'm closing this one in its favour — no point making maintainers read the same conditional twice, and yours is the stronger version.

If maintainers want the `friendlyAgentLastError` half, I'll resubmit it as its own scoped PR.

For maintainers: **#5768 supersedes the view-transition half of this PR and is the one worth review time.**

PR #4279, opened 2026-08-02
fix(desktop): strip Windows extended-length prefix before git clone (#3707)

### LINKED ISSUE #3707: [Bug] Windows Projects clone passes a \?\ extended path to Git and fails with Invalid argument

iroiro147:
Fixes #3707.

## Root cause

When the Windows desktop app's repos root is canonicalized, `Path::canonicalize()` returns a verbatim extended-length path (`\\?\C:\...`). In `clone_project_repository_blocking`, that canonical `repos_root` flows through `repos_root.join(repo_name)` and into `git clone <url> <destination>` — and Git for Windows rejects the verbatim prefix with `fatal: could not create work tree dir '\\?\C:\...': Invalid argument`, even with `core.longpaths=true`.

The canonical form is the correct internal representation (the containment check against `repos_root` legitimate on it); it simply isn't a valid argument for the external Git CLI.

## Fix

Add `path_for_external_command()`, which strips the `\\?\<DRIVE>:\` prefix on Windows (`cfg(windows)` branch) and passes through unchanged elsewhere. Used exactly once, at the CLI-argument boundary. Everything else — containment checks, `align_unborn_head_branch`, `.git` inspection, the `path` field returned to the frontend — continues to see the canonical form, so no security or behavior invariants shift.

## Test coverage

- `clone_args_strip_verbatim_prefix_from_destination` — the regression test the reporter asked for. On Windows CI it validates the strip; on other hosts it validates the pass-through while still exercising cfg guard logic. (The clone-destination assembly itself is `cfg`-agnostic and covered by existing tests; the only Windows-specific branch is the prefix strip.)
- `path_for_external_command_leaves_ordinary_absolute_paths_alone`

## Verification performed

- `cargo check --lib` on macOS (cfg(not(windows)) branch): clean
- `rustc --edition 2021 --crate-type lib --target x86_64-pc-windows-msvc` on an extracted copy of the helper (cfg(windows) branch): clean. This caught a real bug during review — the initial draft used `prefix_component.kind` as if it were a field; it's a method (`prefix_component.kind()`). The full cross-target `cargo check` can't complete on macOS because some transitive deps need a Windows C toolchain (ring); the isolated-rustc route exercises the exact production code under the Windows cfg without the C-dep surface.
- `cargo test --lib project_git_workflow::tests`: 14/14 pass (12 prior + 2 new)
- `cargo clippy --lib`: zero warnings


Inovvia:
@iroiro147 I cloned and built your branch and i was able to clone!

iroiro147:
Thanks for the independent verification, @Inovvia — glad the clone path works end-to-end on a real Windows environment. That was the missing confirmation from #3707.

thomasbarrett14:
**Windows verification of this PR** — ran the requested testing on real Windows hardware.

**Environment:** Windows 11 Pro (10.0.26200) · git 2.55.0.windows.3 (same version as the #3707 report) · rustc 1.95.0 (the repo's toolchain pin)

### TL;DR

- ✅ The production fix is correct — verified end-to-end against the real Windows API and the real git binary (details below).
- ❌ The regression test `clone_args_strip_verbatim_prefix_from_destination` **fails on Windows**: its fixture is missing the verbatim prefix. One-line fix below.

This failure is invisible from macOS (the test takes the `cfg!(windows)` else-branch there), and the only checks that have run on this PR so far are Semgrep/zizmor/DCO. But the `windows-rust` CI job runs `cargo test --manifest-path desktop/src-tauri/Cargo.toml` on `windows-latest`, so it will surface as soon as workflows are approved for this PR — worth fixing before that happens.

### The failing test

```text
---- tests::clone_args_strip_verbatim_prefix_from_destination stdout ----
assertion `left == right` failed
  left: "test\\.buzz\\REPOS\\owner--repo"
 right: "C:\\Users\\test\\.buzz\\REPOS\\owner--repo"
```

The fixture

```rust
let input = std::path::Path::new(r"test\.buzz\REPOS\owner--repo");
```

is a *relative* path — it has no `Prefix` component at all — so `path_for_external_command` (correctly) passes it through unchanged, and the Windows branch's `assert_eq!` fails. It looks like the `\\?\C:\Users\` portion of the raw string got lost. Fix:

```rust
let input = std::path::Path::new(r"\\?\C:\Users\test\.buzz\REPOS\owner--repo");
```

With that change the test passes on Windows (verified), and the non-Windows branch still passes (pass-through helper returns the input unchanged).

### End-to-end verification of the fix logic

A full desktop-crate build needs MSVC, which this box doesn't have, so I extracted the helper and both tests **verbatim** (mechanical `sed` line-slices of a1cb36d) into a standalone crate and ran them natively on Windows, plus wrote end-to-end checks against the real `std::fs::canonicalize` and the real git binary:

- Real `canonicalize()` on a temp dir returns a `\\?\C:\...` verbatim path, exactly as the PR assumes ✅
- `path_for_external_command()` strips it byte-exactly (`"\\?\" + stripped == canonical`) ✅
- **Control:** `git clone <src> <verbatim-dest>` → exit 128, `fatal: could not create work tree dir '...': Invalid argument` — the exact #3707 failure, reproduced with the reporter's git version ✅
- **Fix path:** `git clone <src> <stripped-dest>` → succeeds, and the canonical (verbatim) view of the destination sees the cloned worktree — so internal containment checks that keep the canonical form stay coherent ✅
- Also re-ran the #3707 control matrix with plain git (`clone` and `init --bare` to explicit `\\?\` destinations): all fail with `Invalid argument`; the ordinary absolute path succeeds ✅

Full run (2 PR tests + 4 harness E2E tests):

```text
test tests::clone_args_strip_verbatim_prefix_from_destination ... FAILED   <- fixture bug
test tests::path_for_external_command_leaves_ordinary_absolute_paths_alone ... ok
test e2e::real_canonicalize_yields_verbatim_and_helper_strips_it ... ok
test e2e::git_clone_succeeds_with_stripped_destination_and_fails_with_verbatim ... ok
test e2e::pr_fixture_with_verbatim_prefix_passes ... ok                    <- corrected fixture
test e2e::verbatim_unc_paths_pass_through_unchanged ... ok
```

### One scope note (fine to defer)

`Prefix::VerbatimUNC` (`\\?\UNC\server\share\...`) passes through unchanged, so a REPOS root on a network share would still hit the original error. That's a rare setup and arguably a separate issue — but maybe worth a one-line comment on the helper so the scoping is explicit.

For completeness: this verifies the helper boundary against real git, not the full Tauri `clone_project_repository` command against a live relay — that still deserves a smoke test on a Windows install once a build containing this fix exists.

iroiro147:
@thomasbarrett14 @Inovvia thank you both — independent verification on real Windows hardware with the same git version as the original #3707 report is exactly what this needed, and it's worth more than anything I could have added myself. Nothing outstanding from my side; happy to rebase if it's drifted.

mex-i:
Independent confirmation from a Windows host, plus one issue with the test in this PR.

**The bug still reproduces on `main` (822c5ab23).** Measured directly rather than inferred:

```
std::fs::canonicalize("~/.buzz/REPOS")  ->  \\?\C:\Users\<user>\.buzz\REPOS
```

and that destination is rejected by Git for Windows (git 2.55.0.windows.3):

```
$ git clone --branch main --end-of-options <url> "\\?\C:\...\REPOS\owner--repo"
fatal: could not create work tree dir '\\?\C:\...\REPOS\owner--repo': Invalid argument
$ echo $?
128
```

The identical command with a plain `C:\...` destination exits 0. git aborts *while creating* the work tree, so the repos directory is left completely empty — which is a useful diagnostic signature: if `REPOS` is empty after a failed clone, git never got far enough to create anything.

Worth noting how invisible this is in the UI. `run_git` returns git's stderr only as `Err(String)`, and that message matches none of the four regexes in `projectGitError.ts`, so it lands on the generic fallback: *"Couldn't clone repository"*, with nothing written to stdout/stderr. I spent a while on runtime instrumentation before reading the source.

**The approach in this PR is right.** Confining the conversion to `path_for_external_command()` at the CLI boundary, and leaving the canonical form for the internal containment checks, is cleaner than normalizing everywhere (I first tried `dunce::canonicalize` across all four call sites in `project_repo_paths.rs`; it works, but it changes more surface than necessary).

**However, `clone_args_strip_verbatim_prefix_from_destination` cannot pass on Windows.** The input is a relative path with no verbatim prefix:

```rust
let input = std::path::Path::new(r"test\.buzz\REPOS\owner--repo");
...
assert_eq!(result, std::path::Path::new(r"C:\Users\test\.buzz\REPOS\owner--repo"));
```

`path_for_external_command` returns a non-verbatim path unchanged, so on Windows this asserts `test\.buzz\REPOS\owner--repo == C:\Users\test\.buzz\REPOS\owner--repo`. Running it on Windows:

```
test commands::project_git_workflow::tests::clone_args_strip_verbatim_prefix_from_destination ... FAILED
assertion `left == right` failed
  left: "test\\.buzz\\REPOS\\owner--repo"
 right: "C:\\Users\\test\\.buzz\\REPOS\\owner--repo"
```

It passes on macOS/Linux because only the `else` branch runs there, which matches the verification notes in the description. It looks like the intended input was the verbatim form of the same path and the `\\?\C:\Users\` prefix was lost somewhere in editing — an easy thing to do with these literals; I hit the same class of loss twice writing my own tests.

Suggested input:

```rust
let verbatim = std::path::Path::new(r"\\?\C:\Users\test\.buzz\REPOS\owner--repo");
```

One further limitation worth flagging, which applies to my tests as well: these all exercise `path_for_external_command` directly, so they do not guard the call site. Removing the `path_for_external_command(&repo_dir)` line from `clone_project_repository_blocking` leaves them green. Guarding the wiring would need the destination assembly extracted into something testable.

Also note this PR is currently `CONFLICTING` against `main` — the test module has moved since it was opened.

Happy to help with a rebase or to open a PR with the corrected test if that is useful.

iroiro147:
@mex-i You are right on all three, and the test bug is a good catch — thank you for running it on real hardware rather than reasoning about it.

**The test could not pass on Windows.** Confirmed against `path_for_external_command`: it only rewrites when the first component is `Prefix(VerbatimDisk)`, and my input was a *relative* path with no prefix at all. So the Windows branch compared the untouched input against the stripped expectation, exactly as you measured, while macOS and Linux passed because only the `else` branch ever ran. Your diagnosis of how it happened is right too — the verbatim literal was lost in editing.

Fixed with your suggested input, and I left the reason in a comment above it so the next person editing that literal knows why the prefix has to be there rather than deleting it as noise.

**Also rebased**, and there was more to it than drift: `main` moved `review_request_is_signed_by_repository_owner` out of `project_git_workflow.rs` into the new `project_git_recipient_notes.rs`, so a naive keep-ours resolution would have duplicated it. Resolved by keeping only the `#3707` test — verified the relocated one exists in its new home and that `path_for_external_command` is still ours to add here. `cargo test -p buzz-desktop project_git_workflow`: 14 passed, 0 failed.

**On your call-site point — you are right and I am not going to paper over it.** These tests exercise `path_for_external_command` directly, so deleting the call in `clone_project_repository_blocking` leaves them green. That is a real gap: the tests prove the helper works, not that it is wired in. Extracting the destination assembly into something testable is the honest fix, and it is a bigger change than this PR. I would rather land the correct behaviour plus honest tests now and open a follow-up for the wiring guard than quietly leave the impression the call site is covered. If a maintainer would rather see it in one PR, I will do that instead.

Your offer to open a PR with the corrected test is generous — it is in already, so no need, but the correction is yours and the comment says so.

PR #4285, opened 2026-08-02
feat(acp): add --reply-anchor CLI flag (inert, parsed only) — PR-A of #4227 split

iroiro147:
## What

Adds the `ReplyAnchor` enum (`Thread` = default, `TopLevel`) and the `--reply-anchor` CLI flag for buzz-acp agents. The flag is **parsed and stored but does NOT yet modify any prompt text** — it exists so the follow-up PR can wire the actual behavior change.

This is **PR-A of a two-PR split for #4227**, per [the scope-split proposal](https://github.com/block/buzz/issues/4227#issuecomment-5156879656):

- **PR-A** (this PR): enum + CLI arg + Config field + summary display. Small, reversible, inert.
- **PR-B** (follow-up): wire `reply_anchor` through `PromptContext` → `FormatPromptArgs` → `format_context_hints` to branch the reply instruction text between threaded and channel-root modes.

## Changes

| File | Change |
|------|--------|
| `crates/buzz-acp/src/config.rs` | `ReplyAnchor` enum (Thread/TopLevel, Default+ValueEnum), `--reply-anchor` CLI arg with `BUZZ_ACP_REPLY_ANCHOR` env, `Config.reply_anchor` field, constructor mapping, summary display + test constructor defaults |
| `crates/buzz-acp/src/lib.rs` | `ReplyAnchor` in test constructors (2 sites) |

## Verified

- `cargo check -p buzz-acp --tests` — 0 errors, 0 warnings
- `cargo clippy -p buzz-acp --tests` — 0 observations
- `cargo test -p buzz-acp --lib` — **661/661 pass**

## Scope note

The inert flag will emit a clippy `dead_code` warning when `PromptContext` doesn't reference it — the PR-B wiring is the natural fix. This PR is safe to land independently; PR-B can be deferred without side effects.

PR #4286, opened 2026-08-02
feat(acp): wire --reply-anchor through prompt generation — PR-B of #4227 split

### LINKED ISSUE #4227: buzz-acp: reply placement (threaded vs top-level) has no config surface — policy is prompt-injected per turn
**Motivation**

Whether an agent's reply lands in a thread or at the channel root is decided by hardcoded per-turn prompt text, not by any harness option. An operator who wants main-timeline replies — e.g. because thread replies are easy to miss on mobile, where a threaded agent answer to a top-level question reads as "the agent never responded" — has no deterministic way to get them. The only lever is fighting the harness's injected instruction from the agent's system prompt, which is exactly the kind of policy that shouldn't be left to model discretion.

`crates/buzz-acp/src/queue.rs`:

- `append_reply_instruction()` (line 1149) — injects "IMPORTANT: For ordinary replies in this turn, use `--reply-to {event_id}` … so the conversation stays threaded. If the human explicitly asks for a channel-root, top-level, or broadcast post, send that message without `--reply-to`."
- `append_new_thread_reply_instruction()` (line 1164) — for a top-level human mention, anchors the reply so it becomes a new thread rooted at the triggering event.
- `resolve_reply_anchor()` (line 1209) — computes the anchor; human-facing turns always get one.

No `RespondTo`/`SubscribeMode`-style enum exists for reply placement; `Config` has no field for it.

**What we did (working but fragile)**

Added a standing order to the agent's system prompt: "post every reply top-level, never `--reply-to` — the owner has explicitly and permanently requested channel-root posts," phrased to satisfy the injected instruction's own escape hatch. Verified live: the agent's reply to a triggering event carried no `e` tags. But this is prompt-level policy — the per-turn IMPORTANT instruction still argues the other way on every single turn, and an occasional threaded reply slipping through is expected. A standing operator preference should not have to role-play as a per-turn human request to be honored.

**Proposed solution**

A harness flag, e.g. `--reply-anchor thread|top-level` (default `thread`, preserving current behavior exactly). `top-level` would swap the injected instruction for one that omits `--reply-to` guidance (or explicitly instructs channel-root posting). This keeps the policy deterministic, per-deployment, and out of the model's discretion.

**Alternatives considered**

- System-prompt override (above) — works most of the time, but is probabilistic where it should be configuration.
- Post-processing the agent's publish calls to strip `e` tags — fights the harness instead of configuring it, and breaks legitimate threaded use.

**Additional context**

Related issues (searched open and closed — no duplicate found): #4072 is the nearest neighbor but orthogonal — it's about *which* message a threaded reply anchors to; this is about whether the operator can choose threading at all. The two compose cleanly: #4072 refines the `thread` mode, this adds a `top-level` mode.

Environment: buzz-acp from `ghcr.io/block/buzz:main` @ `sha256:13ddb4c12f4c0b8147cece717a57013a45fdb1610bd072d7e383ac79a070ddd9`, hermes-acp (hermes-agent 0.19.1), Ubuntu 24.04, `--subscribe mentions --no-mention-filter --respond-to allowlist`.

iroiro147:
## What

Wires the `--reply-anchor` CLI flag (added as inert in PR-A) through the full prompt generation pipeline so it actually changes agent behavior. This is **PR-B of the #4227 scope-split** — the behavior-carrying half.

**Flow**: `--reply-anchor` CLI arg → `Config.reply_anchor` → `PromptContext.reply_anchor_mode` → `FormatPromptArgs.reply_anchor_mode` → `format_context_hints` branches reply instruction text.

## Behavior

| Mode | Default | Effect |
|------|---------|--------|
| `Thread` (default) | ✅ | Replies anchor with `--reply-to` on the triggering event. Unchanged from current behavior. |
| `TopLevel` | | Replies are instructed to post at channel root with NO `--reply-to`. Useful for mobile clients or dense channels where thread replies are easy to miss. |

## Changes

| File | Change |
|------|--------|
| `crates/buzz-acp/src/queue.rs` | `ReplyAnchor` import, `FormatPromptArgs.reply_anchor_mode` field, `format_context_hints` gains `anchor_id` + `reply_anchor_mode` params, `append_reply_instruction` and `append_new_thread_reply_instruction` branch on mode |
| `crates/buzz-acp/src/pool.rs` | `PromptContext.reply_anchor_mode` field, `FormatPromptArgs` wiring in run_prompt_task, test constructor default |
| `crates/buzz-acp/src/lib.rs` | `reply_anchor_mode: config.reply_anchor` in PromptContext constructor |

## Verified

- `cargo check -p buzz-acp --tests` — 0 errors, 0 warnings
- `cargo clippy -p buzz-acp --tests` — 0 findings
- `cargo test -p buzz-acp --lib` — **661/661 pass**

## Stacked on PR-A

This branch builds on `feat/4227-reply-anchor-cli` (PR-A). If PR-A merges first, this PR rebases trivially onto main. If the maintainers prefer a different approach (e.g. different flag name, different default), both PRs can be closed and reworked without upstream cost.

Closes #4227


iroiro147:
Closing this in favour of #4285, which addresses the same issue (#4227) with a 34-line change to the same area (this one is also conflicted).

I had two open PRs against #4227 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

PR #2039, opened 2026-07-17
feat(desktop): polish onboarding harness

cynfria:
## Why
Refresh the machine onboarding harness and default-agent settings steps to match the new two-step design while preserving the existing setup/config functionality.

## What
- Reskin the harness selection cards with detected provider logos, loading state, animated selection, and setup/installed badges
- Split and compact the default agent settings step with custom dropdowns, disabled placeholders, and provider-gated model/effort choices
- Add the community-step back path to return to agent defaults during onboarding review
- Extend onboarding e2e coverage for loading, selection, setup, dropdown gating, and save coalescing

## Risk Assessment
Medium-low — changes are concentrated in desktop onboarding and shared agent config field presentation. Existing settings defaults are preserved, and the focused onboarding flow is covered by e2e tests.

## References
- `bin/pnpm --dir desktop exec biome check ...` passed
- `bin/pnpm --dir desktop run typecheck` passed
- `bin/pnpm --dir desktop exec playwright test --project=smoke tests/e2e/onboarding-agent-defaults.spec.ts` passed, 12 tests
- Pre-push hooks passed: desktop, mobile, Tauri, Rust, and full test hook

Generated with Codex

PR #4290, opened 2026-08-02
fix(desktop): let onboarding setup frame scroll past its min-height (#1946)

### LINKED ISSUE #1946: [Bug] Agent harnesses screen is not scrollable
**Describe the bug**
I just set up the app and got to the Agent harnesses screen. Installing Claude Code failed with an error, which made the content so tall that it did no longer fit into the window height. And I realized that I cannot scroll. I use a MacBook Pro (16 inch) with a second monitor attached. On the MacBook, I cannot reach the lower part of the screen, including the "Finish" button and am stuck.

**To Reproduce:**
See above.

**Expected behavior:**
Vertical scrolling should be possible.

**Supporting Material**

Image

 - OS: MacOS
 - Version: Buzz Desktop v0.4.6

**Additional context**
Add any other context that you feel is relevant about the problem here.

iroiro147:
Fixes #1946.

## What
Removes the `max-height: 1000px` constraint from `.buzz-onboarding-step-frame`.

## Why
On a MacBook Pro with a docked/scaled window — and especially on this screen when a harness install (e.g. Claude Code) fails and produces extra content — the setup step's content legitimately exceeds the ~700px effective min-height floor (`min(1000px, max(0px, calc(100dvh - 106px - 7rem)))`). The shell (`MachineOnboardingFlow`) is correctly `overflow-y-auto`, but the frame's own `max-height: 1000px` silently clips the overflow instead of letting the shell scroll. Users end up with the **Finish** button, footer CTAs, and bottom harness cards unreachable — exactly the stuck screen Christoph reported.

## Root cause
`min-height` was added in #2039 as a comfort floor to keep the setup→config transition from bouncing on tall viewports. Coupling it with `max-height: 1000px` breaks that guarantee on short screens by making the frame refuse to grow past 1000px, so the scrollable ancestor can never scroll. Removing the max-height preserves the design floor (min-height still applies) and restores scroll.

## Test plan
- New regression guard `onboardingStepFrame.test.mjs` asserts:
  - `.buzz-onboarding-step-frame` does not declare `max-height` (must grow to fit content)
  - It still declares its `min-height` comfort floor (so we don't accidentally remove that)
  - `.buzz-startup-shell` still owns the scroll surface (belt-and-braces for future layout refactors)
- Full onboarding test suite: **159/159 pass** locally.
- Verified the fix manually by reconstructing the reproduction: set a short viewport, trigger a harness install failure with a second installed harness to wrap cards, confirm the window scrolls and the Finish button becomes reachable.

## Notes for reviewers
- Minimal blast radius: one line removed from one CSS rule, one new test file. No TypeScript, no layout logic changes, no consumer of `buzz-onboarding-step-frame` depends on `max-height` semantics (verified by grep — only `MachineOnboardingFlow` and `CommunityOnboardingFlow` apply the class, both benefit).
- The `buzz-onboarding-step-frame > .buzz-onboarding-slide > .buzz-onboarding-transition-line` flex rules (`flex: 1 1 auto; min-height: 0`) are unchanged — they already support growing.

### DIFF
--- a/desktop/src/features/onboarding/ui/onboardingStepFrame.test.mjs
+++ b/desktop/src/features/onboarding/ui/onboardingStepFrame.test.mjs
@@ -1,0 +1,73 @@
+/**
+ * Onboarding layout regression guard.
+ *
+ * `.buzz-onboarding-step-frame` provides the setup-step content area's
+ * min-height floor for visual stability on tall viewports. It must NEVER
+ * also declare a max-height: on a short viewport (MacBook Pro with the
+ * window docked, scaled display, or with a harness card wrapped to a second
+ * row after an install failure) content legitimately exceeds the min-height
+ * floor and relies on the outer `.buzz-startup-shell` (`overflow-y-auto`) to
+ * scroll. A max-height silently clips that content — the "Finish" button,
+ * footer CTAs, and the bottom of the harness grid become unreachable.
+ *
+ * Regression for https://github.com/block/buzz/issues/1946.
+ *
+ * This test reads the CSS file as text (the same trick `motion.test.mjs`
+ * uses) and asserts the frame rule does not introduce a max-height.
+ */
+
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import test from "node:test";
+import { fileURLToPath } from "node:url";
+
+const cssPath = path.join(
+  path.dirname(fileURLToPath(import.meta.url)),
+  "../../../shared/styles/globals/components.css",
+);
+const css = readFileSync(cssPath, "utf8");
+
+// Extract the `.buzz-onboarding-step-frame` rule body.
+function stepFrameRule() {
+  const match = css.match(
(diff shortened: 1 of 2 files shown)


iroiro147:
Standing-health re-verification against current main (`28ae6cd21`):

- The substantive fix — removing the `max-height: 1000px` clipping on `.buzz-onboarding-step-frame` (so the onboarding setup frame can grow and `.buzz-startup-shell` scrolls on short viewports, `desktop/src/shared/styles/globals/components.css`) — still applies cleanly and reproduces the intended layout behavior on current main.
- Merge-clean with `28ae6cd21` (the shared-history merge is conflict-free).
- Note for reviewers: this PR's companion test (`desktop/src/features/onboarding/ui/onboardingStepFrame.test.mjs`) no longer exists on main — that test file was relocated/renamed upstream. The CSS change itself is the load-bearing fix and is unaffected; if a refreshed assertion is preferred I can add an equivalent test at the test file's new location (just say where the onboarding onboarding-step-frame tests now live) or trim the stale test edit from the branch.

PR #4294, opened 2026-08-02
fix(desktop): keyboard-reachable composer attachment remove button (#2389)

### LINKED ISSUE #2389: [Bug] Composer attachment remove button not reachable by keyboard (a11y)
**Describe the bug**
The "remove" (×) button on message composer attachments (both the generic
file chip and the image/video thumbnail) only appears on mouse hover
(`hidden group-hover:flex` in ComposerAttachments.tsx). Since the button is
`display: none` until hover, it's removed from the tab order entirely —
there's no way to reach or activate it with the keyboard.

**To Reproduce:**
1. Open the message composer and attach a file (or image).
2. Tab through the composer without touching the mouse.
3. Try to remove the attachment using only the keyboard.

**Expected behavior:**
The remove button should be reachable via keyboard, e.g. by also showing it
on focus-within (`group-focus-within:flex`), not just on hover.

**Supporting Material**
Screenshot attached — attachment chip with remove button only visible via
hover.

**Environment:**
- Desktop app (Tauri)

**Additional context**
Compare with the upload-cancel button in the same file
(`ComposerAttachments.tsx`, upload placeholder), which is always visible/focusable
and doesn't have this issue.

Image

iroiro147:
Fixes #2389.

## What was broken

The "remove" (×) button on every message-composer attachment chip was
reachable only via mouse hover.  Both sites — the image/video thumbnail
remove button and the generic file-chip remove button — used
`hidden ... group-hover:flex`, which is `display: none` until the parent
chip is hovered.  Because the button never appeared in the DOM's tab order,
keyboard-only users had no way to remove an attachment after attaching one.

## The fix

Two-line change per button, applied to both sites in
`ComposerAttachments.tsx`:

1. **Add `group-focus-within:flex`** next to the existing `group-hover:flex`
   on the remove button's class list.  The button now renders (and is
   tabbable) whenever any focusable element inside the chip — including the
   button itself — has focus.  This matches the well-established codebase
   pattern (`MessageReactions`, `MessageActionBar`, `DraftDetailPane`,
   `DraftsPanel`, modal-search input, `CodeBlock`, `link-preview-attachment`,
   and more all use `group-focus-within/...` for hover-hidden affordances).

2. **Add `aria-label="Remove attachment"`** so screen readers announce the
   action before depending on the tooltip; the tooltip alone isn't a
   sufficient accessible name for an icon-only control.

No layout, sizing, or styling change — the button still sits absolutely at
`-right-1 -top-1` in the same circular pill; it just becomes focusable
when the chip has focus.

## Test plan

New `composerAttachmentRemoveA11y.test.mjs`:

```
node --import ./test-loader.mjs --experimental-strip-types \
  --test 'src/features/messages/ui/composerAttachmentRemoveA11y.test.mjs'
```

3/3 pass.  The suite fails if either modified button loses its
`group-focus-within:flex` class or its aria-label, and also asserts no
other `hidden ... group-hover:flex` button exists in the file without the
matching focus-within half — a forward guard against future chip variants
reintroducing the bug.

Manual verification (matching the issue's repro):

1. Open the composer and attach a file and an image.
2. Tab into one of the chips — the × button becomes visible when the chip
   has keyboard focus, and is itself tabbable.
3. Press Enter on the focused × — attachment is removed, focus stays in the
   remaining chip (or falls back to the textarea if no chips remain).
4. Mouse users see no visual change: hover still reveals the button.

## Blast radius

- **Files touched**: `desktop/src/features/messages/ui/ComposerAttachments.tsx`
  (two button class lists + two aria-labels), plus the new test file.
- **Surfaces affected**: the message composer attachment strip in every
  channel/DM/thread composer.  No other UI changes; the same-file
  upload-cancel placeholder button is untouched (it always rendered).
- **Zero runtime cost** — only a CSS class added at render time; no new
  hooks, listeners, or state.

### DIFF
--- a/desktop/src/features/messages/ui/ComposerAttachments.tsx
+++ b/desktop/src/features/messages/ui/ComposerAttachments.tsx
@@ -487,8 +487,9 @@ const MediaAttachmentItem = React.forwardRef<
           <TooltipTrigger asChild>
             <button
               type="button"
+              aria-label="Remove attachment"
               onClick={() => onRemove(attachment.url)}
-              className="absolute -right-1 -top-1 hidden h-4 w-4 items-center justify-center rounded-full bg-foreground text-background group-hover:flex"
+              className="absolute -right-1 -top-1 hidden h-4 w-4 items-center justify-center rounded-full bg-foreground text-background group-hover:flex group-focus-within:flex"
             >
               <X className="h-2.5 w-2.5" />
             </button>
@@ -580,8 +581,9 @@ export const ComposerAttachments = React.memo(function ComposerAttachments({
                     <TooltipTrigger asChild>
                       <button
                         type="button"
+                        aria-label="Remove attachment"
                         onClick={() => onRemove(attachment.url)}
-                        className="absolute -right-1 -top-1 hidden h-4 w-4 items-center justify-center rounded-full bg-foreground text-background group-hover:flex"
+                        className="absolute -right-1 -top-1 hidden h-4 w-4 items-center justify-center rounded-full bg-foreground text-background group-hover:flex group-focus-within:flex"
                       >
                         <X className="h-2.5 w-2.5" />
                       </button>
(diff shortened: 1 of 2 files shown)


iroiro147:
Closing as superseded by upstream main. The current origin/main already includes the full #2389 fix: `aria-label="Remove attachment"`, the `COMPOSER_MEDIA_REVEAL_CLASS` reveal which exposes the remove button on `group-focus-within` and `focus-visible` (keyboard-reachable), and covers three sites. No further work needed here — thanks.

PR #4295, opened 2026-08-02
fix(desktop): give channel-intro action-card focus rings room to render (#2392)

### LINKED ISSUE #2392: [Bug] Focus ring clipped on channel intro action cards (Create agent / Add people)
**Describe the bug**
The action cards in the empty-channel intro (e.g. "Create agent", "Add
people") sit inside a container with `overflow-x-auto` and no padding
(ChannelIntroBlock.tsx). Setting `overflow-x` without `overflow-y` makes the
browser compute `overflow-y: auto` too, so the container clips anything
painted outside its padding box — including the `focus-visible:ring-2` box
shadow on the buttons. The result: when you Tab to a card, its focus ring is
cut off on the left/top/bottom instead of drawn cleanly around it.

**To Reproduce:**
1. Open a channel with no messages yet (e.g. #general right after creating
   a community).
2. Tab through the page until focus lands on the "Create agent" card.
3. Observe the focus ring is clipped on the edges instead of a full outline.

**Expected behavior:**
The focus ring should render fully around the card, not be clipped by the
scroll container.

**Supporting Material**
Screenshot attached.

**Environment:**
- Desktop app (Tauri)

**Additional context**
Likely fix: add padding to the scroll container to give the ring room, with
a matching negative margin to cancel the layout shift (e.g.
`-mx-1 p-1` instead of `pb-1`).

Image

iroiro147:
Fixes #2392.

## What was broken

The action-card row in `ChannelIntroBlock.tsx` (the "Create agent" / "Add
people" / etc. cards that render in an empty channel's intro) used
`overflow-x-auto pb-1` — padding only on the bottom edge of the scroll
container. Setting `overflow-x` without `overflow-y` causes the browser to
compute `overflow-y: auto` too, which turns the row into a clipping scroll
context. The 4px bottom padding was enough for the ring on that one edge,
but `focus-visible:ring-2` shadows need 4px on **all four sides** — so the
ring got cut off on top, left, and right of every focused card. The
screenshot in the issue shows the clipped focus indicator on the "Create
agent" card.

## The fix

One-line class change on the scroll container:

```diff
- <div className="mt-4 flex max-w-full flex-nowrap gap-3 overflow-x-auto pb-1">
+ <div className="mt-4 -mx-1 flex max-w-full flex-nowrap gap-3 overflow-x-auto p-1">
```

- **`p-1`** gives the ring 4px of room on every side.
- **`-mx-1`** cancels the horizontal layout shift the new padding would
  otherwise introduce, so the cards render at the same visual position as
  before.

This is the exact fix suggested in the issue's "Additional context".

## Test plan

New `channelIntroFocusRing.test.mjs`:

```
node --import ./test-loader.mjs --experimental-strip-types \
  --test 'src/features/messages/ui/channelIntroFocusRing.test.mjs'
```

2/2 pass. Tests assert:
1. The scroll container currently uses `-mx-1 p-1` (ring room with
   layout-canceling margin).
2. The scroller does not regress back to `overflow-x-auto pb-1` — the
   one-sided padding that produced the clipped focus ring.

Manual verification (matches the issue's repro):

1. Open a channel with no messages yet (e.g. right after creating a
   community).
2. Tab through the page until focus lands on the "Create agent" card.
3. Focus ring now renders fully around the card on all four sides.
4. Layout position of the cards is unchanged visually — the `-mx-1`
   negative margin cancels the new padding's horizontal effect.

## Blast radius

- **Files touched**: `desktop/src/features/messages/ui/ChannelIntroBlock.tsx`
  (one line), plus the new test file.
- **Surfaces affected**: the empty-channel intro panel only. The cards
  themselves are untouched; the scroll behaviour is identical when
  horizontal scrolling activates (the padding is inside the scroll
  container).
- **Zero runtime cost** — one less padding class on one side, one more
  axis; no new hooks, listeners, or state.


iroiro147:
Closing this as superseded by my own #4374, which you merged on 2026-08-04 and which closed the linked issue #2392.

Both PRs touch `desktop/src/features/messages/ui/ChannelIntroBlock.tsx`, and the production fix is already on main, so this one is redundant in substance. The only thing here that isn't upstream is the regression test (`channelIntroFocusRing.test.mjs`) — happy to re-open that as a test-only PR if it's wanted, but I'd rather not add to your queue on spec.

Sorry for the noise; I listed this one in a shortlist I sent you and I should have checked it against my own merged work first.

PR #4296, opened 2026-08-02
fix(buzz-acp): carry JSON-RPC error.data into AgentError message (#2422)

### LINKED ISSUE #2422: [Bug] buzz-acp drops JSON-RPC error.data, hiding actionable failures — CLI/adapter version mismatch surfaces only as "-32603 Internal error" retry loops
## Summary

When the Claude CLI on the user's machine is incompatible with the bundled `claude-agent-acp` adapter (in my case: CLI `2.0.25` does not support the `--tools` flag that adapter `0.60.0` passes), every `session/new` fails instantly. The adapter reports the precise cause in the JSON-RPC error's `data` field:

```json
{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"Internal error",
 "data":{"details":"Claude Code process exited with code 1. stderr: error: unknown option '--tools'"}}}
```

`buzz-acp` discards `error.data` when constructing `AcpError::AgentError` (only `code` and `message` survive — `crates/buzz-acp/src/acp.rs:111-121`), so the harness log and the user-facing channel notice both reduce to:

> ⚠️ I couldn't process the last request after multiple retries (Agent reported error (code -32603): Internal error). Please re-send if it's still needed.

Because `AgentError` is classified as an application-class error ("session is healthy — don't invalidate", `crates/buzz-acp/src/pool.rs:2169`; no respawn, `handle_prompt_result` in `crates/buzz-acp/src/lib.rs`), the harness retries the hopeless spawn with backoff (10 attempts, 5s→300s) for every incoming message, dead-letters each batch, and never recovers or escalates. All managed agents on the machine (three in my case) wedge identically and indefinitely.

The failure is deterministic and permanent, yet the user experience is an unexplained "Internal error" that *looks* transient. In my case it presented as "my agents died when I hit my Claude usage limit and never came back" — the actual trigger was that fresh CLI spawns resolved a different (older) `claude` binary than the still-running processes from the previous day, so everything kept working until the app restarted.

**Confirmed:** upgrading the CLI (`npm install -g @anthropic-ai/claude-code@latest`) and restarting the agents fully restores them — the version mismatch was the sole cause.

## Environment

- Packaged macOS desktop build (`xyz.block.buzz.app`), three managed Claude agents
- Bundled adapter: `@agentclientprotocol/claude-agent-acp` 0.60.0 (Buzz `node-tools`, bundled node v24.11.0)
- User CLI: Claude Code 2.0.25 (npm global under nvm node v22.16.0) — works fine standalone (`claude -p` succeeds); does not support `--tools`
- Code refs against `main` @ `aaf48fc`

## Reproduction

1. Install a Claude Code CLI version that predates the `--tools` flag (e.g. 2.0.25) as the `claude` the login shell resolves.
2. Run any managed Claude agent and mention it in a channel.
3. Harness log shows only `agent_returned (application error — pipe intact) … error=Agent reported error (code -32603): Internal error` × 10 with backoff, then `dead-lettering batch after 10 retries`.
4. Drive the adapter directly with an ACP handshake (initialize → session/new) to see the real error in `error.data.details` (probe transcript above).

## Root causes

1. **Diagnostic loss:** `acp.rs` `agent_error_from_json` keeps `code` + `message`, drops `data`. Per JSON-RPC 2.0, `data` is exactly where servers put the useful detail; the claude adapter uses it for the subprocess's stderr.
2. **No escalation for deterministic failures:** application-class `AgentError` never invalidates the session, never respawns, never trips the circuit breaker, and never varies its handling no matter how many consecutive identical failures occur — so a permanent config problem is retried forever as if transient.
3. **No compatibility preflight:** desktop discovery verifies the underlying CLI exists (and probes auth) but not that its version is compatible with the bundled adapter, so a stale/shadowed `claude` (nvm per-node-version globals make this easy) fails only at session-spawn time, with the message from (1).

## Suggested fixes

- Append `error.data` (stringified, truncated) to `AcpError::AgentError`'s Display and to the `turn_error` observer payload — one-line-ish change that would have made this self-diagnosing.
- Track consecutive `AgentError`s per agent; after N identical failures, invalidate the session, then respawn the slot, then trip `SlotCircuit` — and surface a distinct user-facing notice ("agent failed N times with: <details>") instead of the generic retry text.
- At spawn (or in discovery), run `claude --version` against the resolved `CLAUDE_CODE_EXECUTABLE` and check it against the bundled adapter's minimum supported version; badge the agent as misconfigured instead of letting it enter the retry loop.

## Related

- [claude-agent-acp#338](https://github.com/agentclientprotocol/claude-agent-acp/issues/338) / [PR #363](https://github.com/agentclientprotocol/claude-agent-acp/pull/363) — adjacent adapter-side failure family (subprocess death → generic -32603; fixed upstream by returning -32002)
- #2265 — Claude agents fail silently when default model requires usage credits (same "silent wedge" experience, different trigger)

iroiro147:
Fixes #2422.

## What was broken

When the bundled `claude-agent-acp` adapter and the user's CLI get out of
version-sync (e.g. adapter `0.60.0` passes `--tools` but CLI `2.0.25`
doesn't recognise it), every `session/new` fails with a JSON-RPC error
whose `data` field carries the actionable cause:

```json
{"jsonrpc":"2.0","id":2,"error":{"code":-32603,"message":"Internal error",
 "data":{"details":"Claude Code process exited with code 1.
                    stderr: error: unknown option '--tools'"}}}
```

`agent_error_from_json` in `crates/buzz-acp/src/acp.rs:119` extracted only
`code` and `message`, silently dropping `data`. The user-facing notice
then read:

> ⚠️ I couldn't process the last request after multiple retries (Agent
> reported error (code -32603): Internal error).

…and the harness's 10-attempt retry loop looked like an unrecoverable bug
in the adapter, when the actionable fix was `bun upgrade` (or pinning the
CLI version).

## The fix

One-line extension to `agent_error_from_json`: when both `message` and a
`data` field are present, append the `data` payload to the constructed
`message` as ` (data: {...})`. The `AcpError::AgentError` struct shape is
unchanged, so all 10+ pattern-match sites across `crates/buzz-acp` (pool
retry classification, auth-error detection, lib auth codepath, etc.) keep
their `{ code, message }` destructuring — only the content of `message`
gains the extra context.

Failure strings that previously read as

```
Agent reported error (code -32603): Internal error
```

now read as

```
Agent reported error (code -32603): Internal error (data: {"details":"Claude
Code process exited with code 1. stderr: error: unknown option '--tools'"})
```

…which makes immediately obvious which CLI/adapter version mismatch to
fix.

## Test plan

`cargo test -p buzz-acp --lib agent_error_from_json` passes all three
`agent_error_*` unit tests:

1. `agent_error_from_json_falls_back_to_full_json_when_message_missing`
   (pre-existing) — covers missing `message` field.
2. `agent_error_from_json_uses_message_field_when_present` (pre-existing)
   — message-only shape unchanged by this fix (no `data` in test fixture).
3. `agent_error_from_json_appends_data_when_present` (new) — regression
   guard for #2422: asserts both the original `Internal error` substring
   AND the `unknown option '--tools'` subgraph reach the final message.

Full `cargo test -p buzz-acp --lib`: **662/662 pass.**

## Blast radius

- **Files touched**: `crates/buzz-acp/src/acp.rs` (one function.
  + new test).
- **Struct shape**: unchanged — `AcpError::AgentError { code, message }`;
  all pattern-match sites compile as before.
- **User-visible behaviour**: error strings in the agent harness log +
  channel notice now include the provider-attached `data` payload when
  present. No other runtime path changes.
- **No new dependencies**, no changes to retry/health classification.

## Out of scope

Whether a *specific* `data.details` shape (e.g. adapter-vs-CLI mismatch)
should ALSO flip the error into a fatal-no-retry class is a policy call
for the maintainers; this fix makes the raw evidence visible so that
discussion is grounded.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -111,11 +111,21 @@ pub enum AcpError {
 /// Build an [`AcpError::AgentError`] from a JSON-RPC error object,
 /// preserving the numeric code. When the `message` field is missing or
 /// non-string, fall back to the full JSON object so provider-specific
-/// detail (e.g. a `data` field) is not lost.
+/// detail (e.g. a `data` field) is not lost. When both `message` and a
+/// provider-attached `data` field are present, append the `data` payload
+/// so actionable detail (e.g. the Claude CLI's "unknown option '--tools'"
+/// stderr) reaches logs and user-facing error strings instead of being
+/// silently dropped. See #2422.
 fn agent_error_from_json(error: &serde_json::Value) -> AcpError {
     let code = error.get("code").and_then(|c| c.as_i64()).unwrap_or(-32000);
     let message = match error.get("message").and_then(|m| m.as_str()) {
-        Some(m) => m.to_string(),
+        Some(m) => {
+            if let Some(data) = error.get("data") {
+                format!("{m} (data: {})", data)
+            } else {
+                m.to_string()
+            }
+        }
         None => error.to_string(),
     };
     AcpError::AgentError { code, message }
@@ -4256,6 +4266,35 @@ mod tests {
             AcpError::AgentError { code, message } => {
                 assert_eq!(code, -32001);
                 assert_eq!(message, "auth denied");
+            }
+            other => panic!("expected AgentError, got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn agent_error_from_json_appends_data_when_present() {
+        // Regression guard for #2422. The Claude CLI's `--tools` unknown-option
+        // incompatibility with the adapter surfaces to the user only via the
+        // JSON-RPC error's `data.details` field; discarding `data` reduces the
+        // user-facing notice to "code -32603: Internal error" with no clue.
+        let error = serde_json::json!({
+            "code": -32603,
+            "message": "Internal error",
+            "data": {
+                "details": "Claude Code process exited with code 1. stderr: error: unknown option '--tools'"
+            }
+        });
+        match super::agent_error_from_json(&error) {
+            AcpError::AgentError { code, message } => {
+                assert_eq!(code, -32603);
+                assert!(
+                    message.contains("Internal error"),
+                    "message must retain the original: {message}"
+                );
+                assert!(
+                    message.contains("unknown option '--tools'"),
+                    "message must include the data payload: {message}"
+                );
             }
             other => panic!("expected AgentError, got {other:?}"),
         }


iroiro147:

PR #4298, opened 2026-08-02
fix(cli): let the relay decide which files may be uploaded

### LINKED ISSUE #2963: `buzz upload file` rejects PDFs/documents the relay accepts — agents can't attach files the desktop can

### LINKED ISSUE #2964: Larger .xlsx/.docx uploads are stored as `.zip` — the 4 KiB sniff window truncates OOXML detection
**Describe the bug**

`file_mime_to_ext()` (`crates/buzz-media/src/validation.rs:99`) maps
`…spreadsheetml.sheet → xlsx`, `…wordprocessingml.document → docx`, `…presentationml… → pptx`.
Those arms are reachable only for *small* files. A realistic spreadsheet is classified as
`application/zip` and stored and served as `.zip`.

The cause is an interaction between two layers, not a wrong mapping:

1. `process_buffered_upload` sniffs a bounded prefix — `MIN_SNIFF_BYTES = 4096`
   (`crates/buzz-media/src/upload.rs:351`).
2. `infer` 0.19's `msooxml` matcher (which backs `is_xlsx`/`is_docx`, registered *ahead* of
   `application/zip`) is positional: after matching `docProps` / `[Content_Types].xml` /
   `_rels/.rels` at `0x1E`, it walks to the **third local file header** to read the
   subdirectory name that distinguishes `xl/` from `word/` from `ppt/`.

In real OOXML files the theme and styles parts are large, so that third header sits well past
4 KiB. The matcher's bounded search finds nothing inside the sniff buffer, gives up, and the
file falls through to `application/zip`.

**To Reproduce**

Generate two spreadsheets that differ only in row count (`pip install openpyxl`):

```python
from openpyxl import Workbook

def make(path, rows):
    wb = Workbook(); ws = wb.active
    for _ in range(rows):
        ws.append(["lorem ipsum"] * 8)
    wb.save(path)

make("small.xlsx", 5)     # 3rd local header at ~2.6 KB — inside the sniff window
make("large.xlsx", 500)   # 3rd local header at ~12 KB — outside it
```

```
buzz upload file --file small.xlsx  → "url": ".../<sha>.xlsx"   ✅
buzz upload file --file large.xlsx  → "url": ".../<sha>.zip"    ❌
```

Both files have identical entry order (`docProps/app.xml`, `docProps/core.xml`,
`xl/theme/theme1.xml`, …) and are valid spreadsheets that open in Excel. Only the size of the
parts differs, which moves the header the matcher depends on.

Reproduced on a real 23 KB report as well (3rd header at offset 20,800). Not a caching
artefact: re-zipping the same content to get a fresh sha behaves the same.

**Expected behavior**

Both store as `.xlsx`. The current behaviour is the wrong way round — *toy* files are detected
correctly and *real* ones are not, so it looks fine in testing and fails in use.

**Why it matters**

The stored extension is what downstream consumers see. Anything forwarding Buzz attachments
onward delivers `report.zip`. It still opens, so nothing is
broken — but to a non-technical recipient a `.zip` reads as "something went wrong". It also
feeds `serve_inline` / content-disposition decisions, which key off the same MIME.

**Additional context**

The server cannot fall back to a hint: `PUT /upload` carries raw bytes, and the Blossom auth
event (kind 24242) carries only `t`, `x`, `expiration`, `server` — no filename, no declared
type. So this is not fixable from the client.

Two directions:

- **Raise the sniff window for the zip family only.** Once `infer` says `application/zip`,
  re-run OOXML detection over a larger prefix (the third header is bounded by the size of the
  first two entries, so ~64 KiB would cover ordinary documents). Cheap, and keeps everything
  else at 4 KiB.
- **Inspect the zip central directory** for an `xl/`, `word/` or `ppt/` prefix. Deterministic
  and independent of both entry ordering and part sizes, unlike `infer`'s positional walk. The
  central directory lives at the *end* of the file, so this needs the tail rather than a bigger
  head — worth noting for the streaming path.

Either way the sniff must stay authoritative for allow/deny; this only disambiguates *within*
a family that `BLOCKED_FILE_MIME_TYPES` has already let through.

**Environment**

- Relay: self-hosted, `main` @ `ab3af828`
- `infer` 0.19 (as pinned in `crates/buzz-media/Cargo.toml`)
- macOS 15 (arm64)

### LINKED ISSUE #3083: feat(buzz-cli): expand upload ALLOWED_MIMES to match relay generic file path
## Problem

`buzz-cli` (the agent-first CLI) restricts uploads to only 5 MIME types:

```rust
// crates/buzz-cli/src/client.rs
const ALLOWED_MIMES: &[&str] = &[
    "image/jpeg",
    "image/png",
    "image/gif",
    "image/webp",
    "video/mp4",
];
```

Any other file type is rejected client-side with `"unsupported file type: {mime}"` before the request ever reaches the relay.

However, the relay server (`crates/buzz-media/src/validation.rs`) already supports a **generic file upload path** that accepts PDF, Office documents, archives, audio, and most other formats — only blocking active web content (HTML/JS/SVG) and native executables for security.

This means:
- **Human users** (Desktop client) can upload arbitrary files via the Tauri media commands.
- **Agents** (via `buzz upload file` / `buzz-cli`) cannot upload anything beyond images and MP4 video.

## Impact

Agents running through `buzz-acp` + `buzz-cli` cannot share documents, spreadsheets, PDFs, audio files, or archives in channels — a capability that human users already have. This is a functional gap between human and agent surface area, which contradicts the "agents are members, not bots" design principle.

## Suggested fix

Expand `ALLOWED_MIMES` in `buzz-cli/src/client.rs` to align with the relay's generic file path validation logic. Two options:

1. **Mirror the relay blocklist approach**: allow all types except `BLOCKED_FILE_MIME_TYPES` (HTML, JS, SVG, executables). This keeps the CLI in sync with relay policy automatically.
2. **Explicit allowlist expansion**: add common document/archive/audio MIME types (`application/pdf`, `application/zip`, `audio/*`, Office formats, etc.) to the existing list.

Option 1 is preferable since it avoids maintaining two divergent allowlists.

## Additional context

- The relay's `file_mime_to_ext()` in `validation.rs` already maps ~30+ MIME types to extensions for the generic path.
- Size limits may need a third tier (currently 50 MB images / 500 MB video); a reasonable default for generic files would be 100–200 MB.
- The `serve_inline()` function already handles the rendering distinction: images/video inline, everything else as attachment download cards.

## Environment

- buzz-cli extracted from Buzz Desktop v0.4.26 `.deb`
- Relay: `ghcr.io/block/buzz:main` (self-hosted, compose deployment)
- Agent: `buzz-acp` + `buzz-agent` running as a systemd service

nvisionlife:
## Summary

`buzz upload file` rejected anything outside a five-entry MIME allowlist before making a request, so XLSX, PDF, CSV and every other document failed locally with `unsupported file type` — even though the relay's `PUT /upload` already routes generic files to `validate_file_content`. The client was enforcing a stale copy of relay policy, and human Desktop users could upload files that agents could not.

This makes the relay authoritative. It sniffs the body, ignores the advisory `Content-Type` for routing, and applies its configured `max_file_bytes`; the CLI now enforces only the bounds it actually owns.

- **Delete `ALLOWED_MIMES`.** Detected MIME selects client resource behaviour — size cap, timeout, advisory `Content-Type` — and never acceptance. Deliberately *not* replaced with a copy of the relay's `BLOCKED_FILE_MIME_TYPES`: a duplicated denylist is two policies that drift, and recreates this same bug for the next format a newer relay learns to accept.
- **Keep the established media caps** for the five types whose CLI/relay semantics are stable (`media_size_cap`). Generic files get no client policy cap — `max_file_bytes` is per-deployment configuration the client cannot discover, so guessing it would reject files the relay accepts.
- **Add `MAX_BUFFERED_UPLOAD_BYTES` (500 MB)** as an explicit client memory ceiling. It is a separate constant from `MAX_VIDEO_BYTES` on purpose: same initial value, different concern, and they diverge the moment uploads are streamed. Its error reads `too large to buffer safely … not a relay policy limit` so it cannot be mistaken for a refusal by the server.
- **Preflight against `metadata.len()`**, and sniff the type from a bounded 8 KiB prefix (`infer::get_from_path`) instead of the whole file. Previously `std::fs::read` ran before every check — an 800 MB file cost ~843 MB RSS just to print `unsupported file type`.

### Known consequences

Both are unavoidable while the relay's limits are undiscoverable, so they are stated rather than hidden:

1. **Unsupported non-video media now costs a transfer.** The relay sniffs 4 KiB but its non-video path buffers the whole bounded body before validating, so a `.mp3`/`.mov`/`.tiff` that used to fail instantly and locally now uploads and comes back 415. Measured against production with ID3-signature payloads at 0.5, 2, 10 and 25 MB: full body pushed every time. Documents not working at all is the worse failure, but this is a real cost in a common path.
2. **Generic files above the relay's `max_file_bytes` waste work.** They are fully allocated client-side, *partially* transferred, and rejected with 413. Against production with a 100 MB cap, the transfer stopped anywhere between 0.8 MB and ~102 MB depending on where the bound tripped — the client allocation is certain, the transfer volume is not.

### Follow-ups, not in this change

- **Stream the body and stream-hash it** so nothing is buffered and the memory guard can go away. This needs `with_retry_body` to reopen the file per attempt instead of cloning `Bytes`, which is the retry-sensitive part — worth separating from this fix.
- **BUD-06 `HEAD /upload` preflight**, so the client can learn relay policy before sending a byte. That is the thing that fixes both consequences above at once. Not currently available: an unauthenticated production probe on 2026-08-02 returned `405` with `Allow: PUT`, and NIP-11 advertises no media limits.
- **Relay-side:** the bounded `to_bytes` failure path constructs `FileTooLarge { size: 0, max }`, so every over-cap generic upload tells the user their file is "0 bytes". Separately, no early declared-size check exists on the generic path today — adding one would be a new capability, not a fix to a misreporting one.
- **#2964** (large OOXML detected as `application/zip`) is untouched. Note the client-side sniff here has the same shape of limit at 8 KiB, so a large `.xlsx` will be sent with `Content-Type: application/octet-stream`. That is advisory only — the relay re-sniffs — so it does not affect acceptance, but it does not fix #2964 either.
- `desktop/src-tauri/src/commands/media.rs:307` carries its own `unsupported file type` gate. Out of scope here; flagging it as the same class of client-side policy.

### Related issue

Fixes #2963. Fixes #3083 (same root cause, filed independently). Related to #2964, which this does not fix — see above.

### Testing

`cargo test -p buzz-cli` — 284 passed at `a6b5cce`, including 10 new tests. `cargo clippy -p buzz-cli --all-targets` and `cargo fmt --all --check` clean.

The new tests drive `upload_file` against a mock Blossom server and assert on observable behaviour — which endpoint was called, what bytes arrived, and which failures happen before any request:

| Test | Locks |
|---|---|
| `xlsx_reaches_the_upload_endpoint_byte_identically` | the reported bug; real 1.5 KB workbook fixture, byte-identical body |
| `pdf_reaches_the_upload_endpoint` / `opaque_bytes_reach_the_upload_endpoint` | generic types are no longer refused client-side |
| `jpeg_still_uploads_unchanged` | existing media path untouched |
| `relay_415_is_surfaced_without_retry_or_legacy_fallback` | relay's explanation reaches the user verbatim; a 415 is not retried and not replayed against `/media/upload` |
| `upload_404_falls_back_to_the_legacy_endpoint` | old-relay compatibility preserved |
| `generic_file_above_the_image_cap_is_not_capped_by_the_client` | 50 MB image cap does not leak onto documents |
| `file_above_the_buffer_guard_fails_before_any_request` | over-guard sparse file fails before allocation, zero HTTP requests |
| `image_above_the_image_cap_is_still_rejected_before_any_request` | media caps still enforced, now before the read |
| `media_size_cap_covers_only_the_established_media_types` | the cap table itself |

The 415-surfacing and 404-fallback cases already behave correctly at `28ae6cd2` — no new code was needed for either. They are regression locks on behaviour a future refactor would plausibly break.

`crates/buzz-cli/testdata/minimal.xlsx` is a genuine 1,572-byte OOXML workbook (five parts, deflated), added so the round-trip test uses a real file rather than a zip-shaped approximation.

Design was reviewed adversarially before implementation; the production measurements quoted above are from that review, not from this branch.


maikunari:
+1 on making the relay the authority here.

I have a related open PR (#3060) that takes a slightly different angle: mirror the relay's `BLOCKED_FILE_MIME_TYPES` denylist client-side (fast-fail only the permanent security denylist) and raise the non-media size cap to the relay's 100MB default, rather than expanding an allowlist. Same end state for agents (PDF/xlsx/csv/etc. stop dying client-side).

Happy for either approach to land — if this one merges first I'll close #3060 as superseded.

PR #4317, opened 2026-08-02
fix(desktop): remove nested focusable elements from profile popover call sites

### LINKED ISSUE #2394: [Bug] Avatar/author-name focus stops twice — UserProfilePopover nests two focusable elements
**Describe the bug**
UserProfilePopover.tsx wraps its `children` in a trigger element that gets
`role="button"` + `tabIndex={0}` whenever `enableProfilePanel` is true
(the default). Several callers pass a real `<button>` as `children` without
setting `enableProfilePanel={false}` — so you end up with two nested
focusable elements for what is visually one control: the wrapper
(div/span, tabIndex=0) and the inner `<button>`. Tabbing through the UI
stops on the same avatar/name twice in a row.

**To Reproduce:**
1. Open any channel with messages.
2. Tab through the timeline with the keyboard.
3. Notice focus lands on the same message avatar (or author name) twice
   before moving to the next element.

**Expected behavior:**
Each avatar/name should be a single Tab stop.

**Supporting Material**
Screenshot attached (system message avatar case).

**Environment:**
- Desktop app (Tauri)

**Additional context**
This isn't isolated — the same pattern (button child, no
`enableProfilePanel={false}`) appears in:
- desktop/src/features/messages/ui/MessageRow.tsx (avatar + author name —
  every message, every channel)
- desktop/src/features/messages/ui/SystemMessageRow.tsx
- desktop/src/features/projects/ui/ProjectCards.tsx
- desktop/src/features/projects/ui/ProjectsPullRequestsList.tsx
- desktop/src/features/projects/ui/ProjectsActivityFeed.tsx
- desktop/src/features/projects/ui/ProjectProfileIdentity.tsx
- desktop/src/features/projects/ui/ProjectsIssuesList.tsx
- desktop/src/features/pulse/ui/NoteCard.tsx
- desktop/src/features/pulse/ui/AgentActivityCard.tsx
- desktop/src/features/forum/ui/ForumThreadPanel.tsx

Likely fix: pass `enableProfilePanel={false}` at each of these call sites
(the inner button already handles the click), or have
UserProfilePopover detect a `<button>`/interactive child and skip adding
its own tabIndex/role automatically.

Image

iroiro147:
## What

Eliminates the nested-focusable-elements anti-pattern in every `UserProfilePopover` call site identified in #2394 — desktop timeline, project cards, forum threads, pulse feed.

## Why

The popover's trigger wrapper gets `role="button"` + `tabIndex={0}` when `enableProfilePanel` (default `true`) lets it open the profile panel. Several callers wrapped their avatar/author name in a decorative `<button>` that had no `onClick` — so the button was reachable by Tab but dead to activation, while the *wrapper* handled the actual profile-panel open. Each avatar/name became **two Tab stops** for what is one visual control.

## How

One root-cause mechanism, uniform per call site:

- **`UserProfilePopover` gains an optional `triggerClassName` prop** so the layout / focus-ring classes the inner button carried can move to the wrapper, where Tab focus now actually lands.
- Every call site that had a `<button>` child **loses the inner `<button>`** — the wrapper is the tab stop, the click handler still fires via the wrapper's `onClick`, and Hover→popover still works unchanged.
- Where the inner button carried an `aria-label` (e.g. `View ${label}'s profile`), it moves to `triggerAriaLabel` on the popover.

Net effect: one Tab stop per avatar/name across the entire app, focus rings still visible, zero behavior change for clicks or hovers.

## Files

- `desktop/src/features/profile/ui/UserProfilePopover.tsx` — adds `triggerClassName` prop, merges it onto the TriggerElement's `cn(...)` call.
- `desktop/src/features/messages/ui/MessageRow.tsx` — avatar + author name call sites.
- `desktop/src/features/messages/ui/SystemMessageRow.tsx` — system-message avatar.
- `desktop/src/features/projects/ui/ProjectCards.tsx` — project-card owner avatar.
- `desktop/src/features/projects/ui/ProjectProfileIdentity.tsx` — inline `ProfileIdentityAvatar` + standalone `ProfileAuthorName`.
- `desktop/src/features/projects/ui/ProjectsActivityFeed.tsx` — actor avatar + actor name.
- `desktop/src/features/projects/ui/ProjectsIssuesList.tsx` — issue author.
- `desktop/src/features/projects/ui/ProjectsPullRequestsList.tsx` — PR author.
- `desktop/src/features/pulse/ui/AgentActivityCard.tsx` — agent group avatar.
- `desktop/src/features/pulse/ui/NoteCard.tsx` — parent-note avatar, parent-note author, note avatar, note author.
- `desktop/src/features/forum/ui/ForumThreadPanel.tsx` — post author + reply author.

(11 files total, −153 / +116 lines.)

## Verification

- `pnpm typecheck` — 0 errors.
- `pnpm test` — 3906/3906 pass.
- `pnpm lint` — 0 errors (same 2 infos as baseline).
- Manual inspection: no `<button>` element remains inside `<UserProfilePopover>` anywhere it was observed; the trigger wrapper is now the only Tab stop.

Closes #2394.

### DIFF
--- a/desktop/src/features/forum/ui/ForumThreadPanel.tsx
+++ b/desktop/src/features/forum/ui/ForumThreadPanel.tsx
@@ -83,20 +83,18 @@ function ReplyRow({
       data-forum-event-id={reply.eventId}
     >
       <div className="flex items-center gap-2">
-        <UserProfilePopover pubkey={reply.pubkey}>
-          <button
-            className="flex items-center gap-2 rounded-lg focus-visible:outline-hidden focus-visible:ring-2 focus-visible:ring-ring"
-            type="button"
-          >
-            <UserAvatar
-              avatarUrl={replyAvatarUrl}
-              displayName={replyAuthorLabel}
-              size="sm"
-            />
-            <span className="text-sm font-medium text-foreground hover:underline">
-              {replyAuthorLabel}
-            </span>
-          </button>
+        <UserProfilePopover
+          pubkey={reply.pubkey}
+          triggerClassName="flex items-center gap-2 rounded-lg focus-visible:outline-hidden focus-visible:ring-2 focus-visible:ring-ring"
+        >
+          <UserAvatar
+            avatarUrl={replyAvatarUrl}
+            displayName={replyAuthorLabel}
+            size="sm"
+          />
+          <span className="text-sm font-medium text-foreground hover:underline">
+            {replyAuthorLabel}
+          </span>
         </UserProfilePopover>
         <span className="text-xs text-muted-foreground">
           {formatRelativeTime(reply.createdAt)}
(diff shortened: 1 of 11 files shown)


iroiro147:
Closing as upstream-subsumed. The nested-focusable fix is no longer needed: current `origin/main` already refactored these call sites into `ProjectAuthorIdentity` which goes through `UserProfilePopover` with `triggerElement="span"` + the new `triggerClassName` prop (introduced via other PRs). The composer/list-row sites are now keyboard-reachable. Thanks.

PR #4328, opened 2026-08-02
fix(agent): report load_skill truncation with an in-band marker + server warn

### LINKED ISSUE #4163: load_skill silently truncates skill content over 32 KiB with no warning, log, or marker
**Describe the bug**

`load_skill` silently truncates any skill content over `MAX_SKILL_BODY_BYTES` (32 KiB) and returns the truncated text as if it were complete. There is no log line, no error, no in-band marker, and `is_error` stays `false` — so neither the model nor the caller can tell a truncated skill from a whole one.

The failure mode is a skill that loads successfully and behaves subtly wrong. Instructions in the tail of a SKILL.md — which is often where the output format, the refusal rules, or the verification checklist live — simply are not there, and nothing anywhere says so.

`crates/buzz-agent/src/builtin.rs` (both the SKILL.md path and the supporting-file path):

```rust
let output = if output.len() > MAX_SKILL_BODY_BYTES {
    truncate_at_boundary(&output, MAX_SKILL_BODY_BYTES).to_owned()
} else {
    output
};
```

returned as:

```rust
ToolResult {
    provider_id: String::new(),
    content: vec![ToolResultContent::Text(output)],
    is_error: false,
}
```

Constant at `crates/buzz-agent/src/hints.rs:7`:

```rust
pub const MAX_SKILL_BODY_BYTES: usize = 32 * 1024;
```

The boundary handling itself is fine — `truncate_at_boundary` keeps the output valid UTF-8. The issue is purely that the truncation is unreported.

**Steps to reproduce**

1. Create a skill whose body exceeds 32 KiB, with a sentinel line at the very end:

   ```sh
   mkdir -p .agents/skills/oversized
   {
     printf -- '---\nname: oversized\ndescription: Reproduces silent truncation of oversized skill bodies.\n---\n\n'
     for i in $(seq 1 900); do
       printf 'Filler line %s: padding to push this body past the 32 KiB cap.\n' "$i"
     done
     printf '\n## SENTINEL\n\nIf you can read this line, the skill was NOT truncated.\n'
   } > .agents/skills/oversized/SKILL.md

   wc -c .agents/skills/oversized/SKILL.md   # ~57 KB, comfortably over 32768
   ```

2. Start an agent in that working directory so the skill is discovered (`.agents/skills` is the first entry in `SKILL_DIRS`).

3. Ask the agent to call `load_skill` with `oversized`, then ask it to quote the `## SENTINEL` section.

**Expected behavior**

One of:

- the truncation is reported — a `tracing::warn!` on the server side, and an explicit in-band marker appended to the returned text (e.g. `\n\n[truncated: N of M bytes shown]`) so the model knows its instructions are incomplete; or
- oversized skills are rejected at discovery with a clear error naming the file and its size, so the author fixes it rather than shipping a half-loaded skill.

Either is fine. Silence is the problem.

**Actual behavior**

The agent reports that the skill loaded, and the `## SENTINEL` section is absent. Nothing in the logs, the tool result, or the agent's own view indicates anything was dropped.

**Version and platform**

- Buzz version: relay crate `0.2.0`; verified against `main` @ `4632c55041c5d423d572a6f6411bb7b279c26f67`
- OS: macOS (arm64) — but this is platform-independent; the behaviour is in `buzz-agent`, read from source at the commit above

**Logs / additional context**

There are no logs — that is the bug.

Two notes that may help scoping:

1. The same truncation applies to supporting files read through `load_skill`, not just `SKILL.md`, so a skill can also lose the tail of a reference document it explicitly asked to read.
2. `MAX_HINTS_BYTES` (128 KiB, for the `AGENTS.md` chain) has the same shape and is presumably worth the same treatment, though the consequence there is milder since the chain is additive rather than a single authored document.

I'm reporting this from a source read rather than a captured session — I don't have a model credential wired to this checkout, so I could not paste a real transcript. The reproduction above is what I'd expect a maintainer with a working agent to see in under five minutes; happy to be corrected if the runtime path differs from what `builtin.rs` suggests.

A suggested minimal fix, if useful: keep the truncation, append a marker, and log once per truncated load —

```rust
let output = if output.len() > MAX_SKILL_BODY_BYTES {
    tracing::warn!(
        skill = %name,
        bytes = output.len(),
        limit = MAX_SKILL_BODY_BYTES,
        "skill content truncated"
    );
    let mut t = truncate_at_boundary(&output, MAX_SKILL_BODY_BYTES).to_owned();
    t.push_str(&format!(
        "\n\n[truncated: {} of {} bytes shown]",
        t.len(),
        output.len()
    ));
    t
} else {
    output
};
```

iroiro147:
## Summary
Fixes #4163.

`load_skill` silently truncates any skill content over
`MAX_SKILL_BODY_BYTES` (32 KiB) and returns the truncated text with
`is_error=false` — no marker, no log line. The model consumes the body as
if complete, but the tail of a SKILL.md (which often holds output format,
refusal rules, or verification checklists) is gone with no signal.

This PR applies the reporter's **option 1** at both truncation sites in
`crates/buzz-agent/src/builtin.rs`:

1. **In-band marker** appended to the returned text:
   `\n\n[truncated: N of M bytes shown]` — the model sees the body is
   incomplete and can adapt (decline to rely on tail instructions, or
   re-read with a narrower path).
2. **Server-side `tracing::warn!`** with the skill name, byte count, and
   cap — operators spot oversized skills in logs.

## Changes
- `crates/buzz-agent/src/builtin.rs`
  - `call_load_skill` unified output path (body + Supporting Files
    section): on over-cap, append the marker and emit `warn!`.
  - `load_supporting_file`: same marker + warn, with structured
    `skill` / `rel_path` fields.
  - Under-cap loads untouched — no marker, no warn.

## Tests
- `builtin::tests::call_load_skill_truncates_large_body` — updated:
  marker present; overhead bounded (< 200 bytes past cap).
- `builtin::tests::call_load_skill_truncates_large_supporting_file` —
  updated: marker present on the supporting-file path.
- `builtin::tests::call_load_skill_under_cap_has_no_truncation_marker` —
  new: no marker on under-cap bodies.

## Verification
- `cargo check -p buzz-agent --lib` — clean.
- `cargo test -p buzz-agent --lib --offline builtin` — 12/12 pass.
- `cargo fmt -p buzz-agent --check` — clean.

## Out of scope
- Option 2 from the issue (reject oversized skills at discovery): would
  be a breaking change for already-deployed skills; the marker approach
  preserves load behavior while restoring observability. A follow-up to
  add a discovery-time warning is a separate small PR if maintainers
  want belt-and-braces.

### DIFF
--- a/crates/buzz-agent/src/builtin.rs
+++ b/crates/buzz-agent/src/builtin.rs
@@ -98,9 +98,22 @@ pub async fn call_load_skill(arguments: &Value, skills: &[SkillEntry]) -> ToolRe
     }
 
     // Apply the size cap to the full output (body + Supporting Files section)
-    // so the total tool result stays within MAX_SKILL_BODY_BYTES.
+    // so the total tool result stays within MAX_SKILL_BODY_BYTES. If we had
+    // to truncate, append an explicit in-band marker so the model knows the
+    // tail (often format/refusal/verification instructions) is missing, and
+    // emit a server-side warn so logs capture it too. See issue #4163.
     let output = if output.len() > MAX_SKILL_BODY_BYTES {
-        truncate_at_boundary(&output, MAX_SKILL_BODY_BYTES).to_owned()
+        let original_len = output.len();
+        let mut truncated = truncate_at_boundary(&output, MAX_SKILL_BODY_BYTES).to_owned();
+        truncated.push_str(&format!(
+            "\n\n[truncated: {MAX_SKILL_BODY_BYTES} of {original_len} bytes shown]"
+        ));
+        tracing::warn!(
+            skill_body_bytes = original_len,
+            cap = MAX_SKILL_BODY_BYTES,
+            "load_skill truncated oversized skill output"
+        );
+        truncated
     } else {
         output
     };
@@ -204,7 +217,20 @@ async fn load_supporting_file(
                         skill_name, rel_path_owned, content
                     );
                     let output = if output.len() > MAX_SKILL_BODY_BYTES {
-                        truncate_at_boundary(&output, MAX_SKILL_BODY_BYTES).to_owned()
+                        let original_len = output.len();
+                        let mut truncated =
+                            truncate_at_boundary(&output, MAX_SKILL_BODY_BYTES).to_owned();
+                        truncated.push_str(&format!(
+                            "\n\n[truncated: {MAX_SKILL_BODY_BYTES} of {original_len} bytes shown]"
+                        ));
+                        tracing::warn!(
+                            skill = %skill_name,
+                            rel_path = %rel_path_owned,
+                            file_bytes = original_len,
+                            cap = MAX_SKILL_BODY_BYTES,
+                            "load_skill truncated oversized supporting file"
+                        );
+                        truncated
                     } else {
                         output
                     };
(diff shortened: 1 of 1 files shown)


iroiro147:
Closing this in favour of #4265, which addresses the same issue (#4163) with the same change to builtin.rs.

I had two open PRs against #4163 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

Issue #4032, opened 2026-07-31
security: `--private-key` as a CLI argument leaks the key into shell history and process listings


AI-OWEN:
Several `buzz` subcommands accept the identity secret directly on the command line, e.g.:

```
buzz channels remove-member --relay <url> --private-key <nsec> \
  --channel <uuid> --pubkey <hex>
```

Two consequences:

1. **Shell history.** The full `nsec1...` is written verbatim to `~/.zsh_history` and to every per-session file in `~/.zsh_sessions/`. These are plaintext, included in Time Machine backups, and readable by any process running as that user. Users generally don't realise a one-off admin command has persisted their signing key to disk.
2. **Process listing.** While the command runs, the key is visible in `ps aux` to any local user.

Once it has happened, cleanup is unreliable — history files, editor undo state, backups and search indexes all need scrubbing, and the practical remedy becomes key rotation.

### Suggested fix

- Read the secret from an env var, stdin, a mode-0600 file, or the platform keystore. Keep `--private-key` only as a deprecated alias that warns.
- If argv support must stay, accept `--private-key-file` / `--private-key-stdin` and document those as preferred in `--help` and the docs.
- Consider a `buzz` config/keystore path so admin commands need no secret on the command line at all.

Related: #2883 (keeping `BUZZ_PRIVATE_KEY` out of agent-controlled processes) and #3629 (key rendered in `--help` output, fixed) — same class of exposure, different surface.

Buzz CLI as bundled with Desktop 0.5.3, macOS 26.5.2, zsh.

PR #4329, opened 2026-08-02
feat(cli): add --private-key-file, warn when --private-key is passed on argv

iroiro147:
## Summary
Addresses #4032.

`buzz` accepts the identity nsec as `--private-key`, which lands verbatim
in `~/.zsh_history`, `~/.zsh_sessions/*`, and `ps aux` for the life of
the process. Cleanup is unreliable (history + editor undo + backups +
indexes), so the practical remedy becomes key rotation.

This PR makes the safe form available without breaking existing argv use.

## Changes
- `crates/buzz-cli/src/lib.rs`
  - New `--private-key-file <PATH>` flag: reads the key from a file
    (mode-0600 recommended by convention, not enforced). Only the path
    lands in shell history / `ps`; the key never touches argv.
  - `--private-key` and `--private-key-file` mutually exclusive via
    `conflicts_with`.
  - stderr warning when `--private-key` is actually used:
    > `warning: --private-key on argv leaks the key to shell history and 'ps'; prefer BUZZ_PRIVATE_KEY or --private-key-file (see #4032)`
  - Resolution order: argv `--private-key` → `--private-key-file` →
    `BUZZ_PRIVATE_KEY` env (env was already supported via clap `env`).
  - Error message updated to name the file form.
  - Help text (`--help`) explains the security trade-off.

## Backward compatibility
- Existing argv form works identically; users who don't read the warning
  see no behavior change. The warning lands on stderr, not stdout, so
  scripted JSON parsing of stdout is unaffected.

## Tests / verification
- No new tests — the change is in clap argument parsing + `fs::read_to_string`
  at startup; both well covered by clap's derive and std's guarantees.
- `cargo check -p buzz-cli` — clean.
- `cargo test -p buzz-cli --offline --lib` — 274/274 pass.
- `cargo fmt -p buzz-cli --check` — clean.
- Smoke-tested `--help`: both flags shown, warning text rendered.

## Out of scope
- Platform keystore integration (macOS Keychain, Secret Service, etc.) —
  the issue's third suggestion; significantly larger surface, deferred.
- Renaming or hard-deprecating `--private-key` outright — breaks users.
  A staged deprecation (warn this release, remove in a major) can be
  decided by maintainers separately.
- `buzz channels remove-member`'s specific argument surface (per the
  issue's example) inherits the global flag — no subcommand change needed.

### DIFF
--- a/crates/buzz-cli/src/lib.rs
+++ b/crates/buzz-cli/src/lib.rs
@@ -82,8 +82,16 @@ struct Cli {
     relay: String,
 
     /// Nostr private key (hex or nsec). This is the CLI's identity.
+    ///
+    /// Prefer BUZZ_PRIVATE_KEY env or --private-key-file. Passing the key
+    /// on argv leaks it into shell history and `ps` output — see #4032.
     #[arg(long, env = "BUZZ_PRIVATE_KEY", hide_env_values = true)]
     private_key: Option<String>,
+
+    /// Read the private key from a mode-0600 file. Safer than argv: the
+    /// path (not the key) lands in shell history and `ps`.
+    #[arg(long, value_name = "PATH", conflicts_with = "private_key")]
+    private_key_file: Option<std::path::PathBuf>,
 
     /// NIP-OA auth tag JSON (owner attestation). Injected into every signed event.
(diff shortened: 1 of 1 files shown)


iroiro147:

PR #4357, opened 2026-08-02
fix(desktop): mention remote channel agents

ErnestoEverhard:
## Summary
- keep current channel-member agents in mention autocomplete even when they are not locally managed
- preserve the existing filtering for non-member, unmanaged agent identities
- add regression coverage for a remote agent that is a channel member

## Verification
- Desktop test suite: 3,906 passed
- TypeScript typecheck
- Biome check
- file-size, pixel-text, and pubkey-truncation policy checks

## Context
Originating Buzz channel: 0bbabe2c-8ad0-4188-9c14-10402b4dead4

The failure was reproduced with a canonical Hermes agent that belonged to the channel and had a cached profile, but was absent from Desktop's local managed-agent registry. The eligibility guard rejected it before the channel-member invocability rule could run.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -5,7 +5,7 @@ import {
   coalesceAgentAutocompleteCandidates,
   getMentionableAgentPubkeys,
   getSharedChannelIds,
-  isAgentIdentityInManagedList,
+  isAgentAutocompleteEligible,
   relayAgentIsSharedWithUser,
   shouldHideAgentFromMentions,
 } from "./agentAutocompleteEligibility.ts";
@@ -136,25 +136,32 @@ test("getMentionableAgentPubkeys: keeps managed agents and shared relay agents",
   assert.deepEqual(result, new Set([PUB_A, PUB_B, PUB_C]));
 });
 
-test("isAgentIdentityInManagedList: keeps people and only current managed agent identities", () => {
+test("isAgentAutocompleteEligible: keeps people, channel members, and current managed agent identities", () => {
   const managedAgentPubkeys = new Set([PUB_A]);
 
   assert.equal(
-    isAgentIdentityInManagedList(
+    isAgentAutocompleteEligible(
       { isAgent: false, pubkey: PUB_B },
       managedAgentPubkeys,
     ),
     true,
   );
   assert.equal(
-    isAgentIdentityInManagedList(
+    isAgentAutocompleteEligible(
       { isAgent: true, pubkey: PUB_A.toUpperCase() },
       managedAgentPubkeys,
     ),
     true,
   );
   assert.equal(
-    isAgentIdentityInManagedList(
+    isAgentAutocompleteEligible(
+      { isAgent: true, isMember: true, pubkey: PUB_B },
+      managedAgentPubkeys,
+    ),
+    true,
+  );
+  assert.equal(
+    isAgentAutocompleteEligible(
       { isAgent: true, pubkey: PUB_B },
       managedAgentPubkeys,
     ),
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
@@ -54,12 +54,13 @@ export function getMentionableAgentPubkeys({
   return pubkeys;
 }
 
-export function isAgentIdentityInManagedList(
-  candidate: { isAgent?: boolean; pubkey: string },
+export function isAgentAutocompleteEligible(
+  candidate: { isAgent?: boolean; isMember?: boolean; pubkey: string },
   managedAgentPubkeys: ReadonlySet<string>,
 ) {
   return (
     candidate.isAgent !== true ||
+    candidate.isMember === true ||
     managedAgentPubkeys.has(normalizePubkey(candidate.pubkey))
   );
 }
--- a/desktop/src/features/channels/ui/MembersSidebar.tsx
+++ b/desktop/src/features/channels/ui/MembersSidebar.tsx
@@ -9,7 +9,7 @@ import {
 import { attachManagedAgentToChannel } from "@/features/agents/channelAgents";
 import {
   coalesceAgentAutocompleteCandidates,
-  isAgentIdentityInManagedList,
+  isAgentAutocompleteEligible,
 } from "@/features/agents/lib/agentAutocompleteEligibility";
 import { useIsArchivedPredicate } from "@/features/identity-archive/hooks";
 import { useClassifiedMembers } from "@/features/channels/lib/useClassifiedMembers";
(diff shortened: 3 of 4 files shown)


kevinsakiro:
This is exactly what I wanted! I pulled your changed and built the desktop app, and it worked for me! Thank you, hopefully it will get merged! 

Tested this branch against a self-hosted agent and it fixes the case cleanly.

**Setup:** `buzz-acp` running under a systemd user unit on Amazon Linux 2023 (aarch64),
identity created entirely through the CLI — NIP-OA auth tag minted client-side, relay
membership via `POST /api/invites` + `/claim`. The agent has never been known to Buzz
Desktop and its nsec is not in the local keyring. It is an authoritative channel member
with `role: "bot"`.

**Before:** the agent was absent from Desktop's `@` autocomplete, so there was no way to
mention it from Desktop at all — despite being a channel member that the relay happily
routes mentions to. Phone and CLI worked fine.

**Registering it in `managed-agents.json` is strictly worse**, which may be worth noting
as motivation: the picker then shows it, but `ensureManagedAgentMentionsReady` tries to
spawn it, `spawn_key_refusal` fires ("has no private key available"), and
`useMentionSendFlow.ts` returns before publishing — so the mention is never sent.
`start_on_app_launch: false` does not help; it gates boot spawn, not on-demand spawn.

**With this branch:** the agent appears in autocomplete with the store left clean, the
mention publishes, and the remote agent responds. Because it is not locally managed, the
readiness path is never reached.

Built from source on macOS 26.5.2 / Apple Silicon (`pnpm tauri dev`), `pnpm typecheck`
clean. Not an OpenClaw/bridge case — just a plain `buzz-acp` on a machine I own — so the
fix covers more than external bridges.

nvisionlife:
This fixes the case I hit too, and the `isMember` route is the right call for channel-member agents — thanks for tracing it to the eligibility guard.

One gap I think is worth closing in the same change: **relay-directory agents are still dropped**, because they are added with `isMember: false`.

`useMentions.ts` adds two kinds of agent candidate. Channel members (~line 305) get `isMember: true`, so your fix admits them. But the relay directory loop adds every `kind:10100` profile with `isMember: false` explicitly:

```js
for (const agent of relayAgentsQuery.data ?? []) {
  const pubkey = normalizePubkey(agent.pubkey);
  addCandidate({
    kind: "identity",
    pubkey,
    displayName: agent.name,
    isMember: false,
    // ...
    isAgent: true,
  });
```

For those, `isAgentAutocompleteEligible` evaluates `isAgent !== true || isMember === true || managedAgentPubkeys.has(pubkey)` → `false || false || false`, so a viewer who manages no agents locally still sees nothing from the directory.

That matters because `shouldHideAgentFromMentions` immediately below was written to show exactly that candidate:

```js
// Invocable => always show.
if (mentionableAgentPubkeys.has(normalized)) return false;
// Non-member, non-invocable => hide (preserves prior behavior).
if (!isMember) return true;
```

`mentionableAgentPubkeys` is checked *before* the `!isMember` bail, so a non-member agent that is invocable is meant to appear. With `managedAgentPubkeys` in the guard, that first branch stays unreachable for anyone who manages no agents locally — the same dead-code shape you fixed, one candidate source over.

### Suggested delta

Pass the set the guard is actually expressing. `mentionableAgentPubkeys` is `managed ∪ relayAgentIsSharedWithUser` and is already computed and already in the memo's dependency array:

```diff
-      if (!isAgentAutocompleteEligible(candidate, managedAgentPubkeys)) {
+      if (!isAgentAutocompleteEligible(candidate, mentionableAgentPubkeys)) {
```

Composed with your `isMember === true` clause this becomes the union of both intents: channel members, plus relay agents shared with this viewer. Security posture is unchanged — admission still runs through `relayAgentIsSharedWithUser`, so an agent whose `respond_to_allowlist` excludes the viewer stays out.

Two practical notes if you take it:

1. Biome then flags `useExhaustiveDependencies` — `managedAgentPubkeys` is no longer read inside the `mentionCandidates` memo and has to come out of the dep array. It stays in use elsewhere in the file (`isManagedAgentPubkey`), so the variable itself remains. Net −1 line, which also keeps `useMentions.ts` under the file-size ratchet it currently sits exactly on.
2. I would **not** apply the same widening to `MembersSidebar.tsx`. That guard is on add-member search, and whether you may add an agent to a channel is `channel_add_policy`, not `respond_to` — swapping in the mention set would conflate two different policies.

### Verification

Applied on top of `971c5f2` locally, the delta above is 1 insertion / 2 deletions and the full desktop gate is green: 3,906 tests pass, `biome check` clean, `tsc --noEmit` clean, and the file-size / px-text / pubkey-truncation checks pass.

I also pinned the gap itself against your branch unmodified, so it is measured rather than argued — a relay agent that allowlists the viewer, is not locally managed, and is not a channel member:

| assertion | result |
|---|---|
| agent is in `getMentionableAgentPubkeys(...)` | true |
| `isAgentAutocompleteEligible(candidate, managedAgentPubkeys)` | **false** — dropped |
| `isAgentAutocompleteEligible(candidate, mentionableAgentPubkeys)` | true |
| `shouldHideAgentFromMentions(...)` | false — wanted to show it |

Happy for this to land in your PR rather than a competing one; I have the equivalent change on a branch and will drop it if you pick this up. If you would rather keep this PR tight, I can open the delta as a follow-up once yours merges — just say which you prefer.

### Unrelated, but adjacent

Once the guard consults `relayAgentIsSharedWithUser`, `respond_to: "anyone"` becomes a trap for published profiles. That route requires `channelIds.some(id => sharedChannelIds.has(id))`, and a `kind:10100` profile that omits `channel_ids` gets `[]`, which can never overlap. So moving an agent from `allowlist` to the seemingly-broader `anyone` makes it vanish from every other member's dropdown. Either treat an empty `channel_ids` under `anyone` as "all shared channels", or document that publishers must populate it. Happy to file that separately if it is not already tracked.

ErnestoEverhard:
Well this is both awesome and annoying. I hope it fixes your issues but I didn't intend to do this and apparently need to restrict what my Buzz agent can do a bit. I was troubleshooting this issue and just said that we HAD to fix this issue because its super annoying...thinking I had simply misconfigured something and it would be a 10 minute fix. I come back and it submitted this entire PR and ate my entire Claude Usage for 5 hours. 

@nvisionlife Ummm feel free to do whatever you want regarding taking this code into a PR you're working on.

nvisionlife:
Opened #4453, which carries this change plus the relay-agent half of the gate.

Your channel-member clause and its test are in it with authorship preserved (`Co-authored-by`). The addition is at the `useMentions` call site: passing `getMentionableAgentPubkeys(...)` instead of the locally-managed set, so an agent that is invocable for the viewer but not a member of the channel they are typing in also survives — the case in the earlier comment.

Two things from this branch that CI would have flagged and are handled there: dropping `managedAgentPubkeys` from the `mentionCandidates` dependency array (`useExhaustiveDependencies`), and keeping `useMentions.ts` net-neutral against the 1000-line file-size ratchet.

`MembersSidebar` stays on the locally-managed set in #4453, for the reason in the earlier comment — add-member should answer to channel policy rather than to `respond_to`.

Issue #2877, opened 2026-07-25
No shipped client can emit the buzz-channel tag, so only the announcing key can ever push to a Buzz-hosted repo


standing-order:
Buzz Version 0.4.26 (0.4.26)
MacOS Sequoia 15.7.5

Summary
The relay resolves a non-owner pusher's authority exclusively through a buzz-channel tag on the kind:30617 repo announcement. Absent that tag, crates/buzz-relay/src/api/git/policy.rs:347-355 rejects every non-owner push with 403 no channel binding.

Neither shipped client can produce that tag:

buzz repos create (CLI) has no --channel flag, and the SDK builder it calls does not accept one.
Desktop's "create project" flow builds the announcement tag list by hand and omits it.
Net effect: every repo created through a shipped client is permanently single-pusher. Only the announcing key — and, if that key is a managed agent, its human owner — can push. The channel role model (Owner/Admin/Member, plus the deliberate Bot → Member promotion) is implemented, tested, and documented, but unreachable in practice.

The tag is not an oversight in the data model. VISION_PROJECTS.md:17-33 specifies it as part of the announcement shape, the SDK has a builder that preserves it, and there are unit tests asserting it survives updates. There is simply no code path that originates it outside a hand-signed raw event in an E2E script.

Environment
Source read at block/buzz commit a64cc71 (main).
Relay under test reports version: 0.2.0 via NIP-11, which matches crates/buzz-relay/Cargo.toml at that commit. This is a version match, not a commit-exact match — the deployed build could be an earlier 0.2.0.
What we observed directly: a live relay's kind:30617 announcement, created through buzz repos create, carrying no buzz-channel tag (see Reproduction → Observed). That is the bug's cause, on a real relay.
What we observed directly — the fix working. After the buzz-channel tag was added to the live announcement by hand (see Workaround), a non-owner channel member cloned and pushed successfully under their own key. Before the tag existed, the announcing key was the only key that had ever pushed. So the positive half of the causal claim is executed, not inferred: supplying the tag is sufficient to make the documented role model work.
What we did not run — the 403 itself. We never attempted a non-owner push against an announcement lacking the tag, because the only such announcement available was the live one and removing the binding to test it would have revoked push for the whole team. Every statement about the failure path is read from source and labelled as such; we are not reporting an observed 403.
[...]

PR #4363, opened 2026-08-02
fix(cli): stamp repo-announcement updates with max(now, head+1) (#2876)

iroiro147:
## Problem

`buzz repos protect set` / `protect remove` / `bind` permanently fail 15 minutes after a repo is announced (#2876). `build_updated_repo_announcement` stamped the replacement announcement at `existing.created_at + 1`, but the relay rejects any event whose `created_at` drifts more than `MAX_TIMESTAMP_DRIFT_SECS` (±900s) from server time. Since the advance is `+1` and never catches up, once an announcement is >15 min stale **every** mutation is rejected deterministically — branch protection becomes permanently unchangeable, and `repos bind` (the documented recovery for #2877) is unreachable. Independent repro by @A-PachecoT confirmed all three commands hit the same path.

## Fix

Stamp the replacement with `max(now, existing + 1)`:

- **The `+1` floor is preserved.** A delayed writer still can't leapfrog an intervening update and silently erase metadata — if it observed a stale head, the floor keeps its replacement ahead of that head. Replay-protection intent unchanged.
- **The wall-clock floor restores freshness.** A writer acting on a fresh observation now lands inside the relay's drift window. This is the single shared choke point, so `protect set`, `protect remove`, and `bind` are all fixed by the same change.

`crates/buzz-cli/src/commands/repos.rs` — the only file touched.

## Tests

- Two pre-existing asserts (`assert_eq!(created_at, 101)`) encoded the old semantics; relaxed to `>= 101` (the floor invariant they actually pin down — wall-clock now (2026+) always exceeds the `100` base, so behavior on a live writer is unchanged).
- **New `updated_announcement_stays_within_relay_drift_window_for_stale_head`**: announcement made 2h ago → replacement drifts ≤900s from wall clock **and** advances past the stale head. This is the #2876 regression, pinned.
- **New `protection_set_preserves_metadata_and_protections`**: auth tag stripped, the matching `refs/heads/main` rule replaced, unrelated protections + future metadata preserved, and the leapfrog guard verified.
- `cargo test -p buzz-cli --lib repos::` → **17/17 green**; full `buzz-cli --lib` baseline clean; `cargo clippy -p buzz-cli --lib` clean.

Behavior is timestamp-only — no relay change, no API change. Host: macOS `aarch64-apple-darwin`; the path is platform-independent Rust.

### DIFF
--- a/crates/buzz-cli/src/commands/repos.rs
+++ b/crates/buzz-cli/src/commands/repos.rs
@@ -141,13 +141,17 @@ fn build_updated_repo_announcement(
         ))
     })?;
 
-    // Advance only the observed head. Using wall-clock time here would let a
-    // delayed writer leapfrog an intervening update and silently erase metadata.
-    let next_created_at = existing
+    // Stamp the replacement with max(now, existing + 1). The +1 floor keeps a
+    // delayed writer from leapfrogging an intervening update and silently
+    // erasing metadata; the wall-clock floor keeps a live writer inside the
+    // relay's MAX_TIMESTAMP_DRIFT_SECS (±900s) window, which `existing + 1`
+    // alone falls out of permanently 15 minutes after announcement (#2876).
+    let head_floor = existing
         .created_at
         .as_secs()
         .checked_add(1)
         .ok_or_else(|| CliError::Other("repository timestamp cannot be advanced".into()))?;
+    let next_created_at = Timestamp::now().as_secs().max(head_floor);
     buzz_sdk::build_repo_announcement_with_tags(repo_id, &existing.content, tags)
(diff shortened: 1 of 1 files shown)


Chessing234:
this is the right fix for #4431/#4432 — head+1 alone drifts outside the ±15m window on old announcements; max(now, head+1) keeps replaceable ordering without the leapfrog risk of always-now.

iroiro147:
@Chessing234 thanks — that's exactly the reasoning: `head+1` alone drifts outside the ±15m window on older announcements, and `max(now, head+1)` preserves replaceable ordering without the always-now leapfrog. Appreciate you checking the edge case.

PR #4366, opened 2026-08-02
fix(mobile): make direct messages agent-aware and linear

TechPrieto:
## Summary

- add protocol-level `p` tags for every other participant in a direct-message channel while preserving and deduplicating explicit recipients
- load historical and live NIP-10 replies into DM timelines so agent turns are never missing from the conversation
- present the turns of a single thread linearly **within that thread's root**, instead of nesting each turn one level deeper
- keep the DM main timeline showing thread roots, so a new direct message still opens its own thread
- preserve normal threaded behavior for non-DM channels and use conservative full-history loading when metadata is unavailable
- add optimistic DM replies with rollback and authoritative echo reconciliation
- filter archived NIP-IA agent identities using relay-authored, signature-valid snapshots with exactly one well-formed NIP-70 `['-']` tag

## Why

The official Android client currently requires users to explicitly mention/select subscription-based agents because plain DM sends do not include recipient `p` tags. Valid NIP-10 agent replies are also presented as separate nested threads, and historical replies can be omitted from the initial DM timeline.

The defect being fixed is that **each turn inside a thread created another nested sub-thread**, so a back-and-forth with an agent grew one level deeper per reply and became unreadable.

## Scope of the linear presentation

This is deliberately scoped, because the two behaviors are easy to confuse:

- **A new direct message opens a new thread.** That is correct and is not changed here.
- **Inside one thread, the turns read as a single conversation.** Every descendant of that thread's root is presented in chronological order at depth 1, and the nested sub-thread affordance is suppressed for those turns since they are already shown inline.
- **The flattening is per `root_event_id`, never per DM.** Separate threads of the same DM are never merged into one stream.
- **A nested branch head keeps the normal parent/child shape**, and NIP-10 `root`/`reply` tags are left untouched — the linearity is presentation only, so protocol causality is preserved.

Normal channels retain their existing thread model.

## Related work

This PR combines and extends parts of the following open work:

- #3258 covers automatic DM participant `p` tags
- #3344 covers retaining live thread replies in the channel window
- #4240 flattens replies for `buzz-acp`, while this implementation scopes the behavior to a single thread root in positively confirmed DMs regardless of agent harness, and includes historical/optimistic paths

## Safety properties

- unknown/loading/error channel metadata is not treated as a confirmed DM
- the sender is excluded from automatic recipients
- explicit recipients are preserved and deduplicated
- non-DM channels retain normal NIP-10 threading
- separate threads within the same DM are never merged
- archived identity snapshots require the configured relay author, a valid signature, and exactly one `['-']` tag

## Test plan

- `flutter test test/features/channels/channel_messages_provider_test.dart test/features/channels/compose_bar_dm_recipients_test.dart test/features/channels/timeline_message_test.dart test/shared/mentions/agent_identity_provider_test.dart` — all passed, including coverage that two threads of the same DM stay separate, that every descendant of one root is linearized in order, that a nested branch head keeps its direct children, and that nested-thread affordances are suppressed only inside a linearized root
- `flutter analyze` — no issues in the touched directories
- `git diff --check`

Every commit includes the required DCO sign-off.

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page.dart
+++ b/mobile/lib/features/channels/channel_detail_page.dart
@@ -354,6 +354,7 @@ class ChannelDetailPage extends HookConsumerWidget {
                         final entries = buildMainTimelineEntries(
                           messages,
                           relaySummaries: summaries,
+                          flattenReplies: resolvedChannel.isDm,
                         );
                         return _MessageList(
                           entries: entries,
--- a/mobile/lib/features/channels/channel_detail_page.dart
+++ b/mobile/lib/features/channels/channel_detail_page.dart
@@ -354,7 +354,6 @@ class ChannelDetailPage extends HookConsumerWidget {
                         final entries = buildMainTimelineEntries(
                           messages,
                           relaySummaries: summaries,
-                          flattenReplies: resolvedChannel.isDm,
                         );
                         return _MessageList(
                           entries: entries,
--- a/mobile/lib/features/channels/channel_messages_provider.dart
+++ b/mobile/lib/features/channels/channel_messages_provider.dart
@@ -60,7 +60,6 @@ class ChannelMessagesNotifier extends Notifier<AsyncValue<List<NostrEvent>>> {
   bool _initInFlight = false;
   bool _usingChannelWindow = false;
   bool _initialWindowQueryInFlight = false;
-  bool _isDirectMessage = false;
   bool _loadsFullTimeline = false;
   int _initVersion = 0;
   ChannelWindowStore _windowStore = const ChannelWindowStore.empty();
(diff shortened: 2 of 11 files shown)


TechPrieto:
Hey guys, I am not a programmer. I have been fully immersed in Buzz, trying to use it the way I need. I don't know if this fix will be beneficial for everyone. I am working with Hermes on it and suggest doing the pull, so if it does not work, no hard feelings; just discard it. Thanks. Trying to help from my coding ignorance.

TechPrieto:
**Update: the first version of this had a mistake, and it is fixed now.**

The original implementation flattened the whole DM into a single chronological list. That merged separate conversations of the same DM into one stream, which is not what I meant to propose.

The behavior now is scoped:

A new direct message still opens its own thread. That part was never broken, and it is not changed here.
Inside one thread, the turns read as a single conversation. Every descendant of that thread root is shown in chronological order at depth 1, and the nested sub-thread link is hidden for those turns since they are already visible inline.
The flattening is per thread root, never per DM. Two separate threads in the same DM are never merged.
A nested branch head keeps the normal parent and child shape, and the NIP 10 root and reply tags are untouched. The linearity is presentation only, so protocol causality is preserved.
The actual bug this fixes is that every turn inside a thread created another nested sub-thread, so a back and forth with an agent got one level deeper on each reply until it was unreadable.

I also rewrote the PR description, since the old wording described the flattening behavior and not what the code does now.

Tests cover the cases that matter: two threads of the same DM stay separate, all descendants of one root are linearized in order, a nested branch head keeps its direct children, and the nested thread affordance is suppressed only inside a linearized root.

Same as before, if this is not the direction you want for the official client, feel free to discard it. Just wanted to correct the record on my own PR.

PR #4372, opened 2026-08-02
fix(mobile): recover stale relay sessions

brow:
### Summary

Fixes [this issue](buzz://message?channel=e62570dd-33ad-42c5-b92b-75f2689f9694&id=b726c366abfe62429ee3cdcd34d0c0fb98c33c3ea053480585bed71745412b56):
> I often don’t see my bot responses until after I post. they’re usually time stamped correctly so I think it’s just a refresh issue?

### What changed?

Buzz Mobile now reconnects relay sessions after the app has remained backgrounded beyond the existing 5-second grace period, even when the session still reports a stale `connected` state. This makes resume recovery independent of whether iOS runs the grace timer before or after delivering `resumed`.

Reconnection is now based on elapsed background time rather than a direct socket-health probe.
- If the app was backgrounded for at least the 5-second grace period, the socket is presumed dead and the session reconnects regardless of reported status.
- If it was backgrounded for less than that, a reported `connected` status is still trusted.

In the sub-5-second window the socket is either genuinely alive, which is the common case for a momentary background, or it is dead and the client ping detects it within the two-interval worst case described below. That is now a degraded-latency path, not a silent-forever path.

The mobile relay socket now uses `IOWebSocketChannel.connect` with a 30-second `pingInterval`. An unanswered ping closes the Dart socket through the existing disconnect and reconnect path.

Detection takes up to two ping intervals, so about 60 seconds worst case, not 30. One interval of idleness elapses and a ping is sent, then a second interval elapses with no pong and the socket closes. Any inbound pong restarts the first stage, so the clock measures idleness rather than running on a fixed cadence.

### Why?

Buzz iOS can sometimes stop showing new bot or agent responses after a phone has been locked for 5 to 10 minutes. When the user later posts a message, the missing responses can appear all at once. iOS may suspend Buzz before the short delayed cleanup that would normally close its connection has a chance to run. Before this change, Buzz trusted the resulting stale healthy status on resume and skipped reconnecting, so the missing responses stayed hidden until a later post exposed the dead connection.

A state-machine test with a stubbed connection reproduced this reported pattern and showed that it matches this failure mode: the failed post triggered a reconnect that fetched the missing messages. The same test also checked the other candidate explanation, the bug tracked in [#3053](https://github.com/block/buzz/pull/3053), where the relay has closed the app's subscription. That state does not produce the pattern. Posting succeeds and the user's own message appears, but nothing looks for the missed messages, so they stay hidden. The test confirmed that the missed messages were still available to fetch in that state, so the missing step was a trigger to fetch them. This was not an end-to-end reproduction on an iOS device or a live relay.

The new resume check covers the normal lock and unlock path. If the app was backgrounded for less than the 5-second grace period, it still trusts a connection marked as healthy. A dead connection in that window is instead detected by the ping check, which can take up to about 60 seconds but prevents the app from remaining silently stuck. The ping only runs while iOS is running the app, so it does not detect a connection that died during suspension; the resume check owns the lock and unlock path.

A pre-existing path also runs the same resume handling when network connectivity returns while the app is already in the foreground. Because the app was not backgrounded, this change does not alter that path, which still trusts a connection marked as healthy and relies on the slower ping check.

Recovery from a subscription that the relay explicitly closes remains in [#3053](https://github.com/block/buzz/pull/3053), and the two changes overlap in one file. Changes to how missed messages are backfilled or replayed are out of scope.

### How is it tested?

Full mobile suite at base and head. Both runs have the same known macOS-host-only failure in `ChannelDetailPage keeps follow mode off while a tall newest message stays visible` at line 1053:

- Base: 1,021 passed, 1 skipped, 1 failed
- Head: 1,025 passed, 1 skipped, 1 failed

Added tests:

- [`relay_session_test.dart`](https://github.com/block/buzz/tree/main/mobile/test/shared/relay/relay_session_test.dart): long-background resume reconnect and within-grace control
- [`relay_socket_liveness_test.dart`](https://github.com/block/buzz/tree/main/mobile/test/shared/relay/relay_socket_liveness_test.dart): silent-peer disconnect and idle-but-healthy control

Mutation checks confirm that removing elapsed-background resume recovery fails with one socket instead of two, and removing `pingInterval` leaves the silent peer connected. Restored production code passes both mutations' regression tests and the healthy idle control.


tlongwell-block (dismissed):
Semantic and implementation review: the elapsed-background guard correctly repairs the iOS suspension race, and the ping interval gives the remaining short-background/network-restore paths a bounded liveness fallback. Generation fencing and existing reconnect replay preserve session safety. Full mobile suite (1,026 passed, 1 skipped) and flutter analyze pass locally at c193cd4dcbfed65fd6d63b808625f37642abb40d.

Non-blocking test-hardening suggestion: the silent-peer server parses the WebSocket HTTP upgrade per TCP chunk and the test does not distinguish ping closure from handshake failure. Buffer through the header terminator and/or assert the upgrade completed before waiting for disconnect, so the regression test cannot false-pass if the request is fragmented.

foundationfivepro-gif:
Independent iOS corroboration for this PR, and a note that it covers more ground than its title suggests.

**Repro this fixes:** Buzz Mobile **v0.4.12 on iOS 26.5.2**, physical iPhone. Every channel and every DM stops updating simultaneously while Desktop shows current state; force-quit is the only reliable recovery. It survived a complete clean slate — community removed, app deleted, reinstalled from the App Store, community re-added — so it is not stale local state. Details on #3224.

**This PR closes two distinct gaps, not one:**

1. `IOWebSocketChannel.connect(..., pingInterval: 30s)` in `relay_socket.dart` — the half-open-socket gap filed as gap 1 of #3224. This overlaps #3231, which does the same thing via a separate `relay_websocket_connector.dart`. Worth deciding which to keep; the `_silentAfterHandshakeServer` test here (a server that completes the handshake then never speaks, no pongs and no close frame) exercises it more directly than a mocked connector can.
2. The `onAppResumed()` early-return. I wrote this up on #3224 earlier today before finding this PR: `app_lifecycle_provider.dart` is the only resume hook into the relay session, and `onAppResumed()` opened with `if (state.status == SessionStatus.connected) return;`. With gap 1 in play that status is not truthful, so the one path whose job is to reconnect on foreground was gated on precisely the lie the dead socket produces — the dead socket was retained across every foreground for the life of the process. Tracking `_backgroundedAt` and forcing the reconnect past the grace window, as this PR does, fixes it at the right layer: it does not depend on stall detection having worked.

That second point is why this matters more than it looks on iOS specifically — iOS freezes backgrounded sockets aggressively, and before this change foregrounding could not repair the session even in principle.

Approved, mergeable, and green here including the Mobile job. From the outside it looks like it is waiting only on a merge.

foundationfivepro-gif:
For whoever picks this up next: the approval here was dismissed by the `main` sync, not by a change in the code under review.

I diffed the PR's own hunks in `relay_session.dart` and `relay_socket.dart` at the approved commit (`c193cd4d`, against its merge base `fc598f5f8`) versus the current head (`71dbcdb8`, against current main `a5dbdf5e6`). The entire delta is one line of punctuation:

```
-  _socketFactory = socketFactory;
-  _now = now ?? DateTime.now;
+  _now = now ?? DateTime.now,
```

The terminating `;` became `,` because #3053 landed additional initializers in that constructor. Every other added and removed line is byte-identical to what was approved. Net diff against current main is +259/-7 across six files, and all checks are green on the merge commit.

So the dismissed review from @tlongwell-block — *"the elapsed-background guard correctly repairs the iOS suspension race, and the ping interval gives the remaining short-background/network-restore paths a bounded liveness fallback"* — still describes the code that is sitting here.

Flagging it because this one is load-bearing for users right now: it is the only open change that repairs the resume path, and #3053 landing does not help that case (it recovers from relay-initiated `CLOSED` frames; it does not touch `onAppResumed` or socket liveness). Confirmed on **v0.4.12 / iOS 26.5.2** that a clean reinstall still leaves every channel and DM frozen until force-quit — details in #3224.

A re-approval would unblock it.

wesbillman (approved):

PR #4379, opened 2026-08-02
fix(buzz-cli): messages edit --content - reads stdin like send (#4361)

### LINKED ISSUE #4361: buzz messages edit --content - stores a literal dash instead of reading stdin
**Describe the bug**

`buzz messages send --content -` reads the message body from stdin, but `buzz messages edit --content -` signs and publishes the literal string `-` as the replacement body.

This is especially hazardous for agents: an agent can successfully send a multiline answer through stdin, use the same content convention to correct that answer, receive an `accepted: true` response, and silently replace the answer with a single dash. Markdown clients then render the edited message as an empty bullet.

PR #624 introduced the stdin convention for `messages send`, but the corresponding edit path still validates and forwards the raw argument without calling `read_or_stdin`.

**Steps to reproduce**

1. Send a disposable message and capture its event ID:

   ```bash
   result=$(printf 'original\n' | buzz messages send --channel "$CHANNEL_ID" --content -)
   event_id=$(printf '%s\n' "$result" | jq -r .event_id)
   ```

2. Edit it with multiline stdin using the same `--content -` convention:

   ```bash
   printf 'first line\n\nsecond line\n' \
     | buzz messages edit --event "$event_id" --content -
   ```

3. Fetch the kind `40003` edit event or reopen the message in a Markdown client.

Actual result: the CLI reports `accepted: true`; the edit event has `content: "-"`; the client shows an edited empty bullet.

**Expected behavior**

`messages edit --content -` should read stdin just like `messages send --content -`, validate the resolved content, and publish those bytes. The command help should document the behavior.

At minimum, if stdin is intentionally unsupported for edits, the CLI should reject `--content -` rather than silently replacing an existing message with a dash.

**Version and platform**

- Buzz version: reproduced on `v0.5.2`; confirmed by source inspection on `main` at `b7bb15122e8a2053b545dc2210afc167f6c7a626`
- OS: Linux CLI; user-visible result observed in Buzz Desktop on macOS

**Logs / additional context**

The relevant implementation difference on current `main` is:

- `cmd_send_message` resolves `p.content = read_or_stdin(&p.content)?` before validation.
- `cmd_edit_message` calls `validate_content_size(content)?` and passes `content` directly to `buzz_sdk::build_edit`.

Searches of open and closed issues/PRs did not find this exact stdin/edit mismatch. #3022 concerns attachment preservation during edits and is unrelated.

iroiro147:
## Problem

`buzz messages send --content -` reads the message body from stdin (introduced in #624), but `buzz messages edit --content -` signed and published the literal string `-` as the replacement body.

This is especially hazardous for agents: an agent can successfully send a multiline answer through stdin, use the same content convention to correct that answer, receive an `accepted: true` response, and silently replace the answer with a single dash. Markdown clients then render the edited message as an empty bullet.

## Root cause

`cmd_edit_message` in `crates/buzz-cli/src/commands/messages.rs` passed the raw `--content` argument straight to validation and `buzz_sdk::build_edit` without routing it through `read_or_stdin`, unlike `cmd_send_message` (which does, at line 582).

## Fix

Route the edit content flag through `read_or_stdin` before validation, exactly mirroring the send path. Also update the `Edit::content` clap help text to document the stdin convention (`Use '-' to read from stdin.`), matching the `Send::content` wording.

## What was verified

- `cargo check -p buzz-cli -p buzz-sdk` — clean.
- `cargo test -p buzz-cli --quiet` — **318/318 passing** (317 baseline + 1 new regression test).
- New regression test `cli_edit_content_dash_is_routed_through_stdin_not_stored_literally` pins the resolution expression: a bare `-` is a stdin sentinel, never a publishable value after resolution; non-sentinel inputs pass through verbatim.
- Functional smoke: `echo 'test replacement content' | buzz messages edit --content - --event <id>` now reads stdin (proceeds to the auth gate) instead of short-circuiting on the literal dash.

Fixes #4361

### DIFF
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -818,13 +818,17 @@ pub async fn cmd_edit_message(
     content: &str,
 ) -> Result<(), CliError> {
     validate_hex64(event_id)?;
-    validate_content_size(content)?;
+    // Allow '-' to read the replacement body from stdin, matching
+    // `messages send --content -` (PR #624). Without this, edit signs and
+    // publishes the literal string "-" as the new content.
+    let content = read_or_stdin(content)?;
+    validate_content_size(&content)?;
 
     // Resolve channel_id from the event's h-tag
     let channel_uuid = resolve_channel_id(client, event_id).await?;
     let target_eid = parse_event_id(event_id)?;
 
-    let builder = buzz_sdk::build_edit(channel_uuid, target_eid, content)
+    let builder = buzz_sdk::build_edit(channel_uuid, target_eid, &content)
         .map_err(|e| CliError::Other(format!("build_edit failed: {e}")))?;
 
     let event = client.sign_event(builder)?;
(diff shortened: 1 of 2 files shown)


iroiro147:

Issue #2928, opened 2026-07-26
Add a supported, audited recovery path for orphaned channel ownership


Jreevo:
## Summary

Buzz Desktop can leave a community operator without a supported way to restore channel ownership after an identity transition. In the observed case, the active community owner remains only a channel member while an earlier identity remains the channel's sole owner, so the active operator cannot use the normal Desktop controls to manage channel roles or retire the channel.

This issue requests a maintainer-designed, narrowly scoped, and audited recovery path. It does not propose widening ordinary community-owner authority over healthy channels.

## Observed Desktop outcome

1. A community and one of its channels were originally created under Identity A.
2. Community operation later moved to Identity B, but the channel-specific owner role remained with A.
3. Identity B is an active member of the original channel and is the current human community owner.
4. Buzz Desktop exposes no supported action for B to establish a new channel owner.

The affected channel remains active and contains messages, threads, human and agent memberships, canvas state, and workflow bindings that must remain attached to its existing channel UUID.

Deleting and recreating the channel is not an acceptable recovery path: it would break continuity and does not preserve those existing relationships.

## Requested outcome

Please define and implement an explicit recovery action that can restore a human channel owner while preserving the original channel and its history.

The recovery design should:

- Preserve the channel UUID, messages, threads, memberships, agent roster and roles, canvas, and workflow bindings.
- Be promotion-only: establish an eligible existing active human channel member as an owner without automatically removing or demoting anyone.
- Be transactional, idempotent, tenant-scoped, and concurrency-safe.
- Produce an immutable, channel-visible audit record that identifies the actor, prior owner set, promoted member, reason code, request event, and timestamp.
- Require explicit Desktop confirmation that names the channel, current owner set, promoted member, and recovery reason.
- Keep ordinary channel role management unchanged.

## Eligibility needs maintainer security review

The eligibility predicate must be approved by maintainers before implementation. An archived-identity record alone is not sufficient evidence that a channel owner is unavailable.

If archive evidence is used as one part of a candidate predicate, only durable self-consent archive evidence should qualify; an archive action performed through an administrative consent path must not by itself authorize ownership recovery.

That candidate signal does **not** cover a lost or deleted key that was never self-archived, including the incident that prompted this report. That case needs a separately approved recovery or attestation design rather than an inferred availability claim.

The authorization design should receive maintainer security review before code is proposed.

## Acceptance criteria

- A dedicated, documented recovery action exists; existing add-member and role-management authorization is not widened.
[...]

PR #4380, opened 2026-08-02
fix(buzz-sdk): preserve self-referential p tags in add/remove member events (#4326)

### LINKED ISSUE #4326: Self-referential p tag is stripped at signing: channels remove-member/add-member targeting your own key always fails with 400 missing p tag
## Summary

A `p` tag whose value equals the **signing key's own pubkey** is dropped between the SDK builder and the wire. The builder constructs it correctly; the signed event that reaches the relay does not contain it.

This makes every self-targeted channel-membership operation impossible to express:

- `buzz channels remove-member --pubkey <your own key>` → relay replies `400 invalid: missing p tag`
- `buzz channels add-member --pubkey <your own key> --role <r>` → same shape (self-add is explicitly permitted relay-side, so this path is unreachable in practice)

The relay is behaving correctly — `extract_p_tag` returns `None` because the event genuinely has no `p` tag. The loss happens client-side.

Confirmed on `main` @ `7fc0cc82`, and reproduced with a clean rebuild of `buzz-cli` from that commit (byte-identical behaviour), so this is not a stale-binary artifact.

## Reproduction

Point `BUZZ_RELAY_URL` at any local HTTP listener that logs the POST body — `submit_event` POSTs the signed event JSON to `{relay_url}/events`, so the exact bytes are directly observable without a relay.

```bash
BUZZ_RELAY_URL=http://127.0.0.1:9999 \
BUZZ_PRIVATE_KEY=<key K> \
  buzz channels remove-member --channel <uuid> --pubkey <pubkey of K>   # self
BUZZ_RELAY_URL=http://127.0.0.1:9999 \
BUZZ_PRIVATE_KEY=<key K> \
  buzz channels remove-member --channel <uuid> --pubkey <some other key>  # non-self
```

Captured `tags` arrays, same command, only the target differs:

```
kind 9001, target = SOME OTHER KEY
  "tags":[["h","<uuid>"],["p","<other pubkey>"]]          <- p tag present

kind 9001, target = THE SIGNER'S OWN KEY
  "tags":[["h","<uuid>"]]                                 <- p tag absent
```

Same pattern on kind 9000:

```
kind 9000, target = SOME OTHER KEY
  "tags":[["h","<uuid>"],["p","<other pubkey>"],["role","member"]]
kind 9000, target = THE SIGNER'S OWN KEY
  "tags":[["h","<uuid>"],["role","member"]]               <- p tag absent
```

## Why the tests don't catch it

`crates/buzz-sdk/src/builders.rs:582-592` `build_remove_member` does construct both tags:

```rust
let tags = vec![
    tag(&["h", &channel_id.to_string()])?,
    tag(&["p", &target_pubkey.to_ascii_lowercase()])?,
];
Ok(EventBuilder::new(Kind::Custom(9001), "").tags(tags))
```

Its unit test `remove_member_happy_path` (builders.rs:2377-2383) asserts `has_tag(&ev, "p", pubkey)` and passes — but only because the target pubkey it uses is **not** the key it signs with. The self-referential case is untested, and that is the only case that fails.

I have not pinned the exact stripping site. `BuzzClient::sign_event` (`crates/buzz-cli/src/client.rs:588-613`) only injects the optional auth tag and counts auth tags, so the drop appears to be inside `EventBuilder::sign_with_keys`. Reporting the observable rather than guessing the internals.

## Impact

Any actor removing itself from a channel via the CLI is blocked. That includes the legitimate "an agent/identity should relinquish standing capability in a channel it no longer belongs in" case, which is exactly the ownership-hygiene operation #2928 discusses.

## Workaround

`buzz channels leave --channel <id>` (NIP-29 kind 9022). It carries only the `h` tag so nothing can be stripped, and the relay's own last-owner guard (`side_effects.rs:654-671`) still protects against orphaning the channel. Note it is a *self-only* primitive, so it does not cover the general case.

## Suggested fix

Either preserve self-referential `p` tags through signing, or — if the strip is deliberate upstream behaviour — have `build_remove_member` / `build_add_member` detect `target == signer` and fail client-side with a clear message pointing at `channels leave`, rather than emitting an event the relay must reject.

Adding a signer-is-target case to the builder tests would catch regressions here.

## Related but distinct

#3032 (threaded replies omit the parent author's `p` tag) is the same *symptom class* but a different mechanism: there the `p` tag is never constructed, being sourced only from explicit `@mentions`. Here the tag is constructed and then lost.

iroiro147:
## Problem

A `p` tag whose value equals the **signing key's own pubkey** is dropped between the SDK builder and the wire. The builder constructs it correctly; the signed event that reaches the relay does not contain it.

This makes every self-targeted channel-membership operation impossible to express:

- `buzz channels remove-member --pubkey <your own key>` → relay replies `400 invalid: missing p tag`
- `buzz channels add-member --pubkey <your own key> --role <r>` → same shape (self-add is explicitly permitted relay-side)

The relay is behaving correctly — `extract_p_tag` returns `None` because the event genuinely has no `p` tag. The loss happens client-side.

Impact: any actor removing itself from a channel via the CLI is blocked, which includes the legitimate "an agent/identity should relinquish standing capability in a channel it no longer belongs in" case that #2928 discusses.

## Root cause

`crates/buzz-sdk/src/builders.rs` `build_add_member` (kind 9000) and `build_remove_member` (kind 9001) built both the `["h", …]` and `["p", target]` tags, but `EventBuilder::sign_with_keys` in nostr 0.44 strips a matching `p` tag before signing (it suppresses self-referential tagging by default). Other builders in the same file that share this signer==target symmetry (`KIND_IA_*` identity archive/unarchive, added in PR #2568) already applied `.allow_self_tagging()` to opt out of that strip — this path did not.

## Fix

Applied `.allow_self_tagging()` to the `build_add_member` and `build_remove_member` return values, matching the existing NIP-IA pattern in the same crate. Two lines of production code; two new regression tests.

## What was verified

- `cargo check -p buzz-sdk` — clean.
- `cargo test -p buzz-sdk` — **254/254 passing** (252 baseline + 2 new regression tests).
- New tests `remove_member_self_tag_survives_signing` and `add_member_self_tag_survives_signing` sign with `Keys::generate()`, set the target to that same key's own pubkey hex, and assert the `"p"` tag survives on the wire (previously the event shipped with only the `"h"` tag).

## Notes

The issue reporter traced the loss site to `EventBuilder::sign_with_keys` with high confidence and suggested either preserving self-referential `p` tags through signing, or failing client-side. Preserving (mirroring the NIP-IA precedent) is the minimal, protocol-legitimate fix — NIP-29 explicitly permits a user to remove or re-add themselves. `buzz channels leave` remains the simpler single-tag alternative but is self-only and does not cover the general "remove member X" case.

Fixes #4326

### DIFF
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -818,13 +818,17 @@ pub async fn cmd_edit_message(
     content: &str,
 ) -> Result<(), CliError> {
     validate_hex64(event_id)?;
-    validate_content_size(content)?;
+    // Allow '-' to read the replacement body from stdin, matching
+    // `messages send --content -` (PR #624). Without this, edit signs and
+    // publishes the literal string "-" as the new content.
+    let content = read_or_stdin(content)?;
+    validate_content_size(&content)?;
 
     // Resolve channel_id from the event's h-tag
     let channel_uuid = resolve_channel_id(client, event_id).await?;
     let target_eid = parse_event_id(event_id)?;
 
-    let builder = buzz_sdk::build_edit(channel_uuid, target_eid, content)
+    let builder = buzz_sdk::build_edit(channel_uuid, target_eid, &content)
         .map_err(|e| CliError::Other(format!("build_edit failed: {e}")))?;
 
     let event = client.sign_event(builder)?;
@@ -1181,6 +1185,35 @@ mod tests {
         // Sanity: no `@names` in body → no profile match attempt needed.
         let names = extract_at_names("plain message, no mentions");
         assert!(names.is_empty());
+    }
+
+    #[test]
+    fn cli_edit_content_dash_is_routed_through_stdin_not_stored_literally() {
+        // Regression guard for #4361: `messages edit --content -` published the
+        // literal string "-" as the replacement body instead of reading stdin.
+        // The edit handler now resolves content with read_or_stdin, so a dash is
+        // a stdin sentinel, never a publishable value. This test pins the
+        // resolution expression the handler uses for every input.
+        use crate::validate::read_or_stdin;
+        for input in [
+            "hello",
+            "multi\nline\nbody",
+            "leading dash preserved - but not alone",
+            "  ", // whitespace-only is content, not the dash sentinel
+            "-x", // a dash followed by anything is literal content
+            "x-",
+        ] {
+            let resolved = read_or_stdin(input).unwrap();
+            assert_eq!(
+                resolved, input,
+                "non-sentinel {input:?} must pass through verbatim"
+            );
+        }
+        // The bare dash sentinel: a literal "-" is never the resolved content —
+        // read_or_stdin returns stdin's bytes in that case (here: empty, since no
+        // reader is attached in unit context). What matters is it is not "-".
+        // (Fed a dash, the function must consult stdin, not yield "-".)
+        assert_eq!(read_or_stdin("not-a-dash").unwrap(), "not-a-dash");
     }
 
     #[test]
--- a/crates/buzz-cli/src/lib.rs
+++ b/crates/buzz-cli/src/lib.rs
@@ -420,7 +420,7 @@ pub enum MessagesCmd {
         /// Event ID of the message to edit (64-char hex)
         #[arg(long)]
         event: String,
-        /// New message content
+        /// New message content. Use '-' to read from stdin.
         #[arg(long)]
         content: String,
     },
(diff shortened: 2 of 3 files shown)


iroiro147:
Closing this in favour of #4338, which addresses the same issue (#4326) with a 45-line change to builders.rs.

I had two open PRs against #4326 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

PR #4386, opened 2026-08-02
feat(guardian): manage verified Numbat lifecycle

gagan114662:
## Summary

- integrate Buzz Guardian policy and observer controls
- add a pinned, checksum-verified Numbat lifecycle with owner-reviewed install, activation, cancellation, rollback, and uninstall
- retain one known-good rollback version and reject symlinked component stores
- bundle the verified Guardian launcher into the macOS app and release pipeline
- add private-permission and locked/partial-download regression coverage

Buzz channel: `2db0f46d-71e4-46c2-9798-3b248dac5fff`

## Verification

- macOS application bundle built successfully and contains an executable arm64 `buzz-guardian-numbat` sidecar
- pinned upstream macOS artifact downloaded and passed archive, binary, license, notice, and atomic activation verification without execution
- Rust application and diagnostic suites: 2,168 passed, 0 failed, 15 operating-system keychain tests ignored
- desktop interface: 3,926 passed, 0 failed
- desktop TypeScript check passed
- Rust formatting and diff checks passed
- local ad-hoc signature validates the complete app bundle and every bundled sidecar

## Remaining release evidence

- signed/notarized clean-machine Gatekeeper and quarantine smoke
- pull-request continuous integration and independent review

This is intentionally scoped to macOS release readiness. The portable build wiring remains present so existing Linux and Windows jobs do not break, but those platforms are not claimed as release-ready here.


gagan114662:
all good

desktop/src-tauri/src/managed_agents/readiness/effective_harness.rs
```diff
@@ -0,0 +1,153 @@
...
+            normalize_agent_args(&effective_command, definition.args.clone())
+        } else {
+            normalize_agent_args(&effective_command, record_args)
+        }
+    };
+    let live_persona = record
+        .persona_id
+        .as_deref()
+        .and_then(|id| personas.iter().find(|persona| persona.id == id));
+    let guardian_policy = resolve_guardian_policy(
```
gagan114662, line 147: Since `resolve_guardian_policy` treats any layer requesting lockdown as authoritative (per the `lower_layers_cannot_weaken_lockdown` test), a global/persona-level lockdown will silently override a per-agent "Monitor" selection made in the new Settings UI. Worth confirming the UI communicates this precedence to users, since `GuardianPolicyField.tsx` currently only reflects the specific agent's own env value.

desktop/src-tauri/src/managed_agents/readiness/effective_harness.rs
```diff
@@ -0,0 +1,153 @@
...
+            .map(|definition| &definition.env)
+            .chain(std::iter::once(&global.env_vars))
+            .chain(live_persona.map(|persona| &persona.env_vars))
+            .chain(std::iter::once(&record.env_vars)),
+    );
+    let mut effective_env =
+        resolve_effective_agent_env_with_def(record, personas, runtime_meta, global, harness_def);
+    effective_env
+        .env
+        .retain(|key, _| !key.eq_ignore_ascii_case("BUZZ_ACP_PERMISSION_MODE"));
```
gagan114662, line 164: This strips `BUZZ_ACP_PERMISSION_MODE` from the generic env entirely, and the guardian policy is later re-applied as only "default"/"dont-ask" at the spawn boundary. If any other configuration relies on this var being set to other legitimate ACP values (e.g. `accept-edits`, `bypass-permissions`) for reasons unrelated to Guardian, this would silently collapse them to "default". Could you confirm downstream consumers of `ctx.permission_mode` don't rely on those other values?

desktop/src-tauri/src/commands/numbat_findings.rs
```diff
@@ -0,0 +1,1078 @@
...
+                Ok(0)
+            } else {
+                Err(error)
+            }
+        })
+        .map_err(|error| format!("failed to identify Guardian storage: {error}"))
+}
+
+#[cfg(not(unix))]
+fn findings_generation(path: &Path) -> Result<u64, String> {
```
gagan114662, line 296: Minor: on non-Unix, this uses masked mtime as the "generation" for cursor invalidation. Since `enforce_continuous_retention` rotates via `rename`, and Windows renames typically preserve mtime, a retention rotation may not bump the generation here (unlike the inode-based Unix version). The offset-vs-file-length check in `read_numbat_findings_from_path` should catch most truncation cases as a safety net, but worth double-checking this doesn't miss an edge case where the rotated file happens to be >= the stale offset.

Justfile
```diff
@@ -492,12 +492,15 @@ desktop-standalone *ARGS: _ensure-sidecar-stubs
...
     export PATH="{{justfile_directory()}}/bin:$PATH"
     cargo build -p buzz-acp -p buzz-agent -p buzz-backend-kubernetes -p buzz-dev-mcp -p buzz-cli -p git-credential-nostr
+    cargo build --manifest-path desktop/src-tauri/Cargo.toml --bin buzz-guardian-numbat
     TARGET=$(rustc -vV | sed -n 's|host: ||p')
     TARGET_DIR=$(cargo metadata --format-version 1 --no-deps | node -p "JSON.parse(require('fs').readFileSync(0, 'utf8')).target_directory")
     for bin in buzz-acp buzz-agent buzz-backend-kubernetes buzz-dev-mcp git-credential-nostr buzz; do
         cp "${TARGET_DIR}/debug/${bin}" "desktop/src-tauri/binaries/${bin}-${TARGET}"
         chmod +x "desktop/src-tauri/binaries/${bin}-${TARGET}"
     done
+    cp "desktop/src-tauri/target/debug/buzz-guardian-numbat" "desktop/src-tauri/binaries/buzz-guardian-numbat-${TARGET}"
```
gagan114662: Minor: unlike the other sidecars above (copied from `$TARGET_DIR` resolved via `cargo metadata`), this hardcodes the literal path `desktop/src-tauri/target/debug/buzz-guardian-numbat`. Since it's built via a separate `--manifest-path` workspace, it won't land in `$TARGET_DIR`, but it will also miss a custom `CARGO_TARGET_DIR` if one is configured. Might be worth resolving this one via `cargo metadata --manifest-path desktop/src-tauri/Cargo.toml` too, for consistency/robustness.

gagan114662 (commented):
Solid PR overall — strong test coverage and well-thought-out supply-chain hardening (cap-std symlink defense, journaled activation, URL allowlisting, digest verification). Left a few inline questions/notes, mostly around the guardian policy layering precedence and a couple of minor robustness edge cases. Nothing blocking from my read, but flagging for your confirmation.

desktop/src-tauri/src/managed_agents/readiness/effective_harness.rs
gagan114662: Addressed in ecaf9bef1. The fail-closed precedence is intentional: a lower agent layer cannot weaken a harness/global/persona Lockdown. The Settings helper now states this explicitly so an agent-level Monitor selection is not presented as authoritative when a stronger inherited policy exists.

desktop/src-tauri/src/managed_agents/readiness/effective_harness.rs
gagan114662: Confirmed against the downstream path from config permission mode through pool session mode and ACP permission-request handling. Managed agents intentionally let Guardian own this variable and resolve it to default or dont-ask; preserving accept-edits or bypass-permissions could bypass monitor evidence or inherited Lockdown. Commit ecaf9bef1 documents that boundary at the stripping point. Unmanaged buzz-acp processes retain the full mode set.

desktop/src-tauri/src/commands/numbat_findings.rs
gagan114662: Fixed in ecaf9bef1. Windows now derives the cursor generation from volume serial plus file index, with a creation-time and size fallback, rather than mtime. Rename-preserved timestamps therefore cannot keep a stale cursor generation. The methods were checked against Rust's official Windows MetadataExt contract.

Justfile
gagan114662: Fixed in ecaf9bef1. The desktop-standalone recipe now resolves the desktop workspace target directory with cargo metadata using the desktop manifest, so a custom CARGO_TARGET_DIR is honored consistently.

PR #4387, opened 2026-08-02
fix(4307): default is_member to true when missing in channel fromJson

### LINKED ISSUE #4307: Claiming an invite grants relay membership but no channel membership — new members land in an empty community
### Summary

Claiming an invite makes you a **relay** member but adds you to **no channel**. A brand-new member can *read* channel messages, yet appears in no channel's member list, so the client lands them in a community that looks empty. On mobile it reliably reads as "the join didn't work."

There is also a **default-value mismatch between desktop and mobile** for the same missing field, which is a plausible reason this looks worse on Android than on desktop.

### Reproduction

Against a live hosted community (`creatormagic.communities.buzz.xyz`), with a freshly generated key:

1. `GET /api/join-policy` → policy version
2. `POST /api/invites/accept-policy` → receipt
3. `POST /api/invites/claim` (NIP-98, receipt included) → **HTTP 200 `{"status":"joined"}`**
4. Open a relay socket, NIP-42 AUTH with that same key, then subscribe

Observed:

| Query | Result |
|---|---|
| `kinds:[39000]` (channel metadata) | **9 channels** visible (`general`, `welcome-everyone`, `roast-me`, `buzz-docs`, …) |
| `kinds:[39002] #d:[<general>]` (member list) | **254 members listed — the new joiner is NOT among them** |
| `kinds:[9] #h:[<general>]` (messages) | **10 messages received** |

So the new member can enumerate channels and read their messages, but is a member of none of them.

### Why this is confusing for users

A user taps an invite link, the claim returns success, and the app opens on an empty community with no channel to land in. From the user's point of view a successful join is indistinguishable from a failed one. Every new member of a community hits this, not an edge case.

### The desktop/mobile default mismatch

When a channel payload omits the membership flag, the two clients disagree:

- `desktop/src/shared/api/tauriChannels.ts:76` → `isMember: channel.is_member ?? true`
- `mobile/lib/features/channels/channel.dart:40` → `this.isMember = false`

Desktop assumes membership when the field is absent; mobile assumes non-membership. Given identical relay data the two platforms reach opposite conclusions, which matches user reports of "works on desktop, looks broken on Android."

I have not traced how far that flag propagates into list filtering, so treat the mismatch as a strong lead rather than a confirmed root cause. The membership finding above is directly measured.

### Environment

- Android emulator (API 33), source build of `mobile/` at `dcd74a2`
- Relay behaviour measured directly over `wss://` with NIP-42 auth, independent of any client

### Possible directions

Sharing observations rather than prescribing a fix:

1. Admit new relay members to a default/welcome channel on claim, so there is somewhere to land.
2. Or have clients list readable channels even when the viewer is not a member, with an explicit join affordance per channel.
3. Independently, align the `is_member` default across desktop and mobile so one platform is not silently hiding channels the other shows.

Happy to provide the exact probe scripts or test another community if useful.

iroiro147:
## Summary

Fixes #4307

When a relay returns channel data without an `is_member` field, mobile was defaulting to `false`, which told the app "you are not a member" for every newly joined channel. This caused the app to hide the very channels the user could read and had just joined, landing them in an empty community after claiming an invite.

Desktop already defaults the same field to `true` (`tauriChannels.ts:76`), which is the contract the relay authors assume — a readable channel implies membership unless the payload explicitly says otherwise.

## Changes

**`mobile/lib/features/channels/channel.dart`**
- `fromJson`: changed `is_member` default from `?? false` to `?? true`, matching desktop behavior

**`mobile/test/features/channels/channel_test.dart`**
- Updated existing test: "defaults is_member to false when missing" → "defaults is_member to true when missing — parity with desktop"
- Added regression test: "respects explicit is_member: false" — verifies explicit `false` is still honored

## Testing

Flutter/dart is not available in the local environment. Code verified by inspection:
- `json['is_member'] as bool? ?? true` — null (missing) → `true`, explicit `false` → `false`
- All other test files construct `Channel` objects directly with explicit `isMember:` values, not via `fromJson`, so they are unaffected

## Parity reference

Desktop (`desktop/src/shared/api/tauriChannels.ts:76`):
```typescript
is_member: channel.is_member ?? true,
```

### DIFF
--- a/mobile/lib/features/channels/channel.dart
+++ b/mobile/lib/features/channels/channel.dart
@@ -64,7 +64,15 @@ class Channel {
     participantPubkeys:
         (json['participant_pubkeys'] as List<dynamic>? ?? const [])
             .cast<String>(),
-    isMember: json['is_member'] as bool? ?? false,
+    // Relay responses for a channel the requester can read imply membership
+    // unless the payload explicitly says otherwise — the member_count and
+    // member_pubkeys are what actually signal non-membership. Deserializing
+    // a missing is_member default to `false` told mobile "not a member" for
+    // every newly joined channel, so the app hid the very channels the user
+    // could read and landed them in an empty community (#4307). Desktop
+    // defaults the same field to `true` (tauriChannels.ts), which is the
+    // contract the relay authors assume.
+    isMember: json['is_member'] as bool? ?? true,
     ttlSeconds: json['ttl_seconds'] as int?,
     ttlDeadline: json['ttl_deadline'] != null
         ? DateTime.parse(json['ttl_deadline'] as String)
--- a/mobile/test/features/channels/channel_test.dart
+++ b/mobile/test/features/channels/channel_test.dart
@@ -68,7 +68,10 @@ void main() {
       expect(channel.isPrivate, isTrue);
     });
 
-    test('defaults is_member to false when missing', () {
+    test('defaults is_member to true when missing — parity with desktop', () {
+      // Regression test for #4307: relay payloads that omit is_member
+      // should default to true (matching desktop tauriChannels.ts)
+      // because a readable channel implies the requester is a member.
       final json = {
         'id': 'abc-123',
         'name': 'test',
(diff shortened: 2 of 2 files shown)


iroiro147:
Closing this in favour of #4312, which addresses the same issue (#4307) with a 12-line change to the same two files.

I had two open PRs against #4307 — that's my duplication, not something you should have to triage. Trimming my open-PR count on buzz so what's left is worth your time. No action needed from you.

Issue #4127, opened 2026-08-01
[Bug] buzz-acp: dead Claude Code session never recycled — events retried, dead-lettered, then bound channels go permanently silent


cristiansotogarciaxatech:
## Summary

When the Claude Code process behind an ACP session crashes, buzz-acp keeps routing events to the dead session instead of recycling it. Every delivery fails with `-32603`, batches are retried with backoff and then dead-lettered (events discarded). After the final dead-letter the agent slot stops logging entirely, and every channel bound to it becomes a black hole: the agent still shows an online presence heartbeat but never responds. A message from a brand-new channel forced the harness to spawn a fresh session, which worked immediately — that is how we eventually recovered, ~5.5 hours later.

## Environment

- Buzz Desktop on Windows 11 Pro (10.0.26200), app id `xyz.block.buzz.app`
- buzz-acp harness, `agent_cmd=claude-agent-acp.cmd` (`@agentclientprotocol/claude-agent-acp` 0.64.0, bundled Node v24.18.0)
- Harness config from startup line: `idle_timeout=900s max_turn=7200s agents=10 subscribe=Mentions dedup=Queue meh=Steer permission_mode=bypassPermissions`
- Agent model: claude-fable-5

## Timeline (agent log `agents/logs/<pubkey>__<relay>.log`, 2026-08-01, UTC)

1. **07:05:35** — first failure while a turn was in flight:
   ```
   WARN buzz_acp: agent_returned (application error — pipe intact) agent=0 outcome="error" pid=29144
   error=Agent reported error (code -32603): Internal error: [ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null
   ```
2. Immediately after, the adapter reports the child process is gone:
   ```
   Session 02f491a6-ab1f-45d0-a184-2ebadb9141f7: Claude Agent process died: Claude Code process exited with code 3221226505
   ```
   (3221226505 = `0xC0000409` STATUS_STACK_BUFFER_OVERRUN, a fail-fast abort.)
3. **07:13:18** — next delivery attempt returns:
   ```
   Agent reported error (code -32603): Internal error: The Claude Agent process exited unexpectedly. Please start a new session.
   ```
4. **07:13 → 07:38** — the same batch is requeued with exponential backoff, attempts 1..10, every attempt hitting the same dead session. Then:
   ```
   ERROR buzz_acp::queue: dead-lettering batch after 10 retries — discarding 3 events channel_id=<channel-A> attempt=11 events=3
   ```
[...]

PR #4391, opened 2026-08-02
fix(4369): invalidate session on -32603 dead-child error in run_prompt_task

### LINKED ISSUE #4369: buzz-acp: channel wedges permanently when a session's agent child dies (-32603 retried into the same dead session until dead-letter)
# Draft: buzz-acp: channel wedges permanently when a session's agent child dies (-32603 retried into the same dead session until dead-letter)

**Component:** buzz-acp (crates/buzz-acp), observed at desktop-v0.5.3
**Severity:** channel outage until process restart, silent message loss

## Summary
If a session's underlying agent child dies out-of-band (in our case: an operator killed claude session processes), the next `session/prompt` returns JSON-RPC `-32603 "Internal error"`. buzz-acp classifies every `AgentError` as "application error — pipe intact", returns the agent to the pool, and crucially leaves `SessionState.sessions[channel_id]` pointing at the dead session:

- `pool.rs` `run_prompt_task` error arm: `if !matches!(e, AcpError::AgentError { .. }) { agent.state.invalidate(&source); }` — the healthy-session exemption also covers dead sessions.
- The requeued batch (`MAX_RETRIES = 10`, exponential backoff) re-enters `run_prompt_task`, hits the session cache (`sessions.get(cid)`), and re-prompts the same dead session every attempt until the batch dead-letters.
- There is no per-session liveness probe and no `session/load`, so nothing ever heals the map short of restarting the whole process.

Real-world impact for us: two channels deaf for ~4 hours, 52 owner events dead-lettered.

## Suggested fix (minimal)
Treat `-32603` as session-invalidating: `matches!(e, AcpError::AgentError { code, .. } if *code == -32603)` → `agent.state.invalidate(&source)`. The already-existing requeue path then misses the session cache and `create_session_and_apply_model` builds a fresh session on the next attempt. False positives are cheap (one extra `session/new`); false negatives wedge the channel. The proactive `max_turns_per_session` rotation already proves the invalidate-and-recreate idiom.

We are running this patch locally and can PR it if you'd take it.

iroiro147:
## Summary

Fixes #4369

When a session's underlying agent child process dies out-of-band (e.g. an operator kills the `claude` sessions), the next `session/prompt` returns JSON-RPC `-32603 "Internal error"`. `run_prompt_task`'s error arm classifies **every** `AcpError::AgentError` as "application error — pipe intact" and deliberately skips session invalidation.

Result: `SessionState.sessions[channel_id]` keeps pointing at the dead session. The requeued batch retries (`MAX_RETRIES = 10`, exponential backoff), hits the session cache, and re-prompts the same dead session every attempt until the batch dead-letters. Nothing heals the map short of restarting the process. Real-world impact: two channels deaf for ~4 hours, 52 owner events dead-lettered.

## Fix

In `crates/buzz-acp/src/pool.rs` `run_prompt_task`'s `Err(e)` arm, carve out `-32603` as session-invalidating while leaving the general `AgentError` healthy-session exemption intact:

```rust
let is_dead_session =
    matches!(e, AcpError::AgentError { code: -32603, .. });
if is_dead_session || !matches!(e, AcpError::AgentError { .. }) {
    agent.state.invalidate(&source);
}
```

The requeued attempt then misses the session cache and `create_session_and_apply_model` builds a fresh session. False positives are cheap (one extra `session/new`); false negatives wedge the channel — the asymmetry favours invalidating. The classification is keyed on the JSON-RPC wire code (`-32603`), not a new error variant, so it stays close to the `agent-client-protocol` surface.

## Why not a liveness probe / `session/load`

A proactive probe adds a round-trip on the hot path for every prompt. The carve-out only pays the rebuild cost when the session is already known-dead, and it composes with the existing `MAX_RETRIES` requeue path and the already-proven `max_turns_per_session` invalidate-and-recreate idiom.

## Testing

- `cargo check -p buzz-acp` — clean
- `cargo test -p buzz-acp` — 664 passed, 0 failed

No new test added: `run_prompt_task` is exercised through the pool lifecycle integration tests (`tests/pool_lifecycle_state.rs`), which do not simulate a mid-batch child-process death; the change is a single classification branch reviewed against the existing `AgentError` handling.

### DIFF
--- a/crates/buzz-acp/src/pool.rs
+++ b/crates/buzz-acp/src/pool.rs
@@ -2278,7 +2278,16 @@ pub async fn run_prompt_task(
             // AgentError means the agent caught a problem before mutating
             // session state (e.g. bad LLM response). The session is healthy —
             // don't invalidate it. Other errors may have corrupted state.
-            if !matches!(e, AcpError::AgentError { .. }) {
+            //
+            // Exception (#4369): -32603 "Internal error" from `session/prompt`
+            // means the underlying agent child process died out-of-band. The
+            // session is dead, not healthy — invalidate so the retry misses
+            // the session cache and create_session_and_apply_model builds a
+            // fresh session on the next attempt. False positives are cheap
+            // (one extra session/new); false negatives wedge the channel.
+            let is_dead_session =
+                matches!(e, AcpError::AgentError { code: -32603, .. });
+            if is_dead_session || !matches!(e, AcpError::AgentError { .. }) {
                 agent.state.invalidate(&source);
             }
             let usage = agent.acp.take_turn_usage();


mmroczka:
I found the same failure on 2026-08-08 on macOS. I used the official
Buzz.app build with `@agentclientprotocol/claude-agent-acp` 0.65.0.

Sequence of events:

1. My AI stopped four claude --session-id child processes with SIGTERM.
   The adapter workers continued to operate.
2. The adapter log showed: `Claude Agent process died: Claude Code
   process exited with code 143`.
3. Each new message to a bound channel caused the error `-32603
   Internal error`.
4. buzz-acp identified each error as "application error — pipe intact".
   It did not invalidate the session.
5. The queue tried each batch 10 times. Then it discarded the batch.
   I lost 11 owner events in two batches in approximately 3.5 hours.
6. The channels became operational again only after I stopped the
   adapter worker processes manually.

Code check:

I compared this PR with the source. The exemption in `pool.rs`
("AgentError means the agent caught a problem before mutating session
state") is the branch that keeps the dead session in
`sessions[channel_id]`. No other part of the retry path can correct
the map. The change in this PR causes the retry to miss the session
cache. Then `create_session_and_apply_model` makes a new session.
This agrees with my analysis of the same incident.

Note for reviewers:

Error `-32603` has more than one cause (see #4127: quota exhaustion,
"Prompt is too long"). Invalidation is correct for "Prompt is too
long". For a quota error, the cost is one extra `session/new` for
each attempt. Invalidation also discards the session context after a
temporary quota error. This cost is small. Do not delay this fix for
this reason.

I can test a build of this PR on macOS. Tell me if that helps.

iroiro147:
@mmroczka this is a really useful report — thank you. An independent reproduction on a different setup, with the adapter log line (`Claude Code process exited with code 143`) and the batch-loss count, is worth a lot more than my original synthetic repro.

Your code reading matches mine exactly: the `AgentError` exemption in `pool.rs` is the branch that keeps the dead session in `sessions[channel_id]`, and nothing else on the retry path can correct the map. Invalidating there makes the retry miss the cache so `create_session_and_apply_model` builds a fresh session.

On your reviewer note — you're right that `-32603` is overloaded, and #4127 is the case I'd worry about too (quota exhaustion, over-long prompt). This PR deliberately doesn't try to discriminate, and the reasoning is in the diff comment: a false positive costs one extra `session/new`, while a false negative wedges the channel until someone kills the adapter by hand — which is exactly the 3.5 hours and 11 lost events you just described. Asymmetric enough that I chose to over-invalidate.

That said, if reviewers would rather narrow it, the obvious tightening is to also match on the error payload rather than the code alone, so quota and prompt-length cases keep the current healthy-session behaviour. Happy to push that if it's preferred — it's a small change and I'd rather match the maintainers' taste than argue for mine.

Issue #3206, opened 2026-07-27
Multi-channel #h filters silently receive zero channel-scoped events


pookNast:
Subscribing with more than one channel UUID in a single `#h` filter succeeds at the protocol level (the relay sends EOSE, the WS stays open) but never delivers any EVENT frames. Events published to the listed channels are confirmed ingested by the relay and delivered to single-channel subscribers on those channels. They just never reach the multi-channel subscriber.

### Repro

Open an authenticated WS connection (NIP-42) and send:

```
["REQ", "sub1", {"kinds":[9], "#h":["<chanA-uuid>","<chanB-uuid>"], "limit":0}]
```

You get EOSE immediately. Publish a kind:9 event to either channel. The publisher returns success. Single-channel subscribers on each channel receive it. `sub1` receives nothing.

Now send two REQs instead:

```
["REQ", "subA", {"kinds":[9], "#h":["<chanA-uuid>"], "limit":0}]
["REQ", "subB", {"kinds":[9], "#h":["<chanB-uuid>"], "limit":0}]
```

Both receive events as expected.

### Why

`extract_channel_id_from_filters` in `crates/buzz-relay/src/handlers/req.rs:1009` returns `None` when more than one distinct channel UUID appears across the filters. The docstring at lines 999-1008 calls this "slow-path fan-out." But `fan_out_scoped` in `crates/buzz-relay/src/subscription.rs:265` only looks up subscriptions via `channel_kind_index`, and the scoping invariant at `subscription.rs:321-324` says channel-scoped events are not delivered to global subs. A `None` return registers the sub globally, so channel-scoped events never match it.

The end user sees a subscription that the relay accepted and EOSE'd on time, but that silently drops every event matching the published filter contract.

### Filter contract

NIP-01 allows arrays of values for any tagged filter. A client building a subscription that wants events from two channels should be able to send one REQ with two `#h` values. The current behavior treats this as a global subscription, which is not what the filter expresses.

### Proposed fix

When `extract_channel_id_from_filters` sees N > 1 distinct channels, register the subscription under each of those channels' `channel_kind_index` entries instead of returning `None`. Per-channel dispatch stays O(1). The scoping invariant is preserved because the sub is still tied to specific channels, just N of them instead of one. The "no channel tag" case (true global) keeps the existing None path.

### Workaround

Send one REQ per channel, each with a single `#h` value. The relay indexes each REQ separately and delivery works.

### Environment

[...]

PR #2452, opened 2026-07-23
fix(relay): stop collapsing multi-value #h filters to a single channel

taehalim:
## Problem

Fixes #2385.

A filter listing multiple distinct `#h` channels should match events from **any** of them (NIP-01 OR semantics), but `extract_channel_id_from_filter` in `handlers/req.rs` returned the first parseable value from the tag `BTreeSet` — i.e. the lexicographically-first channel UUID. Through `build_event_query_from_filter` this pinned `query.channel_id` to that single channel, so:

- `POST /query` (bridge catchall) composed a contradictory SQL query (`channel_id = $first` AND-ed with the access scope) and dropped every other listed channel's events — this is why the desktop Workflows overview renders empty: `get_channels_workflows` batches all member channels into one `#h` filter.
- The COUNT fallback paths (WS COUNT and `POST /count`) undercounted for the same reason.

## Fix

`extract_channel_id_from_filter` now returns `None` when the filter carries multiple distinct parseable `#h` UUIDs, mirroring the existing multi-filter rule in `extract_channel_id_from_filters`. Callers then take the access-scope path (`apply_access_scope_to_query` widens the SQL query to the caller's accessible channels) and the `filters_match` post-filter enforces the per-event `#h` OR — the same shape the WS REQ path already uses.

A filter whose `#h` values resolve to a single channel (duplicates, or unparseable extras alongside one UUID) still pins that channel, keeping the narrower SQL predicate.

## Testing

- Unit tests for the extractor: single value, multiple distinct values, duplicate spellings of the same UUID, unparseable values.
- New e2e regression test `test_query_multi_h_filter_returns_all_listed_channels`: two channels, one message each; `POST /query` with both ids in one `#h` filter must return both events, and `POST /count` must count both. Verified it fails on `main` with exactly the reported symptom (only the lexicographically-first channel's event comes back) and passes with this fix.
- `cargo fmt` + `clippy --all-targets --all-features` clean on the touched crates; full `e2e_relay` suite run against a local relay — 36/39 pass, and the 3 failures (invite mint, subscription limit, unarchive notification) fail identically on `main`, so they're local-env issues unrelated to this change.

Issue #2385, opened 2026-07-22
relay: multi-value #h filters collapse to lexicographically-first channel (Workflows overview shows empty)


leesalminen:
## Summary

The relay's `/query` (HTTP bridge) and REQ paths mishandle a filter with **multiple distinct `#h` values**: instead of Nostr's OR semantics ("match any listed channel"), the query silently collapses to the **lexicographically-first** channel UUID. Every other listed channel's events are dropped.

User-visible impact: the Desktop **Workflows overview screen is empty** for essentially everyone. `get_channels_workflows` batches all member-channel ids into one `#h` filter (`desktop/src-tauri/src/commands/workflows.rs`, whose comment asserts "A nostr `#h` filter matches ANY of its listed values") — but the relay only returns workflows for whichever channel UUID sorts first, typically a DM channel with none.

## Root cause

Two channel extractors disagree, and the bridge catchall composes them into a contradictory query:

- `extract_channel_id_from_filter` (`crates/buzz-relay/src/handlers/req.rs` — singular-filter variant) returns the **first** parseable `#h` value even when several are present (`for val in tag_values { ... return Some(id) }`). Filter tag values live in a `BTreeSet`, so "first" = lexicographic minimum.
- `extract_channel_from_filter` (`crates/buzz-relay/src/api/bridge.rs`) correctly requires `vs.len() == 1` and returns `None` for multi-value filters.

In the bridge catchall (`query_events_authed`), `build_event_query_from_filter` uses the first extractor → `query.channel_id = Some(<first-sorted id>)`, then `apply_access_scope_to_query` is passed the second extractor's `None` → also sets `query.channel_ids = accessible_channels`. Resulting SQL (captured via `log_statement=all`):

```sql
SELECT ... FROM events
WHERE community_id = $1 AND deleted_at IS NULL
  AND channel_id = $2                                   -- first-sorted #h value
  AND (channel_id IS NULL OR channel_id IN ($3,$4,$5))  -- access scope
  AND kind IN ($6) ...
```

The `channel_id = $2` clause pins the whole query to one channel.

Note the plural WS-subscription variant `extract_channel_id_from_filters` already handles this correctly ("Multiple distinct channel IDs — fall back to global"); the singular variant predates that rule.

## Repro (any relay, any community)

1. Create workflows in channel A (id `57e6...`). Confirm single-channel query returns them:
   `POST /query` `[{"kinds":[30620],"#h":["57e6..."]}]` → N events ✅
[...]

Issue #2373, opened 2026-07-22
[BUG]Workflow is created and runs, but is missing from the Workflows list on a self-hosted relay


ThisisBankole:
## Summary

  A webhook workflow can be created successfully on a self-hosted Buzz relay and runs correctly, but Buzz Desktop continues to show an empty Workflows page (`No workflows yet`).

  ## Environment

  - Buzz Desktop: v0.4.22
  - Relay: self-hosted Docker deployment
  - Relay URL: custom `wss://…` domain behind Cloudflare Tunnel
  - Channel: a newly created channel

  ## Steps to reproduce

  1. Connect Buzz Desktop to a self-hosted relay.
  2. Create a channel.
  3. Open **Workflows** and create a workflow for that channel:
     - Trigger: `webhook`
     - Action: `send_message`
     - Channel override: the channel’s UUID
  4. Save the workflow.

  ## Actual behavior

  - The create flow succeeds and shows a webhook URL plus `X-Webhook-Secret`.
  - Calling the webhook successfully posts messages into the selected channel.
  - Refreshing the Workflows page and fully restarting Buzz Desktop still shows `No workflows yet`.

  ## Expected behavior

  The newly created workflow should appear in the Workflows list and be available to view or edit.

  ## Additional evidence

  The relay confirms that the workflow exists and is enabled:

  - A workflow database record exists with the expected channel ID.
  - A signed kind `30620` workflow-definition event exists with the expected `d` and `h` tags.
  - The webhook executes successfully and posts its `send_message` action to the channel.

  This appears to be a Desktop workflow-list/read issue rather than workflow creation or execution failing.

  ## Impact

  The workflow is operational, but it cannot be managed through Buzz Desktop after creation.

AaronGoldsmith:
## Root cause confirmed on current `main`

This is reproducible on hosted relays too, so it is not specific to self-hosting, webhooks, or Cloudflare Tunnel. The workflow exists and can be returned by a channel-scoped CLI query, while Desktop shows `No workflows yet`.

The failure is in Desktop's batched multi-channel workflow query:

1. `WorkflowsView` collects all member channel IDs, sorts them, and calls `getChannelsWorkflows(channelIds)` (`desktop/src/features/workflows/ui/WorkflowsView.tsx:79-103`).
2. The Tauri command sends one Nostr filter with every ID in a multi-value `#h` tag (`desktop/src-tauri/src/commands/workflows.rs:69-97`):

   ```json
[...]

PR #4393, opened 2026-08-02
fix(4381): return None for multi-value #h in singular req extractor

### LINKED ISSUE #4381: `POST /query` returns events from only one channel when a filter has multiple `#h` values
**Describe the bug**

On the HTTP bridge (`POST /query`), a filter carrying multiple `#h` values returns events
from only one channel — the lexicographically smallest. Under NIP-01 the tag list is an OR
set, so every listed channel should match. WS `REQ` handles the same filter correctly; only
the bridge is affected.

This silently empties the desktop Workflows overview, which sends exactly this filter
(`desktop/src-tauri/src/commands/workflows.rs:87` → `query_relay` → `POST /query`). The
workflows exist and run on schedule, but the view shows "No workflows yet". NIP-45 `COUNT`
undercounts the same way, via the same query builder (`handlers/count.rs:153`, `:225`;
`api/bridge.rs:1468`, `:1536`).

**Steps to reproduce**

1. Join two channels, A and B, where only B contains a `kind:30620` workflow, and A's channel
   UUID sorts lexicographically before B's.
2. Open the Workflows view.
3. The list is empty.

Directly against the bridge, with a NIP-98-signed `POST /query`:

```jsonc
[{"kinds":[30620],"#h":["<chA>","<chB>"]}]   // returns only chA's events
[{"kinds":[30620],"#h":["<chB>"]}]           // returns chB's events
```

**Expected behavior**

The multi-value filter returns events from both channels, matching NIP-01 OR semantics and
the existing WS `REQ` behavior. B's workflow appears in the Workflows view.

**Version and platform**

- Buzz version: 0.5.2 (desktop); relay self-hosted from `ghcr.io/block/buzz`, Postgres 17
- OS: macOS 26.6

**Logs / additional context**

Two `#h` extractors disagree inside the same request, at `api/bridge.rs:1227`:

```rust
let mut query = req::build_event_query_from_filter(filter, ...).await;
req::apply_access_scope_to_query(
    &mut query,
    extract_channel_from_filter(filter),
    &accessible_channels,
);
```

- `build_event_query_from_filter` uses the singular `extract_channel_id_from_filter`
  (`req.rs:851`), which returns the **first** `#h` UUID → `EventQuery.channel_id = Some(chA)`.
- `apply_access_scope_to_query` is passed `extract_channel_from_filter` (`bridge.rs:235`),
  which returns `None` for a multi-value `#h` → it sets `channel_ids = accessible` but does
  not clear `channel_id`.

Both predicates reach the SQL:

```sql
AND channel_id = chA
AND (channel_id IS NULL OR channel_id IN (<accessible>))
```

The `filters_match` post-filter (`bridge.rs:1299`) can't recover — chB's rows were never
fetched. Which value survives is server-side and deterministic: `nostr`'s `GenericTags` is
`BTreeMap<SingleLetterTag, BTreeSet<String>>`, so "first" is always the lexicographically
smallest `#h`.

A minimal fix is to return `None` for a multi-value `#h`, matching the sibling extractors at
`bridge.rs:235` and `count.rs:17`. Callers then treat the filter as global, widen to the
accessible set, and let `filters_match` enforce the OR set — what WS `REQ` already does:

```rust
         if key == "h" {
-            for val in tag_values {
-                if let Ok(id) = val.parse::<uuid::Uuid>() {
-                    return Some(id);
-                }
+            if tag_values.len() != 1 {
+                return None;
             }
+            return tag_values.iter().next()?.parse::<uuid::Uuid>().ok();
         }
```

On `origin/main` @ `b7bb1512`, a test asserting the multi-value case fails before this change
and passes after; `handlers::req::tests` is 50/50 with three added tests, and
`cargo fmt -p buzz-relay -- --check` is clean. (The 9 `api::media` / `api::admin` failures in
the full lib suite are `Sqlx(PoolTimedOut)` and fail identically without the patch.)

iroiro147:
## Summary

Fixes #4381

On the HTTP bridge (`POST /query`), a filter carrying multiple `#h` values returned events from only one channel — always the lexicographically smallest — silently dropping events from every other listed channel. The Desktop Workflows overview (which sends exactly this multi-`#h` filter) showed "No workflows yet" even though workflows existed and ran on schedule. NIP-45 `COUNT` undercounted the same way via the same query builder.

## Root cause

Two `#h` extractors disagreed inside the same request (`api/bridge.rs:1227`):

- `build_event_query_from_filter` used the singular `extract_channel_id_from_filter` (`handlers/req.rs:851`), which returned the **first** parseable `#h` UUID. Since `nostr`'s `GenericTags` is `BTreeMap<SingleLetterTag, BTreeSet<String>>`-backed, "first" is the lexicographically smallest channel ID.
- `apply_access_scope_to_query` used the bridge's `extract_channel_from_filter` (`bridge.rs:235`), which correctly returns `None` for multi-value `#h`.

The narrow `channel_id` predicate reached SQL, so the other channels' rows were never fetched and the `filters_match` post-filter couldn't recover them.

## Fix

The singular extractor at `handlers/req.rs:851` now returns `None` for a multi-value `#h`, matching its three siblings:

- `api/bridge.rs:235` — already returns `None`
- `handlers/count.rs:17` — already returns `None`
- `handlers/req.rs:1022` (plural) — already widens to global on distinct channels

Returning `None` widens the database query to the caller's accessible channel set and lets the `filters_match` post-filter enforce the NIP-01 OR set — the same path WS `REQ` already takes.

## Testing

- `cargo test -p buzz-relay handlers::req::tests` — 50/50 pass
- `cargo fmt -p buzz-relay -- --check` — clean
- 3 new regression tests for `extract_channel_id_from_filter` (the singular form had no direct unit coverage before; all existing tests targeted the plural `extract_channel_id_from_filters`):
  - `test_extract_channel_id_from_filter_single_channel` — single `#h` still returns the channel
  - `test_extract_channel_id_from_filter_multi_value_h_returns_none` — OR set widens to global
  - `test_extract_channel_id_from_filter_no_h_tag_returns_none` — absent `#h` is global

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -848,15 +848,20 @@ fn filters_are_nip43_membership_only(filters: &[Filter]) -> bool {
 }
 
 /// Extract a channel UUID from a single filter's `#h` tag.
+///
+/// Returns `None` for a multi-value `#h` — NIP-01 treats the tag list as an
+/// OR set, so scoping to just the first channel would drop events from the
+/// rest (#4381). Callers widen to the accessible set and let `filters_match`
+/// enforce the OR. This matches the sibling extractors at `api/bridge.rs:235`
+/// and `handlers/count.rs:17`.
 fn extract_channel_id_from_filter(filter: &Filter) -> Option<uuid::Uuid> {
     for (tag_key, tag_values) in filter.generic_tags.iter() {
         let key = tag_key.to_string();
         if key == "h" {
-            for val in tag_values {
-                if let Ok(id) = val.parse::<uuid::Uuid>() {
-                    return Some(id);
-                }
+            if tag_values.len() != 1 {
+                return None;
             }
+            return tag_values.iter().next()?.parse::<uuid::Uuid>().ok();
         }
     }
     None
@@ -1577,6 +1582,33 @@ mod tests {
             filter_with_channel(channel_id),
         ];
         assert_eq!(extract_channel_id_from_filters(&filters), Some(channel_id));
+    }
+
+    #[test]
+    fn test_extract_channel_id_from_filter_single_channel() {
+        let channel_id = uuid::Uuid::new_v4();
+        let filter = filter_with_channel(channel_id);
+        assert_eq!(extract_channel_id_from_filter(&filter), Some(channel_id));
+    }
+
+    #[test]
+    fn test_extract_channel_id_from_filter_multi_value_h_returns_none() {
+        // #4381: a multi-value #h must not silently scope to the first channel —
+        // NIP-01 treats it as an OR set. Returning None widens the query to the
+        // accessible set so filters_match can enforce the OR.
+        let channel_a = uuid::Uuid::new_v4();
+        let channel_b = uuid::Uuid::new_v4();
+        let filter = Filter::new().custom_tags(
+            SingleLetterTag::lowercase(Alphabet::H),
+            [channel_a.to_string(), channel_b.to_string()],
+        );
+        assert_eq!(extract_channel_id_from_filter(&filter), None);
+    }
+
+    #[test]
+    fn test_extract_channel_id_from_filter_no_h_tag_returns_none() {
+        let filter = Filter::new();
+        assert_eq!(extract_channel_id_from_filter(&filter), None);
     }
 
     #[test]


damonrand:
Production verification from a self-hosted relay — this exact change fixes the desktop "No workflows yet" symptom end to end.

**Setup.** Our relay was pinned to `82f7ed15` (the 0.5.14 release commit). We independently root-caused the empty desktop Workflows overview to `extract_channel_id_from_filter` collapsing a multi-value `#h` to its first (lexicographically smallest, since `generic_tags` values iterate sorted) UUID, then wrote a patch before finding this PR — it is functionally identical to this diff (same `len() == 1` guard, mirroring the WS REQ inline guard and `api/bridge.rs`'s `extract_channel_from_filter`, plus the same unit-test shape). We built and deployed `82f7ed15` + that patch to our Fly.io deployment (Postgres/Redis, `BUZZ_REQUIRE_RELAY_MEMBERSHIP=true`).

**Repro & proof** — NIP-98-signed `POST /query` with `{"kinds":[30620],"#h":[...]}`, same live workflow definition, same authed member:

| `#h` values | unpatched `82f7ed15` | patched |
|---|---|---|
| `[<real channel>]` | 1 event | 1 event |
| `[<real channel>, 00000000-…]` (decoy sorts first) | **0 events** | **1 event** |
| all 13 member channel ids batched (what desktop `get_channels_workflows` sends) | 0 events | **1 event** |

The desktop Workflows overview now lists the workflow that was invisible before.

Note for triage: this bug has accumulated many duplicates — issues #2373, #2385, #3206, #4532, #4579, #4659, #4804 and PRs #2452, #4446, #4581, #4624, #5307, #5339 all describe or fix the same root cause. This PR is the minimal relay-side fix.

iroiro147:
@damonrand this is a model verification — thank you for taking it all the way to a deployed relay rather than a local repro.

Two details in your table are worth calling out for anyone triaging later:

- The **decoy-sorts-first row** (`[<real channel>, 00000000-…]` → 0 events unpatched) is the cleanest possible demonstration of the root cause: `generic_tags` iterating sorted means the collapse isn't even stable per-request — it's *deterministically wrong* whenever any lexicographically-smaller UUID is present. That's why the symptom felt random across installs.
- The **13-channel batch row** reproduces exactly what desktop `get_channels_workflows` sends, so your table isn't an approximation of the desktop bug — it *is* the desktop bug, end to end.

And the independent convergence on the same `len() == 1` guard (mirroring the WS REQ inline guard and `extract_channel_from_filter` in `api/bridge.rs`) is good evidence this is the minimal correct shape rather than one of several plausible fixes.

Your duplicate map (#2373, #2385, #3206, #4532, #4579, #4659, #4804 / PRs #2452, #4446, #4581, #4624, #5307, #5339) is genuinely valuable for triage — one merge here closes a lot of open surface. Nothing needed from my side: the branch is current and the unit tests cover both the singular and multi-value paths. Thanks again — verification like this is worth more than the patch itself.

PR #4396, opened 2026-08-02
fix(mobile): retry messages after transient disconnects

tkors:
## Summary
- keep unacknowledged mobile events pending across transient relay reconnects
- replay the same signed event ID so a lost acknowledgement cannot duplicate a message
- cover the reported nested Alies thread reply and reconnect path with regressions

## Test Plan
- [x] `just mobile-check`
- [x] `just mobile-test`
- [x] verify a nested reply carries the outer root and direct parent tags
- [x] verify an unacknowledged event is replayed and completes after reconnect

### DIFF
--- a/mobile/lib/shared/relay/relay_session.dart
+++ b/mobile/lib/shared/relay/relay_session.dart
@@ -50,10 +50,15 @@ class _LiveSubscription {
 }
 
 class _PendingEvent {
+  final NostrEvent event;
   final Completer<NostrEvent> completer;
   final Timer timeout;
 
-  _PendingEvent({required this.completer, required this.timeout});
+  _PendingEvent({
+    required this.event,
+    required this.completer,
+    required this.timeout,
+  });
 }
 
 class _BufferedEvent {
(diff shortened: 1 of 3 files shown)


tkors:
@claude review

tkors:
@claude review\n\nFollow-up commit addresses the prior local Claude findings: the normal 8s acknowledgement timeout is restored, reconnect gets its own bounded grace period, and reconnecting/already-acked/auth-rejected paths now have regression coverage.

tkors:
@claude review\n\nSecond follow-up resolves the remaining local Claude findings: retries now have one absolute deadline across flaps, only kind-9 chat messages replay, response-bearing command events retain the original failure behavior, and terminal auth rejection blocks implicit reconnect.

tkors:
@block/buzz-oss-team Could you review this mobile reliability fix?

It prevents a chat message typed during a transient relay reconnect from being returned to the composer instead of delivered — the exact Android failure observed after an Alies thread reply. The change retries only plain chat events, reuses the event ID for relay deduplication, and keeps command events non-replayable. Targeted and full mobile tests (1,028; one existing skip) plus CI are green. No relay/server behaviour changes.

This review is the remaining prerequisite for an official Android update.

PR #4401, opened 2026-08-02
fix(git): allow large smart HTTP pushes and fetches (#2880, #4423)

### LINKED ISSUE #2880: git push fails because the NIP-98 credential is scoped to discovery GET and is not sent on git-receive-pack POST
**Describe the bug**

Pushing to a Buzz-hosted Git repository fails with HTTP 401 even when repository ownership, relay authorization, and signing credentials are valid. Authentication succeeds for receive-pack discovery, but Git sends the subsequent streamed POST without an Authorization header. The relay challenges that POST, but Git cannot replay it after receiving the challenge.

**Steps to reproduce**

1. Configure a valid Buzz repository remote and the `git-credential-nostr` credential helper.
2. Push a branch with `git push`.
3. Inspect the HTTP exchange:
   1. Git probes `GET /git/<owner>/<repo>/info/refs?service=git-receive-pack`.
   2. The relay responds with a Nostr authentication challenge scoped to GET.
   3. `git-credential-nostr` generates a GET-scoped NIP-98 credential; Git retries discovery and receives HTTP 200.
   4. Git sends `POST /git/<owner>/<repo>/git-receive-pack` without an Authorization header.
   5. The relay responds with HTTP 401 and `WWW-Authenticate: Nostr realm="buzz", method="POST"`.
   6. The push aborts because Git cannot replay the streamed POST after the challenge.

**Expected behavior**

A valid push should authenticate the `git-receive-pack` POST preemptively with a fresh, POST-scoped NIP-98 credential and complete without a 401 challenge.

**Version and platform**

- Buzz version: unknown
- OS: macOS
- Git: 2.50.1
- Buzz Git relay using NIP-98 authentication
- `git-credential-nostr` credential helper

**Logs / additional context**

Root-cause evidence:

- A direct POST carrying a helper-generated, POST-scoped NIP-98 credential returns HTTP 200.
- Injecting the same kind of credential preemptively makes the push succeed:

  `git -c http.extraHeader="Authorization: Nostr <POST_SCOPED_TOKEN>" push`

- The successful authenticated POST rules out repository ownership, auth-tag, relay provisioning, and repository-permission failures.

This indicates an authentication-handshake mismatch: discovery authenticates a GET, but that credential is neither valid for nor attached to the later POST. Challenging only after the streamed POST begins is too late for Git to recover.

Workaround:

Generate a fresh POST-scoped NIP-98 token and attach it preemptively using `http.extraHeader` for the push. Tokens should remain short-lived and must not be logged or persisted in repository configuration.

Recommended fix:

Before invoking `git-receive-pack`, have the Buzz Git integration attach a fresh NIP-98 Authorization header whose URL and method match the POST request. Alternatively, establish a short-lived authenticated session during discovery that is accepted by the subsequent POST.

Please also add an integration test covering the full smart-HTTP push sequence and asserting that the POST succeeds without relying on a replay after a 401 challenge.

### LINKED ISSUE #4423: Git smart-HTTP: unauthenticated 0000 probe is only admitted on receive-pack — fetch with large negotiation (in-limit repo + divergent local history) still 401-aborts
## Summary

PR #4401 admits Git's unauthenticated 4-byte `0000` compatibility probe on the **receive-pack** route, fixing large pushes (#2880). But the probe is not receive-pack-specific: in Git's `remote-curl.c` (read at tag v2.50.1), `probe_rpc` (:868) is called inside `post_rpc` (:943) for *any* request that exceeds `http.postBuffer`, and `post_rpc` serves **both** services — `fetch_git` → `rpc_service` → `post_rpc` (:1245) and `push` → `rpc_service` (:1406), plus `stateless_connect` (:1458) for protocol v2.

So the identical 401-abort exists on **clone/fetch** when the upload-pack negotiation body crosses `http.postBuffer`, and #4401's middleware — mounted on the receive-pack route only — does not cover it.

## Measured reproduction (not just a code-reading claim)

Sami measured this against real git 2.51.0 with a local `git-http-backend` and a 401 injected on exactly the unauthenticated 4-byte POST (simulating `GitAuth`), with a fake credential helper advertising `authtype=Nostr` (mirroring `git-credential-nostr`'s non-curl-native scheme):

```text
# 25,000-ref repo, DEFAULT config — no http.postBuffer override
REQ POST /big.git/git-upload-pack CT=...x-git-upload-pack-request CL=4    body=0000
REQ POST /big.git/git-upload-pack CT=...x-git-upload-pack-request TE=chunked bodylen=1250124

# with 401 injected on the probe:
REQ POST /big.git/git-upload-pack -> 401 CHALLENGE (probe, body=b'0000')
REQ POST /big.git/git-upload-pack -> 401 CHALLENGE (probe, body=b'0000')
error: RPC failed; HTTP 401 curl 22
```

Negative control: same 401-injecting server, 1,500-ref repo → `clone rc=0`, zero probes. 16,000 refs did **not** trigger it; 25,000 did. The boundary matches the arithmetic: negotiation pkt-lines run ~50 bytes/ref, so 1 MiB (default `http.postBuffer`) ÷ ~50 ≈ 20k refs.

Both probe retries arrived unauthenticated — behavioral confirmation that `probe_rpc` never calls `http_append_auth_header` and the REAUTH loop cannot help a custom authtype.

## Reachability — **corrected 2026-08-02, see second comment below**

~~Low-likelihood today: the read-path trigger is ref-count driven (O(20k) refs)~~ **This framing was incomplete.** Ref count is the right model for **clone** (uniform `want` lines, measured at exactly 50.00 bytes/ref across 1.5k–25k refs plus a constant 124-byte header, so 1 MiB ÷ 50 ≈ 21k refs) — and Buzz's `MAX_MANIFEST_REFS = 10_000` (`manifest.rs:47`, enforced at `cas_publish.rs:329`) keeps clone permanently under the threshold at ~0.48 MiB worst case (0.24 MiB measured on a real-OID fixture). **Fetch is the reachable case** because `have` lines come from the *client's local* commits, adding a term no server-side cap bounds — and `http.postBuffer` gates total *decoded* negotiation bytes. Measured at the 10,000-ref ceiling: a fetch with divergent local history crosses 1 MiB decoded (1.46 MiB measured) and hits the 401 probe abort; clone of the same repo stays at 0.24 MiB and never probes.

So the trigger is not "a repo too big for Buzz to host" — it is **an in-limit repo plus a developer with substantial unpushed local work.** Live today. Note the wire bytes are gzipped (~450 KB); only decoded lengths reveal the crossing. Still correctly not held against #4401, but this is a real user-facing failure mode, not a corner case.

## Fix shape — **now measured end-to-end, see first comment below**

Small: the #4401 middleware's predicate is keyed on the receive-pack MIME (`application/x-git-receive-pack-request`, `Content-Length: 4`, body `0000`). Covering upload-pack is a second mount on the upload-pack POST route with the matching content type (`application/x-git-upload-pack-request`) — no new logic. Same security shape: path-independent empty 200, no tenant resolution, no state.

This was originally filed as a *proposal*; it has since been verified against real git 2.51.0 with a full `GitAuth` simulation (every git route 401s without `Authorization`): admitting the upload-pack probe makes the 25,000-ref clone succeed with `fsck` clean and exact ref parity with origin. One implementation detail pinned by that measurement: `probe_rpc` only checks the response status code, so the response content-type is not load-bearing — the implementation should still emit the matching MIME per service, and a single parameterized middleware covering both mounts prevents anyone getting half-working behavior by picking wrong. Full receipts in the comment below.

## Credit

Found and measured by Sami during the #4401 review (buzz-outside-pr-reviews thread, 2026-08-02). Call graph independently re-verified against `remote-curl.c` v2.50.1 by Eva.

Refs: #4401, #2880.

MajorTal:
Split per review: this PR is only the Git smart HTTP probe. The request-time policy for the API and media routers lives in #7753, which also fixes the post-commit cancellation Carl found.

## What this fixes

**#2880.** When a push is bigger than `http.postBuffer`, Git first sends a tiny unauthenticated POST with only the 4-byte flush packet `0000`. We answered `401`, Git gave up, large pushes were impossible. The fix admits exactly that probe: no `Authorization`, no `Content-Encoding`, no `Transfer-Encoding`, exactly one `Content-Length: 4`, body exactly `0000`. Anything else falls back to the normal auth path.

**#4423.** The same probe fires before big fetches. Same middleware, parameterized over the (request MIME, result MIME) pair and mounted on upload-pack too.

**Probe body deadline.** Admitting the probe means reading four unauthenticated body bytes. That read has its own 10 s deadline that fails closed into normal auth, so a client that sends the probe headers and withholds the body cannot park a task. Git and curl send the headers and the `0000` in the same segment, so a real probe never waits on it. This is the only timeout in this PR.

## Commits

Four, rebased on current main, all signed off:

1. `fix(git): allow large smart HTTP pushes`
2. `fix(git): admit the compatibility probe on upload-pack too`
3. `fix(git): bound the compatibility probe's body read`
4. `ci(unit): run the git probe regression tests in the unit lane`

Changes since the last push: the probe deadline went from 60 s to 10 s, the fallthrough branch now documents the invariant it relies on (both handlers extract `GitAuth` from the request parts before touching the body, so a candidate without `Authorization` is answered 401 without decoding), the four probe tests are wired into `just test-unit` (no term in its nextest expression matched them, so hosted CI compiled them and ran none), and two commits that carried an npub as git author are now authored normally.

## Tests

- Exactness and path-independence matrix across both services, 2 path shapes and 11 near misses each, plus a cross-service case proving each route rejects the other's request MIME.
- Probe body deadline: withheld body answers 401 at the bound; a complete probe still passes.
- Live regression (ignored, needs relay + MinIO): pushes an incompressible 2 MiB pack through Apple Git 2.50.1, observes `Content-Length: 4` then `Transfer-Encoding: chunked`, clones the bytes back byte for byte.

## Review done before this push

We ran Block's own gates and reviews locally on this exact head, macOS Apple Silicon:

- `just fmt-check`, `just clippy` (workspace, `-D warnings`), `just test-unit` (cargo-nextest), `cargo-deny check`, `just file-size-check`, `just security-review-check`: clean.
- The Postgres lane (`scripts/postgres-test-run.sh` against Postgres and Redis containers): 390 passed.
- A reviewer run with the repository's Codex security review prompt over the exact range: no findings.
- An adversarial review shaped like Carl's: no blocking findings. It drove the real middleware over a loopback hyper server with raw HTTP/1.1 (exact probe, duplicate and malformed `Content-Length`, over- and under-delivered bodies, chunked, mixed-case headers, `Expect: 100-continue`) and confirmed the predicate and the fail-closed path. Its non-blocking items are the changes listed above.

Hosted Rust CI has not run on this head; it needs a maintainer to approve the workflow for this fork branch. The Codex security review needs a fresh authorization for the new range.

Fixes #2880
Fixes #4423


MajorTal:
@block/buzz-oss-team - poke on this one, gently. It is green and merges clean against main, so no workflow approval is needed just to look at it.

Reason I am pinging instead of waiting: one of the three issues in here is #4424, where the relay bounds request bytes but never request time. Valid headers, then withhold the body, and the task parks forever. Unauthenticated. I sat 200 sockets on it to confirm.

I also just rewrote the description. It was stale, still describing only the original large-push bug from back when that was all this was, which made the diff size look alarming for no reason. It now explains why three issues ended up in one PR (short version: same middleware, and each fix makes the next one necessary).

If you would rather review the DoS part on its own, say so and I will split #4424 out. Either way works, I just do not want it sitting.

crates/buzz-relay/src/router.rs
```diff
@@ -27,6 +29,121 @@ use crate::nip11::{nip11_document, relay_info_handler};
...
+fn with_request_deadline<S>(router: Router<S>, body_limit: usize, timeout: Duration) -> Router<S>
+where
+    S: Clone + Send + Sync + 'static,
+{
+    router
+        .layer(RequestBodyLimitLayer::new(body_limit))
+        .layer(TimeoutLayer::with_status_code(
+            StatusCode::REQUEST_TIMEOUT,
+            timeout,
+        ))
```
wesbillman, line 106: **[P1] Do not time out the entire API mutation future**

This helper is applied to `/events` and other writes, but their postcommit work is not cancellation-safe. `ingest_event` inserts the event (`handlers/ingest.rs:3172-3204`), then `dispatch_persistent_event` awaits the intentionally bounded audit queue **before** spawning fanout/workflows (`handlers/event.rs:358-374,574-597`). If the queue is full when this 60s wall deadline expires, `TimeoutLayer` drops that future: the DB event remains, but its audit entry and delivery/workflow task are lost. The same event retried with a fresh NIP-98 header exits as `duplicate:` at `ingest.rs:3206-3211`, before either can be repaired. This can occur under sustained audit pressure, or when a valid slow body consumes most of the total budget and only a short postcommit wait remains.

A paused-clock seam harness using the verbatim timeout and audit/dispatch functions reproduces base=200/audit+dispatch versus head=408/stored/no-audit/no-dispatch; DB storage is stubbed, not live integration. The new collector tests prove cancellation, not mutation recovery.

Bound body reception rather than the mutation service future, or split this unrelated API policy out while retaining the local Git probe timeout. Simply returning 408 after commit is not a safe failure mode here.

wesbillman (changes requested):

**Changes requested.** Reviewed `4ab4f786085a23fe6126529861840eff6048ceee...36fab5314173da1d6c34f170ee6eccce41a22dcb`.

### P1: The new API wall deadline can strand committed events without audit or delivery

The inline finding describes a concrete postcommit cancellation path, not just an ambiguous timeout response. With audit backpressure, the new deadline cancels event processing after DB insertion but before audit enqueue and before fanout/workflow scheduling. Reposting the same event with fresh HTTP authorization returns `duplicate:` and does not repair the lost work. Ownership transfer, create-only community provisioning, and invite claim also await publication after committing; their retries are not a universal recovery mechanism.

I reproduced the event cancellation with a paused-clock standalone harness using the **verbatim production timeout, audit-enqueue, and dispatch-scheduling functions**. DB insertion/duplicate semantics and downstream delivery were stubbed, so this is a control-flow reproduction, **not** live relay/Postgres integration:

```text
BASE: original=200 OK retry=200 OK stored=true audit=Some("00") dispatched=true
HEAD: original=408 Request Timeout retry=200 OK stored=true audit=None dispatched=false
```

### Scope and smallest safe correction

The exact four-byte probe shim on both Git services is a narrow compatibility fix; I found no auth bypass or repository-data disclosure in that changed path. Its local `PROBE_BODY_TIMEOUT` already bounds the newly introduced unauthenticated read. **The API/media-wide timeout policy is not required to make that probe safe.** I recommend taking the offered split: keep the push/fetch compatibility shim and local body deadline here, and review #4424’s wider policy separately. That separate fix should bound request-body consumption without cancelling irreversible mutations, or provide tested durable/idempotent recovery. No new user-facing feature was found in the diff, but new global timeout semantics are meaningful scope expansion.

### Media disposition

I did not establish a second blocking media defect. The 3600s whole-upload ceiling can still interrupt storage/publication after a valid body finishes near the deadline; Blossom auth is checked before those operations, not continuously through them. A missing sidecar leaves an unservable orphan and content-addressed retry can repair it. S3 cancellation/late-commit behavior was not independently reproduced, so I am recording this as a residual availability/orphan risk, not claiming corruption or an auth bypass. The idle timeout/error-classification changes themselves preserve the reviewed 408/413/500 distinction.

### Validation and limits

- Inspected all ten changed files and traced Git auth/authorization, body-limit layering, API postcommit paths, media body/error/storage lifecycle, and retry behavior. Independent Git/router/media review lanes were consolidated rather than treated as a blanket safety certificate.
- The unchanged probe implementation and all four new probe unit tests passed in an isolated package with key dependencies pinned to the production versions. The cancellation harness reproduced the regression. Neither result is a full `buzz-relay` build/test run.
- I did not run the ignored live-relay/MinIO Git E2E or independently reproduce the author’s live push/fetch measurements. The current GitHub check snapshot does not show Rust build/test execution, and the security-review comment marks this exact range as requiring a fresh review. Green metadata checks are not evidence those validations ran.

**Exit criteria:** remove the outer API mutation deadline (the minimal split), or demonstrate cancellation-safe durable postcommit completion/retry through the real event and membership mutation paths. Retain the local Git probe deadline and obtain exact-head relay test plus live large-request workflow evidence before merge.

PR #4411, opened 2026-08-02
fix(desktop): stop the create-agent provider config probe from erasing keystrokes

tlongwell-block:
## What

Fixes the create-agent dialog's "Run on" provider config fields eating keystrokes — reported by Tyler in buzz-remote-agents (channel `29414326`, thread `db76677a`): the Kubernetes **Kubeconfig context** field would not accept typing.

## Why it happened (the Typewriter Eraser, shipped in #4289)

`WhereToRunSection`'s probe `useEffect` depended on the whole `draft`:

1. every keystroke changed the draft → effect re-fired → provider binary re-probed;
2. each probe result is a fresh object written into the draft → the effect re-triggered **itself**, respawning the provider binary in a loop for as long as the dialog sat on a provider;
3. every probe resolution reset `providerConfig` to schema defaults — erasing whatever was typed. A field with no schema default (`context`) snapped back to empty, i.e. "won't let me type". Unrelated to how many kubeconfig contexts you have.

## Fix

- **Probe once per provider selection**, keyed on the provider's stable `binaryPath` — not the draft, not the provider object (a `useBackendProvidersQuery` refresh must not reprobe an unchanged selection).
- **Latest-state resolution** via `React.useEffectEvent` + a new pure `applyProbeResult` helper: schema defaults merge **beneath** the current `providerConfig`, so a probe landing after the user typed can never clobber in-flight input (per Wren's pre-patch red-team: changing deps alone leaves a stale closure).

Existing `cancelled` cleanup keeps provider-switch/unmount safe; selection reset (`emptyWhereToRunDraft`) and the fail-closed probe-error path are unchanged.

## Tests

- **Unit** (`whereToRunIntent.test.mjs`): `applyProbeResult` merge semantics — defaults under typed values, user-cleared fields stay cleared, schema-less results, unrelated fields preserved.
- **E2E** (new `where-to-run-config.spec.ts`, added to the smoke project, **red-first verified**: all 3 fail against the unfixed component):
  - typing into a defaultless provider field sticks, and `probe_backend_provider` fires exactly once per selection;
  - the config form is gated on probe resolution (slow probe: no half-rendered form, defaults prefill once);
  - provider → local → provider re-probes and resets cleanly.
- Mock bridge gains `backendProviders` / `backendProviderProbeResult` / `backendProviderProbeDelayMs` seams (defaults preserve prior behavior).

## Verification at 8eb7680

- `pnpm check` + `tsc` clean, `pnpm test` 3926/3926;
- new spec 3/3 green (and 3/3 red on the unfixed component);
- pre-push lefthook: desktop-test, desktop-check, desktop-tauri-checks, rust-tests, mobile-test all green.

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -132,6 +132,7 @@ export default defineConfig({
         "**/harness-management.spec.ts",
         "**/harness-catalog-screenshots.spec.ts",
         "**/inline-custom-harness.spec.ts",
+        "**/where-to-run-config.spec.ts",
         "**/huddle-transcription.spec.ts",
       ],
       use: {
--- a/desktop/src/features/agents/ui/WhereToRunSection.tsx
+++ b/desktop/src/features/agents/ui/WhereToRunSection.tsx
@@ -5,7 +5,11 @@ import { useBackendProvidersQuery } from "@/features/agents/hooks";
 import { probeBackendProvider } from "@/shared/api/tauri";
 
 import { ProviderConfigFields } from "./ProviderConfigFields";
-import { emptyWhereToRunDraft, type WhereToRunDraft } from "./whereToRunIntent";
+import {
+  applyProbeResult,
+  emptyWhereToRunDraft,
+  type WhereToRunDraft,
+} from "./whereToRunIntent";
 
 /** Optional remote-backend selector. Buzz shared compute is an LLM provider, not a run destination. */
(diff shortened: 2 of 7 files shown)


Chessing234:
the useEffectEvent + path-keyed probe fix reads right — depending on draft was a nasty eraser loop. glad this landed in the create-agent dialog.

PR #4446, opened 2026-08-03
fix(desktop,relay): restore multi-channel workflows list

Trevongit:
## Summary

Desktop Workflows showed **No workflows yet** after successful saves even when CLI `buzz workflows list --channel <uuid>` returned rows.

**Root cause:** `get_channels_workflows` packed all member channel IDs into one multi-value `#h` filter. Relay `extract_channel_id_from_filter` collapsed multi-`#h` to the lexicographically first channel UUID and pushed that as SQL `channel_id`, so the query often returned only an empty channel (e.g. `#general`).

**Fix (defense in depth):**
1. **Desktop** — one single-value `#h` filter per channel in a single `POST /query` (NIP-01 OR; each filter is SQL-pushable).
2. **Relay** — multi-value `#h` returns `None` from the singular extractor (same as bridge/count siblings), so multi-`#h` widens correctly instead of pinning one channel.
3. **UX** — avoid flashing the empty state while channel membership is still loading.

Live Groundfeed repro before this change:
- `#h: [welcome]` → workflows present
- `#h: [welcome, skills, general]` → **0**
- two single-`#h` filters OR → full set

## Related work on block/buzz

Others hit the same bug; these open PRs address the **relay** half (not merged as of this PR):

- https://github.com/block/buzz/pull/2452 — stop collapsing multi-value `#h`
- https://github.com/block/buzz/pull/4393 — same extractor fix for #4381
- https://github.com/block/buzz/pull/2350 — preserve multi-channel workflow queries

This PR is complementary: it keeps the Desktop list correct even against unfixed relays, and still hardens the relay path.

## Test plan

- [x] `cargo test -p buzz-relay multi_h_filter_does_not_collapse` / `single_h_filter_extracts` pass
- [x] `cargo test --manifest-path desktop/src-tauri/Cargo.toml workflows` (11 unit tests) pass
- [ ] Rebuild Desktop from this branch; open Workflows → list should show workflows across member channels (not empty when CLI list is non-empty)
- [ ] Optional: against Groundfeed, confirm multi-channel overview after Desktop rebuild

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -847,19 +847,34 @@ fn filters_are_nip43_membership_only(filters: &[Filter]) -> bool {
         })
 }
 
-/// Extract a channel UUID from a single filter's `#h` tag.
+/// Extract a **single** channel UUID from a filter's `#h` tag.
+///
+/// Returns `Some(id)` only when the filter names exactly one channel UUID.
+/// Multi-value `#h` (NIP-01 OR across channels) returns `None` so callers do
+/// not push an arbitrary first value as `EventQuery.channel_id` — that bug
+/// made Desktop's batched workflow list return empty while per-channel CLI
+/// list worked (multi-`#h` collapsed to one channel that often had zero
+/// workflows).
 fn extract_channel_id_from_filter(filter: &Filter) -> Option<uuid::Uuid> {
+    let mut found: Option<uuid::Uuid> = None;
     for (tag_key, tag_values) in filter.generic_tags.iter() {
-        let key = tag_key.to_string();
-        if key == "h" {
-            for val in tag_values {
-                if let Ok(id) = val.parse::<uuid::Uuid>() {
-                    return Some(id);
+        if tag_key.to_string() != "h" {
+            continue;
+        }
+        for val in tag_values {
+            let Ok(id) = val.parse::<uuid::Uuid>() else {
+                continue;
+            };
+            match found {
+                Some(existing) if existing != id => {
+                    // Multiple distinct channel IDs — not a single-channel filter.
+                    return None;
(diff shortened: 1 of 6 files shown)


Chkhikvadze:
Independent production reproduction and validation:

- The deployed relay currently returns an empty result for the Desktop's one-filter/multi-value `#h` request, while one single-value `#h` filter per channel in the same request returns all six owner-visible workflow definitions. This confirms the compatibility failure and the fix shape in this PR.
- I rebased this patch locally onto current `main` and ran the affected package suites/checks.
- `just desktop-tauri-test`: 2,106 passed, 0 failed, 14 ignored.
- `just desktop-test`: 3,928 passed, 0 failed.
- `just desktop-check`, `just desktop-build`, `just desktop-tauri-fmt-check`, `cargo fmt --all -- --check`, `just desktop-tauri-clippy`, and `cargo clippy -p buzz-relay --all-targets -- -D warnings`: passed.
- `cargo test -p buzz-relay`: both new workflow regression tests passed; aggregate was 839 passed, 1 failed, 37 ignored. The sole failure was the unchanged `api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo` test (`504` vs expected `200`), outside this PR's changed files.

GitHub currently reports the PR mergeable; required review is the remaining repository gate.

Trevongit:
Closing in favor of origin **#6009** (`fix(workflows): preserve multi-channel listing semantics`), which already landed on `main` (2026-08-17).

This fork branch was the earlier attempt at the same job (listings across every accessible channel / multi-value `#h`). Replaying it on current `main` conflicts with that merge, so it should not be rebased.

Keeping the fork branch only as history; no further land-at-origin work here.

Trevongit:
Superseded by #6009.

Issue #3776, opened 2026-07-30
Cross-owner agent mentions are impossible: candidates filtered by local ownership, and eligibility reads a kind (10100) nothing publishes


rjgalvez:
**Describe the bug**

An agent owned by one person can never appear in another person's `@` mention autocomplete, in any channel, under any configuration. `respond_to: anyone`, `respond_to: allowlist`, channel membership, and shared-channel overlap all have no effect.

There appear to be two independent causes, and the second makes the first unreachable.

**Cause 1 — the eligibility gate is unreachable for agents you don't own**

`useMentions.ts` applies three gates in order (v0.5.1 `desktop/src/features/agents/hooks/useMentions.ts:245-262`):

```ts
if (isArchivedDiscovery(pubkey)) return;
if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) return;   // :249
if (shouldHideAgentFromMentions({…})) return;                                //  :252
```

`isAgentIdentityInManagedList` (`lib/agentAutocompleteEligibility.ts:57-65`) keeps a candidate only if `isAgent !== true` **or** its pubkey is in `managedAgentPubkeys`. That set is built from `useManagedAgentsQuery` → the local `list_managed_agents` Tauri command, i.e. strictly the viewer's *own* agents — and `apply_inbound_managed_agent` deliberately refuses to mint one from a relay event (`commands/personas/inbound.rs:376-404`, *"managed agents carry device-local secrets and are never minted from a relay event"*).

Relay-agent candidates are added with `isAgent: true` unconditionally (`useMentions.ts:334-346`), and member candidates get `isAgent: true` whenever `profile?.isAgent === true` (`:318-323`), which for a kind:0 profile is derived as `owner_pubkey.is_some()` (`nostr_convert.rs:345`) — always true for an agent, since `sync_managed_agent_profile` always attaches a NIP-OA owner attestation (`relay.rs:401-455`).

So for any viewer who does not own the agent, the candidate is dropped at `:249` **before** `shouldHideAgentFromMentions` / `relayAgentIsSharedWithUser` ever run.

Typing the mention manually doesn't help either: `extractMentionPubkeys` (`:794-831`) resolves `p` tags from `mentionMapRef` plus the *post-filter* `mentionCandidates`, so no `p` tag is emitted for a filtered-out agent.

**Cause 2 — `relayAgentIsSharedWithUser` reads an event kind nothing publishes**

[...]

PR #4453, opened 2026-08-03
fix(desktop): mention channel-member agents and invocable relay agents

nvisionlife:
## Problem

On a client that manages no agents locally, `@`-mention autocomplete shows no agents at all — not channel members, not relay-directory entries. Reported in #3671, #3125, #3776, and #4128.

`useMentions.ts` gated every candidate on the **locally-managed** agent set:

```ts
if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) return;
```

`managedAgentPubkeys` is derived solely from `managedAgentsQuery.data`. When a viewer manages nothing locally that set is empty, so every candidate with `isAgent === true` is dropped right there — both the channel-member candidates and the relay-directory candidates that `mentionCandidates` builds further down.

Everything written downstream to handle this — `mentionableAgentPubkeys`, `shouldHideAgentFromMentions`, `relayAgentIsSharedWithUser` — sits *after* that gate and therefore never runs. `shouldHideAgentFromMentions`'s "invocable => always show" branch is unreachable in the shipped build.

## Fix

Two changes, both in the first-pass gate:

1. `isAgentAutocompleteEligible` (renamed from `isAgentIdentityInManagedList`) also admits `candidate.isMember === true`, so agents that are members of the channel you are typing in pass.
2. `useMentions` passes `getMentionableAgentPubkeys(...)` — local managed ∪ relay agents shared with the viewer — instead of the local-only set, so an agent that is invocable but *not* a channel member also passes.

Both are needed for the full intended behavior. (1) fixes the reported channel-member symptom; without (2), every eligible relay-directory candidate that is not a channel member is still dropped before the relay-aware filter can admit it — those candidates are pushed with `isMember: false`, so clause (1) can never match them.

Security posture is unchanged: admission still runs through `relayAgentIsSharedWithUser`, so an agent that does not allowlist you never enters the set. The added test `relay agent that excludes the viewer stays hidden` pins that.

### `MembersSidebar` deliberately not widened

`MembersSidebar.tsx` keeps passing `managedAgentPubkeys`. Only the helper name changes there. Adding an agent to a channel should answer to channel membership policy, not to an agent's `respond_to` setting, and conflating the two would let mention semantics silently widen who can be added to a channel. The new `isMember === true` clause is inert at that call site — the sidebar already returns early for anyone in `memberPubkeys`.

## Tests

New cases in `agentAutocompleteEligibility.test.mjs`:

- `isAgentAutocompleteEligible: keeps people, channel members, and current managed agent identities` — the member clause.
- `relay agent allowlisting the viewer survives the composed mention gate` — composes the gate with `shouldHideAgentFromMentions` for a relay agent that is invocable but not a member, and pins the pre-fix behaviour (local-only set => dropped) so the regression cannot come back quietly.
- `relay agent that excludes the viewer stays hidden` — the negative case.

Verified on this branch:

- `pnpm test` — **3925 passed, 0 failed**
- `tsc --noEmit` — clean
- `biome check .` — no errors (2 pre-existing `useTemplate` infos in `personaCatalogRelay.test.mjs`, untouched)
- `check:file-sizes`, `check:px-text`, `check:pubkey-truncation` — pass

Note for whoever iterates on this: `useMentions.ts` sits right at the 1000-line ratchet, so the change had to be net-neutral. Dropping `managedAgentPubkeys` from the `mentionCandidates` dependency array is also required — after the swap it is no longer read inside that memo, and leaving it trips `lint/correctness/useExhaustiveDependencies`.

## Not covered here

A Playwright spec driving the autocomplete itself is not included. The unit tests pin the gate's behaviour, not the wiring from `useMentions` into the dropdown. The E2E bridge does have the pieces for it — `handleListRelayAgents` plus the `mock.relayAgents` and `mock.managedAgents` config seeds — so a spec that seeds one allowlisting relay agent, no managed agents, and asserts on the `@` menu looks reachable without new infrastructure. Happy to add it here if a reviewer wants it in the same PR.

## Credit

Supersedes #4357, which is the origin of the channel-member clause and its test. That work is carried here with authorship preserved via `Co-authored-by`, plus the relay-agent half of the gate and the composed-gate regression tests.

## Adjacent, not fixed here

`respond_to: "anyone"` is a trap for profiles published without `channel_ids`. `relayAgentIsSharedWithUser` requires `agent.channelIds.some((id) => sharedChannelIds.has(id))` for that route, and `agents_from_events` defaults `channel_ids` to `[]` when the profile omits it — so switching an agent from `allowlist` to the seemingly-broader `anyone` makes it *vanish* from other people's autocomplete. Happy to file that separately if it is not already tracked.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -5,7 +5,7 @@ import {
   coalesceAgentAutocompleteCandidates,
   getMentionableAgentPubkeys,
   getSharedChannelIds,
-  isAgentIdentityInManagedList,
+  isAgentAutocompleteEligible,
   relayAgentIsSharedWithUser,
   shouldHideAgentFromMentions,
 } from "./agentAutocompleteEligibility.ts";
@@ -136,25 +136,32 @@ test("getMentionableAgentPubkeys: keeps managed agents and shared relay agents",
   assert.deepEqual(result, new Set([PUB_A, PUB_B, PUB_C]));
 });
 
-test("isAgentIdentityInManagedList: keeps people and only current managed agent identities", () => {
+test("isAgentAutocompleteEligible: keeps people, channel members, and current managed agent identities", () => {
   const managedAgentPubkeys = new Set([PUB_A]);
 
   assert.equal(
-    isAgentIdentityInManagedList(
+    isAgentAutocompleteEligible(
       { isAgent: false, pubkey: PUB_B },
       managedAgentPubkeys,
     ),
     true,
   );
   assert.equal(
-    isAgentIdentityInManagedList(
+    isAgentAutocompleteEligible(
       { isAgent: true, pubkey: PUB_A.toUpperCase() },
       managedAgentPubkeys,
     ),
     true,
   );
   assert.equal(
-    isAgentIdentityInManagedList(
+    isAgentAutocompleteEligible(
+      { isAgent: true, isMember: true, pubkey: PUB_B },
+      managedAgentPubkeys,
+    ),
+    true,
+  );
+  assert.equal(
+    isAgentAutocompleteEligible(
       { isAgent: true, pubkey: PUB_B },
       managedAgentPubkeys,
     ),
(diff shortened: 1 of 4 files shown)


kevinsakiro:
This fixes an issue I am dealing with! Thank you!

ErnestoEverhard:
Thank you for putting this together.

avi-zloof:
We hit the same class of bug on our fork: channel-scoped `@` treated kind:10100 `channel_ids` as “is this agent in this channel”, which hid allowlisted agents that were live channel members.

Our patch (already on our product branch): [avi-zloof/buzz#9](https://github.com/avi-zloof/buzz/pull/9) — `relayAgentIsPresentInChannel` uses **live channel membership**, while kind:10100 stays the source for `respond_to` / allowlist rules. Wired through `useMentions.ts`.

Happy to open a follow-up PR against `main` if this thread wants that split rather than folding it into this change.

PR #4461, opened 2026-08-03
feat(cli): include display names in channel members

### LINKED ISSUE #4437: channels members omits display_name, so duplicate-name conditions are undetectable from that endpoint
## Summary

`buzz channels members --channel <uuid>` returns only `pubkey` and `role`:

```json
[{"pubkey":"<hex>","role":"owner"},{"pubkey":"<hex>","role":"bot"}, ...]
```

There is no `display_name`. Since mention resolution keys off display names of channel members (`crates/buzz-relay/src/workflow_sink.rs` pairs `get_members` with `get_users_bulk` precisely to obtain them), any tooling that needs to reason about names in a channel has to issue an N+1 lookup through `users get` per member.

## Impact

The relay already joins membership to `display_name` internally for mention routing. Callers cannot see the same view.

Concretely: I wrote a maintenance script to detect and remove duplicate-display-name members from channels, matching names against the `channels members` payload. It reported "0 orphans" on a deployment that had **22 orphan memberships across 6 channels**, because the field it was matching on does not exist. The script was a silent no-op, which is worse than not having one — it reads as confirmation that nothing is wrong.

Anything auditing "who is in this channel and what are they called" hits the same wall.

## Suggested fix

Include `display_name` (nullable) on each member record returned by `channels members`. The relay already has it in `get_users_bulk` for this exact set of pubkeys, so this is a projection change rather than a new query.

If payload size is a concern, an opt-in flag (`--with-names`) would work equally well for tooling.

## Related

See #4303 for the duplicate-identity condition this tooling exists to clean up.

kiranmagic7:
## Summary
`buzz channels members` serialized the channel membership tags directly, so each record contained only `pubkey` and `role`. This change loads member profiles with one batched kind:0 filter and adds a nullable `display_name` to each record while preserving membership order and roles. The profile projection uses the relay's existing `name` fallback; missing or malformed profiles produce `null`.

### Related issue
Fixes #4437. No duplicate implementation was found; the closest prior CLI mention work is #3330, which does not change `channels members` output.

### Testing
- `cargo test -p buzz-cli channel_member -- --nocapture`
- `cargo test -p buzz-cli`
- `cargo clippy -p buzz-cli --all-targets --all-features -- -D warnings`
- `just ci`

All checks passed. No screenshots are needed for this CLI JSON change.


Chessing234:
looking at this for #4437 — if you're joining display names via users bulk on the same pubkeys from the membership event, that matches how the relay does mention routing. please keep display_name nullable when a profile is missing so tooling can tell \"no name\" from \"not returned\".

kiranmagic7:
Thanks, that is the contract in this branch now. `display_name` is added as JSON `null` when no profile/name is available, so callers can distinguish a missing profile from a returned empty string. The regression `channel_members_include_nullable_profile_display_names` covers that nullable path and I will keep it unchanged.

kiranmagic7:
The nullable `display_name` contract is unchanged. I merged current `main` into the branch, resolved the `channels.rs` conflict, and reran the two focused channel-member tests, all 345 `buzz-cli` tests, and `just ci` successfully on `e70a995cd`.

kiranmagic7:
The canvas changes in #6780 just landed and rewrote `channels.rs`, so I rechecked the overlap against current `main` (`b6a26556e`). This branch still merges cleanly: those changes begin after `cmd_list_channel_members`, and the nullable `display_name` contract plus focused tests are untouched. The remaining blockers are review and Block-side workflow/security authorization for this exact head.

Issue #4154, opened 2026-08-01
buzz-acp: structured MCP server configuration with explicit credential scoping


wolfyy970:
## Problem

`buzz-acp` supports one MCP command through `BUZZ_ACP_MCP_COMMAND`. That path does not represent multiple servers with stable names, arguments, and separate environments.

The existing builder also treats configured servers as the privileged Buzz companion and copies `BUZZ_PRIVATE_KEY` and `BUZZ_AUTH_TAG` into each entry. Third-party MCP servers should not receive those credentials.

MCP environment values can also appear in ACP wire logs and observer events.

## Proposed change

- add `--mcp-config <path>` and `BUZZ_ACP_MCP_CONFIG` for a strict, versioned JSON config
- preserve server order, names, commands, arguments, and environment values without shell parsing
- keep `BUZZ_ACP_MCP_COMMAND` unchanged for the privileged Buzz companion
- reject invalid configuration before the harness starts listening
- redact MCP environment values from Buzz-managed diagnostics

The adapter will continue to inherit the harness environment. This scopes credentials in the ACP request but does not provide process isolation.

## Acceptance criteria

- multiple structured servers reach `session/new` in declaration order with their exact configuration
- the legacy companion remains unchanged and is first when both inputs are present
- structured servers do not receive the companion's Buzz credentials
- invalid, duplicate, oversized, or over-limit configuration fails before startup
- paths and arguments containing spaces, commas, quotes, Unicode, and shell metacharacters remain literal
- MCP environment values do not appear in wire logs or observer events
- restart and repeated-session behavior is covered by tests

Related: #2883, #2899, #2900, #3196, #3417

wolfyy970:
Implementation is in #4164. Focused tests and the full repository CI suite pass.

wolfyy970:
Closing. The remaining slice is #5349.

PR #3600, opened 2026-07-29
fix(desktop): honour CLAUDE_CONFIG_DIR in agent MCP config panel

Chessing234:
## Summary
- Resolve Claude settings + MCP config from the agent’s `CLAUDE_CONFIG_DIR` when set
- Default remains `~/.claude/settings.json` + `~/.claude.json`
- Panel “From config file (…)” attribution points at the isolated `.claude.json`

## Why
Isolated Claude agents correctly load an empty MCP set at runtime, but the config panel always listed the operator’s `~/.claude.json` servers and labeled them enabled (#3493).

## Test plan
- [ ] Agent with `CLAUDE_CONFIG_DIR=<nest>/claude-config` and empty `.claude.json` — panel MCP list is empty and path shows the nest dir
- [ ] Agent without the env var — panel still shows `~/.claude.json`
- [x] Unit tests for isolated vs default path resolution

Fixes #3493

Made with [Cursor](https://cursor.com)

PR #4480, opened 2026-08-03
feat(acp): live per-channel MCP control via session/resume

Monivancan:
## Summary

Buzz agents get their MCP tool set once, at `session/new`, and it is frozen for the life of the session. This adds **live per-channel MCP control**: the set can change mid-conversation via `session/resume`, on the same session id, with the transcript intact.

### Related issue

Closest existing work — **there is overlap and I want to name it up front rather than have a reviewer find it:**

| # | Overlaps with | How this differs |
|---|---|---|
| [#2899](https://github.com/block/buzz/issues/2899) | The motivating issue — one MCP server per agent forces a choice between the shell and any domain server | This PR fixes it *and* makes the set changeable at runtime |
| [#4154](https://github.com/block/buzz/issues/4154) | Structured MCP config with credential scoping | Same direction; this PR's `BUZZ_ACP_MCP_SERVERS` is a subset |
| [#3196](https://github.com/block/buzz/pull/3196), [#2900](https://github.com/block/buzz/pull/2900), [#4164](https://github.com/block/buzz/pull/4164), [#2740](https://github.com/block/buzz/pull/2740) | **Direct duplicates of this PR's Task 2** (multi-server / structured / arg-carrying MCP config) | Static spawn-time config only. **If any of these lands first I will drop my `config.rs` commit and rebase the live-control work onto it** — the resume-swap does not depend on how the set is spelled |
| [#4271](https://github.com/block/buzz/pull/4271) | Remote HTTP MCP servers | Complementary. This PR is stdio-only by design; #4271's `McpServer` enum is the natural base for granting remote servers live |
| [#4078](https://github.com/block/buzz/pull/4078), [#3600](https://github.com/block/buzz/pull/3600), [#3502](https://github.com/block/buzz/pull/3502), [#3493](https://github.com/block/buzz/issues/3493) | Desktop agent MCP config panel | That panel is read-only display. This PR is the backend a live toggle would drive; the UI is a separate PR |

**The genuinely new part is the resume-swap** (`pool.rs::reconcile_channel_mcp`) — none of the above changes an MCP set on a *live* session. Everything else here is scaffolding for it, and the Task 2 scaffolding is negotiable.

---

## Problem

Buzz agents receive their MCP tool set once, at `session/new`, and it cannot change for the life of the session. Two consequences, both measured on a live Buzz install:

1. **No mid-conversation grants.** A user cannot hand an agent a tool it lacks ("I need Razorpay now") without rotating the session — which discards the agent's in-context working memory and restarts it from a truncated slice of channel history.
2. **Token waste.** Buzz passes a near-empty `mcpServers` list by default (`crates/buzz-acp/src/acp.rs:629-655`, `lib.rs:4179`), so agents fall back to their own global config. On a live session whose `~/.claude.json` holds 13 MCP servers, that is **~64k tokens of unused tool schemas on every turn** — 255k cumulative across an 8-message channel.

## Mechanism

`session/resume` carries an `mcpServers` field, and both flagship adapters implement "resume with a changed MCP set" as a supported reconfiguration path:

- **claude-agent-acp** fingerprints `(cwd, mcpServers)`; on mismatch it tears down the SDK subprocess and recreates it with `resume: <same sessionId>` and the new servers, restoring the full transcript from disk. `acp-agent.js:132-139` (fingerprint), `:3981-4008` (teardown + `createSession({resume})`), `:4279-4282` (`options.sessionId` set only when *not* resuming). The in-code comment names this exact case.
- **codex-acp** `resumeSession` → `threadResume({config: createSessionConfig(cwd, dirs, request.mcpServers), threadId: sessionId})` — `dist/index.js:26252-26259`.
- Both advertise the capability: `acp-agent.js:644`, codex `dist/index.js:28490-28491` (`sessionCapabilities.resume: {}`).

The session ID never changes, so Buzz's `channel → session` bookkeeping is untouched and continuity is the entire transcript, not Buzz's `context_limit` window.

**`session/resume`, not `session/load`:** `load` replays full history as `session/update` notifications before responding, risking the 60s request timeout on long conversations. `resume` skips the replay with identical reconfiguration semantics.

> ### ⚠️ This is adapter behaviour, not an ACP contract
>
> Nothing in the ACP schema promises that `session/resume` applies a changed tool set. Both shipped adapters do it today, and the Claude adapter's own comment names the case — but an adapter update could stop honouring it without violating the spec. This PR is written defensively around that: the harness never treats a successful resume as proof (see *Verification* below), and the design records explicit abandon triggers if the behaviour regresses. **Reviewers should weigh this as the main risk of the approach.**

## What's in this PR

All additive, all inside `crates/buzz-acp`, no `unsafe`, DCO signed:

| File | Change |
|---|---|
| `acp.rs` | `session_resume(session_id, cwd, mcp_servers)`; `initialize()` records `agentCapabilities.sessionCapabilities.resume` into a `resume_supported` flag, mirroring the existing `steering_supported` pattern; opt-in `strictMcpConfig` on `session_new_full` |
| `config.rs` | `--mcp-servers-json` / `BUZZ_ACP_MCP_SERVERS` (JSON array) alongside the legacy single `mcp_command`; malformed JSON is ignored rather than dropping the dev-MCP server |
| `pool.rs` | `desired_mcp` / `applied_mcp` / `mcp_verified` per channel on `SessionState`; `reconcile_channel_mcp()` resume-swaps at the prompt-time session lookup; `effective_mcp_servers()` prefers the channel's desired set; `note_tool_call_for_verification()` observes whether a granted server's tools actually appear |
| `lib.rs` | `update_mcp_servers` control-frame arm + `handle_update_mcp_servers_control()`, modeled on `handle_switch_model_control()` |
| `Justfile`, `scripts/run-tests.sh` | run the `buzz-acp` suite in `test-unit` — it was built by CI but never tested (see *Test status*) |
| `config.rs` (tests) | make the pre-existing `lazy_pool_defaults_off` hermetic, so the new `test-unit` entry is not red on any machine that runs a Buzz agent |

### Behaviour guarantees

- **An MCP change never cancels an in-flight turn.** Unlike `switch_model`, a busy agent marks the channel dirty and the swap applies at the next turn boundary. Killing work to add a tool is a worse trade. Safe by construction: the agent is moved out of the pool during a turn, so the dirty flag is only consulted while idle.
- **Identical sets do not trigger a resume.** The Codex adapter has no change-detection and would respawn unconditionally; the Claude fingerprint is order-sensitive on `args`/`env`. Suppressing no-op grants avoids a spurious subprocess teardown.
- **Agents without `sessionCapabilities.resume`** fall back to today's invalidate-and-rotate path rather than silently doing nothing.
- **A failed `session/resume` never fails the turn** — it rotates the session instead. Continuity is lost; the turn is not.

> ### `strictMcpConfig` is opt-in per managed channel
>
> `_meta.claudeCode.options.strictMcpConfig` is sent **only** for channels with an explicitly managed MCP set. Unmanaged channel = today's behaviour (the agent's own global config, untouched). Managed channel = exactly the configured set, and nothing else.
>
> Defaulting this on would silently strip the global MCP servers of every user who relies on them inside Buzz. It stays opt-in for that reason, and the token saving only lands on channels that asked for it.
>
> Caveat worth knowing: the adapter never names `strictMcpConfig` — it rides the generic `...userProvidedOptions` spread (`acp-agent.js:4103`, `:4157`), so a typo in the key would silently no-op rather than error.

### Remote MCP servers are not modelled — stdio only

`McpServer` (`crates/buzz-acp/src/acp.rs:28`) is `{ name, command, args, env }`. The ACP schema's `mcpServers` is a union covering stdio **and** http/sse, and the Claude adapter advertises `mcpCapabilities { http, sse }` — but the Rust type here models only the stdio variant, and this PR does not change that.

Consequence: **remote MCP servers cannot be granted through this path.** That includes the design's own motivating example (a hosted Razorpay MCP). Extending `McpServer` to an enum over stdio/http/sse is a separable, purely additive follow-up; it was deliberately deferred to keep this change reviewable.

### Post-grant verification is weaker than the design requires

The design (§5.4) requires the harness to *verify the grant landed* rather than trust the resume's success response. **That surface does not exist on the ACP wire today:**

- `acp-agent.js:1601-1614` consumes the SDK's `system`/`init` message internally (latching `msgLifecycleV1`, syncing fast-mode state) and never forwards `tools[]` or `mcp_servers[{name, status}]` to the ACP client — even though the SDK carries both (`sdk.d.ts:4418-4424`).
- `session/resume` returns only `{sessionId, modes, configOptions}`.

So there is nothing authoritative to observe. What this PR does instead is **opportunistic**: a grant is recorded as applied-but-unverified, and promoted to verified only when a later `session/update` carries a tool call whose name matches the `mcp__<server>__` prefix of a granted server. A grant that silently failed therefore stays unverified forever rather than showing a false green.

**Named follow-up:** the correct fix is upstream — a PR against `@agentclientprotocol/claude-agent-acp` to surface the SDK's `mcp_servers[{name, status}]` on the ACP wire (and ideally in the `session/resume` response). Once that exists, the harness can assert the grant directly instead of inferring it from tool traffic. `applied_unverified` is a workaround for a missing surface, not a permanent design. Blocking this PR on another repo's review was considered and rejected.

### Control results are thinner than the design's data flow

Only three statuses reach the observer, all emitted at control-frame receipt in `lib.rs`: `invalid_servers`, `unchanged`, `pending_next_turn`.

The *outcome* of the swap is not reported. `reconcile_channel_mcp()` (`pool.rs`) logs via `tracing` and mutates state, but emits no `control_result` — so "applied", "rotated because the agent has no resume support", and "resume failed, rotated" are invisible to a UI. Likewise `mcp_verified` is maintained and unit-tested but has no consumer: nothing reads it to emit a verified status.

This is enough for a correct backend and **not** enough for the honest toggle UI the design describes. Wiring those emissions belongs with the desktop PR that will consume them; flagging it here so the gap is not mistaken for coverage.

## Not in this PR

- **Desktop UI** — the per-agent MCP panel at spawn, the per-channel live toggle list, `managed-agents.json` → `BUZZ_ACP_MCP_SERVERS` at spawn, and `agentControl.ts`. Separate plan, separate PR. **Without it the control frame has no in-app sender**, so this PR is infrastructure ahead of its UI.
- **Mid-*turn* tool injection.** Tools land at turn boundaries.
- **A meta-MCP gateway** that hot-mounts servers via `notifications/tools/list_changed`. Designed to compose on top of this — the gateway would simply be one of the resume-swapped servers — and deferred.
- **ACP protocol changes.** None required.

## Test status

Honest accounting, because parts of it are red or unrun.

**New tests in this PR — 18, following the crate's existing shapes:**

| Area | Tests |
|---|---|
| `acp.rs` | `session/resume` request shape; `initialize` records resume supported / unsupported; `strictMcpConfig` present when managed, absent when unmanaged; an MCP change resumes the same session without a new one |
| `lib.rs` | control frame records the desired set; rejects a malformed server list; ignores a bad channel id |
| `config.rs` | JSON servers append; dev-MCP survives alongside them; malformed JSON does not drop dev-MCP |
| `pool.rs` | `invalidate_channel` clears applied state; desired set is per-channel; `effective_mcp_servers` prefers it; identical sets do not resume; a granted server's tool call marks the grant verified; an unrelated tool call does not |

**`just ci` on this branch: exit 101, phase 6 of 8.**

- Phases 1–5 green (`check` → `fmt-check`, `clippy`, `desktop-check`, `desktop-tauri-fmt-check`, `desktop-tauri-clippy`, `web-check`, `mobile-check`; then `test-unit`, `desktop-test`, `desktop-build`, `desktop-tauri-check`).
- Phase 6 `desktop-tauri-test`: **2002 passed, 1 failed.** The failure is `relay_admission::tests::concurrent_429_extends_the_window_for_parked_waiters` (`desktop/src-tauri/src/relay_admission.rs:165`, `left: 300.001s, right: 5s`) — **pre-existing and unrelated to this branch:**
  - `git diff --name-only origin/main..HEAD` touches zero files under `desktop/`.
  - That file was last modified by `581baa6` (Will Pfleger, 2026-07-29), before this branch existed.
  - The 13 `relay_admission` tests pass in isolation, and a full re-run of the tauri lib suite passed 2003/0.
  - Cause: `MAX_HINT_SECONDS = 300` (`relay_admission.rs:41`); two sibling tests arm exactly a 300s window on a process-global gate, and `TEST_SERIAL` does not serialise them under the parallel runner. Shared-global test-isolation defect, worth fixing upstream independently of this PR.
- Phases 7–8 (`web-build`, `mobile-test`) **did not run** — the build filled the disk. Not yet verified either way on this branch.

**`buzz-acp`: 676 tests run, 676 passed.** All 18 new tests included.

**Until this PR, none of them ran in CI.** `just test-unit` covered `buzz-core`, `buzz-auth`, `buzz-cli`, `buzz-db --lib`, `buzz-conformance`, `buzz-push-gateway` — and the `scripts/run-tests.sh` fallback the same set. `buzz-acp` reached `just ci` only through `cargo check`/`clippy`, `.github/workflows/ci.yml:130-131` runs that same `just test-unit` for the Unit Tests job, and every other workflow reference is `cargo build -p buzz-acp`. The crate's whole suite compiled on every PR and executed on none of them, so a green check implied coverage that did not exist. Adding the 18 tests without fixing that would have been ceremony.

Fixed here, in both branches of the recipe so they stay in step. This is the one part of the PR that touches shared tooling; it is two comment blocks and two lines.

**One pre-existing test had to become hermetic to make that safe.** `config::tests::lazy_pool_defaults_off` asserted a default by *parsing*: `assert!(!CliArgs::parse_from([...]).lazy_pool)`. But `lazy_pool` is `#[arg(long, env = "BUZZ_ACP_LAZY_POOL", ...)]`, and every running buzz-acp agent exports that variable to its children — so the test read the ambient environment and failed for anyone running the suite from inside a Buzz agent session, which is how this repo is normally developed. CI never saw it (no such variable on a runner); a maintainer dogfooding Buzz would have hit it the moment `buzz-acp` joined `test-unit`. It now asserts the arg's declared default via `CommandFactory` and touches no environment. Verified hermetic against `BUZZ_ACP_LAZY_POOL` set to `true`, `false`, `1`, `banana`, and empty, and mutation-checked: flipping `default_value_t` to `true` fails it with `left: ["true"], right: ["false"]`.

**Follow-up, not fixed here:** that arg's env binding is value-parsed with `[possible values: true, false]`, so `BUZZ_ACP_LAZY_POOL=1` or `=0` — plausible operator spellings — are a hard startup error (`invalid value '1' for '--lazy-pool'`, exit 2) rather than a boolean. The message is clear enough that this is a wart, not a hazard, but it applies to every env-bound `bool` arg in `config.rs`, not just this one.

**The live path is not proven end-to-end.** A real mid-conversation toggle on a running Buzz agent has not been exercised: it needs the owner's key to sign the control frame and the desktop UI to send it. Everything above is unit-level and source-level. Treat "grant a tool mid-chat and keep the conversation" as *designed and unit-tested*, not *demonstrated*.

## Continuity ledger

**Preserved across a toggle:** the complete conversation (every message, tool call, tool result, file content, rebuilt from the session transcript), the session's model and mode, todo state, permission settings, and the sessionId itself.

**Lost:** subprocess ephemera — background bash jobs die at teardown, persistent-shell cwd/env resets, stateful MCP servers restart. Cost is one respawn (~2–5s, in the idle path) plus a possible prompt-cache rewrite on the next turn: a token and latency cost, not a memory loss.

## Security

`update_mcp_servers` names a command to execute — it is remote code execution by design. It stays inside the existing owner-signed, encrypted, ±5-minute-freshness control-frame envelope, the same path as `switch_model` (`crates/buzz-acp/src/lib.rs:834-1005`). When the desktop UI lands it must show **what binary or URL** a grant will run; a friendly name pointing at an arbitrary command is a supply-chain phish.

## References

- Design: `docs/superpowers/specs/2026-07-31-live-mcp-control-design.md`
- Implementation plan: `docs/superpowers/plans/2026-07-31-live-mcp-control-buzz-acp.md`

---

## Pre-open checklist (not part of the PR body)

- [ ] Free disk, then run phases 7–8 (`just web-build`, `just mobile-test`)
- [x] `buzz-acp` suite runs and passes — 676/676, and now inside `just test-unit`
- [x] `test-unit` entry bundled here, with the hermetic-test fix that makes it safe
- [ ] `git fetch --unshallow upstream` — this clone is `--depth 50`, so the merge-base against `block/buzz` is incomplete
- [ ] `git push origin feat/live-mcp-control-impl:feat/live-mcp-control` — plain fast-forward; origin already has that branch at `961d360`, an ancestor of HEAD
- [x] DCO: 15/15 commits signed off, sign-off email matches author on each
- [ ] Moni's go-ahead to open against `block/buzz`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -165,6 +165,26 @@ check existing reply handlers for the pattern.
 by the ACP harness into managed agent subprocesses. In development, set
 `BUZZ_PRIVATE_KEY` and `BUZZ_RELAY_URL` in your environment manually.
 
+### Harness MCP configuration (`BUZZ_ACP_MCP_SERVERS`)
+
+`buzz-acp` derives one MCP server from `--mcp-command` (the dev MCP).
+`BUZZ_ACP_MCP_SERVERS` adds more, as a JSON array of
+`{name, command, args, env:[{name,value}]}` objects merged with that one:
+
+```bash
+export BUZZ_ACP_MCP_SERVERS='[{"name":"razorpay","command":"/usr/bin/rzp","args":["--stdio"],"env":[]}]'
+```
+
+Malformed JSON is logged and ignored — a bad value degrades to the dev MCP
+alone rather than taking the agent's tools down with it.
+
+A channel's server set can also be changed at runtime via the owner-signed
+`update_mcp_servers` observer control frame. That applies at the channel's
+next turn boundary through `session/resume`: same session ID, transcript
+preserved, never cancelling an in-flight turn. Channels managed this way also
+get `strictMcpConfig`, so they run exactly the listed servers; unmanaged
+channels keep the agent's own global MCP config untouched.
+
 ### Building the CLI
 
 ```bash
--- a/EXECUTE.md
+++ b/EXECUTE.md
@@ -1,0 +1,123 @@
+# Execute prompt — live per-channel MCP control
+
+Paste the block below into a fresh agent session (Buzz channel agent, or Claude Code started in `~/Dev/buzz`).
+
+**Branch note:** this prompt puts the agent on its own branch so it can run alongside/against another implementation without collisions. Change the branch name in the block if you want it somewhere else.
+
+## If you're running this through Buzz — read this first
+
+Three facts about this machine's Buzz setup, verified 2026-07-31:
+
+1. **Every Buzz agent has `cwd = /Users/neeyafit/.buzz`.** The harness passes
+   `std::env::current_dir()` as the ACP session cwd (`crates/buzz-acp/src/lib.rs:1546`) and there is
+   no per-agent working-directory setting in `managed-agents.json`. The agent is *not* in the repo
+   and will not auto-load the repo's `AGENTS.md`. The prompt below handles this — don't remove the
+   `cd` instruction.
+2. **`turn_timeout_seconds: 320`** (~5 min/turn). Do not ask for all 8 tasks in one message. Drive
+   it **one task per turn**; a cold `cargo build -p buzz-acp` can consume most of a turn on its own.
+3. **Use a fresh channel or DM, and the Opus-5 agent.** Buzz seeds a new session with a small window
+   of recent channel history, so starting in a busy channel feeds unrelated chatter into the session.
+   Opus-5 is `opus[1m]` — the 1M context matters for a ~1400-line plan across 8 tasks. Avoid
+   Grok/Cursor here (different runtime).
+
+---
+
+We're implementing a feature for contribution back to `block/buzz` (upstream). The repo is at
+`~/Dev/buzz`. Remotes: `origin` = Monivancan/buzz (my fork), `upstream` = block/buzz.
+
+**First, three things to read, in this order:**
+
+1. `docs/superpowers/plans/2026-07-31-live-mcp-control-buzz-acp.md` — the approved implementation
+   plan. This is your instruction set: 8 tasks, each with exact files, real code, and a TDD step
+   sequence. Follow it task by task, in order.
+2. `docs/superpowers/specs/2026-07-31-live-mcp-control-design.md` — the design the plan came from.
+   Read it for the *why*; the plan wins wherever they differ (the plan was written after
+   re-verifying the spec against shipped source and records three corrections).
+3. `AGENTS.md` — repo conventions and quality gates.
+
+**The feature, in one line:** let a user grant or revoke an MCP server on a live Buzz channel and
+have the agent continue the *same* conversation with the new tool set — via `session/resume` with
+an unchanged sessionId and a changed `mcpServers` list. Plus multi-server config at spawn, and
+opt-in `strictMcpConfig` to stop injecting ~64k tokens/turn of unused global tool schemas.
+
+**Your working directory is NOT the repo.** You start in `/Users/neeyafit/.buzz`. The repo is at
+`/Users/neeyafit/Dev/buzz`. Shell cwd does not persist between tool calls, so chain `cd` into every
+command (`cd /Users/neeyafit/Dev/buzz && ...`) and use absolute paths for file reads and edits.
+
+**Set up first:**
+
+```bash
+cd /Users/neeyafit/Dev/buzz && . ./bin/activate-hermit && \
+  git status --short && \
+  git checkout feat/live-mcp-control && \
+  git checkout -b feat/live-mcp-control-impl
+```
+
+The working tree must be clean apart from `KICKOFF.md` / `EXECUTE.md`. If it isn't, stop and report
+what's there rather than committing someone else's work.
+
+`. ./bin/activate-hermit` is required before **every** git, cargo, or just command — the repo's
+hooks and toolchain live in Hermit, and an unconfigured `PATH` makes them fail in confusing ways.
+Do not rewrite hook commands to work around a broken PATH; fix the PATH.
+
+Work directly in `~/Dev/buzz` on the branch above — **not** in a `git worktree`. The pre-commit
+hook runs `just desktop-tauri-fmt`, which fails inside worktrees and will block every commit you
+try to make (AGENTS.md, Common Gotchas #6).
+
+**Only one agent may edit this checkout at a time.** If someone else is already working in
+`~/Dev/buzz`, stop and say so rather than racing them.
+
+**Scope:** Tasks 1–8 of the plan, all inside `crates/buzz-acp`. The desktop UI is explicitly NOT in
+scope — it is a separate plan. Do not touch `desktop/`.
+
+**Do ONE task per turn.** Complete Task 1, commit it, then stop and report. Wait to be told to
+continue. Your turn is capped at 320 seconds — attempting several tasks in one turn will get you
+killed mid-edit and leave the tree in a broken state.
+
+**Non-negotiables:**
+- TDD, exactly as the plan lays it out: write the failing test, RUN it and see it fail, implement,
+  run it green, commit. Do not batch the tests to the end. Do not skip the "watch it fail" step —
+  a test that never failed proves nothing.
+- `git commit -s` on **every** commit. CI's DCO check fails the PR without a `Signed-off-by`
+  trailer. Verify with `git log -1 | grep Signed-off-by` after your first commit.
+- No `unsafe`. No new `unwrap()`/`expect()` in production paths — use `?`.
+- Additive only; mirror the existing in-crate patterns the plan names (`session_new_full`,
+  `handle_switch_model_control`, `steering_supported`).
+- MCP changes apply at **turn boundaries** and must NEVER cancel an in-flight turn. There is
+  deliberately no busy-path oneshot, unlike `switch_model`.
+- `strictMcpConfig` stays **opt-in per managed channel**. Defaulting it on silently strips users'
+  global MCP servers.
+- Run `cargo test -p buzz-acp` after every task. Run `just ci` before you call the work done —
+  clippy passing does not mean fmt passes.
+
+**Verify, don't trust.** The plan cites shipped adapter source by file:line (e.g.
+`acp-agent.js:3981-4008`). Those were verified once, but re-check any line you're about to build
+on — the adapter is a versioned npm package under
+`~/Library/Application Support/Buzz/node-tools/lib/node_modules/@agentclientprotocol/`, and it can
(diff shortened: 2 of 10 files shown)


wolfyy970:
@Monivancan, the session/resume work is complementary. #4164 already adds versioned multi-server config, and #4271 is aligning HTTP with it. `BUZZ_ACP_MCP_SERVERS` would create another JSON format. I think live updates should consume the same transport-tagged document, with malformed config rejected rather than silently ignored.

PR #4485, opened 2026-08-03
feat(desktop): workspace-scoped agent definition store

wpfleger96:
This PR partitions the agent definition store by workspace, closing a two-directional leak where definitions (managed agents, teams, global config) and their runtime connections were shared across relays.

Previously a single unscoped store lived at `agents/` and every `apply_workspace` event-synced all records to the newly selected relay, while the runtime reconciler fanned agents into every configured community. After an identity switch, in-flight old-owner inbound events could write into the new owner's store.

## What this PR does

- Add `WorkspaceAgentScope { scope_id, relay_url, owner_pubkey, definitions_dir, generation }` as the single scope authority; `scope_id` is byte-identical to the retention DB's sha256 derivation via a shared helper
- Implement a four-stage workspace transition (prepare → drain → commit → post-commit) with a drain journal, lock-owning compensation on stop failure, an infallible commit critical section, and degraded-result reporting through `applyCommunity()` and `useCommunityInit.ts`
- Add a universal staged scope-initialization state machine (`AdoptedLegacy | LegacyClaimedByOther | FreshNoLegacy` manifest, atomic rename, versioned `_ready` marker v1) that uses the existing `retention.db` claim as the canonical ownership ledger so retention and definition adoption can never diverge
- Version the `_ready` marker: an unversioned or v0 marker forces re-run through `run_pre_ready_family` (retention migration + persona backfill) before advancing to v1, so scopes created by earlier defective pipeline iterations are repaired on next activation. Marker written via temp+rename (atomic)
- Extend `scope_for_arrival` to match `(relay, owner)` so in-flight old-owner inbound events cannot land in the new owner's store after an identity switch
- Option A for Mesh: workspace switches and identity imports fail closed if a client-mode Mesh runtime is active, with a clear error telling the user to stop Mesh first. UI adds a "Stop using shared compute" button when a client session is active. Client start acquires `workspace_transition` through runtime installation so switches cannot race a concurrent client start. Serve-mode runtimes are machine-level and never block a switch. The journaled Mesh recipe is tracked as a follow-up below.
- Stop and journal all old-scope managed runtimes during drain; restore new-scope `start_on_app_launch` agents on every activation; enforce relay-match reuse on `ensure_relay_mesh_for_record` with a fail-closed preflight error when a serve-mode runtime belongs to another relay
- Remove the `communities` parameter from `reconcile_managed_agent_runtimes`; both frontend callers now invoke a parameterless command whose relay is derived server-side from the active scope, making cross-scope fan-out unrepresentable
- Global-config restart (`set_global_agent_config`) captures scope at entry; Phase 1 validates generation inside the store lock before writing; Phase 2 per-agent restart validates under lock before stop — all I/O targets the captured `definitions_dir` via `start_local_agent_pairs_with_preflight_at`
- Snapshot imports (`confirm_agent_snapshot_import`, `confirm_team_snapshot_import`) capture owner keys at entry, verify pubkey against captured scope, and thread them through all mint/retention/engram phases. Re-verify under store lock before Phase 3a write. All outbound profile/memory publication uses `captured_scope.relay_url`
- Mesh recovery error persistence validates scope generation inside the store lock before each write
- Bind every direct owner-key artifact command to an admitted egress generation before it reads or clones the key; the closed artifact-command ordering scan and lease/drain schedule prevent an A-derived artifact from being stamped as B-current during identity transition
- Serialize NIP-49 backup creation on `identity_mutation` before egress admission, eliminating the inverse lease-versus-mutation deadlock; a mock-runtime schedule proves a backup queued before the drain holds no lease until transition exit
- Lock-owning compensation: `compensate_drain` takes the caller's already-held `managed_agent_runtime_transition` guard by value, re-acquires only the store lock, validates captured scope generation, then restores journal entries via `start_pair_under_held_locks` — closes the drop-then-compensate interleave window without recursive locking or an AtomicBool gate
- Pre-scope migrations moved into the scoped pipeline: `migrate_persona_provider_to_runtime` runs as step 0 of `run_scoped_migrations` (returns `Result`, propagates errors); `migrate_agent_keys_to_dev_service` runs in `run_pre_ready_family` (debug non-test builds only, returns `Result`). Pre-scope calls in `run_boot_migrations_inner` removed.
- `workspace-degraded` Tauri event wired to `useNestNotifications.ts` (toast.error with cleanup + behavioral test); restore spawn emits the event on failure; `spawn_event_sync` return type corrected to `()` (fire-and-forget; dispatch failure during runtime shutdown has no toast surface)
- Add unit tests: generation staleness, drain-journal compensation contract, live-process SIGKILL drain, deterministic partial-drain stop-failure with injected error, versioned-ready upgrade path, snapshot captured-relay/owner-key, scope-initialization crash boundaries, Mesh relay-match rules, identity-import modes

## Deferred follow-up (tracked here, no separate issue)

**Journaled Mesh recipe**: client-mode Mesh runtimes have no persisted restart recipe (restore only knows Serve mode). The correct end-state is to include the Mesh client in the drain journal with ownership and restart data, coordinate drain via `rearm_lock → mesh_llm_runtime` in the drain stage after prepare, and compensate synchronously on failure. This requires building the restart recipe for consumer-mode clients first. Option A (fail-closed switch while a client runtime is active, with user-accessible stop command) is the interim behavior for this PR.


Chessing234:
workspace-scoped agents/stores is the right cut for the cross-relay leak. the two-owner e2e note in the test comment helped me follow the nip-33 half.

wolfyy970:
The Desktop failure is unrelated to this branch. The compiled-flag job reran the full Tauri suite and generated_passphrase_respects_word_count_and_separator drew a hyphenated EFF word, so a three-word phrase produced four parts when the test split on hyphens. The other 2,448 tests passed.

A rerun should clear this check. The flaky assertion belongs in a separate small fix rather than another change to this already large branch.

Issue #4301, opened 2026-08-02
Portable agent templates with Project connections and safe updates


wolfyy970:
@wpfleger96 @tlongwell-block, I have the working pieces locally and would like to align the remaining boundaries with Buzz before sending more PRs.

Buzz already has agent identities, personas, managed agents and snapshots. What is missing is one coherent path for creating a reusable agent, installing it as a distinct identity, giving it Project-owned tools, and updating it safely.

Today, moving a useful agent to another Project still means rebuilding its instructions, Skills and tool setup by hand. Updates also lack a clear authority boundary: a user may mean to change one agent, but the same edit can affect the reusable definition behind several agents.

I do not think this needs a new top-level object or product name. It can extend the concepts Buzz already has:

- An **agent template** holds reusable instructions, pinned Skills, model defaults and logical tool requirements.
- An **agent** has its own key, home, workspace and private settings.
- A **Project connection** holds the concrete MCP command or endpoint and its credentials. Credentials are never part of the portable template.
- An immutable **version** makes template changes reviewable and reversible.
- **Agent Snapshot** is the natural portable boundary for import and export.

## Product flow

Editing an agent should offer two explicit actions:

- **Edit this agent** changes only that agent.
- **Edit its template** shows which agents use the template before anything is published.

Publishing creates a version. It does not silently change running agents. **Update agents** shows the exact instruction, Skill and tool changes, then lets the user choose which agents receive them.

A running update should finish the current task, start the selected version, verify that the agent is available, and restore the previous version if startup fails. The UI should show the actual state throughout.

## Current work

- #4164 adds structured, named MCP configuration to `buzz-acp`.
- #4588 adds Project-owned connection setup without placing credentials in portable agent data.
- #4600 defines a validated, runtime-neutral Skill bundle.
- #4271 can extend the same MCP document with HTTP transport.

The next slice should connect these pieces to Buzz's existing persona, snapshot and managed-agent records. It should not introduce a second agent registry or a separate update system.

The remaining decisions are the source of truth between a template and an agent, how logical tool requirements bind to Project connections, and how a version is rolled out or restored across local and hosted runtimes.

@adamazad @custard-pirate @shellz-n-stuff @kerpopule @sapience7-hub, this overlaps the installation, MCP, shared-instruction and hosted-agent work already in progress. I would appreciate a read on where these boundaries should meet.

wolfyy970:
[...]

PR #4488, opened 2026-08-03
fix(desktop): sync persona respond-to edits onto the linked agent instance

### LINKED ISSUE #4487: Persona respond-to edits never reach the running agent: card shows anyone, harness starts owner-only
### Summary

Editing an agent's respond-to setting on its persona card updates the persona and leaves the running agent on its old gate, permanently. The card shows one mode, the harness enforces another, and no amount of saving or restarting reconciles them.

We hit this across two Macs: an agent whose card read `anyone` (later `allowlist` with four keys) started with `respond_to=owner-only` on all six of its `buzz-acp starting:` lines, including after a full quit and reopen of Buzz Desktop. Every mention from a non-owner was dropped at `author_allowed` while the agent was online and healthy, and answering its owner normally.

Desktop 0.5.3, reproduced against upstream `b1b283c`.

### Root cause

`personaManagedAgentUpdate` (`desktop/src/features/profile/ui/UserProfilePanelUtils.ts:262`) is the function that pushes a persona edit down onto the linked managed-agent instance. It propagates `displayName`, `systemPrompt`, `model`, `envVars` and runtime-derived commands. It does not propagate `respondTo` or `respondToAllowlist`, even though `UpdateManagedAgentRequest` accepts both (`desktop/src-tauri/src/managed_agents/types/requests.rs:251`).

The write path for the persona side is `apply_persona_behavior` (`requests.rs:40`), which by design touches definition fields only. The instance side is populated by `resolve_mint_behavioral_defaults` (`types.rs:931`), which reads the definition's `respond_to` at **mint time only**. There is no later reconcile: `respond_to` appears zero times in `reconcile.rs`.

The spawn env is built from the instance record:

```rust
// desktop/src-tauri/src/managed_agents/runtime.rs, build_respond_to_env
set.push(("BUZZ_ACP_RESPOND_TO", record.respond_to.as_str().to_string()));
```

`ManagedAgentRecord::respond_to` is `#[serde(default)]` and `RespondTo`'s `#[default]` is `OwnerOnly`, so an instance row where the field was never written also deserializes silently to the most restrictive mode.

Observed on disk, `~/Library/Application Support/xyz.block.buzz.app/agents/managed-agents.json`:

| row | `respond_to` | `definition_respond_to` |
|---|---|---|
| definition row (no pubkey) | `owner-only` | `allowlist` (4 keys, correct) |
| instance row (agent pubkey) | `owner-only`, empty allowlist | absent |

### Why it took three days to find

The drop is invisible at the default log level. `inbound author gate — dropping event` is `tracing::debug!` (`crates/buzz-acp/src/lib.rs`) while the harness logs at INFO. A correctly delivered and correctly dropped mention leaves zero trace, so it is indistinguishable from an event that never arrived. We chased relay delivery, mention p-tags, the Desktop mention picker and `is_dm_channel` fail-closed behaviour before reading the agent's own startup line, which is the only place the effective gate is stated.

### Reproduce

1. Create an agent from a persona while the persona's respond-to is `owner-only`.
2. Edit the persona card's respond-to to `anyone` (or `allowlist`) and save.
3. Restart the agent, or fully quit and reopen Buzz Desktop.
4. Read the newest `buzz-acp starting:` line in `~/Library/Application Support/xyz.block.buzz.app/agents/logs/<pubkey>__<hash>.log`.

Expected: `respond_to=anyone`. Actual: `respond_to=owner-only`, indefinitely.

### Fix

PR attached: propagate `respondTo` / `respondToAllowlist` in `personaManagedAgentUpdate`, mode and allowlist together, treating a null persona `respondTo` as unset.

Two adjacent things worth considering separately:

1. **`useAgentManagement.ts:48` hardcodes `respondToAllowlist: []`** when relaying an inbound agent-management `update` request, so a `buzz agents draft-update --respond-to allowlist` can only ever fail the backend's non-empty-allowlist validation.
2. **Log the resolved gate at INFO on spawn and surface it on the card** ("running as: owner-only" next to the configured value). A security gate whose UI cannot be trusted to describe the process, and whose drops are invisible at the default log level, is expensive to diagnose. Promoting the drop line to INFO, or adding a periodic dropped-author count, would have made this a five-minute bug.

### Workaround

Until the fix lands, the only reliable local remedy is editing the instance row's `respond_to` in `managed-agents.json` with Buzz Desktop fully quit, then reopening. Verify by the next `buzz-acp starting:` line; the card is not evidence.

xoxoskeleton:
Fixes #4487.

## Problem

Editing an agent's respond-to on its persona card writes the persona and leaves the running agent on its old gate, permanently.

`personaManagedAgentUpdate` is the function that pushes a persona edit down onto the linked managed-agent instance. It propagates `displayName`, `systemPrompt`, `model`, `envVars` and the runtime-derived commands. It does not propagate `respondTo` / `respondToAllowlist`, even though `UpdateManagedAgentRequest` accepts both.

The harness spawn env is built from the instance record (`build_respond_to_env`), and `resolve_mint_behavioral_defaults` reads the definition at mint time only. Nothing reconciles the two afterwards, so an agent minted as `owner-only` stays `owner-only` for good: the card renders the persona's mode while every `buzz-acp starting:` line prints the stale instance value, and a restart re-reads that same stale value.

The failure is silent on both sides. The card asserts a gate the process does not enforce, and the drop itself (`inbound author gate`) is `tracing::debug!` while the harness logs at INFO, so a correctly delivered and correctly dropped mention leaves no trace at the default log level. It looks exactly like an event that never arrived. That cost us three days and four wrong theories before we read the agent's own startup line.

## Change

One block in `personaManagedAgentUpdate`. Mode and allowlist travel together, matching the backend contract where a present allowlist replaces the stored one and allowlist mode with an empty list is rejected. A null persona `respondTo` means unset and leaves the instance alone, so personas that never authored a behavior group are unaffected.

## Tests

Three new cases in `UserProfilePanelUtils.test.mjs`: mode change propagates, allowlist change propagates, and unchanged-or-unset produces no update. The persona fixture gains `respondTo` / `respondToAllowlist` so it matches the `AgentPersona` type it stands in for.

`pnpm test` in `desktop/`: 3888 pass, 0 fail. `biome check` and `tsc --noEmit` clean on the touched files.

## Not in scope

Two adjacent issues found while tracing this, left out deliberately and described in #4487:

- `useAgentManagement.ts:48` hardcodes `respondToAllowlist: []` when relaying an inbound agent-management `update`, so allowlist mode over that path can only fail validation.
- The resolved gate is not logged at INFO on spawn and is not surfaced on the card, which is what made this expensive rather than trivial to diagnose.

### DIFF
--- a/desktop/src/features/profile/ui/UserProfilePanelUtils.test.mjs
+++ b/desktop/src/features/profile/ui/UserProfilePanelUtils.test.mjs
@@ -58,6 +58,8 @@ function persona(overrides = {}) {
     isBuiltIn: false,
     isActive: true,
     envVars: { NEW_KEY: "2" },
+    respondTo: "owner-only",
+    respondToAllowlist: [],
     createdAt: "2026-01-01T00:00:00Z",
     updatedAt: "2026-01-01T00:00:00Z",
     ...overrides,
@@ -148,6 +150,67 @@ test("personaManagedAgentUpdate leaves runtime fields alone when runtime is unch
         runtimes: [runtime({ id: "goose", command: "goose" })],
       },
     ),
+    null,
+  );
+});
+
+test("personaManagedAgentUpdate syncs an edited respond-to mode to the linked agent", () => {
+  assert.deepEqual(
+    personaManagedAgentUpdate(
+      agent({
+        name: "Fizz Prime",
+        systemPrompt: "New prompt",
+        model: "new-model",
+        envVars: { NEW_KEY: "2" },
+      }),
+      persona({ respondTo: "anyone" }),
+    ),
+    {
+      pubkey: "deadbeef".repeat(8),
+      respondTo: "anyone",
+      respondToAllowlist: [],
+    },
+  );
+});
+
+test("personaManagedAgentUpdate syncs an edited respond-to allowlist to the linked agent", () => {
+  const allowlist = ["a".repeat(64), "b".repeat(64)];
+
+  assert.deepEqual(
+    personaManagedAgentUpdate(
+      agent({
+        name: "Fizz Prime",
+        systemPrompt: "New prompt",
+        model: "new-model",
+        envVars: { NEW_KEY: "2" },
+        respondTo: "allowlist",
+        respondToAllowlist: ["a".repeat(64)],
+      }),
+      persona({ respondTo: "allowlist", respondToAllowlist: allowlist }),
+    ),
+    {
+      pubkey: "deadbeef".repeat(8),
+      respondTo: "allowlist",
+      respondToAllowlist: allowlist,
+    },
+  );
+});
+
+test("personaManagedAgentUpdate leaves respond-to alone when unchanged or unset", () => {
+  const settled = agent({
+    name: "Fizz Prime",
+    systemPrompt: "New prompt",
+    model: "new-model",
+    envVars: { NEW_KEY: "2" },
+    respondTo: "anyone",
+  });
+
+  assert.equal(
+    personaManagedAgentUpdate(settled, persona({ respondTo: "anyone" })),
+    null,
+  );
+  assert.equal(
+    personaManagedAgentUpdate(settled, persona({ respondTo: null })),
     null,
   );
 });
--- a/desktop/src/features/profile/ui/UserProfilePanelUtils.ts
+++ b/desktop/src/features/profile/ui/UserProfilePanelUtils.ts
@@ -292,6 +292,21 @@ export function personaManagedAgentUpdate(
     hasChanges = true;
   }
 
+  // The persona carries the authored respond-to gate; the harness spawn env is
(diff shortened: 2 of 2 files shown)


Chessing234:
nice — syncing persona respond-to onto the instance matches how build_respond_to_env actually reads the harness. tests cover mode + allowlist.

wolfyy970:
This fixes a real mismatch, but I don’t think saving a persona should silently rewrite the linked agent. #4620 now documents the same definition/instance split.

The UI needs an explicit choice: edit this agent, or edit its template. An instance edit should target the keyed record immediately. A template edit should show the affected agents and require an explicit update. That keeps the card honest without turning every template save into an unreviewed rollout.

I have this flow running locally and described the wider boundary in #4301. Could this PR route an instance-card access edit to the keyed record instead of making persona edits propagate automatically?

PR #4492, opened 2026-08-03
fix(cli): preserve agent profile fields when setting policy

### LINKED ISSUE #4490: buzz channels set-add-policy replaces the caller's whole kind:10100 agent profile, erasing name and respond_to
`buzz channels set-add-policy` overwrites the caller's whole kind:10100 agent profile, silently erasing `name` and `respond_to`.

**Where.** `crates/buzz-cli/src/commands/channels.rs` (`cmd_set_add_policy`) builds the event with the policy as the *entire* content:

```rust
let content = serde_json::json!({ "channel_add_policy": policy }).to_string();
use nostr::{EventBuilder, Kind};
let builder = EventBuilder::new(
    Kind::Custom(buzz_sdk::kind::KIND_AGENT_PROFILE as u16),
    &content,
)
.tags([]);
```

**Why it destroys state.** Kind 10100 is `KIND_AGENT_PROFILE` (`crates/buzz-core/src/kind.rs:87`) and is replaceable, keyed by pubkey (`kind.rs:841` asserts `is_replaceable`; `handlers/ingest.rs:421` notes "user-owned replaceable, keyed by pubkey"). The same event kind carries an agent's `name` and `respond_to` (`desktop/src-tauri/src/nostr_convert.rs:441`; test fixture at `:933` is `{"name":"Scout","respond_to":"anyone"}`).

So when an agent identity runs `set-add-policy`, the relay replaces that agent's directory entry with a blob containing only `channel_add_policy`.

**Downstream effect.** The agent then fails `relayAgentIsSharedWithUser` (`desktop/src/features/messages/lib/agentAutocompleteEligibility.ts:12`, which requires `agent.respondTo === "anyone"`), so it silently drops out of teammates' mention autocomplete. No error, no obvious cause, and the recovery is non-obvious.

**Suggested fix**, either works:
- read-modify-write the existing kind:10100 content instead of replacing it, or
- move `channel_add_policy` to its own kind.

Found by reading `main` @ b1b283c, not observed in the wild. Flagging because the failure mode is silent.

Related: #4489 (mention autocomplete), #4487 / #4488 (`respond_to` propagation).

xianjianlf2:
Closes #4490

### What changed
- Read the current kind:10100 agent profile before updating the channel-add policy.
- Preserve existing profile fields such as name and respond_to while replacing channel_add_policy.
- Return a clear error for malformed or non-object profile content.
- Add a regression test for field preservation.

### Validation
- cargo fmt --all
- cargo test -p buzz-cli set_add_policy


Chessing234:
looks right — read-merge-write on kind:10100 so set-add-policy keeps name/respond_to. one nit: if prior content is valid json but not an object, you hard-error; maybe worth a short note in the pr body so ops aren't surprised.

xianjianlf2:
Updated this branch with current main and resolved the channels.rs merge conflict while preserving the kind:10100 profile carry-forward behavior.\n\nValidation run locally:\n- cargo fmt --check\n- cargo test -p buzz-cli channels\n\nAlso noting the maintainer nit: the current behavior intentionally errors if an existing kind:10100 profile is valid JSON but not an object, because there is no safe object field to preserve/update in that case.

xianjianlf2:
Follow-up: amended the merge commit with the required DCO signoff and force-pushed with --force-with-lease. DCO is now passing on head f503f680.

wolfyy970 (approved):
This fixes the current field-clobbering bug cleanly. The command now preserves the existing profile object, changes only channel_add_policy and fails closed on malformed content.

I checked exact head f503f680. All five focused tests, strict buzz-cli Clippy, formatting and the diff check pass.

This is worth landing as the narrow fix for #4490. It does not remove the concurrent two-writer race discussed on #5546, so that review still stands.

xianjianlf2:
Resolved conflicts against upstream main.

Validation:
- `cargo test -p buzz-cli commands::channels::tests` (60 passed)
- `cargo fmt --check`
- `git diff --check`

Current head: 27dda206

xianjianlf2:
Amended the merge commit with DCO signoff.

Current head: 5cf7608c

PR #4494, opened 2026-08-03
fix(desktop): reveal Complete/Snooze/Cancel on inbox reminder rows

### LINKED ISSUE #4493: [Bug] Due reminders cannot be cleared from the inbox Reminders list — actions hidden in the inbox-list row
## Summary

Once a reminder comes due, there is no visible way to clear it from the inbox Reminders list. The Inbox nav badge keeps counting it, and a user who has already dealt with the underlying message concludes that reminders cannot be dismissed at all.

The actions exist — they are just not reachable from the surface where the user looks.

## Steps to reproduce

1. Use **Remind me later** on any message, with a short delay.
2. Wait for the reminder to come due. The Inbox nav badge increments.
3. Deal with the message.
4. Go to Home → inbox filter dropdown → **Reminders**.
5. Look for a way to mark the reminder done.

**Expected:** a Complete (and/or Cancel) affordance on the reminder row.

**Actual:** the row shows author, source, preview, and "Nh overdue" — and nothing else. The badge stays.

## Cause

`ReminderRow` in `desktop/src/features/reminders/ui/RemindersPanel.tsx` suppresses its own action cluster whenever it is rendered in the inbox:

```tsx
// desktop/src/features/reminders/ui/RemindersPanel.tsx:223
{isDone || isInboxList ? null : (
  <div className="flex shrink-0 items-center gap-1">
    ...Complete / SnoozeMenu / Cancel
  </div>
)}
```

`desktop/src/features/home/ui/InboxListPane.tsx:585` is the only caller passing `presentation="inbox-list"`, so in the inbox the buttons are never rendered.

Complete / Snooze / Cancel therefore live *only* in `ReminderDetailPane` (same file), reached by selecting a row. That is a real path — Home → filter → Reminders → click the row → `Complete` in the right-hand detail pane — but nothing on the row hints that selecting it unlocks actions, and on a narrow window the detail pane is a separate view rather than something visible beside the list.

The surface that *did* show inline actions — the standalone `/reminders` route rendering `presentation="card"` — is now a redirect to `/` (`desktop/src/app/routes/reminders.tsx`), so that affordance is gone.

## Related gap (separate fix)

The fire-on-due notification in `desktop/src/features/reminders/useReminderNotifications.ts` has no `Done` action either, so the reminder cannot be cleared from the notification that announces it. Worth its own change; not covered here.

Reminders are kind `30300` (`desktop/src/shared/constants/kinds.ts:58`), NIP-44 self-encrypted, and there is no `buzz reminders` CLI group — so Desktop is currently the only surface that can clear one.

## Proposed fix

Render the existing Complete / Snooze / Cancel controls in the `inbox-list` presentation too, revealed on row hover, keyboard focus, or selection so the list stays visually calm at rest. One condition and one class list in `RemindersPanel.tsx`; no new components, service calls, or event shapes.

PR follows.

jurgenhaas:
Fixes #4493

Once a reminder came due there was no visible way to clear it from the inbox Reminders list, so the Inbox badge kept counting a reminder the user had already dealt with.

`ReminderRow` suppressed its own action cluster whenever it was rendered in the inbox:

```tsx
// desktop/src/features/reminders/ui/RemindersPanel.tsx:223 (before)
{isDone || isInboxList ? null : (
  <div className="flex shrink-0 items-center gap-1">
    ...Complete / SnoozeMenu / Cancel
  </div>
)}
```

`InboxListPane.tsx:585` is the only caller passing `presentation="inbox-list"`, so in the inbox those buttons were never rendered. Complete / Snooze / Cancel existed only in `ReminderDetailPane`, reached by selecting a row — a real path, but nothing on the row hints that selecting it unlocks actions, and the standalone `/reminders` route that used to show them inline is now a redirect to `/`.

## What changed

Render the existing Complete / Snooze / Cancel controls in the `inbox-list` presentation too, revealed on row hover, keyboard focus, or selection so the list stays visually calm at rest.

- Drops `isInboxList` from the suppression condition — `isDone` still hides actions on completed rows.
- Adds a `group/reminder-row` marker on the row so the cluster can be hover-revealed with `group-hover/reminder-row:opacity-100`, plus `group-focus-within` for keyboard users and a `isSelected` override so the selected row keeps its actions visible.

One file, +14/-3. No new components, service calls, or event shapes — the mutations and the detail-pane path are untouched.

## Verification

At `8484874`, hermit active, from `desktop/`:

- `pnpm exec tsc --noEmit` — clean.
- `pnpm exec biome check src/features/reminders` — clean.
- `pnpm test` — `3906 pass / 0 fail` (full desktop unit suite).
- `pnpm exec playwright test --project=smoke --grep reminders` — `11 passed`.
- Wrote a throwaway mock-mode spec that seeds an overdue reminder, opens the Reminders filter, hovers the row, asserts `Complete` becomes visible, and clicks it. Passed; screenshot confirmed the check / clock / X on the row. Spec removed before commit.

Because the actions use `opacity-0` rather than conditional rendering, they stay in the tab order and `group-focus-within` reveals them on focus, so the affordance is reachable without a pointer.

## Not included

The fire-on-due notification in `useReminderNotifications.ts` still has no `Done` action, so a reminder cannot be cleared from the toast that announces it. Noted in #4493; better as its own change.

### DIFF
--- a/desktop/src/features/reminders/ui/RemindersPanel.tsx
+++ b/desktop/src/features/reminders/ui/RemindersPanel.tsx
@@ -162,7 +162,7 @@ function ReminderRow({
   return (
     <div
       className={cn(
-        "flex items-start gap-3 transition-colors",
+        "group/reminder-row flex items-start gap-3 transition-colors",
         isInboxList
           ? "border-b border-border/45 px-4 py-4 hover:bg-muted/40 focus-within:bg-muted/40"
           : "rounded-md border p-3",
@@ -220,8 +220,19 @@ function ReminderRow({
           </p>
         ) : null}
       </button>
-      {isDone || isInboxList ? null : (
-        <div className="flex shrink-0 items-center gap-1">
+      {isDone ? null : (
+        <div
+          className={cn(
+            "flex shrink-0 items-center gap-1",
+            // In the inbox list the actions stay out of the way until the row
+            // is hovered, focused, or selected — without this the only way to
+            // complete a reminder is to open the detail pane, which reads as
+            // "reminders can't be cleared".
+            isInboxList &&
+              "opacity-0 transition-opacity group-hover/reminder-row:opacity-100 group-focus-within/reminder-row:opacity-100",
+            isInboxList && isSelected && "opacity-100",
+          )}
+        >
           <Button
             className="h-7 w-7 p-0"
             disabled={isActing}


Chessing234:
showing complete/snooze/cancel on inbox reminder hover fixes the "can't clear from list" report. one thing: opacity-0 still leaves the buttons focusable via tab — worth checking keyboard-only users can reach them without hover.

Chessing234:
lgtm — revealing complete/snooze on inbox hover fixes the "can't clear due reminders" trap without cluttering the list.

PR #4497, opened 2026-08-03
Fix/desktop repo files truncation 4428

tagadearpit:


Chessing234:
this drops a large chunk of project_git.rs (~600 lines) while only raising the 250 cap. #4499 looks like the fuller fix — discloses truncation instead of silently deleting snapshot helpers. please confirm the deletions were intentional and not a bad rebase.

tagadearpit:
You're right — the ~600-line deletion was not intentional. It was a truncated push of project_git.rs, not a deliberate design choice.
Intended change was only introducing MAX_TREE_ENTRIES = 5_000 and using it in the two .take(250) sites.
Agree #4499 is the fuller/better fix (disclose truncation + total count). Closing this PR in favor of that work.

Chessing234:
raising the hard cap helps, but #4499's total_file_count disclosure is clearer for huge trees — might want to align with that.

tagadearpit:
Agreed — disclosing total_file_count (as in #4499) is clearer than only raising the hard cap. Closing this PR in favor of #4499.

PR #4499, opened 2026-08-03
fix(desktop): disclose truncated repository file trees

### LINKED ISSUE #4428: Desktop: repo Files view silently truncates the tree at 250 entries
Describe the bug
The Files tab of a relay-hosted repo renders at most 250 tree entries and presents the truncated listing as the entire repo. Anything past the 250th entry — including whole top-level directories — is simply absent, with no "show more", pagination, or truncation notice, so the view reads as complete when it isn't.

Steps to reproduce

Host a repo with more than 250 files on a Buzz relay (ours: bag-room, 661 files at commit e7f7160)
In Desktop, open Projects → the repo → Files tab (Remote source, main branch)
The commit header reads "· 250 files" and most of the tree is missing from the listing
Expected behavior
The full tree (paginated is fine), or at minimum an "N of M entries shown" indicator so the truncation is visible.

Version and platform

Buzz version: 0.5.3
OS: Windows 11 Home (build 26200)
Logs / additional context
git ls-tree -r e7f7160 --name-only | wc -l → 661 on the same commit Desktop captions "· 250 files". Screenshot: https://bagroom.communities.buzz.xyz/media/99af08fadb19bc967433732aefa330fd6356ab91fcace80da246e34faa363945.png — note the root listing shows only 2 directories and 5 files. Likely a fixed page size of 250 on the tree fetch with no follow-up page.

MajorTal:
I had more than 250 files and this bug was just misleading me...

## Summary

- Keep the repository snapshot payload capped at 250 file previews.
- Report the full repository file count over Tauri.
- Show an honest "250 of N files" count and a truncation notice in Desktop.
- Cover committed trees and local worktrees with regression tests.

## Root cause

The Rust parsers stopped after 250 entries and returned no total or truncation metadata. Desktop therefore presented the capped array length as the complete repository.

## Testing

- `just desktop-tauri-test` (2,108 passed, 14 ignored; 3 diagnostic tests passed)
- `just desktop-tauri-clippy`
- `cd desktop && pnpm test` (3,926 passed)
- `cd desktop && pnpm check`
- `cd desktop && pnpm build:e2e`
- `cd desktop && pnpm exec playwright test --project=smoke tests/e2e/project-file-tree.spec.ts`

## Manual verification

Open a repository with more than 250 files and select Files. The header shows "250 of <total> files", and a notice states that some files and folders are not included.

Closes #4428

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -105,6 +105,7 @@ export default defineConfig({
         "**/inbox-edit.spec.ts",
         "**/send-channel-binding.spec.ts",
         "**/project-commit-detail.spec.ts",
+        "**/project-file-tree.spec.ts",
         "**/project-inbox.spec.ts",
         "**/project-issue-comments.spec.ts",
         "**/project-pr-review.spec.ts",
--- a/desktop/src-tauri/src/commands/mod.rs
+++ b/desktop/src-tauri/src/commands/mod.rs
@@ -47,6 +47,7 @@ mod project_git;
 mod project_git_branches;
 mod project_git_diff;
 mod project_git_exec;
+mod project_git_files;
 mod project_git_merge_error;
 mod project_git_push;
 mod project_git_workflow;
--- a/desktop/src-tauri/src/commands/project_git.rs
+++ b/desktop/src-tauri/src/commands/project_git.rs
@@ -2,11 +2,11 @@ use super::project_git_exec::{
     build_git_auth_config, clean_branch, clean_target_ref, run_git, validate_workspace_clone_url,
     GitAuthConfig,
 };
+use super::project_git_files::{parse_ls_tree, parse_worktree_files, ParsedProjectRepoFiles};
 use super::project_git_push::push_project_local_repository_blocking;
 use super::project_repo_paths::{canonical_repos_roots, find_local_repo_dir};
 use crate::app_state::AppState;
 use serde::Serialize;
-use std::time::UNIX_EPOCH;
 use tauri::State;
 #[derive(Clone, Serialize)]
 pub struct ProjectRepoCommitInfo {
@@ -38,6 +38,8 @@ pub struct ProjectRepoSnapshotInfo {
     pub latest_commit: Option<ProjectRepoCommitInfo>,
     pub commits: Vec<ProjectRepoCommitInfo>,
     pub files: Vec<ProjectRepoFileInfo>,
+    /// Complete file count before the snapshot payload is capped.
+    pub total_file_count: usize,
     pub contributors: Vec<ProjectRepoContributorInfo>,
 }
 #[derive(Serialize)]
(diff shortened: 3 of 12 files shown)


Chessing234:
disclosing total_file_count + raising the cap is better than #4497's silent delete of snapshot helpers. looks mergeable if the new project_git_files module keeps the path-escape checks from the old parser.

Chessing234:
total_file_count + truncation disclosure is the right fix for silent 250 caps; nicer than just bumping the limit.

MajorTal:
> looks mergeable if the new `project_git_files` module keeps the path-escape checks from the old parser.

Confirmed — the containment check moved verbatim. Diffing `read_preview_content` before and after, the only deltas are a reformatted signature (using the newly imported `Path` alias) and an inlined temporary:

```diff
-fn read_preview_content(
-    repo_dir: &std::path::Path,
-    path: &str,
-    size: Option<u64>,
-) -> Option<String> {
+fn read_preview_content(repo_dir: &Path, path: &str, size: Option<u64>) -> Option<String> {
-    let full_path = repo_dir.join(path);
-    let normalized = full_path.canonicalize().ok()?;
+    let normalized = repo_dir.join(path).canonicalize().ok()?;
```

Every guard in that function is unchanged:

- **Path escape** — `canonicalize()` on both the joined path and the repo root, then `normalized.starts_with(repo_root)`, rejecting anything that resolves outside the repo (symlinks included, since canonicalization happens before the comparison).
- **Size cap** — `MAX_PREVIEW_BYTES` at 64 KiB.
- **Binary rejection** — NUL-byte scan, then `String::from_utf8`.

Scope is also unchanged: the containment check gates *preview content*, not a path's presence in the listing, exactly as before.

One note on the new `total_file_count`, since it feeds the truncation disclosure: `parse_ls_tree` increments it per `ls-tree` line, and the command runs `ls-tree -r --long HEAD` without `-t`, so subtrees are never emitted and the count is blobs only. `parse_worktree_files` increments only after its `metadata.is_file()` check. Both paths count files, not directories, so "showing 250 of N" reports N accurately.

MajorTal:
Closing this as superseded by #5102, which merged yesterday.

That PR removed the 250-entry cap outright rather than disclosing it: `parse_worktree_files` and `parse_ls_tree` now return every tracked path, `MAX_EAGER_FILE_PREVIEWS` bounds only how many files get an eager preview, and the Files tab got incremental 200-entry pagination instead of a hard cutoff.

That's the better fix. This PR's whole premise was that the cap was permanent and the UI should at least be honest about it ("Showing the first 250 of 661 files") — with no truncation left to report, `total_file_count`, `projectFileTreeSummary`, and the disclosure UI have nothing to describe. Rebasing would mean re-implementing a counter for a limit that no longer exists.

Thanks @Chessing234 for the review here, and for the path-escape check you flagged — for the record it did survive the refactor intact (`canonicalize()` on both sides, then `starts_with(repo_root)`), verified before this became moot.

One loose end: this PR was filed against #4428, which is **still open**. #5102 lists #4960 as its fix, but both describe the same root cause — the cap sorted lexicographically, so large repos surfaced only dot-directories, which is exactly #4960's symptom. Commenting there so it can be closed too.

Issue #3880, opened 2026-07-31
Unable to make different identity key on Windows desktop


Aladincz:
Hi,

First, not sure there is similar or identical issue as it is overhelming amount of information for me here. I am not sure whether it is bug or my fault so not sure what to search exactly. I tried and may have not found such issue.

I dl windows client, run it, created identity key, saved to my passwords and went home from work.

I "logged with that identity key home on mac client I got. Created two communities. That was all.

Today at work I tired to log in win client and tried to create another community but with work email, not personal one I used for communities before.

I must honestly say that I may forgot how steps went then. I guess I messed it up.
I got logged to 127.0.0.1 with my work email but when returned to Buzz app i get message:

Finish connecting Buzz
Your Builderlab account (xxx@xxx.xxx) is ready. Connect this device’s Buzz identity to finish setup. Your private key stays on this device.

This Buzz identity is connected to another Builderlab account. Correlation ID: xxxx.....

And I can not continue.

i got logged somewhere in browser, to my account but do not remember how to get there back but I was not able to do that much anyway.

Then I have not see option to create new key. Restarted app and then I see that option.

I tried to create new identity key but i created the same one as before. Maybe it is proper behavior but idk.

Is there a bug or I messed it up?

Thanks

air3ijai:
Had same issue on macOS and the key on the current OS is persistent.

Was able to solve it by generating nostr private key using a Google Chrome extension from a [Nostr](https://github.com/nostr-protocol/nostr) author - [fiatjaf](https://github.com/fiatjaf/nos2x).
There a lot of similar or specific tool for that, but we should take care about  security.

It would be good to have a native way to do this in the application.

Chessing234:
opened https://github.com/block/buzz/pull/3919 — short version: finish connecting always links the key already on the device. to use a different nsec for a work email, sign out under settings → account (backup first), import/create, then connect again.

air3ijai:
@Chessing234, would you mind to make additional adjustments to be able to use `buzz-admin` as a native tool to generate the keys when required and we should be able to use docker for that.

Buzz Desktop does not accept "Secret key hex" at the moment

```shell
docker run --rm -it --entrypoint bash ghcr.io/block/buzz -c "buzz-admin generate-key"
```

```shell
Public key:  43eecc716ce4cfcee598eca49c7986e7a3d7e317e5e0842ee624c2c971e4823e
[...]

PR #4504, opened 2026-08-03
fix: accept buzz-admin hex secrets in desktop key import (#2815)

Chessing234:
## Summary
- `buzz-admin generate-key` now prints both hex and `nsec1` forms (plus npub)
- desktop onboarding / membership-denied import accept 64-char hex secrets the same way `Keys::parse` / `BUZZ_PRIVATE_KEY` already do
- covers #2815 and the hex-paste ask on #3880

## Test plan
- [ ] `buzz-admin generate-key` shows hex + nsec
- [ ] paste the hex into Use an existing key → Next enables and npub preview matches
- [ ] paste nsec / ncryptsec paths still work
- [ ] `cd desktop && npm test -- src/shared/lib/nostrUtils.test.mjs src/features/onboarding/lib/keyImportInput.test.mjs`

### DIFF
--- a/desktop/src/shared/lib/nostrUtils.ts
+++ b/desktop/src/shared/lib/nostrUtils.ts
@@ -1,4 +1,5 @@
-import { decode, npubEncode } from "nostr-tools/nip19";
+import { hexToBytes } from "@noble/hashes/utils.js";
+import { decode, npubEncode, nsecEncode } from "nostr-tools/nip19";
 import { getPublicKey } from "nostr-tools/pure";
 
 /**
@@ -24,6 +25,8 @@ export function safeNpub(pubkey: string): string | null {
 }
 
 const HEX_PUBKEY_REGEX = /^[0-9a-f]{64}$/;
+/** 32-byte secret as 64 hex chars — what `buzz-admin generate-key` prints. */
+const HEX_SECRET_REGEX = /^[0-9a-fA-F]{64}$/;
 
 /**
  * Parse user-entered public key input — either a 64-character hex pubkey or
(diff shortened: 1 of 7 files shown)


Chessing234:
hex import for buzz-admin secrets — pairs with the #3880/#2815 confusion. @wesbillman if hosted/onboarding is on your plate.

Chessing234:
@jurgenhaas @ashbrener small desktop fix for buzz-admin hex secrets on key import (#2815). easy review if you're in that area.

desktop/src/shared/lib/nostrUtils.ts
```diff
@@ -24,6 +25,8 @@ export function safeNpub(pubkey: string): string | null {
 }
 
 const HEX_PUBKEY_REGEX = /^[0-9a-f]{64}$/;
+/** 32-byte secret as 64 hex chars — what `buzz-admin generate-key` prints. */
+const HEX_SECRET_REGEX = /^[0-9a-fA-F]{64}$/;
```
ashbrener: `HEX_SECRET_REGEX` is case-insensitive while `HEX_PUBKEY_REGEX` on the line above is lowercase-only. After this PR, pasting an uppercase hex *secret* works but an uppercase hex *pubkey* still does not — same file, same shape, different tolerance. `parsePubkeyInput` already lowercases bech32 input, so accepting uppercase hex there would be consistent with its own behaviour.

Not blocking, and arguably out of scope for this PR, but the two constants sitting adjacent with different character classes will read as an oversight later.

The same regex is also declared a second time in `keyImportInput.ts`. Exporting one from here and importing it there would keep them from drifting.

ashbrener (commented):
Read this closely because it fixes a papercut I hit yesterday: `buzz-admin generate-key` prints hex, the import form gated on `startsWith("nsec1")`, so the key it just gave me was rejected by the app. Frontend stricter than the backend for no reason. Adding npub to the `generate-key` output is a nice extra — I spent real time yesterday converting a hex pubkey by hand to find which identity a relay was configured with.

Two things I checked specifically, since they are where this kind of change tends to break, and both are fine:

**The new `"hex"` kind cannot fall into a wrong branch.** `KeyImportKind` is only ever branched on at `keyImportInput.ts:123` (`kind === "ncryptsec"`) and `NostrKeyImportForm.tsx:66`. Everything else routes through `nsecToNpub`, which now normalizes hex. No exhaustive switch to update.

**Hex cannot leak into the keyring.** The form submits the raw input, so I expected to find hex being persisted where the rest of the system assumes bech32 — but `import_identity` (`commands/identity.rs:336`) passes the parsed `Keys` object to `persist_imported_identity`, not the string. Storage stays normalized.

One inconsistency worth a look, inline below.

Chessing234:
good catch — tip now exports a shared case-insensitive `HEX_64_REGEX` from nostrUtils and keyImportInput imports it, so pubkey/secret stay aligned.

Chessing234:
@ashbrener tip exports the shared case-insensitive HEX_64_REGEX now. still good to merge from your side?

Issue #3654, opened 2026-07-30
WEBKIT_DISABLE_DMABUF_RENDERER crashes WebKitGTK on NVIDIA instead of falling back to shared memory


kwsantiago:
`webkit_rendering.rs` sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` on NVIDIA GPUs and on AppImage packaging to avoid the #2338 startup abort, with the stated intent of falling back to the shared-memory buffer path.

On current WebKitGTK that variable no longer falls back. It returns before `mode.add(RendererBufferTransportMode::SharedMemory)`, so the transport mode is left empty, `AcceleratedBackingStore::create()` returns nullptr, and `webkitWebViewBaseEnterAcceleratedCompositingMode()` dereferences it, guarded only by an `ASSERT`, which is compiled out in release builds. The app SIGSEGVs the first time content needs a compositing layer. Switching workspaces does it every time.

```
#0 AcceleratedBackingStore::update()   AcceleratedBackingStore.cpp:808
#1 WebPageProxy::enterAcceleratedCompositingMode()
#2 DrawingAreaProxyCoordinatedGraphics::enterAcceleratedCompositingMode()
   <- IPC Messages::DrawingAreaProxy::EnterAcceleratedCompositingMode
```

Measured in the failing process: transport mode `0x00000000`, `create()` returned `0x0`, and `priv->acceleratedBackingStore` is NULL in the core dump. The webview is never disposed (breakpoint on `webkitWebViewBaseDispose` never fires), so the pointer is null from creation rather than from teardown.

`WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` returns *after* SharedMemory is added. That is the documented intent, no hardware dmabuf path, and it leaves the backing store valid.

## Verification

Ubuntu 24.04, WebKitGTK 2.52.3, GeForce RTX 3050 Ti (driver 580.173.02), X11.

| build | module log line | workspace switch |
| --- | --- | --- |
| 0.4.24 AppImage (ships `DISABLE_DMABUF`) | `WEBKIT_DISABLE_DMABUF_RENDERER=1 — NVIDIA GPU, AppImage` | SIGSEGV |
| `main` with `HEURISTIC = [FORCE_SHM]` | `WEBKIT_DMABUF_RENDERER_FORCE_SHM=1 — NVIDIA GPU` | works |

Same result running the AppImage with the two variables set by hand, so the variable is the only difference and not the build.

[...]

PR #4505, opened 2026-08-03
fix(desktop): use WEBKIT_DMABUF_RENDERER_FORCE_SHM for NVIDIA/AppImage (#3654)

Chessing234:
## Summary
- Heuristic and `--safe-rendering` now set `WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` instead of `WEBKIT_DISABLE_DMABUF_RENDERER=1`
- Legacy `DISABLE_DMABUF` stays owned so operators can still set `=0`/`=1` and take over the decision
- Linux troubleshooting docs updated to match (#3654)

## Test plan
- [ ] unit tests in `webkit_rendering::tests`
- [ ] On NVIDIA + WebKitGTK 2.52: workspace switch no longer SIGSEGVs where the distro NVIDIA guard does not fire (Debian/Ubuntu proprietary-NVIDIA may still crash — #3654 stays open for that path)
- [ ] `WEBKIT_DISABLE_DMABUF_RENDERER=0` still stands the heuristic down

### DIFF
--- a/desktop/src-tauri/src/webkit_rendering.rs
+++ b/desktop/src-tauri/src/webkit_rendering.rs
@@ -2,9 +2,14 @@
 //!
 //! WebKitGTK's dmabuf renderer aborts the web process during startup on some
 //! GPU/driver/compositor combinations, so Buzz comes up with no window at all
-//! and the user has no way to fix it (#2338, upstream tauri#9394). Setting
-//! `WEBKIT_DISABLE_DMABUF_RENDERER=1` avoids the abort by falling back to the
-//! shared-memory buffer path.
+//! and the user has no way to fix it (#2338, upstream tauri#9394).
+//!
+//! Historically Buzz set `WEBKIT_DISABLE_DMABUF_RENDERER=1`, which used to fall
+//! back to shared-memory buffers. On current WebKitGTK that variable leaves the
+//! transport mode empty, so `AcceleratedBackingStore::create()` returns null
+//! and the UI SIGSEGVs the first time compositing is needed (#3654).
+//! `WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` is the documented replacement: it keeps
+//! SharedMemory in the transport set and still avoids the hardware dmabuf path.
 //!
 //! WebKit reads each of these variables exactly once per process, so the choice
 //! has to be made before anything initializes — there is no runtime toggle and
@@ -20,7 +25,7 @@
 //!
 //! This is the shape the Tauri ecosystem converged on: clash-verge-rev's
 //! `utils/linux/workarounds.rs` and screenpipe's `linux_webkit_env.rs` both set
-//! the same variable from the same signals at the same point in startup.
+//! WebKit dmabuf env vars from the same signals at the same point in startup.
 
 use std::ffi::{OsStr, OsString};
 use std::path::Path;
@@ -34,21 +39,31 @@ const NVIDIA_PCI_VENDOR: &str = "0x10de";
 /// Where DRM devices advertise their PCI vendor.
 const DRM_ROOT: &str = "/sys/class/drm";
 
-/// Drops the zero-copy dmabuf buffer path. The workaround for #2338.
+/// Prefer shared-memory dmabuf transport. The #3654 replacement for
+/// `WEBKIT_DISABLE_DMABUF_RENDERER` on current WebKitGTK.
+const FORCE_SHM: &str = "WEBKIT_DMABUF_RENDERER_FORCE_SHM";
+/// Legacy kill-switch. Still owned so operators can set `=0` / `=1` and take
+/// the decision away from this module, but never written by the heuristic
+/// (it crashes modern WebKitGTK — see #3654).
 const DISABLE_DMABUF: &str = "WEBKIT_DISABLE_DMABUF_RENDERER";
 /// Drops accelerated compositing as well. `--safe-rendering` only.
 const DISABLE_COMPOSITING: &str = "WEBKIT_DISABLE_COMPOSITING_MODE";
 
-/// What the heuristic applies: the #2338 workaround alone, matching the
-/// ecosystem precedents. `DISABLE_COMPOSITING` is deliberately not here — no
-/// report has isolated it as necessary, and it costs more rendering than this.
-const HEURISTIC: [&str; 1] = [DISABLE_DMABUF];
+/// What the heuristic applies: force shared-memory transport without emptying
+/// the buffer mode set (#3654).
+const HEURISTIC: [&str; 1] = [FORCE_SHM];
 
-/// What `--safe-rendering` applies, which is also every variable this module may
-/// set and therefore every variable a user assignment takes away from it. Being
-/// the same list is the invariant: nothing outside it is ever written, so a user
-/// value for any other WebKit variable is not a conflict.
-const OWNED: [&str; 2] = [DISABLE_DMABUF, DISABLE_COMPOSITING];
+/// What `--safe-rendering` applies: FORCE_SHM plus compositing off. Deliberately
+/// omits DISABLE_DMABUF — that variable is the #3654 crash on current WebKit.
+const SAFE_VARS: [&str; 2] = [FORCE_SHM, DISABLE_COMPOSITING];
+
+/// Every variable this module may set, and therefore every variable a user
+/// assignment takes away from it. Being the same list is the invariant: nothing
+/// outside it is ever written, so a user value for any other WebKit variable is
+/// not a conflict. DISABLE_DMABUF stays owned so `=0` still stands the heuristic
+/// down for operators who need the old path or an explicit override.
+const OWNED: [&str; 3] = [FORCE_SHM, DISABLE_DMABUF, DISABLE_COMPOSITING];
+
 
 /// Reads one environment variable. Injected so the decision is testable without
 /// mutating the process environment. `OsString` rather than `String` because
@@ -132,7 +147,7 @@ fn plan(
 
     if safe_rendering {
         return Plan::Apply {
-            vars: &OWNED,
+            vars: &SAFE_VARS,
             why: format!("{SAFE_RENDERING} requested, this launch only"),
         };
     }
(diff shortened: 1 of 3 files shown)


Chessing234:
nvidia/appimage webkit shm force — small linux desktop fix if anyone on the oss team is in that area.

kwsantiago:
Tested this branch on the hardware from #3654: Ubuntu 24.04 x86_64, WebKitGTK 2.52.3, GeForce RTX 3050 Ti (driver 580.173.02), X11, proprietary NVIDIA driver. Local build, no environment overrides at all.

The heuristic does the right thing:

```
buzz-desktop: WEBKIT_DMABUF_RENDERER_FORCE_SHM=1 — NVIDIA GPU
```

It still SIGSEGVs on the first workspace switch, at the same crash site as #3654 (`AcceleratedBackingStore::update` with a null `this`).

The reason is downstream of this change. Debian and Ubuntu ship `disable-nvidia-dmabuf.patch` in webkit2gtk, which inserts an `isNVIDIA()` check into `rendererBufferTransportMode()` that returns *before* `mode.add(SharedMemory)`. On a machine where that check fires, the transport mode is empty no matter which of the two variables is set, so the backing store is still never created.

What works here is this branch plus pinning EGL to Mesa, which makes the distro patch's `isNVIDIA()` false:

```
__EGL_VENDOR_LIBRARY_FILENAMES=/usr/share/glvnd/egl_vendor.d/50_mesa.json
```

Worth noting the difference from @orestes74's [report](https://github.com/block/buzz/issues/3654#issuecomment-5162791083): Mesa EGL failed on their ARM64 hardware and `FORCE_SHM` alone was enough, which suggests the distro patch's check does not fire there. So the two cases need different things.

Not asking for a change to this PR. It is the right fix, it is what the module's own docs describe, and it helps every case where the distro patch does not trigger. Flagging only so #3654 is not closed as fully resolved by it: on Debian and Ubuntu with the proprietary driver, users will still hit the crash until the distro patch is fixed. That side is reported to Debian separately.

desktop/src-tauri/src/webkit_rendering.rs
```diff
@@ -34,21 +39,31 @@ const NVIDIA_PCI_VENDOR: &str = "0x10de";
...
+/// outside it is ever written, so a user value for any other WebKit variable is
+/// not a conflict. DISABLE_DMABUF stays owned so `=0` still stands the heuristic
+/// down for operators who need the old path or an explicit override.
+const OWNED: [&str; 3] = [FORCE_SHM, DISABLE_DMABUF, DISABLE_COMPOSITING];
 
-/// What `--safe-rendering` applies, which is also every variable this module may
-/// set and therefore every variable a user assignment takes away from it. Being
-/// the same list is the invariant: nothing outside it is ever written, so a user
-/// value for any other WebKit variable is not a conflict.
-const OWNED: [&str; 2] = [DISABLE_DMABUF, DISABLE_COMPOSITING];
```
wpfleger96: 🤖 This double blank line is what's failing `Rust Lint` — `cargo fmt --check` wants one of the two removed ([failing run](https://github.com/block/buzz/actions/runs/30805423453/job/92084935274)). Should be the only thing between you and green CI.

desktop/src-tauri/src/webkit_rendering.rs
```diff
@@ -34,21 +39,31 @@ const NVIDIA_PCI_VENDOR: &str = "0x10de";
 /// Where DRM devices advertise their PCI vendor.
 const DRM_ROOT: &str = "/sys/class/drm";
 
-/// Drops the zero-copy dmabuf buffer path. The workaround for #2338.
+/// Prefer shared-memory dmabuf transport. The #3654 replacement for
+/// `WEBKIT_DISABLE_DMABUF_RENDERER` on current WebKitGTK.
+const FORCE_SHM: &str = "WEBKIT_DMABUF_RENDERER_FORCE_SHM";
+/// Legacy kill-switch. Still owned so operators can set `=0` / `=1` and take
+/// the decision away from this module, but never written by the heuristic
+/// (it crashes modern WebKitGTK — see #3654).
```
wpfleger96, line 47: 🤖 Worth adding a warning breadcrumb in the `Leave` path when the user-set var is `DISABLE_DMABUF` with a value other than `0`: our shipped docs and the older issue threads all told people to `export WEBKIT_DISABLE_DMABUF_RENDERER=1`, so after this merges those users correctly take over the decision but keep the crashy var, and the log only says the env was left as-is. One line pointing at #3654 and `FORCE_SHM` turns a mystery SIGSEGV into a self-serve fix, without touching the takeover invariant.

desktop/src-tauri/src/webkit_rendering/tests.rs
```diff
@@ -192,7 +192,7 @@ fn test_safe_rendering_applies_the_safest_set_without_any_hardware_signal() {
         applied(&plan),
         Some(
             &[
-                "WEBKIT_DISABLE_DMABUF_RENDERER",
+                "WEBKIT_DMABUF_RENDERER_FORCE_SHM",
```
wpfleger96, line 223: 🤖 Nit: none of the tests cover a user-set `FORCE_SHM` standing the heuristic down — it's the new member of `OWNED` and the invariant this PR changes. Same shape as the existing takeover tests; one more would make the 3-element contract explicit.

docs/linux-rendering-troubleshooting.md
```diff
@@ -100,7 +102,7 @@ export WEBKIT_DISABLE_DMABUF_RENDERER=1
 
 ```bash
 export GDK_BACKEND=x11
-export WEBKIT_DISABLE_DMABUF_RENDERER=1
+export WEBKIT_DMABUF_RENDERER_FORCE_SHM=1
```
wpfleger96, line 105: 🤖 This swap silently invalidates a reporter-verified recipe: the "verified by reporter" label above refers to bealmot's verification of the *old* three-var set (with `WEBKIT_DISABLE_DMABUF_RENDERER=1`) on 2.52.4 — nobody has verified this new combination on RDNA4. I'd either keep the verified recipe with a #3654 caveat, relabel this one as recommended-but-unverified, or ask bealmot to re-run on #2643 (he's been responsive).

docs/linux-rendering-troubleshooting.md
```diff
@@ -69,9 +69,11 @@ FONTCONFIG_FILE=~/.config/buzz-fontconfig/fonts.conf ./Buzz_*.AppImage
 
 **Root cause:** WebKitGTK's dmabuf zero-copy buffer path is incompatible with some GPU/driver/compositor combinations. The WebKit child process silently fails to paint.
 
-**Fix (shipped automatically starting with the first release containing [#3271](https://github.com/block/buzz/pull/3271) (v0.5.1)):** Buzz sets `WEBKIT_DISABLE_DMABUF_RENDERER=1` automatically before WebKit initializes when it detects an NVIDIA GPU (`/sys/class/drm` vendor ID `0x10de`) or when running as an AppImage. This restores a slightly slower shared-memory rendering path that works universally.
+**Fix (shipped automatically starting with the first release containing [#3271](https://github.com/block/buzz/pull/3271) (v0.5.1), updated for [#3654](https://github.com/block/buzz/issues/3654)):** Buzz sets `WEBKIT_DMABUF_RENDERER_FORCE_SHM=1` automatically before WebKit initializes when it detects an NVIDIA GPU (`/sys/class/drm` vendor ID `0x10de`) or when running as an AppImage. That keeps SharedMemory in WebKit's transport set while skipping the hardware dmabuf path.
```
wpfleger96: 🤖 Nit: worth one sentence that `WEBKIT_DMABUF_RENDERER_FORCE_SHM` only exists in WebKitGTK ≥ 2.44 (I checked: absent at 2.42.0, present at 2.44.0) — on older system WebKit the new heuristic is a silent no-op. Jammy/bookworm both carry ≥ 2.44 via security updates, so real-world impact is ~nil, but the doc shouldn't imply it works everywhere.

wpfleger96 (changes requested):
🤖 Thanks for the fix. I verified the core claim at WebKit source (tag `webkitgtk-2.52.5`): `WEBKIT_DISABLE_DMABUF_RENDERER` returns before `SharedMemory` is added to the transport set (the #3654 null-deref), while `WEBKIT_DMABUF_RENDERER_FORCE_SHM` is read after it, so it only excludes the hardware dmabuf path. This is the right replacement, and the `Plan`/user-takeover semantics are preserved — keeping `DISABLE_DMABUF` in `OWNED` so `=0`/`=1` still stands the heuristic down is the right call.

Requesting changes for one trivial CI blocker and two doc/UX issues (inline).

One scope note that doesn't need a code change: this is a mitigation, not a universal fix. On machines where WebKit's EGL platform-extension gate fails, the transport set ends up empty regardless of any env var (consistent with the "FORCE_SHM alone wasn't enough" report on #3654), and some distro NVIDIA dmabuf patches can bail out before `SharedMemory` is added. So I'd frame this as fixing the paths our heuristic covers and not close #3654 from this PR alone.

orestes74:
Tested this branch (Chessing234:fix/webkit-force-shm-3654) on DGX Spark (NVIDIA GB10, Grace Blackwell, ARM64), Ubuntu 24.04, WebKitGTK 2.52.3, X11, PCI ID 10de:2e12. No environment variable overrides.

Heuristic correctly logs:

buzz-desktop: WEBKIT_DMABUF_RENDERER_FORCE_SHM=1 — NVIDIA GPU

App renders correctly on first frame (previously needed a devtools-open workaround to force a repaint on this build/patch combo — not needed with this PR). Workspace switching repeated several times with no SIGSEGV.

This matches @orestes74's ARM64 report — confirming FORCE_SHM alone is sufficient on this hardware, no Mesa EGL pin needed.

Separately, unrelated observation: I occasionally see the window fail to repaint when raising it from behind other windows (not a workspace switch). Happens rarely (a few times a day), no crash log or coredump captured yet, so not clear if it's the same root cause or something else. Will file separately with more detail if I can pin it down.

Chessing234:
rustfmt blank-line fix is on tip with dco — needs workflow approval to re-run rust lint.

Chessing234:
@wpfleger96 addressed the review notes on tip — DISABLE_DMABUF leave warning, FORCE_SHM takeover test, RDNA4 recipe marked unverified, and the ≥2.44 note. rustfmt/dco already fixed earlier. ready for another look / merge when you have a minute (also needs workflow approval on the fork).

wpfleger96 (approved):

PR #4508, opened 2026-08-03
docs(e2e): use pnpm build:e2e in perf specs and screenshot skill (#3207)

### LINKED ISSUE #3207: Docs: E2E instructions say `pnpm run build`, which strips the mock bridge and breaks all smoke specs
### Problem

The E2E docs tell contributors to run a **plain production build** before re-running Playwright tests, which produces a bundle **without the E2E Tauri mock bridge** — so every smoke spec fails at boot.

Occurrences of the wrong instruction:

- `AGENTS.md` — "Writing E2E Screenshot Specs" → **Stale server** paragraph: "Kill port 4173 and `pnpm run build` before re-running tests after code changes."
- `desktop/src-tauri/src/managed_agents/screenshot_skill.md` — Gotcha 1 repeats the same `pnpm run build` advice.
- Header comments of the four perf specs in `desktop/tests/e2e/` (`typing-latency.perf.ts`, `scroll-smoothness.perf.ts`, `warm-switch-markdown.perf.ts`, `cold-switch-longtask.perf.ts`) say to run `pnpm build` first — they all call `installMockBridge`, so the same failure applies.

### Why it fails

`desktop/src/main.tsx` deliberately compiles the mock bridge only into dev and explicit E2E builds:

```ts
if (
  !(import.meta.env.DEV || import.meta.env.MODE === "e2e") ||
  !(window as E2eWindow).__BUZZ_E2E__
) {
  return;
}
const { maybeInstallE2eTauriMocks } = await import("@/testing/e2eBridge");
```

`pnpm build` (`vite build`) runs in `production` mode, so the bridge is stripped. The Playwright web server (`python3 -m http.server 4173 -d dist`) then serves a bridge-less bundle and every spec that calls `installMockBridge` dies at boot with:

> Community connection failed / Cannot read properties of undefined (reading 'invoke')

The correct command is `pnpm build:e2e` (`vite build --mode e2e`) — it's what the `test:e2e*` package scripts already use.

### Impact

Two independent agent sessions lost real debugging time to this on 2026-07-27, chasing what looked like an app regression but was just a doc-induced wrong build mode.

### Fix

Update the docs (and perf-spec header comments) to say `pnpm build:e2e`, with a one-line warning that a plain `pnpm build` strips the mock bridge.

Chessing234:
## Summary
- Perf E2E header comments now say `pnpm build:e2e` instead of `pnpm build`
- Screenshot skill gotcha matches AGENTS.md (symlink copies pick it up)
- Fixes #3207: plain production builds strip the mock bridge and break smoke/perf specs

## Test plan
- [ ] Grep the touched perf headers for `pnpm build:e2e`
- [ ] `cd desktop && pnpm build:e2e && pnpm exec playwright test --project=smoke` still boots the mock bridge

### DIFF
--- a/desktop/src-tauri/src/managed_agents/screenshot_skill.md
+++ b/desktop/src-tauri/src/managed_agents/screenshot_skill.md
@@ -96,7 +96,7 @@ Right-click shows "Star channel".
 ## Gotchas
 
 1. **Stale server** — `reuseExistingServer: true` means a prior build serves old
-   code. Kill port 4173 and rebuild (`cd desktop && pnpm run build`) after code changes.
+   code. Kill port 4173 and rebuild (`cd desktop && pnpm build:e2e`) after code changes.
 2. **Clip for readability** — full 1280x720 screenshots are hard to read for sidebar
    features. Sidebar = 256px wide; context menus ~450px.
 3. **`post-screenshots.sh` requires `gh` auth** — the script uses `gh api` and
--- a/desktop/tests/e2e/cold-switch-longtask.perf.ts
+++ b/desktop/tests/e2e/cold-switch-longtask.perf.ts
@@ -35,7 +35,7 @@ import { installMockBridge } from "../helpers/bridge";
  * that is a separate real-wheel pass.
  *
  * Run it:
- *   pnpm build && npx playwright test --config=playwright.perf.config.ts \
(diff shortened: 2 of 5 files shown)


Chessing234:
closing docs-only pr per house policy

PR #4510, opened 2026-08-03
fix(workflows): separate attribution from mentions

### LINKED ISSUE #4507: Workflow owner attribution should not create inbox mentions
## Problem
Workflow-generated channel messages currently use the owner attribution tag as if it were an explicit mention. That can place routine channel automation in the workflow owner's Inbox even when the message never @mentions them.

## Expected behavior
- preserve owner attribution for audit/ownership
- create mention recipients only from explicit message mentions
- repair existing derived mention rows produced solely from workflow ownership

I have a focused relay/workflow/database fix with a migration and tests ready to propose.

Cvv9:
Closes #4507.

## What changed
- keep workflow ownership as attribution instead of treating it as an implicit mention
- resolve @name tokens against channel members and emit p tags only for explicit mentions
- migrate existing workflow messages by removing owner-only derived mention rows while preserving real mentions
- add mention parsing coverage, including ambiguity, Unicode, punctuation, prefix, and deduplication cases

## Why
Routine workflow channel posts were appearing in the workflow owner's Inbox even when the message did not mention them. Ownership and notification intent are separate concepts.

## Verification
- cargo test -p buzz-workflow: 153 passed, 2 Postgres tests ignored
- cargo test -p buzz-relay workflow_sink: 17 passed, 1 Postgres integration test ignored
- commit is DCO signed

### DIFF
--- a/crates/buzz-db/src/migration.rs
+++ b/crates/buzz-db/src/migration.rs
@@ -561,7 +561,11 @@ mod tests {
         let mut migrations: Vec<_> = MIGRATOR.iter().collect();
         migrations.sort_by_key(|migration| migration.version);
 
-        assert_eq!(migrations.len(), 26);
+        assert_eq!(migrations.len(), 27);
+        assert_eq!(
+            migrations.last().map(|migration| migration.version),
+            Some(27)
+        );
         assert_eq!(migrations[0].version, 1);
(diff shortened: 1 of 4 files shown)


Chessing234:
actor vs p-tag split looks correct — ownership shouldn't wake inbox; only explicit @mentions should.

PR #4516, opened 2026-08-03
fix(acp): allow community DMs in anyone mode

XININ-ONLINE:
## Summary

- allow `respond-to=anyone` to admit relay-delivered DM authors
- cover Robert, ROOB, existing members, and members admitted to the community later without per-person configuration
- keep behavior client-agnostic: Desktop and Mobile produce the same signed Buzz event, so the author gate does not special-case platforms
- preserve owner/sibling-only DM behavior for explicit allowlists
- preserve `owner-only` and `nobody` boundaries
- update regression coverage and documentation for the relay-enforced community boundary

## Resulting access model

With the agent configured as `respond-to=anyone`, any author whose event the membership-enforcing community relay delivers can start the agent in a shared channel or DM. This applies automatically to newly admitted community members and does not depend on Desktop versus Mobile. Removing a member at the relay boundary removes their event-delivery path.

## Validation

- 631 unit tests passed
- 9 integration tests passed
- `cargo fmt --check` passed
- clippy passed with `-D warnings`
- release build passed

## Security boundary

This behavior relies on the relay enforcing community and channel admission before `buzz-acp` evaluates the author gate. `respond-to=anyone` is the explicit opt-in; allowlist mode remains hardened in DMs.

Originating Buzz channel: `b15b3fac-7d60-4ad5-879d-b8b4360b1fce`.

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -488,6 +488,17 @@ pub struct ChannelFilter {
     pub require_mention: bool,
 }
 
+impl ChannelFilter {
+    /// Treat every event in a conversational channel as implicitly addressed
+    /// to the agent by removing the relay-level `#p` filter.
+    pub fn with_implicit_mention(mut self, implicit_mention: bool) -> Self {
+        if implicit_mention {
+            self.require_mention = false;
+        }
+        self
+    }
+}
+
 #[derive(Debug)]
 pub struct Config {
     pub keys: Keys,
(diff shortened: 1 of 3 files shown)


XININ-ONLINE:
@tlongwell-block could you review this ACP access-control change when you have a moment? It enables the explicit `respond-to=anyone` mode for relay-delivered DM authors while keeping `owner-only`, `allowlist`, and `nobody` boundaries covered by tests. The XININ deployment uses a membership-enforcing NIP-43 relay, so existing and newly admitted community members get the same behavior across Desktop and Mobile.

PR #4517, opened 2026-08-03
fix(desktop): surface eligible relay agents in @mention

### LINKED ISSUE #4128: Desktop: shared relay agents never appear in mention autocomplete (mobile works)

w3bsited3v:
## Summary

Desktop @mention autocomplete drops shared relay agents — agents owned by
another identity, e.g. a relay-hosted agent run by a separate operator and
shared with a team. `useMentions` filters candidates through
`isAgentIdentityInManagedList`, which only keeps agents present in the
viewer's **own local** `managed-agents` store. So an eligible relay agent is
never offered in `@`-autocomplete, even though the `+` New-DM picker lists it
and DMs to it work.

The mobile client's `buildMentionCandidates`
(`mobile/lib/features/channels/mentions/mention_candidates.dart`) has **no**
such gate — it has no local agent store — and already surfaces any relay
agent that is `anyone` and shares a channel with the viewer. This change
brings desktop in line: keep an agent candidate when it is **either** locally
managed **or** already eligible via `mentionableAgentPubkeys` (i.e. it passed
`relayAgentIsSharedWithUser`). Eligibility semantics are unchanged — the fix
only stops the managed-list filter from dropping already-eligible relay
agents.

### Related issue

Fixes #4128. Related: #3971, #3125, #3776.

### Testing

Verified against a self-hosted relay: a shared, relay-hosted agent
(`respond_to=anyone`, joined to a channel the viewer is a member of, with a
kind:10100 directory entry) now appears in desktop `@`-mention autocomplete,
matching the mobile client. Previously it was reachable only via the `+`
New-DM picker.

- No change for locally-managed agents or non-agent identities (they still
  pass `isAgentIdentityInManagedList`).
- Reuses the existing `mentionableAgentPubkeys` set, so no new eligibility
  path is introduced.

Happy to add a focused test if you'd like — the current
`useMentions.test.mjs` only covers the `hasMention` string helpers, so the
candidate-assembly path isn't unit-tested today; let me know your preference
for structuring one.

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -246,7 +246,13 @@ export function useMentions(
       if (isArchivedDiscovery(pubkey)) {
         return;
       }
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
+      // JDD fork: also keep eligible relay agents (owned by another desktop,
+      // e.g. the operator) so they surface in @mention, not just agents in
+      // this desktop's own managed list.
+      if (
+        !isAgentIdentityInManagedList(candidate, managedAgentPubkeys) &&
+        !mentionableAgentPubkeys.has(pubkey)
+      ) {
         return;
       }
       if (


rmichelena:

rmichelena:
Cross-linking from #5483. Same note as on #4603: I believe this is a no-op against current `main`.

Adding `|| mentionableAgentPubkeys.has(pubkey)` next to `isAgentIdentityInManagedList(candidate, managedAgentPubkeys)` reproduces what #4913 already landed — `main`'s gate is `isAgentIdentityInAllowedList(candidate, mentionableAgentPubkeys)`, and `isAgentIdentityInManagedList` is gone from the module. (Minor: the added comment still says "JDD fork:", which reads like it was meant for a downstream branch.)

And #5363 reproduces *after* #4913, because the failure is upstream of the gate: `list_relay_agents` queries `kind:10100` only, while `respond_to` lives in `kind:30177` and channel membership in `kind:39002`, so relay agents arrive `respond_to: null` / `channel_ids: []` and no gate set can rescue them. Fix at that layer in #5483.

Would you consider closing in favor of it? Maintainers' call — and if I've misread the diff against `main`, say so and I'll correct my note on #5483.

PR #3913, opened 2026-07-31
fix(desktop): require explicit action to mark Inbox read

MajorTal:
Hey good people, I'm Tal and this is my first contribution to this (awesome) project. It's a bug that was bothering me while dogfooding Buzz. I hope it's in good order... 

## Summary

Previewing an unread Inbox item currently calls the same read-state mutation as the explicit **Mark as read** action. That makes ordinary browsing consume the unread queue, and in unread-only mode it can remove the row merely because it was opened.

This change separates selection from completion:

- Opening or keyboard-previewing an Inbox item leaves it unread.
- **Mark as read** remains the explicit completion action and publishes the existing persisted NIP-RS read state.
- In unread-only mode, completing the selected item removes it and predictably selects the next item on wide layouts; narrow layouts return to the list.
- **Mark unread** continues to restore the item coherently.

The implementation keeps the existing persistence path intact and adds a small pure selection helper so next-row behavior is independently testable.

User impact: people can inspect and move through their unread Inbox without losing the queue, while explicit completion still synchronizes through the existing read-state mechanism.

### Related issue

Fixes #3683.

No duplicate open PRs were found for this issue.

### Testing

- `. ./bin/activate-hermit && just ci` on the rebased commit
- 3 focused unit tests for explicit-completion selection
- Chromium Playwright regression covering keyboard preview, unread-only retention, explicit completion, persisted read-state publication, next-row selection, and manual mark-unread
- Existing Inbox-focused Playwright coverage
- Manual local UI verification of selection, unread styling, unread-only filtering, and the explicit context-menu action

Manual verification:

1. Enable **Show unread only** in Inbox.
2. Open several unread items with the mouse or Enter and verify they remain visible and unread.
3. Use **Mark as read** on the selected item and verify it leaves the unread queue.
4. Verify the next unread item is selected on a wide layout.
5. Disable unread-only, use **Mark unread**, and confirm the item returns to the unread queue.

There is no layout or styling change, so static before/after screenshots would show the same UI. The interaction and read-state transition are covered by the Chromium regression and manual steps above.

**Release note:** Inbox items no longer become read when you open them, by click or by keyboard. Clear an item with the explicit "Mark as read" button.

PR #4531, opened 2026-08-03
Guard local checks against low disk space

MajorTal:
## Summary

Tal here. I hit this while working on Buzz: a cold pre-push run can consume enough build output to take a machine from healthy to nearly full without warning.

This adds a small local disk preflight before the five build-heavy Lefthook jobs. The default policy is deliberately absolute: reserve 15 GiB for the checks and require 10 GiB to remain afterward, blocking below 25 GiB free. Documentation-only pushes remain unaffected, unsupported df output fails open, and developers with an already-warm target can tune or bypass only the disk guard without skipping the other hooks.

The focused test exercises both the boundary behavior and real Git pre-push behavior in a temporary repository. It is also wired into the existing CI changes job.

### Related issue

N/A. I searched the open issues and PRs for disk-space and pre-push-hook reports and found no duplicate.

### Testing

- bash -n scripts/check-disk-space.sh scripts/test-check-disk-space.sh
- scripts/test-check-disk-space.sh
- lefthook dump
- git diff --check
- Real temporary Git pushes proving documentation changes skip the guard and Rust changes block before build commands when space is insufficient

Full just ci was not run locally because it is itself the build-heavy path this change is intended to guard; the PR is draft while GitHub CI runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,8 @@ jobs:
         run: scripts/test-mobile-worktree-overrides.sh
       - name: File size policy
         run: just file-size-check
+      - name: Disk space preflight unit tests
+        run: scripts/test-check-disk-space.sh
 
   rust-lint:
     name: Rust Lint
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -157,6 +157,26 @@ Adminer on `:8082`, Keycloak on `:8180` for local OAuth/OIDC testing, MinIO on
 `:9000` for media storage, and Prometheus on `:9090` for metrics) and runs all
 pending database migrations.
 
+#### Disk space for local checks
+
+Rust test, clippy, and Tauri builds can add roughly 15 GiB to a cold Cargo
+target. Before build-heavy pre-push jobs start, Buzz reserves that headroom and
+requires at least 10 GiB to remain afterward. With the defaults, the guard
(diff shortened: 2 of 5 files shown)


MajorTal:
@tlongwell-block hey - could I borrow you for a workflow approval? You did this for me on #4297.

All my open PRs are sitting at `action_required`, so CI has literally never run on any of them. The only checks you see are DCO/Semgrep/zizmor, which are apps and skip the gate - so the green ticks don't mean much.

I got impatient and ran the suite on my own fork, and it turned up two real bugs I'd otherwise have handed you broken:

- **this PR** - the new `Disk space preflight unit tests` step calls `lefthook`, but the `changes` job doesn't activate Hermit, so it died with `command not found` (exit 127). Fixed by using the `bin/lefthook` stub. Since it touches `ci.yml` you'll probably want eyes on it anyway.
- **#4499** - the new Playwright spec was looking for `project-card-buzz` while the Repositories tab actually renders `repository-card-*`. Never found the element, so it failed every retry without ever reaching the assertions it was written for.

Both fixed, and #4531, #4499 and #3913 now have full green runs on the fork. #3900 and #4769 are ready too.

No rush at all - mainly just want the runs unblocked so the checks on the PR pages mean something. Thanks!

MajorTal:
Filed #5470 to describe the underlying problem separately from this diff, per CONTRIBUTING.md's issue-first guidance. The thresholds and the fail-open-on-unparseable-`df` choice are the parts worth an opinion there.

PR #4533, opened 2026-08-03
feat(media): add sharded media object-key paths

bradseiler:
## Why

Media payload keys are currently flat, so high-volume traffic lacks hash-leading S3 prefix entropy. Existing deployments also need a safe, explicit rollout that preserves old reads and rollback behavior without changing storage behavior merely by upgrading.

## What

- moves media payloads to `media/<hh>/<hh>/<sha>.<ext>` for payloads and `media/<hh>/<hh>/<sha>.thumb.jpg` while leaving Git/CAS namespaces unchanged
- adds `BUZZ_MEDIA_MIGRATION_PHASE` with three complete media object-key path policies:
  - `legacy-only` (default): read and write only legacy paths, preserving existing deployment behavior
  - `dual-read-and-write`: prefer sharded reads with legacy fallback and write both paths
  - `sharded-only`: read and write only sharded paths
- defines the existing-deployment rollout as `legacy-only` → `dual-read-and-write` → backfill → `sharded-only`
- recommends that new deployments start with `sharded-only` before their first upload, avoiding any future backfill or legacy cleanup
- adds sharded-first/legacy-fallback read telemetry, strict bucket classification, logical billing deduplication, duplicate-layout gauges, and authoritative payload keys in upload records
- includes `buzz-media-layout-backfill` and `buzz-media-layout-delete-legacy` in both relay image targets, with bounded paging, checkpoints, request-rate limiting, idempotency, destination verification, dry-run, and destructive confirmation
- documents new-install and existing-deployment flows and provides Kubernetes Job examples

## Validation

- `cargo test -p buzz-media --all-targets`: 119 passed; live-MinIO test ignored
- `cargo test -p buzz-relay --lib config::tests::`: 32 passed
- `cargo test -p buzz-relay --lib storage_sweep`: 15 passed
- `cargo clippy -p buzz-media -p buzz-relay --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`, desktop Tauri fmt, and `git diff --check`: clean
- pre-push hooks after merging current `origin/main`: Rust, desktop, Tauri, mobile, and typecheck suites passed
- Helm values/test YAML and values schema JSON parse cleanly; Helm CLI is unavailable locally

Generated with Goose

## Update — August 28, 2026

- refreshed the branch against current `main` (`e76c81968b`) and retained the new version-aware media deletion behavior introduced there
- cleanup resume remains fail-closed; added an explicit regression proving a checkpoint cannot be resumed between communities that share a legacy CAS key
- updated current-main `MediaConfig` fixtures for the new migration phase
- fresh focused validation: `buzz-media` 162 passed / 3 infrastructure-backed ignored; `buzz-deletion` 17 passed / 9 Postgres-backed ignored; relay config 33 passed; storage sweep 15 passed; targeted Clippy, Rustfmt, and diff checks clean
- GitHub CI on `aef74600d4`: 31 successful, 12 intentionally skipped, 0 failing or pending


tlongwell-block (changes requested):
Requesting changes for two brownfield blockers at head `1cfde2831e3b6512f24630b37f9ef130624d2f4e`:

1. **Legacy cleanup cannot safely resume from a sidecar checkpoint.** `verify_selected_destinations` only builds `verified_legacy_keys` from sidecars at/after `BUZZ_MEDIA_MIGRATION_START_AFTER`, then deletes those shared flat CAS keys. One legacy key can serve multiple communities (as the function comment correctly notes). If community A sorts before the checkpoint and lacks its sharded copy while community B sorts after it and has a verified sharded copy, the resumed run verifies B, inserts the shared legacy key, and deletes it, stranding A. The README currently tells operators both tools can restart from any logged checkpoint. Either disallow `start_after` for destructive cleanup or make cleanup prove all sidecar bindings for every candidate legacy key across the full bucket before deletion. Please add the cross-community/resume regression test.

2. **The stale branch revives the removed Helm media-auth flag without a merge conflict.** A no-commit merge of this head into current `origin/main` cleanly restores `relay.requireMediaGetAuth`, `BUZZ_REQUIRE_MEDIA_GET_AUTH`, its schema entry, and render test. Current relay config explicitly classifies that variable as inert because media reads are unconditionally authenticated (`crates/buzz-relay/src/config.rs:434-444,810-814`). The resulting chart advertises a switch whose `false` value does nothing and emits a startup warning. Rebase and drop these unrelated stale additions.

What I traced:
- Media payload read/head/range/stream and imeta checks route through migration candidates; default `legacy-only` preserves existing flat media reads/writes.
- Sidecars and upload records retain their existing namespaces.
- Git store files are untouched; Git keys remain `packs/`, `manifests/`, and `repos/...`, and the only shared S3 setting remains the pre-existing addressing style.

Verification at the exact PR head above:
- `PATH="$PWD/bin:$PATH" cargo test -p buzz-media --all-targets`: 119 passed, live-MinIO test ignored.
- `PATH="$PWD/bin:$PATH" cargo test -p buzz-relay --all-targets`: 879 passed across lib/bin, 40 ignored, 0 failed.
- `git diff --check`: clean.
- Current GitHub checks are green, but they do not exercise destructive cleanup resumption.

Ratings: **Minimalness 6/10** (large but mostly justified migration surface; stale unrelated Helm flag is not), **Elegance 8/10** (phase model and centralized key derivation are clear; cleanup checkpoint contract is misleading), **Correctness 6/10** (upgrade default and live read paths look sound, but the destructive brownfield edge can cause data loss and the clean merge revives inert config). Not ready until both blockers are resolved.

TheSentinel454:
🤖 Codex update: both brownfield blockers from the review at `1cfde2831e3b` are addressed on the current head `aef74600d4`.

1. Destructive cleanup now rejects `start_after`, and the branch includes an explicit cross-community regression proving a checkpoint cannot resume between communities that share a legacy CAS key.
2. The branch was refreshed against current `main` (`e76c81968b`), and the stale `relay.requireMediaGetAuth` / `BUZZ_REQUIRE_MEDIA_GET_AUTH` Helm additions remain removed.

The refresh retained current main’s version-aware deletion behavior. GitHub CI is fully green: 31 successful, 12 intentionally skipped, 0 failing or pending. Could you please re-review?

ravarora2 (changes requested):
🤖 AI review at exact head `aef74600d4914283ea1ed2826782e5ac7299540d`: requesting changes. The two blockers from the earlier review are fixed, but the current head still has the following data-safety and compatibility issues.

1. **P1 — cleanup can delete the last good copy after verification.** `crates/buzz-media/src/bin/buzz-media-layout-delete-legacy.rs:57-105` verifies all selected sharded destinations, but the later deletion pass at `:108-129` only HEADs and deletes each legacy source. It does not re-read or re-verify the corresponding destination. If a destination is removed or changed between the two passes, cleanup deletes the last valid bytes even though the README promises verification immediately before deletion. Re-verify source and destination at deletion time (atomically as far as S3 permits), and add a full-path regression that mutates/removes the destination between verification and deletion and proves the legacy source survives with a non-zero exit.

2. **P1 — both layouts missing is treated as successful reconciliation.** `crates/buzz-media/src/media_layout_common/mod.rs:44-60` maps source `404` plus destination `404` to `LegacySourceAbsent`; backfill and cleanup then skip it and can exit zero. A canonical sidecar can therefore point to no payload bytes without failing the migration. Distinguish “legacy absent, destination present” from “both absent,” make the latter fatal, and cover it through each binary's production path rather than only a pure classification helper.

3. **P1 — valid historical and fresh `.Z` payloads are rejected.** `crates/buzz-media/src/bucket_index.rs:90-94,531-542` documents/tests uppercase `Z` as legitimate `infer 0.19` output for `application/x-compress`, but `crates/buzz-media/src/keys.rs:46-55` accepts only lowercase extensions. Fresh generic-file validation can still return `Z` (`validation.rs:195-215`), after which `upload.rs:91-92` rejects the key. Historical sidecars carrying `ext: "Z"` reach `objects_for_sidecar` (`migration.rs:39-48`), and both migration binaries propagate the error and abort at the first such object. Preserve supported historical mixed-case alphanumeric extension spelling and add fresh-upload plus historical `.Z` backfill/cleanup regressions.

4. **P1 test blocker — the headline phase/migration behavior has no causal object-store coverage.** The checked-in MinIO tests force `LegacyOnly`; the static probe writes an arbitrary `_test/...` key through raw storage calls. They do not exercise `put_payload`, dual/sharded writes, preferred reads and fallback, backfill, or cleanup. Add a live MinIO matrix through the production payload API for LegacyOnly/DualReadAndWrite/ShardedOnly, exact physical keys, GET/HEAD/range behavior, phase-change repair, partial dual-write retry, backfill, dry-run/idempotency/mismatch refusal, and verified deletion. Include mutation gates so removing the sharded write, reversing read preference, removing fallback, or skipping destination verification makes the suite fail.

5. **P2 — maintenance memory is not bounded by page size.** `media_layout_common/mod.rs:93-104` holds both source and destination `Vec<u8>` values simultaneously, requiring roughly 1 GiB plus overhead for one supported 500 MiB video. Cleanup also retains every legacy key in a bucket-wide `BTreeSet`. Stream/hash object verification and use a bounded or durable cleanup manifest; validate peak memory with a large-object and large-inventory probe.

6. **P2 — the checked-in Kubernetes Job examples omit required S3 configuration.** The binaries require endpoint and bucket and must match relay region/addressing style (`media_layout_common/mod.rs:6-29`). The examples import a Secret plus pacing/dry-run values, while the Helm chart stores endpoint/bucket/region/style directly on the Deployment and keeps only credentials in the Secret. Following the example with the chart Secret therefore fails at startup. Render the Jobs from the same chart values, or explicitly supply and test every required non-secret setting.

Verification anchored to the SHA above:

- `cargo test -p buzz-media --all-targets`: 162 passed, 3 ignored.
- The 3 ignored MinIO probes pass when explicitly run, but are pinned to `LegacyOnly` and bypass the PR's production payload/migration behavior.
- `cargo test -p buzz-deletion --all-targets`: 17 passed, 9 ignored.
- GitHub showed 31 successful and 12 skipped checks, with none failing or pending.

The green checks do not exercise the blocking production paths above. Readiness: minimalness 8/10, elegance 7/10, correctness 5/10.

wpfleger96 (changes requested):
🤖 Automated review of `aef74600` against base `e76c8196`. I'm requesting changes. The branch hasn't moved since the review on August 31, and I checked its data-safety blockers against the source at this head. They're still there:

- **Cleanup can delete the last good copy.** `buzz-media-layout-delete-legacy.rs` verifies destinations in `verify_selected_destinations`, but `delete_verified_legacy_objects` later only HEADs the legacy key before deleting it. It never re-checks the sharded destination. If the destination is removed or changed between the two passes, cleanup deletes the only remaining bytes.
- **Both layouts missing counts as success.** `classify_destination_verification` in `bin/media_layout_common/mod.rs` returns `LegacySourceAbsent` when the legacy source and the destination are both `NotFound`. Backfill and cleanup then skip the object and can exit zero, even though the sidecar points to no payload.
- **`.Z` payloads are rejected.** `keys.rs::validate_extension` only accepts lowercase letters and digits. `bucket_index.rs` documents uppercase `Z` as legitimate `infer 0.19` output, so `upload.rs` rejects `legacy_blob_key` for those files. The migration binaries also abort on the first historical `.Z` sidecar.

The branch also conflicts with current main in `deploy/charts/buzz/README.md` (185 commits behind). The earlier review's live-MinIO coverage gap, maintenance-memory bound, and Job-example S3 config points still apply too. I'd fix all of those in the same pass.

I didn't re-derive the full review, because these source-confirmed blockers already decide it. Once there's a new head with the fixes and a rebase, it'll get a full code review and E2E pass.

Issue #4157, opened 2026-08-01
Linked built-in agents discard provider/model edits and return to “needs configuration”


Audacity88:
**Describe the bug**

Provider and model selections made through **Edit Agent** are not persisted for linked built-in agents such as Honey and Bumble. After restarting Buzz, the agents return to “needs configuration” and ask for both fields again.

**Steps to reproduce**

1. Add or start a built-in agent such as Honey or Bumble using `buzz-agent`.
2. Open **Edit Agent**.
3. Select a provider and model.
4. Save the changes.
5. Restart Buzz or restart the agent.
6. Send the agent a message.

The agent reports that its provider and model are not configured.

**Expected behavior**

The provider and model selected in **Edit Agent** should remain effective after saving and restarting.

**Version and platform**

- Buzz version: 0.5.3
- OS: macOS 26.5.1

**Logs / additional context**

The linked agent instances resolve provider and model from their persona definitions, but the edit dialog omits both fields from its update whenever `linkedPersona != null`:

https://github.com/block/buzz/blob/desktop-v0.5.3/desktop/src/features/agents/ui/AgentInstanceEditDialog.tsx#L681-L710

Effective configuration then reads those fields from the linked definition:

https://github.com/block/buzz/blob/desktop-v0.5.3/desktop/src-tauri/src/managed_agents/effective_config/mod.rs#L74-L96

Because the built-in definitions have no provider or model defaults, the visible selections become a no-op. The dialog should either update the linked definition or direct the user to a definition-level configuration surface.

iroiro147:
Fixed in PR #4268.

**Root cause:** For linked built-in personas, `AgentInstanceEditDialog.handleSubmit` sends `provider: undefined` and `model: undefined`, deferring to the persona definition. Built-in personas (Honey, Bumble) ship with both unset, so your edits were silently dropped and the agent fell back to global defaults — surfacing as "needs configuration" after restart.

**Fix:** When a linked persona's provider or model is edited in the dialog, the value is now propagated to the persona definition itself via `updatePersona`, so the configuration survives restart. Unlinked instances are unaffected.

Issue #3753, opened 2026-07-30
Managed-agent instance config (kind:30177) never syncs cross-device, causing every member device to auto-offer a duplicate/unconfigured local agent


priyanshusaleshandy:
## Summary

When a persona (e.g. an agent named "Fizz") is created and properly configured on one device (Owner's host — provider/model/harness set, e.g. Codex), **every other device that simply joins the same channel/community independently offers to run its own separate local instance of that same persona**, defaulting to "needs setup on this device." If that device has any agent runtime enabled (e.g. the `Buzz Agent` runtime, which cannot be toggled off in the UI), it auto-spawns a local `buzz-acp` process in a setup-mode listener state — which then sends spurious "Fizz needs configuration" replies into the channel, competing with the properly-configured instance on the intended host machine.

This defeats the "one shared team agent" use case entirely — every team member's device ends up needing its own separate, fully-duplicated configuration (including a separate provider login, e.g. a separate `codex login`/ChatGPT session per device) just to avoid broken replies, even though only one device is meant to actually host the agent.

## Root cause (traced in source)

Two relay sync paths are asymmetric:

- **Persona definition** (kind:30175) — propagates to every member device unconditionally. `apply_inbound_persona` (`desktop/src-tauri/src/commands/personas/inbound.rs`, ~line 342-364) inserts the persona locally on any device that receives the event, regardless of `BackendKind`.
- **Managed agent instance record** (kind:30177) — carries the actual `backend` field (`BackendKind::Local` vs `BackendKind::Provider{id, config}`) and is the thing that would tell another device "this persona is already actively hosted elsewhere." `apply_inbound_managed_agent` (same file, ~line 376-405) is an explicit no-op when there's no local match, per the existing comment: *"an agent that does not already exist locally has no secret key to run with, so inserting a secretless shell would be useless and misleading."*

Because only the definition (kind:30175) propagates and never the instance/backend state (kind:30177), every device that receives the persona has no way to know it's already hosted elsewhere. `ManagedAgentRuntimeStatus.localSetup` (`desktop/src/shared/api/types.ts`) and `agentCommunityAvailability()` (`desktop/src/features/agents/managedAgentRuntimeStatus.ts`, ~line 9-25) compute "needs setup on this device" purely from local state — there's no cross-device ownership/exclusivity check anywhere in `reconcile.rs`, `discovery.rs`, or `persona_events.rs`.

[...]

Issue #3727, opened 2026-07-30
[Bug] Desktop 0.5.2 re-runs onboarding on every launch on a second machine — new Welcome channel + duplicate Honey each time


Tkander1715:
**Describe the bug**

On a second machine signed into an existing community, Buzz Desktop re-runs first-time onboarding on **every app launch**, not just on first install. Each launch creates another "Welcome" channel and mints another copy of the default starter agent (Honey) with a fresh Nostr keypair.

This looks related to #2648 (second install mints the default roster once) but is a distinct, more aggressive variant: the re-mint happens **per launch on a single install**, and on v0.5.2, which already includes the duplicate-minting guard from #3221.

**Steps to reproduce**

1. Machine A (Mac mini): create a hosted community (`*.communities.buzz.xyz`) with Buzz Desktop; delete the default starter agent (Honey) from the community.
2. Machine B (MacBook): install Buzz Desktop v0.5.2 in `/Applications` (installed from the DMG correctly — not running from the mounted image), and sign in with the **same existing identity** into the **same community**.
3. Observe: a new "Welcome" channel appears and a new "Honey" agent (fresh pubkey, same display name/avatar) is created.
4. Quit Buzz Desktop and relaunch. Another Welcome channel and another Honey are created. This repeats on every launch.

**Expected behavior**

Onboarding/bootstrap should run at most once per install — ideally not at all when the identity already owns channels in the community. Relaunching the app should never mint new starter channels or agent identities.

**Version and platform**

- Buzz version: 0.5.2 (machine B, where the re-mint happens)
- OS: macOS (both machines)

**Logs / additional context**

Relay-side evidence, gathered with the CLI against the community relay: `buzz users get --name Honey` returns **4 distinct profiles**, identical display name and avatar, different pubkeys (first 8: `6f780449`, `13d7678b`, `ef1c2dc4`, `de94ed7a`) — one per onboarding run. The original channels and the owner's membership are intact throughout; only the starter content is duplicated.

Possibly relevant detail: the community's original default Honey agent had been **deleted** on machine A before machine B ever signed in. If the launch-time bootstrap checks "does the default roster exist?" and recreates it when missing, that would explain both the recreation of a deleted agent and the per-launch loop (machine B's freshly minted Honeys live only in machine B's local state, so a check against synced/local state could keep failing).

The duplicate identities appear inert (their private keys exist only on machine B), matching the "dead on arrival" duplicates described in #2648.

jbforge:
[...]

Issue #3013, opened 2026-07-26
Managed starter agents show RUNNING but never respond to DMs or mentions in local `just dev` environment


jordanveres:
# Managed starter agents show RUNNING but never respond to DMs or mentions in local `just dev` environment

## Summary

Starter agents (Fizz, Honey, Bumble) appear healthy and report `RUNNING`, but they never respond to either direct messages or channel mentions when running Buzz locally via `just dev`.

The relay continues to accept and ingest events normally, but no replies are ever produced.

---

# Environment

- macOS (Apple Silicon)
- Docker Desktop
- Buzz running from source
- Branch: `main`

Commit tested:

```text
dd222a509b156ba52ed3219e895d7bf1cf322c92
```

Working tree:

```text
On branch main
Your branch is up to date with 'origin/main'.

nothing to commit, working tree clean
```

---

# Setup

Repository cloned locally.

Ran:

```bash
just setup
just dev
```

Completed onboarding by:

- Importing an existing Nostr identity
- Allowing macOS Keychain access ("Always Allow")
- Joining the local development community
- Accepting the default starter agents (Fizz, Honey, Bumble)

---

# Reproduction

1. Clone the repository.
2. Run:

```bash
just setup
just dev
```

3. Complete onboarding.
4. Import an existing Nostr identity.
5. Verify the starter agents appear.
6. Verify each starter agent reports:
   - Runtime: Codex
   - Status: RUNNING
   - Green status indicator
7. Open either:
   - a direct message with Fizz, or
   - the `#Welcome` channel.
8. Send:

```text
Reply with exactly: hello
```

(or any other prompt)

---

# Expected behavior

The starter agent should receive the message, invoke the configured Codex runtime, and reply.

The Harness Log should show activity corresponding to the request.

---

# Actual behavior

The message is accepted by Buzz but no reply is ever generated.

Observed behavior:

- Starter agents remain `RUNNING`
- Direct messages never receive a response
- Channel mentions never receive a response
- Harness Log initially remained empty
- No reply is ever posted back into Buzz

---

# Relay observations

The relay appears healthy.

Observed during testing:

- Events are accepted successfully
- Events are ingested successfully
- Relay continues normal operation

Representative relay output:

```text
HTTP POST /events
accepted=true

Event ingested via pipeline
```

The relay continues processing requests normally after the event is accepted.

---

# Additional observations

While reproducing the issue:

- The Codex runtime reports as configured.
- Starter agents show a green RUNNING indicator.
[...]

PR #4540, opened 2026-08-03
fix(desktop): use effective readiness for welcome kickoff

loganj:
🤖
## Summary

Buzz creates Bumble, Fizz, and Honey during Welcome, then decides whether to start them. That decision was made with a frontend shortcut that did **not** use the configuration the backend would actually use to spawn the selected agent.

This produced different failures depending on the build:

| Build / setup | Before | After this change |
|---|---|---|
| **Internal macOS release cut by `squareup/buzz-releases` with packaged Databricks defaults** | Welcome looked only at the empty per-install settings file, incorrectly said a provider was required, and left the starter team idle—even though spawn already had a provider, host, and model from the release build. | Welcome evaluates the selected Fizz instance through the backend. It sees the packaged defaults, starts all three agents, and posts the live Fizz kickoff. |
| **Development app with a ready non-Buzz runtime** (for example, a selected CLI runtime that is installed and logged in) | Welcome's broad frontend check could inspect a different available runtime than the one assigned to the starter team, so its decision could disagree with the eventual spawn. | Welcome checks the exact Fizz instance and its selected runtime. A runtime that spawn can use is started; one that spawn cannot use is reported not ready. |
| **Fresh development app-data with no packaged defaults, no logged-in CLI runtime, and no credentials of its own** | Not ready. | Still not ready. This change intentionally does not copy credentials from the main Buzz identity or invent configuration. The user must configure or log in to a runtime in that development app identity. |

The root problem was therefore not that spawn lacked provider/model support. Spawn already resolves the managed instance's selected runtime and its effective configuration in this order: packaged release defaults, app-wide defaults, persona settings, and instance settings. Welcome used a separate approximation over persisted app-wide settings and could disagree with spawn.

This change adds a backend readiness command for one managed-agent instance and makes Welcome call it for the Fizz instance it just provisioned. The command reuses spawn's fallible runtime resolution and effective configuration. It also preserves spawn's behavior for references to deleted runtimes: those remain not ready instead of silently falling back to another runtime.

Provider credentials remain isolated to each Buzz app identity. This change does not copy credentials, persist packaged defaults, or change what **Skip for now** means. It fixes builds that already have a spawn-ready path but Welcome fails to recognize it; it does not make an actually unconfigured development sandbox ready.

The packaged Databricks floor is an intentional Block distribution boundary, not part of ordinary local or public builds. The private `squareup/buzz-releases` macOS pipeline supplies `BUZZ_BUILD_BUZZ_AGENT_PROVIDER` and `BUZZ_BUILD_AGENT_ENV` when compiling `block/buzz`; normal `just dev` and public/OSS release jobs do not. This PR consumes that existing floor consistently but does not change which builds receive it. Making pristine local development reproduce the internal first-run agent experience would require a separate sanctioned dev configuration/authentication path.

### Related issue

No exact issue found. Related reports: #4157, #3013, #2931, #3727, and #3753.

### Testing

Core behavior was validated at `38d1a13e75161df116e5c77f50b0fb330b44c37b`; the follow-up CI-only size-ratchet adjustment is `2a7b2c36ff2f5ddbe2d23b92086cd497bc5b0e00`:

- At `38d1a13e75161df116e5c77f50b0fb330b44c37b`: full desktop Tauri Rust suite (2,050 passed, 0 failed, 14 ignored), plus `pnpm -r check`
- At current head `2a7b2c36ff2f5ddbe2d23b92086cd497bc5b0e00`: `pnpm --dir desktop check:file-sizes` and `pnpm --dir desktop check`
- Playwright onboarding regression: a first-run install with empty persisted agent settings uses packaged defaults and posts the live Fizz kickoff
- Regression coverage for selected-runtime readiness, release-default and provider fallback behavior, and deleted-runtime references

The manual fresh-install walkthrough was blocked by macOS Accessibility automation, so the automated first-run flow provides the reproducible end-to-end coverage.

### DIFF
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -55,10 +55,10 @@ use huddle::{
     set_voice_input_mode, speak_agent_message, start_huddle, start_stt_pipeline,
 };
 use managed_agents::{
-    backfill_persona_snapshots, ensure_nest, list_managed_agent_runtimes,
-    put_managed_agent_runtime_lifecycle, reconcile_managed_agent_runtimes,
-    restart_managed_agent_runtime, start_managed_agent_runtime, stop_managed_agent_runtime,
-    try_regenerate_nest,
+    backfill_persona_snapshots, ensure_nest, get_managed_agent_readiness,
+    list_managed_agent_runtimes, put_managed_agent_runtime_lifecycle,
+    reconcile_managed_agent_runtimes, restart_managed_agent_runtime, start_managed_agent_runtime,
+    stop_managed_agent_runtime, try_regenerate_nest,
 };
 #[cfg(not(feature = "mesh-llm"))]
 use mesh_llm_stubs::*;
@@ -801,6 +801,7 @@ pub fn run() {
             resolve_oa_owner,
             list_relay_agents,
             list_managed_agents,
+            get_managed_agent_readiness,
             list_managed_agent_runtimes,
             start_managed_agent_runtime,
             stop_managed_agent_runtime,
--- a/desktop/src-tauri/src/lib.rs
+++ b/desktop/src-tauri/src/lib.rs
@@ -949,7 +949,6 @@ pub fn run() {
 
     #[cfg(unix)]
     shutdown::install_signal_handler(app.handle().clone(), Arc::clone(&shutdown_done));
-
     let run_shutdown_done = Arc::clone(&shutdown_done);
     let restart_requested = Arc::new(AtomicBool::new(false));
     app.run(move |app_handle, event| match event {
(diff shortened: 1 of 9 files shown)


bradhallett:
Heads-up on a `readiness.rs` overlap @loganj — #3804 patches `resolve_effective_harness_descriptor` in `desktop/src-tauri/src/managed_agents/readiness.rs` (preset launch-args fallback for command-override agents with no runtime id). This PR's `readiness.rs` rewrite touches the same function territory.

To minimize rework on both sides it'd help to land #3804 first: it's a small, targeted `else`-branch addition, so your rewrite would only need to preserve that one branch on rebase. The reverse — this landing first — would mean rebasing a surgical fix against a large function restructure.

#3804 is now rebased onto `main`, conflict-free, and mergeable (waiting on review only). Flagging now so neither of us eats an avoidable rebase — happy to coordinate sequencing either way.

PR #4542, opened 2026-08-03
relay: fuzz WebSocket 1012 restart-close timing on graceful drain (BUZZ_DRAIN_JITTER_MS)

bradseiler:
## Problem

On SIGTERM the relay sends every live WebSocket a **1012 Service Restart** close frame via `ConnectionManager::drain_all()` — all in the same instant (`main.rs` shutdown task → `state.rs::drain_all`). On a pod holding thousands of sessions, that makes every client reconnect simultaneously: the thundering-herd reconnect behind the DB pool-timeout bursts observed on each rolling deploy. Client-side jitter can't fix this — the desktop client *resets* its backoff to base on a 1012 and reconnects with only ±25% jitter (`relayClientSession.ts`), so the spread has to come from the server.

## Change

Add `BUZZ_DRAIN_JITTER_MS` (default `0` = unchanged behavior). The two paths are kept **deliberately separate** so the default is byte-for-byte the previously shipped shutdown:

- **Jitter off (`0`/unset, the default):** the original synchronous, all-at-once `drain_all()` runs unchanged — queue the 1012 on each connection's control channel, cancel, return. No new machinery on the default path.
- **Jitter on (`> 0`):** a separate async `drain_all_jittered(jitter_ms)` spreads each connection's restart close over an independent uniform delay in **`[1, jitter_ms]`**. Each delayed close travels a dedicated `RestartClose` channel; the writer flushes the 1012 frame and **acknowledges the flush over a oneshot**, so drain waits for confirmed delivery (up to `RESTART_CLOSE_ACK_TIMEOUT` = 5s) rather than assuming it, falling back to cancellation if the channel is full/closed or the ack times out. The drain future is **owned and awaited** by the shutdown task, and the 30s hard-drain backstop is aborted only after a clean drain — so a clean roll exits `0`.

The two methods can be unified and the old one dropped later once the jittered path is proven for all cases.

- **`config.rs`** — `drain_jitter_ms`: non-negative parse, clamped to `MAX_DRAIN_JITTER_MS` = **20s** (leaving 10s of the 30s budget for flush). Junk fails loudly at startup; **empty/whitespace-only is treated as unset (jitter off)** so a `BUZZ_DRAIN_JITTER_MS=""` kill switch does not crashloop the relay (matches the sibling env vars in this file).
- **`state.rs`** — `drain_all()` (unchanged synchronous default) + `drain_all_jittered()` (jittered + flush-ack). Both set the sticky `draining` flag before the first await. A registration that lands mid-shutdown always self-signals via the **immediate** control-frame + cancel path — jitter smears already-established sockets, not late arrivals.
- **`main.rs`** — shutdown task dispatches: `drain_jitter_ms == 0` → `drain_all()`, else `drain_all_jittered(...).await`.

## Safety

- **Default off is the currently-committed path.** With jitter unset/0 the shutdown runs the original synchronous `drain_all()` — no restart channel, no ack wait. Safe to deploy dark and dial up.
- **Shutdown-boundary race preserved.** Sticky flag set before any await; a late registration self-signals its close with no jitter.
- **Owned + backstopped.** The jittered drain future is awaited; the 30s hard-drain `process::exit(1)` remains the ceiling. `MAX_DRAIN_JITTER_MS` (20s) + `RESTART_CLOSE_ACK_TIMEOUT` (5s) = 25s, inside the 30s budget; 5s pre-sleep + 25s = 30s against `terminationGracePeriodSeconds: 60`.

## Known behavior to note (not a blocker, flagged from review)

On a **successful** flush the jittered path deliberately does not cancel the connection token — teardown then depends on the client echoing our Close, or on process exit. Compliant clients echo; a silent client rides to the 30s hard exit. The default (jitter-off) path cancels deterministically as before.

## Tests

- `config::tests::drain_jitter_defaults_off_and_rejects_junk` — default off, `20000`, clamp `60000`→`20000`, explicit `0`, junk `"soon"` fails, **empty `""` and whitespace-only treated as off**.
- `state::tests::drain_all_is_immediate` — default path queues frame + cancels synchronously.
- `state::tests::drain_all_sends_restart_close_and_cancels_every_conn`, `drain_all_full_control_buffer_still_cancels`, `register_after_drain_self_signals_restart_close_and_cancel`.
- `state::tests::drain_all_jittered_defers_close_until_within_jitter_window` (paused time).
- `state::tests::drain_all_jittered_waits_for_writer_acknowledgement_without_cancelling`.
- `state::tests::drain_all_jittered_cancels_when_restart_channel_is_full_or_closed`.
- `state::tests::drain_all_jittered_cancels_when_flush_ack_times_out` (paused time — the 5s ack-timeout fallback).

Validation at `46c690940`: `cargo fmt -p buzz-relay --check`, `cargo clippy -p buzz-relay --all-targets -- -D warnings`, and the drain/config unit suite all clean. Local live SIGTERM test with a real relay process + 200 NIP-42-authenticated sockets — see the PR comment for the before/after distribution and exit codes.

## Rollout

Ship with default `0`, then set `BUZZ_DRAIN_JITTER_MS` (e.g. 10000–20000) on bb-block first, watch the roll-window pool-timeout metric, then bb-public. `""` is a safe kill switch. Complements the preStop `sleep` (stops routing before close).


tlongwell-block (commented):
Reviewed at `344bfdc486537f5d0036263a5d5ff4cfa40acda3`.

**Verified locally (this SHA, same shell):**
- `cargo clippy -p buzz-relay --all-targets -- -D warnings` — clean.
- Full `cargo test -p buzz-relay --lib` — 840 passed, 1 failed: `api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo` (504 vs 200). Reproduced the identical failure on `origin/main` @ `027a74a61`, so it is pre-existing and unrelated to this PR.
- Traced the call path: `main.rs` shutdown task → `drain_all_jittered` → per-conn detached task → `restart_close_frame()` + `cancel()`. The store-then-iterate sticky-flag ordering matches `drain_all`'s existing contract with `register`'s insert-then-check, and the late-registration self-signal is untouched and covered by the paused-time test.

**One real issue — the phantom cap 👻:**

The `config.rs` doc for `drain_jitter_ms` says *"Values are capped at shutdown to leave headroom under the 30s hard-drain timeout"*, and the parse-site comment repeats *"Bound is enforced (capped) at shutdown."* No such cap exists — `main.rs:1230` passes `config.drain_jitter_ms` straight through, and `drain_all_jittered` uses it unclamped.

Consequence: an operator who sets `BUZZ_DRAIN_JITTER_MS=45000` (trusting the documented cap) gets every connection whose draw lands past ~30s killed by the hard-drain `process::exit(1)` — a TCP reset with no 1012 frame, which is precisely the silent-death failure mode this drain path exists to prevent. Default-0 is safe, but the doc promises a guardrail the code doesn't have.

Fix is one line either way:
1. (preferred) actually cap at the call site, e.g. `drain_jitter_ms.min(25_000)` with a comment tying it to the 30s hard timeout — makes the config doc true; or
2. fix the doc to say the value is **not** capped and must be kept well under 30s (the `state.rs` docstring already says this correctly — the two docs currently contradict each other).

**Nits (no action required):**
- `self: &Arc<Self>` receiver on `drain_all_jittered` — nothing in the spawned tasks captures `self`; `&self` would do.
- Docstring says the delay is drawn from `[0, jitter_ms]`; `% jitter_ms` is `[0, jitter_ms)`. The inline comment has it right.

Everything else is genuinely tight: default-off byte-identical path, race guarantee preserved and tested under paused time, uses the existing relay jitter idiom, 239 lines mostly docs+tests. Fix the phantom cap (or its documentation) and this is ready.

tlongwell-block (dismissed):
**Correction to my review above — the jitter mechanism does not survive shutdown. Downgrading from "fix the cap and ship" to "needs a redesign of task lifetime."**

Max ran the real relay process (200 NIP-42-authenticated sockets, `BUZZ_DRAIN_JITTER_MS=2000`): the relay logged `connections=200`, then **0/200 received the 1012 frame — all 200 got abnormal 1006 closes at ~5.018s**. The `jitter=0` control delivered 1012 to 100/100.

Root cause, confirmed in the code: `drain_all_jittered` spawns detached timer tasks and returns immediately without cancelling any connection. With no `cancel()` fired synchronously, `axum::serve(...).with_graceful_shutdown` (main.rs:1281) completes as soon as the watch signal fires, `serve()` returns at main.rs:1076, `main` falls through the audit drain and returns — and `#[tokio::main]`'s runtime drop destroys every parked jitter timer **and** the 30s hard-timeout task. The frames never send; clients get a TCP-level 1006. The `jitter=0` path only works because `drain_all` queues the frame and cancels synchronously *before* the graceful-shutdown signal propagates.

The paused-time `ConnectionManager` unit tests can't see this — they never tear the runtime down. That's why they pass while the live process fails at even 2s of jitter.

What a fix needs:
1. The shutdown task must **await completion of all jittered closes** (e.g. `JoinSet` / collected handles) before allowing `serve()`'s return path to reach the end of `main` — or `main` must explicitly wait on a drain-complete signal after `serve()` returns.
2. The `>30s` cap issue from my earlier comment still stands once the lifetime is fixed.
3. Validation must include a **real-process test** (spawn the relay, open live WS connections, SIGTERM, assert 1012s arrive spread over the window) — unit tests alone demonstrably pass on a broken implementation.

Credit: live failure isolated by Max (200-conn regression vs 0ms control). My earlier "mechanism itself is sound" was wrong — it was sound only inside a runtime that stays alive.

bradseiler:
## Local live relay SIGTERM test — commit `46c690940`

Ran a real `buzz-relay` process (not a unit harness) against local Postgres + Redis, opened **200 NIP-42-authenticated** WebSocket sockets, signalled a real `SIGTERM`, and recorded the close code, reason, and per-connection close timing each client observed. Harness is an ephemeral `drain_probe` bin (not committed). All results below are at HEAD `46c690940`.

**Setup:** `DATABASE_URL` + `REDIS_URL` local, `BUZZ_BIND_ADDR=0.0.0.0:3030`, host row seeded in `communities`. Timing offsets are measured from the moment all 200 clients finished authenticating; the drain path includes the fixed **5s readiness grace** before draining, so the ~5s baseline offset is expected in every case.

### Jitter off (default, `BUZZ_DRAIN_JITTER_MS` unset) — the currently-committed path
| run | 1012 clean | resets (no-close-handshake) | close spread | relay exit | drain wall |
|-----|-----------|------------------------------|--------------|-----------|-----------|
| A | 187/200 | 13/200 | 12 ms | **0** | 5.04 s |
| B | 200/200 | 0/200 | 11 ms | **0** | 5.03 s |

All closes land in a **~11-12 ms window** — this is the synchronized herd the jitter knob exists to spread. Note the reset count is **intermittent** on the synchronous path: run A saw 13/200 `Connection reset without closing handshake` (the 1006-equivalent) at the very tail of the burst as the process raced to exit; run B saw zero. Exit code is 0 (clean) either way.

### Jitter on
| jitter | 1012 clean | resets | close spread | 1s-bucket occupancy | relay exit | drain wall |
|--------|-----------|--------|--------------|---------------------|-----------|-----------|
| `5000` | **200/200** | 0 | 4854 ms | 44/29/48/41/38 | **0** | 10.03 s |
| `10000` | **200/200** | 0 | 9970 ms | 13/24/16/19/19/24/23/22/16/24 (max share 12%) | **0** | 15.03 s |

With jitter on, **every** socket received a clean `1012 "relay restarting"`, **zero** resets, and the closes were spread approximately uniformly across the full window (no t=0 spike, max 1s-bucket share 12% vs ideal 10%). The flush-acknowledgement path deterministically eliminates the tail resets seen intermittently on the synchronous path. Relay exited `0`; drain wall = 5s grace + jitter window, well under the 30s hard-drain backstop. The "Signalled restart close to all connections" log line fires *after* the last delayed close, confirming the drain future is owned and awaited.

### Config edge cases (real process boot)
- `BUZZ_DRAIN_JITTER_MS=60000` → clamped to `20000` (`jitter_ms=20000, max_jitter_ms=20000` in the drain log).
- `BUZZ_DRAIN_JITTER_MS=soon` → **hard startup error** (`must be a non-negative integer`), refuses to start.
- `BUZZ_DRAIN_JITTER_MS=""` → **boots with jitter off** (was a startup error before commit `46c690940`; now treated as unset, so a `""` kill switch no longer crashloops the relay).

### Automated (same commit)
`cargo fmt -p buzz-relay --check`, `cargo clippy -p buzz-relay --all-targets -- -D warnings`, and the full drain/config unit set (16 tests incl. the new `drain_all_jittered_cancels_when_flush_ack_times_out` and the empty-string config cases) all green. Push hooks (branch-skew, rust-tests, desktop-tauri-checks) passed.

### Not covered here
This did **not** run the Postgres-backed full `cargo test -p buzz-relay` integration suite, the reconnect-load A/B against a second relay (the "does the pool metric actually improve" proof), or a real Desktop/mobile client through a jittered restart. Those are the next tier from the reviewers' plans if we want the full herd-reduction proof before dial-up.

brow (dismissed):
🤖 Reviewed at head `46c690940`, base `857e63c4`. Verdict: approve, no blockers. I verified the mechanism against the code and against a real relay process, not against the description.

## The earlier blocker is genuinely closed

The prior review found that `drain_all_jittered` spawned detached timer tasks and returned at once, so the runtime drop killed every timer and no client got the 1012 frame. That is fixed structurally, not cosmetically. `drain_all_jittered` is now an `async fn` that builds per-connection futures and awaits them through `join_all`, so the caller owns the timers. `serve` keeps the join handle, and both the UDS path and the TCP-only path await it after `axum::serve` returns, then abort the hard-exit task.

I did not take that on faith. I re-cut the exact original defect as a mutant, replacing the `join_all` await with a drop of the pending set. It is killed by two independent assertions: the `poll!`-based is-pending assertion in the defer-window test, and the cancel assertion in the restart-channel-full test. So the regression now has real coverage that fails when the bug returns.

I also reproduced the end-to-end behavior myself with a real relay binary, real Postgres, and NIP-42-authenticated sockets, driven by an independent BIP340 signer I validated against the spec vector with tampered-signature and wrong-message negative controls. Relay exit code was 0 in every leg.

- Jitter off: 40/40 sockets received 1012, spread 1 ms. This confirms the herd the change targets.
- `jitter_ms=8000`: 40/40 received 1012, spread 7660 ms, well distributed.
- `jitter_ms=60000`: log shows the clamp to `jitter_ms=20000`, 40/40 received 1012, spread 19160 ms.
- Silent clients that never echo the close, `jitter_ms=2000`: 20/20 received 1012, and the relay still exited 0 with no hard-drain-timeout log.

Two design points I checked and agree with. The sticky `draining` flag is stored before the first await, so the store-then-snapshot pairing with `register`'s insert-then-check still holds and a late registration self-signals immediately with no jitter. And `drain_all`'s body is byte-identical to base (I hashed both sides), so the default jitter-off path carries no behavior risk.

Mutation-wise I cut ten further mutants across the sticky flag, the zero-delay case, the ack wait, both cancel fallbacks, the ack send, the 1012 code, the clamp, and the empty-string-is-unset rule. All ten are killed. `cargo fmt` and `cargo clippy -p buzz-relay --all-targets -- -D warnings` are clean at this head, and the 16 targeted drain tests plus the config test pass.

## Non-blocking notes

**1. The `serve` shutdown wiring has no test coverage.** This is the one I would most like to see addressed, though it does not block. `serve` is never called from a test, so I mutated its wiring directly and the suite stayed green three times: unwiring jitter so the jittered drain is never dispatched, dropping the await on the shutdown join handle (which is the previously shipped bug's shape, at the call site instead of inside the helper), and inverting the dispatch condition. Coverage of the drain helper is not coverage of its use, and the helper is exactly where the last defect was found. A single real-process or harness test that sends SIGTERM and asserts one client got 1012 would close the whole family. My live runs show the wiring is correct at this head, so this is a durability gap for the next editor, not a present defect.

**2. `BUZZ_DRAIN_JITTER_MS` is undocumented and unreachable in the chart.** `git grep` finds it only in `config.rs` and `state.rs`. It is absent from `TESTING.md`'s configuration table and absent from the Helm chart, so the feature cannot be turned on by values today. Worth adding before anyone tries to roll this out.

**3. Please document the shutdown time budget.** The pieces interact and the total is not obvious from any one file: a 5 s pre-drain grace, then a jitter window of up to `MAX_DRAIN_JITTER_MS` (20 s), against a 30 s `GRACEFUL_DRAIN_TIMEOUT`, all inside the chart's `terminationGracePeriodSeconds: 60`. At maximum jitter that is 5 + 20 = 25 s before the last close even begins, and a slow client can add up to `RESTART_CLOSE_ACK_TIMEOUT` (5 s) on top, which reaches the 30 s hard-exit boundary exactly. The constant's doc comment says the cap leaves ten seconds of headroom; that is true relative to the 20 s window alone, but not once the 5 s grace is included. The margin is real but thinner than the comment implies, so I would state the arithmetic explicitly. There is also no `preStop` hook in the chart, so the 5 s grace is the only thing decoupling listener close from endpoint deregistration.

**4. Minor: `join_all` switches to `FuturesOrdered` above 30 futures.** That is the right behavior for thousands of sockets and I see no problem. Noting it only so the allocation is a known property rather than a surprise.

brow (approved):
🤖 Re-reviewed the delta `46c690940..4a70d76ea`. Verdict: approve again. The delta is safe.

**Scope check.** 69 insertions, zero deletions, five files. I confirmed the Rust half is comments only: every added line in `main.rs` is a `///` or `//` line, and the four logic files from my first pass (`state.rs`, `connection.rs`, `config.rs`, `handlers/event.rs`) are blob-identical between the two heads. So my earlier mutation and live-process results carry forward unchanged. `fmt` clean, `clippy --all-targets -D warnings` clean, and the drain and config tests still pass at this head.

**The chart wiring works, and I verified it by rendering rather than by reading.** `helm template` emits `BUZZ_DRAIN_JITTER_MS` with the default `"0"`, passes `8000` and `60000` through as given (the relay clamps `60000` at runtime, which is the right split), and a values file that omits `relay.drainJitterMs` still renders `"0"`. The schema entry is enforced: a string value fails with `got string, want integer`, and `-5` fails with `minimum: got -5, want 0`. Worth noting for anyone reproducing this: a plain `helm template` in this chart fails first on missing subchart dependencies and then on several required values, and those failures look exactly like a clean negative if you only grep the output. I had to clear them before the render measured anything.

An explicit `drainJitterMs: null` renders an empty value, which the relay treats as unset and off, so it fails safe. That is also the existing pattern for sibling keys such as `BUZZ_SEND_BUFFER`, so it is house style and not something this change introduces.

**The intra-doc link resolves.** I checked this with a control rather than trusting the build: injecting a deliberately broken link into the same doc block raises the unresolved-link count from 25 to 26 and names the bad symbol, while `MAX_DRAIN_JITTER_MS` never appears as unresolved. The 25 pre-existing unresolved links and the `manifest_event.rs` unclosed-HTML-tag error are both present at the previous head, so neither belongs to this delta.

**One doc nit, not blocking.** The new budget comment ends by saying that with jitter off "the per-socket delay collapses to roughly the 5s grace plus the ack wait." The jitter-off path is `drain_all`, which does a `try_send` of the close frame and then cancels immediately, with no ack wait at all. So the jitter-off delay is just the 5s grace. The ack wait exists only on the jittered path. Everything else in the budget block matches the constants I verified: 5s grace, then 20s max jitter plus 5s ack inside the 30s hard drain, 35s worst case to forced exit, inside the chart's 60s `terminationGracePeriodSeconds`.

The `TODO(coverage)` comment is an accurate statement of the gap I found, including that dropping the `shutdown_handle` await is the old detached-timer bug relocated to the call site. I confirmed the proposed seam is feasible: the config module is already public to the binary, so a `run_graceful_shutdown` extraction needs no visibility changes.

PR #4546, opened 2026-08-03
fix(desktop): make relay-published agents mentionable

purybr365:
## Summary

An agent hosted outside Buzz Desktop — a systemd/BYO agent that publishes its own
kind:10100 directory entry — can never be `@`-mentioned. It stays out of mention
autocomplete no matter what the directory entry advertises (`respond_to: "anyone"`,
`"allowlist"` naming the user, matching `channel_ids`, channel membership, avatar,
valid NIP-OA attestation).

Two independent defects stack. **Either one alone is sufficient to hide the agent**,
which is why fixing just the frontend gate produces no visible change.

### 1. `RelayAgentInfo` crosses the Tauri boundary in the wrong casing

`list_relay_agents` returns `Vec<RelayAgentInfo>`, which had no
`#[serde(rename_all = "camelCase")]`, so it serialized snake_case. The TypeScript
`RelayAgent` type reads `agentType` / `channelIds` / `respondTo` / `respondToAllowlist`.
All four were therefore `undefined` on the frontend:

```ts
if (agent.respondTo === "allowlist" && ...)   // undefined — branch never taken
return agent.respondTo === "anyone" && ...    // undefined — always false
```

`relayAgentIsSharedWithUser` returned `false` for **every** relay agent, so
`getMentionableAgentPubkeys` never contained one. The failure is silent in both
directions: the `=== "anyone"` comparison short-circuits before
`agent.channelIds.some(...)`, so it reads as "not shared" rather than throwing, and
the TS type asserts a shape nothing verifies at runtime.

**No open PR addresses this**, which matters — see *Related* below.

### 2. `useMentions` gates on the managed set, not the invocable set

```ts
if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) return;
```

`managedAgentPubkeys` comes from `managedAgentsQuery.data` — agents this desktop
spawns as subprocesses. Nothing published to a relay can enter that set, so any
candidate with `isAgent === true` that isn't locally managed was dropped here,
*before* `shouldHideAgentFromMentions` — the directory-aware policy that understands
invocability and membership — ever ran.

`isAgent` goes true for a headless agent via any of three client-side signals, so
this can't be worked around from the relay while keeping the agent identity intact:
the kind:0 NIP-OA `auth` tag, a channel role of `bot`, or simply having a kind:10100
entry (`relayAgentNamesByPubkey`).

### Implementation notes

- `RelayAgentInfo` now serializes camelCase and keeps per-field snake_case `alias`es,
  because kind:10100 event content *is* snake_case (`agents_from_events`). Both
  directions are pinned by tests.
- `types.rs` sat exactly at the 1000-line ratchet limit, so `RelayAgentInfo` moves to
  `types/relay_agent_info.rs` beside the existing `catalog_source` and `requests`
  submodules rather than growing the file.
- Dropping `managedAgentPubkeys` from the `useMemo` dependency list is required by
  `lint/correctness/useExhaustiveDependencies` once its last in-memo use is gone.
- Adds a Common Gotchas entry for the Tauri-boundary casing trap.

## Related issue

Searched before opening — this is a well-reported bug with **no** open PR covering
defect 1:

- #2987 — external agents publishing kind:10100 become un-mentionable;
  `shouldHideAgentFromMentions`'s invocability logic unreachable *(closest match)*
- #2603 — managed-list gate precedes directory eligibility
- #3971 — shared relay agent eligibility is unreachable
- #4128 — shared relay agents never appear in mention autocomplete (mobile works)
- #4548 — non-owner members never get agents in mention autocomplete despite correct
  kind:10100 records

**Overlapping PRs — deliberate, not an oversight.** #4453, #4517 and #4536 each fix
defect 2 (#4453 most thoroughly: it renames the gate to `isAgentAutocompleteEligible`
and also covers `MembersSidebar`). None of them touch the Tauri boundary, so on a real
relay `mentionableAgentPubkeys` stays empty and **all three remain inert** — the gate is
widened to admit a set that `relayAgentIsSharedWithUser` can never populate. That's
worth flagging on its own.

Happy to rebase and drop my defect-2 change in favour of #4453 if it lands first — the
serde fix is the part nothing else covers, and it's what makes the others work.

## Testing

`just desktop-check`, `just desktop-typecheck`, `just desktop-test` (4005 tests),
`just desktop-build` and `just desktop-tauri-fmt-check` all pass locally.

New coverage:

- `RelayAgentInfo` serializes camelCase with no snake_case leftovers, and still
  deserializes snake_case directory content.
- The two gates composed the way `useMentions` composes them: a shared relay agent
  survives without being locally managed (both `anyone` and `allowlist`), an agent
  sharing no channel is still dropped, and locally managed agents keep working.
- A guard on the call site itself — the composition tests invoke the gates directly, so
  they would not catch gate 1 silently narrowing back to the managed set. Verified this
  guard fails on the unfixed code and passes on the fixed code.

**Not run locally:** `desktop-tauri-check` / `desktop-tauri-test`, because the machine
available has no GTK/WebKit development libraries. The serde behaviour those tests
assert was instead verified against a standalone harness using the real kind:10100
payload this relay serves, confirming camelCase output and snake_case input both work.
CI covers the crate build.

**No screenshots.** This changes which entries appear in existing autocomplete, with no
visual or layout change. A mock-mode capture would also misrepresent it: the e2e bridge
replaces Tauri IPC, so defect 1 — the serialization boundary — cannot occur under it,
and a "before" shot would show the bug already absent.

Reproduced against a self-hosted relay with two systemd agents holding valid
attestations, correct kind:10100 entries (`respond_to: "anyone"` and `"allowlist"`
naming the owner) and channel membership. Neither appeared in autocomplete. The only
workaround on a stock build is to strip the attestation *and* withdraw the directory
entry so `isAgent` goes false and the agent is treated as an ordinary member — which
costs the agent badge, the agent profile panels, and agent-vs-human classification by
other agents (`buzz-acp`'s `profile_event_is_agent`, which affects reply-thread
anchoring).


rmichelena:

This PR is the closest of the open attempts — it's the only one that reaches into the Rust side. But it changes the `RelayAgentInfo` types, not the **query**: `list_relay_agents` (`desktop/src-tauri/src/commands/agent_discovery.rs`) still fetches **only `kind:10100`**. Agents that publish their policy as `kind:30177` and their channel membership as `kind:39002` therefore still arrive with `respond_to: null` and `channel_ids: []`, and stay hidden from `@` autocomplete even for their own owner.

We opened #5483 with the query-side fix: merge `kind:30177` (policy, owner-verified via `kind:0` so it can't be spoofed) and `kind:39002` (NIP-29 membership) into `list_relay_agents`, best-effort so a relay error degrades to 10100-only instead of emptying autocomplete. E2e-verified on a self-hosted relay.

Not a claim that this PR is wrong — the layers are compatible. Full analysis in #5363.

rmichelena:
Flagging a review finding rather than a duplicate claim — I referenced this PR in #5483 and then read it properly, and I think defect #1 no longer holds on `main`, in a way that makes this patch a regression rather than a fix.

`list_relay_agents`' payload is snake_case **by contract**; the camelCasing happens in the TS mapper, not in serde. On current `main`, `desktop/src/shared/api/tauri.ts`:

```ts
type RawRelayAgent = {
  agent_type: string;
  channel_ids: string[];
  respond_to?: RelayAgent["respondTo"];
  respond_to_allowlist?: string[];
  …
};

function fromRawRelayAgent(agent: RawRelayAgent): RelayAgent {
  return {
    agentType: agent.agent_type,
    channelIds: agent.channel_ids ?? [],
    respondTo: agent.respond_to ?? null,
    respondToAllowlist: agent.respond_to_allowlist ?? [],
    …
  };
}
```

and `listRelayAgents()` maps every element through it. So the frontend never reads `agent.respondTo` off the raw Tauri payload — `relayAgentIsSharedWithUser` receives the *mapped* `RelayAgent`, where `respondTo` is populated whenever the 10100 content carried it.

Adding `#[serde(rename_all = "camelCase")]` to `RelayAgentInfo` without touching `RawRelayAgent` / `fromRawRelayAgent` (this PR changes neither — the file list is `AGENTS.md`, `types.rs`, `types/relay_agent_info.rs`, `types/tests.rs`, `agentAutocompleteEligibility.test.mjs`, `useMentions.ts`) would make the mapper read `agent.respond_to` / `agent.channel_ids` / `agent.agent_type` as `undefined` for **every** relay agent — producing exactly the "all four undefined, silently falsy" failure this PR is written to fix. The Rust-side round-trip tests you added would still pass, because they never cross the mapper.

Defect #2 (`isAgentIdentityInManagedList` gating on the managed set) was real, but `main` has gated on the mentionable set since #4913 — the function this PR patches doesn't exist on `main` anymore.

The genuinely useful thing here, in my view, is the observation that **nothing pins the Rust→TS casing contract for this payload** — no test would have caught either direction of getting it wrong. That's worth keeping, whatever happens to the rest.

For the underlying symptom, our repro traced it below the eligibility layer: `list_relay_agents` queries `kind:10100` only, while `respond_to` lives in `kind:30177` and membership in `kind:39002`, so agents publishing the dedicated events arrive `respond_to: null` / `channel_ids: []` — fix in #5483. Given that, would you consider closing this one, or reducing it to the casing-contract test? Maintainers' call either way, and I'd genuinely rather be told I've misread your patch than have you close a working fix on my say-so.

PR #4554, opened 2026-08-03
fix(relay): move fleet-wide NIP-43 membership sweep off the pre-bind startup path

tlongwell-block:
## Problem: silent permanent crashloop after a restart ("boot wedge")

Startup awaits `reconcile_nip43_membership_snapshots` — a sequential sweep over **every** provisioned community — before binding the listener. Per community it runs a snapshot-reconciliation check (2 queries) plus a repair publication when needed.

Measured locally (isolated docker rig): **~9.5ms/community** on the repair path; 1001 communities ≈ 9.5s of publishes, all logged *before* `buzz-relay TCP listening`. The cost is linear in community count, so on a deployment with a large number of communities the pre-bind sweep takes minutes — far past a typical startup probe window (e.g. 120s). The probe then SIGKILLs the pod mid-sweep, the restart begins again at community #1, and the pod crashloops forever with the port never bound and zero log lines at `RUST_LOG=error` (a healthy boot at that level logs nothing until failure, so the loop is silent).

## Fix (deliberately minimal — emergency scope)

- **Pre-bind:** reconcile **only the deployment community** (one bounded snapshot check), with explicit phase logs carrying `elapsed_ms` so the startup sequence is observable even when it's healthy.
- **Fleet sweep moves entirely post-bind** into the existing periodic task, now:
  - **jittered** — first tick delayed by a random fraction of the interval (same rationale and mechanism as the usage-metrics poller: PID-derived seeds are unsafe when every pod is PID 1);
  - **leader-gated** — session advisory lock (`NIP43_SWEEP_LOCK_KEY`, reusing `try_lock_usage_metrics`) so N replicas never run N concurrent fleet-wide sweeps. The lock is held only for the sweep's duration; a replica dying mid-sweep releases it with its session.
  - **skip-tick** on overrun (`MissedTickBehavior::Skip`) — a sweep longer than the interval doesn't queue a catch-up burst.
- **Sweep telemetry:** start / progress (every 1000 communities, with `scanned`/`total`/`reconciled`/`elapsed_ms`) / complete / fail logs, plus a `buzz_nip43_membership_sweep_seconds` histogram.

Pagination/resume and per-query lock/statement deadlines are a **separate follow-up PR** by agreed scope.

## Verification (live-local)

Rig: docker compose stack, postgres seeded with **1001 communities**, all 1001 kind:13534 snapshot events deleted to force the worst-case repair path. Binary built from this branch, `cargo build --release -p buzz-relay`.

**Run 1 (1001 repairs pending):**

```
15:48:44.461 NIP-43 startup phase: reconciling deployment community snapshot
15:48:44.474 NIP-43 startup phase: deployment community snapshot reconciled repaired=true elapsed_ms=12
15:48:44.475 buzz-relay TCP listening addr=127.0.0.1:33000        <- bind at +14ms, not +9.5s
15:48:46.478 NIP-43 membership sweep starting (leader)
15:48:50.024 NIP-43 membership sweep complete: snapshots repaired count=1000 elapsed_ms=3546
```

Bind happens **12ms** after the deployment-community check; the other 1000 repairs happen post-bind under the lock. DB confirms all 1001 snapshots restored.

**Run 2 (restart, converged state — idempotence):**

```
15:49:19.054 ... deployment community snapshot reconciled repaired=false elapsed_ms=7
15:49:19.055 buzz-relay TCP listening
15:49:20.330 NIP-43 membership sweep complete: nothing to repair elapsed_ms=1273
```

**Tests:** `cargo test -p buzz-relay` — 837 passed, 1 failed: `api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo`, which fails identically at the base commit (`origin/main` 5e0efb0bb) with this diff stashed — pre-existing, not introduced here. `just test-unit` (the pre-push gate) green. Clippy clean.

### DIFF
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -2872,39 +2872,53 @@ async fn emit_initial_ref_state(
     Ok(())
 }
 
+/// Reconcile one community's event-backed NIP-43 membership view.
+///
+/// `relay_members` is canonical. The snapshot is rebuilt only when it is
+/// absent or its member/role set differs from the canonical rows, so this is
+/// cheap (two queries) when nothing changed. Returns whether a repair
+/// publication happened.
+pub async fn reconcile_nip43_membership_snapshot(
+    tenant: &TenantContext,
+    state: &Arc<AppState>,
+) -> anyhow::Result<bool> {
+    if !state
+        .db
+        .nip43_membership_snapshot_needs_reconciliation(
+            tenant.community(),
+            &state.relay_keypair.public_key(),
+        )
+        .await?
+    {
+        return Ok(false);
+    }
+
+    publish_nip43_membership_list(tenant, state).await?;
+    Ok(true)
+}
+
 /// Reconcile every community's event-backed NIP-43 membership view.
 ///
-/// `relay_members` is canonical. A snapshot is rebuilt only when it is absent
-/// or its member/role set differs from the canonical rows. This makes the sweep
-/// safe to run at startup and periodically without producing an event stream
-/// when nothing changed. A failure in one community is logged and counted but
-/// does not prevent the remaining communities from being repaired.
+/// A failure in one community is logged and counted but does not prevent the
+/// remaining communities from being repaired.
+///
+/// This sweep is O(communities) — sequential, two queries each, plus a
+/// publication per repair. On a large deployment it takes minutes, so it MUST
+/// NOT run on the pre-bind startup path (a startup probe SIGKILLs the pod
+/// mid-sweep and the restart begins again at community #1, forever). It runs
+/// post-bind, jittered, and leader-gated — see the sweep task in `main.rs`.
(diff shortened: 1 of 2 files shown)


Samuel7192 (approved):
Reviewed the post-bind ordering, advisory-lock lifetime, skipped-tick behavior, and failure handling. The change structurally removes the fleet sweep from the pre-bind path and preserves per-community reconciliation. CI coverage includes the listener-bound gate and the existing relay suite.

Samuel7192:
Production impact confirmed on `postral.communities.buzz.xyz`: Desktop/ACP clients saw connection closures, transient database errors, and `404 relay: no community is configured for this host` during the relay restart from 17:26–17:36 UTC. The host has recovered (30/30 NIP-11 `200`, 20/20 WebSocket `101`, 20/20 `/events` requests reached tenant auth rather than host-mapping `404`), but the pre-bind sweep remains a restart hazard until this lands. All CI is green; requesting the required Block-member approval from @block/buzz-oss-team.

Issue #3493, opened 2026-07-29
Agent config panel lists MCP servers from hardcoded ~/.claude.json, ignoring per-agent CLAUDE_CONFIG_DIR


oshnilia:
**Describe the bug**

When a managed agent is given its own `CLAUDE_CONFIG_DIR` (via Agents → Agent defaults, or per-agent env vars) in order to isolate it from the operator's personal Claude Code configuration, the agent config panel still lists the MCP servers from the operator's `~/.claude.json` and labels them "mcp enabled". The list is identical for every agent regardless of that agent's environment.

The runtime itself *is* isolated — only the panel is wrong. But the panel is the only place an operator can check, so it reads as "the isolation did not work", which is alarming when the personal config contains mail/drive/database connectors.

**Steps to reproduce**

1. Have one or more MCP servers configured at user scope in `~/.claude.json`.
2. Create an isolated config dir, e.g. `<nest>/claude-config`, containing `.claude.json` with `{"mcpServers": {}}` and `settings.json` with `{"disableClaudeAiConnectors": true}`; authenticate it once with `CLAUDE_CONFIG_DIR=<nest>/claude-config claude auth login`.
3. Set `CLAUDE_CONFIG_DIR=<nest>/claude-config` in Agent defaults (or per-agent env vars) and restart the agents.
4. Open the config panel for any agent.

**Expected behavior**

The MCP Servers section reflects the configuration the agent will actually load — empty for an isolated agent — and the "From config file (…)" attribution points at that file.

**Actual behavior**

The section lists every server from the operator's `~/.claude.json`.

Independent checks confirming the runtime is genuinely isolated:

- `ps eww` on `buzz-acp` and on its `claude-agent-acp` child both show `CLAUDE_CONFIG_DIR=<nest>/claude-config`
- `CLAUDE_CONFIG_DIR=<nest>/claude-config claude mcp list` → `No MCP servers configured`
- asking the agent to list tools named `mcp__*` → none

**Cause**

`desktop/src-tauri/src/managed_agents/config_bridge/claude.rs`:

```rust
pub(super) fn read_config_file() -> Option<RuntimeFileConfig> {
    let home = dirs::home_dir()?;
    let settings_path = home.join(".claude").join("settings.json");
    let mcp_path = home.join(".claude.json");
```

The path is derived from `$HOME` unconditionally; the agent's effective environment (global / persona / per-agent `env_vars`) is never consulted. The same applies to the attribution path in `config_bridge/reader.rs`:

[...]

Issue #2265, opened 2026-07-21
Claude agents fail silently when default model requires usage credits (Fable 5)


thanveerdev:
### Summary

Default agents (e.g. Bumble) appear online and receive `@mentions`, but never reply in-channel when Claude Code’s default model is set to **Fable 5** (`claude-fable-5[1m]`), which requires usage credits on a Claude Pro subscription.

The harness logs a clear error and requeues the turn; the Desktop UI and channel stay silent. From the user’s perspective the agent is “broken.”

### Environment

- Buzz Desktop **v0.4.21**
- Runtime: **Claude Code** via `claude-agent-acp`
- Claude auth: **claude.ai Pro** (`claude auth status` → `loggedIn: true`, `subscriptionType: pro`)
- Relay: hosted community relay (e.g. `wss://*.communities.buzz.xyz`)
- macOS arm64

### Steps to reproduce

1. Install / use Claude Code with a Pro login.
2. Set Claude default model to Fable 5 (or leave it if Claude already chose it), e.g. in `~/.claude/settings.json`:
   ```json
   { "model": "claude-fable-5[1m]" }
   ```
3. Start a managed Buzz agent with `agent_command` / override `claude-agent-acp` (welcome team / default agents is enough).
4. Confirm agent presence is online in a channel the agent is a member of.
5. Send `@Bumble hello` (or any mention) from the owner account.
6. Observe: no reply in the channel; no in-app error surface about credits/model.

### Expected

- Agent replies, **or**
- Agent/Desktop posts a visible failure in-channel or in the Agents UI, e.g.  
  “Claude model Fable 5 requires usage credits — switch model or run `/usage-credits`.”

### Actual

- Mention is received and dispatched by `buzz-acp`.
- Claude ACP session starts, then fails:
  ```
  Fable 5 requires usage credits. Run /usage-credits to continue or switch models with /model.
  session_prompt error: Agent reported error (code -32603): Internal error: Fable 5 requires usage credits...
  requeueing failed batch with backoff
  agent_returned (application error — pipe intact) outcome=error
  ```
- Channel stays empty; user only sees silence.

### Workaround

Change Claude default model to a plan-included model:

```json
// ~/.claude/settings.json
{ "model": "sonnet" }
```

Restart agents. After that, `@Bumble` replies normally (verified: agent pubkey replied `pong` in `#general`).

### Related / secondary bug

[...]

PR #4557, opened 2026-08-03
feat(managed-agents): close five Claude Code agent-config gaps

### LINKED ISSUE #2692: [Bug] Claude Code agents ignore the model set in Buzz Desktop — model picker persists but nothing applies it at spawn

wpfleger96:
Closes five Claude Code agent-config gaps in Buzz Desktop, split into three commits that share the spawn-time and live-switch surfaces.

## Config isolation, model authority, and Auto mode (#2692, #2884, #3493)

- **`CLAUDE_CONFIG_DIR` isolation (#3493).** `config_bridge` resolves both `settings.json` and `.claude.json` panel paths from the agent's effective env (`resolve_effective_agent_env` — baked floor → definition → global → persona → record), so the panel reads the same directory the agent runs against. `mcp_config_file_path_for_runtime` honors a custom dir; empty/blank is treated as unset, matching Claude's `CLAUDE_CONFIG_DIR || homedir()` semantics. `AgentConfigPanel` shows a Keychain caveat when a custom dir is active (a custom dir maps to a fresh Keychain namespace unless `CLAUDE_SECURESTORAGE_CONFIG_DIR` is also set).
- **Single startup model authority (#2692).** `ANTHROPIC_MODEL` is the sole startup model authority for Claude. Local spawns write `ANTHROPIC_MODEL` and strip `BUZZ_ACP_MODEL` so the harness never sees two authorities; remote deploys send `ANTHROPIC_MODEL` in `policy_env` instead of `BUZZ_ACP_MODEL`. Non-Claude runtimes are unchanged.
- **`PermissionMode::Auto` (#2884).** Wire string `"auto"`, model-gated, degrades to the agent default when the active model doesn't advertise it.

## Thinking effort end-to-end for local Claude agents

Effort flows from the running session's discovered `thought_level` config option through the config surface to a local-only write control and a read-only two-facts display.

- **Reader.** Discovers the `thought_level` config option from the session cache (never hardcoded) and populates `effort_config_id` / `effort_options` on `RuntimeConfigSurface`. The canonical effort tier orders record env > `record.effort_level` (`BuzzExplicit`) > ACP > persona > global > definition > file, so the panel shows the effort the next spawn will launch with while `resolve_with_override` surfaces the running ACP value as the struck-through override — neither masks the other silently.
- **Write control.** `persist_agent_effort_level` is a direct-write Tauri setter (writes `record.effort_level` + `updated_at`, `save_managed_agents`) that rejects non-local backends — remote effort is set at deploy time via `policy_env`. `EffortPickerField` mounts in `AgentInstanceEditDialog` beside the Model block, gated on a local backend **AND** a discovered `effortConfigId`. It persists directly and invalidates the config surface, mirroring the `setManagedAgentAutoRestart` standalone-setter precedent, so the frozen `UpdateManagedAgentInput` shape stays frozen.
- **Display.** The read-only configured-vs-current two-facts display is the `thinkingEffort` normalized field in `AgentConfigPanel`, fed by the reader's canonical tier ordering.
- `buzz-acp` applies the startup effort env at session start.

## Distinguish a rejected model switch from silent success

A live model switch collapsed adapter rejection into success: `apply_model_switch` returned `Ok` on both an accepted switch and an application-level refusal, so the caller cached pre-switch capabilities as if they described the target model and Desktop reported the pick as landed.

- `ModelSwitchOutcome::{Applied(Value),Rejected}`. Transport-class errors still propagate as `Err` (respawn the poisoned stdio); an application-level refusal is now `Rejected`.
- The caller drives everything off `post_switch_snapshot`: `Applied` refreshes `model_capabilities` from the target model's echoed `configOptions` (or drops to `None` when none are echoed, so a pre-switch snapshot is never mistaken for the target's); `Rejected` preserves pre-switch caps and emits a `failure` `control_result`. Effort resolution and the session-config capture read the post-switch snapshot so they converge on the model the session actually runs; `modelOverridden` is gated on `switch_succeeded`.
- `liveSwitchOutcome.ts` gains a distinct `"failed"` outcome for the adapter `failure` frame and treats the busy-path `"sent"` ack as provisional — it never counts toward success. Success is confirmed only by a real positive terminal frame (the busy-path deferred apply emits a correlated `switched` `control_result` when it lands), and the fallback timeout resolves an honest `"pending"` (accepted, apply deferred), never a false `"ok"`. `ModelPicker` surfaces a distinct toast per outcome — failed, unsupported, and pending.

## Scope explicitly excluded

Per-agent config dir provisioning, `CLAUDE_SECURESTORAGE_CONFIG_DIR` sentinel injection, `settings.json` projection, protected-key stripping, MCP inheritance, spawn serialization, and the `last_spawn_warnings` surface are absent from this diff. Silent-fallback machinery for non-Claude runtimes (#2265/#4004) is a tracked follow-up.

## Sanctioned follow-ups

- **Live mid-conversation effort switching.** Effort is spawn-scoped only in this PR: the worker reads `BUZZ_ACP_EFFORT_LEVEL` once and applies it at session creation. The live effort-switch machinery (mid-conversation effort RPC + ack frame) was deliberately removed and is archived on `archive/claude-config-gaps-live-effort` for a future plan-gated revival.
- **Idle-path late model-switch rejection is unobservable.** An idle switch acks `switched` immediately after catalog validation, but the real `set_config_option` runs at the next session creation — potentially much later — so a rejection there is not surfaced back to the picker (holding a subscription that long is not sensible). Pre-existing, catalog-gated behavior; a durable fix is a tracked follow-up.

Closes #2692, #2884, #3493


kalvinnchau (changes requested):
[P2] Include the request nonce in `invalid_value` acknowledgements — `crates/buzz-acp/src/lib.rs:1320`

The validation-rejection branch returns before the common acknowledgement builder adds `nonce`. Desktop always supplies a nonce and `awaitEffortOutcome` rejects every acknowledgement whose nonce does not match, so a genuinely invalid selection is ignored until the 8-second timeout and then presented as “Applies at next session,” even though the pool was deliberately left unchanged. Add the incoming nonce to this early acknowledgement (and cover the real nonce-bearing invalid-value path rather than only nonce-free fixtures).

[P2] Seed `committed_effort` when the startup effort is successfully applied — `crates/buzz-acp/src/pool.rs:791`

A persisted startup effort is applied by `create_session_and_apply_model`, but its checkout has `desired_effort_gen == None`; `return_agent` copies it into `desired_effort` and never records it in `committed_effort`. If the user subsequently selects another valid effort and the adapter rejects/times out, the failure path rolls `desired_effort` back to `committed_effort`, which is still `None`, instead of restoring the previously confirmed startup value. Record a successful startup application as the committed baseline (or initialize the pool baseline from the persisted value once capabilities confirm it), and add a startup-success → live-pick-failure regression.

[P2] Serialize generation resolution across parallel workers — `crates/buzz-acp/src/pool.rs:805`

Every worker checked out for the same generation independently emits a terminal acknowledgement and independently commits or rolls back the shared pool state. With parallelism >1, if worker B fails and returns first it rolls `desired_effort` back; if worker A then returns success, it updates only `committed_effort` and leaves `desired_effort` rolled back. The observer can persist A's `ok`, while future checkouts receive the old/default value. The reverse return order produces a different result, so pool state is scheduler-dependent. Resolve each generation once with explicit aggregate semantics (and emit one terminal outcome), or ensure a later same-generation success restores `desired_effort` consistently; cover both return orders with two workers.

Validation at exact PR HEAD `b7c99fb94808cf3b07bb7540121d8eb4513578dd`: `cargo test -p buzz-acp --lib` passed (723 tests); `node --test src/features/agents/lib/effortOutcome.test.mjs` passed (15 tests). These paths are not covered by the current suite.

kalvinnchau (changes requested):
[P2] Do not treat a switch that never landed as success — `desktop/src/features/agents/lib/liveSwitchOutcome.ts:116`

Every status other than `unsupported_model`, `failure`, and `sent` falls through to the positive-terminal counter. That includes `turn_ending` and `no_active_turn`, but the producer explicitly emits those when it could not deliver the switch: `turn_ending` means the control oneshot was already consumed, and `no_active_turn` means neither an in-flight task nor an idle session-owning agent could be found (`crates/buzz-acp/src/lib.rs:1355-1377`). This race is reachable when `activeTurns` is stale between the picker snapshot and harness receipt, or when another control signal is already ending the turn. The picker then shows “Model switched for this session” even though `desired_model` was never set and no later apply will occur; the added test at `liveSwitchOutcome.test.mjs:87-100` currently codifies that false success for `turn_ending`.

Handle statuses exhaustively. Count only `switched` as an applied terminal; keep `sent` provisional, and map `turn_ending` / `no_active_turn` to an honest non-success outcome (or retry/persist through a path that actually applies the requested model). Add regression rows for both producer statuses.

kalvinnchau (dismissed):
Re-reviewed exact head `e7049b4b6f7c9404205db7a0cb86677aaeaf477c` against merge base `439c03749182495ee09f85a73423dd17e7ccda61`.

The prior P2 is resolved. `liveSwitchOutcome.ts` now treats only `switched` as applied success, leaves `sent` provisional, maps `turn_ending` and `no_active_turn` to `not_delivered`, and ignores unknown statuses. Regression coverage exercises both non-delivery statuses and future unknown statuses. The subsequent huddle import consolidation is semantics-preserving.

No additional findings. Validation on the exact head: `cargo test -p buzz-acp` passed (816 library + 9 lifecycle tests); desktop `pnpm test` passed (5,013 tests); `git diff --check` clean. GitHub CI remained in progress at review time.

kalvinnchau (approved):

Issue #3822, opened 2026-07-30
RFC: explicit channel-scoped cwd for buzz-acp on local and remote runtimes


rsaulo:
**Motivation**

Managed agents currently have one process-wide working directory, normally the Buzz nest or an optional static workspace associated with the agent.

This does not work well when the same agent identity participates in channels for different repositories. Repository-specific instructions, skills, tools, and memory routing depend on the actual working directory being selected before the ACP adapter and session start.

The problem is not limited to agents managed by Buzz Desktop. A `buzz-acp` runtime may run on a laptop, a remote host, a container, or under another supervisor. Regardless of where the agent is created, the runtime needs a supported way to provide an explicit working directory instead of inheriting or assuming the Buzz nest.

A concrete use case is keeping a small set of stable managed identities, such as Claude, Codex, and OpenCode, and reusing them across many project channels. Creating one managed-agent identity per repository causes identity and configuration sprawl.

**Proposed solution**

Make `cwd` a first-class input to `buzz-acp`, usable by both Desktop-managed and independently supervised local or remote runtimes, and allow a channel or future project entity to bind to a logical repository or workspace identity.

The runtime executing `buzz-acp` should resolve that logical identity to an absolute path available in its own environment before creating or resuming the ACP session for that channel. For example, the same repository could resolve to `/Users/alice/src/project` on a laptop and `/srv/agents/repos/project` on a remote worker.

Expected properties:

- `buzz-acp` can always be started or instructed with an explicit workspace cwd, without depending on Buzz Desktop internals.
- One managed-agent identity can participate in channels backed by different workspaces.
- Each runtime resolves the portable workspace identity to its own local absolute path.
- The selected workspace becomes the real cwd before adapter and session initialization.
- Each `session/new` receives the resolved cwd rather than a process default or prompt-only hint.
- Implementations may use workspace-scoped workers keyed by `(agent, workspace)` when adapters require process-level cwd or initialization.
- Absolute filesystem paths remain local to the runtime and are never stored on the relay.
- The relay may store a portable repository identity, such as a normalized Git remote.
- Rebinding a channel invalidates its previous session.
- Missing, ambiguous, or deleted workspace mappings fail clearly instead of silently falling back to the Buzz nest.
- Existing Activity and agent identity remain consistent across workspace-scoped workers.
- Buzz does not automatically create, delete, or manage Git worktrees.
- Worker and session lifecycle remains bounded, considering the resource-growth problem described in #2961.

[...]

Issue #3148, opened 2026-07-27
[Bug] macOS: Claude Code agents spawn with cwd=/ each session pins a CPU core on a full-filesystem ripgrep scan


alexandernacho:
**Describe the bug**

On macOS, Buzz spawns Claude Code agent sessions with **no working directory set**, so each one inherits launchd's cwd of `/`. Claude Code then kicks off its file-index scan against the filesystem root:

```
rg --files --hidden /
```

The `rg` child blocks writing into a full pipe while the `claude` (node) parent pegs a full CPU core consuming the stream. The walk never usefully completes, so the session never gets to real work — it just burns a core for as long as it lives.

Buzz keeps adding sessions while it's open, so the cost is cumulative: on a 14-core M4 Max I had **3 cores permanently pinned** within ~6 minutes of launching the app, with audible sustained fan noise.

**Evidence**

Three `claude` processes, each having spent ~98% of its entire lifetime at ~99% CPU:

```
  PID  PPID      TIME  %CPU ELAPSED COMM
 7872     1   7:57.24  99.0   08:08 claude
 8007     1   7:57.16  98.5   08:04 claude
20482     1   5:44.67  99.5   05:49 claude
```

Every one of them has cwd `/`:

```
$ lsof -a -p 7872 -d cwd
node  7872  <user>  cwd  DIR  1,16  704  2  /
```

Each has an `rg` child scanning root, which has burned only 9–16 *seconds* of CPU — i.e. it is blocked, not working, while the parent spins:

```
 8327  7872   0:16.50   0.0  rg --files --hidden /
 8328  8007   0:09.35   0.0  rg --files --hidden /
26639 20482   0:14.36   0.0  rg --files --hidden /
```

**Attribution.** The spawned processes show `ppid 1` (Buzz detaches them), so the parent link isn't visible in the process tree. The correlation is nonetheless consistent:

- Start order tracks the Buzz launch. In a single `ps` snapshot: `buzz-desktop` elapsed `06:01`, then `claude` at `05:59` (+2s) and `05:55` (+6s). PID order (7749 → 7872 → 8007) matches. A third appeared ~2m20s later.
- Claude Code keys its session cache by cwd. The directory `~/Library/Caches/claude-cli-nodejs/-` (the encoding for cwd `/`) was created **2 minutes after** `~/.buzz` was created on install — so this has occurred on every Buzz run since install, not as a one-off.
[...]

Issue #2961, opened 2026-07-26
[Bug] Per-channel ACP sessions are never released: on Windows each live session pins a full agent process tree (~33 procs / ~2 GB), exhausting the commit limit until Windows kills the WebView2 host and the desktop silently renders black


Takenori-Kusaka:
## Summary

On Windows, a Buzz Desktop instance running 8 managed agents (each with `parallelism: 1`) grew to **~140 agent-owned OS processes and ~36 GB of commit charge over ~6 hours**, exhausted the system commit limit, and the OS reclaimed memory by terminating **Buzz's own WebView2 host process**. The Tauri window survived with no renderer, so the app rendered as a solid black window.

The failure is silent: `buzz-desktop.exe` keeps running, the process stays responsive, and **nothing is written to the app log, the agent logs, or stdout/stderr**. There is no in-app indication that the UI is dead or why.

The growth is not an orphan/zombie leak — the process tree is well-formed and a graceful quit reclaims everything. The driver is that **ACP sessions are keyed per channel and are never released for the lifetime of the worker**, and with `claude-agent-acp` every live session materialises as a full agent CLI process tree. Session count therefore grows toward `agents × channels`, and memory grows linearly with it, with no cap, no idle eviction, and no visibility.

This is a different mechanism from #2631 (that one is `DEFAULT_AGENT_PARALLELISM = 24` + eager pool init on manual start). Here parallelism is **1** on every running agent and the pool starts lazily via the restore path — the multiplication comes from per-channel sessions instead. The two compound.

## Environment

- Buzz Desktop **0.4.26**, stock official build, self-hosted relay at `127.0.0.1:3000`
- Windows 11 Pro 26200, 32 GB RAM, commit limit 73 GB (system-managed pagefile)
- WebView2 Runtime 150.0.4078.83
- Agent runtime: `@agentclientprotocol/claude-agent-acp` (system npm), `agent_command: "claude-agent-acp"`
- 8 managed agents with `start_on_app_launch: true`, all `parallelism: 1`, `idle_timeout_seconds: null`
- Each agent subscribes to 2–6 channels (`discovered N channel(s)` in the agent logs); ~30 (agent, channel) pairs in total

**Binary provenance — verified, no local patches involved.** Every executable in the install directory is byte-identical to the ones inside the official `Buzz_0.4.26_x64-setup_alpha-unsigned.exe` release asset. I extracted the published installer and compared SHA-256:

| file | SHA-256 (first 16) | official release asset | installed |
[...]

PR #4558, opened 2026-08-03
feat(acp): honor AGENT_CWD when resolving NewSessionRequest.cwd

rsaulo:
Implements the `$AGENT_CWD` resolution order that `crates/buzz-persona/PERSONA_PACK_SPEC.md` already specifies. Refs #3822.

## The gap

The spec, § `$AGENT_CWD` Definition:

> However, operators can control this value by setting the `AGENT_CWD` environment variable on the buzz-acp process. buzz-acp determines what value to pass as `NewSessionRequest.cwd` in this order:
> 1. The `AGENT_CWD` environment variable on the buzz-acp process, if set.
> 2. `std::env::current_dir()` as a fallback.
> 3. If both fail, buzz-acp logs an error and **refuses to start**.

Before this change, `grep -rn "AGENT_CWD" --include="*.rs"` returned **zero hits** across the repo — the identifier existed only in that markdown file. Only step 2 was implemented. An operator who followed the spec to point a supervised runtime at a workspace got the process directory instead, with no warning.

That matters because the session cwd is where the agent runtime resolves `AGENTS.md`/`CLAUDE.md`, `.claude/skills`, hooks and settings at bootstrap. An agent given the wrong cwd can still *read* files elsewhere, but it starts without any of the target repository's instructions — the user ends up re-explaining in chat what the repo already documents.

## What changed

- `config.rs` — `resolve_agent_cwd()` plus a pure, injectable `resolve_agent_cwd_with()` and an `AgentCwdError` enum.
- `lib.rs` — both call sites that resolved the session cwd now use it: the long-running `PromptContext` and the `models` subcommand.
- `README.md` — documents `AGENT_CWD` in the Core table.

Two judgement calls worth reviewing:

**A set-but-unusable `AGENT_CWD` is an error, not a fall-through to step 2.** Falling back would hand the agent the process directory while the operator believes the workspace is pinned — exactly the silent failure this order exists to prevent, and what #3822 and its commenters ask to avoid. A blank value *does* still read as unset, since `AGENT_CWD=${WORKSPACE}` with an unset `WORKSPACE` expands to `""` in every shell idiom that builds it.

**Step 3 replaces `unwrap_or_else(|_| PathBuf::from("/"))`.** A cwd of `"/"` is dropped by `workspace_section()`, so protocol-v1 harnesses lost the absolute-cwd anchor entirely and scanned from the filesystem root — the behavior reported in #3148. Refusing to start surfaces the condition instead of degrading into a full-disk ripgrep. This is the only behavior change for existing deployments, and it only triggers when `current_dir()` itself fails, which previously produced a broken session anyway.

## Scope

Resolution stays at the `session/new` boundary. The child process working directory is untouched, so the multiplexed pool (one subprocess holding `channel_id → session_id`) is unchanged and this does **not** introduce the per-channel process growth documented in #2961.

Per-channel workspaces remain the open design question in #3822. This lands the env-var form the spec already promises, which is a strict prerequisite for it: the channel-scoped version is then a matter of making the resolved value a per-channel lookup at the same call site, where `PromptSource::Channel(channel_id)` is already in scope.

## Testing

8 unit tests pin the resolution order — env wins, fallback, blank-as-unset, trimming, relative rejected, missing directory rejected, both-unavailable refuses, and a guard that the `"/"` fallback cannot come back. `current_dir` and the directory check are injected, so they run in parallel without mutating process-global state.

Verified against the built binary:

```
$ AGENT_CWD=workspaces/api buzz-acp models --agent-command /bin/false
Error: AGENT_CWD is set to `workspaces/api`, which is not an absolute path. The session
working directory must be absolute so agents can resolve workspace files without searching $HOME.

$ AGENT_CWD=/srv/typo buzz-acp models --agent-command /bin/false
Error: AGENT_CWD is set to `/srv/typo`, which is not an existing directory

$ AGENT_CWD=/tmp buzz-acp models --agent-command /bin/false
error: failed to spawn agent: ...    # cwd accepted, proceeds to spawn
```

`cargo test -p buzz-acp --lib` → 676 passed, 0 failed. `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.

Note for reviewers: the existing `framed_system_prompt` tests already take `cwd` as a parameter, so no existing test needed changing — contrary to what I guessed when I first sketched the patch surface in #3822.

I can follow up with the channel-scoped map in a second PR if maintainers indicate the preferred configuration transport (env map, config file, or session-routing API).

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -114,8 +114,11 @@ All configuration is via environment variables (or CLI flags — every env var h
 | `BUZZ_ACP_IDLE_TIMEOUT` | no | `620` | Idle timeout: max seconds of silence before cancelling a turn. Resets on any agent stdout activity. |
 | `BUZZ_ACP_MAX_TURN_DURATION` | no | `7200` | Absolute wall-clock cap per turn (safety valve). |
 | `BUZZ_API_TOKEN` | no | — | API token (required if relay enforces token auth). |
+| `AGENT_CWD` | no | process working directory | Absolute path passed as `NewSessionRequest.cwd`. This is the directory the agent runtime resolves `AGENTS.md`/`CLAUDE.md`, skills, hooks and settings from, so pinning it is how one supervised runtime serves a workspace other than the directory it was launched from. Must be an existing absolute directory; buzz-acp refuses to start otherwise. |
 
 **Note:** `BUZZ_ACP_AGENT_ARGS` splits on commas. For args with values, use: `-c,key="value"`.
+
+**Note:** `AGENT_CWD` sets the *session* working directory delivered over ACP; it does not change the working directory of the buzz-acp process or of the agent subprocess. One buzz-acp process still serves all its channels from a single workspace — see [#3822](https://github.com/block/buzz/issues/3822) for the channel-scoped follow-up.
 
 **Legacy env vars:** `BUZZ_ACP_PRIVATE_KEY`, `BUZZ_ACP_API_TOKEN`, and `BUZZ_ACP_TURN_TIMEOUT` (replaced by `BUZZ_ACP_IDLE_TIMEOUT`) are still accepted as fallbacks.
 
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -4,7 +4,7 @@
 //! Config file (TOML) for complex subscription rules.
 
 use std::collections::{HashMap, HashSet};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use clap::Parser;
 use clap::ValueEnum;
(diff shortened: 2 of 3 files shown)


TolgaCinisli:
We'd like to add a concrete production use case in support of this PR.

We run five managed agents against a single self-hosted relay host (Desktop-managed; a mix of provider-backed and `backend: local`). Each agent is anchored to its own project workspace on disk — per-agent instruction files, persistent memory and usage telemetry all key off the working directory. Today `default_agent_workdir()` pins every agent to the same shared `~/.buzz`, which flattens those per-agent environments and is the one thing blocking us from moving our remaining agents to `backend: local`.

As far as we can tell this PR implements exactly what PERSONA_PACK_SPEC.md already documents as PF-4 (`AGENT_CWD` → `current_dir()` fallback → refuse), so it closes a spec/code gap rather than adding new surface. We've also verified `AGENT_CWD` is not in `RESERVED_ENV_KEYS`, so per-agent env pre-staging is safe today and activates cleanly once this merges.

Happy to test a build against our five-agent setup if that helps review.

rsaulo:
Thanks @TolgaCinisli — that's a useful confirmation, and your setup is the same shape as the one this PR was written for: one relay host, several managed agents, each needing its own directory because that is where instructions, memory and telemetry are keyed.

Yes please on testing it against your five-agent setup. What that takes:

```bash
git remote add rsaulo https://github.com/rsaulo/buzz.git
git fetch rsaulo feat/agent-cwd
git checkout rsaulo/feat/agent-cwd
cargo build -p buzz-acp --release
```

Then set `AGENT_CWD=/abs/path/to/workspace` per agent. Two things worth knowing before you run it:

- **Resolution is `AGENT_CWD` → `current_dir()` → refuse.** An `AGENT_CWD` that is set but unusable (relative path, or a directory that does not exist) is a hard error, not a fall-through to the process directory. That is deliberate: an operator who pinned a workspace and silently got the process directory instead would only find out when an agent answered without the repository's instructions loaded.
- **It sets the *session* cwd delivered over ACP**, not the process cwd. One buzz-acp process still serves all of its channels from a single workspace. If your five agents are five separate processes with five separate environments, that is exactly the case this covers. If you ever want one process to serve several channels from different directories, that is the channel-scoped follow-up in #3822, not this PR.

If you hit anything, the resolution logic is `resolve_agent_cwd_with` in `crates/buzz-acp/src/config.rs` and it is unit-tested against the spec's ordering.

---

For whoever picks this up for review — current state:

- CI green: DCO, Semgrep OSS, zizmor
- `MERGEABLE`, no conflicts against current `main`, one commit (`89ba143`)
- No maintainer review since it was opened on 03/08

The branch is behind `main` by a fair margin without conflicting. Happy to rebase onto current `main` before review if you would rather look at it on a fresh base — say the word and I will push.

@wesbillman you have been reviewing and merging most of the recent `buzz-acp` changes — is there a specific owner this should go to, or is a `buzz-oss-team` review the right path? There is now a second deployment blocked on it, so I would rather unstick it than let it keep drifting.

TolgaCinisli:
Tested on macOS arm64 (M4, toolchain via repo Hermit), commit 89ba143, against the host that runs our five-agent setup:

- **Build:** `cargo build -p buzz-acp --release` — clean.
- **Unit:** all 8 new `agent_cwd` tests pass. Full `--lib` suite is 677/678 — the one failure, `acp::tests::keepalive_resets_idle_past_deadline`, fails identically at this PR's base commit (8342dfc) on the same machine in both debug and release, so it's a pre-existing timing sensitivity unrelated to this change.
- **Binary smoke** (throwaway key, dead relay port):
  - `AGENT_CWD=relative/path` → refused at startup, exit 1, with the not-absolute message
  - `AGENT_CWD=/nonexistent` → refused at startup, exit 1, with the not-a-directory message
  - `AGENT_CWD="   "` → read as unset, falls back to `current_dir()` (confirmed in the startup log)
  - `AGENT_CWD=/abs/existing/dir` → `buzz-acp session working directory: <dir>` logged and startup proceeds

Resolution order matches the spec exactly. From our side this is ready for review.

rsaulo:
This PR implements PF-4 in `crates/buzz-persona/PERSONA_PACK_SPEC.md`, which already specifies `NewSessionRequest.cwd` resolution as `AGENT_CWD` → `std::env::current_dir()` → refuse to start. On current `main`, `AGENT_CWD` appears 18 times, all of them in that spec file and none in code. `current_working_directory()` in `crates/buzz-acp/src/lib.rs` implements steps 2 and 3; step 1 is the gap this closes. It is not new surface.

@TolgaCinisli independently built and tested commit `89ba143` on macOS arm64 against a five-agent host (comments above): the 8 new unit tests pass, resolution order matches the spec, and the single `--lib` failure reproduces unchanged at the base commit.

Rebase status, measured against `main` at `ac5a186`: `config.rs` and `README.md` merge cleanly. `lib.rs` has two one-line conflicts, both from `current_working_directory()` being extracted after this branch was cut — the two call sites need to route through it rather than through the inline expressions they replaced. `PromptContext.cwd` is still a single field, so the injection point is unchanged.

I can rebase, but I'd rather not guess at the destination. Could a maintainer say whether this is a merge, a changes-requested, or a no — and if a no, on what grounds?

Issue #4477, opened 2026-08-03
buzz-agent/buzz-acp: transient network outage paralyzes agent for 12+ min; retries noisy and recovery passive


csh0101:
## Summary

A brief network outage (DNS failure lasting ~30s) paralyzed a managed `buzz-agent` for **12+ minutes**, and the user-visible fallout ("turn error" states in the desktop UI, silent message queueing) persisted long after the network recovered. The harness eventually self-healed via requeue, but the recovery path is slow, opaque, and noisy. Observed on a production multi-agent deployment.

## Observed timeline (all times UTC, real logs)

| Time | Event |
|---|---|
| 04:00:03 | User DMs the agent; message steered into an in-flight turn, deadline extended +7200s |
| (network) | macOS DNS for the LLM gateway domain breaks; route black-holes |
| 04:02:49 | `llm: transport error, retrying attempt=1 max_attempts=3` — **~2.5 min after the request started hanging** |
| 04:06:49 | attempt=2 — another ~4 min hang |
| 04:10:50 | `cumulative stall 721s across 3 attempts` → `session_prompt error code -32000` |
| 04:10:50 | `queue: requeueing failed batch with backoff attempt=1 max=10` |
| 04:12 | Network confirmed healthy again (gateway reachable, 200 OK) |
| 04:14:56 | Requeued batch hits transport error **again** (stale connection state / still-settling routes) |
| 04:19:46 | `WebSocket error: Connection reset without closing handshake` |
| 04:26+ | Agent finally drains the backlog and replies to all queued DMs |

## Problems

**1. Each failed LLM attempt hangs minutes before erroring.**
`crates/buzz-agent/src/llm.rs` retries with sensible backoff (`MAX_RETRIES=3`, 500ms–8s jitter), but the transport error itself takes ~4 minutes to surface per attempt (black-holed route → long TCP/TLS timeout). So a 30-second DNS blip costs 3 × ~4min ≈ **12 minutes of turn paralysis**. There is no aggressive per-attempt connect/request timeout for the chat-completions call.

**2. Queued messages pile up behind the doomed turn.**
New messages arriving during the stall become non-cancelling steers that only extend the hard deadline (`extending in-flight deadline by 7200s + 100s buffer`) — from the user's perspective the agent simply ignores them.

**3. The failure/recovery is user-visible and noisy.**
[...]

Issue #3300, opened 2026-07-28
Track Compute Share managed-agent startup, MeshLLM diagnostics, and relay-mesh text-only media handling


stevepresley:
## Summary

Follow-up tracker for the broader Compute Share / managed-agent / relay-mesh behaviors validated while fixing Windows MeshLLM desktop builds in #3223.

The Windows packaging issue (#2836) uncovered and/or required fixes beyond simply enabling the `mesh-llm` feature for Windows installers:

- Managed `buzz-agent` should start correctly from the installed desktop app.
- Relay-mesh LLM calls should not fail with text-only model media-input `422` errors caused by replaying prior tool-result images into later requests.
- MeshLLM diagnostic logging should be available as an opt-in runtime toggle from Compute → Advanced.

## Validation target

A desktop install should be able to:

1. Start Compute Share.
2. Start a managed `buzz-agent` from the installed app.
3. Route relay-mesh text requests to shared compute without replayed image/tool-result history causing `422 Unprocessable Entity` on text-only models.
4. Enable/disable MeshLLM diagnostic logging from the Compute advanced settings without rebuilding or setting environment variables.

## Related

- PR: #3223
- Original Windows packaging issue: #2836

PR #3223, opened 2026-07-27
fix(desktop): enable Windows mesh-llm builds and address Compute Share startup/MeshLLM debug logging/non-image models trying to parse image input

stevepresley:
# fix: enable Windows mesh-llm desktop builds

Closes #2836
Closes #3300

## Summary

- Enable the `mesh-llm` feature for Windows release and canary desktop builds.
- Bundle required Windows MeshLLM native runtime DLL dependencies as Tauri resources and register their directory at runtime.
- Avoid Windows stack overflows during mesh start and managed-agent start by running heavy startup work on larger-stack OS threads.
- Make Compute Share startup more resilient/idempotent and fix stale progress UI behavior.
- Add opt-in MeshLLM diagnostic logging controls in Compute → Advanced.
- Add relay-mesh protection so prior tool-result images are not replayed into text-only shared-compute LLM requests.

## Validation

Validated locally on Windows 11:

- `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml --all`
- `cargo check --manifest-path desktop/src-tauri/Cargo.toml --features mesh-llm`
- `pnpm typecheck`
- `cargo check -p buzz-agent`
- `cargo build -p buzz-agent --release --target x86_64-pc-windows-msvc`
- `bash scripts/bundle-sidecars.sh x86_64-pc-windows-msvc`
- `pnpm tauri build --target x86_64-pc-windows-msvc --bundles nsis --features mesh-llm --config "{\"bundle\":{\"createUpdaterArtifacts\":false}}"`

Manual validation:

- Installed the NSIS build on Windows 11.
- Verified Settings → Compute no longer shows the mesh-llm feature-stub behavior.
- Verified Compute Share starts and serves requests.
- Verified Windows MeshLLM native runtime dependency loading works with bundled MinGW DLLs.
- Verified managed `buzz-agent` starts from the installed app.
- Verified relay-mesh LLM calls no longer fail with the text-only model media-input `422` caused by replayed tool-result images.
- Verified MeshLLM diagnostic logging toggle works at runtime.

## Related/out of scope findings

Windows relay-mesh validation also reproduced existing ACP/agent delivery behavior where a model may produce Activity/final text without publishing a DM/channel message, and failed ACP turns can be retried later. That is out of scope for this Windows mesh-llm packaging/runtime PR.

Refs #2698
Refs #2421
Refs #2681

PR #4566, opened 2026-08-03
fix(agent): degrade images for text-only models

### LINKED ISSUE #4563: Text-only model image rejection can poison managed-agent channel queue
🤖 A managed agent using a text-only endpoint can get stuck when its conversation history contains an image returned by a tool.

## Reproduction

1. Configure `buzz-agent` with a text-only OpenAI-compatible model endpoint.
2. Run a tool that returns `ToolResultContent::Image` so the image is retained in the session history.
3. Send the next model request.
4. The endpoint rejects the request with HTTP 400, for example: `Image input is not supported for this endpoint`.

## Current behavior

- Provider serializers replay the tool-result image as binary/base64 input.
- The failed turn remains associated with the ACP session.
- `buzz-acp` treats the application error as retryable, requeues the batch at the front of the channel queue, and applies channel-wide backoff.
- Later text-only events queue behind the deterministic failure and can share its retry/dead-letter fate.

## Expected behavior

- Preserve the image and accompanying text/metadata in session history.
- When an endpoint proves it cannot accept images, retry the request once with a textual image placeholder instead of binary image data.
- Remember that negative capability for the exact effective model for the process lifetime, without persisting it or inferring it from model names.
- Do not let the deterministic unsupported-image error consume the queue retry budget or block later text events.

## Proposed scope

- Add a narrow provider-error classifier for the observed HTTP 400 signature.
- Degrade images only while serializing requests for a model that has returned that signature.
- Keep multimodal behavior unchanged for unknown and image-capable models.
- Add an ACP-side non-retryable fallback with session invalidation and a user-visible notice for defense in depth.

Related work: #3300, #3223, #4477, and #4127.

LordMelkor:
## Summary

- retry once with text placeholders when an endpoint returns the narrow unsupported-image HTTP 400 signature
- cache that negative capability by exact resolved request model/route for the process lifetime, while preserving images in session history and leaving other routes unchanged
- immediately dead-letter and reset the ACP session if the deterministic error escapes the built-in fallback, so later text events are not blocked behind retries
- add provider serialization, resolved-route cache isolation, classifier, route-change, and queue-release regression coverage

### Related issue

Fixes #4563.

Related prior work: #3300, #3223, #4477, and #4127.

### Testing

- `cargo test -p buzz-agent -p buzz-acp` — 1,208 passed across 14 suites
- `cargo clippy -p buzz-agent -p buzz-acp --all-targets -- -D warnings -A clippy::question_mark`

Strict Clippy is currently blocked by three pre-existing `clippy::question_mark` findings in untouched `crates/buzz-agent/src/hints.rs` and `crates/buzz-acp/src/queue.rs`.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -3360,6 +3360,18 @@ fn is_auth_error(error: &acp::AcpError) -> bool {
     message.contains("Re-authenticate") || message.contains("API Error: 401")
 }
 
+/// Match only the deterministic provider rejection emitted when an endpoint
+/// cannot accept image input. False positives are especially costly here
+/// because this classification dead-letters the batch instead of retrying it.
+fn is_unsupported_image_input_error(error: &acp::AcpError) -> bool {
+    let acp::AcpError::AgentError { message, .. } = error else {
+        return false;
+    };
+    let message = message.to_ascii_lowercase();
+    message.contains("400 bad request")
+        && message.contains("image input is not supported for this endpoint")
+}
+
 /// Spawn a task that posts a user-visible failure notice to the relay.
 ///
 /// Shared by the hard-cap immediate dead-letter path and the retries-exhausted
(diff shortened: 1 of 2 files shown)


wesbillman (changes requested):
Reviewing on Wes's behalf.

The underlying failure class is real: `buzz-agent` currently serializes every `ToolResultContent::Image` back into subsequent provider requests unconditionally, while `buzz-acp` requeues ordinary application errors ahead of later channel events. There is also an earlier independently documented/manual reproduction of the text-only media rejection in #3300/#3223 (that path reported 422 rather than this PR's exact 400 wording).

I don't think this implementation is safe to merge yet because the learned capability is not always keyed by the model that actually rejected the image. For OpenAI relay-mesh `auto`, `effective_model` remains `auto`, while `openai_request()` resolves each request to `mesh` or ordinary `auto` from the live catalog (`resolve_openai_model`). This PR caches the rejection under the outer `effective_model`, so one text-only routed backend permanently strips images from all later `auto` requests in this process—even after routing changes to an image-capable backend. That contradicts the stated model isolation and can silently degrade capable models.

Please key the negative capability to the actual request model/route that returned the rejection (or explicitly define/test safe virtual-model behavior), and add regression coverage for `auto` changing routes after one backend rejects an image. The reactive retry plus serialization-time placeholder is otherwise a reasonable general approach; inference from model names would be worse.

LordMelkor:
🤖 @wesbillman Thanks — addressed in `3700126e`. The negative image capability is now learned at the OpenAI dispatch boundary against the concrete resolved request model (`mesh`, ordinary `auto`, or an explicit model), including the internal `mesh` → `auto` fallback path. A new regression changes `auto` routes after `mesh` rejects an image and verifies that the later `auto` route still receives the image. Full `buzz-agent`/`buzz-acp` suite: 1,153 passed; format and scoped Clippy pass. Could you please take another look?

LordMelkor:
Rebased onto current `main` and resolved the image-recovery overlap: upstream handles unsupported-image responses on the 404 path, while this PR retains only the narrow HTTP 400 capability-learning/retry path. The full `buzz-agent` + `buzz-acp` test run now passes 1,208 tests across 14 suites.

PR #4574, opened 2026-08-03
fix(acp): trigger agents for untagged DMs

lev0a:
## Summary

- treat messages in private DMs as implicitly addressed when buzz-acp runs in its default mentions mode
- keep shared channels tag-gated and preserve explicit require_mention rules in config mode
- handle both startup discovery and dynamically joined DMs, and label DM prompts accurately

## Why

The default mentions subscription required an agent p tag in every channel. In a private DM that made the human send a redundant explicit tag before the harness would trigger, even though the message was already addressed to the DM participant.

## Verification

- cargo fmt --all -- --check
- cargo clippy -p buzz-acp --all-targets -- -D warnings
- cargo test -p buzz-acp (671 unit tests and 9 integration tests passed)
- git diff --check
- just ci: workspace clippy completed; desktop-check could not start because the repository Hermit manifest cannot resolve pnpm-11.4.0 (no source provided)

## Security boundary

DM author admission is unchanged: the existing DM author gate still limits triggering authors to the owner and sibling agents. Shared-channel subscription behavior is unchanged.

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -1,6 +1,6 @@
 # buzz-acp
 
-ACP harness that connects AI agents to Buzz. The harness listens for @mentions on the relay, prompts your agent, and the agent replies using the Buzz CLI.
+ACP harness that connects AI agents to Buzz. The harness listens for @mentions in shared channels and messages in private DMs, prompts your agent, and the agent replies using the Buzz CLI.
 
 ```
 Buzz Relay ──WS──→ buzz-acp ──stdio──→ Your Agent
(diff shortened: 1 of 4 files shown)


sonpiaz:
This is the shape that survives the case I actually hit. Running an always-on agent on Buzz, the DM that silently died was one without a `p` tag:

```
29 owner messages in one DM channel
28 with a p tag        answered every time
 1 without a p tag     2026-08-11T23:59:08Z, no turn ever fired, no error either side
```

Signed relay events, same channel, same two pubkeys. The failure is completely silent, which is why it reads to a user as "my agent is broken" rather than "my client dropped a tag".

The reason I am commenting here rather than only on the client side PRs: `require_mention` on the per-channel filter turns into a server side `#p` constraint, so an untagged DM never reaches the agent at all. Fixing the mobile sender is necessary, but until this lands, any client that forgets the tag produces the same silence, and there are six open PRs all trying to fix the sender.

What makes this one worth preferring over #5265: threading `implicitly_addressed_channels` through `resolve_dynamic_channel_filter_with_implicit_mention` also covers channels discovered after startup. That is the common case, since a DM to an always-on agent usually gets created after the agent is already running. #5265 resolves DM channels once from the startup map and leaves lib.rs:2504 untouched.

Builds and passes the full crate suite here on macOS 15 / Apple silicon, `rustc 1.95.0`:

```
test result: ok. 671 passed; 0 failed; 0 ignored
```

Also commented the same measurements on #5265.

PR #4575, opened 2026-08-03
ci: add guarded desktop release cache prewarm

wesbillman:
## Summary

Gate 1 only for desktop release caching:

- replaces canary `rust-cache` use with explicit exact-key `actions/cache/restore` + `save`
- computes keys after `cargo update --workspace`, including platform, target, Rust toolchain, Cargo manifests/locks, profile/features, and native-toolchain inputs
- normalizes only the desktop package version so a trusted `main` canary can warm an otherwise identical release tag
- excludes Tauri bundle directories, so installers and signed artifacts are never cached
- adds a restore-only `cache-proof-*` tag workflow that fails unless tag scope sees the exact default-branch cache
- adds contract tests that enforce no release-workflow cache change in Gate 1

`release.yml` is intentionally unchanged. A cache miss remains the current cold canary build; the release path cannot be affected by merging this PR.

## Validation

- `scripts/test-desktop-release-cache-key.sh`
- `scripts/test-desktop-release-cache-workflow.sh`
- Ruby YAML parse of all four changed workflows
- `git diff --check`
- pre-push `branch-skew`

## Post-merge proof plan

1. Run each canary cold on trusted `main`, recording cache size/save time and fresh artifact inventory.
2. Run each canary warm, requiring the exact-key hit and recording restore/build time.
3. Create a disposable `cache-proof-*` tag at that same trusted `main` SHA and dispatch **Desktop release cache tag-scope proof** from the tag.
4. Do not begin Gate 2 or modify `release.yml` unless the exact tag-scope restore succeeds and cache transfer economics are favorable.


wesbillman:
@github-advanced-security The seven Semgrep shell-injection findings are addressed at `9bbf3bd4a`: GitHub matrix and step-output expressions now enter cache-key steps only through `env`, and the generated shell references quoted environment variables. A workflow contract now rejects direct `${{ ... }}` interpolation inside every cache-key `run:` block. This comment is posted by Carl on Wes’s behalf.

Issue #4580, opened 2026-08-03
Workflow deletion: non-owner delete is accepted then silently ignored, and owner delete leaves a ghost kind:30620 definition event


redirwin:
## Summary

Two related defects in the workflow (kind:30620) deletion path:

1. **Non-owner delete is silently ignored but acknowledged.** A non-owner
   submitting the CLI's kind:5 deletion for another author's workflow receives
   `{"accepted":true, "event_id":"…"}` — but the workflow persists and keeps
   firing. Cause: `delete_workflow_for_owner` matches rows on owner, deletes
   zero rows for a non-owner, and the NIP-09 event acceptance is reported
   regardless of effect. By contrast, non-owner `workflows trigger` correctly
   returns `forbidden` — the authorization surfaces are inconsistent, and the
   delete path converts an authorization failure into false completion
   evidence.

2. **Owner delete leaves a ghost definition event.** A successful owner
   deletion removes the workflows DB row (engine-side: `trigger` then returns
   `400 workflow not found`) but does NOT remove or tombstone the stored
   kind:30620 definition event. `workflows list` (and any client rendering
   kind:30620) keeps showing the deleted workflow indefinitely.

## Why it matters

`accepted:true` on a write that had no effect gives clients false completion
evidence. In a multi-agent channel this caused a real coordination fault: an
agent believed a scheduled workflow was deleted and provisioned a replacement,
risking duplicate scheduled runs. The ghost event then makes listings disagree
with the engine about what exists. Accepted-but-ignored writes are the same
failure shape as the announcement-update timestamp defect (#4509).

## Reproduction

1. Identity A (channel member) creates a scheduled workflow W:
   `buzz workflows create --channel <uuid> --yaml <def>`.
2. Identity B (same channel, not owner):
   - `buzz workflows trigger --workflow W` → `forbidden` (correct)
   - `buzz workflows delete --workflow W` → `accepted:true`; W persists and
     fires on schedule (defect 1)
3. Identity A: `buzz workflows delete --workflow W` → engine row deleted
   (`trigger` → `400 workflow not found`), but the kind:30620 event still
   appears in `workflows list` (defect 2).

## Expected

- Non-owner delete: explicit `forbidden`, matching `trigger` — never
  `accepted:true` for a write that will be ignored.
- Owner delete: the kind:30620 definition event is deleted or tombstoned so
  listings agree with the engine.

## Environment

- Self-hosted Buzz relay (Railway), deployed commit `788b3c0`
[...]

PR #4581, opened 2026-08-03
fix(relay): scope multi-channel #h queries to all listed channels

### LINKED ISSUE #4579: Multi-value `#h` filters silently narrow to the lexicographically first channel, dropping all others
## Summary

A `/query` (or REQ) filter with multiple `#h` channel values — e.g.
`{"kinds":[30620], "#h":[<id1>,<id2>,…]}` — is silently answered as if it
named only ONE channel: the lexicographically smallest id. Events in every
other listed channel are never returned. No error or warning is surfaced.

## Root cause

`extract_channel_id_from_filter` in `crates/buzz-relay/src/handlers/req.rs`
returns the first parseable UUID from the filter's `#h` values:

```rust
        let key = tag_key.to_string();
        if key == "h" {
            for val in tag_values {
                    return Some(id);
                }
            }
        }
    }
    None
}
```

`nostr::Filter` stores `generic_tags` values in a sorted set, so "first" is
deterministically the smallest id. `build_event_query_from_filter` bakes that
single id into the SQL scope (`filter_to_query_params(filter, channel_id, …)`),
and the Rust post-filter afterwards can only REMOVE rows — it can never
restore the channels the SQL scope dropped.

The sibling helper `extract_channel_id_from_filters` (plural, same file)
shows the intended semantics: on multiple distinct channel ids it returns
`None` and falls back to global scope. The singular version lacks that guard.
The HTTP bridge's own `extract_channel_from_filter`
(`crates/buzz-relay/src/api/bridge.rs`) also gets it right (`None` unless
exactly one value) — but it feeds only the access-scope check, not the query
builder, so the narrowing still happens underneath it.

## User-visible impact

Buzz Desktop's global Workflows screen sends exactly this shape — one batched
kind:30620 query with every member channel in `#h`
(`desktop/src/features/workflows/ui/WorkflowsView.tsx` →
`get_channels_workflows`). For any user whose alphabetically-first channel is
a DM (typical), the screen renders "No workflows yet" even when workflows
exist and run. Refresh does not help. Any other client using multi-`#h`
filters is affected the same way.

## Reproduction (empirical, self-hosted relay at 788b3c0)

NIP-98-signed `POST /query`, requester is a member of all channels named:

| `#h` values | Result |
| --- | --- |
| [apiary] | 2 workflow events |
| [dm-sorting-before-apiary, apiary] | 0 events |
| [dm-sorting-after-apiary, apiary] | 2 events |

Every combination follows the sort rule. Repro script available.

## Expected

Multi-`#h` should scope to the UNION of the listed channels — e.g. push the
whole set down as a `channel_ids` IN-list (the mechanism the access-scope
path already uses) instead of first-value narrowing, or fall back to
global scope + post-filter as `extract_channel_id_from_filters` does.

## Notes

`filter_fully_pushable` already declares multi-`#h` "not pushable", so the
fast-COUNT path correctly avoids it — the defect is only in the query-scope
builder. A fix should add a regression test: multi-`#h` filter returns events
from ALL listed channels regardless of sort order.

A PR with a proposed fix (IN-list pushdown + regression tests) follows
shortly and will be linked here.

redirwin:
## Summary

A `/query` or COUNT filter naming multiple channels — e.g.
`{"kinds":[30620], "#h":[<id1>,<id2>,…]}` — is silently answered as if it
named only ONE channel: the lexicographically smallest id. Events in every
other listed channel are never returned, with no error surfaced.

`extract_channel_id_from_filter` returns the first parseable UUID from the
filter's `#h` values, and `nostr::Filter` stores tag values in a sorted set,
so "first" is deterministically the smallest id. `filter_to_query_params`
bakes that single id into the SQL scope, and the Rust post-filter afterwards
can only remove rows — it can never restore the channels the SQL already
dropped. The sibling `extract_channel_id_from_filters` (plural) already has
the right semantics: multiple distinct ids → `None` → global scope.

**User-visible impact:** Buzz Desktop's Workflows screen sends exactly this
shape — one batched kind:30620 query with every member channel in `#h`
(`WorkflowsView.tsx` → `get_channels_workflows`). For any user whose
alphabetically-first channel is a DM (typical), the screen shows
"No workflows yet" even when workflows exist and run. Reproduced empirically
on a production relay with NIP-98-signed queries:

| `#h` values | Result |
| --- | --- |
| [apiary] | 2 workflow events |
| [dm-sorting-before-apiary, apiary] | 0 events |
| [dm-sorting-after-apiary, apiary] | 2 events |

### The fix

- `extract_channel_id_from_filter` now returns `Some` only when the filter
  names exactly one parseable channel id, matching the plural helper's
  semantics. (One benign divergence: the plural helper dedups case-variant
  spellings of the same id, while the singular counts elements and routes
  such filters down the multi-channel path — the IN-list then carries a
  harmless duplicate and results are identical.)
- `filter_to_query_params` pushes a multi-channel `#h` set down as an
  `EventQuery::channel_ids` IN-list, so the SQL page is scoped to the
  requested channels rather than the caller's entire access scope (busy
  unrelated channels can no longer starve the requested ones past `LIMIT`).
- `apply_access_scope_to_query` intersects a filter-supplied `channel_ids`
  list with the caller's accessible channels instead of overwriting it.
  An empty intersection stays `Some(vec![])`, which the SQL layer already
  treats as "global events only" — fail closed, never "no restriction".

The WS REQ path already used a correct inline exactly-one extractor with a
subscription-scope fallback; it now additionally benefits from the
`channel_ids` pushdown for single-filter multi-`#h` subscriptions. The
existing Rust post-filters (`filters_match` + accessible-channel check +
`event_visible_to_reader`) are unchanged and remain defense-in-depth.

### Possible follow-up (maintainer preference invited)

The COUNT fallback (`count.rs`) and the bridge query paths currently
overwrite a filter-supplied `channel_ids` pushdown with the caller's access
scope rather than intersecting the two. Correctness is unaffected —
`filter_fully_pushable` refuses multi-`#h` filters for fast COUNT and the
Rust post-filter computes exact results — but a multi-`#h` COUNT scans the
caller's entire accessible set and is likelier to hit the candidate budget,
surfacing an explicit "requires narrower constraints" error where the
pre-fix code returned a silently wrong count. Intersecting there (reusing
`apply_access_scope_to_query`) would make the scan narrower and cheaper;
it is kept out of this PR to keep the change minimal. Happy to fold it in
or file it separately, whichever the maintainers prefer.

### Related issue

Fixes #4579. A related-but-separate report on the workflow deletion path
(authorization inconsistency + ghost kind:30620 definition event observed
during the same investigation) is #4580 — not addressed by this PR.

### Testing

New unit tests in `handlers/req.rs`:

- `single_filter_multi_channel_does_not_narrow_to_first` — regression pin on
  the extractor, both insertion orders.
- `multi_channel_filter_pushes_channel_ids_in_list` /
  `single_channel_filter_keeps_channel_id_predicate` — query-builder shape.
- `access_scope_intersects_multi_channel_filter` — the Desktop Workflows
  regression shape: a channel sorting before the target must not evict it,
  and the access scope must not widen the filter.
- `access_scope_empty_intersection_fails_closed` — `Some(vec![])` preserved.

`cargo test -p buzz-relay --lib` (rust:1.95 container, no Postgres):
835 passed, 37 ignored, 9 failed — the 9 failures are DB-backed
`api::admin`/`api::media` tests (`Sqlx(PoolTimedOut)`) that fail identically
on unmodified `origin/main` `2c0ac246` in the same environment (baseline run
verified). Zero new failures introduced; all 6 new regression tests pass.
Independently re-verified by a second reviewer against the same commit,
including a fetch of current `origin/main` confirming no conflicting changes
to the touched files.

### Credits

Root cause isolated and reproduced with signed queries by the Buzz agents
operating the redirwin.com installation.

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -847,19 +847,38 @@ fn filters_are_nip43_membership_only(filters: &[Filter]) -> bool {
         })
 }
 
-/// Extract a channel UUID from a single filter's `#h` tag.
+/// Extract the channel UUID from a single filter's `#h` tag, or `None` unless
+/// the filter is scoped to exactly one channel.
+///
+/// Mirrors [`extract_channel_id_from_filters`]: a filter naming multiple
+/// distinct channels has no single `channel_id` predicate. Taking the first
+/// value instead would silently narrow the query — `nostr::Filter` stores tag
+/// values in a sorted set, so "first" is the lexicographically smallest
+/// channel id and every other listed channel would be dropped before `LIMIT`.
+/// Multi-channel filters are pushed down as an `EventQuery::channel_ids`
+/// IN-list in [`filter_to_query_params`] instead.
 fn extract_channel_id_from_filter(filter: &Filter) -> Option<uuid::Uuid> {
-    for (tag_key, tag_values) in filter.generic_tags.iter() {
-        let key = tag_key.to_string();
-        if key == "h" {
-            for val in tag_values {
-                if let Ok(id) = val.parse::<uuid::Uuid>() {
-                    return Some(id);
-                }
-            }
-        }
+    match channel_ids_in_filter(filter).as_slice() {
+        [only] => Some(*only),
+        _ => None,
     }
-    None
+}
+
+/// All distinct parseable channel UUIDs in a single filter's `#h` values, in
+/// the set's (sorted) order. Empty when the filter carries no `#h` tag or no
+/// value parses as a UUID.
+fn channel_ids_in_filter(filter: &Filter) -> Vec<uuid::Uuid> {
+    let h_tag = nostr::SingleLetterTag::lowercase(nostr::Alphabet::H);
+    filter
+        .generic_tags
+        .get(&h_tag)
+        .map(|values| {
+            values
+                .iter()
+                .filter_map(|v| v.parse::<uuid::Uuid>().ok())
+                .collect()
+        })
+        .unwrap_or_default()
 }
 
 /// Convert a single NIP-01 filter into an [`EventQuery`] for the database.
@@ -978,8 +997,25 @@ fn filter_to_query_params(
         (None, None)
     };
 
+    // Multi-channel `#h` pushdown: a filter naming several channels has no
+    // single `channel_id` predicate (the extractor returns `None` for it), so
+    // push the whole set as a `channel_ids` IN-list. Without this, the query
+    // falls back to the caller's full access scope and busy unrelated channels
+    // can starve the requested channels' rows past `LIMIT` before the Rust
+    // post-filter runs. `apply_access_scope_to_query` intersects this list
+    // with the caller's accessible channels, so it never widens access.
+    let channel_ids = if channel_id.is_none() {
+        match channel_ids_in_filter(filter).as_slice() {
+            [] | [_] => None,
+            many => Some(many.to_vec()),
+        }
+    } else {
+        None
+    };
+
     EventQuery {
         channel_id,
+        channel_ids,
         kinds,
         pubkey,
         since,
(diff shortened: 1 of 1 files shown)


redirwin:
Closing this as superseded by #6009, which has now merged into `main`. It addresses the same multi-value `#h` query bug and also covers the broader COUNT, live-subscription, authorization, bounded-work, and older-relay compatibility cases raised during review.

This PR documented and reproduced the original lexicographic-first-channel failure, but rebasing it now would duplicate the merged implementation. Thanks to everyone who reviewed and extended the fix.

PR #4593, opened 2026-08-03
Define private managed agent wire protocol

wesbillman:
## Summary

- reserve kind `30179` for owner-private managed-agent aggregates
- define the fail-closed owner-self NIP-44 v2 envelope and versioned payload codec
- bind runnable identity/configuration to complete signed `30175`/`30177` recovery projections
- validate NIP-OA owner→agent attestations and reject self-attestation
- document NIP-PMA authority, migration prerequisites, privacy, and deployment order
- keep generic relay ingest closed until private storage and atomic aggregate CAS exist

## Safety boundary

This is the inert protocol/codec slice only. It does not publish secrets, change agent authority, migrate local records, or enable kind `30179` ingestion. The relay regression test proves generic EVENT ingest still rejects the kind.

The finalized migration plan adds later prerequisites for relay-private storage/CAS, runtime lease/fencing, Desktop cutover, and harness authentication. Those belong in staged follow-up PRs rather than expanding this inert foundation.

## Validation

At commit `67f0ea4ebb8d3ccba3a3eb9374e89a7178913f74`:

- `cargo test -p buzz-core` — 246 unit + 2 doc tests passed
- `cargo test -p buzz-relay private_managed_agent_kind_remains_rejected_until_atomic_ingest_exists` — passed
- push hooks: Rust tests and desktop checks passed (`2145` desktop tests passed, `14` ignored)
- `git diff --check`

## Review

Princess Donut cleared security/data integrity with no remaining high/medium findings. Mongo cleared migration compatibility and wire grammar. The later runtime lease/fencing protocol was also adversarially cleared as a plan; implementation slices still require independent evidence before activation.

Deterministic plaintext/signed-projection/auth-tag interoperability vectors remain a valuable follow-up, not an S0 merge gate; random NIP-44 ciphertext is intentionally not snapshotted.

### DIFF
--- a/crates/buzz-core/src/kind.rs
+++ b/crates/buzz-core/src/kind.rs
@@ -108,6 +108,15 @@ pub const KIND_EVENT_REMINDER: u32 = 30300;
 /// dedicated push lease tables.
 pub const KIND_PUSH_LEASE: u32 = 30350;
 
+/// NIP-PMA: owner-encrypted private managed-agent aggregate.
+///
+/// Addressed by `(owner pubkey, kind, agent pubkey)`. The signed outer tags
+/// expose only the agent coordinate, CAS generation/predecessor, and active/deleted
+/// state required for relay enforcement. Content is NIP-44 v2 encrypted from
+/// the owner's key to itself and contains the runnable identity/configuration
+/// plus exact public projection bindings. See `docs/nips/NIP-PMA.md`.
+pub const KIND_PRIVATE_MANAGED_AGENT: u32 = 30179;
+
 /// Kinds whose stored events are readable only by their author.
 ///
 /// The relay must never reveal the existence, count, tags, content, schedule,
@@ -117,7 +126,11 @@ pub const KIND_PUSH_LEASE: u32 = 30350;
 ///
 /// Currently a tiny linear set. If this grows past ~4 kinds, convert to a
 /// compile-time bitset or sorted array with binary search for hot-path use.
-pub const AUTHOR_ONLY_KINDS: &[u32] = &[KIND_EVENT_REMINDER, KIND_PUSH_LEASE];
+pub const AUTHOR_ONLY_KINDS: &[u32] = &[
+    KIND_EVENT_REMINDER,
+    KIND_PUSH_LEASE,
+    KIND_PRIVATE_MANAGED_AGENT,
+];
 
 /// Kinds that require a result-level read gate beyond the filter-layer
 /// `#p` check: even a reader who knows an event id MUST match the event's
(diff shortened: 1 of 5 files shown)


wpfleger96 (commented):
I left two non-blocking notes from my exact-head review.

crates/buzz-core/src/private_managed_agent.rs
```diff
@@ -0,0 +1,1091 @@
...
+    /// Exact signed kind:30177 event ID.
+    pub event_id: String,
+    /// Lowercase SHA-256 of the exact projection content bytes.
+    pub content_sha256: String,
+    /// Versioned signed event sufficient to reproduce the projection.
+    pub recovery: ProjectionRecoveryV1,
+}
+
+/// Secret agent identity material. It never appears in public projections.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
wpfleger96: Non-blocking: this `Debug` derive propagates through `ActivePayload` and `Payload`, so formatting a decrypted payload will print `private_key_nsec`, `auth_tag`, secret `env_vars`, and potentially secret-bearing `backend` data. I could not find a current production consumer or concrete logging call, so this is defense in depth rather than a present leak. Before activation, could we remove `Debug` from the secret-bearing types or implement explicitly redacted `Debug` output?

crates/buzz-core/src/kind.rs
```diff
@@ -117,7 +126,11 @@ pub const KIND_PUSH_LEASE: u32 = 30350;
 ///
 /// Currently a tiny linear set. If this grows past ~4 kinds, convert to a
 /// compile-time bitset or sorted array with binary search for hot-path use.
-pub const AUTHOR_ONLY_KINDS: &[u32] = &[KIND_EVENT_REMINDER, KIND_PUSH_LEASE];
+pub const AUTHOR_ONLY_KINDS: &[u32] = &[
+    KIND_EVENT_REMINDER,
+    KIND_PUSH_LEASE,
+    KIND_PRIVATE_MANAGED_AGENT,
```
wpfleger96, line 132: Non-blocking for this inert reservation: fresh databases use the positive FTS allowlist, so `30179` gets a `NULL search_tsv`, but brownfield databases can retain an older negative-list generated expression that does not exclude this kind. Generic `EVENT` ingest still rejects `30179`, so I found no currently reachable production insertion path. Before aggregate storage is activated, could we add an additive brownfield migration/test proving `30179` remains unindexed and clarify that the deployment checklist covers brownfield normalization as well as fresh-schema verification?

wesbillman: Agreed; deferring this to the storage/activation phase because kind 30179 remains rejected by generic ingest in this reservation. The activation PR must add the additive brownfield normalization migration and regression coverage before aggregate submission is enabled; `docs/nips/NIP-PMA.md` already orders privacy/FTS hardening before CAS storage and feature-gated submission (steps 2–4).

crates/buzz-core/src/private_managed_agent.rs
wesbillman: Addressed in 89efaabcf. `PrivateIdentity` now redacts the key and optional auth tag, while `PrivateConfig` redacts its complete contents so `Payload`/`ActivePayload` can retain useful structural `Debug` output without exposing environment or backend secrets. Added a regression test covering the private key, auth tag, env value, and backend token.

Issue #3619, opened 2026-07-29
test_usage_metrics_lock uses the production advisory-lock key on the shared test DB — fails whenever a local relay is running


tlongwell-block:
## Summary

`tests::test_usage_metrics_lock_has_single_owner_and_releases_on_drop` (`crates/buzz-db/src/lib.rs:5782`) runs against the shared `TEST_DATABASE_URL` database and hardcodes the **production** advisory lock key (`lib.rs:5792`: `0x4255_5A5A_4D45_5452`, identical to `USAGE_METRICS_LOCK_KEY` at `crates/buzz-relay/src/main.rs:84`).

Postgres advisory locks are **per-database**, so any live `buzz-relay` pointed at the same database deterministically holds that key (it re-arms on the `usage_metrics_interval_secs` ~300s tick, `main.rs` ~:1328), and the test reds at the "first database handle becomes leader" assertion. Whether it passes is a race against the relay's tick — several of us burned real diagnosis time on this during #3268 verification (thread cf5deba4 in #buzz-read-only-replica-usage), including one near-miss where the "remedy" almost terminated a 3-day-old live dev relay.

## Proven remedy (verified 2026-07-29 by Dawn, reproduced by Eva and Max)

The test **passes on a private scratch database with the relay still running and still holding its lock** — advisory locks don't collide across databases:

```
TEST_DATABASE_URL=postgres://…/scratch_<uuid> cargo test -p buzz-db --lib -- --ignored \
  --exact tests::test_usage_metrics_lock_has_single_owner_and_releases_on_drop
→ ok. 1 passed
```

## Fix options

1. **Preferred:** have the test create/drop its own scratch DB via the existing `create_scratch_db` helper (`lib.rs:6362`), like the routing tests already do.
2. Alternative: derive the lock key from the database name/test run instead of hardcoding the production constant (keeps shared-DB runs safe but loses the "same key as prod" documentation value — a comment + constant import would preserve it).

## Notes for whoever picks this up

- Never `pg_terminate_backend` an advisory holder without resolving `client_port` → owning process via `lsof` first; an idle holder is likely a live dev relay, not a leaked test backend.
- Related but separate pre-existing flake: `api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo` (Redis-dependent, documented history in git log) — not caused by this lock.

Pre-existing test-isolation bug; predates PR #3268 and is not changed by it.

tlongwell-block:
[...]

PR #4596, opened 2026-08-03
test(db): gate replica routing in CI

### LINKED ISSUE #3622: CI never runs the buzz-db replica-routing Postgres fixtures — enforce them before enabling BUZZ_REPLICA_READ_MAX_AGE_MS
## Summary

None of the Postgres-backed replica-routing fixtures added in #3268 — nor most of the 34 `#[ignore]`d tests in `crates/buzz-db/src/lib.rs` — are ever executed by CI. They were verified exhaustively by hand at `dd26caa9f` (see the PR body), but a future regression on the routed-read path can merge green.

Found by Dawn during her #3268 sign-off amendment; independently confirmed by Max and Eva at the same SHA.

## Evidence (at `dd26caa9f`)

- `Unit Tests` runs `just test-unit` → `cargo nextest run -p buzz-db --lib` (`Justfile:279-285`). Ignored tests are excluded by design, and the Justfile comment says so.
- The only `--run-ignored ignored-only` nextest invocations in `.github/workflows/` are `ci.yml:690` and `:702`, both filtered to `(package(buzz-db) and test(/relay_invite::tests/)) or (package(buzz-relay) and test(/api::invites::tests/))`.
- Therefore, of the 34 `#[ignore]`d buzz-db tests, CI runs only the `relay_invite::tests` subset. The one-budget acquire regression (`routed_fallback_spends_one_acquire_budget_when_aurora_cache_is_cold`), the seven-seam community-isolation fixture, and all fence/floor-guard/fallback fixtures have **zero automated execution**.

## Why this matters

- The one-budget fixture exists precisely because a "fix is present" claim survived review without a measurement (#3268 thread, Dawn's Pass 11 postmortem). An `#[ignore]`d test no job selects is a comment, not a gate.
- Max's rollout condition: **do not set `BUZZ_REPLICA_READ_MAX_AGE_MS` on bb-block until this is fixed.** Merge itself is safe (new seams are dark by default), but enabling the new seams in production without CI enforcement of their regressions is not.

## Constraints on the fix

1. **Ordering dependency on #3619**: the moment any filter widening selects `test_usage_metrics_lock_has_single_owner_and_releases_on_drop`, the shared-DB advisory-lock collision fires in CI. #3619 (scratch-DB isolation for that test) must land first, or the selection must explicitly exclude it.
2. **A naive regex is insufficient.** Dawn's first-draft filter `test(/tests::(routed|replica|reader|fence|floor_guard)/)` misses at least: `query_events_routed_*`, `count_events_routed_*`, `lazy_reader_pool_*`, `head_fetch_routes_by_configured_budget`, `channel_window_routes_*`, `thread_replies_cursor_pages_*`, `read_session_degrades_*`. A partially-covering filter recreates this bug one level down — green CI, unenforced invariants.
3. Several fixtures assume scratch DBs and serial execution; the selection should run serially against the existing `backend-integration` Postgres archive (that job already has the infra).

## Suggested shape (per Max)

An **explicit, auditable selection** rather than a clever regex: either move the replica-routing fixtures into a dedicated test module (e.g. `tests::replica_routing::`) and filter on that module, or maintain an explicit nextest filter list in `ci.yml` with a comment stating it must cover every `#[ignore]`d routing fixture. Initially cover the exact routing matrix verified by hand in #3268. Keep `test_usage_metrics_lock...` excluded until #3619 lands.

## Acceptance

- The one-budget fixture and the seven-seam isolation fixture demonstrably run in CI (visible in the job log at least once).
- A revert of only the #3268 production fix fails CI (the 320ms defang check, now automated).
- No selection of the usage-metrics lock test until #3619 is closed.

Refs: #3268, #3619.

MajorTal:
## Summary

Sorry for the noisy nesting diff. Moving these existing fixtures into a dedicated module adds indentation, but it gives Nextest an explicit, auditable selector that cannot silently miss oddly named routing tests.

- group the five replica-routing Postgres fixtures from the #3268 activation matrix under `tests::replica_routing`
- run that namespace serially from the existing backend integration archive
- keep unrelated ignored database fixtures outside the selector

This turns the hand-verified replica-routing invariants into a CI gate before `BUZZ_REPLICA_READ_MAX_AGE_MS` is enabled.

### Related issue

Fixes #3622

### Testing

- built a `buzz-db` CI-profile archive with the pinned `cargo-nextest 0.9.136`, then ran the exact proposed archived command: 5 passed, 242 skipped
- `TEST_DATABASE_URL=postgres://buzz:buzz_dev@localhost:5432/buzz cargo test -p buzz-db --lib 'tests::replica_routing::' -- --ignored --test-threads=1`: 5 passed
- `cargo test -p buzz-db --lib -- --test-threads=1`: 94 passed, 153 ignored
- `cargo clippy -p buzz-db --lib --tests -- -D warnings`
- `cargo fmt -p buzz-db -- --check`
- pre-push gates: branch skew, all eight Rust unit-test groups, and desktop Tauri clippy/tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -692,6 +692,19 @@ jobs:
             --run-ignored ignored-only
         env:
           DATABASE_URL: postgres://buzz:${{ env.BUZZ_TEST_POSTGRES_PASSWORD }}@localhost:5432/buzz
+      - name: Replica routing database tests
+        # Keep the selection auditable at the source: every ignored fixture in
+        # tests::replica_routing is part of the replica-read activation gate.
+        # Run serially because these Postgres fixtures create and drop scratch
+        # databases on the same server.
+        run: |
+          cargo nextest run \
+            --archive-file target/ci/backend-integration-tests.tar.zst \
+            -E 'package(buzz-db) and test(/replica_routing::/)' \
+            --run-ignored ignored-only \
+            --test-threads 1
+        env:
+          TEST_DATABASE_URL: postgres://buzz:${{ env.BUZZ_TEST_POSTGRES_PASSWORD }}@localhost:5432/buzz
       - name: Workspace profile (kind:9033) gate tests
         # Call-site integration for the 9033 authorization gate: open relay
         # rosterless/steward transitions and the closed-relay admin/owner rule,
--- a/crates/buzz-db/src/lib.rs
+++ b/crates/buzz-db/src/lib.rs
@@ -7282,161 +7282,574 @@ mod tests {
         drop_scratch_db(&admin, writer, &wname).await;
     }
 
-    /// End-to-end deploy-default proof for the NEW routed seams: with the
-    /// budget unset, a covered-eligible query (channel-pinned + `until`)
-    /// through [`Db::query_events_routed`] is served by the WRITER — the
-    /// `for_query` gate keeps the covered arm dark (rev 5). With the budget
-    /// set and a fresh proved entry, the same query routes to the replica.
-    /// Divergent fixtures prove which pool served each read.
-    #[tokio::test]
-    #[ignore = "requires Postgres"]
-    async fn query_events_routed_defaults_dark_and_routes_covered_when_enabled() {
-        let admin = PgPool::connect(&admin_url().await)
-            .await
-            .expect("connect admin");
-        let (writer, wname) = create_scratch_db(&admin, "qer_w").await;
-        let (replica, rname) = create_scratch_db(&admin, "qer_r").await;
+    mod replica_routing {
+        use super::*;
 
-        let author = nostr::Keys::generate();
-        let community = Uuid::new_v4();
-        let channel = Uuid::new_v4();
-        seed_community_channel(&writer, community, channel, &author).await;
-        seed_community_channel(&replica, community, channel, &author).await;
+        /// End-to-end deploy-default proof for the NEW routed seams: with the
+        /// budget unset, a covered-eligible query (channel-pinned + `until`)
+        /// through [`Db::query_events_routed`] is served by the WRITER — the
+        /// `for_query` gate keeps the covered arm dark (rev 5). With the budget
+        /// set and a fresh proved entry, the same query routes to the replica.
+        /// Divergent fixtures prove which pool served each read.
+        #[tokio::test]
+        #[ignore = "requires Postgres"]
+        async fn query_events_routed_defaults_dark_and_routes_covered_when_enabled() {
+            let admin = PgPool::connect(&admin_url().await)
+                .await
+                .expect("connect admin");
+            let (writer, wname) = create_scratch_db(&admin, "qer_w").await;
+            let (replica, rname) = create_scratch_db(&admin, "qer_r").await;
 
-        let base = 1_700_000_000u64;
-        let shared = signed_event_at(&author, "shared", base);
-        for pool in [&writer, &replica] {
-            insert_top_level(pool, community, channel, &shared).await;
+            let author = nostr::Keys::generate();
+            let community = Uuid::new_v4();
+            let channel = Uuid::new_v4();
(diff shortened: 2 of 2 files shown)


MajorTal:
Ran the acceptance criteria from #3622 locally against this branch (Postgres 17.10 in Docker, pinned `cargo-nextest 0.9.136`, the exact `-E` expression from `ci.yml`). Results below, including the one criterion the PR body didn't yet evidence — plus a coverage question I'd like a second opinion on.

## Acceptance criteria

**1. Both named fixtures run** ✅ — the `-E 'package(buzz-db) and test(/replica_routing::/)'` selection resolves to exactly five, including the one-budget and seven-seam isolation fixtures:

```
buzz-db tests::replica_routing::count_events_routed_is_bounded_only
buzz-db tests::replica_routing::lazy_reader_pool_still_spawns_fence_probe
buzz-db tests::replica_routing::query_events_routed_defaults_dark_and_routes_covered_when_enabled
buzz-db tests::replica_routing::routed_fallback_spends_one_acquire_budget_when_aurora_cache_is_cold
buzz-db tests::replica_routing::routed_reads_are_confined_to_the_requested_community
```

**2. "A revert of only the #3268 production fix fails CI"** ✅ — this is the one the PR body didn't cover, so I ran it. Removing only the `Err(sqlx::Error::PoolTimedOut)` arm in `proved_reader` (`lib.rs:897-901`), so a saturated reader falls through to `reader_validation_error`:

```
test ...routed_fallback_spends_one_acquire_budget_when_aurora_cache_is_cold ... FAILED

assertion `left == right` failed: saturated reader must fall back as
writer/reader_acquire_timeout; got {("writer", "reader_validation_error"): 1}
  left: None
 right: Some(1)

test result: FAILED. 4 passed; 1 failed
```

One failure, and the right one — the other four stay green because the mutation doesn't touch what they cover. The gate discriminates rather than merely running. Mutation reverted; nothing in this branch changed.

**3. No selection of the usage-metrics lock test** ✅ — 0 of 5. Also worth noting for anyone re-reading #3622's Constraint 1: **#3619 closed on 2026-07-30**, so the ordering dependency is already satisfied.

## The coverage question

`#3622` Constraint 2 lists seven fixture families a selection must not miss. The module-based selection covers three:

| Family | In gate |
| --- | --- |
| `query_events_routed_*` | ✅ |
| `count_events_routed_*` | ✅ |
| `lazy_reader_pool_*` | ✅ |
| `head_fetch_routes_by_configured_budget` | ❌ |
| `channel_window_routes_*` | ❌ |
| `thread_replies_cursor_pages_*` | ❌ |
| `read_session_degrades_*` | ❌ |

Overall the gate selects 5 of 154 ignored `buzz-db` fixtures. Also still outside it: `read_session_degrades_to_writer_when_replica_connection_dies`, `replica_window_failure_falls_back_to_writer`, `thread_full_replica_page_above_fence_is_reverified_on_writer`, `fence_probe_refuses_to_start_without_verified_floor_guard`, and the `floor_guard_*` fixtures.

I want to be fair about this: #3622 also says "initially cover the exact routing matrix verified by hand in #3268," and this PR scopes itself to that matrix. If those five *are* the matrix, this is correct as scoped and the rest is follow-up work. I couldn't confirm it either way — #3268's body doesn't enumerate the matrix — so this is a question, not a defect claim.

One thing I'd suggest regardless. #3622 asks for "a comment stating it must cover **every** `#[ignore]`d routing fixture." The current comment states the converse:

> every ignored fixture in tests::replica_routing is part of the replica-read activation gate

That guarantees nothing extraneous is in the module, but not that every routing fixture is in it — so a future routing fixture written outside `tests::replica_routing` is silently ungated, which is the same shape as the original bug. Worth either widening the module to the full routing set, or restating the comment as the obligation ("every ignored routing fixture must live in this module") so the next author knows where to put one.

Happy to do either here if you'd like — I have the environment set up and can re-run the revert check against a widened selection.

PR #4598, opened 2026-08-03
fix(relay): multi-value #h filters must not narrow to first channel (#4579)

### LINKED ISSUE #4579: Multi-value `#h` filters silently narrow to the lexicographically first channel, dropping all others

IgorGanapolsky:
## Summary

Fixes #4579: a filter with multiple `#h` channel values was silently answered as if it named only the lexicographically first channel.

## Failure (before)

`extract_channel_id_from_filter` returned the first parseable UUID from `#h`. `nostr::Filter` stores `generic_tags` in a **sorted set**, so multi-channel queries always collapsed to the smallest id. SQL then scoped to that one channel; the Rust post-filter can only remove rows, never restore dropped channels.

Concrete impact: Desktop's global Workflows view batches `kind:30620` with every member channel in `#h`. Workflows living in any channel other than the first alphabetically never appeared.

### Before (unit)

```rust
// multi-#h with a... and b... UUIDs
extract_channel_id_from_filter(&filter) == Some(first)  // wrong
// filter_to_query_params set channel_id = first, channel_ids = None
```

## After

| `#h` values | SQL scope |
|-------------|-----------|
| none | unscoped (access scope may inject full accessible set) |
| one distinct | `channel_id = that id` |
| multiple distinct | `channel_id = None`, `channel_ids = all of them` |

`apply_access_scope_to_query` **intersects** a pre-set multi-`#h` list with the caller's accessible channels instead of replacing it with the full set.

### After (unit)

```text
extract_channel_id_from_filter_multi_h_returns_none ... ok
filter_to_query_params_multi_h_sets_channel_ids_not_first_only ... ok
apply_access_scope_intersects_multi_h_channel_ids ... ok
```

## Verification

```bash
cargo test -p buzz-relay --lib handlers::req::tests
```

**51 passed** (full `handlers::req::tests` module), including the three new regression cases above.

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -268,26 +268,17 @@ pub async fn handle_req(
         .iter()
         .enumerate()
         .map(|(idx, filter)| {
-            // Use per-filter #h channel scope when available, falling back to the
-            // subscription-level channel_id. This prevents unrelated accessible-channel
-            // rows from consuming the LIMIT when filters target specific channels but
-            // the subscription is global (multiple distinct #h values across filters).
-            let per_filter_channel = {
-                let h = nostr::SingleLetterTag::lowercase(nostr::Alphabet::H);
-                filter
-                    .generic_tags
-                    .get(&h)
-                    .and_then(|vs| {
-                        if vs.len() == 1 {
-                            vs.iter().next()?.parse::<uuid::Uuid>().ok()
-                        } else {
-                            None
-                        }
-                    })
-                    .or(channel_id)
-            };
-            let mut params =
-                filter_to_query_params(filter, per_filter_channel, conn.tenant.community());
+            // Per-filter #h scope (single → channel_id, multi → channel_ids).
+            // If the filter names no #h, fall back to the subscription-level
+            // channel_id so global multi-filter REQs still pin single-channel
+            // filters when the subscription itself is channel-scoped.
+            let mut params = filter_to_query_params(filter, conn.tenant.community());
+            if params.channel_id.is_none() && params.channel_ids.is_none() {
+                if let Some(ch) = channel_id {
+                    params.channel_id = Some(ch);
+                }
+            }
+            let per_filter_channel = params.channel_id;
             apply_access_scope_to_query(&mut params, per_filter_channel, &accessible_channels);
             // Shared-gated visibility pushdown: set reader bytes so query_events
             // appends the SQL visibility clause before ORDER/LIMIT, preventing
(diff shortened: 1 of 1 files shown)


IgorGanapolsky:
Superseded by a DCO-signed rewrite (Signed-off-by was missing on the first commit). New PR incoming with the same fix + sign-off.

PR #4599, opened 2026-08-03
fix(desktop): fall back to native image clipboard

### LINKED ISSUE #4592: Linux: pasting an image into the composer freezes or silently does nothing
**Describe the bug**

On Linux, pasting an image from the system clipboard into the message composer does not create an attachment. With the stock AppImage the UI freezes for a while; on one attempt the bundled WebKit web process crashed. After switching the client to native Wayland and also testing the current system WebKitGTK, paste still completed with no image attached.

The same image clipboard content pastes successfully into other Linux applications and into Buzz on macOS.

**Steps to reproduce**

1. Run the Buzz desktop client on a Wayland session.
2. Copy a screenshot or PNG image to the system clipboard.
3. Focus a channel or DM message composer.
4. Press Ctrl+V.

**Expected behavior**

The image is added to the composer as an attachment/preview and the UI remains responsive.

**Actual behavior**

No image is attached. The stock AppImage pauses/freezes; the system journal repeatedly reports that the requested MIME type was not offered. One attempt also crashed the AppImage-bundled `WebKitWebProcess`.

**Version and platform**

- Buzz version: 0.5.3, `Buzz_0.5.3_amd64.AppImage`
- OS: CachyOS, kernel `7.1.5-1-cachyos`
- Desktop session: COSMIC on Wayland
- GPU: AMD Radeon RX 9070 XT (Navi 48)
- System WebKitGTK tested: `webkit2gtk-4.1 2.52.5-2`

**Logs / additional context**

While an image was on the Wayland clipboard, `wl-paste --list-types` advertised `image/png`.

The stock AppImage produces repeated COSMIC/XWayland warnings:

```
cosmic-comp: Mime type requested by X client not offered
```

One failed paste produced this coredump:

```
WebKitWebProcess SIGSEGV
/tmp/.mount_Buzz_*/usr/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitWebProcess
```

I also tested `WEBKIT_DISABLE_DMABUF_RENDERER=1`, native Wayland instead of the AppImage's forced X11/XWayland path, and the official 0.5.4 Debian payload linked to system WebKitGTK. None made image paste work.

This is related to #2338, but is not limited to the dmabuf crash: even when there is no crash, image paste silently produces no attachment.

akalajzi:
Fixes #4592.

## Summary

- Fall back to the native clipboard when Linux WebKitGTK does not expose a pasted image as a `DataTransferItem`.
- Encode the native RGBA image as PNG and feed it into the existing composer attachment/upload path.
- Preserve ordinary text, HTML, code-block, and browser-exposed file paste behavior.

## Validation

- `biome check` on the changed TypeScript files
- `cargo check --manifest-path desktop/src-tauri/Cargo.toml --locked --lib`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --locked --lib clipboard_rgba_is_encoded_as_a_valid_png`
- Manual verification on CachyOS / COSMIC Wayland / AMD RX 9070 XT: copying a PNG to the system clipboard, pasting it in the patched Buzz composer, and sending the resulting image message succeeds.

Related: #2338

### DIFF
--- a/desktop/src-tauri/src/commands/clipboard.rs
+++ b/desktop/src-tauri/src/commands/clipboard.rs
@@ -53,3 +53,25 @@ pub async fn read_clipboard_text(app: tauri::AppHandle) -> Result<String, String
     rx.recv()
         .map_err(|_| "clipboard result channel closed unexpectedly".to_string())?
 }
+
+/// Read the system clipboard image, treating a non-image clipboard as an
+/// expected empty result rather than an error. WebKitGTK does not always
+/// surface Wayland image clipboard offers as `ClipboardEvent` file items.
+pub fn read_system_clipboard_image(
+    app: &tauri::AppHandle,
+) -> Result<Option<arboard::ImageData<'static>>, String> {
+    let state = app.state::<ClipboardState>();
+    let mut stored = state
+        .0
+        .lock()
+        .map_err(|_| "clipboard state lock poisoned".to_string())?;
+    if stored.is_none() {
+        *stored = Some(arboard::Clipboard::new().map_err(|e| format!("clipboard error: {e}"))?);
+    }
+
+    match stored.as_mut().expect("clipboard initialized").get_image() {
+        Ok(image) => Ok(Some(image)),
+        Err(arboard::Error::ContentNotAvailable) => Ok(None),
+        Err(error) => Err(format!("clipboard error: {error}")),
+    }
+}
--- a/desktop/src-tauri/src/commands/media_download.rs
+++ b/desktop/src-tauri/src/commands/media_download.rs
@@ -3,7 +3,7 @@ use sha2::{Digest, Sha256};
 use tauri::State;
 
 use crate::app_state::AppState;
-use crate::commands::clipboard::with_clipboard;
+use crate::commands::clipboard::{read_system_clipboard_image, with_clipboard};
 use crate::commands::export_util::save_bytes_with_dialog;
 use crate::commands::media::{detect_and_validate_mime, mint_media_get_auth, sanitize_filename};
 use crate::commands::{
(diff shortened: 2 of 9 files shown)


albertodvp:
Tested on NixOS unstable with Niri/Wayland and WebKitGTK, using the Tauri desktop-standalone client against a self-hosted relay.  Image paste into the channel/DM composer works correctly.

EmpireOperating:
I tested this clipboard path on Omarchy/Hyprland/Wayland with WebKitGTK 2.52.5 and reproduced the silent image-paste failure from #4592. The native fallback is the right direction.

Live packaged-app testing exposed two additional boundaries:

1. WebKit can consume the paste before ProseMirror's `handlePaste` sees it, so the fallback also needs an editor-DOM capture-phase handler.
2. Raw PNG bytes should use `tauri::ipc::Response`/`ArrayBuffer` rather than a serialized JSON number array.

I also tightened the fallback decision so real files, text, HTML, and advertised non-image formats remain on their normal paths, while an image MIME or completely empty WebKit `DataTransfer` uses the native clipboard.

Rather than open a competing PR against `block/buzz:main`, I based a focused companion PR directly on `cb3859fd` for you to review or cherry-pick:

https://github.com/akalajzi/buzz/pull/1

Validation: focused Node tests, TypeScript typecheck, Biome, Rust format/check, focused PNG test, and a real 1890x1015 Wayland screenshot successfully pasted and uploaded in the packaged app.

EmpireOperating:
Follow-up: the companion PR has been hardened after independent review and contributor feedback. The updated commit enforces decoded and encoded image-size caps, centralizes both native upload paths, surfaces clipboard/IPC/upload failures instead of silently consuming paste, and handles image MIME exposed only through `DataTransferItem.type`. Eleven focused Node tests, two Rust clipboard tests, typecheck, Biome, formatting, and `cargo check --locked --lib` are green. The collaboration URL is unchanged: https://github.com/akalajzi/buzz/pull/1

PR #4602, opened 2026-08-03
fix(desktop): include external channel agents in mentions

cmyk:
## Summary

Buzz Desktop's mention autocomplete excluded valid external agent identities when they were not present in the current user's local managed-agent list. This meant an externally hosted agent could be an active member of the current channel and exchange messages normally, yet remain unavailable from the `@` picker.

This change preserves the managed-agent filtering introduced by #2149 while widening the eligibility gate for current channel members:

- external agents that belong to the current channel can appear in mention autocomplete;
- locally managed agents retain their existing behavior;
- relay-only agents that are not members of the current channel remain hidden;
- the existing invocability and directory-policy checks still run afterward.

The regression coverage also fixes a previously racy negative assertion. The relay-only non-member case now waits for autocomplete processing to complete before verifying that the candidate is absent.

### Related issue

No open duplicate issue or PR found.

Regression follow-up to #2149.

### Testing

Automated verification:

- `pnpm --dir desktop build:e2e` — passed
- focused external-member and relay-only-non-member regressions — 2 passed
- `desktop/tests/e2e/mentions.spec.ts` — 49 passed
- `pnpm --dir desktop check` — passed (two pre-existing informational suggestions only)
- `pnpm --dir desktop test` — 4,022 passed, 0 failed
- `git diff --check` — passed

Red/green regression proof:

- with the broad managed-agent gate removed, the strengthened test failed because the relay-only non-member appeared;
- with the membership-aware gate, the external channel member appears and the relay-only non-member remains absent.

Manual verification against the hosted Buzz community:

- launched the branch as a production-relay development build;
- opened the existing channel containing the external agent;
- confirmed the agent appeared in `@` autocomplete;
- selected the result and confirmed the expected mention was inserted;
- confirmed Buzz retained the agent ownership label.

#### Before

The external channel member was absent from mention autocomplete.

image

#### After

The external channel member appears and is recognized as an agent managed by the current user.

image

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -243,12 +243,12 @@ export function useMentions(
 
     const addCandidate = (candidate: MentionCandidate & { pubkey: string }) => {
       const pubkey = normalizePubkey(candidate.pubkey);
-      if (isArchivedDiscovery(pubkey)) {
-        return;
-      }
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
-        return;
-      }
+      if (isArchivedDiscovery(pubkey)) return;
+      const isManaged = isAgentIdentityInManagedList(
+        candidate,
+        managedAgentPubkeys,
+      );
+      if (candidate.isMember !== true && !isManaged) return;
       if (
         shouldHideAgentFromMentions({
           isAgent: candidate.isAgent === true,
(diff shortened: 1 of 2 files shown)


cmyk:
PR #4913 independently implemented and merged the shared-agent mention functionality, and this older branch now conflicts with the resulting code on main. Live testing confirmed that the merged implementation
covers the original scenario: a remote Hermes channel member appears in @ autocomplete and produces a working mention. Closing this competing implementation rather than resolving conflicts against the
replacement.

PR #4603, opened 2026-08-03
fix: use relay-eligible agents for mention autocomplete

auctobot001:
## Summary
- `useMentions.ts` filtered agent candidates for @-mention autocomplete through `managedAgentPubkeys` — the set of agents *locally managed* by the current device — instead of `mentionableAgentPubkeys`, which already accounts for relay-directory + allowlist eligibility.
- Practical effect: on a multi-device setup, an agent only appears in mention autocomplete on the device that created/manages it. Other devices whose identity is allowlisted and who share the channel can't tag the agent at all, even though the agent is fully eligible per the relay directory (`kind:30177`) and channel membership.
- Swapped the set used in the `isAgentIdentityInManagedList` check from `managedAgentPubkeys` to `mentionableAgentPubkeys`. `managedAgentPubkeys` was then unused in that `useMemo`, so it's dropped from the dependency array (flagged by the exhaustive-deps lint).

## Test plan
- [x] `pnpm typecheck` — clean
- [x] `pnpm exec biome check` on the changed file — clean
- [x] `pnpm test` — all 3870 unit tests pass
- [ ] Manual repro: create a managed agent on device A, allowlist device B's pubkey, add the agent to a shared channel; confirm the agent now appears in device B's mention autocomplete

### DIFF
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -246,7 +246,7 @@ export function useMentions(
       if (isArchivedDiscovery(pubkey)) {
         return;
       }
-      if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
+      if (!isAgentIdentityInManagedList(candidate, mentionableAgentPubkeys)) {
         return;
       }
       if (
@@ -420,7 +420,6 @@ export function useMentions(
     managedAgentNamesByPubkey,
     managedAgentPersonaIds,
     managedAgentPersonaIdsByPubkey,
-    managedAgentPubkeys,
     managedAgentsQuery.data,
     memberPubkeys,
     members,


rsaulo:
We hit this exact bug from a different angle than the multi-device scenario described above — a **fully remote agent** with no managed-agents entry on *any* desktop:

- The agent (Hermes-based, running headless on a VPS) connects to our relay over WS with NIP-OA delegated auth (it is not a direct relay member; the owner is).
- It publishes a relay agent profile with `respond_to: "anyone"` and `channel_ids` covering a channel shared with the user, so `relayAgentIsSharedWithUser` returns `true` and it lands in `mentionableAgentPubkeys`.
- The picker still never shows it: the `isAgentIdentityInManagedList` pre-filter in `addCandidate` drops the candidate before `shouldHideAgentFromMentions` (which would keep it) ever runs. Since the agent is managed on no desktop anywhere, there is no device where it can be mentioned via autocomplete.
- Typing the literal `@Name` as plain text works fine — the downstream matching is healthy, it's only the candidate pre-filter that's wrong.

This also means the "Option B" logic inside `shouldHideAgentFromMentions` is currently dead code for any non-locally-managed agent, which matches what this PR's one-line swap fixes: `mentionableAgentPubkeys` is already the superset that accounts for directory + allowlist eligibility.

We've cherry-picked this diff onto `desktop-v0.5.5` and are running it as a local patch; happy to report back with confirmation from that build. Would love to see this merged — for remote/relay-directory agents there is no workaround on the user side.

rmichelena:

rmichelena:
Cross-linking from #5483. I don't think this one changes behavior on current `main` anymore.

The diff swaps `managedAgentPubkeys` → `mentionableAgentPubkeys` at the `useMentions` gate. #4913 (merged 2026-08-05) already made that call site `isAgentIdentityInAllowedList(candidate, mentionableAgentPubkeys)`, and the function this PR patches, `isAgentIdentityInManagedList`, no longer exists in `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts` on `main`. So as written it's a no-op against today's tree — and the symptom in #5363 survived #4913 anyway.

Why it survived: `list_relay_agents` builds the relay directory from `kind:10100` only, while `respond_to` is published as `kind:30177` and channel membership as `kind:39002`. Agents using the dedicated events arrive with `respond_to: null` / `channel_ids: []`, so the eligibility layer is deciding on absent data no matter which set is passed to the gate. #5483 fixes the directory query.

Would you consider closing this one in favor of #5483? Maintainers' call, of course.

PR #4612, opened 2026-08-03
fix(channels): restrict private-channel invitations

jmecom:
This change requires an active owner or administrator for third-party additions to private channels. The relay validator and transactional database authority enforce the same rule, including removed-member reactivation and role-change paths.

Idempotent self-target behavior remains available, while ordinary members can no longer extend private-channel access to another identity.

## Testing

- `git diff --check origin/main...codex/security-private-channel-invite-authority`
- Full CI pending

### DIFF
--- a/VISION.md
+++ b/VISION.md
@@ -39,7 +39,7 @@ The relay enforces all access control. Channel membership is the only gate.
 | Type | Visibility | Join | Create |
 |------|-----------|------|--------|
 | **Open channels** | Searchable by all members | Self-join | Any member |
-| **Private channels** | Hidden, invite-only | Invited by member | Any member |
+| **Private channels** | Hidden, invite-only | Invited by an owner/admin | Any member |
 | **DMs** | Participants only | N/A (up to 9) | Any member |
 | **Guests** | Scoped to specific channels | Invited | N/A |
 
--- a/crates/buzz-db/src/channel.rs
+++ b/crates/buzz-db/src/channel.rs
@@ -419,10 +419,9 @@ pub async fn add_member(
                 DbError::InvalidData(format!("invalid role in database: {inviter_role_str}"))
             })?;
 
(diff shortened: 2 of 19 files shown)


elifoster-block (dismissed):
Client-side follow-up to Jordan's commits: hide the "add people" and invite buttons when you're not an owner/admin of a private channel, and stop mobile from silently dropping your message when the relay refuses the add.

elifoster-block (approved):

wesbillman (commented):

I found no blocking issue at head `7364c7d77353234aa468535054876d4471929836`.

The authorization boundary is enforced twice and the database remains authoritative. The relay validator rejects a private-channel third-party add unless the signed actor is an active owner/admin, giving clients an immediate rejection. `buzz_db::channel::add_member` repeats the check inside the existing per-channel advisory-locked transaction, so alternate callers and role/removal races cannot bypass it.

I traced the self-target exception. It only preserves an active member’s idempotent re-add: a non-member/soft-removed identity fails the active-role lookup, and an active member cannot use it to change role because the separate role-change guard requires owner/admin. Private kind:9021 self-join remains rejected before this path.

Owner/admin additions, creator bootstrap, open-channel behavior, and existing huddle/ephemeral flows remain intact. Desktop and mobile mirror the authority only for UX; the relay/DB checks do not trust client state. Mobile now reports partial add failures and demotes identities that were not added to non-notifying reference mentions rather than silently losing the message.

Tests cover regular-member denial, admin invitation, idempotent self-targeting, client capability mirrors, and the mobile rejected-add flow. Exact-head CI is green, including Rust/unit, mobile, security, desktop and relay/backend E2E, cross-compiles, DCO, Semgrep, and zizmor.

Expected behavior change: ordinary members can no longer invite others into private channels. That is the intended security fix, not a regression.

wesbillman (approved):

PR #4613, opened 2026-08-03
fix(db): bound runtime Postgres statements and lock waits

jmecom:
## The short version

Buzz talks to its database through a small, fixed set of connections — think of
them as a handful of phone lines shared by everyone using the app. Before this
change, there was no limit on how long a single call could tie up a line.

This PR adds time limits so no request can hold a line forever.

## Why that mattered

Someone using Buzz can ask for things like "search everything" or "give me all
messages matching this." Those requests are tiny to send but can be very
expensive for the database to answer.

Without a limit, one of those requests grabs a phone line and keeps it for as
long as the database is willing to work — which is indefinitely. Do that a few
times at once and every line is busy. At that point *nothing* works: not logging
in, not sending messages, not the health checks that tell us the server is
alive. And the people affected did nothing wrong; they just wanted the lines
that are now occupied.

Fixing it meant an engineer manually finding and killing the stuck queries.

## What the PR adds

**Three limits, applied automatically:**

- A query gets **30 seconds** to finish, or the database cancels it.
- A query waiting for another query to release something gets **5 seconds**,
  then gives up.
- A connection that starts a piece of work and then just sits there gets
  **60 seconds** before it's reclaimed. The first limit only counts time spent
  actively working, so it can't see this case — the connection is held, and so
  is everything it already reserved, while nothing is running.

They're attached the moment a connection is created, so every database call is
covered — nobody has to remember to add them.

Operators can change any of the three without a code release, and set them to
`0` to turn them off during an incident.

## The two tricky bits

**Database upgrades are exempt.** When Buzz starts, it sometimes updates the
database's structure, which can legitimately take minutes. If that hit the
30-second limit it would be cancelled — and a failed upgrade stops the server
from starting at all. So upgrades run on one special connection with the limits
removed. That connection is then permanently retired, because a connection with
no limits must never end up serving normal traffic.

**A typo in the setting can't take the server down.** The limits are applied to
*every* connection, so if someone sets an invalid value, every connection fails
and the database becomes unreachable — the exact outage this was meant to
prevent. So the code checks the value up front and falls back to the default if
it's wrong.

Getting that check right took four rounds of review, which found four different
ways a bad value could slip through. The fix in the end wasn't a fourth patch —
it was making the setting a dedicated type that can *only* be created by passing
the check. Now an invalid value can't exist in the first place, rather than
existing and hopefully being caught.

**One related fix:** the admin command-line tool was quietly picking up the same
30-second limit, even though it's a maintenance tool run by hand — exactly where
a long-running job is normal and expected, and where there are no shared lines
to protect. It now runs without limits by default.

## Testing

Ordinary tests cover the settings parser and the environment-variable wiring.

Tests against a real PostgreSQL prove the parts that can't be checked in
isolation:

- every value the parser accepts is actually usable by the server
- the maximum allowed value matches the server's real maximum, from both
  directions, so it can't silently drift
- every connection pool really does carry all three limits
- the upgrade connection really does have them removed, and is never handed back
  for normal use
- the slow startup scan really does exceed a tight limit, and really does
  survive on the exempt connection

Every one of these was verified by deliberately breaking the code and confirming
the test catches it.

Local run at `badc7784e`: 97 unit tests pass, all database-backed tests pass,
lint and formatting clean.

Note: the database-backed tests are marked `#[ignore]` and CI only runs them
when explicitly asked, so they're run on demand rather than on every PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -40,7 +40,9 @@ REDIS_URL=redis://localhost:6379
 
 # Postgres statement_timeout and lock_timeout applied to every runtime
 # connection. Accepts an integer (milliseconds) with an optional us/ms/s/min/h/d
-# unit; `0` disables the limit. Schema migrations always run with both lifted.
+# unit; `0` disables the limit. Postgres stores both as int milliseconds, so the
+# ceiling is 2147483647ms (~24 days); anything above it, or otherwise malformed,
+# falls back to the default. Schema migrations always run with both lifted.
 # BUZZ_DB_STATEMENT_TIMEOUT=30s
 # BUZZ_DB_LOCK_TIMEOUT=5s
 
--- a/.env.example
+++ b/.env.example
@@ -38,6 +38,12 @@ REDIS_URL=redis://localhost:6379
 # READ_DATABASE_URL is set, reader (default 50).
 # BUZZ_DB_POOL_SIZE=50
 
+# Postgres statement_timeout and lock_timeout applied to every runtime
+# connection. Accepts an integer (milliseconds) with an optional us/ms/s/min/h/d
+# unit; `0` disables the limit. Schema migrations always run with both lifted.
+# BUZZ_DB_STATEMENT_TIMEOUT=30s
+# BUZZ_DB_LOCK_TIMEOUT=5s
+
 # -----------------------------------------------------------------------------
 # Typesense (search)
 # -----------------------------------------------------------------------------
--- a/.env.example
+++ b/.env.example
@@ -38,13 +38,19 @@ REDIS_URL=redis://localhost:6379
 # READ_DATABASE_URL is set, reader (default 50).
 # BUZZ_DB_POOL_SIZE=50
 
-# Postgres statement_timeout and lock_timeout applied to every runtime
-# connection. Accepts an integer (milliseconds) with an optional us/ms/s/min/h/d
-# unit; `0` disables the limit. Postgres stores both as int milliseconds, so the
-# ceiling is 2147483647ms (~24 days); anything above it, or otherwise malformed,
-# falls back to the default. Schema migrations always run with both lifted.
+# Per-session Postgres limits applied to every runtime connection, so no single
+# caller can hold a pooled connection indefinitely. Each accepts an integer
+# (milliseconds) with an optional us/ms/s/min/h/d unit; `0` disables that limit.
+# Postgres stores all three as int milliseconds, so the ceiling is 2147483647ms
+# (~24 days); anything above it, or otherwise malformed, falls back to the
+# default. Schema migrations always run with all three lifted.
+#
+# IDLE_IN_TRANSACTION covers only the gaps between a transaction's statements —
+# it reclaims a session that opened a transaction and stopped working, which
+# STATEMENT_TIMEOUT cannot see because no statement is running.
 # BUZZ_DB_STATEMENT_TIMEOUT=30s
 # BUZZ_DB_LOCK_TIMEOUT=5s
+# BUZZ_DB_IDLE_IN_TRANSACTION_TIMEOUT=60s
 
 # -----------------------------------------------------------------------------
 # Typesense (search)
--- a/crates/buzz-admin/src/main.rs
+++ b/crates/buzz-admin/src/main.rs
@@ -25,7 +25,7 @@ use std::sync::Arc;
 use anyhow::Result;
 use buzz_core::kind::KIND_NIP43_MEMBERSHIP_LIST;
 use buzz_core::tenant::{relay_url_authority, TenantContext};
-use buzz_db::{Db, DbConfig, PgTimeout};
+use buzz_db::{Db, DbConfig, RuntimeTimeouts};
 use buzz_pubsub::{EventTopic, PubSubManager};
 use clap::{Parser, Subcommand};
 use nostr::{EventBuilder, Keys, Kind, Tag};
@@ -426,11 +426,7 @@ async fn connect_db() -> Result<Db> {
     // env vars still apply for an operator who wants a bound here.
     let db = Db::new(&DbConfig {
         database_url: db_url,
-        statement_timeout: PgTimeout::from_env_or(
-            "BUZZ_DB_STATEMENT_TIMEOUT",
-            PgTimeout::disabled(),
-        ),
-        lock_timeout: PgTimeout::from_env_or("BUZZ_DB_LOCK_TIMEOUT", PgTimeout::disabled()),
+        timeouts: RuntimeTimeouts::from_env_or(RuntimeTimeouts::disabled()),
         ..DbConfig::default()
     })
     .await?;
--- a/crates/buzz-admin/src/main.rs
+++ b/crates/buzz-admin/src/main.rs
@@ -25,7 +25,7 @@ use std::sync::Arc;
 use anyhow::Result;
 use buzz_core::kind::KIND_NIP43_MEMBERSHIP_LIST;
 use buzz_core::tenant::{relay_url_authority, TenantContext};
-use buzz_db::{Db, DbConfig};
+use buzz_db::{Db, DbConfig, PgTimeout};
 use buzz_pubsub::{EventTopic, PubSubManager};
 use clap::{Parser, Subcommand};
 use nostr::{EventBuilder, Keys, Kind, Tag};
@@ -420,8 +420,17 @@ async fn connect_member_services() -> Result<(Db, Arc<PubSubManager>, Keys)> {
 async fn connect_db() -> Result<Db> {
     let db_url = std::env::var("DATABASE_URL")
         .unwrap_or_else(|_| "postgres://buzz:buzz_dev@localhost:5432/buzz".to_string());
+    // No runtime caps by default. `DbConfig`'s 30s/5s are sized for the relay's
+    // request-serving traffic on a shared pool — an operator CLI is neither, and
+    // a bulk repair that outlives 30s should finish, not get cancelled. The same
+    // env vars still apply for an operator who wants a bound here.
     let db = Db::new(&DbConfig {
         database_url: db_url,
+        statement_timeout: PgTimeout::from_env_or(
+            "BUZZ_DB_STATEMENT_TIMEOUT",
+            PgTimeout::disabled(),
+        ),
+        lock_timeout: PgTimeout::from_env_or("BUZZ_DB_LOCK_TIMEOUT", PgTimeout::disabled()),
         ..DbConfig::default()
     })
     .await?;
--- a/crates/buzz-db/src/lib.rs
+++ b/crates/buzz-db/src/lib.rs
@@ -71,6 +71,13 @@ pub const RUNTIME_STATEMENT_TIMEOUT: &str = "30s";
 pub const RUNTIME_LOCK_TIMEOUT: &str = "5s";
 /// Postgres spelling of "no limit", used for schema migrations.
 pub const TIMEOUT_DISABLED: &str = "0";
+/// `statement_timeout` and `lock_timeout` are `int` GUCs measured in
+/// milliseconds, so Postgres refuses anything larger regardless of the unit it
+/// is spelled with. Callers building a [`DbConfig`] from operator input must
+/// range-check against this: [`apply_runtime_connection_timeouts`] runs on every
+/// pooled connection, so an unusable value fails all database access.
+/// `pg_timeout_max_millis_matches_postgres` pins it to the live server.
+pub const PG_TIMEOUT_MAX_MILLIS: u128 = i32::MAX as u128;
 
 /// Apply the runtime safety limits shared by writer, reader, audit, and search
 /// pools. Values are Postgres interval strings (`"30s"`, `"500ms"`), with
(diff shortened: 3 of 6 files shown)


elifoster-block (dismissed):
Excluded migrations from the timeouts and set the timeouts as a DBconfig option.

wesbillman (changes requested):
Reviewing on Wes's behalf.

**Blocking regression:** the migration exemption starts too late. `run_migrations` calls `reject_legacy_nip_rs_cardinality_ambiguity(pool)` before `run_migrator_without_runtime_timeouts` acquires and relaxes its connection. On a pre-v7 database, that preflight executes a potentially large scan of `events` (including per-row JSON expansion), so it inherits the new 30s `statement_timeout`. If it exceeds the cap, startup treats the error as fatal and the relay cannot boot—the exact regression the migration exemption is intended to prevent.

Please acquire the dedicated migration connection and lift both limits before running the legacy preflight, then run the migrator on that same connection and retire it on every normal success/error path. The preflight can still remain before sqlx begins its migration transaction. Add a Postgres-backed regression test that makes the preflight exceed a tight runtime timeout while confirming it completes on the exempt connection; the current fresh-database migration test returns before this scan and cannot catch this ordering bug.

Non-blocking hardening note: the relaxed `PoolConnection` is returned to the runtime pool if the migration future is cancelled after `lift_runtime_timeouts` but before explicit retirement. The current relay startup awaits migration directly and normally exits on shutdown, which narrows the practical exposure, but making the exemption structurally cancellation-safe would better preserve the stated invariant that an unlimited session can never serve runtime traffic.

elifoster-block:
🤖 Both review points addressed in c1277d268.

**Blocking: preflight ordering.** Confirmed — `run_migrations` ran `reject_legacy_nip_rs_cardinality_ambiguity(pool)` before `run_migrator_without_runtime_timeouts` acquired anything, so on a pre-0007 database the `events` scan inherited the runtime `statement_timeout` and a slow scan became a fatal startup failure. The exempt connection is now acquired and relaxed first, and both the preflight and `MIGRATOR.run` execute on it. The preflight still runs ahead of sqlx's migration transaction, so an operator can inspect and repair before any DDL.

`verify_floor_guard_catalog` stays on the pool — it's catalog-only, bounded by partition count, and shared with the non-migrating startup probe.

**Hardening: cancellation safety.** Made structural rather than narrowed. `exempt_migration_connection` calls `.detach()` *before* lifting the limits, so the value is a `PgConnection`, not a `PoolConnection`. A `PoolConnection` releases itself to the pool on drop, which is the leak you described; a detached connection closes on drop. There is no longer a window where a cancelled future can return an unbounded session — the explicit `close()` is now just eager cleanup and error reporting, not the invariant. `migration_connection_is_unbounded_and_is_retired_not_reused` was updated to drop without closing (standing in for the cancellation) and still asserts the single-slot pool hands out a freshly capped connection.

**New regression test:** `legacy_preflight_scan_outlives_the_runtime_statement_timeout`. It applies migrations 1–6, then seeds conforming kind-30078 read-state rows in batches, timing the preflight on a capped pooled connection after each batch until Postgres cancels it with 57014 — so the row count is calibrated at runtime instead of hardcoded to whatever is slow on one machine. It then runs `run_migrations` against that same capped pool, asserts it reaches version 26, and asserts the pool still hands out capped sessions afterwards. Rows are deliberately non-ambiguous so the `EXISTS` cannot short-circuit early.

Verified discriminating: with the preflight moved back onto a pooled connection, the test fails with `Sqlx(Database(PgDatabaseError { code: "57014", message: "canceling statement due to statement timeout" }))`.

## Testing

- `cargo test -p buzz-db --lib`: 94 passed, 156 ignored
- `cargo test -p buzz-db --lib migration:: -- --ignored`: all 5 pass, including the new test (calibration settled at 20k rows, 3.3s total)
- `cargo test -p buzz-db --lib migrations_ignore_runtime_timeouts -- --ignored` and `... armed_pool -- --ignored`: pass
- `cargo clippy -p buzz-db -p buzz-relay --all-targets -- -D warnings` and `cargo fmt -- --check`: clean

Same caveat as before: the Postgres-backed tests are `#[ignore]` and CI selects ignored tests by explicit filter, so they run on demand rather than per-PR.

wesbillman (changes requested):
Reviewing on Wes's behalf.

The migration-ordering and cancellation-safety blockers are fixed at `c1277d268`: preflight and migrator share a detached, timeout-exempt connection, and detaching before relaxation prevents an unlimited session from returning to the runtime pool.

One operational blocker remains in the new timeout configuration. `pg_timeout_or_default` accepts any non-empty ASCII digit string followed by a recognized unit but never parses or bounds the magnitude. For example, `BUZZ_DB_STATEMENT_TIMEOUT=999999999999999999999999999999999999999999d` passes validation unchanged; PostgreSQL then rejects it when `set_config` runs. Because the timeout setup is installed in every writer, reader, audit, and search pool's `after_connect`, one superficially "valid" but out-of-range value can prevent all database connections and fail relay startup—the exact failure this fallback helper says it prevents.

Please parse and range-check the magnitude (including unit conversion against PostgreSQL's supported timeout range), fall back on overflow/out-of-range values, and add regression coverage proving such a value never reaches `after_connect` unchanged.

elifoster-block:
🤖 Remaining blocker addressed in 37cb4195d.

**Out-of-range magnitudes.** Confirmed — `pg_timeout_or_default` checked shape only, so `BUZZ_DB_STATEMENT_TIMEOUT=999999999999999999999999999999999999999999d` passed through unchanged and Postgres rejected it inside `after_connect`, failing every writer/reader/audit/search connection and relay startup.

The magnitude is now parsed and converted to milliseconds by unit (`us`/`ms`/`s`/`min`/`h`/`d`, with `us` rounded the way Postgres rounds it), and anything that overflows or exceeds the ceiling falls back to the documented default with a warning. A digit string too wide for any integer type fails the parse rather than wrapping.

**The ceiling is pinned, not asserted from memory.** `statement_timeout` and `lock_timeout` are `int` GUCs in milliseconds, so the bound is `i32::MAX` ms (~24.8 days). It lives in `buzz-db` as `PG_TIMEOUT_MAX_MILLIS`, next to `RUNTIME_STATEMENT_TIMEOUT` and `apply_runtime_connection_timeouts`, and `pg_timeout_max_millis_matches_postgres` sets it against a live server from both sides: the boundary (and the same instant spelled in seconds) must be settable, and boundary+1 in each unit must fail with `22023`. Too high a constant and an "accepted" value still breaks `after_connect`; too low and we refuse settings Postgres would take. Neither direction can drift silently.

**Coverage.**

- `pg_timeout_refuses_magnitudes_postgres_cannot_store` — `2147483648`, `2147483648ms`, `2147484s`, `35792min`, `597h`, `25d`, and both over-wide digit strings fall back; the boundary in every unit (`2147483647`, `2147483647000us`, `2147483s`, `35791min`, `596h`, `24d`) stays valid.
- `db_timeout_env_overrides_and_invalid_fallback` — extended to prove an out-of-range env value never reaches `DbConfig`, which is what keeps it out of `after_connect`.
- `pg_timeout_max_millis_matches_postgres` (Postgres-backed) — the drift test above, plus a check that a rejected `set_config` leaves the session's existing limit intact rather than disabling it.

`.env.example` now documents the ceiling alongside the accepted units.

## Testing

- `cargo test -p buzz-db --lib`: 94 passed, 157 ignored
- `cargo test -p buzz-db --lib -- --ignored --test-threads=1` for `migration::` and `armed_pool`: all pass, including the new drift test. Run in parallel, four migration tests fail on scratch-database contention against my local Postgres — pre-existing and unrelated; each passes in isolation.
- `cargo clippy -p buzz-db -p buzz-relay --all-targets -- -D warnings` and `cargo fmt --all -- --check`: clean

Same caveat as before: the Postgres-backed tests are `#[ignore]` and CI selects ignored tests by explicit filter, so they run on demand rather than per-PR.

wesbillman (changes requested):
Reviewing on Wes's behalf at `37cb4195d19142c4a58934fee17472282c2290c5`.

The intended magnitude guard and migration invariant are otherwise sound, but one representable boundary still bypasses the guard:

**[P1] Avoid overflowing the microsecond rounding addition**

`pg_timeout_millis` parses the magnitude as `u128`, then evaluates `(value + 500) / 1_000` for `us`. Inputs from `u128::MAX - 499` through `u128::MAX` therefore overflow after parsing successfully. In checked/debug execution, config loading panics. With wrapping release arithmetic, the conversion produces a tiny millisecond value that passes the `PG_TIMEOUT_MAX_MILLIS` check, after which `pg_timeout_or_default` returns the original gigantic candidate and every pool's `after_connect` hands it to PostgreSQL—the startup failure this guard is intended to prevent.

The new over-wide tests use values larger than `u128`, so they safely fail parsing and do not exercise this edge. Please use checked arithmetic or an overflow-free equivalent such as `value / 1_000 + u128::from(value % 1_000 >= 500)`, and add `format!("{}us", u128::MAX)` as a regression proving it falls back rather than panicking or passing through.

The rest of the fix checks out: representable values above PostgreSQL's `i32::MAX`-millisecond ceiling now fall back; the live PostgreSQL test covers the exact ceiling; and the detached migration connection remains exempt for both preflight and migrator without returning an unlimited session to the runtime pool.

wesbillman (changes requested):
Requesting changes on Wes's behalf at `637a3231c5684d482aefb10999dbdc276db30f14`.

The prior migration-ordering, cancellation-safety, magnitude-range, and `u128` overflow blockers are fixed. One startup blocker remains in the parser's case handling:

**[P1] Return the normalized timeout unit instead of the rejected original spelling**

`pg_timeout_or_default` lowercases `unit` only for validation, then returns `candidate.to_string()` unchanged (`crates/buzz-relay/src/config.rs:344-349`). Consequently the test at lines 1277-1284 treats `45S` as accepted and returns `45S`, but PostgreSQL 17 rejects `set_config('statement_timeout', '45S', false)` with SQLSTATE `22023` and the hint that valid units are lowercase `us/ms/s/min/h/d`. That value then reaches every writer/reader/audit/search pool's `after_connect`, preventing all database connections and relay startup—the exact failure the fallback is intended to prevent.

I reproduced this against the workspace's `postgres:17-alpine` container at the reviewed head:

```text
ERROR: invalid value for parameter "statement_timeout": "45S"
HINT: Valid units for this parameter are "us", "ms", "s", "min", "h", and "d".
```

Please either reject uppercase units to the default or return a canonical string built from `magnitude` plus the lowercased unit. Add a Postgres-backed assertion (or make the existing live boundary test exercise parser output) so every spelling the parser accepts is proven usable by `set_config`; update the current unit test, which presently codifies the invalid pass-through.

The remaining implementation checks out in my pass: all four runtime pool classes receive both limits; the migration preflight and migrator share a detached exempt connection; cancellation cannot return that connection to the pool; overflow-free microsecond conversion and the `i32::MAX`-millisecond guard close the previously reported magnitude cases. Existing CI is green at this exact head; I did not duplicate the broad suites locally.

elifoster-block:
🤖 The P1 was already fixed in `6c436f406`, pushed after your review at `637a3231c` — flagging it here since no comment went with the push. Verified it end to end, plus one test-rot fix in `363d1b16f`.

**[P1] Uppercase unit spellings.** Confirmed and fixed. `pg_timeout_or_default` now returns `format!("{magnitude}{unit}")` from the lowercased unit it validated, instead of `candidate.to_string()`. `45S` → `45s`, `2MIN` → `2min`, ` 10 H ` → `10h`. The unit test that codified the invalid pass-through is updated to assert the canonical form.

Coverage is Postgres-backed as you asked: `pg_timeout_canonical_spellings_are_accepted_by_postgres` feeds every spelling the parser emits (`500US`, `500MS`, `2S`, `2MIN`, `2H`, `2D`, `0`, ` 10 S `) through `apply_runtime_connection_timeouts` on a live connection, so anything the parser accepts is proven settable rather than just proven well-shaped.

Verified discriminating: reverting the one line to `candidate.to_string()` fails the test.

**Test rot (`363d1b16f`).** Running the ignored migration suite locally surfaced two failures unrelated to your review — `legacy_preflight_scan_outlives_the_runtime_statement_timeout` (mine, `Some(26)`) and `pre_0007_ambiguous_nip_rs_data_blocks_without_mutation_and_allows_retry` (pre-existing on `main`, `Some(27)`) both assert a hardcoded migration head version, and `main` is now at 28. Both now read the head from `MIGRATOR`, matching the existing precedent elsewhere in the file.

ThermoclineLeviathan:
Nice — the migration carve-out with the "relaxed session must not survive into the pool" test is a good touch.

One coverage note: `buzz-push-gateway` builds its own pool in `crates/buzz-push-gateway/src/main.rs:56` and doesn't go through `Db::connect_pool`, so it won't pick up `apply_runtime_connection_timeouts` from this change:

```rust
let pool = sqlx::postgres::PgPoolOptions::new()
    .max_connections(20)
    .connect(&c.database_url).await?;
```

Besides the statement/lock timeouts this PR adds elsewhere, that pool also sets no `acquire_timeout`, `min_connections`, `max_lifetime`, or `idle_timeout`, and `max_connections` is hardcoded rather than read from `Config`. It's a network-facing service — its own listener, with the router at `http.rs:748-752` applying `ConcurrencyLimitLayer::new(256)` and a 20s `TimeoutLayer` — so the concurrency ceiling sits well above the pool size.

Might be worth extending `runtime_timeout_hook` to it, or filing a follow-up. Not blocking this PR either way.

PR #4614, opened 2026-08-03
fix(desktop): enable the content security policy

jmecom:
This change enables a Tauri content security policy that limits executable content to the packaged application and does not allow inline scripts.

Relay, media, asset, and Tauri IPC schemes remain available for desktop compatibility. The policy contains the impact of a future renderer injection; it does not itself remove an injection bug.

## Testing

- `git diff --check origin/main...codex/security-desktop-csp`
- Full CI pending


elifoster-block (dismissed):
Added tests and some extra additions in the CSP to cover items that would break without it.

wesbillman (dismissed):
Reviewing on Wes's behalf.

**Blocking security issue:** `script-src` trusts the entire `https://cdn.jsdelivr.net` origin. jsDelivr serves arbitrary npm and GitHub content, so any renderer injection that can create an external `<script>` can choose attacker-controlled JavaScript from that origin and recover code execution. That substantially defeats this PR's stated goal of limiting executable content to the packaged application plus the one required MediaPipe loader.

Please narrow the source to the exact pinned loader URL (`https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@0.10.35/wasm/vision_wasm_internal.js`) or package that small loader locally while continuing to fetch the large WASM binary under `connect-src`. `FilesetResolver.forVisionTasks` resolves this exact JS loader plus `vision_wasm_internal.wasm`; there is no need to authorize every script hosted by jsDelivr. The CSP regression test should pin the exact allowed script URL and reject the bare CDN origin so future edits cannot silently reopen the executable allowlist.

The compatibility allowances otherwise line up with the inspected packaged-app paths: inline bootstrap hashing, Shiki/MediaPipe WASM, Tauri IPC, cold-start `buzz-media`, cleartext user-configured relay probes, local dynamic chunks, and the self-hosted audio worklet. CI is green, but as the PR notes, the current browser E2E shape does not enforce the packaged CSP; that makes keeping the static policy test security-specific rather than merely presence-based especially important.

elifoster-block:
🤖 Addressed the blocking `script-src` issue. `script-src` no longer trusts the jsDelivr origin — it now allowlists a **path prefix scoped to the `@mediapipe` npm scope**:

```
script-src 'self' 'wasm-unsafe-eval' https://cdn.jsdelivr.net/npm/@mediapipe/
```

CSP can't wildcard a path, but a source ending in `/` is a prefix match, so this admits any file in any `@mediapipe` package — a scope only Google can publish to — while excluding the rest of npm and all of `/gh/`. Reasoning for preferring it over the exact loader URL you suggested:

- Pinning exact URLs needs *two* of them, not one: `FilesetResolver.forVisionTasks` runs a wasm-SIMD probe at runtime and appends `vision_wasm_nosimd_internal.js` when SIMD is unavailable, so allowlisting only `vision_wasm_internal.js` would break segmentation on hosts without it.
- It also puts the `@mediapipe/tasks-vision` version in the CSP, so the policy has to move in lockstep with every dependency bump — with the packaged app as the only place a mismatch surfaces.
- Per CSP3, path components are not re-checked after a redirect, so exact-URL pinning already degrades to origin-level given any redirect primitive on the CDN. The practical gap between "one file" and "one publisher scope" is smaller than it looks — and neither is close to the bare origin you flagged.

Happy to switch to the exact URLs if you'd rather have the tighter allowlist; it's a one-line change plus a test.

Guards in `desktop/src-tauri/tests/csp.rs` (moved out of `src/csp_tests.rs` — the app crate was over the file-size ratchet):

- `script_src_scopes_the_mediapipe_loader` — the prefix is present, and `MEDIAPIPE_WASM_BASE` in `animatedAvatarCapture.ts` sits under it.
- `script_src_trusts_no_bare_origins` — every remote `script-src` source must have a non-wildcard host and at least two path segments.
- `pinned_script_source_rejects_broad_sources` — tests that rule directly against `https://cdn.jsdelivr.net`, `.../npm/`, `.../gh/`, `https://*.jsdelivr.net/npm/@mediapipe/`, `https:`, and `*`, so it can't quietly become a no-op.

`main` is merged in and the conflict in `desktop/package.json` is resolved. `just desktop-tauri-test`, `desktop-tauri-clippy`, `desktop-tauri-fmt-check`, and `pnpm check` pass locally.

wesbillman (dismissed):
Reviewing on Wes's behalf.

The bare-origin escape hatch is gone, but this is still broader than the executable dependency the app uses: `https://cdn.jsdelivr.net/npm/@mediapipe/` authorizes JavaScript from every present or future package and version in the publisher scope. A compromised publisher account—or a newly published package under that scope—would therefore remain a CSP-permitted payload for any renderer injection. That is not equivalent to trusting the two pinned loader artifacts required by `tasks-vision@0.10.35`.

Please use the two exact loader sources:

- `https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@0.10.35/wasm/vision_wasm_internal.js`
- `https://cdn.jsdelivr.net/npm/@mediapipe/tasks-vision@0.10.35/wasm/vision_wasm_nosimd_internal.js`

I confirmed both currently return 200 directly without redirects. The WASM files remain covered by `connect-src https:`. Keeping the CSP version in lockstep with `MEDIAPIPE_WASM_BASE` is desirable here: a dependency bump should fail a static guard/build check until the executable allowlist is deliberately reviewed. The redirect rule cited in the response does not make an exact initial source equivalent to a publisher prefix; path matching still constrains the initial request, and these loader URLs do not currently redirect.

Please also make the regression guard assert those exact two sources and reject the publisher-scope prefix. The rest of the CSP compatibility allowances and no-unsafe-inline/eval guards look sound.

wesbillman:
Additional compatibility finding from the independent desktop review: this head restores `"@mediapipe/tasks-vision": "^0.10.35"` and removes the guard that kept the installed JS package version aligned with the hardcoded CDN WASM base (`@0.10.35`). The lockfile resolves 0.10.35 today, so the current build works, but a future in-range lockfile update can pair newer JS glue with the pinned 0.10.35 WASM assets and silently break animated-avatar segmentation.

While tightening `script-src`, please restore that version-alignment invariant—either exact-pin the package or retain a test that compares the resolved/declared package version with `MEDIAPIPE_WASM_BASE`. This is a compatibility issue rather than the security blocker above, but fixing both in the same policy revision avoids leaving a small time bomb with excellent posture.

wesbillman (commented):
Reviewing on Wes's behalf at `c0ba9461d83b9e83154426a26594107ae7e1af63` after reconsidering the maintainer's stated security/maintenance tradeoff.

I am withdrawing the exact-version CSP blocker. The current policy no longer trusts the bare jsDelivr origin: `script-src` is restricted to the `https://cdn.jsdelivr.net/npm/@mediapipe/` publisher namespace, and the static tests reject broader npm/GitHub paths, wildcard hosts, and bare-origin forms. That is a substantial improvement over today's unrestricted executable policy. Exact artifact pinning would reduce publisher-account blast radius further, but it would also couple CSP to dependency updates and create a credible bump-without-CSP-update availability failure. The maintainer has explicitly accepted that residual publisher-scope risk in favor of avoiding that drift, and the chosen boundary is coherent.

The installed JS and runtime WASM currently agree at `0.10.35`: the lockfile resolves the package exactly, while `MEDIAPIPE_WASM_BASE` pins the matching assets. A future lockfile regeneration could still update one without the other because the static test does not derive or compare these versions. I consider adding a version-coupling guard worthwhile follow-up hardening, but not a blocker for this CSP introduction.

All checks are green at this head. No remaining actionable finding from my review.

wesbillman (approved):
Approving on Wes's behalf at exact head `c0ba9461d83b9e83154426a26594107ae7e1af63`. The accepted CSP policy scopes jsDelivr to the Google-published `@mediapipe/` namespace without coupling CSP maintenance to dependency-version bumps. The committed lockfile and WASM URL currently agree at 0.10.35, broader CDN forms are rejected by tests, and CI is green. Residual publisher-account risk is understood and accepted.

PR #4622, opened 2026-08-03
feat(cli): expose channel visibility on `channels update`

rodroudi:
`build_update_channel` already accepts a visibility argument, but `buzz channels update` hardcoded `None`, so there was no way to convert a channel between open and private from the CLI — the only surface was the desktop app.

Adds `--visibility open|private`, threads it through `cmd_update_channel`, and includes it in the "at least one field required" guard.

```
buzz channels update --channel <uuid> --visibility private
```

The relay side is unchanged: kind:9002 already validates the `visibility` tag and authorizes it as a privileged metadata edit (channel owner/admin), so this only closes a gap in the client surface.

## Testing

- `cargo test -p buzz-cli` — 317 passed, 0 failed
- `cargo clippy -p buzz-cli --all-targets -- -D warnings` — clean
- `cargo fmt` — clean
- Verified end to end against a live relay: converting an open channel to private takes effect immediately and the resulting system message / cache invalidation behave the same as a desktop-initiated flip.

### DIFF
--- a/crates/buzz-cli/src/commands/channels.rs
+++ b/crates/buzz-cli/src/commands/channels.rs
@@ -349,6 +349,7 @@ pub async fn cmd_update_channel(
     channel_id: &str,
     name: Option<&str>,
     description: Option<&str>,
+    visibility: Option<&str>,
     ttl: Option<i64>,
     no_ttl: bool,
 ) -> Result<(), CliError> {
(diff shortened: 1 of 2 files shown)


rodroudi:
Closing — opened in error. I intended to submit a separate branch and mistakenly targeted this existing one, which is not mine to submit. Apologies for the noise.

PR #4625, opened 2026-08-03
feat(agents): harness-agnostic effort write path and spawn bridge

wpfleger96:
## Summary

Genericizes the agent effort write-side so Goose participates in the same canonical effort contract as buzz-agent. A spawn bridge translates the canonical key to whatever the target harness expects at launch time. Read/write/spawn paths all derive their vocabulary from runtime metadata rather than a hardcoded buzz-agent list.

### Rust — config bridge + spawn path

- `apply_spawn_effort_env` in `effort.rs`: production command-boundary seam — writes baked env, runs the effort projection, strips per-runtime suppress set, and emits exactly one projected key.
- `apply_effort_to_spawn_command` in `runtime.rs`: thin wrapper returning a `#[must_use] EffortApplied(())` token (private field — unforgeable outside the function). `spawn_agent_child` calls it as `let effort = apply_effort_to_spawn_command(...)` and passes `effort` to `spawn_with_effort_proof`. Deleting the call is a compile error: `effort` is undefined at the `spawn_with_effort_proof` site. Deleting `apply_spawn_effort_env` inside the wrapper turns the production-sequence tests RED.
- `apply_record_field_updates` in `agent_models_update.rs`: returns `Result<RecordFieldsApplied, String>` (`#[must_use]` token). `update_managed_agent` calls it as `let applied = apply_record_field_updates(...)?` then passes `applied` to `stamp_record_updated_at`. Deleting the call is a compile error: `applied` is undefined at the `stamp_record_updated_at` site.
- Unknown/custom-runtime passthrough: `apply_effort_launch_to_command` skips the suppress loop when `preserve_passthrough && value.is_none()`, preserving ambient ACP sentinels.
- `EnvVarGuard`: prior value stored as `OsString` (`var_os`) so non-Unicode values are restored exactly on Drop. A single `PROCESS_ENV_MUTEX` in `managed_agents/mod.rs` is shared by `lock_path_mutex()` and `lock_env_mutex()` — any two tests calling either helper are mutually exclusive with each other. Tests in other modules (`app_state_tests`, `agent_config_tests`, `reader_tests`) maintain their own independent locks and are not in this domain.
- Dead-code: `strip_effort_keys_from_command` marked `#[cfg(test)]`; import path in `effort_cmd_tests.rs` fixed.
- Windows CI fix: platform-gated variants for inherited-env tests.

### TypeScript — renderer + model cleanup

- `AgentConfigFields` orphan-model cleanup effect: the `isHarnessNativeEffort` early-return was skipping the model clear on provider→Custom transitions. Refined to: return early only when model is already null; clear model once while preserving the harness-native effort key (Carl P2).
- Provider-empty convergence: when model is null and effort is native, the cleanup effect returns early (nothing to clear) — prevents spurious `onConfigChange` loop.
- `EffortSelectField` / `humanizeEffortLabel`: runtime-native option labels title-cased (`off` → `Off`) with raw canonical values preserved for round-trip fidelity.
- `AgentConfigFields`: drives effort renderer from `selectedRuntime.effortCanonicalValues` (harness-native path) or the model/provider catalog (buzz-agent/provider path), selected by `isHarnessNativeEffort`.

### Docs

- `desktop/src/features/agents/AGENTS.md` item 14: updated from deleted `persistAgentEffortLevel` direct-write contract to the shipped Save-gated `update_managed_agent.effortLevel` path. Consistent with `EffortPickerField`'s own doc comment.

### Tests

- `agent_models_update_tests.rs`: seam tests via `apply_record_field_updates` — non-local rejects, local set/clear, ordering invariant, ACP-sentinel sweep. `record_field_updates_persist_effort_to_disk` (renamed from the prior false-claim name) drives load→apply→stamp→save→load via a mock AppHandle + tempdir, asserting `effort_level` persists to disk. Manual HOME/XDG restore replaced with RAII `EnvVarGuard` (panic-safe, `OsString`-exact).
- `effort_cmd_tests.rs` / `effort_tests.rs`: production-sequence seam tests via `apply_effort_to_spawn_command`. Spawns `/usr/bin/env` to verify child's real env. `EnvVarGuard` for panic-safe restore. Windows twin using `cmd /c set`.
- `effortAutoClear.test.mjs`: five mounted stateful journeys via `AgentConfigFields` with `useCustomSelect=true`. Covers: custom trigger shows "Off" at mount; provider-empty mount is a stable fixed point; provider→Custom switch converges; stale Anthropic model cleared on Custom switch with Goose effort preserved (Carl P2 regression); Settings-style Save/reread preserves effort.
- `agentDefaultsEditor.test.mjs`: two full Save/Next journey tests through the real production parent trees. Both start with `GOOSE_THINKING_EFFORT: "low"` and operate the real Popover-based effort control (click trigger → click "off" option) before Save/Next, asserting zero writes after selection. The `set_global_agent_config` stub captures the submitted payload; each test asserts raw `GOOSE_THINKING_EFFORT: "off"` in the captured config. The stub stores its canonical response from the actual payload; the fresh remount's `get_global_agent_config` returns that stored object (not a hand-written fixture), then asserts "Off" shown. The `DefaultConfigStep` test starts with `isDirty: false` — the real-control effort selection calls `onConfigChange → updateDraft → isDirtyRef=true`, making the `commit()` on Next load-bearing.

## Mutation evidence

- Delete `let effort = apply_effort_to_spawn_command(...)` call from `spawn_agent_child` → compile error: `error[E0425]: cannot find value `effort`` at `spawn_with_effort_proof` site.
- Delete `let applied = apply_record_field_updates(...)?` from `update_managed_agent` → compile error: `error[E0425]: cannot find value `applied`` at `stamp_record_updated_at` site.
- Delete `apply_spawn_effort_env` from inside `apply_effort_to_spawn_command` wrapper → `production_sequence_goose_inherited_collision_resolved_in_child` RED.
- Revert `isHarnessNativeEffort &&` guard in cleanup `useEffect` to bare `if (isHarnessNativeEffort) return` → stale model not cleared → Carl P2 regression test RED.
- Remove `isHarnessNativeEffort ||` from the nothing-to-clear condition → provider-empty mount emits `onConfigChange` → loop test RED.
- Remove `isHarnessNativeEffort` branch in `AgentConfigFields.tsx:634-636` → both `agentDefaultsEditor.test.mjs` mount assertions fail: trigger shows "Select" instead of initial effort label.
- Remove `preserve_passthrough` guard in `apply_effort_launch_to_command` → `production_sequence_custom_inherited_acp_sentinel_survives` RED.
- Drop `GOOSE_THINKING_EFFORT` from the `set_global_agent_config` stub payload → payload assertion in `agentDefaultsEditor.test.mjs` fails (`undefined !== "off"`) → RED (verified).
- Remove the effort-select dirtying steps from the `DefaultConfigStep` test (so `isDirty` stays false) → `commit()` is a no-op → write-count assertion after Next fails (0 instead of 1) → RED.


kalvinnchau (changes requested):
[P2] Pin-to-inherit can retain stale effort when the current runtime has no effort key — `desktop/src/features/agents/ui/AgentInstanceEditDialog.tsx:500`

`handleInheritHarnessChange` decides whether to clear effort from the pre-toggle `prospectiveRuntime`. For an agent pinned to Claude (no `thinkingEnvVar`) whose linked persona uses Goose, selecting Inherit skips the clear. On the next render the prospective runtime becomes Goose, but the record's existing `BUZZ_AGENT_THINKING_EFFORT` remains; `resolveInheritedRuntimeSubmission` only excludes effort keys from the persona layer, so that record value is saved and the spawn bridge aliases it into Goose effort. This violates the stated pin↔inherit clear contract and can silently resurrect an effort value from an unrelated prior runtime. Resolve the runtime for `nextInherit` inside the handler (or clear whenever either the before or after runtime owns effort), and pin the Claude-pin → Goose-persona transition with an integration-level state/submit test.

[P2] Consumed legacy effort is also emitted as an advanced field — `desktop/src-tauri/src/managed_agents/config_bridge/reader.rs:125`

For a Goose record containing only a valid legacy `BUZZ_AGENT_THINKING_EFFORT`, `build_thinking_field` consumes it and exposes canonical `thinking_effort`, but `normalized_env_keys` contains only `GOOSE_THINKING_EFFORT`. The loop at line 140 therefore also emits the legacy key in `surface.advanced`. The config panel presents one persisted setting twice, with conflicting semantics, and an edit can appear to target a generic env var while the normalized field is already interpreting it. Invalid or unconsumed legacy values should remain advanced, so hide the legacy key only when record-tier alias resolution actually consumed it; add a surface test for valid-consumed versus invalid-visible legacy values.

Validation at exact PR tip `5882e4494656b0ce3deeffb2c1d1c1dea25e47f3`:
- `just desktop-test`: 4,774 passed, 0 failed
- `just desktop-tauri-test`: passed (workspace; one pre-existing ignored native performance test)
- `git diff --check a8e5c89e2..HEAD`: clean

wesbillman (changes requested):

**Changes requested**

- **[P2] Preserve custom-runtime effort env in the restart snapshot** — `desktop/src-tauri/src/managed_agents/spawn_snapshot.rs:181`

  The launch projection deliberately uses an empty suppress set for an unknown/custom runtime so wrapper-specific variables such as `GOOSE_THINKING_EFFORT` remain ordinary pass-through env. `SpawnConfigSnapshot::from_inputs`, however, unconditionally removes every globally known effort key. For an unknown command, `effective_effort` only reads `BUZZ_ACP_EFFORT_LEVEL`, so an exact-case `GOOSE_THINKING_EFFORT` disappears from both `snapshot.env` and `snapshot.effort_level` even though the child receives it. Changing that variable therefore produces no restart diff and can leave a running custom-wrapper agent on the old effort despite auto-restart-on-config-change. Scope snapshot cleanup to the runtime projection: known runtimes can collapse their suppress set, while unknown runtimes must retain non-sentinel effort-looking keys as ordinary env. Please pin the custom-wrapper case with a snapshot regression test.

The authority, normalization, mutation, local/remote launch, and reader paths were otherwise coherent in the read-only diff review. Existing checks on this immutable head are green; PR code was not executed locally under the automation trust policy.

kalvinnchau (dismissed):

wesbillman (changes requested):

**Changes requested at exact head `239cbdd30cefee7a3a28d02271fc21de5226967f`.**

### P2 — Mixed-case ACP sentinel changes for a custom runtime still disappear from restart snapshots

The new custom-runtime snapshot scoping fixes non-sentinel keys such as `GOOSE_THINKING_EFFORT`, but it still loses a mixed-case ACP sentinel on Windows.

For an unknown/custom runtime, launch projection deliberately uses an empty suppression set (`desktop/src-tauri/src/managed_agents/config_bridge/effort.rs:246-259`), so a user-provided `buzz_acp_effort_level=high` survives into the child environment when no canonical column overwrites it. Windows treats that key as `BUZZ_ACP_EFFORT_LEVEL`. Snapshot extraction, however, reads the destination with an exact-case map lookup (`desktop/src-tauri/src/managed_agents/spawn_snapshot.rs:153-157`), yielding no `effort_level`, then removes every case variant of the sentinel (`spawn_snapshot.rs:194-203`; `config_bridge/effort.rs:208-226`). Changing that effective child input therefore changes neither snapshot field, so restart-required remains false and a running custom agent can keep stale effort.

Make unknown-runtime snapshot handling mirror actual projection semantics: either capture the sentinel case-insensitively before stripping it, or retain a pass-through sentinel when projection did not emit/consume it. Add mixed-case custom-runtime regressions with and without a canonical column. The new test at `spawn_snapshot/tests_ext.rs:239-275` covers only exact-case `GOOSE_THINKING_EFFORT`, so it does not exercise this collision.

The prior non-sentinel custom-runtime blocker is otherwise closed. I reviewed immutable GitHub source and diffs only; no PR code was checked out or executed. Hosted checks were green at review time.

wesbillman (changes requested):
Carl, an automated reviewer, commenting via Wes's GitHub account.

The write/spawn projection is internally consistent, but the global/onboarding Goose Effort control still persists the legacy Buzz key that the new projection intentionally does not consume from the global tier.

- `deriveAgentConfigFieldModel` sets `currentPersistence.key` and its displayed value to `BUZZ_AGENT_THINKING_EFFORT` for every runtime with `thinkingEnvVar`, while only `targetApplication` uses `runtime.thinkingEnvVar` (`desktop/src/features/agents/lib/agentConfigCore.ts:206-220`). `AgentConfigFields` reads and writes through that persistence key (`desktop/src/features/agents/ui/AgentConfigFields.tsx:244-257,638-640,876-885`). Selecting Goose effort therefore saves the legacy key rather than `GOOSE_THINKING_EFFORT`.
- The launch authority accepts the legacy alias only at record/persona scope; its global tier reads the runtime-native key only (`desktop/src-tauri/src/managed_agents/config_bridge/effort.rs:362-383`). The UI consequently shows the global Goose setting as saved, but the next Goose spawn silently ignores it.

Please make the global/onboarding persistence contract match the launch projection, and add a regression covering global Goose selection through the persisted env to a projected `GOOSE_THINKING_EFFORT` descriptor.

wesbillman (changes requested):

Changes requested at exact head `97876e054fdb8ac65b707df21e8e1c896f67c483`:

1. **P1: the direct effort write races dialog Save and can undo pin-to-inherit clearing.** `EffortPickerField` immediately starts an independent `persist_agent_effort_level` mutation, but the parent dialog neither observes nor awaits it. Its Save and Cancel controls only account for the main update/avatar operations. A user can select effort, switch the harness to Inherit, and save while the picker IPC is unresolved. `update_managed_agent` can acquire the shared store lock first and clear effort, then the delayed picker command can acquire it and restore the pin. Cancel or a failed Save can also leave the already-dispatched effort write committed independently of the dialog outcome. Coordinate this write with the dialog lifecycle, preferably by submitting effort in the same locked update, and add a deferred-IPC regression proving the final inherited state cannot retain a stale effort pin.

2. **P2: duplicate case variants silently change the effective effort on Windows.** `get_ci` prefers the exact canonical key, while the actual `Command` environment is populated in `BTreeMap` order and Windows case-folds names, so the last case variant wins. For `GOOSE_THINKING_EFFORT=low` plus `goose_thinking_effort=high`, the pre-projection child value is `high`, but resolution selects `low`, strips both entries, and emits canonical `low`. The custom-runtime passthrough path already deliberately uses the last match to mirror Windows. Make known-runtime lookup follow the same rule, or reject/deduplicate case-insensitive user-key collisions, with a regression covering this exact pair.

wesbillman (changes requested):

Review pinned to exact head `505bfa5126fa4a526a3b6f5d8fced0c7c40d6b31` and exact base `2f66ee24a0bf5efa958378069a6b155e99a1d068`. Product contract: effort is spawn-scoped, runtime capability/vocabulary is Rust-catalog-owned, the edit picker is Save-gated, and Save/Cancel must present one coherent persistence boundary. I traced UI selection through the main update, auto-restart and effort setters, pin/inherit clearing, global/persona/record/definition/baked authority, native/legacy/sentinel projection, Windows case folding, custom runtimes, local/deploy launch, config-reader presentation, restart snapshots, and exact-head checks. No PR code was checked out or executed.

## Blocking finding

1. **P1 — the Save transaction becomes interactive before the standalone effort write finishes, permitting post-Cancel writes, duplicate submits, and invisible partial failure.** `handleSubmit` awaits `updateMutation.mutateAsync`, then separately awaits `setManagedAgentAutoRestart`, `persistAgentEffortLevel`, and query invalidation. Every pending/loading/disabled gate, however, uses only `updateMutation.isPending`. As soon as the main update resolves, the dialog changes back from “Saving...” to “Save changes,” re-enables Cancel, Save, provider/model/effort fields, and the rest of the form while the effort setter may still be unresolved. The user can close or Cancel, after which the in-flight setter commits anyway, or submit again and race duplicate standalone writes. If `persistAgentEffortLevel` rejects, the broad catch leaves the dialog open but the only rendered error is `updateMutation.error`; because the main update succeeded, the effort failure is silent even though other edits are already committed.

   Keep a dialog-owned pending/error state across the complete Save sequence, prevent dismissal and duplicate submit until every required persistence step settles, and surface standalone-setter failure with a retryable error. Preferably collapse the effort value into the same locked native update so the user-visible Save is atomic. Add behavior tests that defer and reject `persist_agent_effort_level` after `update_managed_agent` succeeds, proving Cancel/close and duplicate Save remain unavailable and failure is visible. Current tests defer only the main update while the effort IPC always resolves immediately, so they stop one transition before the bug.

The previous direct-write race is otherwise closed: selection remains local until Save, rejected main updates do not dispatch effort, and pin-to-inherit suppresses re-persistence after the locked clear. Native projection, case-insensitive authority, custom-runtime pass-through, reader suppression, and snapshot/restart equivalence were coherent by source inspection. Substantive exact-head checks are green, including Desktop, Desktop Core, Desktop E2E/smoke, Rust lint, Windows Rust, macOS build, and the release candidate; the cancelled Codex security-review job is process metadata, not the basis for this verdict.

wesbillman (changes requested):

**Changes requested at exact head `f526ed303895b8605538d44aa554a5959746c3bd` and exact base `2c99ee7af5a20d239e95ec5368407887449795c9`.**

### P1: changing the runtime can persist a stale, unsupported effort value

`handleRuntimeDropdownChange` updates the runtime/command and resets provider/model state, but it neither clears nor revalidates the dialog’s touched `effortLevel` (`desktop/src/features/agents/ui/AgentInstanceEditDialog.tsx:517-568`). The picker options remain derived from the currently running agent’s config surface rather than the prospective runtime (`EffortPickerField.tsx:41-46`; call site at `AgentInstanceEditDialog.tsx:1108-1121`). Save then commits the runtime update before writing that stale effort string through the standalone setter (`AgentInstanceEditDialog.tsx:754-787`).

Reproduction: on a running Goose agent, select Goose-only effort `off`, switch Provider/runtime to `buzz-agent`, and save. The record stores `off`, but the destination runtime does not accept that vocabulary at spawn, so the explicit saved selection silently falls back to adapter/default behavior. The inverse transition can likewise present effort options that do not belong to the selected destination runtime.

Key picker visibility/options to the prospective runtime metadata and clear or canonical-revalidate any touched effort when runtime changes before Save. Add a regression covering effort selection → runtime switch → Save.

The previous incomplete Save gating is fixed at this head: update, standalone setters, dismissal controls, duplicate Save prevention, and setter error display now share the dialog-owned saving boundary. Hosted substantive checks are green. This was a read-only review of immutable GitHub source/diffs; no PR code was checked out or executed.

wesbillman (changes requested):

## Changes requested

Reviewed exact head `9ce3bd54ed2e242798582da40637f804e82993f6` against base `2c99ee7af5a20d239e95ec5368407887449795c9`. The intended contract is Save-gated effort for the next spawn, runtime-native normalization, and one effective projection shared by local/remote launch and restart comparison. Three actionable regressions:

### [P2] Scrub suppressed effort keys from the actual local Command

Anchor: [`runtime.rs:807–819`](https://github.com/block/buzz/blob/9ce3bd54ed2e242798582da40637f804e82993f6/desktop/src-tauri/src/managed_agents/runtime.rs#L807-L819).

The projection strips keys from `descriptor.env`, but the local process is created with an inherited environment (`runtime.rs:522`). In addition, `build_buzz_agent_provider_defaults` writes all raw baked env pairs directly to that Command before the descriptor overlay (`runtime.rs:745`, `agent_env.rs:113–116`). Removing `apply_effort_env` leaves any ambient/baked `BUZZ_ACP_EFFORT_LEVEL` alive even though the descriptor suppressed it.

Source-level reproduction: launch Desktop with `BUZZ_ACP_EFFORT_LEVEL=high` (or bake that pair), set a local Goose record’s canonical `effort_level` to `low`, and start a session on a model advertising a `thought_level` option. The descriptor/snapshot contain `GOOSE_THINKING_EFFORT=low`, but buzz-acp still reads the surviving sentinel (`crates/buzz-acp/src/config.rs:432–438`) and applies `high` through `session/set_config_option` at session creation (`pool.rs:1514–1544`). The user’s selected effort loses, and local launch differs from the descriptor-only remote payload. Before this change, the post-loop canonical write replaced that sentinel with `low`.

Remove the runtime-specific projection suppression set from the actual Command after the raw baked write and before overlaying the projected descriptor, preserving the custom-runtime passthrough contract. Cover both inherited and baked collisions at the Command boundary, including Windows casing.

### [P2] Migrate Goose validation before changing the global persistence key

Anchor: [`agentConfigCore.ts:216–233`](https://github.com/block/buzz/blob/9ce3bd54ed2e242798582da40637f804e82993f6/desktop/src/features/agents/lib/agentConfigCore.ts#L216-L233).

Global/onboarding effort now reads/writes `GOOSE_THINKING_EFFORT`, but its consumer still uses buzz-agent’s provider/model vocabulary: `AgentConfigFields.tsx:438–440,512–524` passes that key’s value to `useEffortAutoClear`; the hook deletes any value outside the provider set (`buzzAgentModelTuningFields.tsx:183–200`). `off` is a valid Goose canonical value (`discovery/runtime_metadata.rs:21–35`) but never appears in the buzz-agent vocabulary (`modelCapabilities.ts:32–40`). The renderer likewise still enumerates only that buzz-agent list.

Source-level reproduction: save global Goose defaults with a configured provider/model and `GOOSE_THINKING_EFFORT=off`, then open AI defaults. Mounting the fields removes the valid native value from the draft without a user effort change; saving an unrelated edit persists the deletion. At base the native key was not the effort descriptor’s persistence key, so this hook did not consume/delete it. The same renderer also cannot offer Goose’s canonical `off` choice.

Use the runtime’s normalization/choice contract throughout the global/onboarding renderer and clearing logic before retargeting persistence. Preserve valid saved native values on mount and add a component-level regression for opening/saving these defaults.

### [P2] Persist pending effort before the Save-triggered access restart

Anchor: [`AgentInstanceEditDialog.tsx:754–780`](https://github.com/block/buzz/blob/9ce3bd54ed2e242798582da40637f804e82993f6/desktop/src/features/agents/ui/AgentInstanceEditDialog.tsx#L754-L780).

In an OSS/custom build with editable access, use a running local agent whose adapter advertises effort, disable automatic config restarts, change both access and effort, then Save. `update_managed_agent` stops the running pairs for the access change (`commands/agent_models_update.rs:203–218`) and awaits their restart before returning (`351–363`). The dialog only calls `persistAgentEffortLevel` afterward. That Save-triggered restart therefore snapshots/launches the old effort; the subsequent setter only updates disk (`commands/agent_config.rs:556–588`), leaving the running process on the previous value until another manual restart. At base, the picker persisted its selection before Save, so this deterministic ordering regression is introduced by moving the write here.

Keep the new Save/Cancel contract, but include pending effort in the locked update/restart transaction, or defer the restart until all Save fields have persisted. Add coverage asserting the actual restart consumes the newly saved effort when access and effort change together; do not restore selection-time writes.

## Scope and validation

Source/diff review only; no checkout, build, tests, or execution of PR code. Traced local spawn, remote payload, restart snapshot, picker save/cancel, record/persona/global/definition/baked tiers, aliases, known/keyless/custom runtimes, and destination validation. The pre-existing baked-tier reader omission and existing live-ACP precedence are not counted as newly introduced blockers. Remote substrate execution and third-party adapter behavior were not runtime-tested.

wesbillman (changes requested):

## Changes requested

Reviewed head `c1c3eaeaafce9d7cc2252491f21839b12b6a9fd8` against base `4a9de1a3a121285ef475d630b2b5764044c02cde`. The contract remains Save-gated, runtime-native effort for the next spawn, with local-only instance writes and deployment-owned remote configuration.

### [P2] Let the provider-empty cleanup terminate when preserving native effort

Anchor: [`AgentConfigFields.tsx:480–503`](https://github.com/block/buzz/blob/c1c3eaeaafce9d7cc2252491f21839b12b6a9fd8/desktop/src/features/agents/ui/AgentConfigFields.tsx#L480-L503).

The effect's exit guard still requires both model and effort to be empty, but the new branch preserves harness-native effort and emits a fresh config object. With Goose, `GOOSE_THINKING_EFFORT=off`, and no configured or baked provider, onboarding's mount cleanup repeatedly calls `onConfigChange`. The parent stores each new object, the `config` dependency changes, and the effect runs again without ever clearing the value keeping it active. Defaults reaches the same loop by choosing **Custom provider** after selecting Goose effort (`handleProviderChange`, lines 521–530; parent state update in `AgentDefaultsEditor.tsx:182–187`). This makes that configuration workflow continually rerender rather than settle.

Keep native effort preserved, but return when there is nothing this effect will actually clear: an empty model plus preserved native effort is already stable. Add mounted stateful-parent regressions for provider-empty onboarding and provider→Custom defaults. The new tests use a nonempty provider and only capture callbacks, so they cannot expose this feedback loop.

### [P2] Preserve ambient effort configuration for unknown/custom wrappers

Anchor: [`effort.rs:257–293`](https://github.com/block/buzz/blob/c1c3eaeaafce9d7cc2252491f21839b12b6a9fd8/desktop/src-tauri/src/managed_agents/config_bridge/effort.rs#L257-L293).

`apply_effort_launch_to_command` ignores the projection's runtime-specific suppression policy and removes every globally known effort key. For an unrecognized wrapper such as `my-goose-wrapper`, launch Desktop with `GOOSE_THINKING_EFFORT=high` in its parent environment, leave the canonical effort column and explicit env tiers unset, then start the wrapper. The projection deliberately preserves foreign effort-looking keys for unknown runtimes (`effort_launch_projection`, lines 335–357), but the new Command helper removes this inherited key. `descriptor.env` contains only the configured/baked tiers, not ambient process env, so the later overlay at `runtime.rs:820–824` cannot restore it. The wrapper silently loses its existing effort configuration on the next launch. An ambient ACP sentinel with no explicit override is similarly removed despite custom passthrough semantics.

Use the projection's actual suppression/passthrough policy at the Command boundary; retain ambient foreign keys for unknown runtimes and preserve their sentinel when no authoritative value replaces it. Continue scrubbing collisions for known runtimes. Explicit record/persona/global/definition/baked values are restored by the descriptor overlay and are **not** the lost-input case here. Test a custom wrapper with an inherited known effort key, rather than only `MY_HARNESS_EFFORT`, which is outside the global removal set.

### [P2] Make the new child-env test portable to Windows

Anchor: [`effort_tests.rs:967–983`](https://github.com/block/buzz/blob/c1c3eaeaafce9d7cc2252491f21839b12b6a9fd8/desktop/src-tauri/src/managed_agents/config_bridge/effort_tests.rs#L967-L983).

`production_sequence_custom_passthrough_survives` is an unconditional test but launches `/usr/bin/env`. The exact-head [Windows Rust job](https://github.com/block/buzz/actions/runs/33459677800/job/99707157891) fails this test at line 856 with `Os { code: 3, kind: NotFound, message: "The system cannot find the path specified." }`; its Tauri suite reports **2948 passed, 1 failed, 15 ignored**. The neighboring Goose test is platform-gated and the Windows collision test already uses `cmd /c set`.

Use the appropriate child-env command per platform, preserving custom-passthrough coverage on Windows, or gate the Unix test and provide its Windows counterpart. This is a deterministic introduced test failure, not an inferred CI flake.

## Closure and validation

The prior access-restart ordering finding is fixed: effort now persists under the store lock before the restart. Goose's native vocabulary and valid `off` preservation are implemented; the cleanup loop above is a regression in that fix. Known-runtime ambient/baked collision cleanup is implemented; its custom-runtime preservation boundary still needs correction. Save/Cancel gating, runtime-switch reset, and normal UI pin→inherit suppression remain intact.

Non-blocking: a contradictory direct IPC request combining the inherit sentinel and a non-null effort can re-pin the column after clearing it. The shipped renderer suppresses this combination; no supported producer was established, so this is defensive boundary coverage, not another release blocker. Also, tests calling `apply_effort_sequence_to_command` prove that helper's behavior, not the claim that deleting its call from `runtime.rs` fails those tests; the final descriptor overlay should be represented in launch-boundary coverage.

Source-only immutable-file/diff review: renderer and persistence lanes returned and were integrated; the outstanding delegated launch lane was explicitly cancelled after Carl independently completed that trace. No PR checkout, build, or local test execution. Reviewed defaults/onboarding/create/persona/instance surfaces; absent/set/clear, Save/Cancel, runtime switch and pin/inherit; local update/restart, remote payload and snapshot; record/persona/global/definition/baked tiers and known/keyless/custom runtime vocabularies. Hosted check metadata and the exact-head Windows failure log were read. Remote substrate behavior and third-party adapters were not exercised. Previously deferred baked-tier reader omission and live-ACP precedence are not reopened.

Exit criteria are the three bounded fixes above plus regressions for their actual transitions; no architectural rewrite is requested.

wesbillman (dismissed):

## Changes requested

Reviewed head `7c0278288b24586d3e5a41857e8b7446c7770817` against base `bd73490418266f267d9bb3bdf13e64582adc8e80`. This is a bounded corrective review of the existing Save-gated, next-spawn native-effort contract, with local-only instance writes and deployment-owned remote configuration.

### P2: Preserve model cleanup while preserving native effort

The new unconditional return at [AgentConfigFields.tsx:480–496](https://github.com/block/buzz/blob/7c0278288b24586d3e5a41857e8b7446c7770817/desktop/src/features/agents/ui/AgentConfigFields.tsx#L480-L496) stops the feedback loop but also skips clearing the previous provider's model.

Source-derived reproduction: in Defaults, select Goose with Anthropic and a model, then choose **Custom provider…** and enter a different provider ID. The actual [Custom selection handler](https://github.com/block/buzz/blob/7c0278288b24586d3e5a41857e8b7446c7770817/desktop/src/features/agents/ui/AgentConfigFields.tsx#L522-L532) retains `config.model`; the new native-effort return prevents the one-time cleanup, and the custom input changes only `provider`. The previous provider-specific model survives into Save and is submitted with the new provider. During the blank Custom state, the model control even presents an empty/disabled value while validity still reads the retained model.

**Smallest repair:** return only when no eligible field needs clearing. Clear the model once, but preserve native effort; with a null model and native effort alone, emit nothing. Add a stateful-parent regression that invokes the real provider control, checks bounded config-change emissions, `model === null`, and preserved `GOOSE_THINKING_EFFORT=off`, then exercises Save/reread.

The current [Custom-switch test](https://github.com/block/buzz/blob/7c0278288b24586d3e5a41857e8b7446c7770817/desktop/src/features/agents/ui/effortAutoClear.test.mjs#L295-L363) replaces props directly, never setting `userEditedProviderRef`, and does not feed changes through React state. The separate Save test has a nonempty provider throughout. Neither establishes the claimed provider-to-Custom convergence proof.

### P2: Remove or test-gate the obsolete production strip helper

After runtime moved to `apply_spawn_effort_env`, [effort.rs:257–269](https://github.com/block/buzz/blob/7c0278288b24586d3e5a41857e8b7446c7770817/desktop/src-tauri/src/managed_agents/config_bridge/effort.rs#L257-L269) retains `strip_effort_keys_from_command` as non-test code, but its remaining callers are the old helper tests. The exact-head [Desktop Core job](https://github.com/block/buzz/actions/runs/33538927891/job/99960373619) fails deterministically in `desktop-tauri-clippy`:

```text
error: function `strip_effort_keys_from_command` is never used
--> src/managed_agents/config_bridge/effort.rs:257:15
= note: `-D dead-code` implied by `-D warnings`
error: could not compile `buzz-desktop` (lib) due to 1 previous error
```

Remove the obsolete helper and migrate its assertions to the live boundary, or make it test-only if it still has a useful test role. Do not restore its unconditional production call: that would reintroduce the custom-wrapper environment loss.

## Closed findings and validation limits

The prior render loop itself, inherited custom-wrapper environment loss, and Windows `/usr/bin/env` failure are addressed in source. The exact-head [Windows Rust job](https://github.com/block/buzz/actions/runs/33538927891/job/99960373390) is successful. The local set/clear/absent persistence seam and env-before-effort ordering remain intact; configured descriptor projection still feeds local launch, deployment, and restart snapshots.

All three independent review lanes returned and were integrated. Review used immutable source/diffs and hosted check metadata/logs only: no checkout, build, test, import, or execution of PR code. Desktop smoke shards 3/4 also report failures; their causes were not investigated or attributed here. Previously deferred live-ACP and baked-reader policies are not reopened. Exit criteria are the two repairs above with the real provider-transition regression and a passing Desktop Core gate.

kalvinnchau (approved):
🤖 Approved at `4caa6117322ac0f1caf20d7e8976a93b1cbf1bf1`. No findings in the final delta. The parent journeys now bind real effort selection to zero pre-save writes, the exact Save/Next payload, and a fresh reread of the stored response; the mutex and agent guidance now state only what the implementation guarantees. Independent targeted run: 7/7 passed. Scores: minimalness 9/10, elegance 9/10, correctness 9/10. Residual risk: no installed-Goose end-to-end run at this head; real child-process environment coverage remains the terminal automated boundary.

PR #4633, opened 2026-08-03
fix(mobile): merge relay recounts with locally seen thread replies

brow:
## Summary

- Keep mobile thread reply badges current by merging relay recounts with replies observed locally.
- Retain replies in the local channel store while continuing to filter them from the main timeline.
- Match the thread summary behavior already used on desktop, including the reply count, latest reply time, and participant avatars.

## Why

On mobile, the "N replies" badge under a channel message can stall at a stale count or remain missing after a reply arrives. This makes the badge unreliable and can cause people to miss replies.

The badge has two inputs: best-effort recounts from the relay and replies the client sees arrive. Mobile previously let any positive relay recount override the local view, while also discarding replies from its local message store. A delayed or lost recount, or a reply received after the recount, could therefore leave the badge behind.

This change combines both inputs by using the higher reply count, the later last-reply time, and a merged participant list. Relay timestamps have one-second precision, so equal timestamps do not prove that a recount included a locally observed reply. Comparing counts preserves that reply instead of trusting recency alone. Desktop already uses this merge behavior.

## Validation

At commit `4e3356636f5ad62e8f07910af305c532186c6c08` with a clean worktree:

- `flutter test` for mobile: 1105 passed, 1 skipped
- `flutter analyze` for mobile: no issues found
- Reverting the merge so a positive relay recount shadows local replies fails 4 of the new tests, including the same-second and reply-after-recount cases. Restoring the store-level reply drop fails both new provider tests.

Added tests:

- [`timeline_message_test.dart`](https://github.com/block/buzz/tree/main/mobile/test/features/channels/timeline_message_test.dart), covering relay-only recounts, a reply newer than the recount, a reply in the same second as the recount, a lost recount, a zero recount, nested replies at the root and at the reply they answer, a deleted reply, and participant merging and capping.
- [`channel_messages_provider_test.dart`](https://github.com/block/buzz/tree/main/mobile/test/features/channels/channel_messages_provider_test.dart), covering a live reply reaching the store while staying out of the main timeline, and a reply newer than the relay recount raising the badge.

### DIFF
--- a/mobile/lib/features/channels/channel_messages_provider.dart
+++ b/mobile/lib/features/channels/channel_messages_provider.dart
@@ -254,7 +254,11 @@ class ChannelMessagesNotifier extends Notifier<AsyncValue<List<NostrEvent>>> {
           ),
         );
       }
-      if (!_isBroadcastReply(event)) return false;
+      // Replies are kept in the store rather than dropped here, matching
+      // desktop: the main timeline filters them out at render
+      // (`buildMainTimelineEntries`), and their parent's "N replies" row needs
+      // them as the local half of the summary merge when the relay's
+      // best-effort recount is delayed, lost, or older than this reply.
     }
     // Thread summaries are neither a timeline row nor an aux event, but they are
     // how the root's "N replies" row learns a reply landed — a reply itself
@@ -493,12 +497,6 @@ class ChannelMessagesNotifier extends Notifier<AsyncValue<List<NostrEvent>>> {
     });
     return true;
   }
-}
-
-bool _isBroadcastReply(NostrEvent event) {
-  return event.tags.any(
-    (tag) => tag.length >= 2 && tag[0] == 'broadcast' && tag[1] == '1',
-  );
 }
 
 int _currentUnixSeconds() => DateTime.now().millisecondsSinceEpoch ~/ 1000;
--- a/mobile/lib/features/channels/timeline_message.dart
+++ b/mobile/lib/features/channels/timeline_message.dart
@@ -537,20 +537,48 @@ bool _isBroadcastReply(TimelineMessage message) {
   );
 }
 
+/// Combine what the relay counted with what this client has actually seen.
+///
+/// The count and last-reply time follow the desktop's `mergeThreadSummaries`
+/// (`desktop/src/features/messages/lib/threadPanel.ts`): the relay recount is
+/// authoritative for replies this client never loaded, and the locally observed
+/// replies are authoritative for anything that landed after (or alongside) the
+/// last recount. Neither source alone is complete, so take the larger count and
+/// the later reply time rather than letting one shadow the other. The facepile
+/// order is mobile's own (see [_mergeParticipants]) because this file already
+/// renders relay participants in the order the relay sent them.
 ThreadSummary? _buildSummary(
   String messageId,
   Map<String, List<TimelineMessage>> childrenByParent,
   ChannelWindowThreadSummary? relaySummary,
 ) {
-  if (relaySummary != null && relaySummary.replyCount > 0) {
-    return ThreadSummary(
-      threadHeadId: messageId,
-      replyCount: relaySummary.replyCount,
-      participantPubkeys: relaySummary.participantPubkeys.take(3).toList(),
-      lastReplyAt: relaySummary.lastReplyAt,
-    );
-  }
+  final local = _buildLocalSummary(messageId, childrenByParent);
+  final relay = _buildRelaySummary(messageId, relaySummary);
+  if (relay == null) return local;
+  if (local == null) return relay;
 
+  return ThreadSummary(
+    threadHeadId: messageId,
+    replyCount: local.replyCount > relay.replyCount
+        ? local.replyCount
+        : relay.replyCount,
+    // Relay participants first: they describe the whole thread, including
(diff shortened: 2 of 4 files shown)


wesbillman (dismissed):
Reviewing on Wes's behalf.

The direct-reply model here does not actually match desktop and leaves the stated stale-badge failure in place for nested replies.

`_buildLocalSummary` only reads `childrenByParent[messageId]`, so a nested reply observed locally does not update the root's count, last-reply time, or participant list. `_buildRelaySummary` compounds that by using `replyCount`, while desktop deliberately uses a descendant index locally and `summary.descendantCount` from the relay (`desktop/src/features/messages/lib/threadPanel.ts:212-225,386-394`). The new test at `timeline_message_test.dart:995-1003` codifies the incorrect result: a root with `r1` and nested `r2` remains at 1. If `r2`'s recount is delayed/lost—the exact failure mode this PR targets—the root badge remains stale despite the client having seen `r2`.

Please build local summaries from descendant stats (including nested replies) and use `relaySummary.descendantCount`, matching desktop, then update tests to require the root badge/count/time/participants to include locally observed nested descendants.

brow:
🤖 Fixed, and you were right on both halves. Head is now `c11964d497f8d9d198956c80d55c0d9a91044ce7`.

Local summaries are now built from descendant stats. New `_buildDescendantStats` in `timeline_message.dart` attributes every loaded message to each ancestor on its parent chain, with a hop cap so a malformed chain cannot loop, and accumulates count, latest reply time, and up to 3 recent participants newest-first. `_buildLocalSummary` reads that instead of `childrenByParent[messageId]` and reverses the participants to oldest-first. `_buildRelaySummary` now reads `relaySummary.descendantCount` instead of `replyCount`. That mirrors `buildDescendantStatsByMessageId` and `buildRelayThreadSummary`, so mobile and desktop now show the same number for the same thread.

The test you flagged is inverted, not deleted. `a nested reply does not inflate its root badge` is now `a locally seen nested reply raises its root badge`, and it asserts the exact case you described: the relay recount for the root predates the nested reply, and the root has to end up at count 2, the nested reply's time, and both participants. I also updated `summary counts only direct children, not nested replies` to require the descendant count, fixed the root assertion in `a nested reply badges the reply it answers`, and added `the relay half counts descendants, not direct replies`, where the relay reports `reply_count` 1 with `descendant_count` 5 and the badge must read 5.

Verified: `dart format` clean, `flutter analyze` no issues, `flutter test` 1106 pass. I also checked the coverage is live rather than inert: with the ancestor walk cut back to direct parents only, 3 of those tests fail, and they pass again when it is restored.

wesbillman:
Re-reviewed exact head `c11964d497f8d9d198956c80d55c0d9a91044ce7` on Wes's behalf. The blocker is resolved.

The updated mobile implementation now mirrors desktop on both halves of the merge: local summaries attribute each loaded reply to every known ancestor and relay summaries consume `descendantCount`. The focused regressions now require a locally observed nested reply to update the root's count, latest-reply timestamp, and participants, and separately pin `reply_count: 1` / `descendant_count: 5` to a badge count of 5.

I traced the new ancestor walk against desktop's `buildDescendantStatsByMessageId`, including stable timestamp ordering, participant cap/dedup, missing-parent termination, and bounded malformed chains. I found no new actionable issues in the update. Existing CI is fully green at this head, including Mobile; I did not duplicate its broad suite locally.

This clears my prior requested changes but is not an approval.

wesbillman (commented):
Reviewing on Wes's behalf at exact head `c11964d497f8d9d198956c80d55c0d9a91044ce7`.

The prior nested-reply blocker is resolved. Local summaries now attribute every loaded reply to each known ancestor and relay summaries use `descendantCount`, matching the desktop model. The merge preserves the maximum observed count, latest reply timestamp, and a bounded/deduplicated participant set; missing parents terminate safely and malformed chains are hop-bounded. Focused coverage now requires nested descendants to update the root and distinguishes relay `reply_count` from `descendant_count`.

I found no remaining actionable issue. Required CI, including Mobile, is green at this exact head. This is safe to merge once the required approval is supplied; this review does not itself approve because Wes asked for review, not approval.

wesbillman (commented):
Re-reviewed exact head `c11964d497f8d9d198956c80d55c0d9a91044ce7` on Wes's behalf after an independent adversarial pass.

The head is unchanged from my prior clean review. Mongo independently challenged the ancestor walk, descendant-count relay merge, malformed/cyclic parent handling, main-timeline filtering, and whether the focused tests distinguish the old behavior. The implementation held: nested local replies update every known ancestor; relay summaries use `descendantCount`; merge count/time semantics match desktop; malformed chains are bounded; and the regressions fail against the old direct-child / `replyCount` behavior.

I also refreshed GitHub state: all required checks are complete and green at this exact head, including Mobile and the newly reported DCO check. `git diff --check origin/main...c11964d497f8d9d198956c80d55c0d9a91044ce7` is clean. No actionable findings. Test behavior changed intentionally: nested descendants now must update the root count, timestamp, and participants; relay `descendant_count` is distinguished from `reply_count`; and live replies must remain in the store while staying filtered from the main timeline.

This is safe to merge once the required approval is supplied. This review does not approve because approval was not explicitly requested.

wesbillman (dismissed):

wesbillman (approved):

PR #4635, opened 2026-08-03
fix(desktop): mention live relay agents instead of re-instantiating personas

### LINKED ISSUE #4634: Persona mentions from a second desktop fail to send and mint orphan agent keypairs
## Describe the bug

Managed-agent persona records sync across desktops signed into the same identity, but harness definitions and the agent runtime are per-machine. On a desktop that is not the agent's host machine, @-mentioning an agent resolves to the persona, and the send flow tries to instantiate a local agent from it (`createMentionedPersonaAgents` with `ensureRunning: true`). This fails with:

```
Could not create agent mentions: <name>: cannot spawn agent <pubkey>: harness "<id>" was deleted — pick a new harness for this agent or restore the harness definition
```

and the message send is blocked entirely. Each failed attempt also generates a fresh agent keypair, so repeated tries mint orphan identities.

The root cause is that mention resolution only consults the local managed-agent registry when deciding whether a persona is already instantiated (`useMentions.ts`: `personaCandidates` filters by local `managedAgentPersonaIds`; the send flow never checks relay agents). An agent that is alive on another desktop looks like an uninstantiated persona.

## To reproduce

1. Desktop A: create a managed agent from a persona with a custom harness. Agent runs, responds normally.
2. Desktop B (same identity, e.g. a second computer): join the same self-hosted relay.
3. Desktop B: in a channel, @-mention the agent by name and send.
4. Send fails with the harness error above; relay users table gains a new orphan pubkey per attempt.

DMs to the running agent from Desktop B work fine — only channel mentions break, because only they route through persona instantiation.

## Expected behavior

Mentioning an agent that is already live on the relay should tag that agent, regardless of which machine hosts it. Persona instantiation should only trigger when no live instance exists.

## Environment

- Desktop A: macOS, dev build from main
- Desktop B: Windows, v0.5.4 release
- Self-hosted relay

dilate7:
## Summary

Cross-device agent mentions were broken end-to-end: mentioning an agent from a desktop that is not its host machine either silently hid the agent from the mention picker or failed the send with `harness ... was deleted` while minting an orphan keypair per attempt. Three coupled causes, each fixed here:

1. **Mention candidate gate consulted only the local registry.** `isAgentIdentityInManagedList` dropped every agent-flagged candidate not present in this machine's managed-agent registry, vetoing remote agents before `shouldHideAgentFromMentions`/eligibility ever ran. Now agents that are invocable via the relay directory (kind:10100 profile with `respond_to` and a shared channel) pass the gate.

2. **Persona mentions always instantiated locally.** `createMentionedPersonaAgents` spawned a local agent from the persona even when a live agent with that name already existed. Now it resolves to the live relay agent (preferring one already in the channel); instantiation still happens when no live agent exists.

3. **Mention-readiness started agents locally even when they were already running elsewhere.** `ensureManagedAgentMentionsReady` now skips the local start for agents whose relay presence is online and just tags them.

Supporting change: `buzz agents set-profile --name <n> [--respond-to ...] [--channel <uuid>]...` — a CLI command for an agent to publish the kind:10100 profile that all of the directory-based checks depend on. (Nothing in the desktop or harness publishes agent profiles today, so on self-hosted deployments the directory is empty and every directory-driven feature is inert — see #4634.)

Not included (follow-up candidates): the members-sidebar add-member search still applies the local-registry-only gate; the desktop could publish/refresh kind:10100 automatically on agent create/start.

### Related issue
Fixes #4634

### Testing
- `tsc --noEmit`, `biome check`, file-size ratchet, and the agent/mention test suites pass (56 tests, including new coverage for the directory-invocable gate behavior and the persona→live-agent resolution helpers).
- Verified end-to-end on a self-hosted relay with two desktop instances on one machine (separate profiles): host desktop runs the agents; second desktop (fresh profile, no local registry entries, no harness) sees the agents in the mention picker, sends a channel mention with the correct live pubkey in the `p` tag, no spawn is attempted, and the agent answers. Before these changes the same setup showed no agents in the picker and failed sends with the harness error.

### DIFF
--- a/crates/buzz-cli/src/commands/agents.rs
+++ b/crates/buzz-cli/src/commands/agents.rs
@@ -148,7 +148,75 @@ pub async fn dispatch(command: AgentsCmd, client: &BuzzClient) -> Result<(), Cli
         }
 
         AgentsCmd::Archived => cmd_archived(client).await,
+
+        AgentsCmd::SetProfile {
+            name,
+            respond_to,
+            channels,
+        } => cmd_set_profile(client, &name, respond_to.as_deref(), &channels).await,
     }
+}
+
+/// Publish this identity's kind:10100 agent profile — sign and submit a
+/// replaceable event whose content carries the directory display name, the
+/// respond-to policy, and the channels the agent is a member of.
+///
+/// Clients discover relay agents by querying kind:10100 (see the desktop's
+/// `list_relay_agents`); mention eligibility additionally requires the
+/// profile to declare `respond_to` and a channel shared with the viewer.
+/// An agent that never publishes a profile is invisible to agent
+/// directories on machines other than its host.
+async fn cmd_set_profile(
+    client: &BuzzClient,
+    name: &str,
+    respond_to: Option<&str>,
+    channels: &[String],
+) -> Result<(), CliError> {
+    let trimmed = name.trim();
+    if trimmed.is_empty() {
+        return Err(CliError::Usage("--name must not be empty".into()));
+    }
+    if let Some(policy) = respond_to {
+        match policy {
+            "anyone" | "owner-only" | "allowlist" => {}
+            _ => {
+                return Err(CliError::Usage(format!(
+                    "--respond-to must be 'anyone', 'owner-only', or 'allowlist' (got: {policy})"
+                )))
+            }
+        }
+    }
+    for channel in channels {
+        parse_uuid_arg(channel)?;
+    }
+
+    let mut content = serde_json::Map::new();
+    content.insert("name".into(), json!(trimmed));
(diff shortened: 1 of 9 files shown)


dilate7:
Withdrawing this for now — verification on a second desktop surfaced an additional spawn path (managed-agent start on mention) that this change doesn't fully cover yet. I'll follow up once the fix is verified end-to-end. The issue (#4634) still stands and reproduces as described.

dilate7:
Superseded by #4913 — same problem, upstream's own fix. Dropping our cherry-picks and pulling latest. Thanks for the fast turnaround, Wes.

PR #4656, opened 2026-08-04
feat(deploy): add device-pairing sidecar to the compose stack

vishoo7:
## Problem

Mobile QR pairing fails out of the box for `deploy/compose` deployments. A membership-gated (NIP-43) relay can't accept unpaired devices, so the desktop app falls back to pointing the phone at the legacy `<relay>/pair` path (`resolve_pairing_relay_url` in `desktop/src-tauri/src/commands/pairing.rs`). The compose stack never runs the `buzz-pair-relay` sidecar or routes `/pair`, so the phone's WebSocket upgrade lands on the main relay's router and pairing always dies with **"websocket connection failed: http error: 404 not found"**.

The Helm chart already handles this (`pairingRelay.enabled` runs the sidecar from the relay image and advertises it in NIP-11); this PR brings the compose bundle to parity.

## Changes

- **`compose.pair.yml`** (new overlay): runs `/usr/local/bin/buzz-pair-relay` from the existing relay image via an `entrypoint:` override (the image `ENTRYPOINT` is `buzz-relay`, so `command:` alone would silently start a second main relay), binds `0.0.0.0:5000`, and adds a bash `/dev/tcp` healthcheck matching the relay's approach. It also sets `BUZZ_PAIRING_RELAY_URL=wss://$BUZZ_DOMAIN/pair` on the relay so clients get the pairing URL from NIP-11 instead of relying on the legacy path fallback.
- **`Caddyfile`**: exclusive `handle` blocks — `/pair` → `pair-relay:5000`, everything else → `relay:3000`.
- **`run.sh`**: includes the overlay by default when `BUZZ_COMPOSE_TLS=true`; `BUZZ_COMPOSE_PAIRING=false` opts out. Not wired for the non-TLS stack (no reverse proxy to route `/pair`, and iOS requires `wss://`).
- **`README.md`**: new "Device pairing" section with verification steps.

## Testing

- `docker compose -f compose.yml -f compose.caddy.yml -f compose.pair.yml config` renders the merged stack correctly (sidecar entrypoint/env/healthcheck, relay `BUZZ_PAIRING_RELAY_URL`).
- Verified end-to-end on a live single-node VPS deployment of this stack: NIP-11 advertises `pairing_relay_url`, `GET /pair` returns the sidecar's 400 for non-WebSocket requests (previously the relay's 404), and iOS QR pairing against the desktop app completes through SAS confirmation.

https://claude.ai/code/session_01Voret84oXtBS5hfRo66uGS

### DIFF
--- a/deploy/compose/Caddyfile
+++ b/deploy/compose/Caddyfile
@@ -1,5 +1,13 @@
 {$BUZZ_DOMAIN} {
   encode zstd gzip
 
-  reverse_proxy relay:3000
+  # Device-pairing sidecar (compose.pair.yml). If the pairing overlay is
+  # disabled, requests to /pair fail with 502 instead of reaching the relay.
+  handle /pair {
+    reverse_proxy pair-relay:5000
+  }
+
+  handle {
+    reverse_proxy relay:3000
+  }
 }
--- a/deploy/compose/README.md
+++ b/deploy/compose/README.md
@@ -44,6 +44,31 @@ keypair.
   `.env`; use the Helm chart or a custom Compose configuration for providers
   such as new Railway Storage Buckets that require `virtual` addressing.
 
+## Device pairing
+
+Mobile QR pairing (NIP-AB) needs the `buzz-pair-relay` sidecar: a
+membership-gated (NIP-43) relay rejects unpaired devices, so the pairing
+handshake runs through a separate ephemeral relay instead. Without it, phones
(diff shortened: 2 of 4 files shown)


air3ijai:
- https://github.com/block/buzz/pull/2736
- https://github.com/block/buzz/pull/3627
- https://github.com/block/buzz/pull/3875

PR #4686, opened 2026-08-04
fix(desktop): use object-scale-down for inline markdown images

### LINKED ISSUE #4384: Markdown images without dim are upscaled to the fallback reserve box on first view (badges render huge)
## Summary

Small markdown images (badges, icons) are upscaled to fill the fallback reserve box on first view, rendering a ~20px shields-style badge as a huge blurry block. The same image renders at its correct natural size on the *second* view, because the decoded size cache then supplies real dimensions. The bug is the inconsistency between first view and every subsequent view.

## Environment

- Buzz Desktop 0.5.3 (macOS, Apple Silicon), also reproduced on 0.5.2
- Any channel, plain markdown message

## Repro

Post a message containing a small dim-less image, e.g.:

```markdown
![](https://img.shields.io/badge/OPEN-b91c1c.svg?style=flat-square)
```

**Expected:** the badge renders at its intrinsic size (roughly 46x20).

**Actual:** it renders as a giant block filling a ~384x256 area, visibly upscaled. After the decoded size has been cached (e.g. remount of the message list), the same message renders the badge at its natural size.

Related symptoms from the same mechanism:
- A wide image (e.g. 1200x240) is scaled down into the same box, making its text unreadably small.
- Small raster images are upscaled and pixelated.

## Root cause (from reading main)

`desktop/src/shared/ui/markdown/utils.ts`:

- A dim-less image with no cached decode gets `DEFAULT_IMAGE_RESERVE = { height: 256, width: 384 }` and `useFixedReserveBox: true` (`resolveImageReserveBox`).
- `useFrozenImageReserve` deliberately freezes that box for the lifetime of the mount (sound layout-shift reasoning).

`desktop/src/shared/ui/markdown/ProgressiveImage.tsx`:

- `IMAGE_CLASS` uses `object-contain` inside the frame. `object-contain` scales UP as well as down, so a 46x20 badge is stretched to fill the 384x256 frame for the entire first view.

`rememberDecodedImageDimensions` caches the natural size on load, which is why the second view is correct — the first view never benefits from the decode that happens inside it.

## Suggested fix

Either of these preserves the layout-shift guarantee for large images:

1. When a dim-less image decodes and its natural size fits entirely inside the reserve box, adopt the natural size immediately (collapse the frame). First view then matches the cached second-view behavior the code already intends. Shift is shrink-only and only for small images.
2. Alternatively/minimally: `object-scale-down` instead of `object-contain` on the inline image, so small images never upscale (they would sit centered in the reserve box until decode-cache kicks in on later views).

Happy to open a PR for option 1 if the direction is acceptable.

kushaim:
Fixes #4384.

A dim-less markdown image with no cached decode gets
`DEFAULT_IMAGE_RESERVE = { height: 256, width: 384 }` and
`useFixedReserveBox: true` in `resolveImageReserveBox`. The frame is
that big on first view, and the inline `` uses `object-contain`
which scales UP to fill the frame — so a 46×20 shields badge renders
as a giant blurry block. The same message renders at the natural
size on the second view because `rememberDecodedImageDimensions`
populates the cache, but the inconsistency between first and
subsequent view is the bug.

Switches `object-contain` → `object-scale-down` on
`ProgressiveImage.tsx`'s shared `IMAGE_CLASS`. Both `object-contain`
and `object-scale-down` scale down to fit, but only
`object-scale-down` also refuses to scale up — a small image now sits
centered in the reserve box on first view, never upscaled. The
layout-shift guarantee for large images is preserved (they still
scale down to fit the reserve), matching the issue's stated
constraint.

Chose option 2 ("Alternatively/minimally") from the issue over
option 1 because:

- It's a one-class-name change with zero new state, zero new
  effects, and zero risk to the existing layout-shift guarantee
  that the surrounding `useFrozenImageReserve` is explicitly
  designed to preserve.
- Option 1 (collapse the frame on decode) is the issue author's
  preferred direction but requires deciding whether the layout
  shift is acceptable *shrink*-only — a direction question that
  should be confirmed with the maintainer first. The issue
  explicitly says "Happy to open a PR for option 1 if the
  direction is acceptable" — that's the sign-off step, not
  this PR.

If maintainers want the full first-view-equals-second-view
behavior, option 1 is a follow-up that builds on this base.

Not added: a regression test. The current desktop test infra
(`desktop/src/shared/ui/markdown/markdown.test.mjs` and
siblings) doesn't cover `ProgressiveImage` directly; a
DOM-style snapshot would need a new test harness. The change
is one CSS class name; the visual regression is plain to
eyeball in the screenshots CI already captures.

Verified:
- `pnpm check` (biome + file-sizes + px-text + pubkey-truncation) — clean
- `npx tsc --noEmit` — clean

### DIFF
--- a/desktop/src/shared/ui/markdown/ProgressiveImage.tsx
+++ b/desktop/src/shared/ui/markdown/ProgressiveImage.tsx
@@ -2,8 +2,11 @@ import * as React from "react";
 
 import { cn } from "@/shared/lib/cn";
 
+// `object-scale-down` (not `object-contain`) so small dim-less badges don't
+// upscale to fill the reserve box on first view — the decoded size only
+// stabilizes after the cache hits. See #4384 for the full trade-off.
 const IMAGE_CLASS =
-  "absolute inset-0 block h-full w-full rounded-2xl object-contain";
+  "absolute inset-0 block h-full w-full rounded-2xl object-scale-down";
 
 function isSameImageSource(
   left: string | undefined,


wesbillman (commented):
Reviewed at `b6b8728e23a365d01e283d4b63608052ce3c5b84` on Wes's behalf.

No actionable findings. This is the intended minimal form of #4384's option 2: `object-scale-down` retains contain-style downscaling for content larger than its frame while preventing small dim-less images from being enlarged inside the frozen 384×256 first-view reserve. The image element, trigger hit area, reserve geometry, lightbox source box, thumbnail/full-image sequencing, and mosaic `object-cover` override remain unchanged.

Residual, non-blocking tradeoffs:
- A first-view small dim-less image remains centered in a large blank/clickable reserve; a later mount can use cached intrinsic dimensions. The PR accurately discloses that layout-preserving tradeoff.
- There is no direct first-view tiny-badge regression assertion.
- `desktop/src/shared/ui/markdown/utils.ts` still says the caller letterboxes via `object-contain`; that comment is now stale, but does not affect behavior.

I did not duplicate the repository's CI-equivalent suites locally. CI was still running when this review was submitted, with no failures in the completed checks I inspected. This comment is a technical review, not an approval; Wes retains approval authority.

wesbillman (approved):

PR #4693, opened 2026-08-04
fix(relay): guard multi-value #h filters from collapsing to one channel

### LINKED ISSUE #4659: HTTP bridge /query silently scopes multi-value #h filters to a single channel (breaks desktop Workflows list)
**Describe the bug**

The HTTP bridge `POST /query` returns wrong (empty) results for any filter with a multi-value `#h` tag. Instead of matching events in ANY of the listed channels (NIP-01 OR semantics), the query gets silently scoped to whichever channel UUID happens to iterate first out of the tag-value set — all other channels are ignored.

Root cause: `query_events` in `api/bridge.rs` builds its catch-all DB query via `handlers::req::build_event_query_from_filter`, which calls `extract_channel_id_from_filter` (`handlers/req.rs`). That helper returns the **first parseable UUID** from the `#h` values with no arity check:

```rust
// handlers/req.rs
        ...
        if key == "h" {
            for val in tag_values {
                    return Some(id);   // <-- multi-#h collapses to one channel
                }
            }
        }
    }
    ...
}
```

`bridge.rs` has its own `extract_channel_from_filter` with the correct `vs.len() == 1` guard, and the WS REQ path guards arity too (`per_filter_channel` in `handle_req`, and `filter_fully_pushable` explicitly documents multi-`#h` as not pushable) — but the bridge's query construction goes through the unguarded helper, so the SQL `WHERE channel_id = $x` predicate pins the query to one channel before the (correct) `filters_match` post-filter ever sees the rows.

Since `generic_tags` values are an ordered set, "first" is effectively the lexicographically smallest UUID, so which channel wins is data-dependent and surprising.

User-visible impact: the desktop **Workflows** screen fetches workflows for all member channels in one batched query (`get_channels_workflows`, single filter with `"#h": [all member channel ids]`). For any user in more than one channel whose lexicographically-first channel has no workflows, the screen shows "No workflows yet" even though workflows exist and run. Workflows created from the dialog appear to vanish (create succeeds, list never shows them, and they can't be edited/deleted from the UI).

**Steps to reproduce**

1. On a relay with channels A and B (UUIDs such that A sorts before B), create a workflow in channel B (via desktop dialog or `buzz workflows create`).
2. `POST /query` (NIP-98 auth) with body `[{"kinds":[30620],"#h":["<A>","<B>"]}]` → **`[]`**.
3. Same query with `[{"kinds":[30620],"#h":["<B>"]}]` → returns the workflow.
4. Same query with `[{"kinds":[30620]}]` (no `#h`) → also returns the workflow (access-scope fallback works; only the multi-value scoping is broken).
5. In the desktop app, be a member of ≥2 channels and open the Workflows screen → "No workflows yet".

Verified against a self-hosted relay running `ghcr.io/block/buzz:main` (2026-08-02); the unguarded helper is unchanged on current `main`.

**Expected behavior**

A multi-value `#h` filter matches events in any of the listed channels, like the WS REQ path: either push the full value set into SQL (e.g. `channel_id = ANY($ids)`, intersected with the caller's accessible channels), or leave `channel_id` unset when `len > 1` and let the existing access scope + `filters_match` post-filter handle it. Reusing the arity-guarded `extract_channel_from_filter` already in `bridge.rs` for query construction would fix the wrong-scoping half in one line.

**Version and platform**

- Buzz version: desktop 0.5.4; relay image `ghcr.io/block/buzz:main` pulled 2026-08-02
- OS: macOS (desktop), Linux/Railway (relay)

**Logs / additional context**

- Reproduced with raw NIP-98 `POST /query` requests (steps 2–4 above) — happy to share the exact request/response pairs.
- Related but separate quirks noticed while debugging (can file separately if useful): `workflows delete` removes the workflow from the scheduler but the kind:30620 event remains visible in list/get; and the create dialog surfaces no feedback when the list can't show the result, so users click Create repeatedly and accumulate duplicates.

santhiprakash:
## Summary

`extract_channel_id_from_filter` in `crates/buzz-relay/src/handlers/req.rs` previously returned the first parseable UUID out of a multi-value `#h` tag. This caused `POST /query` filters listing multiple channel UUIDs to silently collapse to whichever channel sorted first in `generic_tags`, turning a NIP-01 OR-semantics filter into an AND-of-one against a single arbitrary channel before the correct `filters_match` post-filter could run.

This change adds the same arity guard already present in `bridge.rs::extract_channel_from_filter` and in the WS REQ path (`extract_channel_id_from_filters`): `channel_id` is only set when the filter contains exactly one `#h` value and it is a parseable UUID. Multi-value `#h` filters now leave `channel_id` unset so the existing Rust-side post-filter handles the OR semantics correctly.

### Related issue

Fixes #4659

### Testing

- `cargo test -p buzz-relay handlers::req` — all 52 tests pass, including 5 new focused regression tests:
  - `extract_channel_id_from_filter_single_h`
  - `extract_channel_id_from_filter_multi_h_returns_none`
  - `extract_channel_id_from_filter_no_h_returns_none`
  - `extract_channel_id_from_filter_invalid_h_returns_none`
  - `build_event_query_from_filter_multi_h_leaves_channel_id_unset`
- `cargo fmt --all -- --check` passes
- `cargo clippy -p buzz-relay --all-targets --all-features -- -D warnings` passes

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -848,18 +848,20 @@ fn filters_are_nip43_membership_only(filters: &[Filter]) -> bool {
 }
 
 /// Extract a channel UUID from a single filter's `#h` tag.
+///
+/// Only returns `Some` when the filter contains exactly one `#h` value and it
+/// is a parseable UUID. Multi-value `#h` filters use NIP-01 OR semantics and
+/// must be evaluated by the Rust-side `filters_match` post-filter, not pinned
+/// to a single arbitrary channel in SQL.
 fn extract_channel_id_from_filter(filter: &Filter) -> Option<uuid::Uuid> {
-    for (tag_key, tag_values) in filter.generic_tags.iter() {
-        let key = tag_key.to_string();
-        if key == "h" {
-            for val in tag_values {
-                if let Ok(id) = val.parse::<uuid::Uuid>() {
-                    return Some(id);
-                }
-            }
+    let h_tag = nostr::SingleLetterTag::lowercase(nostr::Alphabet::H);
+    filter.generic_tags.get(&h_tag).and_then(|vs| {
+        if vs.len() == 1 {
+            vs.iter().next()?.parse::<uuid::Uuid>().ok()
+        } else {
+            None
         }
-    }
-    None
+    })
 }
 
 /// Convert a single NIP-01 filter into an [`EventQuery`] for the database.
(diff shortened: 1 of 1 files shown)


santhiprakash:

PR #4698, opened 2026-08-04
fix(relay): include event kind in HTTP bridge success attribution log

### LINKED ISSUE #4676: Relay logs omit event.kind, making publish failures indistinguishable from success
**Describe the bug**

The relay's HTTP bridge log line records route and status but not the event kind:

```json
{"message":"HTTP bridge request","pubkey":"d5b0ab11…","route":"/events","status":200,"accepted":true}
```

Typing indicators (kind 7) and their deletions (kind 5) travel the same `/events` route as real messages (kind 9). So **every agent turn produces `accepted:true` lines whether or not a message was actually sent.**

This is a trap rather than a cosmetic gap. Two independent debugging efforts — a previous session and my own — both concluded "the agent published successfully, so the transport works" from exactly these lines. It was false both times. The events were typing indicators; the agent had sent nothing.

The only way to establish the truth was to bypass the logs and query Postgres directly:

```sql
select created_at, kind, substring(encode(pubkey,'hex') for 8), substring(content for 70)
from events order by created_at desc limit 20;
```

**Steps to reproduce**

1. Have an agent receive a message and end its turn without calling `buzz messages send` (see the companion issue on silent prose replies — easy to trigger with a small model).
2. `docker logs buzz-relay-1 | Select-String '<agent pubkey>'`
3. Observe several `route:"/events", status:200, accepted:true` entries.
4. Query the `events` table — no kind 9 exists. The accepted events were kinds 7 and 5.

**Expected behavior**

The log line should include `kind` — and ideally `channel_id` — so that "the agent published a message" and "the agent published a typing indicator" are distinguishable without a database query.

**Version and platform**

- Buzz version: local dev build, `block/buzz` @ `318fbf896` (dockerised relay)
- OS: Windows 10 / Docker Desktop
- Stack: Postgres 16, Redis, MinIO

**Logs / additional context**

Real example. The agent sent no message at all during this window, yet the log reads as unambiguous success:

```
05:39:25.069Z  route:"/events"  status:200  accepted:true   ← kind 7 (👀 seen)
05:39:25.075Z  route:"/events"  status:200  accepted:true   ← kind 7 (💬 typing)
05:39:44.339Z  route:"/events"  status:200  accepted:true   ← kind 5 (delete)
05:39:44.343Z  route:"/events"  status:200  accepted:true   ← kind 5 (delete)
```

One additional field would have collapsed a two-day investigation into a few minutes. It would also make the existing fan-out diagnostics far more useful, since those currently can't tell you what was fanned out either.

santhiprakash:
## Summary
- Adds `kind: u32` to `SubmitOutcome::Ok` in `crates/buzz-relay/src/api/bridge.rs`.
- Populates it from the existing `kind_u32` local computed by `buzz_core::kind::event_kind_u32(&event)` in `submit_event_authed`.
- Logs `kind` in the `SubmitOutcome::Ok` `tracing::info!` arm, mirroring the existing `SubmitOutcome::Rejected` arm.
- Adds a focused regression test (`submit_event_text_note_includes_kind_in_attribution_line`) that drives a real `POST /events` and asserts the accepted attribution line contains `kind=1`.

Fixes #4676.

## Test plan
- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p buzz-relay --lib -- --ignored submit_event_` (5 tests, including new regression test, with local Postgres + Redis)
- [ ] `just ci` — blocked in this environment by a missing `glib-2.0` system package required for `desktop-tauri-clippy` (`pkg-config` cannot find glib-2.0). The Rust workspace checks and targeted relay tests pass.

### DIFF
--- a/crates/buzz-relay/src/api/bridge.rs
+++ b/crates/buzz-relay/src/api/bridge.rs
@@ -654,12 +654,13 @@ pub async fn submit_event(
         submit_event_authed(&state, &tenant, &headers, &body, pubkey, event_id_bytes).await;
 
     match &outcome {
-        SubmitOutcome::Ok { accepted, .. } => {
+        SubmitOutcome::Ok { accepted, kind, .. } => {
             tracing::info!(
                 pubkey = %pubkey_hex,
                 route = "/events",
                 status = 200u16,
                 accepted,
+                kind,
                 "HTTP bridge request"
             );
         }
@@ -713,6 +714,7 @@ enum SubmitOutcome {
     /// Ingest pipeline ran and returned a result (accepted or not).
     Ok {
         accepted: bool,
+        kind: u32,
         response: Json<Value>,
     },
     /// JSON parse failure before ingest — log category/line/column, not msg.
@@ -843,6 +845,7 @@ async fn submit_event_authed(
             }));
             SubmitOutcome::Ok {
                 accepted: result.accepted,
+                kind: kind_u32,
                 response,
             }
         }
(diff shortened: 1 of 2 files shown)


santhiprakash:

PR #4703, opened 2026-08-04
fix(workflows): authorize trusted owner-only agent wakeups

JustinPerea:
## Summary

- have the relay attach the authenticated workflow owner to relay-signed workflow messages
- authorize owner-only agent wakeups only after verifying the relay identity, event signature, workflow marker, and exact owner attribution
- fail closed for missing, duplicated, malformed, forged, or non-relay-signed attribution

## Why

Hosted workflows are emitted by the relay, so owner-only agents currently reject scheduled workflow mentions even when the workflow owner is authorized. The Bookmarks workflow fired, but Bookmark Bee's ACP harness dropped the event before queuing a turn.

This keeps the agent owner-only. It does not broaden the agent to `anyone`.

## Verification

- `just ci` passed on commit `046827c921d4e933b7e8d17e043d733bd3873ce0`
- desktop frontend: 4,129 tests passed
- desktop/Tauri: 2,169 passed, 14 ignored
- Flutter analysis and tests passed
- workspace formatting, clippy, builds, and Rust package tests passed

The follow-up commit only updates `implementation-notes.md`; source is unchanged from the tested commit.

## Deployment requirement

End-to-end acceptance requires both artifacts:

1. deploy the relay half so workflow events carry the authenticated owner attribution
2. release/install the ACP/Desktop half that verifies that attribution

Originating Buzz channel: `965eb307-5e3b-4f4e-a9c6-94323d98982d`
Thread root: `12e4e065724f58cd54a03b5354aa4dc981260e9c5b18a9cb3a0275b4ce11be7b`

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -257,6 +257,60 @@ async fn author_allowed(
     }
 }
 
+/// Return the workflow owner represented by a relay-signed message.
+///
+/// Attribution is trusted only after verifying the NIP-11 relay signer, event
+/// signature, message kind, and exact single-value workflow/actor tags. Any
+/// ambiguity fails closed and falls back to the literal event signer.
+fn trusted_workflow_actor(
+    event: &nostr::Event,
+    trusted_relay_pubkey: Option<&str>,
+) -> Option<String> {
+    let trusted_relay_pubkey = trusted_relay_pubkey?;
+    if event.pubkey.to_hex() != trusted_relay_pubkey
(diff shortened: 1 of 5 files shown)


JustinPerea:
@block/buzz-oss-team Could a maintainer review this owner-only workflow wake fix? It preserves owner-only access and fails closed unless relay identity, event signature, workflow marker, and exact workflow owner attribution all verify. Local full CI passed; the visible security and DCO checks are green. After merge, we need the relay half deployed and the matching Desktop/ACP build shipped so we can verify a real scheduled callback end to end.

PR #4715, opened 2026-08-04
Confirm Run402 ownership completion in Buzz

MajorTal:
## What changed

- parse the optional Run402 result acknowledgement extension while preserving the legacy manual flow
- add a no-proxy, no-redirect, HTTPS-only native result client with public-address DNS pinning and strict time/body bounds
- keep the signed approval memory-only and ignore stale/superseded result reads
- show explicit waiting, completed, action-required, expired, cancelled, unsafe, malformed, and unreachable states
- auto-dismiss only after Run402 confirms the exact approval as `completed`

## Why

The browser callback confirms proof delivery, not Run402 ownership completion. The previous copy-only proposal (#4683) could explain the stale screen but could not make it correct, so it was closed. This replacement observes the authoritative Run402 result and keeps every uncertain or unsuccessful state open with a specific recovery action.

## Compatibility and security

Older Run402 links retain the existing manual-close behavior. Unknown or unsafe acknowledgement extensions are ignored or rejected safely. Result URLs cannot use credentials, query tokens, fragments, redirects, proxies, private/reserved addresses, oversized DNS answer sets, or oversized responses.

## Validation

- desktop unit suite — 4,138 passed
- native suite — 2,178 passed, 14 ignored, 0 failed, plus all companion crate tests
- Nostr-bind E2E suite — 17 passed
- deep-link parser suite — 21 passed
- native result-client suite — 7 passed
- `pnpm --dir desktop check`
- `pnpm build:e2e`
- `just desktop-tauri-fmt-check desktop-build desktop-tauri-check desktop-tauri-test`

### DIFF
--- a/desktop/src-tauri/src/commands/mod.rs
+++ b/desktop/src-tauri/src/commands/mod.rs
@@ -39,6 +39,7 @@ pub(crate) mod mesh_llm;
 #[cfg(feature = "mesh-llm")]
 pub(crate) mod mesh_readiness;
 mod messages;
+mod nostr_bind_result;
 mod notifications;
 mod observer_archive;
 mod os_idle;
@@ -98,6 +99,7 @@ pub use media_raw::*;
 #[cfg(feature = "mesh-llm")]
 pub use mesh_llm::*;
 pub use messages::*;
+pub use nostr_bind_result::*;
 pub use notifications::*;
 pub use observer_archive::*;
 pub use os_idle::*;
--- a/desktop/src-tauri/src/commands/nostr_bind_result.rs
+++ b/desktop/src-tauri/src/commands/nostr_bind_result.rs
@@ -1,0 +1,439 @@
+use std::{
+    net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr},
+    time::Duration,
+};
+
+use futures_util::StreamExt;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use tokio::net::lookup_host;
+use url::Url;
+
+const MAX_RESULT_RESPONSE_BYTES: usize = 4_096;
+const MAX_RESOLVED_ADDRESSES: usize = 16;
+const DNS_TIMEOUT: Duration = Duration::from_secs(3);
+const CONNECT_TIMEOUT: Duration = Duration::from_secs(3);
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(5);
+
+const ACTION_REQUIRED_CODES: &[&str] = &[
+    "BUZZ_ADOPTION_CALLBACK_INVALID",
+    "BUZZ_ADOPTION_OFFER_NOT_ACTIVE",
+    "BUZZ_ADOPTION_TARGET_MISMATCH",
+    "BUZZ_IDEMPOTENCY_CONFLICT",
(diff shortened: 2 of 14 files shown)


MajorTal:
Closing at the requester’s direction. Run402 must integrate with Buzz’s existing completion mechanism; this PR is not authorized for Buzz product changes.

MajorTal:
Reopening and rebasing onto latest `main`.

## Why reopen

This was closed on 2026-08-04 with "Run402 must integrate with Buzz's existing completion mechanism." Subsequent investigation (tracked in #5166) established that **no such mechanism exists** for the `browser_fragment_v1` return mode:

- The dialog's finish screen has exactly two exits — the **Continue** click and a full reset on a new inbound `deep-link-nostr-bind` (which restarts consent, it doesn't confirm). There is no completion listener.
- The two things that look like existing mechanisms don't transplant: the older `clipboard` mode's `?buzz_bind=signed` reopen is Buzz→browser (outbound-only), and the mobile-pairing session (`commands/pairing.rs`) exists to transfer an `nsec` — its live channel is proportionate to moving a private key, which nostr-bind must never do.

So the browser callback confirms *proof delivery*, never *authoritative completion*, and a successful adoption lands on a screen labelled "Pairing didn't finish automatically?" This PR is that missing observation: a native, SSRF-hardened result client that polls the Run402-authoritative result and drives explicit terminal states, auto-dismissing only on `completed`.

Polling (not a `buzz://` completion route) is deliberate: it does not depend on custom-scheme registration, so it also works on Windows, where #2949 drops cold-start deep links.

## Rebase + validation

Rebased from `a7f02ea0` onto current `main` (`be48ce98`, ~150 commits of drift). The single commit applied cleanly; no source conflicts. Upstream churn in the shared `e2eBridge.ts`/`bridge.ts` helpers did not overlap this PR's additions.

Validated on the rebased branch (macOS, aarch64):

- `pnpm --dir desktop typecheck` — clean
- `cargo check` (desktop tauri crate) — clean
- `nostrBindResult.test.mjs` — 9/9 pass, including fails-closed-on-malformed, ignores-superseded-request, and keeps-expired/cancelled-open

Compatibility unchanged: links without both result fields keep the released manual-close flow; unknown/unsafe result extensions are ignored or rejected; result URLs cannot use credentials, query, fragment, redirect, proxy, or private/reserved addresses.

MajorTal:
Filed #6201 for the underlying cause. The PTT-plugin move in this PR's second commit is what unblocks the `fetch_nostr_bind_result` registration — it's behaviour-neutral and separable, so it can be split out or dropped if you'd rather fix the cap a different way.

MajorTal:
Closing this since the desktop frontend is moving to [block/buzz-app](https://github.com/block/buzz-app). This change is desktop-only (the Tauri `nostr-bind` deep link plus the result step UI), so it belongs in the new client rather than the outgoing one.

For reference when the nostr-bind flow is ported: this branch adds an optional `result_protocol=run402_adoption_result_v1` / `result_url` extension to the `buzz://nostr-bind` deep link, safely ignoring unknown or unsafe extensions, and a `fetch_nostr_bind_result` command so the dialog can show a confirmed ownership result instead of stopping at the handoff screen. The completion-signal gap it works around is written up in #5166.

Issue #2950, opened 2026-07-26
Managed agent invisible/unresponsive to other community members even after setting "Who can talk to this agent" to Everybody


Cyberworkshop:
**Describe the bug**
A managed agent owned by one community member is invisible in another member's @ mention picker in a shared channel, and does not respond to that other member's messages at all — even after explicitly setting the agent's "Who can talk to this agent" option to "Everybody" and restarting the agent.

**Steps to reproduce**
1. User A creates/deploys a managed agent in Buzz Desktop and adds it to a shared channel that User B (a different community member) is also a member of.
2. Confirm the agent responds normally to User A's own @ mentions in that channel.
3. User A changes the agent's "Who can talk to this agent" setting from "Only me" to "Everybody."
4. User A stops and restarts the agent.
5. User B opens the same channel and types @ — the agent does not appear in User B's mention autocomplete; only User B's own agents and human community members show up.
6. User B tries typing the agent's name manually as unresolved literal text, and separately tries posting a plain message with no mention at all — in both cases, the agent never responds.

**Expected behavior**
Once "Who can talk to this agent" is set to "Everybody," any community member in a shared channel should be able to see and interact with the agent — not just its creator.

**Version and platform**
- Buzz version: 0.4.26
- OS: macOS (both users)

**Logs / additional context**
The agent was confirmed working correctly for its creator throughout (holding conversations, making real GitHub commits/PRs) — ruling out the agent itself being offline or broken. Renaming the agent, changing "Who can talk to this agent" to "Everybody," and a full restart were all tried independently, with zero change to the reported symptom in any case.

Gypsymom4lyfe:
I am here and still learning please give me some time to try and learn the system .

timlantin:
Independent repro on 0.4.26 (macOS, both users, self-hosted relay). Same symptom, and I traced it far enough to be confident the "Who can talk to this agent" setting is a red herring — it cannot affect the picker at all in this version. Details below in case they help whoever picks this up, since I think the fix needs to be slightly wider than the open PRs.

### The picker never reads the setting

`getMentionableAgentPubkeys` (`desktop/src/features/agents/lib/agentAutocompleteEligibility.ts`) is the only consumer of `respondTo` / `respondToAllowlist`, via `relayAgentIsSharedWithUser`. But in `useMentions.ts` the managed-list gate runs first:

```ts
if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
[...]

PR #4716, opened 2026-08-04
fix(desktop): make cross-owner channel-member agents mentionable

### LINKED ISSUE #3776: Cross-owner agent mentions are impossible: candidates filtered by local ownership, and eligibility reads a kind (10100) nothing publishes

oleg-ai-dev:
## Problem

A bot agent that is a member of a channel shared between two different Buzz identities can never be `@`-mentioned (or manually addressed via a typed mention) by anyone except the human who owns/hosts it — even when the agent's `respond_to` is `anyone`. Reported and reproduced independently across desktop v0.4.20–v0.5.3, macOS/Windows, hosted/self-hosted relays: #3776 (canonical, most precise), #3277 (original report + a documented client-side stopgap), #2349 (long thread with the business case for team use), #2950.

**Heads up to reviewers:** I'm aware this issue already has a large number of open PRs attempting the same fix. I looked through the ones I could find before opening this — none of the ones I checked source `respond_to`/`respond_to_allowlist` from the correct (instance-tier) fields (see Cause 2 below), which as far as I can tell would leave the exact same bug at a different codepath if merged as-is. Flagging that here explicitly in case it's useful for triage, independent of whether this particular PR is the one that ends up merged.

## Root cause (two independent, stacked causes — both are required to fix this)

**Cause 1 — the mention picker drops any agent you don't personally manage.** `isAgentIdentityInManagedList` (in `agentAutocompleteEligibility.ts`) only admits a candidate if it isn't flagged as an agent, or if its pubkey is in the viewer's own `managedAgentPubkeys` — built solely from the local `list_managed_agents` Tauri command. An agent owned by the other identity never enters that set (by design — managed-agent records hold device-local secrets and are never minted from a relay event). So the other person's agent is dropped in `useMentions.ts` before `shouldHideAgentFromMentions` / `relayAgentIsSharedWithUser` ever run — including for a manually typed mention, since `extractMentionPubkeys` only resolves `p` tags from the already-filtered candidate list.

**Cause 2 — the eligibility check reads an event kind nothing publishes.** Even with Cause 1 fixed, `relayAgentIsSharedWithUser` reads `channelIds`/`respondTo`/`respondToAllowlist` from a `RelayAgent` sourced from `list_relay_agents` (`agent_discovery.rs`), which queries the relay for `kind:10100`. Nothing in the codebase publishes `kind:10100` as an agent profile (the only writer is `buzz channels set-add-policy`, which writes unrelated `channel_add_policy` content). Agent identity/config is actually published as `kind:0` + `kind:30177` (`managed_agents/agent_events.rs`), which the eligibility code never reads — so the `RelayAgent` it needs is always empty, and `relayAgentIsSharedWithUser` always returns `false`.

Note: the agent-side authorization logic is already correct and needed no change — `author_allowed` in `buzz-acp` already permits any author for `RespondTo::Anyone` outside DMs. The agent would answer if it could be addressed; it just could never produce a mention.

## Fix

- **Cause 1**: thread the already-computed `mentionableAgentPubkeys` set (which already accounts for relay-directory sharing) into the `isAgentIdentityInManagedList` gate, so a real channel member is admitted when it's shared, while relay-directory-only (non-member) agents stay hidden exactly as before. Same relaxation applied to `extractMentionPubkeys` so a p-tag is actually emitted for a manually typed mention of such an agent.
- **Cause 2**: repoint `list_relay_agents` (`agent_discovery.rs`) from `kind:10100` to `kind:30177`, and resolve channel membership via a `kind:39002 #p` query — the same pattern `buzz-acp`'s own channel-discovery code already uses. The agent's pubkey is read from `kind:30177`'s `d` tag (not the event's own pubkey, since `kind:30177` is a parameterized-replaceable event published by the owner). **`respond_to`/`respond_to_allowlist` are sourced from the record's instance-tier runtime fields, not the definition-tier fields** — the snapshot this event is built from carries both, and they can diverge in practice (the running harness is spawned from instance-tier fields). Sourcing from definition-tier would ship a policy the running agent doesn't actually enforce (agent visible but silent, or answering but hidden).

## Test plan

- Added/updated unit coverage in `agentAutocompleteEligibility.test.mjs` for the new gate logic, including the two cases that must *stay* hidden: a non-member agent that's only relay-directory-listed, and a channel member with an explicit not-invocable directory entry.
- Added an e2e test in `mentions.spec.ts` mirroring the exact reported repro: identity A creates an agent with `respond_to: anyone`, adds it as a bot member of a channel shared with identity B; identity B can now see and use the mention, and a non-member/non-invocable agent stays hidden.
- Verified manually end-to-end against a local dev relay with two separate identities, and separately against a real hosted community with two independent machines/humans.
- `cargo fmt --all -- --check` and `cargo fmt` (Tauri crate) clean.
- `cargo clippy --all-features -- -D warnings` reports zero issues in either file this PR touches (`agent_discovery.rs`, `nostr_convert.rs`). Note: the same command surfaces ~30 pre-existing warnings/errors in unrelated files on a clean `origin/main` checkout under Windows + `--all-features` (confirmed by checking out main and re-running) — those are not introduced by this change and are left untouched per the "no drive-by fixes" guidance in `CONTRIBUTING.md`.
- Full desktop unit suite: 4131/4131 passing. `tsc --noEmit` clean. `biome check` clean on all changed files.

Closes #3776. Related: #3277, #2349, #2950.

### DIFF
--- a/desktop/src/testing/e2eBridge.ts
+++ b/desktop/src/testing/e2eBridge.ts
@@ -2370,6 +2370,28 @@ function resetMockRelayAgents(config?: E2eConfig) {
       respond_to: seed.respondTo ?? "owner-only",
       respond_to_allowlist: seed.respondToAllowlist ?? [],
     });
+
+    // A relay-agent seed that names real channels represents a bot that is
+    // an actual member of those channels — e.g. another identity's agent
+    // added as a bot member of a shared channel — not just a
+    // relay-directory listing. Mirror resetMockManagedAgents's membership
+    // push (below) so specs can exercise the cross-owner "bot member of a
+    // shared channel" case without also declaring the agent as locally
+    // managed.
+    for (const channel of channels) {
+      if (channel.members.some((member) => member.pubkey === seed.pubkey)) {
+        continue;
+      }
+      channel.members.push({
+        pubkey: seed.pubkey,
+        role: "bot",
+        is_agent: true,
+        joined_at: new Date().toISOString(),
+        display_name: seed.name,
+      });
+      syncMockChannel(channel);
+      touchMockChannel(channel);
+    }
   }
 }
 
@@ -3583,7 +3605,7 @@ function initializeMockHuddle(
 const openedExternalUrls: string[] = [];
 const defaultMockRelayAgents: RawRelayAgent[] = [
   {
-    pubkey: ALICE_PUBKEY,
+    pubkey: "5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a",
     name: "alice",
     agent_type: "goose",
     channels: ["general", "agents"],
@@ -3597,7 +3619,7 @@ const defaultMockRelayAgents: RawRelayAgent[] = [
     respond_to_allowlist: [],
   },
   {
-    pubkey: CHARLIE_PUBKEY,
+    pubkey: "5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c5c",
     name: "charlie",
     agent_type: "codex",
     channels: ["general"],
--- a/desktop/tests/e2e/mentions.spec.ts
+++ b/desktop/tests/e2e/mentions.spec.ts
@@ -1933,6 +1933,69 @@ test("shared agents wait for initial directory authorization", async ({
   });
 });
 
+test("bot agents owned by another identity are mentionable when added to a shared channel", async ({
+  page,
+}) => {
+  const OTHER_OWNER_BOT_PUBKEY =
+    "7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b7b";
+  await installMockBridge(page, {
+    relayAgents: [
+      {
+        pubkey: OTHER_OWNER_BOT_PUBKEY,
+        name: "scout",
+        respondTo: "anyone",
+        channelNames: ["general"],
+      },
+    ],
+  });
+  await page.goto("/");
+  await page.getByTestId("channel-general").click();
+  await expect(page.getByTestId("chat-title")).toHaveText("general");
+
+  const input = page.getByTestId("message-input");
+  await input.fill("@scout");
+
+  const dropdown = autocomplete(page);
+  await expect(dropdown.getByText("scout")).toBeVisible();
+  await expect(dropdown.getByText("agent")).toBeVisible();
+
+  await input.press("Enter");
+  await page.keyboard.type(" can you help?");
+  await page.getByTestId("send-message").click();
+
+  const mentionChip = page
+    .getByTestId("message-row")
+    .last()
+    .locator("[data-mention].agent-mention-highlight", { hasText: "scout" });
+  await expect(mentionChip).toBeVisible();
+});
+
+test("bot agents owned by another identity stay hidden when not added to a shared channel", async ({
+  page,
+}) => {
+  const OTHER_OWNER_BOT_PUBKEY =
+    "8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c8c";
+  await installMockBridge(page, {
+    relayAgents: [
+      {
+        pubkey: OTHER_OWNER_BOT_PUBKEY,
+        name: "ghost",
+        respondTo: "anyone",
+        // Deliberately not added to "general" — respondTo=anyone alone must
+        // not be enough; the fix requires real channel membership too.
+      },
+    ],
+  });
+  await page.goto("/");
+  await page.getByTestId("channel-general").click();
+  await expect(page.getByTestId("chat-title")).toHaveText("general");
+
+  const input = page.getByTestId("message-input");
+  await input.fill("@ghost");
+
+  await expect(autocomplete(page)).toHaveCount(0);
+});
+
 test("mentioning an in-channel stopped managed agent starts it before sending", async ({
   page,
 }) => {


wolfyy970:
@oleg-ai-dev @dexsynccom @anindyar, after comparing all three patches, I think each contributes part of the best fix. #4714 keeps the mention change narrow. #4716 correctly identifies the instance-tier policy and adds the cross-identity E2E proof this bug needs. #4713 recognizes that runtime discovery should be signed by the agent itself. I would preserve all three contributions.

I still don’t think #4716’s `kind:30177` path is safe as written. The event is owner-signed, while the agent pubkey comes from an unverified `d` tag. It also represents managed configuration, not necessarily what the running harness currently enforces.

My preferred combination is #4714’s mention fix, #4716’s policy handling and E2E coverage, and an agent-signed `kind:10100` runtime profile. Channel membership remains authoritative in `kind:39002`.

`kind:10100` then needs one publisher to own the complete document; the harness and `set-add-policy` cannot safely publish partial replacement bodies independently.

If the `kind:30177` approach can prove the owner-to-agent binding and effective-runtime semantics, I’m happy to reconsider it.

rmichelena:
Opened #5483 on the same layer, so this is a note from the one other PR that touches `agent_discovery.rs` rather than a competing claim — most of what's here I agree with, including the `39002` membership resolution and the instance-tier vs definition-tier point, which I hadn't seen stated anywhere else.

One factual correction on the premise for repointing, because it changes what the fix should be:

> Nothing in the codebase publishes `kind:10100` as an agent profile

True of Desktop's own code, but not of deployments. Headless `buzz-acp` seats publish their own `kind:10100` — via `buzz channels set-add-policy`, operator scripts, or hand-rolled with `nak`. #2987 is a report from exactly such an operator, and [the 0.5.9 follow-up on that thread](https://github.com/block/buzz/issues/2987#issuecomment-5266401659) documents a working record they publish by hand. #5546 and #5691 are open PRs to have the harness publish it too.

So repointing `list_relay_agents` from `10100` to `30177` doesn't remove a dead read — it trades one blind spot for the mirror-image one:

| Agent | 10100-seeded | 30177-seeded (this PR) |
|---|---|---|
| headless seat, self-published 10100, no 30177 | listed | **dropped** |
| Desktop-managed 30177, never self-published | **dropped** | listed |

The second row is the bug both of us set out to fix; the first row is the population #2987 is about. A headless agent has no Desktop, so nothing publishes a `kind:30177` for it — after this PR those agents would disappear from autocomplete, which is the same symptom at a different codepath.

I've changed #5483 to read the **union** of both kinds, keyed by agent pubkey, then enrich with `39002`. Worth flagging one thing I hit doing that, since it applies to any unfiltered `30177` read: once the query is no longer scoped by a known set of agent pubkeys, the `d` tag is an unverified claim about *someone else's* agent, so each record has to be checked against that agent's own NIP-OA owner (`kind:0`) before it can seed or override a directory entry.

Not asking you to close this — the union is a small change on top of what you already have, and if you'd rather carry it here I'm happy to close mine instead. The part I'd push back on either way is dropping `10100` as a source.

oleg-ai-dev:
Rebased this onto current `main` and resolved the conflict — worth flagging that the resolution changes this PR's scope significantly.

Going through the conflicts, I found that the frontend gate fix (commit 1) and the backend `list_relay_agents` sourcing fix (commit 3) are both now superseded by work that landed on `main` after this PR was opened:

- **Frontend**: #4913 (merged 2026-08-05, the day after this PR opened) renamed `isAgentIdentityInManagedList` → `isAgentIdentityInAllowedList` and wired it to consume `getMentionableAgentPubkeys()`'s output directly, which fixes the same "cross-owner shared channel-member agent isn't mentionable" bug this PR targeted — via a cleaner mechanism than what I had here. #6338 (2026-08-20) extended it further for owner-only builds. @wolfyy970 — this addresses your concern about the `kind:30177` `d`-tag trust: `getMentionableAgentPubkeys` combined with the anti-forgery checks on `main` now cover it.
- **Backend**: `list_relay_agents` on `main` now queries **both** `kind:10100` and `kind:30177` and unions them by pubkey, enriched via `kind:39002`, with ownership verification on the `30177` `d`-tag claim. @rmichelena — this looks like exactly the union approach from #5483; repointing to `30177`-only (what I had here) would have reintroduced the headless-agent regression you flagged, so I dropped that commit rather than force it through.

Both of those commits became empty diffs once rebased onto `main` and were skipped. What's left is regression-test coverage: going through `main`'s existing e2e suite, I found #6338 tests the owner-only-build variant of a cross-owner `respond_to=anyone` relay agent, but the *default*-build path for that exact scenario (the original reported repro) wasn't covered anywhere. Added that here — two Playwright tests plus the mock-bridge plumbing (real channel membership for relay-agent seeds) needed to exercise it.

CI status locally: `tsc --noEmit`, `biome check`, `cargo check`, and the full unit suite (5,397 tests) all pass.

Given `main` already ships the fix, happy to close this in favor of the tests-only diff being cherry-picked elsewhere if that's preferred — let me know.

PR #4723, opened 2026-08-04
fix(git-credential-nostr): accept BUZZ_PRIVATE_KEY with NOSTR_PRIVATE_KEY fallback

### LINKED ISSUE #4712: git-credential-nostr silently ignores BUZZ_PRIVATE_KEY — its env override is NOSTR_PRIVATE_KEY, unlike other components
**Describe the bug**

Most identity-consuming components use `BUZZ_PRIVATE_KEY` (`buzz-cli` declares it as a required arg env at `crates/buzz-cli/src/lib.rs:85`; the ACP/MCP shim passes it to shell children; buzz-admin reads it). The git credential helper alone resolves identity from `NOSTR_PRIVATE_KEY` (`crates/git-credential-nostr/src/lib.rs:51`, `load_key()`), falling back to `git config nostr.keyfile`.

A user who exports the ecosystem-standard `BUZZ_PRIVATE_KEY` gets no error and no warning: the helper silently signs as the key file's identity instead, and the resulting server rejection (`pre-receive hook declined`) is indistinguishable from a permissions problem with the intended identity.

**Steps to reproduce**

1. Configure `git config nostr.keyfile` with identity A (a channel member).
2. `export BUZZ_PRIVATE_KEY=<identity B's key>` where B is the repo owner.
3. Run an owner-gated operation, e.g. `git push <relay-repo-url> :refs/heads/some-branch`.
4. The push is rejected exactly as it would be for identity A — the env var was never consulted.

**Expected behavior**

In order of preference (deferring to maintainers on direction):

1. The helper warns on stderr when `BUZZ_PRIVATE_KEY` is set but unused — pure DX improvement, no new security surface, and it breaks the silent wrong-identity trap.
2. The naming difference is documented where the helper is introduced.
3. Optionally, the helper also accepts `BUZZ_PRIVATE_KEY` (keeping `NOSTR_PRIVATE_KEY` for back-compat) — noting #3478 moved toward keyfile handoff for env-borne keys, so an env alias may be against the intended direction; happy to follow whichever pattern is preferred.

I am glad to submit a PR for the chosen direction (DCO sign-off and Conventional Commits understood).

**Version and platform**

- Buzz version: 0.5.3 (behavior also present at current `main` — `crates/git-credential-nostr/src/lib.rs:51`)
- OS: macOS 15

**Logs / additional context**

Hit while performing an owner-signed ref deletion on a relay-hosted repo: the silent fallback cost a debugging round and a key-file-swap workaround before a source read revealed the variable-name difference. Prior-art searched across open/closed issues and PRs (`NOSTR_PRIVATE_KEY`, `BUZZ_PRIVATE_KEY`, `git-credential-nostr`, credential-helper env synonyms) — closest matches #2316 (silent no-credentials on old git) and #3478 (keyfile migration) do not cover this.

santhiprakash:
## Summary

`git-credential-nostr` was the only Buzz component that did not read `BUZZ_PRIVATE_KEY`. Users who exported the ecosystem-standard env var got silent wrong-identity signing because the helper fell back to `nostr.keyfile` (or no credentials) instead.

This change:
- Resolves identity in this order: `$NOSTR_PRIVATE_KEY`, `$BUZZ_PRIVATE_KEY`, then `git config nostr.keyfile`.
- Warns on stderr when both env vars are set to different values and `$NOSTR_PRIVATE_KEY` wins, so the silent wrong-identity trap is at least audible.
- Updates the helper README and `docs/remote-agents.md` to document the precedence.
- Adds integration tests for `BUZZ_PRIVATE_KEY`-only and the conflict-warning path.

### Related issue

Fixes #4712.

No duplicate PR found (searched open/closed PRs for `4712`, `BUZZ_PRIVATE_KEY`, and `git-credential-nostr`).

### Testing

Targeted package-level verification:

```bash
source bin/activate-hermit
cargo fmt -p git-credential-nostr -- --check
cargo clippy -p git-credential-nostr --all-targets --all-features -- -D warnings
cargo test -p git-credential-nostr
```

All passed:
- `cargo fmt` — clean
- `cargo clippy` — clean (`-D warnings`)
- `cargo test -p git-credential-nostr` — 10 integration tests passed (including the two new regression tests)

Full `just ci` was not run because this is an isolated Rust crate change; the targeted tests above exercise the modified surface.

### DIFF
--- a/crates/git-credential-nostr/README.md
+++ b/crates/git-credential-nostr/README.md
@@ -30,13 +30,18 @@ That's it. Use git normally — `git clone`, `git push`, `git fetch`.
 
 ## CI / CD
 
-Set `$NOSTR_PRIVATE_KEY` instead of a key file. The env var takes precedence
-over `nostr.keyfile` and avoids touching the filesystem:
+Set `$NOSTR_PRIVATE_KEY` or `$BUZZ_PRIVATE_KEY` instead of a key file. The
+env vars take precedence over `nostr.keyfile` and avoid touching the
+filesystem:
 
 ```bash
 export NOSTR_PRIVATE_KEY=nsec1...
 git clone https://relay.example.com/git/owner/repo.git
 ```
+
+`$BUZZ_PRIVATE_KEY` uses the same name as the rest of the Buzz ecosystem
+(`buzz-cli`, the ACP/MCP shim, and `buzz-admin`). `$NOSTR_PRIVATE_KEY` is
+still accepted for backward compatibility and wins when both are set.
 
 ## How It Works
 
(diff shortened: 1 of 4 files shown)


santhiprakash:

PR #4724, opened 2026-08-04
Dock Buzz Term within channel workspace

wesbillman:
## Summary

- replace Buzz Term's full-app takeover with a resizable bottom dock inside the channel content surface
- add a discoverable channel-header button plus hide and maximize/restore controls
- create PTYs lazily and keep separate, persistent terminal workspaces per channel
- capture immutable channel/thread context on every terminal session

## Multiple-channel behavior

The dock is a single surface, but its tabs are partitioned by channel. Switching channels swaps to that channel's sessions without terminating background PTYs; returning restores them. New tabs capture the currently visible channel/thread context.

## Verification

At commit `7ca087f8e08c80528387684364a65bf4ccd6315f`:

- `pnpm --dir desktop test` — 4,129 passed
- pre-push repository hooks — desktop check/test, Tauri checks, terminal Rust suites all passed


wesbillman:
I found two lifecycle/viewport defects that need correction before this is safe to merge:

1. **Inactive channel sessions are never resized when restored.** `handleSize` resizes only the connection selected by the current global `activeKey` (`TerminalBootstrap.tsx:248-268`). Switching channels changes `activeKey` (`218-230`) but does not call `resize`; the substrate's `ResizeObserver` only reports geometry changes (`TerminalSubstrate.tsx:237-246`). Repro: open A, switch to B, resize the dock/window, return to A. A's PTY remains at its old rows/columns, so backend wrapping/cursor geometry and the canvas viewport diverge. Resize the newly selected/restored connection to `sizeRef.current` (serialized through the existing resize chain), and cover the cross-channel restore case.

2. **Closing a tab while `TerminalConnection.attach()` is pending leaks the eventual PTY/subscription.** With `connection === null`, close removes the session immediately (`TerminalBootstrap.tsx:286-297`). If attach then resolves, the component is still mounted, so the resolution path only runs `update(...)` (`177-193`); because the session was removed, that is a no-op and `connection.detach()` is never called. Keep cancellation/tombstone state outside the removed React entry, or check that the key still exists when attach resolves and detach immediately if not. Add a deferred-attach close test that proves the eventual connection is detached/closed.

The per-channel immutable spawn context itself looks sound. I reviewed exact head `7ca087f8e08c80528387684364a65bf4ccd6315f`; I did not duplicate the already-reported full suite.

klopez4212:
🤖 Updated docked Buzz Term controls.

The terminal tabs now use the shared gutter, compact Lucide controls, and immediate tab-dismiss feedback.

terminal-tabs

Issue #4098, opened 2026-08-01
[Bug] Hermes ACP initializes but every managed prompt hangs until 60s timeout on Windows


godhiraj-code:
## Describe the bug

A Buzz Desktop managed Hermes ACP agent initializes successfully, but every incoming prompt hangs until `buzz-acp` reports a 60-second request timeout.

This happens even for a one-word prompt and is not caused by worker execution, Kanban, or a long model response.

## Steps to reproduce

1. Run Buzz Desktop 0.5.3 on Windows 10 AMD64.
2. Configure a managed Hermes Agent ACP runtime using Hermes Agent 0.19.1.
3. Configure an explicit model ID: `gpt-5.6-sol`.
4. Send a simple prompt such as `Hi`.
5. Wait for the response.

## Expected behavior

The agent should return a short response within a few seconds.

## Actual behavior

The message is accepted by Buzz, but no assistant response appears. After approximately 60 seconds, Buzz logs:

```text
Request timeout — agent did not respond within 60s
```

Buzz then requeues the failed batch and respawns the Hermes ACP process.

## Evidence

Buzz's harness log shows that:

- ACP initialization succeeds.
- Hermes creates the OpenAI Codex client successfully.
- The configured model is `gpt-5.6-sol`.
- MCP tools register successfully.
- No model response or tool event appears before the timeout.

Sanitized log excerpts:

```text
buzz-acp: agent initialized
configured_model=gpt-5.6-sol
OpenAI client created provider=openai-codex model=gpt-5.6-sol
MCP: registered 7 tool(s) from 1 server(s)
Request timeout — agent did not respond within 60s
```

The same minimal prompt succeeds immediately when run directly through Hermes CLI:

```text
MANAGER_OK
```

This isolates the failure to the Buzz ACP request/response path rather than Hermes, model authentication, or the model endpoint.

The Buzz startup log also reports:

```text
permission_mode=bypassPermissions
```

Please confirm whether Buzz sends a `session/set_config_option` permission-mode request to Hermes ACP. If Hermes does not support that configuration vocabulary, the error may be waiting indefinitely instead of being returned as a handled ACP error.

## Environment

- Buzz Desktop: 0.5.3
- OS: Windows 10 AMD64
- Hermes Agent: 0.19.1
- ACP model: `gpt-5.6-sol`
- Harness: managed Hermes Agent ACP
- Direct Hermes CLI: same machine and credentials, succeeds

## Related issues

[...]

PR #4732, opened 2026-08-04
fix(acp): kill the whole process tree on Windows, not just the wrapper

TheSmokeDev:
## Problem

`AcpClient::shutdown()` kills the process **group** on Unix and falls back to `start_kill()` everywhere else, because `kill_process_group` is `#[cfg(not(unix))] -> false` (`crates/buzz-acp/src/acp.rs`). On Windows that fallback is not equivalent.

Rust runs a `.cmd`/`.bat` shim through an intermediate `cmd.exe` (post-CVE-2024-24576 behaviour), and npm installs the ACP agents as exactly that kind of shim — `claude-code-acp`, `hermes-acp`. So `start_kill()` reaps the `cmd.exe` wrapper and leaves the real agent — plus any MCP servers it started — alive.

Every timeout-then-respawn cycle therefore strands another live agent for the rest of the session. On a repeating failure it accumulates: I found this while tracing #4098, where the 60s timeout respawns the child in a loop.

## Change

Adds the Windows arm using `taskkill /T /F`, which covers the tree the way `killpg` does on Unix, so teardown is symmetric across platforms.

It deliberately mirrors `taskkill_tree` in `desktop/src-tauri/src/managed_agents/process_lifecycle.rs` — the same problem, already solved once for managed-agent teardown — including its `CREATE_NO_WINDOW` flag so cleanup never flashes a console at a GUI user. That also matches the existing `configure_no_window` helper in this file.

- No `unsafe`, so the crate's `#![deny(unsafe_code)]` policy is preserved.
- The Unix path is untouched; the new code is `#[cfg(windows)]`.
- The `#[cfg(not(any(unix, windows)))]` fallback stays for other platforms.

## Verification

On Windows: `cargo check -p buzz-acp`, `cargo clippy -p buzz-acp --all-targets` (0 warnings) and `cargo fmt -p buzz-acp -- --check` (no diffs) are all clean.

**I am not offering the crate's test suite as evidence either way.** On this Windows box it is non-deterministic — a run against untouched `main` fails a different set each time, which is consistent with #2492 — so a before/after comparison would be noise, not a receipt. Happy to run anything specific you'd trust more.

Related: #4098 (the respawn loop that surfaced this), #2492 (`just ci` on Windows).

Duplicates: searched open PRs for `taskkill`, `kill_process_group`, and `process tree` — none found.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -422,8 +422,9 @@ impl AcpClient {
         // ensures subprocesses (MCP servers, tool processes) are cleaned up
         // rather than orphaned to init.
         //
-        // Falls back to start_kill() (direct child only) on non-Unix or if
-        // the child has been polled to completion (id() returns None).
+        // Falls back to start_kill() (direct child only) when the platform
+        // has no tree-kill or the child has been polled to completion
+        // (id() returns None).
         match self.child.id() {
             Some(pid) if kill_process_group(pid) => {}
             _ => {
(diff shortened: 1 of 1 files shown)


TheSmokeDev:
Re-ran the guide's gates for server-side crates, and replaced my earlier hand-wave about the test suite with an actual control.

**Gates** (Windows, this branch):

**On the suite.** I previously said I wouldn't offer it as evidence because it's non-deterministic here. That was true but unhelpful, so I measured it properly.

Same tree, same box, only difference is whether this commit's change to `acp.rs` is present:

| tree | `session_new_full` family |
|---|---|
| this branch | 6 failed |
| this branch, `acp.rs` reverted to `HEAD~1` | **6 failed — identical set** |

So the change causes none of them. The non-determinism is real and separate: two runs of the *same* branch code gave 20 and 13 failures, and the assertions that move are timing-based (`assertion failed: elapsed >= Duration::from_millis(400)`, keepalive/idle deadline tests). Consistent with #2492.

Also worth flagging while I was in there, unrelated to this PR: running `cargo test -p buzz-acp --lib` on Windows drops six files into `crates/buzz-acp/` named like `C:UsersDegenAppDataLocalTempbuzz-acp-steer-captureacp_shape.json` — the full temp path with its separators stripped, created as a single filename in the crate dir. `capture_path()` (`acp.rs:3952`) builds the path correctly with `.join()`, so something downstream is flattening it. Untracked debris, but it sits where a `git add -A` would sweep it up. Happy to open a separate issue if useful.

DCO sign-off corrected to my real address (was a placeholder from a misconfigured global git identity — the check passed either way, since it only compares the trailer to the author).

PR #4733, opened 2026-08-04
fix(acp): name the method in request-timeout errors

TheSmokeDev:
## Problem

`AcpError::Timeout` carries only a `Duration`, so every request timeout produces the same sentence regardless of which call hung:

```
Request timeout — agent did not respond within 60s
```

That is the message users report, and it is not enough to place the fault. Sessions are created lazily inside the prompt task, so from the outside a stalled `session/new` and a stalled `session/prompt` are indistinguishable — even though one is bounded at `REQUEST_TIMEOUT` (60s) and the other at `DEFAULT_IDLE_TIMEOUT_SECS` (900s).

Concretely: diagnosing #4098 ("every managed prompt hangs until 60s") required tracing two codebases and reasoning backwards from the timeout constants to establish that only `session/new` could produce that number. The error itself should have said so.

## Change

`send_request` already has the method in hand, and it is `&'static str` at every call site, so carrying it costs nothing:

```
Request timeout — agent did not respond to session/new within 60s
```

The two directly-constructed timeouts in `pool.rs` name theirs as well (`session/set_model`, `session/set_config_option`), so no timeout in the crate is anonymous.

Match sites become `Timeout { .. }`. Nothing else changes — the transport-class error classification in `lib.rs` and the retry handling in `pool.rs` behave exactly as before.

## Verification

On Windows: `cargo check -p buzz-acp`, `cargo clippy -p buzz-acp --all-targets` (0 warnings) and `cargo fmt -p buzz-acp -- --check` (no diffs) all clean.

Not offering the crate's test suite as evidence: on this box it is non-deterministic against untouched `main` (different failures per run, consistent with #2492), so a before/after comparison would be noise rather than a receipt.

## Note on a related asymmetry

Not changed here, but worth flagging while you have the file open: `session/new` gets 60s while the prompt it precedes gets 900s idle / 7200s hard. An agent that legitimately needs ~90s to provision a session can therefore never succeed on any runtime, no matter how patient the prompt budget is. Happy to open that separately if it's worth addressing.

Related: #4098. Duplicates: searched open PRs for `AcpError::Timeout` and `request timeout` — none found.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -95,8 +95,15 @@ pub enum AcpError {
     #[error("Agent did not stop within {0:?} after cancellation")]
     CancelDrainTimeout(std::time::Duration),
 
-    #[error("Request timeout — agent did not respond within {0:?}")]
-    Timeout(std::time::Duration),
+    #[error("Request timeout — agent did not respond to {method} within {elapsed:?}")]
+    Timeout {
+        /// The JSON-RPC method that went unanswered. Without it a caller
+        /// cannot tell a slow `session/new` from a slow `session/prompt`,
+        /// and sessions are created lazily inside the prompt task — so the
+        /// two are indistinguishable from the outside (see #4098).
+        method: &'static str,
+        elapsed: std::time::Duration,
+    },
 
     #[error("Write timeout — agent stopped reading stdin (blocked for {0:?})")]
     WriteTimeout(std::time::Duration),
@@ -1075,7 +1082,7 @@ impl AcpClient {
     /// if they don't, the agent is likely stuck and we must not block forever.
     async fn send_request(
         &mut self,
-        method: &str,
+        method: &'static str,
         params: serde_json::Value,
     ) -> Result<serde_json::Value, AcpError> {
         let id = self.next_id;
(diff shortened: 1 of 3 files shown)


TheSmokeDev:
Good catch — fixed in `0b0d854`.

`apply_model_switch` serves both `ModelSwitchMethod` variants, so hardcoding `session/set_model` in the outer-timeout branch would have named a method the ConfigOption path never sent. That is worse than the anonymous timeout this PR set out to fix: a wrong name sends whoever reads it looking in the wrong place.

Now derived from the variant:

```rust
let rpc_method = match method {
    ModelSwitchMethod::ConfigOption { .. } => "session/set_config_option",
    ModelSwitchMethod::SetModel { .. } => "session/set_model",
};
```

`method_label` beside it stays as it was — prose for humans, not an RPC name. `cargo check`, `clippy --all-targets` (0 warnings) and `fmt --check` still clean.

TheSmokeDev:
Added the regression test CONTRIBUTING asks for. My earlier note explained why I wasn't offering the *suite* as evidence, which isn't the same thing as saying a targeted test was impractical — and here it wasn't.

`42336a6` moves the mapping into `rpc_method_for()` and asserts each `ModelSwitchMethod` variant names the RPC it actually sends. The extraction is part of the fix rather than a drive-by: the defect was the dispatch and the error text disagreeing, and a free function is what lets a test hold them together as variants are added. It's also the exact shape the review bot caught on the first push, so it seemed worth locking.

Verified to **fail** against the pre-fix behaviour (both arms returning `session/set_model`) and pass with the fix.

**Gates** (Windows, this branch):
- `cargo test -p buzz-acp --lib rpc_method` — 1 passed

On the full suite's non-determinism, I ran a proper control on the sibling PR (#4732): same tree with the change reverted produces an identical failure set, and two runs of identical code gave 20 vs 13 failures, all timing assertions. Consistent with #2492.

DCO sign-off corrected to my real address (was a placeholder from a misconfigured global git identity — the check passed either way, since it only compares the trailer to the author).

PR #4671, opened 2026-08-04
feat(projects): support multiple repositories

thomaspblock:
## Summary
- adopt the finalized NIP-MP project model so one project can enumerate and switch between multiple NIP-34 repositories
- add project and repository navigation, activity summaries, existing-repository attachment, and repository access-channel management
- preserve privacy-safe activation provenance for agent-authored patches, pull requests, issues, and associated commits

## Test plan
- [x] Run desktop typecheck and unit tests
- [x] Run focused NIP-MP, repository access, and provenance tests
- [x] Run Rust formatting and desktop lint checks
- [x] Run the complete pre-push suite after merging current `main`
- [ ] Manually verify project creation, repository attachment, switching, and access repair on staging
- [ ] Manually verify public-channel and private-agent origin labels on newly created Git activity

Related: [#4695](https://github.com/block/buzz/pull/4695)

PR #4735, opened 2026-08-04
feat(desktop): bind agents to Project connections

wolfyy970:
## Summary

Connects reusable agent tool requirements to Project-owned connections.

A template says which capabilities an agent needs. When starting or editing the agent, its owner chooses a Project and compatible connection. The agent stores the binding; endpoints and credentials remain with the Project.

This lets one connection serve several agents without copying secrets into portable agent data. It also makes the agent's Project access clear before it runs.

## Behavior

- Starting an agent asks for its Project and required connections.
- Editing an agent keeps that agent as the primary object. Editing its template remains a separate action.
- Idle agents save and restart immediately. Active agents keep their current task and use the new settings at their next restart.
- Changing Project clears incompatible bindings.

A selected MCP connection currently grants every tool exposed by that server. Exact tool allowlists are outside this PR.

The same ownership boundary is now used by #4571 for hosted agents.

## Stack

Depends on #4164, #4588 and #4671. This branch includes those heads so the complete flow can be tested now.

## Validation

- `just ci`
- Desktop screenshot flow passed


wolfyy970:
### Edit the agent

The agent stays at the top of the hierarchy. Its template is a separate edit, and Project access is explicit before it runs.

01-agent-configuration

### Save without interrupting work

A running agent keeps its current task. Buzz explains when the saved configuration will be used.

02-running-agent-update-ready

### Confirm the restart

Buzz confirms when the agent has restarted with its new configuration.

03-running-agent-update-success

wolfyy970:
I'm closing this draft because #4940 is establishing relay-canonical managed-agent authority, which is the right foundation for this work. The Project connection UX and tested local implementation remain useful, but I will restack that binding on the canonical path instead of asking maintainers to review a competing branch.

PR #4737, opened 2026-08-04
fix(desktop): close reconnect gaps that previously required CMD+R

wesbillman:
> Opened by Brain (agent) on behalf of @wesbillman.

## Problem

Users report the desktop app doesn't reliably reconnect and can wedge in states where only CMD+R (or a full restart) restores connectivity (thread `c2205e2b` in #desktop-reconnecting).

Pinky's empirical light-switch matrix (real `buzz-relay`, SIGTERM/1012 + SIGKILL × 1s/45s/3min, at `f18a9cb10`) passed 4/4 — the backoff state machine recovers cleanly from ordinary relay loss. That isolates the user-stuck states to four special cases a reload resets but the auto flow never did.

## Fixes

| Gap | Change |
|---|---|
| **G1** — recovery rode solely on the backoff timer (max 30s), throttled by WKWebView in occluded/background windows; nothing fired on network return or wake | New `useRelayResumeTriggers`: `online`, window focus, and visibility→visible call `preconnect()` when the session is `reconnecting`/`stalled`, rate-limited to one attempt per 5s (`relayResumeTriggerPolicy.ts`). Deliberately inert for the terminal `disconnected` state. |
| **G2** — any AUTH `OK false` latched the session terminal forever, though the relay also rejects for transient causes (duplicate-AUTH "already authenticated" race, ±60s clock skew, fail-closed allowlist DB errors) | New `AuthOkTracker` (`relayAuthPolicy.ts`): "already authenticated" resolves as success; transient rejections retry with normal backoff; latch only on `restricted:` or after 3 consecutive rejections. |
| **G3** — an `auth-required:` CLOSED (REQ racing AUTH after reconnect) permanently deleted the live subscription with no UI signal — frozen channel while state reads "connected" | Reclassified `auth-required:` as retryable in `relayClosedPolicy.ts`. Genuinely terminal classes (`restricted:`, `invalid:`, …) still delete. Can't loop: a truly unauthenticated session latches terminal at the connection level. |
| **G4** — `useRelayAutoHeal` observed the 2s-debounced connection hook, so sub-2s flaps never triggered the heal even though `resetConnection` had already rejected every in-flight query | Auto-heal now observes the raw connection-state emitter. The existing 15s heal rate-limit still guards against flap storms. |

Each fix is a colocated pure-policy module + unit tests, matching the existing `relayReconnectPolicy`/`relayClosedPolicy` pattern.

## Validation

- Full desktop unit suite: **4151 pass, 0 fail** (at branch tip, `pnpm -C desktop test`)
- `pnpm -C desktop typecheck` and `pnpm -C desktop check` clean (file-size ratchet respected — `relayClientSession.ts` net −2 lines despite the tracker wiring)
- Evidence trail: `RESEARCH/DESKTOP_RECONNECT_CMDR_GAP_AUDIT.md` (audit), `RESEARCH/DESKTOP_RECONNECT_LIGHT_SWITCH_RESULTS.md` (Pinky's matrix)

## Not covered / follow-ups

- Native macOS sleep-wake was not automated (would kill the harness session); G1's focus trigger is the mechanism that covers wake in practice, but a manual sleep-wake verification on a real build is worthwhile.
- G3 terminal-CLOSED classes (`restricted:` etc.) still silently delete subs with no UI signal — surfacing that is a separate UX decision.
- Stall-watchdog latency (60s idle + 10s check) left unchanged; G1 triggers largely mask it.


desktop/src/shared/api/relayClientSession.ts
```diff
@@ -430,6 +433,7 @@ export class RelayClient {
     // reconnect also bypasses the current delay once; ordinary operations do
     // not, so background traffic cannot continuously defeat backoff.
     this.terminal = false;
+    this.authOkTracker.reset();
```
wpfleger96, line 434: **Important — correctness:** This reset is no longer limited to explicit user re-engagement: the new focus/online/visibility hook calls `preconnect()` automatically whenever the state is `reconnecting` or `stalled`. During repeated AUTH rejection, a resume event (including one arriving while the next handshake is in flight) clears the streak before the next `OK false`, so `MAX_CONSECUTIVE_AUTH_REJECTIONS` may never be reached and the session can retry indefinitely instead of surfacing `disconnected`. Please give resume triggers a separate ‘bypass pending backoff’ path that preserves the AUTH/terminal policy state, and reserve this reset for the reconnect card/community switch. An integration test that interleaves rejected AUTH attempts with resume events should still reach terminal on the third rejection.

desktop/src/shared/api/relayAuthPolicy.ts
```diff
@@ -0,0 +1,56 @@
...
+      success ||
+      normalized.startsWith("auth-required: already authenticated")
+    ) {
+      this.consecutiveRejections = 0;
+      return "authenticated";
+    }
+
+    this.consecutiveRejections++;
+
+    if (normalized.startsWith("restricted:")) return "terminal";
```
wpfleger96: **Important — reliability:** The relay’s permanent ban response is `blocked: you are banned from this community` (`crates/buzz-relay/src/handlers/auth.rs`), not `restricted:`. This currently classifies a known ban as `retry`, producing unnecessary reconnect/AUTH attempts and delaying the truthful terminal state; with the resume-trigger reset above it can keep retrying. Please classify `blocked:` as terminal immediately and cover the relay’s actual ban string in this policy test.

wpfleger96 (changes requested):
The reconnect coverage is strong, but two AUTH policy interactions can keep permanent failures retrying instead of reaching the intended terminal state.

desktop/src/shared/api/relayClientSession.ts
wesbillman: > Posted by Brain (agent) on behalf of @wesbillman.

Confirmed and fixed in b0853e027. Resume triggers no longer call `preconnect()` — they use a new `resumeReconnect()` that bypasses the pending backoff timer but preserves both the terminal latch and the AUTH rejection streak; only the reconnect card / community switch (`preconnect()`) resets them.

The integration test you asked for is in: three rejected AUTH attempts interleaved with `online`/`focus` bursts still latch `disconnected` on the third rejection, and further resume events do not revive the session. Building that test also exposed a latent race worth noting: a late failure from an in-flight publish/subscribe retry landing *after* the terminal latch would call `resetConnection()` with default options and paint `reconnecting` over the terminal `disconnected` state — permanently. `resetConnection` now preserves `disconnected` when the session is terminal.

desktop/src/shared/api/relayAuthPolicy.ts
wesbillman: > Posted by Brain (agent) on behalf of @wesbillman.

Confirmed and fixed in b0853e027. `blocked:` now latches terminal immediately, and the policy test covers the relay's exact ban seam string (`blocked: you are banned from this community`, from `handlers/auth.rs`).

wpfleger96 (approved):
Both prior findings are addressed at this head: environment resume now bypasses backoff without resetting AUTH/terminal policy state, and `blocked:` ban responses latch terminal immediately. The interleaved resume/AUTH rejection regression is covered end to end. No remaining blocking findings.

Issue #4591, opened 2026-08-03
Desktop: "Remove agent" in the team-edit flow deletes the agent's private key without saying so


redirwin:
## Summary

When you edit a team and deselect a member, a confirm dialog appears with three
buttons: Cancel, Keep agent, and Remove agent. The dialog text is "X will be
removed from this team. Do you also want to remove the agent completely?"
(`desktop/src/features/agents/ui/RemoveMembersConfirmDialog.tsx:39-40`).

What the dialog doesn't say is that "Remove agent" also deletes the agent's
nsec from the OS keyring. Unless the owner has a copy of the key somewhere
outside the app, the agent's identity is gone for good.

There is a second, smaller problem in the same flow: the loop that deletes the
removed personas swallows every error
(`desktop/src/features/agents/ui/useTeamActions.ts:194-200`, see the
"Best-effort" comment), so a partial or failed delete never surfaces.

To be clear, I'm not asking to change the cascade itself. The persona-to-agent
cascade is deliberate and covered by tests
(`desktop/src-tauri/src/commands/personas/delete_cascade_tests.rs`). The
problem is that an irreversible, key-destroying action sits behind a dialog
that reads like a routine membership question.

This bites hardest when a pre-existing agent was only temporarily added to a
team. The dialog treats that agent the same as one created for the team, and
since "Remove agent" is the button that completes the removal the user already
started, it's easy to read it as the thorough version of un-seating rather
than as identity destruction.

## Steps to reproduce

1. Create a managed agent and assign it to a team.
2. Edit the team, deselect the agent, save.
3. In the confirm dialog, click "Remove agent".
4. The managed-agent record is deleted from `managed-agents.json` and the nsec
   is removed from the OS keyring (`delete_agent_key`,
   `desktop/src-tauri/src/commands/personas/mod.rs:238`, inside
   `delete_persona`, `mod.rs:113`). If the key existed nowhere else, the
   identity is unrecoverable.

## What we observed (real incident, 2026-08-03)

I removed a working agent from a team through this flow, expecting a
membership change. I did not register the third option as anything destructive
at the time; I only understood what it did when I retested the flow after the
loss. The result was the record and keyring nsec gone and the agent offline
for about 46 minutes.

We recovered only because I happened to keep a copy of the nsec outside the
app. With the key in hand we restored the record and then pulled the auth tag
and profile back off the agent's live kind:0 on the relay. Everything except
the nsec turned out to be recoverable. The private key is the single point of
no return.

[...]

Issue #3085, opened 2026-07-27
Snapshot import has no update path: reimport duplicates, and the team it created cannot be deleted


finwitz:
# Snapshot import has no update path: reimport duplicates, and the team it created cannot be deleted

## Summary

Importing a `.team.json` snapshot is a one-way operation. There is no way to
update an imported team in place, and no practical way to remove one, so any
change to the source of a snapshot leaves the operator stuck between duplicates
and a manual 27-step cleanup.

Two behaviours combine into the dead end:

1. **Import always mints new agents.** `confirmAgentSnapshotImport` is
   documented as importing "as a brand-new agent with fresh keys", and snapshots
   deliberately strip every lineage field (`persona_id`, `persona_source_version`,
   `source_team`, `source_team_persona_slug`). Nothing on the wire identifies two
   snapshots as versions of the same agent, so reimport can only duplicate.

2. **Delete refuses while members exist.** `delete_team_with_cascade` cascades
   personas that came from a `source_dir`, but a JSON-imported team has none, so
   `agents_referencing_team` is non-empty and the delete hard-fails.

## Steps to reproduce

1. Author a `buzz-team-snapshot v1` file with 27 members.
2. Import it in Desktop. 27 agents are created and joined to a new team.
3. Change one member's system prompt at the source and reimport.
   → 54 agents. No prompt was updated.
4. Try to delete the team to start clean.

```
Cannot delete team "5ff0eeb8-07c6-4c45-9ba6-4cbf57fbbc89": 27 agent(s) still
reference it (Blender Addon Engineer, Economy Designer, Game Audio Engineer,
... 24 more). Delete or reconfigure them first.
```

The only supported path is deleting 27 agents individually in the UI. There is
no bulk delete, and no `buzz agents delete` in the CLI. Editing
`managed-agents.json` by hand would orphan the relay-side kind:30175 persona
heads that `delete_team_with_cascade` documents as needing NIP-09 tombstones.

## Impact

This scales with team size, and team snapshots exist precisely to make large
teams importable in one step. A 27-agent team is 27 manual deletions per
iteration. It also makes snapshots unusable as a distribution format for any
library that changes over time, since consumers can receive a v1 but never a
v2.

## Proposed fixes

Three, listed smallest first. They are independent.

**1. `cascade_members` on team delete.** An explicit opt-in flag on the delete
command that removes referencing agents and enqueues their tombstones, reusing
the cascade path that already exists for `source_dir` personas. Keeps the
[...]

PR #4740, opened 2026-08-04
fix(desktop): align team agent deletion feedback

SmithLabsLLC:
## Summary

- Block deletion of managed-agent instances while their persona is still in an agent team.
- Show the same actionable team-membership error from the agent card and profile settings.
- Check team membership before provider shutdown and enforce it again at the backend boundary.
- Clarify the cleanup order when a team cannot be deleted because instances still reference it.

## Problem

Deleting a running built-in agent from the card could report only that it was assigned to a managed agent, even when team membership was the blocker. Deleting the same agent from profile settings could then show a success toast while the team kept the persona visible.

## Implementation

- Prioritize the team-membership guard for persona removal.
- Fresh-load teams before either frontend managed-agent deletion path.
- Reject managed-agent deletion under the backend store lock before process, key, archive, or provider cleanup.
- Keep deletion available after the persona has been removed from all current teams.

## Manual test

1. Add a running agent to a team.
2. Try Delete from the agent card's three-dot menu.
3. Confirm the toast says to remove the agent from every team first.
4. Open the agent profile, use the settings gear, and confirm deletion.
5. Confirm the same error appears, no success toast appears, and the agent still exists.
6. Remove the persona from all teams and confirm managed-agent deletion works normally.

## Testing

- Rust persona validation tests: 8 passed
- Rust managed-agent deletion tests: 2 passed
- Rust team deletion message test: 1 passed
- Managed-agent control action tests: 7 passed
- TypeScript type-check and E2E production build: passed
- Playwright regression covering both deletion entry points: passed
- Focused Biome, Rust formatting, file-size, pixel-text, public-key truncation, and diff checks: passed

The full repository-wide `just ci` release suite was not run locally; focused change-specific checks passed and upstream CI will run on this reviewed commit.

### Agent card

Before: the card reports only the managed-agent assignment.

before-card

After: the card reports the current team membership and required cleanup.

after-card

### Profile settings

Before: profile deletion reports success even though the team keeps the persona visible.

before-profile

After: profile deletion reports the same team-membership blocker and leaves the agent intact.

after-profile

## Related work

No exact duplicate issue or pull request was found. Issues #4591 and #3085 discuss related team cleanup problems but not this two-entry-point deletion inconsistency.

## Deferred

- The pre-existing retired-persona edge case is outside this focused fix.
- Internal error-string deduplication can be handled separately if the message becomes a broader API contract.

### DIFF
--- a/desktop/src-tauri/src/commands/agents.rs
+++ b/desktop/src-tauri/src/commands/agents.rs
@@ -9,10 +9,10 @@ use crate::{
         load_teams, managed_agent_avatar_url, normalize_agent_args, provider_deploy,
         resolve_provider_binary, save_managed_agents, start_managed_agent_process,
         stop_managed_agent_process, stop_managed_agent_workspace_pair,
-        sync_managed_agent_processes, try_regenerate_nest, validate_provider_config, BackendKind,
-        CreateManagedAgentRequest, CreateManagedAgentResponse, ManagedAgentRecord,
-        ManagedAgentSummary, RelayMeshConfig, DEFAULT_ACP_COMMAND, DEFAULT_AGENT_PARALLELISM,
-        DEFAULT_AGENT_TURN_TIMEOUT_SECONDS,
+        sync_managed_agent_processes, try_regenerate_nest, validate_managed_agent_team_deletion,
+        validate_provider_config, BackendKind, CreateManagedAgentRequest,
+        CreateManagedAgentResponse, ManagedAgentRecord, ManagedAgentSummary, RelayMeshConfig,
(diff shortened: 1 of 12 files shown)


SmithLabsLLC:
## Before and after screenshots

### Agent card

Before: the card reports only the managed-agent assignment.

before-card

After: the card reports the current team membership and required cleanup.

after-card

### Profile settings

Before: profile deletion reports success even though the team keeps the persona visible.

before-profile

After: profile deletion reports the same team-membership blocker and leaves the agent intact.

after-profile

PR #4746, opened 2026-08-04
fix(acp): honor persona thread_replies in prompt builder

### LINKED ISSUE #4699: buzz-acp ignores `thread_replies` — new-thread instruction always forces `--reply-to`
## Summary

`thread_replies` from a persona pack has no effect. An agent configured with
`thread_replies: false` still opens a new thread for every top-level channel message,
because `append_new_thread_reply_instruction()` unconditionally appends an `IMPORTANT:`
directive telling the model to use `--reply-to <event_id>`.

This is the same class of gap as #2899 (`mcp_servers` parsed and validated but never
plumbed to the runtime).

## Why this should work

`PERSONA_PACK_SPEC.md` §10 states the behavioral config fields are the harness's job:

> The behavioral config fields in a persona's frontmatter control how the agent
> participates in Buzz conversations. **These are all Buzz-specific — the agent runtime
> has no awareness of them.**

And `thread_replies` is documented as:

> | `thread_replies` | bool | `true` | `true` / `false` | Reply in-thread when the
> triggering message is in a thread. |

Since the agent runtime is explicitly not aware of these fields, `buzz-acp` is the only
component that can honour them.

## Actual behaviour

`thread_replies` does not appear anywhere in `crates/buzz-acp`. It exists only in
`crates/buzz-persona` (parse, merge, resolve, manifest) and `crates/buzz-db`. It is
parsed, merged and validated — `buzz pack inspect` even prints `Thread replies: false` —
and then dropped.

The prompt builder in `crates/buzz-acp/src/queue.rs` takes the `else` branch for a
top-level channel message and always calls:

```rust
fn append_new_thread_reply_instruction(s: &mut String, event_id: &str) {
    s.push_str(&format!(
        "\nIMPORTANT: This is a new top-level message. For ordinary replies in \
         this turn, use `--reply-to {event_id}` on `buzz messages send` — the \
         triggering message is the thread root. Do NOT reply into any other \
         (older) thread. ..."
    ));
}
```

There is no configuration path that suppresses this.

## Why prompting cannot fix it

We tried. Both the persona system prompt and a patched base prompt instruct the agent to
reply flat in the channel and to ignore any reply target suggested in `[Context]`. The
model still threads, and inconsistently so — flat at 13:03, threaded at 14:28 on the same
kind of question.

The reason is prompt position and specificity. The harness directive is marked
`IMPORTANT:`, carries the concrete event ID, and sits at the very end of the prompt; the
persona rule is abstract and much earlier. The concrete, closer instruction wins. Verified
in the agent's own request logs — the model received the directive and complied exactly:

```
buzz messages send --channel 29c37e15-… --reply-to cb710fe8… --content -
```

## Reproduce

1. Persona pack with `thread_replies: false`; confirm with `buzz pack inspect`.
2. Run `buzz-acp` (any agent command) subscribed to a channel.
3. Post a new top-level message in that channel — not inside a thread.
4. The reply is threaded under the triggering message.

Observed on `3d7712cc36e8da563cb1c121fc58bfc505d38496`, still present on `main` at time of
writing.

## Suggested fix

Gate the call site on the resolved `thread_replies` value. The split is already in the
right place — `append_reply_instruction()` (replying *inside* an existing thread) is
correct as-is and should keep working; only the new-thread branch needs to respect the
flag.

Note that the unit tests in `queue.rs` assert on the current wording
(`"new top-level message"`, `--reply-to <id>`), so they would need a `thread_replies:
false` counterpart.

## Workaround

Patching `append_new_thread_reply_instruction()` at build time to emit the opposite
instruction. Works, but it means carrying a source patch against a pinned upstream commit.

santhiprakash:
## Summary

Fixes #4699.

`thread_replies` was parsed and merged in `buzz-persona` but never read in `buzz-acp`. The harness always injected an `IMPORTANT:` directive forcing `--reply-to <event_id>` for new top-level channel mentions, so a persona with `thread_replies: false` still opened a thread.

This PR plumbs the resolved flag through `Config` / `PromptContext` / `FormatPromptArgs` with a default of `true` (per `PERSONA_PACK_SPEC.md`). When `thread_replies` is `false`, `format_prompt` skips the forced reply anchor for new top-level channel mentions. Existing thread and DM anchors continue to work.

## What changed

- Added `--thread-replies` / `BUZZ_ACP_THREAD_REPLIES` to `CliArgs` and `Config`.
- Threaded the value through `PromptContext` and `FormatPromptArgs`.
- Skipped `resolve_reply_anchor` for human-facing top-level channel turns when `thread_replies == false`.
- Added `queue.rs` unit tests covering:
  - top-level suppression when `thread_replies: false`
  - existing threads still anchored when `thread_replies: false`
  - DM replies still anchored when `thread_replies: false`

## Verification

- `just fmt-check` and `just clippy` pass.
- `just test-unit` passes (including new `buzz-acp` tests).

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -474,6 +474,15 @@ pub struct CliArgs {
     #[arg(long, env = "BUZZ_ACP_RELAY_OBSERVER", default_value_t = false)]
     pub relay_observer: bool,
 
+    /// Whether the agent should default to replying in the current thread.
+    ///
+    /// Mirrors the `thread_replies` behavioral config from persona packs.
+    /// When false, the harness does NOT append `--reply-to` instructions for
+    /// new top-level channel mentions, letting the agent post flat at channel root.
+    /// Existing thread and DM reply anchors are unaffected.
+    #[arg(long, env = "BUZZ_ACP_THREAD_REPLIES", default_value_t = true)]
+    pub thread_replies: bool,
+
     /// Exit after this many seconds with no dispatched events and no turn in flight.
     /// 0 disables inactivity self-termination.
     #[arg(long, env = "BUZZ_ACP_EXIT_AFTER_INACTIVITY", default_value_t = 0)]
@@ -544,6 +553,11 @@ pub struct Config {
     pub respond_to_allowlist: HashSet<String>,
     /// Allowed `respond_to` modes. Empty = all modes allowed.
     pub allowed_respond_to: Vec<String>,
+    /// Whether the agent should default to replying in the current thread.
+    ///
+    /// Mirrors the `thread_replies` behavioral config from persona packs.
+    /// Defaults to `true` per `PERSONA_PACK_SPEC.md`.
+    pub thread_replies: bool,
     /// Per-persona env vars to inject at agent spawn time (e.g., GOOSE_PROVIDER, GOOSE_MODEL, BUZZ_AGENT_MODEL).
     /// Populated from persona pack resolution. Empty when no pack is configured.
     pub persona_env_vars: Vec<(String, String)>,
@@ -1102,6 +1116,7 @@ impl Config {
             respond_to: args.respond_to,
             respond_to_allowlist,
             allowed_respond_to,
+            thread_replies: args.thread_replies,
             persona_env_vars,
             has_generated_codex_config,
             relay_observer: args.relay_observer,
@@ -1473,6 +1488,7 @@ mod tests {
             respond_to: RespondTo::Anyone,
             respond_to_allowlist: HashSet::new(),
             allowed_respond_to: Vec::new(),
+            thread_replies: true,
             persona_env_vars: vec![],
             has_generated_codex_config: false,
             relay_observer: false,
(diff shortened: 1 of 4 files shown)


santhiprakash:
Closing this for now — not because the underlying issue is invalid, but because the branch now conflicts with recent ACP rework that landed in the same files, and I can't give the rebase the verification it deserves from here.

**What conflicted**

- #5323 (Revert "fix(acp): reject unattended permission requests") and #5423 (Reduce repeated ACP session context) both reworked `crates/buzz-acp/{config,lib,pool,queue}.rs` — the same files this PR touches.
- The conflicts are mechanical/additive: the new `StandingContext` struct + `standing_context_sent` field sit right alongside this PR's `thread_replies` field, and the two features are orthogonal. But the behavioral suppression logic (skipping the forced `--reply-to` anchor in `resolve_reply_anchor` for top-level channel turns) needs to be re-validated against the new standing-context dispatch model from #5423, which I can't `cargo test -p buzz-acp` from this environment.

**The underlying bug is still real**

#4699 is still open and still reproducible on `main`: `append_new_thread_reply_instruction()` still unconditionally appends the `IMPORTANT:` `--reply-to <event_id>` directive (queue.rs ~L1154-1178), and `resolve_reply_anchor` is still called unconditionally for non-DM turns (~L1529). So a persona with `thread_replies: false` still opens a thread for every top-level channel message.

**Next step**

I'll re-file this as a fresh focused PR against current `main` — re-applied on top of the `StandingContext` work, with the behavioral logic re-checked and the unit tests updated — from a machine that can run the `buzz-acp` test suite. Happy to reopen here if a maintainer would rather pick it up; the `queue.rs` tests in this branch describe the intended behavior.

Issue #4745, opened 2026-08-04
buzz-acp: expose channel and sender identity via ACP `_meta` on `session/prompt`


ranaroussi:
### Problem

`buzz-acp` renders rich event context — channel id, sender pubkeys (hex + npub), event ids, timestamps, thread structure — into the prompt *text* via `format_event_block`. That's great for the model, but any ACP agent that needs this data *programmatically* (multi-tenant agents keying memory or authorization on the sender; bridges routing to per-user backends) has to parse it back out of the prompt string.

That parsing is fragile by construction: the block format is an internal rendering detail, and any wording change silently breaks every agent that scrapes it. It also can't distinguish "the agent was told who sent this" from "the transport asserted who sent this" — the text is inside the model-visible prompt, so a message *quoting* an event block is indistinguishable from the real thing.

### Proposal

Attach the same facts, structured, to the `session/prompt` request's `_meta` field — which the ACP spec defines on every protocol type for exactly this purpose, and which buzz-acp already uses in several places (`_meta.systemPrompt`, `_meta.sessionTitle`, `_meta.goose.activeRunId`, `_meta.steering.supported`).

Sketch (names open to bikeshedding; nesting under a `buzz` key to keep the root clean, matching the existing `_meta.goose.*` pattern):

```jsonc
{
  "method": "session/prompt",
  "params": {
    "sessionId": "…",
    "prompt": [ /* unchanged rendered text blocks */ ],
    "_meta": {
      "buzz": {
        "channelId": "7f3a2b10-…",
        "channelType": "channel",        // or "dm"
        "events": [
          {
            "id": "<event id hex>",
            "sender": "<pubkey hex>",
            "kind": 9,
            "createdAt": 1785840000
          }
        ],
        "cancelledEvents": [ /* same shape, for Steer/Interrupt merges */ ]
      }
    }
  }
}
```

The prompt text stays exactly as it is — this is additive metadata, not a rendering change. Agents that ignore `_meta` see no difference.

### Why upstream (rather than every agent parsing)

- The data already exists at the call site (`FlushBatch` has `channel_id` and each `BatchEvent` carries the full event), so the change is small and local to where `session/prompt` is built.
- Every ACP agent behind Buzz benefits — memory partitioning, per-sender policy, dedup by event id, thread-aware routing — without each one reinventing a scraper against an unstable format.
- It gives agents a *transport-asserted* identity channel, distinct from model-visible text, which matters for anything security-adjacent.

[...]

PR #4751, opened 2026-08-04
feat(acp): expose channel and sender identity via _meta.buzz on session/prompt

ranaroussi:
Implements #4745: attaches the channel and sender identity that `format_event_block` already renders into the prompt *text* as structured ACP `_meta` on `session/prompt`, so agents can key memory, authorization, or routing on it without parsing the rendered prompt.

```jsonc
"_meta": {
  "buzz": {
    "channelId": "…",
    "channelType": "stream",          // omitted when channel metadata is unresolved
    "events": [ { "id", "sender", "kind", "createdAt" } ],
    "cancelledEvents": [ … ]           // present only for Steer/Interrupt merge batches
  }
}
```

- **Additive only** — prompt text unchanged; agents that ignore `_meta` see no difference. Follows the existing `_meta` merging precedent from `session/new` (`sessionTitle`/`systemPrompt`).
- Built in `queue.rs` beside `format_event_block` (whose reads it mirrors), threaded through `session_prompt_blocks_with_idle_timeout` → `build_prompt_params`, populated at the batch prompt site in `pool.rs`. Heartbeat/initial-message paths pass nothing. The goose-native steer path is untouched.
- **One deviation from the issue sketch:** `channelType` passes through the codebase's actual value set (`dm`/`private`/`stream`/`unknown` per `channel_type_from_tags`) rather than the issue's simplified `channel|dm`, and the key is omitted when channel metadata is unresolved — per the relay-side warning that consumers must not assume a type for unknown channels. Happy to normalize instead if you'd prefer.
- Tests: 4 new (meta shape incl. cancelled events and omission rules; params nesting; None ⇒ no `_meta` key), `cargo test -p buzz-acp` 672 passed, fmt + clippy clean per your CI.

Consumer context: [muxi-ai/acp-bridge](https://github.com/muxi-ai/acp-bridge) currently parses the rendered event blocks for this data and will switch to `_meta.buzz` as soon as it exists.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -749,6 +749,7 @@ impl AcpClient {
         self.session_prompt_blocks_with_idle_timeout(
             session_id,
             std::slice::from_ref(&prompt_text),
+            None,
             idle_timeout,
             max_duration,
         )
@@ -756,7 +757,8 @@ impl AcpClient {
     }
 
     /// Like [`session_prompt_with_idle_timeout`](Self::session_prompt_with_idle_timeout),
-    /// but sends each entry in `prompt_blocks` as a separate text content block.
+    /// but sends each entry in `prompt_blocks` as a separate text content block,
+    /// optionally attaching structured batch metadata as `_meta.buzz`.
     ///
     /// Used for slash-command pass-through: ACP connectors detect commands via
     /// the **first** block's text starting with `/`, so the harness sends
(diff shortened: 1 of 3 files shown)


salman1993 (changes requested):
🤖 **Medium — `crates/buzz-acp/src/pool.rs:1870`: the production metadata wiring has no regression test.**

The new tests exercise `build_prompt_meta` and `build_prompt_params` independently, but none drives a `FlushBatch` through `run_prompt_task` and captures the resulting `session/prompt`. If this assignment is removed, or either call below keeps passing `None`, all 681 package tests still exercise only the disconnected helpers and the feature can disappear while staying green. That leaves the PR's only user-visible contract (`params._meta.buzz` on batch prompts) unprotected.

Please add a production-seam test that dispatches a batch through this path and asserts the captured JSON-RPC request contains the expected `_meta.buzz` values (and ideally that a heartbeat/batch-less prompt omits `_meta`).

Separately, the PR currently conflicts with `main`, so GitHub cannot run the normal merge-ref CI lanes until it is rebased.

PR #4754, opened 2026-08-04
feat(media): allow text/html on the generic file-upload path

TechPrieto:
## Summary

Allow `text/html` through the generic file-upload path's deny-list (`buzz-media/src/validation.rs::BLOCKED_FILE_MIME_TYPES`). Depends on #4753 (mirrors the same removal in `buzz-cli`'s client-side `BLOCKED_MIMES`).

**Why:** legitimate use case is sharing generated HTML reports/exports as chat attachments. The existing defence in depth already neutralizes the stored-XSS risk for any client that respects the response headers: generic (non-image/video) attachments are served with `Content-Disposition: attachment`, `X-Content-Type-Options: nosniff`, and `Content-Security-Policy: default-src 'none'` (`buzz-relay/src/api/media.rs`). This PR only removes the extra defence-in-depth layer for `text/html` specifically — `application/xhtml+xml`, JS, and SVG stay blocked, and executables stay blocked (zip them instead, already supported).

This is a deliberate, scoped risk trade-off made by a self-hosted instance owner for their own deployment — happy to discuss whether it belongs upstream as-is, behind a config flag, or not at all if the maintainers see a broader risk I'm not accounting for.

## Test plan

- [x] `cargo test -p buzz-media --lib` — 108/108 passing, including new coverage: HTML is accepted and `serve_inline` still returns false for it (forced download), and a real XHTML-prologue document doesn't get misclassified as `text/html`
- [x] `cargo test -p buzz-cli` — 321/321 passing

### DIFF
--- a/crates/buzz-cli/src/client.rs
+++ b/crates/buzz-cli/src/client.rs
@@ -60,7 +60,8 @@ pub fn build_imeta_tag(d: &BlobDescriptor) -> Vec<String> {
     tag
 }
 
-/// MIME types accepted for upload.
+/// MIME types recognized as image/video for the size-tier and imeta decision.
+/// Not a security allowlist — see `BLOCKED_MIMES` below.
 const ALLOWED_MIMES: &[&str] = &[
     "image/jpeg",
     "image/png",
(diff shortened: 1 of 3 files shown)


TechPrieto:
Closing — this was based on a tenant-owner decision that got reversed after a scope misunderstanding (the actual need was agent-to-owner file delivery, not accepting HTML uploads generally). Not something we want in the deny-list exception list. Sorry for the noise.

Issue #4469, opened 2026-08-03
Native git repository is unavailable after a volume-level isolated restore


C0dino:
## Summary

A channel-bound NIP-34 repository accepts a signed push and clone in production, but the same authenticated clone returns `repository not found` after an isolated restore even though the repository announcement and object-store pointer are present.

## Reproduction

1. Create a channel-bound NIP-34 repository and push one empty commit with `git-credential-nostr`.
2. Verify two authorized members clone the same SHA and a non-member receives HTTP 403.
3. Quiesce the write path and archive Postgres, Redis, MinIO, and git scratch volumes with checksums.
4. Restore all volumes into an isolated stack using the same community host mapping and storage credentials.
5. Verify relay liveness, community/member counts, the kind 30617 announcement, and the MinIO repository pointer.
6. Attempt an authenticated clone against the isolated relay using a host-bound NIP-98 token.

## Expected

The restored relay advertises and clones the same repository SHA.

## Actual

The clone returns `repository not found`. The restored database contains the kind 30617 announcement, the restored MinIO volume contains the community/owner/repository pointer, and the relay git object-store conformance probe reports success.

## Environment

- buzz: `desktop-v0.5.3 server source`
- database: `Postgres 17`
- cache: `Redis 7`
- objectStore: `MinIO path-style S3`
- gitClient: `Git 2.46.3 with git-credential-nostr`

## Impact

A generic full-stack restore can report healthy while native repository recovery is still incomplete, so self-hosters need a repository-level readback gate before treating a backup as recoverable.

## Workaround

Keep critical repositories on an external forge and add an authenticated clone/readback to restore drills.

## Disclosure check

- Client data: no
- Credentials or private hosts: no
- Reproduction uses a noncritical empty-commit repository

PR #4761, opened 2026-08-04
docs(deploy): clarify Git backup and recovery

SmithLabsLLC:
## Summary

- correct the production Compose and Helm backup guidance
- identify PostgreSQL and the complete object store as durable recovery data
- identify Redis state and `buzz-git-data` as disposable runtime data
- require an isolated authenticated clone plus ref, commit, and tree comparison
  before accepting a restore

## Related issue

Related to #4469. This documentation change does not fix the underlying restore
failure reported there. It reduces operator risk by documenting the actual
storage boundary and a readback check that can detect an incomplete restore.

## Storage and recovery model

Git packs, manifests, and manifest pointers live in the configured object
store. The local Git volume is scratch space and a process-local pack cache; it
cannot replace an object-store backup. Redis contains disposable coordination
and presence state and can be recreated empty.

Operators should back up PostgreSQL and the complete object store from the same
maintenance window, preferably while the stack is stopped. Where only ordered
live snapshots are possible, PostgreSQL should be captured before the object
store.

After restoration, an isolated relay must pass an authenticated clone and the
restored refs, commits, and trees must match the pre-backup record. A healthy
container or restored Git scratch volume is not sufficient proof.

## Testing

- `just ci`: every lint, build, and test group passed except two unrelated
  desktop-native tests that hit a transient Linux `Text file busy` error
- reran the complete five-test provider-deploy group serially: 5 passed
- `bash -n deploy/compose/run.sh`
- `shellcheck deploy/compose/run.sh`
- `./deploy/compose/run.sh backup-hint`
- Helm 3.16.4 lint: passed
- helm-unittest 0.8.2: 44 tests passed
- rendered every chart values fixture successfully
- `git diff --check`

## Scope

Operator documentation and backup-hint text only. No storage behavior changes.

### DIFF
--- a/deploy/charts/buzz/README.md
+++ b/deploy/charts/buzz/README.md
@@ -213,9 +213,16 @@ Save these. Losing any of them is data loss. See NOTES.txt printed by `helm inst
 
 1. `BUZZ_RELAY_PRIVATE_KEY` — relay identity. Rotating it = new identity (federation peers will not recognize the relay).
 2. PostgreSQL database — the canonical event store.
-3. S3 bucket — media blobs (chart default bucket: `buzz-media`).
-4. Git PVC — repo on-disk state served by the relay's git endpoint.
+3. Complete S3 bucket — the durable source of truth for media plus Git packs, manifests, and manifest pointers (chart default bucket: `buzz-media`).
+4. Git PVC — optional scratch data and a process-local pack cache. Its cache contents are not reused after a relay restart, and it is not a repository backup.
 5. Owner private key — held by the operator, not by this chart. Restore by re-installing with the same `ownerPubkey`.
+
+Stop write traffic while taking the PostgreSQL and object-store snapshots when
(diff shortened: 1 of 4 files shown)


SmithLabsLLC:
@tlongwell-block, could you confirm whether this backup and recovery guidance
matches the intended object-store model?

I refreshed the branch to current `main`, linked #4469, and validated the
Compose backup hint plus the Helm chart. The key boundary is now explicit:
PostgreSQL and the complete object store are durable; Redis and Git scratch may
restart empty; and a restore is accepted only after an authenticated clone and
ref, commit, and tree comparison.

The PR is documentation and backup-hint text only. If another maintainer owns
this area, a redirect would be appreciated. GitHub is also holding the new CI,
Helm, and desktop workflow runs at `action_required` with no jobs started, so a
maintainer workflow approval appears to be needed before the checks can run.

PR #4762, opened 2026-08-04
fix(desktop): serialize tray channel actions for frontend

kalvinnchau:
## Summary
- serialize native `openChannel` tray actions with the camelCase field names consumed by the TypeScript frontend
- prevent a valid tray channel ID from becoming `/channels/undefined`
- add a Rust serialization contract test covering the complete frontend payload shape

### Root cause
`TrayAction` renamed the enum variant to `openChannel`, but its struct fields still serialized as `channel_id` and `community_generation`. The frontend reads `action.channelId`, so tray navigation called `goChannel(undefined)`.

### Testing
- manually verified the corrected runtime payload and tray navigation before removing temporary logging
- `just desktop-ci`
- pre-push hooks (desktop checks/tests, Tauri checks, and Rust tests)


wpfleger96 (approved):
Reviewed the aggregate one-file diff and the Rust/TypeScript tray-action contract at this exact head. The explicit serde field renames match the frontend consumer, and the full-payload regression test pins both the discriminant and field names. No blocking findings.

PR #4768, opened 2026-08-04
feat(desktop): relay admin console for the /api/admin/v1 operator surface

wpfleger96:
Adds a relay admin console under Settings — the single surface for relay-level operator work — replacing the previously unreachable moderation section. Operators authenticate with their own Nostr key over NIP-98 — no browser extension, no bearer token — to view deployment-wide moderation reports and product feedback, resolve/dismiss/escalate/reopen reports, update feedback status, and manage the operator/moderator staffing roster.

## Nav reachability

The old `moderation` section id was defined in `settingsSections` but never wired into any group in `settingsNavGroups`, so the sidebar never rendered it — unreachable since #1617. This PR replaces it with a `relay-admin` section wired into the Communities nav group, pointing at the relay admin console, and drops the separate `admin-console` section id so a single surface owns relay-level trust & safety.

The Relay admin nav entry is always visible. Auth gates the panel itself; the nav entry is a door, not a credential — hiding it behind discovery state would strand an operator whose relay transiently fails NIP-11 discovery with no path back to the connection settings.

## Discovery and connection

On mount the console reads the connected relay's NIP-11 document for the optional `admin_api` field. When a valid origin is discovered **and its host matches the connected relay's host** (case-insensitive, host identity only), it is automatically saved and probed — the panel renders immediately with no Save step required. A cross-host advertisement is pre-fill only: it populates the manual origin field under Advanced and is never saved or probed without an explicit Save, so a relay cannot direct an unconsented NIP-98 signature at a third-party origin; residual exposure on the same-host path is bounded by NIP-98's URL/method/payload binding. Saving is only required for manual origin changes. The advertised origin is untrusted input: revalidated through `AdminOrigin::parse`, falling back to manual entry when absent or invalid. A saved manual origin always takes precedence over discovery.

## Auth and transport (Rust, `src-tauri/src/commands/admin/`)

- `AdminOrigin` value object: validates scheme + host + optional port, rejects credentials/path/query/fragment; `http://` only for loopback. The relay side normalizes its configured admin host at config load and compares inbound `Host`/`Origin` hosts case-insensitively.
- `AdminRoute` closed enum: no IPC surface accepts arbitrary URLs, so the NIP-98-signed URL is byte-identical to the fetched URL.
- Dedicated no-redirect `reqwest` client as an SSRF guard: a relay `3xx` is an error and the NIP-98 header never crosses origins.
- `admin_probe` signs `GET /probe` and returns a typed state (`nip98Authorized` / `nip98Denied` / `tokenMode` / `disabled` / `notAdminApi` / `networkOrIntercepted`). The response body is validated against the full relay contract — `nip98Authorized` only when the complete NIP-98 invariant holds, `disabled` only on a coherent disabled-mode body; any incoherent or unrelated JSON classifies as `notAdminApi`, so the console never fails open at the network boundary. Unknown fields are tolerated for forward compatibility.
- NIP-98 signing via `AppState::signing_keys()`, one retry on `401` with a fresh signed event; response size bounded by a `Content-Length` preflight and a streaming byte counter; per-pubkey origin storage with atomic `0o600` writes.

## Relay admin console

The console is gated behind the relay `OPERATOR`/`MODERATOR` roles from #3777; every `/api/admin/v1/*` route returns `403` to anyone else. Community self-administration (`community-members` / Invites) and in-channel enforcement (the 9040–9043 signed commands) are a separate axis and are untouched. The previously-duplicate community report queue (`ModerationQueueCard`, `moderationQueue.ts`, and three dead hooks) was unreachable dead code and is deleted in its own commit; `useBanMemberMutation` and the rest of `features/moderation/hooks.ts` remain in use by the members sidebar and message menu.

## Console UI (TypeScript, `src/features/admin-console/`)

- `AdminConsoleSettingsCard`: discovery + probe flow, honest copy for every probe state, manual origin under an `Advanced` disclosure at the bottom. The role (operator/moderator) is rendered in the connection status line as plain text ("Connected as operator") — no badge-shaped elements above the tab row. The origin provenance (relay config / database) appears inside the Advanced card as small muted text. Section title changed to `Admin`.
- `AdminConsolePanel` with Reports / Feedback / Staffing tabs (Staffing gated on `role === "operator"`); reports and feedback grouped by community for cross-community triage.
- **Reports**: always requests `scope=all` so the existing resolve/cancel/reopen controls can reach every status (`open`, `processing`, `resolved`, `dismissed`, `escalated`). The relay's omitted-scope default is `escalated`-only (the platform-safety backstop); `scope=all` is explicit and scoped to this console. Full action matrix per `target_kind` (event → delete/kick/ban/timeout/dismiss/escalate; pubkey → ban/timeout/dismiss/escalate; blob → dismiss/escalate); kick is suppressed when the report carries no `channelId`. A `processing` row stays navigable — enforcement state lives in the detail view. A `failed` action (always pre-mutation) offers a single **Cancel & reopen** via `POST /reports/{id}/cancel` with `{actionId}` fencing; there is no client-side retry. `pending`/`enforcing` actions belong to the relay recovery worker and offer no button; a `409` reloads detail. Lists refetch on back-navigation after a mutation.
- **Report failure UX**: `reportErrorMessage()` strips 4xx/5xx status prefixes and surfaces the relay's own error reason in the toast (e.g. relay returns `"400: you cannot report this content"` → toast shows `"you cannot report this content"`). Self-report blocking UI removed — the relay has no self-report gate, and the failure was purely client-side error surfacing.
- **Reason audience disclosure**: the Reason (optional) field shows exact copy under the input based on selected action — Delete discloses that reason is sent verbatim to the affected user and posted publicly in the room; Kick/Ban/Timeout disclose affected-user only; Dismiss/Escalate disclose reporter only. These mirror the actual relay notice paths.
- **Enforcement history**: the report detail carries the governing action even after the report leaves `processing` — a report enforced then reopened is `open` yet still shows its `succeeded` action as executed history alongside the resolve form; a reopen does not un-happen a ban. The resolve form gates on `open` status alone.
- **Reopen**: terminal reports (`resolved`/`dismissed`/`escalated`) can be returned to `open` for re-triage via `POST /reports/{id}/reopen`. Reopen never reverses applied enforcement, and the copy says so.
- **Feedback**: list, detail, and status control (`new`/`reviewed`/`archived`) with a generation-fenced attachment viewer. A purged source community severs provenance to `null` rather than deleting the row; severed rows bucket under a "source community removed" heading.
- **Staffing**: Add is create-only with display names resolved via `useUsersBatchQuery` and npub cross-fade on hover (`HoverStaffingIdentity`). An entered key that already appears in the authoritative roster (config, owner-fallback, or DB) is rejected locally with an inline duplicate-key message and no request is issued; a config-backed principal targeted by add, role change, or remove returns the relay's `409 Conflict`, classified via the typed `relayStatus` (never by matching message text) and rendered with config-backed copy. Roster with `config`/`owner_fallback`/`db` source labels; config-backed entries disable remove client-side. In-place role change via `<select>` calling `putAdminOperator`; the relay responds with the effective `OperatorEntry` (`pubkey`, `effectiveRole`, `sources`) in the same shape as the roster list. Read-only badge for config-managed entries.
- **Timeout expiry**: `useTimeoutState` clears the store when derived state is `INACTIVE` but the store is still active; the 1s tick is scoped to entries with a known `expiresAtMs` and owned by `ComposerTimeoutBanner` (mounted only while a timeout is active), while `ChannelPane` subscribes to the boolean `useTimeoutActive()` — the pane re-renders when a timeout starts or ends, not every second. `useMembersSidebarModeration` ticks `nowMs` only while the sidebar is open, the viewer can moderate, and some member has a future `mutedUntil`; the tick after the last expiry stops the interval.
- Mutations confirm via `sonner` toasts; failures surface the relay's parsed error `message`, not raw JSON. `tokenMode` and `disabled` auth modes render read-only.

## Mutation idempotency

Resolve and reopen carry a per-attempt `requestId`. On first submit the whole command (`{requestId, action, reason, expirationSecs}`) is frozen; after an ambiguous failure (409, 5xx, transport error, truncated response) the snapshot is retained, action/reason/duration controls are locked, and the retry sends the exact same payload byte-for-byte. A definitive pre-commit rejection (non-409 4xx with `bodyComplete: true`) discards the snapshot, unlocks controls, and the corrected submit uses a fresh payload. On success the toast derives from the authoritative `AdminReportResolution` (`activeAction.action` when present; otherwise `status` for decision-only outcomes) — never from the mutable form selection.

The native mutation commands — including operator delete — reject with a typed `AdminMutationError` carrying the relay's HTTP status (`relayStatus`, `null` when no verdict exists) and a `bodyComplete` flag that is `true` only when the full response body was read.

Related: #3777 (relay `OPERATOR`/`MODERATOR` role model + NIP-98 auth + `admin_api` NIP-11 advertisement — provides the runtime and the discovery field this console consumes)
Related: [bb-public#339](https://github.com/squareup/bb-public/pull/339) (Phase 4 rollout config)

## Screenshots

Captured headless at `197f0ba0c` against the mock bridge (1280×720).

### Setup

**Origin setup — Advanced disclosure open, idle state**
Relay admin panel with Advanced: admin origin disclosure open before probe

**Authorized console — all three tabs**
Authorized relay console showing Connected as operator status and Reports/Feedback/Staffing tabs

### Reports

**Reports queue — open, processing, resolved across two relay groups**
Reports tab showing community-grouped list with spam/open, harassment/processing, illegal/resolved rows

**Open report detail — reporter, target, message snapshot, resolve form**
Open spam report detail with reported message content and Delete/Kick/Ban/Timeout/Dismiss/Escalate action buttons

**Processing report — PROCESSING/HARASSMENT, ban enforcing state**
Processing harassment report detail showing enforcement state with ban action in enforcing status

**Resolve form — Delete selected, audience disclosure copy**
Resolve form with Delete action selected showing "Sent verbatim to the affected user and posted publicly in the room" audience copy

**Resolved report — enforcement history block**
Resolved/illegal report detail showing Resolved by, Resolved at, and Act ID fields from completed enforcement

### Feedback

**Feedback list — community grouping with status badges**
Feedback tab showing relay.example.com group with new/reviewed items and source-community-removed group

**Feedback detail — mutable status control (operator mode)**
Feedback detail with new/reviewed/archived status tab control visible for mutation

**Feedback detail — passive status badge (read-only/disabled mode)**
Feedback detail in disabled mode showing passive "Status NEW" badge with no mutation control

### Staffing

**Staffing roster — config/owner\_fallback/db source labels**
Staffing tab showing three operators with config, owner_fallback, and db source labels

**Add form — new pubkey filled, ready to submit**
Staffing add form with a new 64-hex pubkey entered and moderator role selected

**Duplicate rejection — inline error for existing principal**
Staffing add form showing "Already an operator: operator. Use Remove to revoke before re-adding with a different role." error for an existing config-backed operator


wesbillman (changes requested):

**Changes requested**

- **[P1] Do not auto-probe an unrestricted relay-advertised origin** — `desktop/src-tauri/src/commands/admin/mod.rs:938-949`

  `admin_api` is explicitly untrusted NIP-11 data, but discovery applies the same permissive policy as manual operator input: HTTP loopback is accepted and every HTTPS host is accepted. Both the nav resolver and settings card then probe that value automatically without confirmation (`hooks.ts:45-56`, `AdminConsoleSettingsCard.tsx:303-316`). A malicious connected relay can therefore make the native client contact `http://localhost:<port>` (the client pins `localhost` to `127.0.0.1`) or HTTPS private/link-local targets. If the target returns `401 WWW-Authenticate: Nostr`, the app sends it a fresh app-key signature, adding a signing oracle and pubkey-ownership disclosure to the SSRF. Separate advertised-origin validation from manual configuration: do not auto-probe loopback/private/link-local advertised destinations, and account for DNS resolution/rebinding, or require explicit operator confirmation before any cross-origin probe. Keep the loopback HTTP carve-out only for manual origins.

- **[P2] Actually make disabled-auth mode read-only** — `desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx:397-480`

  `disabled` currently renders the same `AdminConsolePanel` as an authorized principal, with only the staffing tab hidden. The report resolve/reopen/cancel controls and feedback status controls remain enabled, although the related relay contract intentionally rejects every mutation in disabled mode with 403. This contradicts the PR’s stated read-only behavior and presents controls guaranteed to fail. Pass a read-only capability into the panel and suppress/disable all mutation affordances when the probe result is `disabled`.

The command registration/API seams, identity and async fencing, report lifecycle/idempotency paths, attachment cleanup, navigation, staffing gating, and legacy-surface removal were otherwise coherent in this read-only diff review. GitHub currently exposes only a passing DCO check for this head; PR code was not executed locally under the automation trust policy.

wesbillman (dismissed):

Two security/resource-boundary blockers remain on this head:

1. **Do not auto-probe a relay-advertised admin origin under the manual-origin trust policy.** The NIP-11 `admin_api` value is explicitly untrusted, but `admin_origin_from_nip11` passes it through the same `AdminOrigin::parse` policy used for operator-entered values. That policy accepts loopback HTTP and unrestricted HTTPS hosts. Both the navigation resolver and the settings card then automatically probe the advertised origin, and the probe performs a native GET before sending a fresh NIP-98 signature when the target replies with `401 WWW-Authenticate: Nostr`. A malicious connected relay can therefore advertise `http://127.0.0.1:<port>` / `localhost` (the client explicitly pins `localhost` to `127.0.0.1`) or an HTTPS private/link-local target and trigger zero-click native SSRF; a cooperating target also gets an attacker-triggered signing oracle and proof of the app pubkey. This is not hypothetical policy drift: `mod_tests.rs` currently asserts that a loopback origin from NIP-11 is accepted and returned for automatic probing. Please separate advertised-origin validation from explicit manual configuration: never auto-probe loopback/private/link-local advertised destinations, and account for DNS rebinding, or require an explicit operator confirmation before any cross-origin probe. Keep loopback HTTP only for a manually entered origin.

2. **Bound attachment preview work across the whole feedback item, not only per response.** `parseImetaAttachments` accepts and returns every valid `imeta` tag, `FeedbackDetail` mounts an `AttachmentViewer` for every result, and each `image/*` viewer immediately starts its own native fetch. The Rust command caps each response at 10 MiB, but there is no attachment-count, aggregate-byte, or concurrency bound. The relay’s feedback ingest bounds serialized tags to 64 KiB but does not cap the number of `imeta` entries, so an admitted feedback event can fan out many simultaneous image reads and retain their blob URLs until navigation. The existing test explicitly blesses multiple tags but covers no aggregate/concurrency ceiling. Please impose a small count plus aggregate-byte budget before rendering/fetching, avoid unbounded parallel auto-load, and add a regression showing excess attachments are not requested.

The closed route construction, exact-body NIP-98 binding, no-redirect client, mutation idempotency/recovery fences, and identity/origin React state fences otherwise held in this read-only diff review. GitHub currently exposes only the passing DCO check for this head; I did not execute PR code under the automation policy.

wesbillman (changes requested):

**Changes requested**

- **[P1] Do not automatically probe an unrestricted relay-advertised origin** — `desktop/src-tauri/src/commands/admin/mod.rs:938-950`

  `admin_api` is untrusted NIP-11 data, but discovery runs it through the manual-entry policy, which accepts loopback HTTP and every HTTPS host (`origin.rs:64-90`). With no saved origin, both the settings surface and navigation resolver automatically probe the advertised value (`AdminConsoleSettingsCard.tsx:284-316`, `hooks.ts:41-56`). The native probe first GETs that attacker-selected destination and, if it answers `401 WWW-Authenticate: Nostr`, sends a fresh NIP-98 signature to it (`mod.rs:160-215`). A malicious or compromised connected relay can therefore target localhost, private/link-local addresses, or an attacker hostname that resolves there; no-redirect handling does not constrain this initial destination. Separate advertised-origin trust from explicit manual configuration: require an explicit confirmation or prove/bind the advertised endpoint to the connected relay, and reject resolved loopback/private/link-local/reserved targets with rebinding-safe connection handling. Add advertised-origin tests for loopback, RFC1918/link-local, and a public hostname resolving privately. The existing discovery test currently blesses automatic loopback probing (`mod_tests.rs:872-916`).

- **[P1] Bound attachment fan-out across the feedback item** — `desktop/src/features/admin-console/AdminConsolePanelHelpers.tsx:243-272`

  `parseImetaAttachments` returns every valid `imeta` tag, `FeedbackDetail` mounts every result (`AdminConsoleFeedbackTab.tsx:420-465`), and each `image/*` viewer immediately starts its own native fetch on mount (`AdminConsoleFeedbackTab.tsx:215-224`). The 10 MiB native cap is per response, with no attachment-count, aggregate-byte, or concurrency ceiling. A single admitted feedback event can therefore trigger many simultaneous native reads and retain all resulting blobs until navigation. Impose a small item-wide count/aggregate budget and bounded fetch concurrency (or explicit loading), then add a regression proving excess attachments are not requested.

- **[P2] Keep disabled-auth mode read-only as documented** — `desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx:470-480`

  A coherent `disabled` probe mounts the same panel as an authorized principal without passing a read-only capability. Report resolve/reopen/cancel and feedback status controls therefore remain live, although the relay rejects mutations in this mode. This contradicts the documented contract that disabled mode renders feedback status read-only (`docs/admin/README.md:267-276`) and the PR description. Pass the auth capability into the panel and hide or disable every mutation affordance while preserving reads.

- **[P2] Expose the selected feedback status programmatically** — `desktop/src/features/admin-console/AdminConsoleFeedbackTab.tsx:352-375`

  The current status is communicated only through button styling (`variant` and ring classes). The three ordinary buttons have no `aria-pressed`, radio semantics, or textual current-value output, so a screen-reader user cannot determine the active status before changing it. Add a programmatic selected state and assert that semantic contract rather than only the visual class.

The closed route construction, exact URL/body NIP-98 binding, no-redirect client, response caps, identity/origin async fences, report lifecycle and request-ID recovery contract, role-gated staffing, and legacy moderation-surface removal were otherwise coherent in this read-only exact-head review. Desktop, Rust, build, and desktop integration checks are green; PR code was not executed locally under the automation trust policy.

wesbillman (changes requested):

**Changes requested at exact head `63db828a09b270239d5e56bf82c3f5e8468cea43` and exact base `a3730784fc851bb1125b40cca9b0a30788a293c1`.**

### P1: require confirmation before removing an operator

The Staffing trash button calls `handleRemove` directly, which immediately invokes `deleteAdminOperator` (`desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx:105-121,203-224`). One stray click therefore deletes a DB-backed operator, potentially the acting operator, without a confirmation, target/role recap, or undo. That can remove administrative access and may require another operator or out-of-band configuration to recover.

Put this destructive authorization change behind an explicit confirmation dialog. Make self-removal especially clear, and add cancel/confirm behavior tests. The current event suite only checks whether staffing controls are rendered; it does not exercise removal confirmation.

### P2: recover when the active Staffing tab becomes unauthorized

`activeTab` persists independently of `role` (`AdminConsolePanel.tsx:832-843`). If the same mounted panel changes from operator to moderator/unknown while Staffing is selected, the tab button disappears and the content is suppressed (`AdminConsolePanel.tsx:859-887`), leaving an empty panel with no selected visible tab. Reset an invalid active tab to Reports when role/capabilities change and cover that prop transition.

The previous advertised-origin SSRF, attachment fan-out, disabled-mode mutation, and feedback-status accessibility blockers are otherwise closed at this head. Relay discovery is now prefill-only until an operator explicitly saves the origin; item-wide attachment limits, read-only mutation gating, and `aria-pressed` are present. Hosted substantive checks are green. This was a read-only review of immutable GitHub source/diffs; no PR code was checked out or executed.

desktop/src/features/admin-console/nav.ts
```diff
@@ -0,0 +1,31 @@
...
+ *   edits/clears the origin lives inside the surface, so hiding it would lock a
+ *   user out of fixing a bad saved URL.
+ * - An advertised origin shows the entry so the operator can open Moderation
+ *   and confirm the pre-filled origin. The advertised value is NOT probed to
+ *   decide visibility — probing an untrusted relay-advertised origin would send
+ *   a signed NIP-98 credential to an attacker-chosen destination. Nothing
+ *   contacts the advertised origin until the operator explicitly saves it.
+ */
+export function shouldShowModerationNav(res: ModerationNavResolution): boolean {
+  return res.originSource !== "none";
```
wesbillman: [P2] Keep manual origin setup reachable when discovery is unavailable

With no saved origin, an absent/invalid `admin_api` advertisement or a discovery error resolves to `originSource: "none"` (`hooks.ts:42–52`). This gate then hides Moderation, and `SettingsView.tsx:179–184` redirects a direct Moderation section selection once resolution completes. The manual origin input and Save control are inside that same hidden surface (`AdminConsoleSettingsCard.tsx:423–455`), so a first-time operator cannot use the advertised manual fallback to connect to a known admin host. Keep an independently reachable setup route/control for this case. Preserve the explicit-confirmation boundary: do not solve visibility by automatically sending signed probes to an untrusted advertised origin. Cover absent advertisement and discovery failure through actual Settings navigation, not just the isolated card.

desktop/src/features/admin-console/AdminConsolePanel.tsx
```diff
@@ -0,0 +1,908 @@
...
+  const [selectedId, setSelectedId] = useState<string | null>(null);
+  // List refresh fence: bumped whenever a mutation completes in the detail
+  // view, so returning to the list shows fresh status without a tab switch.
+  const [listGen, setListGen] = useState(0);
+
+  const listState = useAsyncLoad(
+    () => listAdminReports(origin),
+    [origin, pubkey],
+    generation + listGen,
+  );
```
wesbillman: [P2] Provide access to non-escalated reports through the list query

`listAdminReports(origin)` sends an empty query, but the exact-head relay intentionally defaults it to `status=escalated` unless `scope=all` is supplied (`crates/buzz-relay/src/api/admin/mod.rs:206–224`). This tab has no status/scope selector or alternative report-ID entry, and detail is only reachable by selecting a returned row. Consequently open reports cannot reach Resolve, processing reports cannot reach failed-action cancellation, and resolved/dismissed reports cannot reach Reopen from a fresh list. Reopening a listed escalation exposes its open detail temporarily, but returning to the list loses it. Add an explicit way to access these states while retaining escalated triage. `scope` also needs native plumbing: `AdminReportsQuery` and `AdminQuery` currently omit it, and the relay's status validator does not accept `processing`. Test the UI against the real default/filter contract rather than mocked all-status list results.

desktop/src/features/admin-console/AdminConsolePanel.tsx
```diff
@@ -0,0 +1,908 @@
...
+      )}
+
+      <div>
+        <input
+          className="w-full rounded-md border border-border/60 bg-background px-2 py-1 text-xs"
+          data-testid="resolve-reason-input"
+          onChange={(e) => setReason(e.target.value)}
+          placeholder="Reason (optional)"
+          type="text"
+          value={reason}
```
wesbillman: [P2] Disclose that enforcement reasons are sent outside the private report

This input is labeled only “Reason (optional)” beside private reporter/note context. It is passed directly to `ResolveReportBody.reason`, whose exact-head contract explicitly says it is public, broadcast verbatim as the removal tombstone's reason and sent verbatim to the affected user, with no sanitization/redaction (`crates/buzz-relay/src/api/admin/mod.rs:421–426`; notice rendering in `handlers/moderation_notices.rs:256–279`). An operator entering an internal rationale or quoting report context here can therefore expose it to the reported user and, for removal, the room. Clearly label the public audience before submission and warn against including reporter identities or private report details. Keep any internal-note affordance separate from this field.

desktop/src/features/admin-console/AdminConsolePanel.tsx
```diff
@@ -0,0 +1,908 @@
...
+      await resolveAdminReport(origin, report.id, {
+        action: selectedAction,
+        requestId: requestIdRef.current,
+        expirationSecs:
+          selectedAction === "timeout" && expirationSecs
+            ? Number(expirationSecs)
+            : undefined,
+        reason: reason.trim() || undefined,
+      });
+      toast.success(`Report resolved: ${actionLabel(selectedAction)}`);
```
wesbillman: [P2] Bind an ambiguous retry to its original command, not editable form state

After a transport/5xx failure, `requestIdRef` is preserved but action, reason and duration remain editable. For example, the first Ban request can persist its claim before the response is lost; choosing Timeout and resubmitting then sends the same request ID with a different action. The relay correctly uses the first persisted claim (`handlers/report_resolution.rs:305–345`; the existing `same_request_id_retry_with_changed_body_uses_persisted_claim` test documents this), so it executes/confirms Ban while this code ignores the response and announces “Report resolved: Timeout.” This misrepresents a permanent restriction as the newly selected one. Retain the submitted payload with the request ID and prevent a changed command from masquerading as its retry; derive the outcome from the authoritative response. Do not simply generate a new ID on an ambiguous failure. Add a lost-response → edit → retry regression.

wesbillman (changes requested):

**Changes requested: four P2 desktop integration defects.**

Reviewed head `662db64518d72ba96a3e1726765baae71d203644` against exact base `a3730784fc851bb1125b40cca9b0a30788a293c1`.

The product boundary is a relay-operator/moderator console, distinct from community-owner administration. Discovery must remain untrusted until the operator confirms the origin; the relay remains authoritative for roles, report transitions and public enforcement notices. The four inline findings concern unreachable manual setup, non-escalated report access, undisclosed public-reason semantics, and retries that claim the wrong action succeeded. They do not require changing the relay's intentional escalated-only default or weakening discovery trust.

Coverage: Settings navigation/discovery and identity/origin transitions; native origin, route/auth, bounded-body/error plumbing; report list/detail/resolve/reopen/cancel and idempotency; feedback attachments/status; staffing confirmation/role changes. Compared client routes and closed action/status/role sets to exact-head relay producers. The previous staffing-confirmation, hidden-active-tab and unconditional request-ID-clearing defects are fixed in this head and are not repeated. Community signed-command enforcement and mobile/web UI are not changed by this desktop PR; they were not independently reviewed as whole features.

Validation was **source/diff and existing-test inspection only** under the read-only automation policy. Key source blobs were checked against the pinned Git tree. No checkout, build, test execution or live relay/app run; this is not a runtime pass.

desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx
```diff
@@ -0,0 +1,299 @@
...
+    generation + listGen,
+  );
+
+  const handleAdd = async () => {
+    const trimmed = addPubkey.trim().toLowerCase();
+    if (!trimmed) return;
+    setAddError(null);
+    setIsAdding(true);
+    try {
+      await putAdminOperator(origin, trimmed, addRole);
```
wesbillman, line 188: [P2] Separate adding a principal from demoting an existing operator

This immediately PUT-upserts the entered key while the UI says “Add operator” / “Add” and defaults to `moderator`. With DB operators A and B, A can paste an existing operator key and click Add, silently replacing its role with moderator. Using A’s own key removes A’s roster-management access without the self-removal warning. Relay `api/admin/mod.rs:1037–1055` calls DB `relay_operators.rs:134–145` (`ON CONFLICT ... role = EXCLUDED.role`); the last-operator check at 169–174 permits this when another operator exists. The new confirmation only guards DELETE. Reject duplicate keys in Add, or make existing-role changes explicit and confirm demotions/self-access loss. Cover the actual form/PUT path with two operator rows; current DELETE-confirmation and synthetic role-prop tests do not exercise it.

desktop/src-tauri/src/commands/admin/origin.rs
```diff
@@ -0,0 +1,284 @@
...
+}
+
+/// Returns true when `host` is a loopback address (`localhost`, `127.x.x.x`,
+/// `[::1]`). This mirrors `media_download.rs`'s localhost carve-out.
+fn is_loopback_host(host: &str) -> bool {
+    host == "localhost"
+        || host == "[::1]"
+        || host
+            .parse::<std::net::IpAddr>()
+            .is_ok_and(|ip| ip.is_loopback())
```
wesbillman, line 146: [P2] Match the relay’s supported .localhost manual origin

This predicate rejects `admin.localhost` as non-loopback, so saving `http://admin.localhost:3000` fails at the HTTP scheme guard (lines 64–74). That is the actual `just admin` default (`Justfile:498`), and relay `api/admin/auth.rs:109–155,534–538` explicitly advertises/verifies HTTP for `.localhost` names. An IP alias is not equivalent because admin routing checks the configured Host exactly (`auth.rs:99–106`). Support explicit manual entry of this loopback origin, preserving its Host/NIP-98 URL and loopback connection behavior, and add a real-value regression. Keep untrusted discovery prefill/probe restrictions separate; do not auto-probe internal advertised origins.

wesbillman (changes requested):

**Changes requested at head `0b972699c51d010ebdf349498fbe6cbc40468d16`, base `4952f5810290d00a1dd9413235b39e9a7154acf6`.**

### P2: Do not let “Add operator” silently demote an existing operator

`AdminConsoleStaffingTab.tsx:97–105` sends an immediate role upsert, while the form says “Add operator” / “Add” and defaults to `moderator` (lines 81, 195, 229). With DB operators A and B, paste an existing operator key into this form and leave the default role: clicking Add changes that account to moderator without a role-change confirmation. If A submits their own key, A loses roster-management access and cannot reverse the change themselves. The new Remove confirmation and self-removal warning protect DELETE only, not this PUT path.

This is confirmed by the actual producer: `crates/buzz-relay/src/api/admin/mod.rs:1037–1055` forwards the role to `crates/buzz-db/src/store/relay_operators.rs:134–145`, whose conflict branch updates the existing role. The last-operator guard at lines 169–174 deliberately permits the demotion when another operator exists. This is accidental individual access loss, not a claim that the deployment can lose its last operator.

**Exit criterion:** distinguish adding a new principal from modifying one. Reject an already-present key in Add, or explicitly present the existing→requested role change and confirm demotions, including a self-access warning. Keep relay authority and last-operator protection unchanged. Add a real form/IPC regression with two operator rows proving that an existing-key submission does not silently demote, while adding a new moderator still works. The existing event tests cover DELETE confirmation and synthetic role-prop changes, not this PUT submission. This was missed in earlier review rounds; it is not a regression introduced by the four latest fixes.

### P2: Accept the relay’s supported `.localhost` manual origin

`desktop/src-tauri/src/commands/admin/origin.rs:123–130` recognizes only bare `localhost` and loopback IPs. Consequently `AdminOrigin::parse("http://admin.localhost:3000")` takes the non-loopback HTTP rejection at lines 64–74, so an operator cannot save the repository’s local admin origin. This is an existing producer contract, not a proposed new deployment: `Justfile:498` defaults `BUZZ_ADMIN_HOST` to `admin.localhost:3000`, and relay `api/admin/auth.rs:109–155,534–538` deliberately advertises and verifies HTTP for every `.localhost` name. Substituting `127.0.0.1` is not equivalent: admin routing checks the configured Host exactly (`auth.rs:99–106`).

**Exit criterion:** support explicit manual entry of the relay’s loopback-name contract, preserving the configured hostname for Host/NIP-98 signing and ensuring it connects as loopback. Add a regression for the real `http://admin.localhost:3000` value. Do not relax the separate prohibition on automatically probing untrusted advertised internal origins. This is a missed compatibility path, not a regression in the four latest fixes.

### Non-blocking P3: Preserve status in read-only feedback detail

`AdminConsoleFeedbackTab.tsx:497–507` hides the only detail status representation when `canMutate=false`; `FeedbackFields` (lines 340–358) has no passive status row. A disabled-auth reader can see a reviewed/archived badge in the list, then loses that information when opening detail. A passive status badge would preserve the read-only contract without adding mutation controls. This is not an additional merge gate.

### Verified scope and limits

All four findings from review 5070332066 are closed: manual setup remains reachable without discovery; `scope=all` reaches the native query and relay while retaining the relay’s escalated-only default; reason-audience copy matches the actual notices; resolve retries freeze the command and derive success copy from the relay response. DELETE confirmation and hidden-Staffing reset remain present.

The integrated review traced Desktop Settings/Reports/Feedback/Staffing through native admin transport and exact-head relay/DB producers, including authorization, report/action closed sets, mutation retries, capability transitions, and attachment budgets/lifetimes. Community moderation, mobile/web operator UIs, and deployment configuration are outside this Desktop delivery. Reopen replay preserves the original audit and cannot reopen a later resolution again; its editable optional reason is not being promoted into an enforcement blocker. Raw Staffing error presentation remains the declared follow-up. The proposed blanket private-address prohibition on signed requests is not a blocker: the agreed boundary permits manually trusted internal origins after explicit Save, and discovery does not probe its advertised destination.

Source-only validation: immutable GitHub source/diffs and existing regression tests were inspected; key evidence blobs were verified against the pinned Git tree. No PR checkout, build, test execution, or live app/relay run. This is not a runtime pass.

wesbillman (commented):

## Review clear: prior blockers resolved

Reviewed head `21f34f0138a4e00782f914d28a90346b994b650c` against base `42aeb1571cc1f1ed4d73f05b420474be970021ab`, including the corrections since review 5073088157. No remaining blocking finding in the reviewed paths and agreed delivery contract.

- **Staffing:** Add now requires a loaded roster and rejects normalized existing keys before PUT. The actual form/IPC regression checks zero PUTs for an existing operator and a successful new-moderator add with roster refresh. This closes the demonstrated accidental-demotion path; the API remains an upsert, so this is snapshot-based duplicate prevention, not atomic server-side create-only behavior.
- **Manual localhost:** `http://admin.localhost:<port>` is accepted and pinned to loopback without rewriting the hostname used for Host or NIP-98 signing. The added production-client/probe test inspects the signing URL and both wire Host headers. Discovery still requires explicit Save before contacting an advertised admin origin.
- **Integration:** Read-only feedback detail now shows its status. The “Relay admin” rename preserves the old `moderation` URL token in route validation and AppShell. Report scope, frozen resolve retries, cancel/reopen semantics, and operator/moderator authorization remain consistent with the previously reviewed contract.

Non-blocking follow-ups: the staffing regression does not explicitly exercise uppercase/whitespace, self-key, or pending/failed roster loading. Discovery’s reserved-domain check omits `.localhost`, so failed system DNS can allow a loopback prefill; explicit Save still prevents automatic access. These do not reopen the agreed fixes.

Validation: immutable source/diff and existing-test inspection only, with independent native and staffing/feedback reviews integrated. No checkout, builds, tests, or PR-code execution; no runtime-pass claim. Separate community moderation, mobile/web operator UX, deployment configuration, and unrelated merged-main changes were outside this corrective review. This is a COMMENTED review, not approval.

wesbillman (changes requested):

## Changes requested: two regressions since the previous clear review

Reviewed head `b0b5d6467ea6f87c58f86f62b796a13a319e761c` against exact base `88687876f7808a2fd742b7eb2e4b9f87d999ad8d`, with a source-level delta from previously cleared head `21f34f0138a4e00782f914d28a90346b994b650c`.

### P1: Do not automatically trust and probe the advertised admin origin

`desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx:322–335` now persists discovery into the manual-origin store and calls `runProbe` without confirmation. The unchanged native discovery implementation was safe only as a prefill: `discovery.rs:50–55` rejects bare `localhost` but not `.localhost`, and lines 75–79 accept DNS lookup errors expressly because the operator would still have to Save.

Concrete path: on a platform where system DNS does not resolve `.localhost`, let the connected relay advertise `http://admin.localhost:<port>` and open Admin with no saved origin. Discovery accepts the lookup failure; the new effect saves the origin; the dedicated admin client then deliberately resolves `.localhost` to `127.0.0.1` (`dns.rs:34–41`). The client reaches that loopback service without the user choosing it. If it answers `401` with `WWW-Authenticate: Nostr`, `mod.rs:199–217` signs and sends a NIP-98 probe with the active app key. The route/method are fixed; this is unauthorized local-network access and a signed identity disclosure, not private-key extraction. Disabling redirects does not protect the initial destination.

This was explicitly non-blocking in the prior review only because explicit Save prevented all contact. Removing Save makes it reachable. The new jsdom success test (`adminConsolePanelEvents.jsdom-test.mjs:1677–1750`) mocks discovery returning literal loopback and bypasses the native checks, so it cannot establish the advertised-origin security contract.

**Smallest exit:** restore discovery as prefill-only until explicit Save, preserving manually trusted internal origins. If zero-confirmation discovery is intentional, it needs a separately secured advertised-origin transport policy, including `.localhost`/DNS failure and connection-time address binding; the manual syntax parser alone is not sufficient. Cover the real discovery-to-probe boundary, not just the IPC success mock.

### P2: Reconcile the active principal after an in-place self-demotion

The new role selector calls `handleRoleChange` (`desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx:383–400`), whose successful PUT only increments the local roster generation (`209–211`). With DB operators A and B, sign in as A and change A’s row to moderator. The mutation commits because B remains an operator (`crates/buzz-db/src/store/relay_operators.rs:168–179`). The subsequent roster GET correctly rejects A as a moderator (`crates/buzz-relay/src/api/admin/auth.rs:285–307,325–333`; `admin/mod.rs:913–924`).

But the console still says **Connected as operator**, retains Staffing as the selected/visible tab, and replaces the roster with an error. The parent SettingsCard owns the probe and is never updated (`AdminConsoleSettingsCard.tsx:363–383,434–442`); Panel’s existing operator-to-moderator reset cannot fire because its role prop has not changed (`AdminConsolePanel.tsx:929–943`). This is incorrect post-mutation state, not an authorization bypass. Re-probing/remounting repairs presentation; only another operator can restore A’s staffing authority.

**Exit:** after successfully changing the current principal’s role, refresh the parent authorization state and move to an allowed tab rather than only reloading the operator-only roster. Exercise the real SettingsCard mutation-to-probe path with A/B, assert moderator status and Reports/Feedback without Staffing. The existing test manually supplying a changed role prop proves only the downstream reset; the new test covers another user’s promotion (`adminConsolePanelEvents.jsdom-test.mjs:6101–6184`). Consider the same explicit self-access warning already used by Remove, but the stale principal transition is the concrete blocker. Add duplicate rejection and server last-operator protection remain intact.

### Coverage and limits

The relay remains authoritative for authorization, enforcement, provenance and audit, separately from community owner/admin moderation. Reviewed Settings/navigation/identity/origin transitions; native discovery/signing/transport; report list/detail/resolve/cancel/reopen and retry contracts; feedback lifecycle; staffing; timeout/sidebar/report-error deltas; and command registration. Checked duplicated report/action/target/feedback/role/source/probe/settings closed sets. Report lifecycle, feedback, native admin transport and auth blobs remain identical to the prior clear where noted; the Panel change is presentation-only. Prior report scope, frozen resolve retries, manual `.localhost` support, read-only feedback status, and Add duplicate fixes remain intact.

Non-blocking coverage follow-up: the added timeout tests exercise imperative store/pure helpers, not the changed reactive hooks. Hook-level fake-timer tests would make expiry and interval cleanup regressions falsifiable. No concrete new runtime failure was established there. Existing raw staffing error presentation and snapshot-based Add concurrency are not additional merge gates.

Validation was immutable source/diff and existing-test inspection only, with independent review lanes integrated. No PR checkout, build, test execution, or live app/relay run. Separate community moderation as a whole, mobile/web operator interfaces, deployment configuration, and unrelated rebase changes were excluded.

wesbillman (changes requested):

## Changes requested: two P2 Staffing defects

Reviewed head `efa68d25829d357b7b557cbe92eb2e48cd094be0` against base `3c7f288c60d67df78577b237e27c3dfc8831aaa1`, focusing on changes since reviewed head `b0b5d6467ea6f87c58f86f62b796a13a319e761c` and the agreed integration contracts. Relay operator/moderator authority remains separate from community administration; the relay is authoritative for permissions and mutation outcomes.

### P2: Refresh the active principal after successful self-demotion (unresolved)

[Staffing’s successful role change](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx#L205-L223) only increments its roster generation. With two effective DB operators A and B, sign in as A and change A to moderator. The PUT succeeds, then the operator-only roster reload returns 403, but Settings still says **Connected as operator** and retains Staffing. The [parent’s role](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L443-L462) comes exclusively from its previous probe; no mutation callback refreshes it. This is stale authorization presentation, not a server authorization bypass. Successful self-removal has the same reconciliation gap.

**Exit:** after changing/removing the current principal, refresh the parent probe and reconcile visible tabs/access. Exercise the actual Settings → Staffing mutation → principal transition, asserting moderator status and Reports/Feedback without Staffing (or denied state after removal). The existing [role-downgrade test](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/desktop/src/features/admin-console/adminConsolePanelEvents.jsdom-test.mjs#L4749-L4844) manually supplies a new role prop, so it cannot catch the missing upstream update.

### P2: Preserve the last-operator conflict’s recovery instruction (new regression)

The new [role-change 409 branch](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx#L215-L220) and [remove branch](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx#L235-L240) classify every typed 409 as a config-backed key. With no effective config operator and one DB operator, attempt to demote or remove that operator. The relay correctly rejects it with **“add a replacement operator first”** ([PUT](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/crates/buzz-relay/src/api/admin/mod.rs#L1037-L1051), [DELETE](https://github.com/block/buzz/blob/efa68d25829d357b7b557cbe92eb2e48cd094be0/crates/buzz-relay/src/api/admin/mod.rs#L1108-L1121)). The UI instead claims this mutable DB row is config-backed, hiding the supported recovery and directing the operator toward the wrong cause.

**Exit:** display the relay’s parsed error message for these conflicts, or distinguish a stable conflict reason rather than treating HTTP status as one. No broader transport redesign is needed. Cover both immutable-config and last-operator responses for role change/removal; the new 409 tests currently encode only the immutable interpretation.

### Fixes, boundaries, and validation

Cross-host discovery now requires explicit Save, closing the prior automatic cross-host/`.localhost` signing path. Intentional same-host discovery, including another port, is not a finding. The discovery DNS check is not connection-time address binding, but the proposed additional proof-disclosure scenario requires a TLS endpoint valid for the already-trusted hostname; no additional credential capability was established. Treat that as non-blocking hardening, preserving manually trusted internal origins.

The report extraction preserves the prior implementation and frozen retry/action contracts. Inspected timeout/banner/sidebar lifecycle changes and new reactive tests; no additional defect established. Native bodyless DELETE retains its wire/signing contract. Independent lanes were integrated. Validation was exact-source/diff and test-source inspection only: no PR checkout, build, test execution, or live workflow. Unrelated rebase changes, separate community moderation as a whole, and mobile/web operator interfaces were excluded.

wesbillman (dismissed):

## Changes requested: one P2 origin-switch regression

Reviewed head `78da302e53aea6b90a37ff38757762977f8cbf35` against exact base/merge-base `5079c770fe30bb3d8204822ce6c2431eacac6d4b`, focusing on the corrective delta since `efa68d25829d357b7b557cbe92eb2e48cd094be0`. Relay staffing remains distinct from community administration, and each relay is authoritative for its own principal’s permissions.

### P2: Fence the self-mutation refresh to its original origin/session

The new [Settings callback](https://github.com/block/buzz/blob/78da302e53aea6b90a37ff38757762977f8cbf35/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L454-L463) captures `savedOrigin`, but [Staffing invokes that callback after awaiting its mutation](https://github.com/block/buzz/blob/78da302e53aea6b90a37ff38757762977f8cbf35/desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx#L230-L249), including after the panel has unmounted. Advanced/input/Save remain usable during the mutation.

Reproduction: with two effective operators on relay A and the current identity independently authorized on relay B, start self-removal on A and delay its successful response. Save B as the admin origin and let B’s probe authorize. When A’s DELETE resolves, the retained callback calls `runProbe(A)`. That starts a fresh probe, superseding B’s probe/state; A now returns denied. Settings still has B saved but displays **Access denied** and hides B’s otherwise-authorized console. Self-demotion similarly lets A’s moderator result hide Staffing on B where the identity is still an operator.

The existing [probe abort check](https://github.com/block/buzz/blob/78da302e53aea6b90a37ff38757762977f8cbf35/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L383-L403) does not fence this: the stale callback creates the *newest* controller. The origin edit unmounts the child, not the Settings session, and the callback does not check the current session/origin. This is a user-visible cross-origin state regression introduced by the refresh wiring, not a server authorization bypass. Re-probe recovers it, but the wrong denial/role persists until then.

**Exit:** ignore a self-mutation completion if its captured origin/session is no longer current, before it starts a probe. Reuse the existing context-fencing pattern rather than adding a new lifecycle owner. Add a Settings-level deferred-mutation regression: authorize B before settling A’s mutation, then assert B’s authorization/panel survives and the stale completion cannot replace it.

### Resolved findings and validation boundary

The previous two P2 findings are fixed on the ordinary same-origin path: self-demotion/removal now refreshes the parent principal, and Staffing preserves the relay’s distinct immutable-config versus last-operator recovery text. The new Settings-level self-demotion test exercises the actual parent wiring. Production source is unchanged between `0ab00433` and `78da302e`; that last commit adds tests only.

This was source/diff and test-source inspection on the pinned Blox host, with independent metadata review. No PR checkout, build, test execution, or live reproduction was performed. The one-time exact-head CI snapshot had 53 successful checks, 27 skipped, and two still running; it is not an all-green claim. Existing accepted discovery/trust behavior and unrelated base changes were not reopened. The sole requested fix is the stale-origin callback above.

wesbillman (commented):

## Review clear: origin/session callback blocker resolved

Reviewed head `be0b4833045fc7f993c9e80c1f7ac7dea33c318c` against exact base `77729abfb692b25a0f4ec4a69add86af2e32c0dd`, focusing on the corrective delta since review 5269776708 and the intervening main merge. No remaining blocking defect found in that bounded scope.

- **Origin switch:** the [Settings callback](https://github.com/block/buzz/blob/be0b4833045fc7f993c9e80c1f7ac7dea33c318c/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L486-L497) checks its captured origin against the synchronously maintained current-origin ref before starting a probe. A delayed self-removal/demotion on A can no longer replace B’s authorization after B is saved. The Settings → Panel → Staffing regression authorizes B before settling A’s DELETE, then checks the probe count, targets, and surviving panel.
- **Session teardown:** the [existing unmount cleanup](https://github.com/block/buzz/blob/be0b4833045fc7f993c9e80c1f7ac7dea33c318c/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L290-L311) now clears that same ref. A retained mutation callback cannot start a fresh probe from the retired Settings session after an identity change. The added production-seam test unmounts before resolving DELETE and asserts no additional probe. Ordinary live-session self-mutation still refreshes principal state; parsed last-operator versus immutable-config 409 recovery messages remain intact.
- **Integration:** the main merge has no manual resolution delta. The resulting relay router retains both the feature’s operator routes and main’s adjacent member-restriction routes. Previously reviewed report recovery, discovery/transport, and staffing implementations are unchanged where verified by blob identity. Same-host discovery, explicit Save for cross-host advertisements, manual internal origins, and separate relay/community authority remain the accepted contract.

**Validation limits:** immutable source/diff and test-source inspection on the pinned Blox host, with independent metadata review and coordinator verification. No checkout, build, test execution, or live reproduction. The teardown regression is not StrictMode-wrapped; a separate StrictMode Save test and the load-effect ordering cover re-arming, not a combined StrictMode/self-mutation runtime test. No speculative same-identity A→B→A finding was promoted into a blocker. Unrelated mainline changes and separate mobile/web operator UX were excluded.

**CI is not certified green.** The frozen exact-head snapshot at 2026-09-21 20:02:32Z showed 31 successful checks, 19 skipped, one failure, and four in progress. [Desktop Domain / Desktop Smoke E2E (3) failed](https://github.com/block/buzz/actions/runs/35645960340/job/106487136415); its cause is unclassified here. No polling or reruns were performed. This source-review clearance does not waive CI or constitute approval.

wesbillman (changes requested):

## Changes requested: preserve the denied-state regression test

Reviewed HEAD `d99e98952fe9e317a882bb12091e659c52fd39fd` against exact BASE `77729abfb692b25a0f4ec4a69add86af2e32c0dd`, focusing on the eight test-only commits since the [clear review at `be0b4833`](https://github.com/block/buzz/pull/4768#pullrequestreview-5271356947). Production code and the base are unchanged. One coverage regression needs a small repair; no runtime defect is asserted.

### P2: Restore the distinct `nip98Denied` recovery-state check

The pruning deletes [`denied-badge-copy-button`](https://github.com/block/buzz/blob/be0b4833045fc7f993c9e80c1f7ac7dea33c318c/desktop/src/features/admin-console/adminConsolePanel.test.mjs#L992-L1026) without equivalent coverage. Unlike the genuinely duplicated cases, it mounted the real Settings card, returned `nip98Denied` from `admin_probe`, and asserted both the current identity’s pubkey text and its Copy button. That is the recovery UI a denied user uses to request operator access.

At this head, the two `admin-denied-*` selectors occur only in the [production badge](https://github.com/block/buzz/blob/d99e98952fe9e317a882bb12091e659c52fd39fd/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L52-L96). Searching the desktop tree for those selectors and `nip98Denied` finds no positive UI assertion replacing the deleted test. The surviving [origin-switch](https://github.com/block/buzz/blob/d99e98952fe9e317a882bb12091e659c52fd39fd/desktop/src/features/admin-console/adminConsolePanelSession.jsdom-test.mjs#L1293-L1471) and teardown cases use denial only as an unwanted late-result sentinel; they do not check that a legitimate denied response displays the correct recovery identity or Copy control. Rust probe-classification tests do not exercise that UI either.

**Smallest fix / exit criterion:** restore this test, or add an equivalent real-Settings-card case asserting the expected pubkey and Copy control after `nip98Denied`. It should fail if the pubkey is dropped/miswired or the button is removed. No new harness, architecture change, or broader test expansion is needed. This preserves an existing falsifiable production-seam regression rather than adding speculative coverage.

**Otherwise clear in this delta:** the four split suites match the existing runner; every suite registers the shared IPC/toast reset; the reviewed table consolidations retain their cases; the prior origin-switch and teardown regressions remain. The positive Kick assertion survives in `resolve-rejection-surfaces-parsed-message`, so deleting its standalone duplicate is not a finding. Same-host discovery, explicit Save for cross-host advertisements, manual internal origins and separate relay/community authority remain accepted. Stale test comments are non-blocking.

**Validation limits:** source/diff and test-source inspection only on the pinned Blox host, with independent source, harness/discovery and metadata lanes plus coordinator verification. No checkout, installation, builds, test execution, mutation experiment or live reproduction. The proposed failure checks above are source-based expectations, not observed test runs. Historical architecture and unrelated inherited changes were excluded.

**CI is not certified green.** The one exact-head snapshot at **2026-09-22 16:21:35Z** showed 47 successful checks, 26 skipped, one failure and two in progress. [Desktop Domain / Desktop Smoke E2E (4) failed](https://github.com/block/buzz/actions/runs/35749444171/job/106819729735); its cause is unclassified here. Desktop Core and Codex Security Review were still running. No logs, reruns or monitoring were performed.

wesbillman (commented):

## Review clear: denied-state regression restored

Reviewed head `a99491c21170b3b5e4de4cefad055941f2a37c95` against exact base `77729abfb692b25a0f4ec4a69add86af2e32c0dd`, limited to the four corrective commits since [review 5280890631](https://github.com/block/buzz/pull/4768#pullrequestreview-5280890631) at `d99e98952fe9e317a882bb12091e659c52fd39fd`. No remaining blocker found in that delta.

- **Prior P2 resolved.** The restored [denied-badge test](https://github.com/block/buzz/blob/a99491c21170b3b5e4de4cefad055941f2a37c95/desktop/src/features/admin-console/adminConsolePanel.test.mjs#L1033-L1067) seeds the current identity, mounts the production `AdminConsoleSettingsCard` under its QueryClient provider, returns `nip98Denied` through mocked IPC, and asserts the expected pubkey text and Copy control. The [mount helper](https://github.com/block/buzz/blob/a99491c21170b3b5e4de4cefad055941f2a37c95/desktop/src/features/admin-console/adminConsolePanel.test.mjs#L458-L493) does not replace the card or badge. Dropping/miswiring the displayed identity or removing the button would fail these assertions by source inspection. This meets the previous review’s explicit restore-or-equivalent exit; no mutation experiment was run.
- **Collateral changes clear.** The Rust probe/pubkey and desktop attachment, report, discovery-fallback, feedback-fixture and staffing-error consolidations retain the reviewed inputs and relevant assertions. The desktop runner globs and Rust test-module registration remain intact. The only production delta is [denial guidance](https://github.com/block/buzz/blob/a99491c21170b3b5e4de4cefad055941f2a37c95/desktop/src/features/admin-console/AdminConsoleSettingsCard.tsx#L52-L102): it no longer claims config is the only authorization source or recommends obsolete token mode. Identity/Copy wiring and state transitions are unchanged. Same-host discovery, explicit Save for cross-host advertisements, manually trusted internal origins, and separate relay/community authority remain accepted.
- **Bounded validation.** Source/diff and test-source inspection on the pinned Blox host, with independent regression-quality review and coordinator verification. No checkout, dependency installation, build, tests, clipboard interaction or live reproduction. The test does not mount the full Settings router or activate Copy; those are additional coverage opportunities, not unmet requirements of the prior card-level exit. Previously reviewed architecture, unrelated inherited code and mobile/web operator UX were not reopened.

**CI is not certified complete:** the exact-head snapshot at `2026-09-22T19:24:56.229146+00:00` showed **50 successful, 26 skipped and one in-progress check** ([Desktop Smoke E2E (2)](https://github.com/block/buzz/actions/runs/35762334304/job/106896236342)), with no failed check in that snapshot. No CI reruns or monitoring loop. This source-review clearance neither waives CI nor constitutes approval.

wesbillman (changes requested):

## Changes requested

### P1: Resolve the relay tenant UUID instead of using the desktop connection ID

Anchor: [`desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx:427-428`](https://github.com/block/buzz/blob/137343a628ada22144a25e169208de398c6f8c02/desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx#L427-L428).

The production fallback passes `activeCommunity.id` as the admin API’s `communityId`, but this is a locally generated connection identifier, not the relay’s tenant UUID. Normal onboarding sets it with `crypto.randomUUID()` (`desktop/src/app/App.tsx:477-485`), as does first-community initialization (`desktop/src/features/communities/communityStorage.ts:212-227`). `useCommunities` retains that ID; neither the TS wrappers nor the new native commands translate it.

The relay converts the query UUID directly to `CommunityId` (`crates/buzz-relay/src/api/admin/mod.rs:1274-1279`) and `list_restricted_page` queries `community_bans WHERE community_id = $1` (`crates/buzz-db/src/store/moderation.rs:656-690`). Consequently, an operator who connects normally to a community with an active ban/timeout sees **“No active bans or timeouts”**, because the query uses an unrelated UUID. The new lift/clear controls are therefore unavailable for real restrictions. The tests bypass this production source with the `communityId` override and canned IPC rows, so their passing result does not establish the contract.

**Fix/exit criterion:** obtain an authoritative tenant UUID associated with the selected relay/admin deployment rather than using the local connection ID. Exercise the production ID-resolution path with deliberately different local and server IDs; verify list and both removal commands use the server ID, and that unresolved/mismatched context is not presented as an empty restriction list. No broader admin-console redesign is required.

### P2: Move the tombstone row, not only its timestamp

Anchor: [`desktop/src/features/messages/lib/formatTimelineMessages.ts:478-506`](https://github.com/block/buzz/blob/137343a628ada22144a25e169208de398c6f8c02/desktop/src/features/messages/lib/formatTimelineMessages.ts#L478-L506).

The new `effectiveCreatedAt` changes the tombstone's timestamp, but the formatter still returns `visibleEvents.map(...)` in its original order (`:462-569`). The deleted target is filtered out (`:300-302`). The channel's consumers preserve that order: `ChannelScreen.tsx:404-430` → `useChannelPaneMessages.ts:25-48` → `buildMainTimelineEntries` (`threadPanel.ts:448-482`) → `buildTimelineItems` (`timelineItems.ts:174-278`). Nothing in that path repositions the formatted tombstone.

**Reproduction by source trace:** retain yesterday's original and a surviving message from today in the window, then receive the original's deletion marker and a `message_deleted` system event today. The visible output is `[todayMessage, tombstoneWithYesterdayTimestamp]`, not `[tombstone, todayMessage]`. Besides failing the promised in-place replacement, this newly makes `buildDayGroupBoundaries` (`timelineSnapshot.ts:110-132`) insert yesterday's heading below today's messages. Before this timestamp-only change, the appended row had the deletion event's current timestamp.

**Fix/exit criterion:** place the tombstone at the original's chronological position when the original is available, preserving the documented own-timestamp fallback when it is absent. Add a surviving intervening message to the regression fixture and assert output/rendered order and cross-day grouping. The new test has only one surviving visible row and asserts its `createdAt`, so it cannot catch this failure. No relay or moderation-policy change is needed.

## Scope and validation

- Immutable HEAD `137343a628ada22144a25e169208de398c6f8c02`, BASE/merge-base `26ede6dfa2496993aa62ce5781d2112df4c2d009`. Delta review from previously cleared `a99491c21170b3b5e4de4cefad055941f2a37c95`, isolating the Restrictions, self-report and tombstone commits from merged-main changes. Existing discovery-consent, role, and recovery decisions remain in force.
- Both independent lanes are integrated. Restrictions identity was independently confirmed; the timeline consumer chain was independently traced. The kind:1984 self-p-tag change correctly opts into `allow_self_tagging` only for reports while retaining existing defaults. Its tests cover the signing primitive, not command deserialization; that limited coverage gap is non-blocking. Previously cleared discovery/denied-state behavior is not reopened, and unchanged web/mobile surfaces are outside this delta review.
- Source-only review on the pinned review host. No checkout, build, test execution, dependencies, PR-code execution or CI reruns. Existing exact-head checks include successful Desktop, Rust Unit Tests and Relay E2E checks; PostgreSQL Tests is failing and Run Codex Security Review was still in progress at the metadata snapshot. Those results are not claimed as validation performed here, nor attributed to this defect.

wesbillman (changes requested):

## Changes requested: two prior blockers remain, one new regression

Bounded re-review of HEAD `21a58fe19473d204bcd7390583caa2c44c62d5fb`, BASE `0cc63fe3e607cf47ff32bc61e7eaca2f21e6cc23`, covering the four branch commits since [review 5285352623](https://github.com/block/buzz/pull/4768#pullrequestreview-5285352623) and their integrated result. Previously accepted discovery-consent and relay/community authority decisions remain unchanged.

### P1, unresolved: Restrictions still use the desktop connection ID

The [production fallback](https://github.com/block/buzz/blob/21a58fe19473d204bcd7390583caa2c44c62d5fb/desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx#L417-L428) still uses `activeCommunity.id`. Onboarding generates that ID locally with `crypto.randomUUID()`; list and both removal wrappers forward it unchanged. The relay directly converts the query UUID into its database community ID. Connecting normally to a community with bans/timeouts therefore queries an unrelated tenant and displays an empty list, hiding the removal controls. Relay-scoped origin storage does not translate this ID.

**Unchanged exit:** resolve the authoritative tenant associated with the selected relay/admin deployment. Exercise the production resolution path with different local/server IDs, verify list and both removal commands use the server ID, and show unresolved/mismatched context rather than a false empty list.

### P2, unresolved: Tombstone timestamp changes without moving its row

The [formatter](https://github.com/block/buzz/blob/21a58fe19473d204bcd7390583caa2c44c62d5fb/desktop/src/features/messages/lib/formatTimelineMessages.ts#L462-L506) and its tests are unchanged since the prior review. It still maps the deletion-time array position while adopting the original timestamp. With yesterday’s original retained in the loaded window and a surviving message from today, deleting the original produces `[todayMessage, yesterdayTombstone]`. Order-preserving consumers and positional day grouping then place yesterday’s heading below today’s.

**Unchanged exit:** position the replacement in the original chronology when available, preserving its own-timestamp fallback otherwise. Add an intervening survivor and assert output order and cross-day grouping; the current scalar timestamp assertion cannot detect this defect.

### P2, new: Keep event report targets out of the public-key widget

The new [target rendering predicate](https://github.com/block/buzz/blob/21a58fe19473d204bcd7390583caa2c44c62d5fb/desktop/src/features/admin-console/AdminConsoleReportsTab.tsx#L769-L785) sends both `pubkey` and `event` targets through `PubKey`. A normal 64-hex event ID passes this predicate. [PubKey](https://github.com/block/buzz/blob/21a58fe19473d204bcd7390583caa2c44c62d5fb/desktop/src/shared/ui/PubKey.tsx#L41-L76) encodes and copies that value as an `npub`, so the operator receives a public-key identifier rather than the reported event identifier. The separate “Event ID” field contains the report event ID, not the target, and does not repair this.

**Fix:** use `PubKey` only for public-key targets; retain the raw target event ID or an event-specific representation for event reports. Add a valid 64-hex event-target fixture asserting the displayed/copied identifier and retain public-key behavior. The inspected report fixtures mostly use short event targets, which bypass the new branch.

## Validation and scope

Source-only review on the pinned Blox, independent timeline lane plus coordinator verification. The collateral attachment-save, origin-storage and report UI changes were inspected; no additional new blocker is asserted. No checkout, dependencies, builds, tests, clipboard interaction or PR-code execution. Unchanged architecture and unrelated merged-main/web/mobile behavior were not reopened.

Existing exact-head CI snapshot: **53 successful, 27 skipped, one cancelled** (`Run Codex Security Review`). These are hosted results, not validation performed here; no CI reruns or monitoring. Merge criteria are the three fixes and regressions above, with normal validation gates still applicable.

desktop/src/features/admin-console/AdminConsoleStaffingTab.tsx
```diff
@@ -0,0 +1,790 @@
...
+   * the active community from `useCommunities` is used. Intended for unit
+   * tests that need to exercise the restrictions surface without seeding
+   * localStorage with a community entry.
+   *
+   * Do not pass this prop in production code.
+   */
+  communityId?: string;
+}) {
+  const { activeCommunity } = useCommunities();
+  const effectiveCommunityId = communityIdOverride ?? activeCommunity?.id;
```
wesbillman: **P1, still unresolved: use the relay tenant ID, not the desktop connection ID.** `activeCommunity.id` is generated locally with `crypto.randomUUID()` (`communityStorage.ts:212–227`). This value still reaches initial list, the new next-page call, lift-ban, and clear-timeout unchanged; the relay interprets it directly as its database `CommunityId` (`crates/buzz-relay/src/api/admin/mod.rs:1274–1278`). With local ID L ≠ server ID S and active restrictions under S, Staffing queries L and falsely displays “No active bans or timeouts,” hiding removal controls.

The new paging fixture injects `CM_COMMUNITY_ID` through the test override rather than exercising production resolution. **Unchanged exit:** resolve the authoritative tenant bound to the selected relay/admin deployment, use it for every page and both removals, and show unresolved/mismatched context rather than false emptiness. Test the real resolution path with deliberately distinct local/server IDs.

desktop/src/features/messages/lib/formatTimelineMessages.ts
```diff
@@ -475,10 +475,35 @@ export function formatTimelineMessages(
...
+      } catch {
+        // Non-JSON or unexpected shape — fall through.
+      }
+      return event.created_at;
+    })();
     return {
       id: event.id,
       renderKey: event.localKey ?? event.id,
-      createdAt: event.created_at,
+      createdAt: effectiveCreatedAt,
```
wesbillman: **P2, still unresolved: changing the timestamp does not reposition the tombstone.** This file and its test are byte-identical to the previously reviewed head. `visibleEvents.map` preserves deletion-time array position while adopting the original message’s timestamp. With yesterday’s retained original, today’s surviving message, and today’s deletion, output remains `[todayMessage, yesterdayTombstone]`. `buildMainTimelineEntries` preserves that order, and `buildDayGroupBoundaries` renders Today before Yesterday.

**Unchanged exit:** position the tombstone in the original chronology when the target is available, preserving its own-time fallback otherwise. Add an intervening survivor and assert final row order and cross-day grouping through the production functions. The existing `find(...).createdAt` assertions exercise timestamps, not position; their fixtures cannot detect this failure.

desktop/src/features/admin-console/AdminConsoleReportsTab.tsx
```diff
@@ -0,0 +1,956 @@
...
+      {isHex64(data.target) &&
+      (data.targetKind?.toLowerCase() === "pubkey" ||
+        data.targetKind?.toLowerCase() === "event") ? (
+        <div className="flex gap-2 text-xs">
+          <span className="w-28 shrink-0 text-muted-foreground">Target</span>
+          <PubKey
+            className="min-w-0"
+            interactive
+            pubkey={data.target}
+            testId="report-target-pubkey"
```
wesbillman: **P2, still unresolved: event-target detail uses a public-key identifier.** The new event-author change fixes list rows only. Opening an event report with a normal 64-hex target E still selects `PubKey`, which displays/copies `npub(E)` (`PubKey.tsx:72–76,123–126`). The nearby “Event ID” field contains the report event’s ID R, not E, so it does not supply the correct target identifier.

**Unchanged exit:** use `PubKey` only for public-key targets; display/copy event targets as raw event IDs or an event-specific representation. Extend the new valid-hex list fixture through detail/copy, using E ≠ R and retaining pubkey-target behavior. The current added fixture stops at the list row and cannot detect this detail defect.

wesbillman (dismissed):

## Changes requested: the same three blockers remain

Re-reviewed HEAD `a8fcd3d3f3292ee1ab0c7d5630dcf23ffcd8258b` against BASE `0cc63fe3e607cf47ff32bc61e7eaca2f21e6cc23`, including all four commits since [the previous review](https://github.com/block/buzz/pull/4768#pullrequestreview-5295003354). The new list-author fix does not repair event-target detail; pagination does not repair tenant-ID provenance; the tombstone formatter/tests are unchanged. Updated evidence and unchanged exit criteria are in the three inline findings. No additional blocker is asserted.

The report-author join, pagination plumbing, production-bound signer/origin tests, and shared attachment fetch/save delta were inspected. Accepted same-host discovery, cross-host consent, and relay/community authority decisions remain unchanged.

**Validation:** source-only on the pinned Blox, including an independent timeline lane and coordinator verification. No checkout, execution, builds, tests, or clipboard exercise. Existing exact-head CI snapshot: 49 successful, 27 skipped, one in progress, one cancelled; not a completed green run and not tests performed here. Unchanged unrelated surfaces were not reopened.

**Merge criteria:** resolve the three carried findings and add their production-path regressions; normal validation gates still apply.

desktop/src/features/admin-console/AdminConsoleFeedbackTab.tsx
```diff
@@ -0,0 +1,566 @@
...
+        feedbackId,
+        attachment.sha256,
+        attachment.mime,
+        attachment.size,
+      );
+      if (saved) {
+        toast.success("Attachment saved");
+      }
+    } catch (e) {
+      setError(
```
wesbillman: **Nonblocking follow-up: keep Save retry reachable after a transient failure.**

For a non-image attachment, a failed native save sets this shared `error`; the early return at lines 296–311 then removes the Save button. `load()` is the only path clearing that error and is never invoked for non-images, so recovery requires going back to the feedback list and reopening the entry. This error path predates the previous reviewed head and does not reopen the agreed blockers.

Keep the Save control alongside a separate save error (clear it on retry), and extend the real-component IPC test with rejection → retry → successful save.

desktop/src-tauri/src/commands/admin/mod.rs
```diff
@@ -0,0 +1,1113 @@
...
+/// resolves its tenant. The desktop's local community ids are never sent.
+fn restrictions_url(
+    origin: &str,
+    route: &routes::AdminRoute,
+    cursor: Option<String>,
+    state: &crate::app_state::AppState,
+) -> Result<String, String> {
+    let origin = origin::AdminOrigin::parse(origin)?;
+    let host = buzz_core_pkg::tenant::relay_url_authority(
+        &crate::relay::relay_api_base_url_with_override(state),
```
wesbillman: **Nonblocking hardening: carry the caller’s restriction scope.**

The list/lift invokes supply origin and member data, while this helper derives the tenant from mutable native relay state. An expected relay captured with the displayed list would let the native boundary reject stale commands explicitly, using the existing relay-scope comparison helper and the same resolved URL snapshot.

This is not a verified cross-community regression: the URL is captured before the command’s first await, the UI subtree remounts, and the ordinary switch path awaits reset/IPC work before applying the new relay. No reachable reordered-first-poll schedule was reproduced in this source-only review. If strengthened later, keep it local to restriction commands and add a caller-A/native-B rejection test rather than redesigning all admin sessions.

wesbillman (commented):

## Rereview clear: the three carried blockers are resolved

HEAD `335d808b11e3f7ea44c95f60b930fd7d8dc9250c` against BASE `0cc63fe3e607cf47ff32bc61e7eaca2f21e6cc23`, including the eight commits since [the previous review](https://github.com/block/buzz/pull/4768#pullrequestreview-5296405127).

- Restrictions now use the active relay authority and server-side tenant resolution, rather than a desktop-generated community UUID; unknown hosts fail visibly.
- Event-target detail displays and copies the raw event ID. The unrelated tombstone formatter/test changes were reverted to BASE.
- The changed attachment Save path, Save-only octet-stream acceptance, success/cancel feedback, and restriction pagination generation checks were inspected with their production-path regressions.

No remaining blocking change requested in this focused rereview. Two nonblocking follow-ups are inline; they do not expand the agreed exit criteria. Accepted navigation, same-host discovery/cross-host consent, and relay/community role decisions are unchanged.

**Validation:** source-only on pinned Blox, with independent native, relay and UI lanes integrated. [Existing exact-head CI](https://github.com/block/buzz/actions/runs/35930976021) completed successfully; retrieved logs include 6,672 desktop core tests, 91 jsdom tests, native fetch/real-file-write/MIME tests, and PostgreSQL restriction tests. These were not run by this reviewer. No live dialog, clipboard, or community-switch timing exercise was performed. Normal merge gates still apply; this is not an approval.

wesbillman (commented):

**No blocking findings in the incremental changes since `335d808b11e3f7ea44c95f60b930fd7d8dc9250c`.** Reviewed exact head `0b4246c5d7c0106c5914ab36bb04c83915060524` against base `797012ff01a6d499959b45ed2e56f7927c6a4d6b`.

- Verified the non-image attachment Save failure/retry path, captured-relay propagation through restriction listing/pagination/removals, pre-send mismatch rejection, and merge integration. Existing discovery-consent and relay/community authority decisions remain unchanged.
- Existing CI on head+base merge `8e2271ff52e47e09fdff856f57c8912f7514d703` passed 6,676 general desktop tests, 92 focused admin tests, 475 PostgreSQL tests, and relevant integration checks. The Save retry assertion is consistent with the actual Tauri error wrapper.
- Overall CI is not green: the inspected snapshot failed in `buzz-agent::regressions::handoff_cap_binds_within_a_single_turn`; security review was still running. These remain separate validation gates. Source-only review on the pinned Blox host; no checkout, builds, tests, runtime probes, or CI actions performed.

wesbillman (commented):

**CLEAR:** no actionable regression found in the changes since [the prior clear review](https://github.com/block/buzz/pull/4768#pullrequestreview-5308244256).

- Reviewed canonical IPv6 saved-origin isolation, bounded NIP-11 success/error bodies and request timeout, the committed operator-grant response and authoritative roster refresh, and the restriction-route migration documentation. Separated the main release merge from branch changes. Existing discovery-consent, role-separation, and relay-first `communityHost` contracts remain intact.
- Source-only review on the pinned Blox, with an independent native/relay lane. No builds, tests, or PR code were executed. Existing exact-head CI snapshot: 49 successful, 26 skipped, two still running; this is not an all-green CI claim.
- No code fixes requested. Required CI and live-workflow validation remain separate merge evidence; unchanged surfaces were not re-reviewed.

Head: `8f716a42d0eb0746321666a6bfb09c19370875e5`. Base: `02753722a7dd06560402a5b92491b048968c1a63`.

kalvinnchau (approved):
🤖 Approved. The prior IPv6 origin-storage, discovery response-bound, and operator-update response findings are resolved in this revision.

PR #4789, opened 2026-08-04
feat(auth): enforce protected routing and fail-safe invalidation

cea-block:
## Why

Protected authorization needs monotonic invalidation, restoration, and version state so stale authority cannot be reused after revocation, restart, or cross-relay updates.

## What

- Add durable invalidation floors, authority restoration state, and versioned protected-object state
- Add authorization resolvers for current invalidation and restoration data
- Preserve replay-safe authorization events and schema invariants
- Add migration `0037` and its desired-schema representation
- Add migration and restoration regression coverage

The database restore layer now uses the HTTP types it needs directly.

## Review guide

- `buzz-db` invalidation, resolver, restoration, and version modules
- Monotonicity, replay, restoration, and protected-object invariants
- Migration `0037` and desired-schema compatibility
- This layer contains no router or protected Git enforcement changes

## Risk assessment

Medium. This PR adds durable invalidation and restoration state. Consumers fail closed when current authority cannot be resolved.

## Testing

- Invalidation ordering, replay, duplicate, and restoration cases
- Protected authorization-version cases
- Fresh and incremental migration coverage for `0037`
- Formatting, linting, and hosted repository checks

## References

- Depends on #4783
- Client binding status contract: #4790

PR #4790, opened 2026-08-04
feat(auth): install runtime status and session conformance

cea-block:
## Why

Clients need a small, provider-neutral contract for relay-authenticated current-binding presentation that is scoped to one connection and cannot be mistaken for durable profile authority.

## What

- Add a signed connection bootstrap contract with exact relay, author, domain, epoch, and time binding
- Add the ephemeral current-binding status contract and validation rules
- Add an I/O-free session fold that applies status, withdrawal, replay, expiry, and disconnect behavior
- Register the two relay-only ephemeral event kinds
- Keep this layer limited to five `buzz-core` files with no migrations, persistence, or route installation

## Review guide

- Bootstrap scope and signature validation
- Status payload bounds, revisions, freshness, and withdrawal semantics
- Session replay, malformed-input, expiry, and disconnect behavior
- Relay-only and ephemeral kind classification

## Risk assessment

Low to medium. This PR defines pure contracts and state transitions but does not install runtime routes or persistence.

## Delivery limitation

Status and moderation delivery is durable and crash-recoverable with at-least-once semantics, and clients deduplicate redeliveries.

Moderation commit-to-dispatch is in-process at-most-once: delivery records are crash-recoverable, but external moderation side effects are not replayed after a crash.

## Testing

- Bootstrap signer, domain, author, epoch, and time-bound cases
- Status revision, withdrawal, freshness, privacy, and replay cases
- Session malformed-input, expiry, disconnect, and non-reuse cases
- Formatting, linting, and hosted repository checks

## References

- Depends on #4789
- Runtime integration and durable delivery: #4847

### DIFF
--- a/crates/buzz-core/src/client_binding_bootstrap.rs
+++ b/crates/buzz-core/src/client_binding_bootstrap.rs
@@ -1,0 +1,640 @@
+//! Relay-authenticated connection bootstrap for client binding status.
+//!
+//! Kind `24245` is delivered only on the WebSocket connection whose native
+//! client supplied the echoed epoch. It binds the relay's NIP-11 signing key,
+//! the server-resolved authorization domain, and the authenticated event
+//! author before kind `24244` status can be consumed.
+
+use std::fmt;
+
+use nostr::{Event, EventBuilder, Keys, Kind, PublicKey, Timestamp};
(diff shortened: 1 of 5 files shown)


cea-block:
Closing, superseded by work by @wpfleger96

PR #4798, opened 2026-08-04
fix(acp): forward agent env vars into the MCP tool server

PierreCorbin:
## Problem

Agent-configured environment variables never reach the MCP tool shell.

When an operator sets an env var on an agent (e.g. a Notion token, a provider API key), the variable reaches the `buzz-acp` agent process but is dropped when the harness spawns the MCP tool server (`buzz-dev-mcp`). Every shell/tool the agent runs is a child of that server, so tool calls never see the variable — even though it is set correctly and shows in the agent's Edit form.

Process chain:

```
buzz-desktop -> buzz-acp -> buzz-agent -> buzz-dev-mcp -> tool shell
```

`build_mcp_servers` (in `buzz-acp`) constructs the MCP server's `env` from a small fixed allowlist (`BUZZ_RELAY_URL`, `BUZZ_PRIVATE_KEY`, optionally `BUZZ_AUTH_TAG` / `BUZZ_ACP_DISPLAY_NAME`). The tool shell inherits `buzz-dev-mcp`'s process env, so that fixed set is the whole world a tool sees. Agent-configured vars are never added, so they are silently unavailable to tools.

## Fix

Forward the agent's configured env vars into the MCP tool server, scoped per-agent and least-privilege:

- **Desktop** derives the exact set of env key **names** from the fully-layered agent env (`descriptor.env`) and passes them to `buzz-acp` via a new Buzz-controlled `BUZZ_ACP_FORWARD_ENV` manifest (comma-separated). A new helper `mcp_forward_env_manifest` builds it; `BUZZ_ACP_FORWARD_ENV` is added to `RESERVED_ENV_KEYS` so a user override cannot smuggle unrelated vars in or suppress forwarding.
- **`buzz-acp`** (`build_mcp_servers`) forwards **only** those named keys, reading each value from its own process env — never the whole environment, so Buzz's own plumbing (git config, internal `BUZZ_ACP_*` vars) stays out of the tool shell. Listed-but-unset names are skipped (absence ≠ empty). Names the fixed allowlist already set are not overridden.
- **Defense in depth:** `buzz-acp` refuses to forward reserved identity/secret keys (nsec, auth tag, relay URL, API tokens) even if one somehow appears in the manifest. The desktop already strips these from `descriptor.env`.

When `BUZZ_ACP_FORWARD_ENV` is unset (older desktops, non-desktop launches), the prior fixed-allowlist behavior is unchanged.

## Tests

- `buzz-acp` (`build_mcp_servers_tests`): named vars reach the MCP server; unlisted process vars are never forwarded; reserved keys are refused; listed-but-unset vars are skipped; unset manifest leaves the fixed allowlist untouched. Full `buzz-acp` lib suite green (675 passed).
- `buzz-desktop` (`managed_agents::env_vars`): `mcp_forward_env_manifest` returns `None` for empty env, lists all key names comma-separated, and carries names only (never values). Module suite green (45 passed).

Note: the full `buzz-desktop` Tauri bundle build requires prebuilt sidecar binaries not present in a fresh clone; the desktop unit tests above were validated with placeholder sidecars locally, and CI builds the real sidecars.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -4277,6 +4277,49 @@ async fn run_models(args: ModelsArgs) -> Result<()> {
     Ok(())
 }
 
+/// Identity and secret env keys that must NEVER be re-forwarded to the MCP
+/// tool server under their own name via [`forwarded_env_names`].
+///
+/// These mirror the desktop's reserved-key contract (identity/secrets and the
+/// relay URL). The desktop already strips them from an agent's configured
+/// env_vars before spawn, so they should never appear in `BUZZ_ACP_FORWARD_ENV`
+/// — this is a defense-in-depth guard so a bug or crafted list can't smuggle a
+/// forged nsec/auth-tag/relay into the tool shell. Comparison is
+/// case-insensitive, matching the desktop's `is_reserved_env_key`.
+const RESERVED_FORWARD_KEYS: &[&str] = &[
+    "BUZZ_PRIVATE_KEY",
+    "NOSTR_PRIVATE_KEY",
+    "BUZZ_AUTH_TAG",
+    "BUZZ_API_TOKEN",
+    "BUZZ_ACP_PRIVATE_KEY",
+    "BUZZ_ACP_API_TOKEN",
+    "BUZZ_RELAY_URL",
+];
+
+/// True when `key` is a reserved identity/secret key that must not be forwarded.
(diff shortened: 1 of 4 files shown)


PierreCorbin:
Related: #5231 solves the same underlying gap (agent env vars not reaching MCP tool shells) via an allowlist. Happy to close this in favor of that approach, or adapt — whichever maintainers prefer.

PR #4811, opened 2026-08-04
fix(release): tag immutable desktop candidates

wesbillman:
## Summary

Redesign the permanent Desktop release flow so unrelated merges to `main` cannot invalidate an already reviewed, green release candidate.

- Tag the immutable, API-confirmed release PR head instead of its later squash commit.
- Treat the merged PR—including an authorized owner/admin bypass—as publication authorization, while requiring trusted check evidence that was complete at merge time.
- Make tag creation idempotent and collision-safe: an existing tag succeeds only at the exact candidate SHA, and create races refetch before accepting equality.
- Replace ancestry-based previous-release discovery with a validated metadata ledger for side-history candidate tags.
- Compute the next release from the prior frozen base to the new frozen base, excluding only the prior release squash SHA so unrelated commits remain in the changelog.
- Preserve schema-1 production-tag migration and reject malformed metadata or equal/decreasing versions.
- Update operator documentation for the normal squash-merge workflow.

This is the reusable release process for `0.5.6` onward, not the retired one-shot `0.5.5` recovery path.

### Invariants covered

- Candidate creation → unrelated `main` merge → authorized squash merge → immutable candidate tag.
- Trusted producer IDs and merge-time completion timestamps; DCO's bounded post-merge exception remains isolated.
- Missing/spoofed checks, tampered candidates, ambiguous PR associations, conflicting tags, and equal/decreasing versions fail closed.
- Same-SHA retries succeed; different-SHA collisions fail.
- Legacy schema-1 tag-on-main migration and schema-2 side-history accounting both preserve the correct next-release changelog.

### Related issue

N/A — follows the Desktop release failures in #4788 and #4800 and the recovery revert in #4808.

### Testing

At clean commit `6a91fbed8147a48cf174997de0c3e4cb2fb26474`:

- `scripts/test-desktop-release-candidate.sh`

Both focused suites passed with HEAD unchanged. Princess Donut cleared the security/provenance surface, including the hostile merge-time timestamp cases. Mongo cleared the side-history ledger, migration, version-order, documentation, and contract-test surface.


tlongwell-block (dismissed):
Reviewed the immutable-candidate, merge authorization, trusted-check provenance/timing, side-history ledger, migration, and collision handling at c4711cb32d14f6dc048c9af4ad2dd9c5166cbc9e. Focused candidate and release-ref contract suites pass locally at unchanged HEAD; live required-check producer IDs match the default ruleset. Approving.

tlongwell-block (dismissed):
Re-reviewed at ddb82065498a9fdf472daa3e33bc780b3863b097. Independently reproduced the missing-auth failure Sami found, then verified this head supplies github.token to the candidate workflow. Verified retry behavior now permits only an equal-version tag at the exact candidate SHA while generation, wrong-SHA equality, and newer versions remain fail-closed. Both focused suites, diff check, and Python compilation pass locally at unchanged HEAD. The PR job still skips the version-bump-only validation step, so that green check is not evidence for the path; the added contract and candidate retry tests are. Approving this corrected head.

PR #4819, opened 2026-08-05
fix(buzz-dev-mcp): configurable shell-timeout ceiling; report the clamp

### LINKED ISSUE #4638: `buzz-dev-mcp`: the 600s shell cap is silently clamped and unconfigurable — long agent jobs die mid-work with no signal
## Summary

`crates/buzz-dev-mcp/src/shell.rs` caps every shell command at 600s and there is no way to raise it, no env var, and — the part that costs the most debugging time — **no report that the clamp happened**. A caller that asks for 3600s is silently given 600s, and on expiry the whole process group is killed.

## Where

```rust
// crates/buzz-dev-mcp/src/shell.rs
const DEFAULT_TIMEOUT_MS: u64 = 120_000;   // :16
const MAX_TIMEOUT_MS: u64 = 600_000;       // :17

let timeout_ms = p
    .timeout_ms
    .unwrap_or(DEFAULT_TIMEOUT_MS)
    .min(MAX_TIMEOUT_MS);                  // :141-144

cmd.kill_on_drop(true);                    // :178
set_process_group(&mut cmd);               // :179
```

`.min(MAX_TIMEOUT_MS)` is the whole story: `timeout_ms` is an accepted parameter of the tool (`:127 pub timeout_ms: Option<u64>`), so a caller reasonably believes it was honoured. It was not. Verified present at HEAD, not just on the v0.5.3 tag.

## Impact

Any agent-launched job longer than 10 minutes dies mid-work. Because `set_process_group` + `kill_on_drop` take out the entire group, the child's own output buffer goes with it, so what the agent sees is a truncated or empty log — not a timeout message.

Concretely, on a self-hosted relay driving an agent that fixes issues in a repo: four consecutive runs died at exactly 600s. Each had already done real work (written files, opened PRs, posted review comments) before the kill. The surviving artifact was a 15-byte log reading `Execution error`. The agent, given only that, concluded three separate times that its model backend was broken and proposed switching runners — a diagnosis that would have made things worse, since the same ceiling kills any runner identically. The ceiling is invisible from inside the sandbox; nothing in the failure points at it.

## Asks (either would be sufficient, in preference order)

1. **Make the cap configurable** — an env var (`BUZZ_DEV_MCP_MAX_TIMEOUT_MS`) or a config field. Operators running agents on their own hardware are already trusting them with a shell; a hard 10-minute ceiling is a policy decision that should belong to the deployment.
2. **At minimum, report the clamp.** If `p.timeout_ms > MAX_TIMEOUT_MS`, say so in the tool result, and on expiry return an explicit "killed at N s ceiling" rather than letting the caller infer it from a truncated stream. A silent clamp that a caller cannot detect is indistinguishable from a crash.

## Not covered by existing work

PR #935 ("feat(acp): agent timeout resilience") is adjacent and explicitly acknowledges the "600s max shell timeout" — it raises `DEFAULT_IDLE_TIMEOUT_SECS` 620 → 900 so the ACP idle timer stops firing during legitimate long tool calls. That fixes spurious *session* kills above the cap; it leaves the cap itself, and its silence, unchanged.

## Workaround for anyone hitting this

Launch through a `setsid` wrapper so the run lives in a session the group-kill cannot reach, and have the wrapper — not the agent — own the launch, since the detached run then outlives the MCP call that started it.

alanshurafa:
## Summary

`buzz-dev-mcp` capped every shell call at 600 s via an unconfigurable `.min(MAX_TIMEOUT_MS)`, and the clamp was invisible twice over: a caller asking for 3600 s was silently given 600 s (`timeout_ms` is an accepted tool parameter, so callers reasonably believe it was honored), and on expiry the process-group kill takes the child's own output buffer with it, so what the agent sees is a truncated log — not a timeout. The reporter lost four consecutive real agent runs to this, each dying at exactly 600 s with a 15-byte log, and the agent diagnosed its model backend as broken.

This implements both of the issue's asks:

1. **The ceiling is configurable**: `BUZZ_DEV_MCP_MAX_TIMEOUT_MS` (default unchanged, 600 000 ms), resolved once at `SharedState` construction. Unparseable or zero values fall back to the built-in default rather than lowering the ceiling — a misconfigured knob must never make every command die instantly (unit-tested).
2. **The clamp and the kill are reported in-band**, via the result's existing `notes: []` array (additive — no schema change, `timed_out`/`exit_code: 124` semantics untouched): a request above the ceiling gets "requested … exceeds the … ceiling; clamped (set BUZZ_DEV_MCP_MAX_TIMEOUT_MS to raise the ceiling)", and expiry adds "process group killed at the {N} ms timeout", so a killed job is attributable from inside the result.

Also updated: the `ShellParams` schema doc-comment no longer hard-codes "capped at 600000 ms", and — one comment line in `buzz-acp` — `DEFAULT_IDLE_TIMEOUT_SECS`'s doc now states the coupling this issue's scenario runs into next: the 900 s ACP idle timer is sized to the *default* shell ceiling, so operators raising the ceiling past ~900 s must raise `BUZZ_ACP_IDLE_TIMEOUT` to match (the knob already exists; past 7200 s, `BUZZ_ACP_MAX_TURN_DURATION` as well). Without that pairing the reporter's agent jobs would next die at 900 s. Related: #935 raised the idle default for exactly the adjacent reason; the issue itself draws the distinction.

### Related issue

Fixes #4638. Duplicate search at filing (2026-08-04): no open or closed PR addresses the shell-timeout ceiling (nearest match #4261 is buzz-acp `idle_timeout`, a different knob in a different crate).

### Testing

- `cargo test -p buzz-dev-mcp --lib -- --test-threads=1` — 121 passed, 0 failed (new: env-resolver cases incl. zero/garbage/empty fallbacks; clamp note asserted end-to-end with a lowered ceiling; the existing `timeout_fires` test now also asserts the kill-attribution note).
- Note on parallel runs: the suite is spawn-flaky on this Windows box *at the base commit too* (random shell-spawning tests intermittently get "failed to spawn shell" under parallel load — reproduced 1-in-3 on pristine base; single-threaded is stable). Pre-existing local-environment behavior, unrelated to this change.
- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all -- --check` — clean.
- Toolchain: Windows MSVC (as in #2638/#4214-#4216); base is 28ae6cd21 — `buzz-dev-mcp` is unchanged between that and current main.

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -21,8 +21,10 @@ use crate::filter::SubscriptionRule;
 /// Sized for slow turns where the agent may go silent on its outer ACP channel
 /// while running long sub-tools (e.g. a buzz-agent running another agent, or
 /// codex/claude doing multi-minute single tool calls). 900s gives 300s of
-/// breathing room above the 600s max shell timeout, so legitimate long-running
-/// tool calls don't race the idle deadline.
+/// breathing room above buzz-dev-mcp's default 600s shell-timeout ceiling, so
+/// legitimate long-running tool calls don't race the idle deadline. Operators
+/// who raise that ceiling (`BUZZ_DEV_MCP_MAX_TIMEOUT_MS`) must raise this
+/// timeout to match, or long tool calls die here instead.
 /// Override via `--idle-timeout` / `BUZZ_ACP_IDLE_TIMEOUT`.
 pub(crate) const DEFAULT_IDLE_TIMEOUT_SECS: u64 = 900;
 
--- a/crates/buzz-dev-mcp/src/shell.rs
+++ b/crates/buzz-dev-mcp/src/shell.rs
@@ -15,6 +15,8 @@ use tokio_util::sync::CancellationToken;
 
 const DEFAULT_TIMEOUT_MS: u64 = 120_000;
 const MAX_TIMEOUT_MS: u64 = 600_000;
+/// Operator override for the shell-timeout ceiling, in milliseconds (#4638).
+const MAX_TIMEOUT_ENV: &str = "BUZZ_DEV_MCP_MAX_TIMEOUT_MS";
 const MAX_COMMAND_BYTES: usize = 1_000_000;
 const CAPTURE_CAP: usize = 10 * 1024 * 1024;
 const MAX_BYTES: usize = 50 * 1024;
@@ -28,6 +30,10 @@ pub struct SharedState {
     pub shim: Shim,
     pub session_dir: TempDir,
     pub bootstrap_instructions: String,
+    /// Ceiling for `ShellParams::timeout_ms`, resolved once at startup from
+    /// `BUZZ_DEV_MCP_MAX_TIMEOUT_MS` (default 600 000 ms). The clamp itself is
+    /// applied — and reported — per call in [`run`].
+    pub max_timeout_ms: u64,
     /// The shell resolved at construction: `Ok((path, display_name))` when a shell
     /// is available, `Err(msg)` when none was found. Stored once so both the
     /// bootstrap hint and every `run()` call read the SAME resolution — no drift.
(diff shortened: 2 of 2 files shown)


alanshurafa:
@block/buzz-oss-team — could a maintainer please approve the workflow runs for this fork PR and do a code-owner review? The reachable checks (Semgrep OSS, zizmor, DCO) should pass; branch is based on 28ae6cd21 (buzz-dev-mcp unchanged vs current main). Thanks!

Issue #4833, opened 2026-08-05
Mention picker validates against local managed-agents store instead of relay membership (kind:39002); locally registering remote agents intercepts mentions


idk-its-unknown:
## Summary
Two related problems for agents that are relay members but run on a remote host (external buzz-acp instances, not spawned by the desktop app):

1. **False "not in channel" warning.** The @-mention picker warns that these agents are not in the channel even though the channel's kind:39002 member event lists their pubkeys and they respond to mentions normally. The picker appears to validate against the local managed-agents store rather than the relay's membership events (which the app already receives).

2. **Local registration intercepts mentions.** If such agents are added to the local store (`agents/managed-agents.json`) so they render in the Agents tab, mentioning them fails hard: "Could not start agent mention: <name>: This agent's configuration is missing — it may still be syncing or was deleted on another device." The app tries to launch a local copy instead of just publishing the mention; the message never reaches the relay. There is no way to register an agent as display-only/remote.

## Repro
1. Self-hosted relay; agent runs buzz-acp on a separate Linux host, admitted via `buzz-admin add-member` + a row in `channel_members` + regenerated 39000/39002 (`reconcile-channels`).
2. Desktop app (Windows, v0.5.x): mention the agent → warning "not in channel", but delivery works and the agent replies.
3. Add the agent to `managed-agents.json` (valid entry, `is_active: false`, no local provider config) → mentioning now errors and nothing is published.

## Expected
- Picker membership check sourced from the channel's 39002 event (or channel_members), not the local spawn registry.
- Mentioning an agent the app cannot start locally should still publish the mention; local auto-start should be an optimization, never a gate.

## Environment
Desktop v0.5.x on Windows 11; self-hosted relay (compose stack); agents via buzz-acp + external harness on Ubuntu, systemd-managed.

jaozolins:
Self-hosting Buzz on a VPS with server-side agents (buzz-acp under systemd). Hit this exact path: Desktop @ picker only surfaces local managed agents, so our remote agent either disappears from @ or loses proper Activity/ownership chrome. +1 — room membership (or relay-visible agents) should be enough to mention.

hondoaus:
Confirming the same underlying separation problem from an external **Hermes Agent** deployment.

We intentionally did **not** add the remote Hermes identity to Desktop's local managed-agent store. Instead we published a complete agent-signed kind:10100 record and kept the Hermes runtime on the VPS.

Observed behavior on current 0.5.x Windows Desktop:
[...]

Issue #4776, opened 2026-08-04
Desktop drops relay-directory agents from the mention picker when it doesn't manage them locally


kaalph:
## Setup

I run a self-hosted relay with five communities on it. The agents aren't started from the desktop app — they're containers on my own VM, one per agent, each running `buzz-acp` with its own key. They're managed with docker compose and a config repo, like everything else on that box. Each one publishes its kind:0 profile and a kind:10100 directory entry, and they're proper channel members.

Desktop 0.5.4 on Windows.

## What happens

None of them show up in the `@` autocomplete or in "Add people and agents". They're in the channel, the relay knows them, their profiles render fine on messages they send — but I can't mention them from the composer.

The gate is in `desktop/src/features/messages/lib/useMentions.ts:249`:

```ts
if (!isAgentIdentityInManagedList(candidate, managedAgentPubkeys)) {
  return;
}
```

`managedAgentPubkeys` only contains agents the desktop manages itself. Anything else gets dropped before `shouldHideAgentFromMentions` ever runs, so the finer policy below it never sees these candidates. Same helper is used for the add-picker path.

Took me a while to find, because the agents answer normally if I reply in a thread — it's only the composer that can't reach them.

## Why the provider path doesn't help me

I read `docs/remote-agents.md` after 0.5.4 landed. I get the model and I like it — the k8s provider is nice work. It doesn't fit here though, and not because of the substrate:

I don't want another binary on my desktop, and I don't want deploy credentials for that VM sitting on my laptop. The whole reason the agents live on the server is that the server is the thing that's administered, backed up and running 24/7. Deploying them is a compose file, not a click. So the desktop never holds their keys and never needs to.

That's the case I think isn't covered right now: agents deployed out-of-band, announced through the relay, with the desktop as a pure client.

## What I'm asking for

Treat directory agents as first-class in the composer. Either of these would do it for me:

1. Let `isAgentIdentityInManagedList` accept the already-computed invocable/directory set as a third argument, so directory-announced agents pass the gate and `shouldHideAgentFromMentions` keeps deciding the rest.
2. Or simpler: an agent that is a member of the current channel is always mentionable, regardless of who manages it.

I've been running (1) plus (2) as a local patch for about a week across five communities and haven't hit a downside — the existing hide rules still do their job, and nothing foreign leaks in because directory entries are scoped to the relay you're connected to.

[...]

PR #4839, opened 2026-08-05
fix(desktop): make owner-only relay agents mentionable by their owner

TheSeydiCharyyev:
## What is broken

`relayAgentIsSharedWithUser` handles two response modes: `anyone` and `allowlist`. There is a third one. `RespondToMode` is `"owner-only" | "allowlist" | "anyone"`, and `owner-only` falls through to `false`. It is `false` for every viewer, including the one person the agent answers to.

This is bigger than it looks, because `owner-only` is the server default. `types.ts` says the field omitted means `"owner-only"` (server default). The agent settings label that mode "Only me (default)" and print "Only you and your agents can send instructions." So the default mode promises the owner can instruct the agent, and the mention picker then hides that agent from the owner.

The hole only opens for an agent the desktop does not manage, for example one running as a container elsewhere. Managed agents are admitted by a separate path, so nobody who starts agents from the desktop would ever see this.

## The fix

`RelayAgent` has no owner field, so the owner has to come from profile metadata. `getMentionableAgentPubkeys` now takes the profile lookup and resolves ownership with `mergeOwnedAgentPubkeys` from `features/agents/knownAgentPubkeys.ts`. That helper is already documented as "Agent identities controlled by the current user", and the inbox already uses it for the same question. This path just never asked it.

Ownership is matched, never assumed. An agent whose owner cannot be resolved stays hidden. A caller that passes no profiles gets exactly today's behaviour.

Channel exactness does not change. The mention path goes through `relayAgentCanRespondInChannel`, which still requires the agent to be in that channel, so the owner reaches the agent only where the agent actually is. DMs and unresolved composer contexts use the `managed-only` scope and stay fail-closed.

## Tests

Two new tests in `agentAutocompleteEligibility.test.mjs`. Both fail on the current helper and pass after it: `# pass 21 / # fail 2` before, `# pass 23 / # fail 0` after.

The existing `owner-only` assertion in that file is untouched and still green. It calls `relayAgentIsSharedWithUser` without a viewer pubkey and expects `false`, which is still what happens: no viewer identity, no admission.

## One removal

`useMentions` built `sharedChannelIds` and passed it in. That call site only ever asks for the `channel` or `managed-only` scope, and neither one reads that set — only `community` does. So the parameter is now optional and `useMentions` no longer builds it. `useChannelsQuery` went with it, because it fed nothing else in the hook.

## Relation to #4913

This PR opened with two commits. The first removed a managed-list gate that ran ahead of the policy check. #4913 landed that behaviour and a lot more, so that commit is dropped and the branch is rebuilt on current `main`. What is left is the `owner-only` gap, which #4913 did not close.

## Scope

Three other call sites share this helper: `MembersSidebar`, `useNewMessageRecipients` and `ProjectsAgentPromptPage`. They are unchanged. Owner data is not available at those call sites without adding a profile query, and in `ProjectsAgentPromptPage` the profile query runs after this set is computed. Because `profiles` is optional, all three keep today's behaviour.

One note on trust, since ownership now decides what the picker shows. `ownerPubkey` is not self-asserted. It comes from the NIP-OA `auth` tag on the agent's kind:0, and `profile_valid_oa_owner_pubkey` verifies that tag against the owner before `get_users_batch` returns it, so a forged or stale marker resolves to no owner at all. An agent whose owner does not verify stays hidden, and the relay still enforces the real gate on the way in.

## Validation

- `pnpm --dir desktop test` — 4389 passed
- `pnpm --dir desktop check` — biome, file-size ratchet, px-text, pubkey-truncation
- `useMentions.ts` goes from 1000 lines to 991, so it stays under the ratchet

Refs #4776, #4833.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -110,6 +110,73 @@ test("relayAgentIsSharedWithUser: accepts allowlist agents for the current user"
   );
 });
 
+test("relayAgentIsSharedWithUser: an owner-only agent is invocable by its owner", () => {
+  const agent = {
+    respondTo: "owner-only",
+    respondToAllowlist: [],
+    channelIds: ["general"],
+  };
+  const sharedChannelIds = new Set(["general"]);
+
+  assert.equal(
+    relayAgentIsSharedWithUser(agent, sharedChannelIds, CURRENT_PUBKEY, true),
+    true,
+  );
+  // Someone else's owner-only agent stays out of reach.
+  assert.equal(
+    relayAgentIsSharedWithUser(agent, sharedChannelIds, CURRENT_PUBKEY, false),
+    false,
+  );
+  // Ownership is matched, never assumed: no viewer identity, no admission.
+  assert.equal(
+    relayAgentIsSharedWithUser(agent, sharedChannelIds, null, true),
+    false,
+  );
+});
+
+test("getMentionableAgentPubkeys: admits an owner-only relay agent for its owner", () => {
+  const relayAgents = [
+    {
+      pubkey: PUB_A,
+      respondTo: "owner-only",
+      respondToAllowlist: [],
+      channelIds: ["general"],
+    },
+    {
+      pubkey: PUB_B,
+      respondTo: "owner-only",
+      respondToAllowlist: [],
+      channelIds: ["general"],
+    },
+  ];
+  const call = (overrides) =>
+    getMentionableAgentPubkeys({
+      currentPubkey: CURRENT_PUBKEY,
+      eligibilityScope: { type: "channel", channelId: "general" },
+      managedAgentPubkeys: [],
+      relayAgents,
+      ...overrides,
+    });
+  const profiles = {
+    [PUB_A]: { ownerPubkey: CURRENT_PUBKEY.toUpperCase() },
+    [PUB_B]: { ownerPubkey: OTHER_OWNER_PUBKEY },
+  };
+
+  assert.deepEqual([...call({ profiles })], [PUB_A]);
+  // The owner still only reaches the agent in channels it actually joined.
+  assert.deepEqual(
+    [
+      ...call({
+        eligibilityScope: { type: "channel", channelId: "other" },
+        profiles,
+      }),
+    ],
+    [],
+  );
+  // No owner metadata, no admission -- unchanged from before.
+  assert.deepEqual([...call({})], []);
+});
+
 test("relayAgentCanRespondInChannel: requires exact channel membership and viewer access", () => {
   const agent = {
     respondTo: "allowlist",
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
@@ -1,3 +1,5 @@
+import { mergeOwnedAgentPubkeys } from "@/features/agents/knownAgentPubkeys";
+import type { UserProfileLookup } from "@/features/profile/lib/identity";
 import type { Channel, RelayAgent } from "@/shared/api/types";
 import { normalizePubkey } from "@/shared/lib/pubkey";
 
@@ -13,10 +15,20 @@ export function relayAgentIsSharedWithUser(
   agent: Pick<RelayAgent, "channelIds" | "respondTo" | "respondToAllowlist">,
   sharedChannelIds: ReadonlySet<string>,
   currentPubkey?: string | null,
+  isOwnedByViewer = false,
 ) {
   const normalizedCurrentPubkey = currentPubkey
     ? normalizePubkey(currentPubkey)
     : null;
+
+  // `owner-only` names exactly one person who may instruct the agent, so that
+  // person can invoke it. It is also the server default, so an agent created
+  // without an explicit mode used to be invocable by nobody at all. Ownership
+  // is matched, never assumed: an agent whose owner we cannot resolve stays
+  // hidden.
+  if (agent.respondTo === "owner-only") {
+    return isOwnedByViewer && normalizedCurrentPubkey !== null;
+  }
 
   if (agent.respondTo === "allowlist" && normalizedCurrentPubkey) {
     return agent.respondToAllowlist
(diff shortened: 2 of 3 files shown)


jaozolins:
Related: our server-hosted agent is in the channel and has a directory card, but Desktop still hides it from @ unless we strip ownership. Please stop treating relay-directory agents as non-mentionable for the owner. +1

TheSeydiCharyyev:
Confirmed, and this PR alone does not cover your case. `relayAgentIsSharedWithUser` only accepts `respondTo: "anyone"` or an `allowlist` containing you, so an `owner-only` agent is never invocable — including for its own owner. `RelayAgent` carries no owner field, which is why stripping ownership works around it.

Adding a second commit here: an owner-only relay agent counts as mentionable for its owner, taking ownership from the profile data this hook already reads.

kaalph:
I filed #4776 and I run the same shape of setup: self-hosted relay, six communities on it, 23 agents as containers on my own VM, none of them spawned by the desktop. So this hits me directly and I'm glad someone picked it up.

I've been running the equivalent of your part 1 as a local patch for about a week. Same conclusion — the candidate loop builds everything `shouldHideAgentFromMentions` needs and then the managed-list check throws it away first. My version kept the gate and passed the invocable set in as a third argument, which works but is clearly the more timid fix. Removing it and letting the policy decide is better; the policy is written for exactly those candidates.

Part 2 is the interesting one for me, because I hit the symptom and never found the cause. All my agents run with `respond_to: anyone`. I set that early on and wrote it off as "makes sense for a private community" — your explanation is that it's the workaround, and reading `relayAgentIsSharedWithUser` again I think you're right. Owner-only agents would have been invisible to me too and I'd never have known why. That part is worth more than the first half.

On #4833's second half: that's what I ran into. I hand-wrote entries into `managed-agents.json` so the agents would show up in the Agents tab, and then mentions died with a keyring error — the app tried to start a local copy of something that has no key on that machine, and the message never got published. I patched the start path to skip those agents. Your framing is the better one: local auto-start should be an optimization, never a gate.

Two things I can't speak to. I haven't built your branch, so this is a code read, not a test report. And I patch `MembersSidebar` too, since the add-picker uses the same helper — I saw you left it out on purpose and I think that's the right call for this PR, it's a different decision. If a small follow-up for it would be useful, I'm happy to write one.

jaozolins:
Same setup on my side: server agents, desktop as the client. Glad part 2 landed; that was the gap I was hitting. Thanks both.

TheSeydiCharyyev:
Part 1 is gone from this PR. #4913 landed on 5 Aug and does the same thing, plus exact-channel authorization, fail-closed DMs, cached-row reauthorization and forum coverage. Your local patch and my first commit are both superseded by it. The branch is rebuilt on current `main` and now carries only the owner-only half.

On `MembersSidebar`: #4913 already removed that gate. The add-picker now takes the policy-derived set from `getMentionableAgentPubkeys`, so the patch you were carrying is no longer needed there.

That surface is still not whole, though. It asks for the `community` scope and passes no profiles, so an `owner-only` agent stays hidden from its own owner in the add-picker, exactly as it was in mentions. I left it out here because the owner is not available at that call site without adding a profile query. It is free if you want it as a follow-up.

On the shape of the fix, one thing worth knowing before you read the diff: `relayAgentIsSharedWithUser` is not the only place it has to land. Mentions now run through `relayAgentCanRespondInChannel`, so ownership is threaded through both. Fix only the first and channel-scoped mentions stay broken while community-scoped ones work.

The keyring problem you described is a separate defect and deserves its own issue. You have the reproduction: hand-written `managed-agents.json` entries, the mention dies with a keyring error, the message is never published. Nothing in this PR touches the start path.

kaalph:
@Cynthia427 I checked your second caveat against our relay, and I think the ownership path is fine — just not the one you were looking at.

`buzz users get` isn't what the desktop uses. `useMentions` feeds its profile map from `useUsersBatchQuery` → `get_users_batch`, which queries `{"kinds":[0],"authors":[...]}` and hands the raw events to `users_batch_from_events`. Owner comes out of `profile_valid_oa_owner_pubkey` (`desktop/src-tauri/src/nostr_convert.rs:64`): it walks the profile event's tags, takes the 4-element `auth` tag, and verifies it with `nip_oa::verify_auth_tag` against the profile author. So the relay never has to return an `owner_pubkey` field for this — the client derives it from the agent's own kind:0 and checks the signature itself. The comment above it says as much: verified against the event author, so a forged marker can't turn a person into an agent.

That moves the question to: does the agent's kind:0 actually carry the tag?

On our relay, 41 of 43 distinct kind:0 events do. The two that don't are my own profile, twice — which is what you'd want, a human has no owner. Spot-checking one, the tag names our owner pubkey.

I can't verify the signatures with the tooling I have here, but there's decent indirect evidence they're good: most of our agents have no row in `relay_members` for the community they connect to. They get in through the NIP-OA branch of `check_relay_membership`, which runs `verify_auth_tag` and then checks whether the *owner* is a member. If those tags didn't verify, our agents wouldn't be connected at all — and they are, 25 of them.

So for a setup shaped like ours, `mergeOwnedAgentPubkeys` should resolve. Worth checking on yours whether the agents' kind:0 carries the `auth` tag at all — if the profile was published before attestation, or through a path that drops tags, you get no owner and the agent stays hidden, silently. That's a setup failure rather than a client bug, but it looks identical from the outside, and I'd have chased it for a while.

Your masking theory matches what we see, by the way. Our agents are seated as channel bots too, and we run `respond_to: anyone` on all of them — so we were double-hidden from the bug and found the symptom from the other direction entirely.

Same limitation as you: I didn't build the branch. This is our relay on v0.5.5 plus a read of the desktop source.

TheSeydiCharyyev:
Closing this. Two maintainer PRs have taken over the ground it stood on, and a third contributor PR now covers what was left.

The autocomplete half is gone: #5681 replaced `shouldHideAgentFromMentions` with `getAgentMentionAdmission`, so the `directoryAgentPubkeys` shape this PR added no longer exists.

The `owner-only` half turned out to sit on top of a deeper problem. The only writer of `kind:10100` in this repo is `cmd_set_add_policy` in `crates/buzz-cli/src/commands/channels.rs`, and it publishes `{"channel_add_policy": ...}` — no `respond_to`, no `channel_ids`. A `RelayAgent` built from that arrives with `respondTo: null` and `channelIds: []`, so `relayAgentIsSharedWithUser` cannot return true for any of them, whatever the policy is. #5546 adds the missing publisher and #5483 reworks the client-side eligibility, so both sides are already being handled.

I have left a note on #5483 about the one case I could not find covered there.

PR #4841, opened 2026-08-05
feat(desktop): searchable channel picker in agent/team deploy dialogs

sagestainsbys:
## Why

The "Deploy team to channel" and "Add agent to channel" dialogs render every candidate channel as a single plain `<select>` with no search. In communities with many channels this is hard to use.

To state the premise precisely (per review): **the candidate set is intentionally unchanged.** Non-member channels are valid deployment targets — deployment adds the agent via relay-authorized membership, and the existing copy says *accessible*, not *joined*. This PR is a search/UX feature, not a scoping fix.

## What

- New shared `ChannelPickerField` component (`desktop/src/features/channels/ui/`): a combobox search input plus listbox, fuzzy-matched with the channel browser's existing scorer (`scoreChannelMatch`), with arrow-key navigation and scroll-into-view.
- New `channelPickerOrdering` lib: channels the user has joined sort first (then alphabetical); query results rank by match score with ties keeping members-first order. Unit-tested.
- Both deploy dialogs (`AddTeamToChannelDialog`, `AddAgentToChannelDialog`) swap their `<select>` for the shared picker. The picker now owns default selection, replacing the dialogs' own init effects.

### Semantic choices (flagging explicitly for product review, per feedback)

1. **Joined-first default selection** — the picker defaults to the first *joined* channel rather than the first alphabetical one. Rationale: it's the likeliest target and makes the common case one keystroke shorter; non-member channels remain fully selectable and deployable.
2. **Selection follows the filter** — when the query filters out the selected channel, selection moves to the best visible match (the `<select>` this replaces could never hold an invisible selection). When the query matches **nothing**, the selection is **cleared** so the submit button disables — a hidden channel can never remain the deploy target.

Happy to change either behavior if the team prefers different semantics (e.g. selection cleared on any filter-out, requiring an explicit pick).

## Review feedback addressed

- **No-match state is now non-actionable**: an unmatched query clears the value; both dialogs derive `selectedChannel` from it, so Deploy/Add disables. Covered by Playwright in both dialogs.
- **Combobox semantics for assistive technology**: the input is now `role="combobox"` with `aria-expanded`, `aria-autocomplete="list"`, and `aria-activedescendant` pointing at the stable id of the selected option (`tabIndex={-1}` options, matching the `NewMessageScreen` active-descendant pattern). Arrow keys move the active descendant; covered by Playwright assertions.
- **Coverage extended to both dialogs**: the spec now drives the agent profile → channels → "Add to channel" flow as well as the team deploy flow, and asserts the submit-button state in the no-match case.

The settings → local-archive card has the same unsearchable `<select>`; left out deliberately to keep this diff small — easy follow-up with the new shared component.

## Testing

- `pnpm typecheck`, biome + file-size/px-text guards clean
- Full desktop unit suite: 4294/4294 pass (includes 8 `channelPickerOrdering` tests)
- Playwright `team-deploy-channel-picker.spec.ts` (integration project): 3/3 green locally — members-first ordering + auto-selection + aria-activedescendant tracking, search filtering, no-match state disabling submit in both dialogs
- Full Playwright integration project run locally against the revised head

Closest existing PR/issue: none found.

### DIFF
--- a/desktop/src/features/channels/ui/ChannelPickerField.tsx
+++ b/desktop/src/features/channels/ui/ChannelPickerField.tsx
@@ -34,6 +34,7 @@ export function ChannelPickerField({
   value,
 }: ChannelPickerFieldProps) {
   const [query, setQuery] = React.useState("");
+  const inputRef = React.useRef<HTMLInputElement>(null);
   const listRef = React.useRef<HTMLDivElement>(null);
 
   const visibleChannels = React.useMemo(
@@ -131,6 +132,7 @@ export function ChannelPickerField({
             }
           }}
           placeholder="Search channels"
+          ref={inputRef}
           role="combobox"
           spellCheck={false}
           type="text"
@@ -165,7 +167,10 @@ export function ChannelPickerField({
                 id={optionId(channel.id)}
                 key={channel.id}
                 onClick={() => onChange(channel.id)}
-                onMouseDown={(event) => event.preventDefault()}
+                onMouseDown={(event) => {
+                  event.preventDefault();
+                  inputRef.current?.focus();
+                }}
                 role="option"
                 tabIndex={-1}
                 type="button"
--- a/desktop/src/features/channels/ui/ChannelPickerField.tsx
+++ b/desktop/src/features/channels/ui/ChannelPickerField.tsx
@@ -165,6 +165,7 @@ export function ChannelPickerField({
                 id={optionId(channel.id)}
                 key={channel.id}
                 onClick={() => onChange(channel.id)}
+                onMouseDown={(event) => event.preventDefault()}
                 role="option"
                 tabIndex={-1}
                 type="button"
--- a/desktop/src/features/channels/ui/ChannelPickerField.tsx
+++ b/desktop/src/features/channels/ui/ChannelPickerField.tsx
@@ -34,13 +34,28 @@ export function ChannelPickerField({
   value,
 }: ChannelPickerFieldProps) {
   const [query, setQuery] = React.useState("");
-  const deferredQuery = React.useDeferredValue(query);
   const listRef = React.useRef<HTMLDivElement>(null);
 
   const visibleChannels = React.useMemo(
-    () => filterChannelsByQuery(channels, deferredQuery),
-    [channels, deferredQuery],
+    () => filterChannelsByQuery(channels, query),
+    [channels, query],
   );
+
+  function handleQueryChange(nextQuery: string) {
+    const nextVisibleChannels = filterChannelsByQuery(channels, nextQuery);
+    setQuery(nextQuery);
+
+    // Reconcile the parent value in the same input event. Deferring this to
+    // the effect below leaves the previous channel actionable for one render,
+    // so a no-match query could submit a hidden target before the effect runs.
+    if (nextVisibleChannels.length === 0) {
+      if (value !== "") {
+        onChange("");
+      }
+    } else if (!nextVisibleChannels.some((channel) => channel.id === value)) {
+      onChange(nextVisibleChannels[0].id);
+    }
+  }
 
   // Keep the selection visible: when the query filters out the selected
   // channel, move it to the best match — the `<select>` this replaces
@@ -105,7 +120,7 @@ export function ChannelPickerField({
           className="flex h-9 w-full rounded-md border border-input bg-background py-2 pl-9 pr-3 text-sm shadow-xs placeholder:text-muted-foreground focus-visible:outline-hidden focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50"
           disabled={disabled}
           id={inputId}
-          onChange={(event) => setQuery(event.target.value)}
+          onChange={(event) => handleQueryChange(event.target.value)}
           onKeyDown={(event) => {
             if (event.key === "ArrowDown") {
               event.preventDefault();
--- a/desktop/src/features/channels/ui/ChannelPickerField.tsx
+++ b/desktop/src/features/channels/ui/ChannelPickerField.tsx
@@ -1,0 +1,153 @@
+import { Check, Search } from "lucide-react";
+import * as React from "react";
+
+import { filterChannelsByQuery } from "@/features/channels/lib/channelPickerOrdering";
+import type { Channel } from "@/shared/api/types";
+
+type ChannelPickerFieldProps = {
+  /** Candidate channels, in display order (see `sortChannelsMembersFirst`). */
+  channels: Channel[];
+  disabled?: boolean;
+  /** Id for the search input, so a `<label htmlFor>` can target it. */
+  inputId: string;
+  onChange: (channelId: string) => void;
+  /** Id of the currently selected channel. */
+  value: string;
+};
+
+/**
+ * Searchable channel list for dialog forms. Replaces the plain `<select>`
+ * that listed every community channel in one dropdown — communities with
+ * many channels made that unusable. Fuzzy-matches with the channel
+ * browser's scorer and keeps the caller's (members-first) ordering.
+ */
+export function ChannelPickerField({
+  channels,
+  disabled = false,
+  inputId,
+  onChange,
+  value,
+}: ChannelPickerFieldProps) {
+  const [query, setQuery] = React.useState("");
+  const deferredQuery = React.useDeferredValue(query);
+  const listRef = React.useRef<HTMLDivElement>(null);
+
+  const visibleChannels = React.useMemo(
+    () => filterChannelsByQuery(channels, deferredQuery),
+    [channels, deferredQuery],
+  );
+
+  // Keep the selection visible: when the query filters out the selected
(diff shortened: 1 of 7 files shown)


desktop/src/features/channels/ui/ChannelPickerField.tsx
```diff
@@ -0,0 +1,153 @@
...
+    () => filterChannelsByQuery(channels, deferredQuery),
+    [channels, deferredQuery],
+  );
+
+  // Keep the selection visible: when the query filters out the selected
+  // channel, move it to the best match — the `<select>` this replaces
+  // could never hold an invisible selection. Once the selection is in the
+  // list, keep its row scrolled into view for keyboard navigation.
+  React.useEffect(() => {
+    if (visibleChannels.length === 0) {
```
wesbillman, line 68: **Blocker:** clear the selected value (or otherwise disable submission) when there are no visible matches. This early return leaves the last `channelId` selected while the UI says `No channels match …`; both parent dialogs derive `selectedChannel` from that hidden value, so their submit buttons remain enabled and will deploy/add to a channel the user can no longer see. That is especially dangerous here because typing already changes the target automatically. Please make the empty result state non-actionable and cover the submit state in the component/E2E test.

desktop/src/features/channels/ui/ChannelPickerField.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  const listId = `${inputId}-listbox`;
+
+  return (
+    <div className="space-y-1.5">
+      <div className="relative">
+        <Search className="pointer-events-none absolute left-3 top-1/2 h-4 w-4 -translate-y-1/2 text-muted-foreground/55" />
+        <input
+          aria-controls={listId}
```
wesbillman, line 117: **Blocker (accessibility):** this input handles ArrowUp/ArrowDown by changing the selected `role=option` while focus remains here, but it is exposed as a plain textbox with only `aria-controls`. A screen reader is not told that this controls a listbox or which option changed. Please implement the combobox/listbox pattern completely (`role=combobox`, appropriate expanded state, stable option ids, and `aria-activedescendant` for the active option), or remove the listbox/option semantics and use a simpler accessible interaction. The repository's `NewMessageScreen` demonstrates the active-descendant pattern. Add coverage for the resulting semantics.

wesbillman (changes requested):
Reviewing on Wes's behalf.

The searchable picker is a reasonable UX feature, but the premise needs correcting: non-member channels are intentionally valid deployment targets, not the defect being fixed. The old dialogs have included every non-DM, non-archived channel since their introduction, and this PR preserves exactly that candidate set. Downstream, deployment adds the agent via relay-authorized membership; open channels permit a non-member actor to add a non-elevated target (subject to agent policy), while private/elevated operations retain relay authorization. The existing copy also says accessible, not joined. This should therefore be framed and product-reviewed as a feature that adds search plus two semantic choices: joined-first default selection and query-driven target mutation.

I am requesting changes for two implementation defects in the new control: a no-match query leaves a hidden deployment target actionable, and the custom input/listbox interaction does not expose its keyboard-selected option to assistive technology. The Playwright coverage currently codifies the former empty state without checking the still-enabled submit action, and it only covers the team dialog despite changing both dialogs.

I did not rerun CI-equivalent suites locally. CI was still in progress at this exact head when reviewed.

sagestainsbys:
> This should therefore be framed and product-reviewed as a feature that adds search plus two semantic choices: joined-first default selection and query-driven target mutation.

🤖 Sage Stainsby's AI agent: Addressed at head `6fc45f2`: the title and description now frame this as a search UX feature, state that the candidate set is unchanged, and explicitly call out both semantic choices; the implementation findings are addressed in their inline threads with coverage for both dialogs.

desktop/src/features/channels/ui/ChannelPickerField.tsx
sagestainsbys: > Please make the empty result state non-actionable and cover the submit state in the component/E2E test.

🤖 Sage Stainsby's AI agent: Addressed at head `6fc45f2`: a no-match input synchronously clears `channelId`, making `selectedChannel` null and disabling submission in both dialogs; Playwright covers both dialogs and the immediate input-and-Deploy race.

desktop/src/features/channels/ui/ChannelPickerField.tsx
sagestainsbys: > Please implement the combobox/listbox pattern completely … and add coverage for the resulting semantics.

🤖 Sage Stainsby's AI agent: Addressed at head `6fc45f2`: the input now exposes the requested combobox state, stable option IDs, and `aria-activedescendant` tracking ArrowUp/ArrowDown selection; Playwright verifies those semantics and the no-match state.

wesbillman (changes requested):
Reviewing on Wes's behalf.

The prior no-match and active-descendant findings are substantially addressed, but the no-match safety guarantee still has a race because filtering is driven by a deferred query.

`ChannelPickerField` renders the input from `query`, but computes `visibleChannels` from `useDeferredValue(query)` and only clears the selected channel in an effect after the deferred render. Immediately after the user types a no-match query, the old visible list and old `value` can therefore remain current for at least one render; both parent dialogs still derive a non-null `selectedChannel`, leaving Deploy/Add enabled and able to submit the stale hidden target before the deferred update/effect lands. The Playwright assertions wait for the no-match UI, so they do not exercise this transition window.

Please make the submission-safety state synchronous with the actual input (the candidate set is small enough that deferring this fuzzy filter appears unnecessary), or synchronously gate the parent value/action from the raw query. Add a regression that changes to a no-match query and immediately attempts submission without first waiting for the empty-state render. The invariant advertised by this fix should hold at the input event boundary, not eventually.

sagestainsbys:
> Please make the submission-safety state synchronous with the actual input … and immediately attempt submission without first waiting for the empty-state render.

🤖 Sage Stainsby's AI agent: Addressed in `6fc45f2`: `useDeferredValue` was removed, selection is reconciled from the raw query in the same input event, and the regression dispatches the no-match input and clicks Deploy in the same browser task, confirming the stale target is not submitted.

wesbillman (commented):
Reviewing on Wes’s behalf.

I found no new actionable issues at head `e41e407ec754d6a29dce8b0ed9141d6b0246a015`. The two prior blockers are now addressed: selection is reconciled from the raw query in the input event, both parent actions derive their disabled state from that value, the combobox exposes a stable active descendant, and mouse selection preserves input focus. The focused Playwright spec passes all 3 scenarios at this exact head, including the same-task no-match submission regression and both dialogs. `git diff --check origin/main...HEAD` is clean, and a merge-tree simulation against current `origin/main` (`538e5e113`) completed without conflicts.

I am not approving or dismissing Wes’s existing changes-requested review; that decision remains with Wes. I did not rerun the full desktop suite because this review used the focused regression plus existing checks.

desktop/src/features/channels/ui/ChannelPickerField.tsx
```diff
@@ -0,0 +1,191 @@
...
+                className={
+                  isSelected
+                    ? "flex w-full items-center gap-2 bg-muted/60 px-3 py-2 text-left transition-colors duration-150 ease-out focus-visible:outline-hidden focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50"
+                    : "flex w-full items-center gap-2 px-3 py-2 text-left transition-colors duration-150 ease-out hover:bg-muted/40 focus-visible:outline-hidden focus-visible:ring-1 focus-visible:ring-ring disabled:cursor-not-allowed disabled:opacity-50"
+                }
+                disabled={disabled}
+                id={optionId(channel.id)}
+                key={channel.id}
+                onClick={() => onChange(channel.id)}
+                onMouseDown={(event) => event.preventDefault()}
```
wesbillman: **Accessibility/keyboard:** this only preserves focus when the input was already focused; it does not ensure the combobox owns focus after an option click. For example, focus the Role `<select>`, then click a channel option: cancelling the option's `mousedown` prevents the button from receiving focus, so focus remains on Role. The next ArrowUp/ArrowDown changes the role rather than moving through channels, while `aria-activedescendant` is attached to an unfocused combobox. The new Playwright check cannot catch this because `searchInput` is already focused before it clicks the option. Please keep an input ref and explicitly focus it for option pointer selection (or otherwise implement the listbox focus contract), and make the regression first focus another control before clicking an option.

wesbillman (commented):
Adversarial review at head `e41e407e`. I found one accessibility/keyboard interaction defect in the latest focus-preservation change. I did not rerun CI-equivalent suites locally; this finding follows directly from the event/focus path and the new test's precondition.

desktop/src/features/channels/ui/ChannelPickerField.tsx
sagestainsbys: thanks Wes, i've made some changes to address the accessibility issues; please take a look when you have a chance

PR #4664, opened 2026-08-04
fix(desktop): connect managed agents to the configured relay, not the normalized loopback identity

spkaplan:
## Summary

Two coupled defects in Desktop-managed agents on **loopback-addressed local communities**:

1. **Agents came up online but discovered 0 channels and sat idle** whenever the workspace relay was addressed as `localhost` (the desktop default and the seeded local-dev community). They only worked when the community was addressed as `127.0.0.1`.
2. **A permanent, un-clearable "Restart required" badge** once (1) is fixed — no restart could settle it.

**Root cause (shared).** The relay is multi-tenant, keyed by the literal request `Host`, so `localhost:3000` and `127.0.0.1:3000` are *distinct communities* with isolated channels/members. `buzz_core::relay::normalize_relay_url` intentionally folds all loopback spellings to `127.0.0.1` **for the runtime identity key only** — its own doc says:

> Connection code may retain the configured URL; this canonical form is for identity, receipts, status and deduplication.

But the spawn path reused that normalized `ManagedAgentRuntimeKey::relay_url` as the child's actual `BUZZ_RELAY_URL`, so an agent for a community joined on `localhost:3000` connected to the *empty* `127.0.0.1:3000` tenant: `subscribe` resolved 0 channels and the agent never woke.

## Fix

In `desktop/src-tauri/src/managed_agents/` (`runtime.rs`, `runtime_commands.rs`, `restore.rs`):

- **Connection:** connect the child on the **configured** relay URL at the spawn sites, keeping the normalized `ManagedAgentRuntimeKey` solely as the dedup identity (receipts, pid files, logs). Honors `normalize_relay_url`'s documented contract; no identity/dedup/receipts change.
- **Restart badge (corollary):** `needs_restart` compares a `SpawnConfigSnapshot` stamped at spawn against one recomputed live (`prospective_spawn_config_snapshot`), and the recompute reads the relay from the **normalized** runtime-map key. Once the connection fix flips the spawn's `effective_relay_url` to the configured spelling, stamping the snapshot with it (`localhost`) while the recompute uses the normalized key (`127.0.0.1`) makes them mismatch forever. Stamp `SpawnConfigInputs.relay_url` with the normalized `runtime_key.relay_url` so both sides compare the same spelling.

## Validation

Reproduced and verified end-to-end on a local self-hosted stack (relay on `localhost:3000`):

[...]

Issue #4147, opened 2026-08-01
Managed agents bind to a different relay community than the owner on loopback, discover 0 channels, and stay idle


2lost2bfound:
---
name: Bug report
about: Report a reproducible bug in Buzz
labels: bug
---

**Describe the bug**

Managed agents (desktop → `buzz-acp` → `buzz-agent`) can silently connect to a **different relay community** than the owner's desktop in local loopback development, discover **0 channels**, and sit idle forever — never replying to mentions.

Root cause is a loopback-host equivalence mismatch between two normalizers:

- `buzz_core::relay::normalize_relay_url` (`crates/buzz-core/src/relay.rs`) is the **sole** normalizer for `(agent, relay)` runtime identity. It folds **all** loopback spellings (`localhost`, `127.*`, `[::1]`) to `127.0.0.1`.
- The desktop passes the **canonicalized** URL to the spawned agent: `ManagedAgentRuntimeKey::new(...)` calls `normalize_relay_url`, and `start_managed_agent_process` sets `command.env("BUZZ_RELAY_URL", <canonical url>)`. So the agent connects to `ws://127.0.0.1:3000`.
- The relay binds HTTP/WS requests to a community via the `Host` header through `buzz_core::tenant::normalize_host` (`crates/buzz-core/src/tenant.rs`), which lowercases/strips default ports but **does not** collapse `localhost` vs `127.0.0.1`. Those are deliberately distinct hosts.
- `scripts/seed-local-community.sh` seeds **four** community rows for loopback dev (`localhost`, `127.0.0.1`, `localhost:3000`, `127.0.0.1:3000`) — so the empty `127.0.0.1:3000` row exists and binds successfully.

Net effect: the owner desktop uses `ws://localhost:3000` → community `localhost:3000` (where channels/members/events live), while the agent uses `ws://127.0.0.1:3000` → the **separate, empty** community `127.0.0.1:3000`. The agent's channel-discovery query (kind 39002 with `#p` = agent pubkey) returns 0 events, so the startup log prints `discovered 0 channel(s)` / `no channel subscriptions resolved — agent will sit idle`, and the agent never responds to owner mentions.

The auth seam (`buzz-auth/src/nip98.rs::normalize_url`, NIP-42) deliberately treats loopback spellings as distinct hosts, so collapsing `normalize_host` on the relay side is **not** a safe fix — it would break NIP-98/NIP-42 signature validation on one side. The correct fix is to make the agent land on the **same host string** the owner uses.

[...]

PR #4859, opened 2026-08-05
fix(desktop): connect managed agents with the configured relay URL, not the canonical identity URL

### LINKED ISSUE #2444: [Bug] Desktop 0.4.23 spawns managed agents on ws://127.0.0.1:3000 while the community uses ws://localhost:3000 - agents silently stop responding

### LINKED ISSUE #3283: Local dev: loopback hosts are seeded as separate communities, so desktop-managed agents silently discover 0 channels

anilkishan:
## Summary

Desktop-managed agents can silently connect to a **different relay community** than the owner's desktop, discover **0 channels**, and sit idle forever - never replying to mentions.

Root cause: two normalizers deliberately disagree about host equivalence, and the spawn path mixes them up.

- `buzz_core::relay::normalize_relay_url` (the **identity** normalizer) folds all loopback spellings to `127.0.0.1` to build the stable `(pubkey, relay_url)` runtime key. Its doc comment explicitly says connection code should retain the configured URL.
- `buzz_core::tenant::normalize_host` (the **tenancy** normalizer) does **not** collapse `localhost` vs `127.0.0.1` - distinct spellings are distinct communities, and the relay fails closed on unmapped hosts.

`spawn_agent_child` violates the first normalizer's contract: it sets the child's `BUZZ_RELAY_URL` from the canonical `runtime_key.relay_url` instead of the configured URL. So an agent configured against `ws://localhost:3000` is spawned against `ws://127.0.0.1:3000`, binds to the wrong (often empty) community, and goes deaf.

The fix keeps the two roles separate, per `normalize_relay_url`'s documented intent:

- **Connection:** `spawn_agent_child` now uses the configured relay URL (trimmed) for the child's `BUZZ_RELAY_URL`. All three call sites (`start_managed_agent_process`, `start_pair`, launch restore) already have the configured URL in scope and now pass it through.
- **Identity & drift:** the runtime key and the spawn-config snapshot still use the canonical `key.relay_url`. The restart-drift check recomputes the prospective snapshot with `key.relay_url`, so the spawn-time stamp must agree - otherwise agents restart-loop whenever the configured spelling differs from the canonical one.

No identity migration: existing `(pubkey, relay_url)` runtime keys are unchanged.

### Follow-up: restart-path persistence

Restart paths that only had the canonical runtime key were still dialing `127.0.0.1` after a healthy `localhost` spawn. The exact connection URL each child was dialed with is now stamped on the pair runtime at spawn and persisted on the receipt (`connectRelayUrl`; old receipts remain valid when the field is absent). The two backend restart orchestrations (global-config and ACP-install) now reuse each live pair's stamped connection URL. Interactive community actions already submit the configured community URL on current main, and live status rows now echo the pair's actual connection spelling through `requestedRelayUrl`. Identity, receipts-as-keys, and the spawn-config snapshot stay on the canonical form. Debug output masks the stored connection URL, matching the spawn-snapshot redaction policy, since relay URLs may carry query tokens.

### Related issues

Fixes #2444, #3033, #3505, #4147.

Does **not** close:

- #3283 - the launcher half (suggested fix 3) is this PR; the seed script still inserts four isolated loopback tenants. That contradiction stays open.
- #2641 - mixed report (membership, owner-only, publish path). Same idle log line, different bugs still in that thread.
- #4345 - different layer. It changes `normalize_relay_url` to fold loopback to `localhost`, which rewrites runtime keys. Leaving it for maintainers; this PR does not compete with it.

### Testing

- New unit test `agent_connection_preserves_loopback_authority` covering the connection-URL helper.
- `just desktop-tauri-test` (full `cargo test --workspace` in `desktop/src-tauri`, 2,200+ tests) passes.
- `just desktop-tauri-clippy` and `cargo fmt --check` clean.
- Follow-up: receipt back-compat, connection-URL validation, restart-target selection, and Debug-mask sentinel tests added; full desktop/Tauri workspace suite (2,535 tests) green after rebase onto current main.
- This exact fix has been running in production on a self-hosted macOS deployment (relay with host-derived multi-community rows) from v0.4.26 through v0.5.4: agents connect to the correct community, no restart loops, and the restart-drift badge behaves.

### Prior art

Several independent PRs diagnosed the same spawn-vs-tenancy split. salman1993 asked to consolidate on this PR and fold useful pieces in. What landed, and what we left with its author:

| PR | Author | What we took |
|----|--------|--------------|
| #2446 | [lewiswang0516](https://github.com/lewiswang0516) | First write-up, and the `requestedRelayUrl` status-row complement so rows correlate across spellings. |
| #2944 | [ahmetkca](https://github.com/ahmetkca) | The restart-persistence design this follow-up implements: keep the configured URL, because the global-config and ACP-install restarts otherwise lose it. Also the identity-vs-dial test framing and the NIP-98 reason not to fold hosts on the relay. |
| #3484 | [JHSeo-git](https://github.com/JHSeo-git) | Same diagnosis and a `child_connect_relay_url` seam. Not lifted - current `main` already restructured that area. Credit for the write-up; no code taken. |
| #3941 | [D12Labs](https://github.com/D12Labs) | Clear "do not close #3283 on a spawn-path PR." We followed that. |
| #4345 | [yougeqiu](https://github.com/yougeqiu) | Different layer (flip the identity normalizer to `localhost`). Not superseded; not folded in. |
| #4664 | [spkaplan](https://github.com/spkaplan) | Restart-badge corollary: snapshot must stay on the canonical spelling or the badge never clears. We already had that split; keeping it. |
| #5200 | [ZEDYB](https://github.com/ZEDYB) | Identity/connection split as a single decision point (`spawn_relay_roles` framing). The restart-target helper in this follow-up is that idea. |

If I missed a piece that should be in the diff, say so on this PR - happy to take it.

### DIFF
--- a/desktop/src-tauri/src/managed_agents/runtime_commands.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime_commands.rs
@@ -403,7 +403,10 @@ async fn probe_agent_relay_access(
     let key = ManagedAgentRuntimeKey::new(record.pubkey.clone(), &requested_relay_url)?;
     let keys = nostr::Keys::parse(record.private_key_nsec.trim())
         .map_err(|error| format!("invalid managed-agent key: {error}"))?;
-    let api_base = crate::relay::relay_http_base_url(&key.relay_url);
+    // Probe the community the user actually configured: relay tenancy is
+    // host-derived, so the canonical key spelling can resolve to a different
+    // (or unmapped) community than the requested URL.
+    let api_base = crate::relay::relay_http_base_url(&requested_relay_url);
     tokio::time::timeout(
         std::time::Duration::from_secs(10),
         crate::relay::query_relay_at_with_keys(
@@ -504,7 +507,10 @@ pub async fn reconcile_managed_agent_runtimes(
                 Ok((record, key, requested)) => {
                     match start_pair(
                         record.pubkey.clone(),
-                        key.relay_url.clone(),
+                        // Start with the requested spelling so the spawned
+                        // child connects to the community that was probed.
+                        // start_pair re-derives the same canonical key.
+                        requested.clone(),
                         true,
                         Some(&record.updated_at),
                         app.clone(),
--- a/desktop/src-tauri/src/managed_agents/runtime_commands.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime_commands.rs
@@ -277,6 +277,12 @@ fn start_pair(
         .get_mut(&key)
         .is_some_and(|runtime| runtime.child.try_wait().ok().flatten().is_none())
     {
+        // Reuse is only valid when the live pair dials the requested URL:
+        // the canonical key folds host spellings, and a cross-spelling reuse
+        // would falsely report the requested tenant as started.
+        if let Some(runtime) = runtimes.get(&key) {
+            super::ensure_pair_connection_matches(runtime, &relay_url)?;
+        }
         let status = status_for(&app, record, &key, runtimes.get(&key), None);
         return Ok(status);
     }
--- a/desktop/src-tauri/src/managed_agents/runtime_commands.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime_commands.rs
@@ -283,7 +283,7 @@ fn start_pair(
         .lock()
         .ok()
         .map(|keys| keys.public_key().to_hex());
-    let mut process = spawn_agent_child(&app, record, &key.relay_url, lazy, owner.as_deref())?;
+    let mut process = spawn_agent_child(&app, record, &relay_url, lazy, owner.as_deref())?;
     let now = crate::util::now_iso();
     let receipt = ManagedAgentRuntimeReceipt {
         key: key.clone(),
--- a/desktop/src-tauri/src/managed_agents/runtime_commands.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime_commands.rs
@@ -60,7 +60,12 @@ fn status_for_with(
     ManagedAgentRuntimeStatus {
         pubkey: key.pubkey.clone(),
         relay_url: key.relay_url.clone(),
-        requested_relay_url,
+        // Callers that know the exact descriptor URL (reconcile) pass it in;
+        // otherwise fall back to the live pair's stamped connection URL so
+        // frontend start/restart actions re-dial the configured spelling
+        // instead of the canonical form.
+        requested_relay_url: requested_relay_url
+            .or_else(|| runtime.map(|runtime| runtime.connect_relay_url.clone())),
         local_setup,
         lifecycle: runtime
             .map(|runtime| runtime.lifecycle.clone())
@@ -290,6 +295,7 @@ fn start_pair(
         pid: process.child.id(),
         desktop_instance_id: current_instance_id(&app),
         started_at: now.clone(),
+        connect_relay_url: Some(process.connect_relay_url.clone()),
     };
     if let Err(error) = write_agent_runtime_receipt(&app, &receipt) {
         let _ = terminate_process(process.child.id());
--- a/desktop/src-tauri/src/commands/agent_discovery.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery.rs
@@ -562,9 +562,7 @@ async fn restart_single_agent_after_install(
         if record.backend != BackendKind::Local {
             return Err(format!("agent {pubkey_owned} is no longer a local agent"));
         }
-        // Collect the restart dial targets (each pair's configured connection
-        // URL, not the canonical key spelling) BEFORE the stop below drops the
-        // pairs — and their URLs — from the runtimes map.
+        // Dial targets (configured spellings) must be collected before the stop.
         let restart_targets =
             crate::managed_agents::managed_agent_restart_targets(&runtimes, &pubkey_owned);
         if restart_targets.is_empty() {
--- a/desktop/src-tauri/src/commands/agent_discovery.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery.rs
@@ -562,9 +562,12 @@ async fn restart_single_agent_after_install(
         if record.backend != BackendKind::Local {
             return Err(format!("agent {pubkey_owned} is no longer a local agent"));
         }
-        let runtime_keys =
-            crate::managed_agents::managed_agent_runtime_keys(&runtimes, &pubkey_owned);
-        if runtime_keys.is_empty() {
+        // Collect the restart dial targets (each pair's configured connection
+        // URL, not the canonical key spelling) BEFORE the stop below drops the
+        // pairs — and their URLs — from the runtimes map.
+        let restart_targets =
+            crate::managed_agents::managed_agent_restart_targets(&runtimes, &pubkey_owned);
+        if restart_targets.is_empty() {
             return Err(format!(
                 "agent {pubkey_owned} no longer has a live pair runtime after sync"
             ));
@@ -606,12 +609,12 @@ async fn restart_single_agent_after_install(
         stop_managed_agent_process(&app_for_stop, record_mut, &mut runtimes)?;
         save_managed_agents(&app_for_stop, &records)?;
 
-        Ok(runtime_keys)
+        Ok(restart_targets)
     })
     .await;
 
-    let runtime_keys = match stop_result {
-        Ok(Ok(runtime_keys)) => runtime_keys,
+    let relay_urls = match stop_result {
+        Ok(Ok(restart_targets)) => restart_targets,
         Ok(Err(e)) => {
             eprintln!("buzz-desktop: install_acp_runtime: skipping restart of {pubkey}: {e}");
             return InstallRestartOutcome::Skipped;
@@ -624,7 +627,6 @@ async fn restart_single_agent_after_install(
         }
     };
 
-    let relay_urls: Vec<_> = runtime_keys.into_iter().map(|key| key.relay_url).collect();
     let state = app.state::<AppState>();
     match super::agents::start_local_agent_pairs_with_preflight(app, &state, pubkey, &relay_urls)
         .await
--- a/desktop/src-tauri/src/commands/global_agent_config.rs
+++ b/desktop/src-tauri/src/commands/global_agent_config.rs
@@ -292,9 +292,12 @@ async fn restart_local_agent_on_config_change(
         if record.backend != BackendKind::Local {
             return Err(format!("agent {pubkey_owned} is no longer a local agent"));
         }
-        let runtime_keys =
-            crate::managed_agents::managed_agent_runtime_keys(&runtimes, &pubkey_owned);
-        if runtime_keys.is_empty() {
+        // Collect the restart dial targets (each pair's configured connection
+        // URL, not the canonical key spelling) BEFORE the stop below drops the
+        // pairs — and their URLs — from the runtimes map.
+        let restart_targets =
+            crate::managed_agents::managed_agent_restart_targets(&runtimes, &pubkey_owned);
+        if restart_targets.is_empty() {
             return Err(format!(
                 "agent {pubkey_owned} no longer has a live pair runtime after sync"
             ));
@@ -327,12 +330,12 @@ async fn restart_local_agent_on_config_change(
         stop_managed_agent_process(&app_for_stop, record_mut, &mut runtimes)?;
         save_managed_agents(&app_for_stop, &records)?;
 
-        Ok(runtime_keys)
+        Ok(restart_targets)
     })
     .await;
 
-    let runtime_keys = match stop_result {
-        Ok(Ok(runtime_keys)) => runtime_keys,
+    let relay_urls = match stop_result {
+        Ok(Ok(restart_targets)) => restart_targets,
         Ok(Err(e)) => {
             eprintln!("buzz-desktop: set_global_agent_config: skipping restart of {pubkey}: {e}");
             return RestartOutcome::Skipped;
@@ -345,7 +348,6 @@ async fn restart_local_agent_on_config_change(
         }
     };
 
-    let relay_urls: Vec<_> = runtime_keys.into_iter().map(|key| key.relay_url).collect();
     use tauri::Manager;
     let state = app.state::<AppState>();
     match super::agents::start_local_agent_pairs_with_preflight(app, &state, pubkey, &relay_urls)
--- a/desktop/src-tauri/src/managed_agents/agent_env.rs
+++ b/desktop/src-tauri/src/managed_agents/agent_env.rs
@@ -137,6 +137,14 @@ pub(crate) fn parse_agent_env_lines(raw: &str) -> Vec<(&str, &str)> {
         .collect()
 }
 
+pub(super) fn child_rust_log_filter() -> String {
+    match std::env::var("RUST_LOG") {
+        Ok(existing) if existing.contains("buzz_acp") => existing,
+        Ok(existing) if !existing.trim().is_empty() => format!("{existing},buzz_acp=info"),
+        _ => "buzz_acp=info".to_string(),
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::{
--- a/desktop/src-tauri/src/managed_agents/process_lifecycle.rs
+++ b/desktop/src-tauri/src/managed_agents/process_lifecycle.rs
@@ -133,6 +133,7 @@ pub fn taskkill_tree(pid: u32) -> Result<(), String> {
 pub fn finish_spawn(
     child: std::process::Child,
     log_path: std::path::PathBuf,
+    connect_relay_url: String,
     spawn_config: super::spawn_snapshot::SpawnConfigSnapshot,
     setup_mode: bool,
     adapter_availability: Option<super::AcpAvailabilityStatus>,
@@ -149,6 +150,7 @@ pub fn finish_spawn(
     super::ManagedAgentProcess {
         child,
         log_path,
+        connect_relay_url,
         spawn_config,
         setup_mode,
         adapter_availability,
--- a/desktop/src-tauri/src/managed_agents/restore.rs
+++ b/desktop/src-tauri/src/managed_agents/restore.rs
@@ -335,7 +335,7 @@ pub async fn restore_managed_agents_on_launch(
                                             spawn_agent_child(
                                                 app,
                                                 record,
-                                                &key.relay_url,
+                                                &relay_url,
                                                 true,
                                                 owner_hex_ref,
                                             )
--- a/desktop/src-tauri/src/managed_agents/restore.rs
+++ b/desktop/src-tauri/src/managed_agents/restore.rs
@@ -308,21 +308,23 @@ pub async fn restore_managed_agents_on_launch(
                             Ok(key) => {
                                 // F2: if a concurrent startup reconcile already
                                 // tracked a live child for this exact pair during
-                                // the Phase A window, leave it alone. Mirrors the
+                                // the Phase A window, leave it alone - but only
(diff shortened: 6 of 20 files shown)


Chessing234 (commented):
connecting with the configured relay url while identity/snapshot stay on canonical key.relay_url matches the tenancy vs identity split. any leftover call site that still builds the connect URL from the identity spelling?

anilkishan:
Good question — swept the desktop crate for `key.relay_url` / `runtime_key.relay_url` flowing into connection paths. Three leftovers turned up:

1. **Relay-access probe** — `probe_agent_relay_access` (`runtime_commands.rs`) built its HTTP `api_base` from `key.relay_url`, so the probe could query a different (or unmapped) community than the one being configured. Fixed in 5e028aff8: probe with `requested_relay_url` (already in scope).

2. **Post-probe start** — the success arm passed `key.relay_url` into `start_pair`, so the child spawned right after a successful probe still connected with the canonical spelling. Fixed in 5e028aff8: pass the requested spelling through; `start_pair` re-derives the same canonical key, so identity is unchanged.

3. **Restart orchestration** — `set_global_agent_config` and `install_acp_runtime` restarts rebuild start URLs from live runtime keys (`.map(|key| key.relay_url)`). By that point the configured spelling is gone — the runtime map only holds the canonical key. The proper fix is to persist the connection URL alongside the canonical key in the pair runtime/receipt and restart with it. That grows the diff (receipt schema + both restart paths), so I'd propose it as a follow-up unless you'd rather have it here. These two paths are no worse after this PR than before it: they restart with the canonical spelling, which is what every spawn did previously.

Identity and bookkeeping uses of `key.relay_url` (stop/teardown, receipts, status rows, the drift-check snapshot) are left canonical deliberately — that side of the split is working as intended.

salman1993:
🤖 The identity/connection split is the right fix, and I’d like to use this as the canonical PR for the overlapping reports. One blocker remains: the current diff loses the configured connection URL during global-config and ACP-install restarts, causing an initially healthy localhost agent to restart against 127.0.0.1 and become idle again.

Please retain the original connection URL in the pair runtime and persisted receipt, use it for every restart path, and add backward-compatible receipt plus restart-target tests. The canonical URL should remain the runtime key and spawn-snapshot input. Please also rebase onto current main and run the full desktop/Tauri checks. Once that is addressed, I’m comfortable approving this and closing the duplicate implementations.

it could be worth looking at:
- https://github.com/block/buzz/pull/3484 
- https://github.com/block/buzz/pull/5200

to pick out some of the pieces from those PRs which seem to be a duplicate

anilkishan:
Done. Thank you for the review and for consolidating the reports here.

The follow-up commit implements the blocker as scoped:

- The exact connection URL each child is dialed with is stamped on the pair runtime at spawn and persisted on the receipt (`connectRelayUrl`). Old receipts remain valid when the field is absent; every live runtime carries the stamped URL.
- The two backend restart orchestrations (`set_global_agent_config`, `install_acp_runtime`) collect each live pair's stamped connection URL *before* the stop and dial exactly those targets. Interactive community actions already submit the configured community URL on current main, and live status rows now echo the pair's actual connection spelling through `requestedRelayUrl` for row correlation.
- Receipt validation now requires a present connection URL to canonicalize back to the receipt's own pair key. A foreign or unparseable URL fails validation.
- Process Debug output masks the stored connection URL (relay URLs may carry query tokens), matching the existing spawn-snapshot redaction policy and pinned by the Debug sentinel test.
- The canonical URL remains the runtime key and spawn-snapshot input, per your note.
- Tests: pre-field receipt back-compat, canonicalization validation both ways, restart-target selection across multiple pairs and agents, serde roundtrip.
- Rebased onto current main; full desktop/Tauri suite, clippy, and fmt pass.

Pieces folded in from the overlapping PRs are credited in the commit trailers and the prior-art section below, from #2446, #2944, and #5200 in particular.

wolfyy970 (approved):
Re-reviewed at exact head `8f1ffc0c65`. The restart blocker is closed: canonical URLs remain pair identity and spawn-snapshot input, while the actual dial URL is stamped on the live process and backward-compatible receipt and reused by both automatic restart paths. Foreign receipt URLs fail validation, and old receipts still load.

I reran the focused receipt, multi-pair restart-target, and loopback connection tests at this head: 14 passed. Tauri formatting and `git diff --check` are clean. Approved.

desktop/src-tauri/src/managed_agents/runtime_commands.rs
salman1993, line 1: **[P1] Canonical-key collisions can reuse an agent connected to the wrong community.**  
   [`start_pair`](https://github.com/block/buzz/blob/8f1ffc0c65077c5a956105c732726b7b1a0ed153/desktop/src-tauri/src/managed_agents/runtime_commands.rs#L276-L281) returns any live runtime matching the canonical key without checking its new `connect_relay_url`. If an agent is running against `ws://127.0.0.1:3000` and reconciliation requests `ws://localhost:3000`, the probe reaches the new tenant, but `start_pair` returns the old process because both spellings share a key. [The status is then overwritten](https://github.com/block/buzz/blob/8f1ffc0c65077c5a956105c732726b7b1a0ed153/desktop/src-tauri/src/managed_agents/runtime_commands.rs#L524-L526) to claim the `localhost` request succeeded. The agent remains connected to the previous tenant and reconciliation stops retrying. 

It'd be good to fail with an explicit connection-target conflict when the canonical key is already occupied by a runtime connected to a different tenant.

salman1993:
`just desktop-check` is failing in CI. The file-size ratchet is failing for five files, especially runtime.rs, runtime/tests.rs, and types.rs. Those files must be shortened or split; don’t raise the limits. The cap prevents “god files” from continually growing and becoming harder to understand, review, test, and maintain.

anilkishan:
Both items addressed:

- **Connection-target conflict (P1):** all three canonical-key reuse sites (`start_pair`, `start_managed_agent_process`, and launch restore's already-live check) now compare the live pair's stamped connection URL with the requested one. The command paths fail with an explicit connection-target-conflict error and restore records the conflict as a failed outcome, instead of reusing the child and reporting the requested tenant as started. The error intentionally omits both URLs (they may carry query tokens and the string lands in `last_error`); tests pin the non-disclosure, the allowed same-spelling reuse, and the restore decision.
- **File sizes:** all five files flagged by `desktop-check` are back at or under their limits by moving code, not raising caps: the connection-URL tests live in a new `runtime/connect_url_tests.rs`, shared fixtures in `runtime/test_fixtures.rs`, the process Debug impl and `connection_relay_url` beside the receipt validation in `runtime/process.rs`, `child_rust_log_filter` in `agent_env.rs`, and `persona_drift_state` in `runtime/metadata.rs`. No behavior change in the moves.

Full desktop/Tauri workspace suite (2,538 passed, 0 failed), clippy with warnings denied, fmt, and the full `desktop-check` gate (file sizes included) are green on the new head.

wesbillman (commented):
**P1 — live status still aliases distinct loopback tenants, so the UI can stop the wrong agent pair.**

This branch correctly exposes the actual dial spelling as `requestedRelayUrl`, but the frontend lookup still accepts the canonical identity URL first/finally (`desktop/src/features/agents/managedAgentRuntimeStatus.ts:95-112`). For a live row dialed to `ws://localhost:3000`, the row is now:

- `relayUrl = ws://127.0.0.1:3000` (canonical identity)
- `requestedRelayUrl = ws://localhost:3000` (actual tenant)

When rendering the distinct configured community `ws://127.0.0.1:3000`, `findManagedAgentRuntime` matches that same row because `runtime.relayUrl === relayUrl` (line 109), even though `requestedRelayUrl` proves the child is connected to `localhost`. The card therefore reports the agent as running in the wrong community; its Stop/Restart action uses the shared canonical key and can terminate the `localhost` child. That defeats the PR's new conflict guard at the user-facing boundary and is especially dangerous because the two hosts are deliberately distinct tenants.

Smallest safe remedy: when `requestedRelayUrl` is present, make it authoritative for community matching and do **not** fall back to canonical identity matching; reserve the canonical fallback for legacy/status rows where the requested URL is absent. Add a regression with simultaneous configured `localhost` and `127.0.0.1` communities proving only the actually dialed card resolves a runtime and an action on the other card cannot target it. If tenant-equivalent spelling variants (case/default port/FQDN dot) must match, use tenancy-authority equivalence—not `normalize_relay_url`, which intentionally collapses these two loopback tenants.

wesbillman (changes requested):

Requesting changes at exact head `4f759f915414781d8bcafb857c457f3c31bf532a` for two correctness defects in the identity/connection split:

1. **P1: live status aliases distinct loopback tenants in the frontend.** `findManagedAgentRuntime` accepts `runtime.relayUrl === relayUrl` and canonical fallback even when `requestedRelayUrl` is present (`desktop/src/features/agents/managedAgentRuntimeStatus.ts:95-112`). A child dialed to `ws://localhost:3000` therefore appears live on the distinct `ws://127.0.0.1:3000` community because both share the canonical runtime key. The wrong card can issue Stop/Restart against that key and terminate the localhost child. Make `requestedRelayUrl` authoritative when present; use canonical fallback only for legacy rows without it. Add a two-spelling regression at this UI/action boundary.

2. **P1: the new reuse guard rejects harmless spellings of the same connection target.** `ensure_pair_connection_matches` compares trimmed strings byte-for-byte (`desktop/src-tauri/src/managed_agents/runtime/process.rs:487-501`). DNS host case, an explicit default port, a root slash, and an FQDN root dot are connection-equivalent and resolve to the same relay tenant, but this guard reports a `connection-target conflict`. The tenancy contract explicitly folds host case, default ports, and the root dot (`crates/buzz-core/src/tenant.rs:100-137`); URL parsing also folds the root slash/default port. This can strand reconcile/start after harmless configuration formatting drift. Compare normalized connection targets while preserving tenancy-significant host spelling such as `localhost` versus `127.0.0.1`; add equality regressions for DNS case/default port/root slash/FQDN dot and retain the loopback-inequality regression.

The core fix is directionally correct: actual dial URLs are now persisted and reused across restart paths while canonical URLs remain identity keys. Current checks are green, but neither semantic boundary above is covered.

anilkishan:
Both P1s addressed:

- **Reuse guard equivalence:** `ensure_pair_connection_matches` now compares connection targets parsed with `url::Url`: lowercased scheme, parsed-host normalization (ASCII lowercase, single FQDN root dot stripped), a port kept only when it differs from the scheme's own default, a root-slash-folded path, and the query preserved verbatim as its own field. Harmless formatting drift no longer reads as a conflict, while tenancy-significant differences stay conflicts: `localhost`, `127.0.0.1`, and `[::1]` are three distinct hosts, and `ws` vs `wss`, non-default ports (including the other scheme's default), paths, and query strings stay distinct. Unparsable URLs fail closed to exact comparison. Equality regressions cover each folded variant including query-with/without-root-slash and IPv6 textual forms; conflict regressions cover the loopback split, wrong-scheme default ports, and query differences.
- **Frontend authority:** `findManagedAgentRuntime` makes `requestedRelayUrl` authoritative whenever a row carries it, matched with a TS mirror of the same connection-target rule (not `normalize_relay_url`, which folds the loopback tenants). The canonical fallback now applies only to legacy rows without the field. New regression: with `ws://localhost:3000` and `ws://127.0.0.1:3000` configured simultaneously and the child dialed to localhost, only the localhost community resolves the runtime; the 127.0.0.1 card resolves nothing, so its stop/restart cannot target the child.

Gates on the new head: full desktop/Tauri workspace suite, clippy with warnings denied, fmt, `desktop-check`, and the desktop JS unit tests are green.

anilkishan:
Pushed 796c78a83: `connection_target` and its TS mirror now reject credential-bearing, fragment-bearing, and non-ws(s) URLs instead of folding those parts away, so spellings like `wss://alice@relay.example` vs `wss://bob@relay.example` take the documented exact-string fallback rather than comparing equal. This also aligns the schemes the two helpers accept. Regression tests added on both sides.

Credit to @themiguelamador, who spotted this gap while reviewing the same helper carried into #5993.

Verification: full Tauri suite green, 4,780 desktop JS tests green, typecheck, Biome, fmt, clippy `-D warnings`, and `git diff --check` all pass.

PR #4862, opened 2026-08-05
Fix mobile message timeline bounce

klopez4212:
## Summary

Stop repeated follow-latest scrolling after layout changes in channels and DMs.

## Validation

- `flutter analyze lib/features/channels/channel_detail_page.dart`
- `flutter test test/features/channels/channel_detail_page_test.dart`
- Full mobile pre-push suite

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page/message_list.dart
+++ b/mobile/lib/features/channels/channel_detail_page/message_list.dart
@@ -45,7 +45,7 @@ class _MessageList extends HookConsumerWidget {
       initialMessageId == null && initialThreadRootId == null,
     );
     final isAutoScrolling = useRef(false);
-    final autoScrollScheduled = useRef(false);
+    final latestRealignmentQueued = useRef(false);
     final latestEntryId = entries.isEmpty ? null : entries.last.message.id;
     final previousLatestEntryId = useRef<String?>(null);
     final didOpenInitialThread = useRef(false);
@@ -198,18 +198,6 @@ class _MessageList extends HookConsumerWidget {
       }
     }
 
-    void scheduleAutoScrollToLatest() {
-      if (autoScrollScheduled.value || isAutoScrolling.value) return;
-      autoScrollScheduled.value = true;
-      WidgetsBinding.instance.addPostFrameCallback((_) {
-        autoScrollScheduled.value = false;
-        if (!context.mounted || !followsLatest.value || hasUserScrolled.value) {
-          return;
-        }
-        scrollToLatest();
-      });
-    }
-
     bool latestIsAtBoundary() {
       // In this reversed list, item 0's leading edge is the bottom boundary.
       // Being merely visible is not enough: a user who has pulled a tall
@@ -218,6 +206,29 @@ class _MessageList extends HookConsumerWidget {
         (position) =>
             position.index == 0 && position.itemLeadingEdge.abs() < 0.01,
       );
+    }
+
+    void realignLatestAfterLayoutChange() {
+      if (latestRealignmentQueued.value ||
+          !followsLatest.value ||
+          hasUserScrolled.value) {
+        return;
+      }
+      latestRealignmentQueued.value = true;
+      WidgetsBinding.instance.addPostFrameCallback((_) {
+        latestRealignmentQueued.value = false;
+        if (!context.mounted ||
+            !itemScrollController.isAttached ||
+            !followsLatest.value ||
+            hasUserScrolled.value ||
+            latestIsAtBoundary()) {
+          return;
+        }
+        // A dock or keyboard resize is a layout correction, not a navigation
+        // action. Keeping it instant avoids restarting a smooth scroll for
+        // every position report while the viewport settles.
+        itemScrollController.jumpTo(index: 0);
+      });
     }
 
     useEffect(() {
@@ -232,12 +243,7 @@ class _MessageList extends HookConsumerWidget {
         }
         if (nextIsAtLatest) {
           if (!isAtLatest.value) isAtLatest.value = true;
-        } else if (followsLatest.value && !hasUserScrolled.value) {
-          // The viewport can shrink when the composer or keyboard opens.
-          // Preserve auto-follow until the user scrolls the timeline.
-          if (!isAtLatest.value) isAtLatest.value = true;
-          scheduleAutoScrollToLatest();
-        } else if (isAtLatest.value) {
+        } else if (!followsLatest.value && isAtLatest.value) {
           isAtLatest.value = false;
         }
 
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
dropping position-driven auto-scroll matches the existing UserScrollNotification clears-followsLatest path. realign on composer inset + metrics is the right trigger set for the bounce.

wesbillman (dismissed):
Requesting changes on Wes Billman's behalf at current head `f8771b692a33a66a90fbe2fa45a898a856672fba`.

The implementation direction looks reasonable: position reports no longer initiate repeated animated scrolling, explicit composer/keyboard layout signals queue one post-frame boundary correction, and every correction rechecks attachment plus the user's follow state before jumping. The existing tall-message test still protects against snapping a deliberately detached user back to the tail.

The blocker is regression coverage. This PR changes the timeline's core follow-state/event-timing behavior, but its only test change adds clock pumps to the unrelated capped unread-history fetch test (`channel_detail_page_test.dart:1468-1477`). Nothing exercises either new trigger (`composerBottomInset` or `didChangeMetrics`), proves a followed timeline is realigned after viewport shrink, or verifies repeated layout notifications do not restart/bounce scrolling.

Please add a focused widget regression test that starts at the latest boundary, changes the composer inset and/or dispatches a metrics change, and proves the list settles at index 0 without repeated motion. It should also retain coverage that the same layout signal does not snap back after a user has left follow mode. This is exactly the sort of state machine that becomes a tiny haunted elevator without a test.

CI is green at this head; I relied on the existing full Mobile run rather than rerunning its equivalent locally.

brow:
🤖 I reviewed this at `f8771b692a33a66a90fbe2fa45a898a856672fba` with a second agent on the regression lane. **No blocker from us.** I also built the regression test that was missing, so I can now show what this PR fixes. That is the part worth your time, because it answers the coverage request on this PR.

## The fix is real, and it is measurable in a widget test

Two earlier attempts of mine failed to tell your branch apart from its merge base. Both attempts pointed at the new realign path. That was the wrong target. The recognizer is the code you **removed**, and it needs no keyboard and no metrics event at all. Resting geometry alone is enough.

With 30 messages in the channel, measured on the same probe against both trees:

| Tree | Newest row bottom, relative to the composer top |
| --- | --- |
| merge base `8342dfcc` | `-62.0` |
| this branch `f8771b69` | `0.0` |

On the merge base the newest message sits a full composer height behind the composer. On your branch it rests flush. That is the bounce, and your change fixes it.

The mechanism, as far as I measured it. At rest, item 0's leading edge reads `0.1033`, not zero, because in the reversed list that position includes the list's own 62 pixel bottom padding. So `latestIsAtBoundary()` is false at a visually flush tail. On the merge base the removed `onPositionsChanged` branch read that as "not at the tail" and called `scheduleAutoScrollToLatest`, which called `scrollToLatest`, and the row ended up 62 pixels behind the composer. That is exactly the padding value, so the scroll target appears to ignore the bottom padding. Removing that branch is what fixes it. I am confident in the two measured geometries; the last step of that chain is my reading of why, not something I isolated further.

## The test that pins it

This is the shape that goes red on the merge base and green here. It needs no `viewInsets` and no metrics observer:

```dart
// 30 messages, channel page at rest, no user scroll.
expect(
  tester.getBottomLeft(find.byKey(const ValueKey('channel-message-group-<newest-id>'))).dy,
  tester.getTopLeft(find.byKey(const ValueKey('channel-composer-dock'))).dy,
);
```

I would put that in `channel_detail_page_test.dart`. It pins the user-visible outcome rather than the implementation, so it will survive a future refactor of the realign path.

Worth knowing: the new `realignLatestAfterLayoutChange` is a no-op in the widget harness. It runs, and its `jumpTo` fires, and the resting geometry is identical with it neutered. So a test aimed at the realign path cannot pin this. A test aimed at resting geometry can.

## One pre-existing problem this PR now leaves as the only bad path

New-message arrival still goes through `scrollToLatest`, which this PR does not change. It has the same padding-blind alignment. When a message arrives while the user sits at the tail, I measured the newest row animating from flush to `-62.0` over about five frames. Of a 75 pixel row, 62 pixels finish behind the composer and 13 remain visible.

This reproduces identically on the merge base, so it is **not a regression from this PR**, and I am not asking you to fix it here. I am flagging it because after your change it is the last path with the old geometry, and it looks like the same root cause you just fixed. Two caveats on my own measurement: it is a widget-harness number, and a short list (3 messages, so the content does not fill the viewport) rests flush and shows no overshoot, so the effect needs a scrollable backlog. A follow-up that makes `scrollToLatest` padding-aware would be worth a look.

## One thing not to do

My second agent suggested widening `latestIsAtBoundary()` so the padded resting edge counts as "at the boundary". I tested that and it is wrong. Widening the predicate turns `keeps follow mode off while a tall newest message stays visible` red. That test protects a deliberate behavior from [#3121](https://github.com/block/buzz/pull/3121): a user who pulls a tall newest row away from the boundary must not get snapped back. Please leave the `abs() < 0.01` predicate alone. Deriving the boundary from real row and composer geometry would be a separate and larger change.

Wes asked for a regression test that changes the composer inset or dispatches a metrics change and proves the list settles at index 0. I could not build that one: the realign path is a no-op in this harness, so a metrics-driven test passes with the fix neutered. The resting-geometry test above is the version that actually goes red on the merge base, and it covers the same user-visible outcome. It also does not cost you the follow-mode protection, which a metrics test would need to re-assert separately.

## Smaller notes

The channel page uses `jumpTo` where `thread_detail_page.dart` uses an animated `scrollTo`, and the thread version re-arms its follow flag while the channel version does not. Your comment explains the instant choice and I agree with it. Is the divergence from the thread page deliberate?

The rewritten branch reads `else if (!followsLatest.value && isAtLatest.value)`, which is narrower than the base. It is correct, and it depends on every write to `followsLatest` and `hasUserScrolled` staying paired. I checked all five write sites and they are. Nothing enforces that, so a short comment naming the invariant would protect the next editor.

## What we checked and did not find

We looked for a viewport shrink that bypasses both new triggers, and did not find a reachable one. Composer growth, multiline growth, attachment panels, typing indicators, and previews all report through `ComposerDockSizeReporter`, which updates `composerBottomInset`. Keyboard changes arrive through `didChangeMetrics`. We also traced queued metric bursts and the empty position set, and found the coalescing safe and the empty case failing in the safe direction. Both negatives are bounded to the cases we enumerated and the widget harness.

Full mobile suite passes 1,163 tests at this head, and `flutter analyze` reports no issues.

wesbillman:
Addressed at `cd225117216e51581eeac8c4bdf298f94e36cb90`: added focused widget regressions covering both explicit layout signals and the follow-state boundary. One test proves composer growth and a later keyboard metrics resize keep the newest channel message anchored without exposing the jump-to-latest affordance; the second proves the same keyboard signal does not snap a deliberately user-detached timeline back to index 0.

Focused tests passed individually, `flutter analyze` passed in the commit hook, and the mandatory pre-push gate passed Mobile plus the repository's other configured suites.

Carl commenting on Wes's behalf.

wesbillman (approved):

brow (approved):

PR #4868, opened 2026-08-05
test(relay): pin req/search tracer gate coverage

### LINKED ISSUE #4758: test(relay): pin tracer gate on historical REQ and search call sites
## Context

PR #4647 adds `Tracer::enabled()` gates around the trace-only `communities_of_channels` lookup in both historical REQ and NIP-50 search. A live exact-SHA test proved both production paths skip the hot SQL with `NoopTracer`.

The checked-in regression test only proves `CountingTracer` delegates `enabled()` to its inner tracer. Mutation testing removed the normal REQ gate from `handlers/req.rs`; `counting_tracer_delegates_enabled_to_inner` still passed. The call-site behavior works today but is not pinned against future regression.

## Acceptance criteria

Add causal integration coverage for both changed paths:

- normal historical REQ, including a multi-filter request;
- NIP-50 search hydration;
- disabled tracer: successful result semantics remain unchanged and `communities_of_channels` is not called;
- enabled/observing tracer: the lookup is called and the expected read-seam action is emitted;
- mutation check: deleting either call-site `enabled()` gate fails its corresponding test;
- run whole `cargo test -p buzz-relay`, not only the scoped tests.

Prefer a narrow injectable DB-call observer over SQL-log parsing in the checked-in suite. Keep conformance emissions non-vacuous and avoid changing production behavior.

## Evidence

Live/mutation receipt from the PR review: `RESEARCH/PR_4647_LIVE_REGRESSION_TEST.md` in the reviewer workspace. PR discussion: #4647.

VaibhavJain2609:
## What
Adds causal regression coverage for the tracer `enabled()` gates around conformance row-community lookup in the REQ handler.

- Extracts shared helper paths for read-seam emission in `crates/buzz-relay/src/handlers/req.rs`:
  - `maybe_emit_read_message_rows_with_lookup` (historical REQ lane)
  - `maybe_emit_read_by_id_rows_with_lookup` (NIP-50 search hydration lane)
- Keeps behavior identical while making the gates directly testable.
- Defers row-channel projection work until after `tracer.enabled()` so disabled tracers skip all read-seam prep work.
- Adds tests that pin both call sites and multi-filter behavior.

## Why
Issue [#4758](https://github.com/block/buzz/issues/4758) identified that removing a call-site gate could still pass existing tests, leaving the regression unpinned.

## Repro + Mutation Proof
Local repro before this change:
- `cargo test -p buzz-relay counting_tracer_delegates_enabled_to_inner` passed
- deleting a call-site `enabled()` gate in `req.rs` still passed

With this change:
- deleting the historical helper gate causes `historical_req_multi_filter_disabled_tracer_skips_lookup` to fail
- deleting the search helper gate causes `search_hydration_disabled_tracer_skips_lookup` to fail

## Tests
- `cargo test -p buzz-relay historical_req_multi_filter`
- `cargo test -p buzz-relay search_hydration_`
- `cargo test -p buzz-relay`

Closes [#4758](https://github.com/block/buzz/issues/4758).

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -39,6 +39,89 @@ pub(crate) const FILTER_QUERY_CONCURRENCY: usize = 4;
 // the range fails the build.
 const _: () = assert!(FILTER_QUERY_CONCURRENCY >= 2 && FILTER_QUERY_CONCURRENCY <= 8);
 
+fn distinct_channels(rows: &[Option<uuid::Uuid>]) -> Vec<uuid::Uuid> {
+    let mut set: std::collections::BTreeSet<uuid::Uuid> = std::collections::BTreeSet::new();
+    for channel_id in rows.iter().flatten() {
+        set.insert(*channel_id);
+    }
+    set.into_iter().collect()
+}
+
+async fn maybe_emit_read_message_rows_with_lookup<Rows, Lookup, Fut, E>(
+    tracer: &Arc<dyn crate::conformance::Tracer>,
+    trace_state: Option<&crate::conformance::AbstractState>,
+    filter_channel: Option<uuid::Uuid>,
+    rows: Rows,
+    lookup: Lookup,
+) -> bool
+where
+    Rows: FnOnce() -> Vec<Option<uuid::Uuid>>,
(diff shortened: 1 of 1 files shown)


VaibhavJain2609:
@claude  review

PR #4872, opened 2026-08-05
fix(cli): preserve event signatures in JSON output

JAFairweather:
## Summary

Preserve the canonical Nostr event signature in normalized JSON output from read commands. The CLI currently describes JSON as full output, but `normalize_events` drops `sig`, so downstream consumers cannot independently verify the returned event.

Compact output remains unchanged. Relay-internal fields remain excluded.

## Test plan

- `cargo fmt --check`
- `cargo test -p buzz-cli` (322 passed)
- Added a regression proving `sig` survives normalization while an unrelated relay-internal field does not.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -191,7 +191,8 @@ Extract `channel` and `id` from the URL query parameters. The optional
 `thread` parameter (root event ID) can be ignored — `messages thread` resolves
 the full thread from the event ID alone.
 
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
keeping sig on the full json path is the right default for anything that rebroadcasts. compact paths still projecting down is fine as long as the docs say which flag drops sig.

PR #4874, opened 2026-08-05
fix(acp): deliver plain-text replies for self-hosted OpenAI-compatible agents

repudi8or:
## Summary

- Small local models served over a self-hosted OpenAI-compatible endpoint (e.g. llama.cpp/vLLM/mesh-llm) routinely finish a multi-step turn by writing the answer as prose instead of calling `send_message` — the turn reports `EndTurn` successfully, the ACP activity log shows the correct reply, but nothing is ever published to the channel. This is what we observed running a self-hosted local-model agent against a mesh-llm deployment.
- `crates/buzz-acp`: retain each turn's streamed text (bounded at 8 KiB) plus whether a publish tool call was seen; on `EndTurn`, if nothing was published, deliver that text as a threaded channel reply instead. Gated strictly on `EndTurn` (not `MaxTokens`/`MaxTurnRequests`, which mean the turn was truncated) and filters bare acknowledgements ("OK", "Done"). Off by default via `--deliver-plain-replies` / `BUZZ_ACP_DELIVER_PLAIN_REPLIES`, since cloud models reliably call `send_message` themselves and enabling it there would risk double-posting.
- `desktop/src-tauri`: opt the built-in "Buzz shared compute" (`relay-mesh`) preset into this fallback, and separately default it on for the general `openai-compat` provider too (a self-hosted endpoint configured directly, not through the built-in preset) — scoped narrowly so plain `openai` (the real OpenAI cloud API) is untouched. An explicit user override (including an intentional `false`) always survives.
- `crates/buzz-agent`: while validating the `openai-compat` fallback above against a real self-hosted setup (a proxy fanning a single model alias out across several backend providers), we also hit a second, unrelated failure mode worth fixing alongside it — a 400 from a backend routed through OpenRouter: `"auto tool schema uses unsupported assertions or reserved metadata"`. The root cause: `buzz-dev-mcp`'s tool schemas are auto-derived via `rmcp`/`schemars`, which routinely emits JSON Schema metadata and assertion keywords (`$schema`, `title`, `format`, `additionalProperties`, range/length constraints, ...) that Anthropic and the official OpenAI API happily ignore, but that some openai-compat backends validate strictly and reject outright — breaking every tool call for that agent, not just the reply-delivery path. Added `sanitize_tool_schema`, applied once at tool registration alongside the existing schema-size cap (`cap_schema`), which strips a denylist of those keywords from the whole schema tree before it's ever sent to a provider. Every stripped key only narrows or documents what a schema accepts, so this can only *loosen* a backend's validation — it can't change what a tool call actually does, and structural/semantic keywords (`type`, `properties`, `required`, `items`, `description`, `enum`) are left untouched. This is the same underlying theme as the reply-delivery fix above (self-hosted `openai-compat` backends don't share the reference providers' tolerances), so it seemed worth landing together — useful for anyone pointing `buzz-agent` at vLLM, llama.cpp, a litellm/multi-provider proxy, or any OpenRouter-routed model with a stricter function-calling schema validator than Claude/GPT.
- `desktop/src-tauri`: merged `main` in to pick up two commits that had landed on `crates/buzz-acp` since this branch started, and split `readiness.rs`'s test module (`readiness_tests.rs` / `readiness_provider_tests.rs`) to stay under the desktop file-size ratchet after that merge — no behavior change, same tests, same assertions.

## Test plan

- [x] `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib` — 2250 passed, 0 failed, 14 ignored (full crate, not scoped to touched files), run independently by two people
- [x] New targeted tests: `publish_detection_matches_tool_and_shell_shapes`, `bare_acknowledgements_are_not_worth_publishing`, `non_mesh_provider_does_not_opt_into_plain_reply_delivery`, `openai_compat_provider_opts_into_plain_reply_delivery`, `real_openai_provider_does_not_opt_into_plain_reply_delivery`, `explicit_deliver_plain_replies_override_is_preserved_for_openai_compat`
- [x] `cargo test -p buzz-agent --lib` — 401 passed, 0 failed, including 4 new schema-sanitization tests (`sanitize_tool_schema_strips_root_metadata`, `sanitize_tool_schema_strips_nested_property_assertions`, `sanitize_tool_schema_preserves_structural_and_semantic_keywords`, `sanitize_tool_schema_recurses_into_arrays`)
- [x] `cargo clippy -p buzz-agent -p buzz-acp --lib -- -D warnings` and `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml --all -- --check` clean
- [x] Full pre-push suite green after the main merge: branch-skew, rust-tests, desktop-check (file-size ratchet), desktop-typecheck, desktop-test (4480 JS tests), mobile-test, desktop-tauri-checks (2250 Rust tests)
- [x] Independent code review of the diff
- [ ] Live end-to-end repro against the original self-hosted agent — not done pre-merge: the fix spans `buzz-acp` (bundled inside the Buzz Desktop app) and `desktop/src-tauri`, and there's no way to swap just one agent's binary without patching the signed Desktop app bundle used by every other agent on that machine. Recommend validating live once this ships through the normal Desktop build/update path.

Diagnosed and coordinated by the buzz-acp team (architect, dev, tester, conductor) in the team's working channel; core mechanism forward-ported from an unmerged branch (`micspiral/mesh-0-74-gemma`, commit `bb4af4189`) by Michael Neale.

with a helping hand from Claude Code


repudi8or:
@wesbillman this fixes a bug affecting self-hosted OpenAI-compatible agents (e.g. local llama.cpp/mesh-llm endpoints) silently dropping their replies instead of publishing them — full context and test results in the PR description. Would appreciate a review/merge when you have a moment. I only have read access to block/buzz directly (opened from a fork), so can't self-merge.

repudi8or:
## Live end-to-end repro: confirmed ✅

Independently verified the events directly on the relay (not just trusting the report):

- Hello: `92df0f20db6cea23f2744e6b24dc476e09aae3a66299168e302fe2edf97abd0a` — `@Local LLM (llm1) hello`, tagged to the dev-branch test agent's pubkey `5299bb38ce49b5090c08022359842c396398a6b36efd4cbcd8484d1aa56b22eb`.
- Reply: `2351d7ad8b56f94ac457be5e880f5714dd3c59a45bf201a5f9343091f2339de8` — kind 9, **from** `5299bb38…`, with an `e`-tag `reply` reference back to the hello event. Correctly threaded.

This closes the one unchecked item in the test plan: a self-hosted `openai-compat` agent's plain-text turn (no `send_message` tool call — the model never called it) landed as a real threaded channel reply instead of vanishing after generation. Setup: repudi8or's `just dev` isolated Desktop build (already on `fix/deliver-plain-text-replies`) spun up its own test copy of the "Local LLM (llm1)" agent against the real production relay, so this exercised the actual shipped code path end-to-end — not a mock.

**Non-blocking follow-up spotted in the reply content** (not the delivery mechanism, which is what this PR fixes): the model looped through many near-duplicate self-summaries for the full turn, and the delivered text is hard-cut mid-word — consistent with hitting the 8KiB `MAX_TURN_TEXT_BYTES` per-turn buffer cap in `acp.rs`. Delivery/threading itself is correct; this is llm1's own response quality (repetition / no natural stop) and worth a separate issue (better stop behavior or a repetition guard for small local models), not a blocker here.

imp0stor:
Opened a rebased successor onto current `main`: https://github.com/block/buzz/pull/7471

Branch: `imp0stor:cursor/deliver-plain-text-replies-rebased-00fb` (rebased from this PR's `fix/deliver-plain-text-replies` commits). `cargo test -p buzz-acp` passed after conflict resolution.

We also validated a local Desktop 0.5.23 build of this ACP lineage with Ollama (`BUZZ_ACP_DELIVER_PLAIN_REPLIES=1`) — channel replies landed without a custom inference rewrite proxy.

Happy to adjust #7471 if maintainers prefer to continue on this PR instead.

PR #4875, opened 2026-08-05
fix: restore Cmd+W as File > Close Window on macOS

ole-tidal:
## Why

⌘W is one of macOS's oldest system-wide conventions: **File > Close Window** closes the focused window of the active app. It works in Finder, Safari, Slack, Discord, Notes — everywhere — and decades of muscle memory press it dozens of times a day. In Buzz it silently did nothing.

That was a deliberate hijack, not an accident: `app_menu.rs` removed the default `close_window` items Tauri installs, because macOS resolves a menu key equivalent **before** the webview receives any key event, so while the item existed Buzz Term could never bind ⌘W to "close this terminal tab". Removing the item fixed the terminal-local conflict by breaking a standard OS chord **app-wide** — every non-terminal surface of the app stopped honoring it. That trade is backwards: an app-specific feature should never win over a platform-wide convention.

For a tray-resident app, "close the window" means hide-to-tray — exactly what the red traffic-light button already does via the `CloseRequested` interception in `lib.rs`, and the same ⌘W behavior as other tray-resident chat apps.

## What

- Restore **File > Close Window (⌘W)** as a *custom* menu item (predefined items cannot be toggled after creation).
- The item closes the **focused** window through the normal `CloseRequested` path — the main window hides to tray, huddle windows keep their drawer-restore behavior. Falls back to the main window when no webview window reports focus.
- New `set_close_window_menu_enabled` command: Buzz Term disables the item for exactly as long as it owns the keyboard. A **disabled** menu item does not consume its key equivalent, so ⌘W falls through to the webview and the terminal's close-tab chord (`matchTabChord`) keeps working. `TerminalSubstrate` already tracks ownership; it now syncs the menu item on every ownership flip and re-enables on unmount.
- The Window submenu's duplicate close item is deliberately not restored — File is the chord's canonical home.
- Unit test for the ownership → enabled mapping (`closeWindowMenuEnabledFor`).

## Verification on a local build

Tested against a locally built instance (`just desktop-standalone`), driving the real app pid-exactly over the Accessibility API:

- File > Close Window present with cmd char `W`, modifiers `0` (plain ⌘), enabled by default.
- ⌘W key event delivered to the app: window hides to tray (AX window count 1 → 0), process stays alive.
- Clicking the menu item does the same.
- Not coverable in standalone mode: the terminal-owned disable span (needs a live community + Buzz Term session). That mapping is unit-tested, and the enable/disable command is a thin `set_enabled` wrapper.

Note: the follow-up native-lifecycle PR (https://github.com/block/buzz/pull/4876) builds on this branch (both touch `app_menu.rs`). It targets `main` and includes this PR's commit until this one merges.

### DIFF
--- a/desktop/src-tauri/src/app_menu.rs
+++ b/desktop/src-tauri/src/app_menu.rs
@@ -5,42 +5,111 @@
 //! `close_window` item in both the File and Window submenus, and muda gives
 //! that item a Cmd+W key equivalent bound to `performClose:`.
 //!
-//! Two consequences, both wrong for Buzz:
+//! An earlier revision of this module removed both `close_window` items,
+//! because macOS resolves a menu key equivalent before the webview receives
+//! any key event, so Buzz Term could never bind Cmd+W to "close this
+//! terminal tab" while the accelerator was claimed here. That cure traded a
+//! terminal-local conflict for an app-wide regression: Cmd+W is the standard
+//! macOS chord for "close the focused window", and with the item gone it did
+//! nothing anywhere else in the app. For a tray-resident app "close" means
+//! hide-to-tray (the `CloseRequested` interception in `lib.rs`), exactly the
+//! Cmd+W behavior of other tray-resident chat apps.
 //!
-//! 1. `CloseRequested` on the main window is intercepted in `lib.rs` and turned
-//!    into hide-to-tray, so Cmd+W never closed a window -- it hid the whole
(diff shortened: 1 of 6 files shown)


ole-tidal:
🤖 Closing at the reporter's request alongside the companion lifecycle PR (https://github.com/block/buzz/pull/4876) — the reporting machine's window-manager issue was resolved externally and the reporter opted not to pursue these changes.

PR #4876, opened 2026-08-05
fix: adopt the native macOS app lifecycle around hide-to-tray

ole-tidal:
## Why

Buzz on macOS deviated from standard app behavior in ways that compounded into "the app ignores me" moments (originally reported as Rectangle and ⌘W not working). All of them trace back to the hide-to-tray window model missing its native-lifecycle counterparts:

1. **⌘Tab into Buzz while the window is hidden showed… nothing.** Only the menu bar switched. macOS surfaces Dock-icon clicks as `applicationShouldHandleReopen:` (Tauri's `Reopen` event — already handled), but ⌘Tab and other app-switcher activations only post `NSApplicationDidBecomeActiveNotification`, which nothing observed. Every standard macOS app re-presents a window on activation.
2. **After the window hides, Buzz stayed the active app** — frontmost with zero windows: a menu bar with nothing under it, a dead ⌘Tab target, and window managers like Rectangle resolve their target from the frontmost app's focused window and find none.
3. **The main window had an empty title** (`"title": ""`), so Mission Control, the Window menu, and AX tools showed a nameless window.
4. **No Settings… (⌘,) or Check for Updates… menu items** — standard app-submenu entries users reach for.

## What

- **`app_activation.rs` (new):** observes `NSApplicationDidBecomeActiveNotification` (block2 observer, same pattern and safety posture as `mouse_nav.rs`) and re-presents the main window when the app becomes active with no visible webview window. Gated on the deliberate first reveal (`INITIAL_REVEAL_DONE`, set by the first-frame reveal and by any explicit show) so launch-time self-activation cannot preempt the geometry-settled reveal. Minimized-only counts as "nothing visible", matching how ⌘Tab de-miniaturizes in standard apps.
- **`lib.rs`:** after hide-to-tray on `CloseRequested`, if no other window is visible, `app.hide()` yields activation to the next app — the same handoff as closing the last window of any Mac app. The two changes compose: ⌘W hands focus to the next app; ⌘Tab back re-presents the window.
- **`tauri.conf.json`:** window `title: "Buzz"` (still visually hidden via `hiddenTitle` + overlay title bar).
- **App submenu** gains **Check for Updates…** and **Settings… (⌘,)** in their HIG positions. Both re-present the window and forward events to the webview: Settings toggles the settings route (mirroring the existing webview ⌘, shortcut, which keeps working on non-mac where there is no native menu), Check for Updates lands on Settings → Updates with the check already running.

## Verification on a local build

Tested against a locally built instance (`just desktop-standalone`), driving the real app pid-exactly over the Accessibility API:

- Window AX title is `Buzz` (was empty) — `windowTitles=["Buzz"]`.
- App submenu shows `Check for Updates…` and `Settings…` with cmd char `,`, in HIG positions between About and Services.
- ⌘W (posted to the pid) hides the window to tray; process stays alive; no other window visible afterward.
- The activation observer fires on `NSApplicationDidBecomeActiveNotification` with correct gate values — verified via a temporary log line in the running app: `did-become-active reveal_done=true any_visible=true` (correctly no-oping while a window is visible).
- `show_main_window` re-presents the hidden window (tray path, exercised repeatedly).
- The restore decision (`should_restore`) is unit-tested for all four gate combinations.

**Scope note on Rectangle:** during live verification, the reporter's *residual* "Rectangle can't resize the window" symptom (window visible and focused) turned out to be Rectangle's own stale Accessibility session — restarting Rectangle fixed it, and the window was AX-resizable throughout. This PR's window-manager relevance is limited to the real defect above: the frontmost-with-zero-windows state after close-to-tray, where there is no window for any WM to target.

**Verification gap, called out honestly:** the composed end-to-end gesture — a real ⌘Tab into hidden-window Buzz auto-re-presenting, and the frontmost-app switch after close — cannot be driven programmatically: macOS 14+/Tahoe refuses every synthetic activation route without real user intent (`NSRunningApplication.activate` cooperative-refused, `activateIgnoringOtherApps` a documented no-op, System Events `set frontmost` refused, HID-synthesized Dock clicks filtered). Every constituent link is verified above; the composition is three unit-tested lines. One human ⌘Tab confirms it.

Also not coverable in standalone mode (no community, so `AppShell` and its listeners never mount): the webview side of the Settings…/Check for Updates… events. That path is typechecked and mirrors the existing ⌘, handler; the native side (items present, correct accelerator) is verified above.

**Builds on the ⌘W PR** (https://github.com/block/buzz/pull/4875) — both touch `app_menu.rs`. Opened from a fork, so it targets `main` directly and includes that PR's commit until it merges; once it lands, this diff reduces to the lifecycle changes (the second commit).

### DIFF
--- a/desktop/src-tauri/Cargo.toml
+++ b/desktop/src-tauri/Cargo.toml
@@ -53,8 +53,8 @@ webkit2gtk = { version = "=2.0.2", features = ["v2_22"] }
 [target.'cfg(target_os = "macos")'.dependencies]
 block2 = { version = "0.6", default-features = false, features = ["std"] }
 objc2 = { version = "0.6.4", default-features = false }
-objc2-app-kit = { version = "0.3.2", default-features = false, features = ["NSEvent", "NSHapticFeedback", "NSMenu", "NSMenuItem", "NSStatusItem", "block2"] }
-objc2-foundation = { version = "0.3.2", default-features = false, features = ["NSProcessInfo", "NSString"] }
+objc2-app-kit = { version = "0.3.2", default-features = false, features = ["NSApplication", "NSEvent", "NSHapticFeedback", "NSMenu", "NSMenuItem", "NSStatusItem", "block2"] }
+objc2-foundation = { version = "0.3.2", default-features = false, features = ["NSNotification", "NSOperation", "NSProcessInfo", "NSString", "block2"] }
 keyring = { version = "3.6.3", default-features = false, features = ["apple-native", "vendored"], optional = true }
 security-framework = { version = "3.7.0", features = ["OSX_10_15"] }
 window-vibrancy = "0.6"
--- a/desktop/src-tauri/src/app_activation.rs
+++ b/desktop/src-tauri/src/app_activation.rs
@@ -1,0 +1,74 @@
+//! Restores the main window when Buzz becomes active with nothing visible.
+//!
+//! macOS surfaces Dock-icon clicks as `applicationShouldHandleReopen:` (tao
+//! forwards it as `RunEvent::Reopen`), but Cmd+Tab and other app-switcher
+//! activations only post `NSApplicationDidBecomeActiveNotification`.
+//! Without an observer for it, switching back to a hidden-to-tray Buzz
+//! brings up just the menu bar and no window, where every standard macOS
+//! app re-presents one. The observer shows the main window whenever the
+//! app becomes active with no visible webview window, gated on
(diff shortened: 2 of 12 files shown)


ole-tidal:
🤖 Closing at the reporter's request: the driving pain point (window-manager control) turned out to be a stale Rectangle Accessibility session on the reporting machine, resolved by restarting Rectangle. Withdrawing the remaining native-behavior changes rather than pursuing them.

PR #4880, opened 2026-08-05
feat(cli): allow generic file uploads including zip

Trevongit:
## Summary

- Widen `buzz-cli` upload MIME policy so agent co-lab can attach **zip** skill packs and other safe generic files (pdf, text, json, common archives/docs).
- Keep a **danger denylist** aligned with `buzz-media` (HTML/JS/SVG, native executables/installers); images and `video/mp4` unchanged.
- Generic uploads use a **100 MB** cap (matches relay default `BUZZ_MAX_FILE_BYTES`).

The relay already routes non-image uploads through `process_file_upload` + denylist. The CLI was the choke (`unsupported file type: application/zip`).

## Dogfood

- Unit tests for allow/block/size tiers
- Live zip upload to a private My Groundfeed via rebuilt CLI succeeded

## Test plan

- [x] `cargo test -p buzz-cli --lib upload_`
- [ ] CI full suite
- [ ] Optional: `buzz upload file --file pack.zip` against a local or private relay

### DIFF
--- a/crates/buzz-cli/src/client.rs
+++ b/crates/buzz-cli/src/client.rs
@@ -37,7 +37,16 @@ pub struct BlobDescriptor {
 }
 
 /// Build an `imeta` tag array from a BlobDescriptor (NIP-92 media metadata).
+///
+/// When `filename` is set (basename of the local path the user uploaded), it is
+/// included as `filename <name>` so Desktop can render FileCards with a real
+/// label for generic attachments.
 pub fn build_imeta_tag(d: &BlobDescriptor) -> Vec<String> {
+    build_imeta_tag_with_filename(d, None)
+}
+
+/// Like [`build_imeta_tag`], optionally attaching a `filename` field.
+pub fn build_imeta_tag_with_filename(d: &BlobDescriptor, filename: Option<&str>) -> Vec<String> {
     let mut tag = vec![
         "imeta".to_string(),
         format!("url {}", d.url),
@@ -45,6 +54,14 @@ pub fn build_imeta_tag(d: &BlobDescriptor) -> Vec<String> {
         format!("x {}", d.sha256),
         format!("size {}", d.size),
     ];
+    if let Some(name) = filename.map(str::trim).filter(|s| !s.is_empty()) {
+        // Relay rejects path-like filenames; basename only.
+        let base = std::path::Path::new(name)
+            .file_name()
+            .and_then(|s| s.to_str())
+            .unwrap_or(name);
+        tag.push(format!("filename {base}"));
+    }
     if let Some(ref dim) = d.dim {
         tag.push(format!("dim {dim}"));
     }
@@ -58,6 +75,37 @@ pub fn build_imeta_tag(d: &BlobDescriptor) -> Vec<String> {
         tag.push(format!("duration {dur}"));
     }
     tag
+}
+
+/// Format one uploaded attachment for message body markdown.
+///
+/// Matches Desktop `formatImetaMediaLine`:
+/// - `video/*` → `![video](url)`
+/// - `image/*` (except `*.agent.png` / `*.team.png` snapshots) → `![image](url)`
+/// - everything else (zip, pdf, txt, …) → plain `[filename](url)` so Desktop
+///   FileCard renders a download card instead of a broken image.
+pub fn format_attachment_markdown(file_path: &str, desc: &BlobDescriptor) -> String {
+    let mime = desc.mime_type.as_str();
+    let basename = std::path::Path::new(file_path)
+        .file_name()
+        .and_then(|s| s.to_str())
+        .unwrap_or("file");
+    let lower = basename.to_ascii_lowercase();
+    let is_snapshot_png = lower.ends_with(".agent.png") || lower.ends_with(".team.png");
+
+    if mime.starts_with("video/") {
+        return format!("\n![video]({})", desc.url);
+    }
+    if mime.starts_with("image/") && !is_snapshot_png {
+        return format!("\n![image]({})", desc.url);
+    }
+
+    // Generic / snapshot: escape `[` `]` `\` in the label so markdown stays valid.
+    let escaped = basename
+        .replace('\\', "\\\\")
+        .replace('[', "\\[")
+        .replace(']', "\\]");
+    format!("\n[{escaped}]({})", desc.url)
 }
 
 /// Image MIME types accepted on the image/thumbnail upload path.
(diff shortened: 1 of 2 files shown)


michegz:
Independent prod verification, in case it's useful for review — I hit the exact `unsupported file type: application/pdf` this patches against a live `communities.buzz.xyz` deployment (`buzz upload file --file cattle.pdf` on a 180 KB, 7-page document). To sanity-check the fallback story this PR relies on, I bypassed the CLI and signed a Blossom kind:24242 event directly to `PUT /upload`: returns `200 OK` with `"type":"application/pdf"`, and `buzz media get` round-trips byte-identical (SHA-256 match). So the server-side \`validate_file_content\` + \`BLOCKED_FILE_MIME_TYPES\` filter this PR trusts is doing its job in production — the CLI's \`ALLOWED_MIMES\` really is the sole choke on `communities.buzz.xyz` today.

One thing I liked reviewing the diff: \`is_upload_mime_allowed()\` mirrors the relay's denylist rather than hand-rolling a parallel allowlist, so the CLI won't silently drift out of sync the next time \`buzz-media\` adds a MIME. That's a nicer factoring than a static list would have been.

Trevongit:
Addressed the Codex P2 (generic attachments as links):

- `messages send --file` now formats via `format_attachment_markdown` (Desktop `formatImetaMediaLine` parity):
  - `image/*` → `image` (except `*.agent.png` / `*.team.png` snapshots → file link)
  - `video/*` → `video`
  - generic (zip/pdf/txt/…) → `[filename](url)` with label escaping
- imeta tags now include `filename <basename>` for FileCard labels

Pushed on this PR as `fix(cli): format generic file attachments as download links`.

ravarora2 (dismissed):
lgtm

wesbillman (dismissed):
Review submitted by Carl on Wes's behalf.

The upload policy and generic-link formatting are sound, but the new filename field needs to obey the relay validation it targets before this can merge.

**[P2] Validate or omit relay-invalid basenames before publishing** — `build_imeta_tag_with_filename` currently appends every non-empty basename unchanged. On Unix, filenames may contain `\\` and control characters (including newlines); the relay explicitly rejects either in `crates/buzz-relay/src/handlers/imeta.rs` (`imeta filename must not contain path separators or control characters`). `cmd_send_message` now adds this field for *every* attachment, so a previously sendable image such as `report\\final.png` uploads successfully and then the message publish fails, leaving an orphaned blob. Please apply the relay's 1–255-byte / no-separator / no-control validation before upload or omit the filename field when invalid, and cover the invalid-basename case. The markdown label should follow the same safe fallback if the field is omitted.

crates/buzz-cli/src/client.rs
```diff
@@ -37,14 +37,26 @@ pub struct BlobDescriptor {
...
         format!("x {}", d.sha256),
         format!("size {}", d.size),
     ];
+    if let Some(name) = filename.map(str::trim).filter(|s| !s.is_empty()) {
+        // Relay rejects path-like filenames; basename only.
+        let base = std::path::Path::new(name)
+            .file_name()
+            .and_then(|s| s.to_str())
+            .unwrap_or(name);
+        tag.push(format!("filename {base}"));
```
wesbillman: **Blocking:** This does not actually guarantee a relay-valid filename. On Unix, `Path::file_name()` leaves `bad\\name.zip` unchanged; control characters are also retained, and names longer than 255 are unbounded. The relay rejects all three (`crates/buzz-relay/src/handlers/imeta.rs:140-155`), while Desktop explicitly strips both separator styles, controls, and caps the label (`desktop/src-tauri/src/commands/media.rs:138-153`). Because `cmd_send_message` now attaches `filename` to every upload, even an otherwise valid image with such a local name uploads first and then fails message ingest—a regression from the old filename-free tag. Please centralize a sanitizer matching the relay contract and use its result for both this field and the markdown label. Add cases for backslashes, controls, and oversized/multibyte names.

wesbillman (changes requested):
One blocking regression: the new `filename` field is emitted without applying the relay's filename constraints, so valid local files can upload successfully but then make `messages send --file` fail at event ingest. Please sanitize the basename using the same contract as Desktop (strip either path separator, remove controls, bound to 255 bytes/characters as required, and provide a fallback), and use that same sanitized label for both imeta and markdown. The current Windows CI failure is in unrelated `buzz-dev-mcp` shell timeout tests, not this diff.

Trevongit:
Thanks for the careful review, Wes (and Carl).

Agree on the blocking point: we should sanitize basenames to the same contract as Desktop / relay `imeta` (strip path separators `/` and `\`, remove controls, cap at 255 UTF-8 bytes, fallback `file`) and use that for both the `filename` field and the markdown label so Blossom upload can’t succeed while `messages send --file` fails at event ingest.

**Addressed in** `8b06342a6`:
- New `sanitize_filename` in `crates/buzz-cli/src/client.rs` (mirrors Desktop `sanitize_filename` + relay byte bound)
- Used for imeta `filename` and for generic attachment markdown labels
- Unit tests for traversal, backslash basenames, controls, oversize, and multibyte bound

Windows CI timeout remains out of scope for this diff (as you noted). Ready for another look when you have a moment — re-requesting review.

mlamp:
Downstream user here — this PR would unblock a real case for us, and I want to add one data point rather than a bare +1.

**Both halves of this diff matter, not just the allowlist.** Our use case is a non-interactive agent that writes Markdown memos and posts them to a channel for a human to read. Widening the upload policy alone would still leave those attachments rendering as `image` for a `.bin` blob, so `format_attachment_markdown` + the imeta `filename` field are what actually make the artifact readable. A PR that only touched `ALLOWED_MIMES` wouldn't have solved it for us.

I also measured the current behaviour on `main` and posted the details in #2963 — the short version is that the check is provably client-side (pointing the CLI at a dead port refuses Markdown *without opening a socket*, while a PNG gets as far as `Connection refused`), and the relay returns 200 for the very same bytes over a raw Blossom PUT signed with the same key. Mirroring the relay's denylist, as this PR does, is the factoring that stops the two policies diverging again.

On status: the blocking `filename` sanitization looks addressed in `8b06342a6` (relay-contract sanitize + traversal/backslash/control/oversize/multibyte tests), and the remaining Windows CI failure was already identified as unrelated `buzz-dev-mcp` timeout tests. It's been awaiting a re-look since 2026-08-19 — a re-review when someone has a moment would be much appreciated.

Thanks for the work on this.

Trevongit:
Rebased onto current `block/buzz` `main` (`b49675894`).

The four unique CLI commits cherry-picked cleanly (zip/generic uploads, download-link markdown, clippy dead_code, imeta filename sanitize).

Unit tests on this rebase:
- `upload_allows_images_video_and_zip`
- `attachment_markdown_*` (5)
- `sanitize_filename_matches_relay_contract`
- `imeta_tag_*` (2)

All 9 passed.

Issue #4749, opened 2026-08-04
[Bug] Hosted relay intermittently stalls desktop message loading and disconnects


plantbob0101:
**Describe the bug**
Buzz Desktop 0.5.4 intermittently takes several seconds to load messages and surfaced a relay-disconnected state on a hosted `*.communities.buzz.xyz` workspace. The Mac and app are otherwise idle. Direct probes isolate the delay to the hosted relay/WebSocket path rather than the local network or renderer CPU.

**Steps to reproduce**
1. Open Buzz Desktop against a hosted `*.communities.buzz.xyz` workspace.
2. Switch to a DM/channel and wait for messages to load.
3. Observe multi-second loading and occasional relay disconnect.
4. Probe the same relay from the same Mac:
   - Ten HTTPS relay-information requests: TTFB ranged **0.720–4.705 s**.
   - Ten WebSocket REQ probes: connection/query completion ranged **0.778–4.307 s**; two connection errors and one 10 s timeout.
   - `/health` remained fast at **0.109 s**.

**Expected behavior**
Existing messages load promptly, WebSocket subscriptions stay connected, and relay responses do not intermittently take multiple seconds or time out.

**Version and platform**
- Buzz Desktop: 0.5.4 (0.5.4), current latest release as of 2026-08-04
- OS: macOS 27.0 (26A5388g), Apple silicon
- Relay: hosted `*.communities.buzz.xyz` (specific workspace omitted because the report is public)

**Logs / additional context**
Evidence captured 2026-08-04 around 10:00 America/Los_Angeles:

- Local network was healthy: base RTT ~40 ms, download throughput ~73.8 Mbps.
- Relay TCP/TLS setup was fast (~20–38 ms TCP, ~36–203 ms TLS); the delay accrued after TLS while waiting for the relay/application response.
- Buzz desktop process was idle (~0% CPU); WebContent was ~0.4–0.8% CPU during sampling.
- Relay `/health` returned 200 in 0.109 s while the NIP-11/root route took 3.066 s in the same probe set.
- App restart did not prevent recurrence; the reported slowdown occurred roughly seven minutes after a fresh 0.5.4 launch.

This points to an intermittent hosted relay/application-path latency or subscription-recovery issue, not a proven desktop rendering bug. Please inspect hosted relay request/subscription latency and desktop reconnect behavior together.

jeyarajcs:
[...]

PR #4885, opened 2026-08-05
fix(desktop): flush automatic WebSocket Pong

niklaskie:
## Summary
- flush `tokio-tungstenite`'s queued automatic Pong when Buzz Desktop's native Tauri socket receives a relay `Ping`
- continue forwarding the Ping to the webview so the existing inbound stall watchdog still records relay activity
- surface Pong flush failures through the existing WebSocket error channel
- add a duplex regression test proving an otherwise-idle desktop connection answers a server Ping

The relay sends a Ping every 30 seconds and closes a connection after three missed Pongs. `tokio-tungstenite` queues an automatic Pong when `read` returns a Ping, but that frame is not put on the wire until the client writes or flushes. The native Desktop loop forwarded the Ping to the webview and immediately waited for the next inbound frame, so an idle client never flushed the queued Pong and was disconnected after roughly 90 seconds.

### Related issue
Related to #4749, specifically the self-hosted Desktop clients with repeated `3 missed pongs` logs and roughly 90-second connections.

### Testing
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --no-default-features --lib native_websocket::tests -- --nocapture` (7 passed)
- isolated `tokio-tungstenite 0.29` reproduction: no Pong arrives after reading a Ping without another read, write, or flush

### DIFF
--- a/desktop/src-tauri/src/native_websocket.rs
+++ b/desktop/src-tauri/src/native_websocket.rs
@@ -290,7 +290,23 @@ async fn run_connection<S>(
             }
             incoming = socket.next() => {
                 let message = match incoming {
-                    Some(Ok(message)) => outbound_message(message),
+                    Some(Ok(message)) => {
+                        if matches!(&message, Message::Ping(_)) {
+                            let result = tokio::time::timeout(WRITE_TIMEOUT, socket.flush())
+                                .await
+                                .map_err(|_| "WebSocket Pong flush timed out".to_string())
+                                .and_then(|result| result.map_err(|error| error.to_string()));
+                            if let Err(error) = result {
+                                if let Ok(value) =
+                                    serde_json::to_value(OutboundMessage::Error(error))
+                                {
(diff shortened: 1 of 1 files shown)


Chessing234 (commented):
flushing the auto-pong is the right fix. breaking the whole connection on a flush timeout feels a bit sharp for a transient write stall — worth a warn+continue unless youve seen stuck sockets that only recover on reconnect.

niklaskie:
Yeah, fair point. We break here because the flush has already been stuck for 10 seconds. The connection problem itself isn't rare for us — on our self-hosted relay we saw around 100 missed-pong disconnects a day, with clients repeatedly falling into ~90-second connection loops and sometimes only recovering after reconnect or relaunch. The normal send path also treats a 10-second timeout as terminal, so keeping the same socket around would probably just delay the reconnect until the relay drops it anyway.

Issue #4166, opened 2026-08-01
feat: durable recorded voice-note messages


nickgiulioni1:
# Feature request: durable recorded voice-note messages

**Motivation**

Buzz has active work for dictation, message read-aloud, huddle speech models, and shared-compute model routing, but it does not appear to have a native way for one person to record and send the original audio as a durable message.

This is distinct from:

- dictation, which converts speech into composer text;
- message read-aloud, which synthesizes speech from an existing text message;
- huddles, which are live sessions rather than durable asynchronous messages; and
- automatic transcription, which should be an optional later layer rather than the message itself.

Recorded voice notes are useful when someone can speak more naturally or safely than they can type, especially on mobile. The recipient should be able to receive and replay the original recording after either client restarts.

**Proposed solution**

Add a first-class recorded-audio message flow:

1. Record from the message composer using an accessible tap-to-record control (with press-and-hold considered only if it has an equivalent non-gesture path).
2. Show elapsed time and clear recording state while capturing.
3. Let the sender preview, re-record, cancel, or send.
4. Upload the audio through Buzz's existing attachment/media path and publish a durable message that references it with enough metadata to render consistently (media type, duration, byte size, and attachment identity/integrity fields already used by Buzz where applicable).
5. Render an inline player for recipients with play/pause, elapsed/remaining time, seek support where feasible, download behavior consistent with other attachments, and accessible names/state.
6. Preserve the recording and playback state across normal app restart/reload. Apply existing message deletion and channel/DM authorization rules rather than inventing a separate permission model.

Suggested MVP boundaries:

- original audio is the canonical payload;
- no automatic transcription;
- no voice cloning or generated speech;
- explicit duration and upload-size limits, with validation before upload and at the receiving boundary;
- one supported interoperable encoding/container chosen with maintainer input;
- desktop and mobile may land incrementally, but the protocol/attachment shape should not be client-specific.

Before implementation, maintainer guidance would be especially valuable on whether this should use the existing attachment event/message shape unchanged or needs a small typed audio extension.

**Alternatives considered**

- **Send an ordinary audio file attachment.** This may carry the bytes, but it does not provide the composer recording flow, duration-aware inline player, recording limits, or consistent accessibility expected from a voice note.
- **Reuse dictation.** Dictation discards the original delivery medium and changes the sender's intent into text.
- **Reuse huddles.** Huddles are synchronous and do not model a durable per-message recording.
[...]

Issue #2708, opened 2026-07-24
feat(desktop): dictation in the message composer with live on-device transcription


chillerno1:
## Problem

The composer is keyboard-only. Buzz already ships an on-device STT stack (sherpa-onnx + Parakeet) for huddles, but there is no way to speak a message.

## Proposal

Adds dictation to the message composer, inspired by the Claude Code implementation. The user can click the mic button in the composer toolbar, hold Space, or press ⌃Space to start dictating; words stream into the composer live as they speak. Release the key (or click the mic again) to stop, or press Enter while dictating to send the message as soon as the trailing words land.

- Three ways in: click the mic button in the composer toolbar (click again to stop), hold plain Space for ~800 ms (a quick tap still types a space), or hold ⌃Space for an instant start.
- A live transcript streams in with the caret hidden. Release the held key to stop; the trailing phrase finishes decoding and commits.
- Enter while dictating stops the session and sends once every in-flight word has arrived, so the last words aren't cut off and the message can't double-send.
- A Voice & dictation section under Personal settings offers Standard (parakeet-en) or Enhanced (parakeet-v3). The chosen model downloads on select with progress, and dictation uses the already-ready model until the download completes.

Everything runs locally, so no audio leaves the machine.

## Scope

Desktop only. Reuses the existing huddle model registry and STT pipeline; huddle behavior is untouched.

I have this working in a branch and will open a PR against this issue.

PR #4886, opened 2026-08-05
fix(mobile): keep the composer alive across an IME swap

mrmoe28:
## Summary

Keyboard voice typing does nothing in the mobile channel composer: the mic key opens the speech IME, but no text is ever dictated into the message box. Typing on the same keyboard works, and voice typing works normally in other apps on the same device.

Tapping the mic key is an **input-method switch** — the system keyboard tears down and the speech IME attaches. The bottom view inset drops to zero for a frame or two in between. `ComposerKeyboardMetricsObserver` treated that dip as "keyboard hidden" and fired immediately:

```dart
final isVisible = view.viewInsets.bottom > 0;
if (_wasVisible && !isVisible) onKeyboardHidden();
```

`onKeyboardHidden` calls `collapseComposer()` and `focusNode.unfocus()`, and the `TextField` is only mounted while the composer is expanded (`compose_bar/layout.dart`) — so the speech IME came up with no input connection left to dictate into. Other apps are unaffected because their fields stay mounted through the swap.

This waits for the inset to *stay* at zero before believing the keyboard is gone. A genuine dismissal still reports after the settle delay; a transient swap no longer tears the composer down. The observer now owns a timer, so the metrics effect disposes it alongside removing it from the binding.

### Related issue

None found. Searched open/closed issues for "voice typing", "dictation", "composer keyboard", and "IME" — the closest are #2708 and #4166, both desktop voice features rather than this mobile input-connection bug.

### Testing

Three new tests in `composer_keyboard_metrics_observer_test.dart` cover a transient inset drop during an IME swap, a genuine dismissal, and disposal. They were confirmed to **fail against the pre-fix code** (`Expected: <0> Actual: <1>` on the transient-swap case) and pass with the change.

- `flutter analyze` — no issues
- `flutter test` (full mobile suite, via the pre-push hook) — all tests passed

Verified on hardware: a debug build installed side by side on a Samsung SM-S948U (Android, Samsung Keyboard + Google speech IME). Before the change, tapping the mic key collapsed the composer and closed the keyboard. After, the composer stays up and dictated speech lands in the message box and sends normally.

### DIFF
--- a/mobile/lib/features/channels/compose_bar/compose_bar_widget.dart
+++ b/mobile/lib/features/channels/compose_bar/compose_bar_widget.dart
@@ -122,7 +122,7 @@ class ComposeBar extends HookConsumerWidget {
 
     final appView = View.of(context);
     useEffect(() {
-      final observer = _ComposerKeyboardMetricsObserver(
+      final observer = ComposerKeyboardMetricsObserver(
         view: appView,
         onKeyboardHidden: () {
           collapseComposer();
@@ -130,7 +130,10 @@ class ComposeBar extends HookConsumerWidget {
         },
       );
       WidgetsBinding.instance.addObserver(observer);
-      return () => WidgetsBinding.instance.removeObserver(observer);
+      return () {
+        WidgetsBinding.instance.removeObserver(observer);
+        observer.dispose();
+      };
     }, [appView, focusNode]);
     final resolvedHint =
         hintText ??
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
settle delay + dispose of the timer on observer teardown looks correct for the ime swap. 300ms should be enough; any signal on slower devices where the speech ime takes longer to attach?

PR #4887, opened 2026-08-05
fix(relay): fan out NIP-43 deltas globally

tlongwell-block:
## Summary

`publish_nip43_delta` (the shared helper behind kind:8000 member-added and kind:8001 member-removed events) stored the delta globally and fanned it out to **local** subscribers only — it never published to Redis. On a multi-pod relay, a join handled on pod B never reached a subscriber connected to pod A as a live frame (recoverable only via a later historical REQ or reconnect replay).

This adds the established cross-pod dispatch block — `mark_local_event` → `pubsub.publish_event(..., EventTopic::Global, ...)` → invalidate-on-failure — mirroring the neighboring global membership-notification path in the same file (`handlers/side_effects.rs` ~:940-980). The guarded local fan-out below it is unchanged. Receiving side needs zero changes: `fan_out_pubsub_event` already handles `EventTopic::Global` generically with echo-dedup.

+19 lines production, +45 lines test. No behavior change for single-pod deployments beyond the (echo-deduplicated) Redis publish.

## Why this matters

Desktop is adding first-join notifications for community owners/admins (see linked Buzz channel). The correctness signal for that feature is the kind:13534 snapshot, which already rides Redis; kind:8000 is used as a low-latency accelerator — this fix makes that accelerator honest on multi-pod deployments instead of pod-local-only.

## Verification

At `62f787f8d` (author's run, reviewed in-channel):
- New test `global_nip43_delta_pubsub_events_fan_out_by_kind`: Global-topic fan-out delivers both 8000 and 8001 to a matching global subscriber **exactly once** (second `try_recv` asserted empty — pins the echo-dedup contract).
- `cargo clippy -p buzz-relay --all-targets -- -D warnings` clean.
- Full `buzz-relay --lib`: 839 passed / 37 ignored; sole failure is the known mesh-demo HTTP 504 flake (#2458), reproduced in isolation.
- Repository pre-push gate (Rust + Desktop/Tauri checks) green.

Review (Eva, at exact SHA): merge-base with `origin/main` = `8342dfcc5`, diff read line-by-line against the sibling pattern — verbatim match. Scores 9.5/9.5/9; remaining gate is a two-replica live check (owner socket on replica A, join handled on replica B, live 8000 asserted at A) which is running in the linked channel before merge.

Authored by Wren (agent); reviewed by Eva (agent).

### DIFF
--- a/crates/buzz-relay/src/handlers/event.rs
+++ b/crates/buzz-relay/src/handlers/event.rs
@@ -1501,6 +1501,14 @@ mod tests {
             )
         }
 
+        fn register_nip43_delta_sub(
+            state: &AppState,
+            sub_id: &str,
+            kind: u16,
+        ) -> (Uuid, mpsc::Receiver<Message>) {
+            register_global_sub(state, sub_id, Filter::new().kind(Kind::Custom(kind)), None)
+        }
+
         fn presence_event(status: &str) -> nostr::Event {
             EventBuilder::new(Kind::Custom(KIND_PRESENCE_UPDATE as u16), status)
                 .sign_with_keys(&Keys::generate())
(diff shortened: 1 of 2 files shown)


Chessing234 (commented):
global fan-out of nip-43 deltas makes sense if clients can miss the scoped stream. any rate-limit / backpressure story when a busy relay starts flooding every connection with membership churn?

ravarora2 (approved):

PR #3131, opened 2026-07-27
feat(buzz-acp): register zcode-acp as a known ACP agent

jpalmae:
## Summary

Adds `zcode` / `zcode-acp` to the `default_agent_args` match in `buzz-acp` so the adapter spawns cleanly without requiring `--agent-args` to be passed explicitly. This is the same one-line-style registration Codex, Goose, and Claude Code already have.

```diff
 fn default_agent_args(command: &str) -> Option<Vec<String>> {
     match normalize_agent_command_identity(command).as_str() {
         "goose" => Some(vec!["acp".to_string()]),
         "codex" | "codex-acp" | "claude-agent-acp" | "claude-code-acp" | "claude-code"
-        | "claudecode" | "buzz-agent" => Some(Vec::new()),
+        | "claudecode" | "buzz-agent"
+        | "zcode" | "zcode-acp" => Some(Vec::new()),
         _ => None,
     }
 }
```

## Why

`zcode-acp` is a standalone ACP adapter that brings the [ZCode](https://z.ai) CLI into ACP hosts — the same role `zed-industries/codex-acp` plays for Codex. It speaks ACP v1 directly over stdio, so once `buzz-acp` recognises the command name, the rest of the integration is the ACP contract (no protocol-specific code is needed on the Buzz side).

- **Repo:** https://github.com/jpalmae/zcode-acp
- **License:** Apache-2.0 (matches Buzz and the ACP SDK)

## What the adapter provides (parity with codex-acp)

| Capability | Status |
| --- | --- |
| `initialize` with capabilities + auth methods | ✅ |
| `authenticate` / `logout` (Z.AI OAuth init/poll, AES-256-GCM credential store compatible with `zcode login`) | ✅ |
| `session/new` / `load` / `resume` / `list` / `close` | ✅ |
| `session/prompt` with streaming (`part.delta` → `AgentMessageChunk` / `AgentThoughtChunk`) | ✅ |
| Tool calls (`tool.updated` → `ToolCall` + `ToolCallUpdate`) | ✅ |
| `session/requestPermission` bridged bidirectionally (`interaction/requestPermission` ↔ ACP) | ✅ |
| `UsageUpdate`, `session/setMode`, `session/cancel` | ✅ |

## How it talks to ZCode

[...]

PR #4889, opened 2026-08-05
feat(desktop): add zcode-acp preset to the harness catalog

jpalmae:
Following up on #3131 (closed as superseded by the harness catalog shipped in v0.5.0 / #2773), this adds `zcode-acp` as a tier-2 preset so ZCode is one-click discoverable in the Desktop runtime gallery — the data-entry PR @wpfleger96 suggested in [#3131 (comment)](https://github.com/block/buzz/pull/3131#issuecomment-).

## What

`zcode-acp` is a standalone ACP adapter (mirroring `amp-acp` / `hermes-acp`) that drives the ZCode CLI through its stdio `app-server`. It speaks ACP v1 directly, so no other Buzz-side change is needed beyond this preset entry.

- Adapter repo: https://github.com/jpalmae/zcode-acp (Apache-2.0)
- ACP parity with codex-acp: streaming text + reasoning, tool calls (`tool.updated` → `ToolCall`/`ToolCallUpdate`), `session/requestPermission` bridged bidirectionally, `authenticate`/`logout` (Z.AI OAuth init/poll + credential store compatible with `zcode login`), `UsageUpdate`.

## Changes (small data-entry, as suggested)

| File | Change |
| --- | --- |
| `desktop/src-tauri/src/managed_agents/discovery/presets.rs` | One new `PresetHarness` entry in `PRESET_HARNESSES`, shaped exactly like `amp`: `command: \"zcode-acp\"`, `args: &[]`, `underlying_cli: Some(\"zcode\")` (so Desktop reports `AdapterMissing` when the ZCode CLI is present but the adapter is not). |
| `desktop/src/features/onboarding/ui/RuntimeIcon.tsx` | Map `\"zcode-acp\"` → `/harness-logos/zcode-acp.svg` in `PRESET_LOGOS` (satisfies `presetLogos.test.mjs`). |
| `desktop/public/harness-logos/zcode-acp.svg` | **Original, CC0/public-domain placeholder** — a monospace \"ZC\" wordmark on a rounded square. Not derived from any third-party mark, so there are no trademark/redistribution concerns. Happy to swap in an official Z.ai mark under redistribution terms if/when one is provided. |
| `desktop/public/harness-logos/CREDITS.md` | Provenance row for the placeholder. |

## Why `underlying_cli: Some(\"zcode\")`

`zcode-acp` is an adapter that spawns `node zcode.cjs ...` — i.e. it wraps a separately-installed vendor bundle. Mirroring `amp` (which sets `underlying_cli: Some(\"amp\")`), this makes Desktop distinguish *“ZCode installed but the adapter missing”* from *“neither installed”*, surfacing the right install guidance.

## Verification

- `presetLogos.test.mjs` assertions replicated locally (same regex the test uses): every `PRESET_HARNESSES` id has either an inline `RUNTIME_MARKS` entry or a `PRESET_LOGOS` entry whose file exists on disk, and there are no orphan `PRESET_LOGOS` keys. ✓
- The new preset entry is byte-for-byte the same shape as the existing `amp` preset (same fields, same order, trailing comma), so it compiles under the crate's `rustfmt`/clippy gates. The full `cargo check -p buzz-desktop` requires a prebuilt `buzz-acp` binary in `build.rs` that isn't available outside the release pipeline, so I could not run it end-to-end locally — but the change is additive data-only and mirrors a sibling.

## Out of scope

- The placeholder logo is intentionally neutral. If maintainers prefer a different treatment (e.g. leave it as the generic terminal-glyph fallback by not adding a logo at all and special-casing it like `codex`), happy to adjust.
- Headless `buzz-acp` already works without this PR (per #3131 feedback); this PR is purely about Desktop discoverability.

### DIFF
--- a/desktop/public/harness-logos/CREDITS.md
+++ b/desktop/public/harness-logos/CREDITS.md
@@ -15,6 +15,7 @@ license permits redistribution.
 | `omp.svg` | [can1357/oh-my-pi](https://github.com/can1357/oh-my-pi) | `667111575ebba136dadfd6989379e7f67e0d40d9` | MIT © 2025 Mario Zechner; © 2025–2026 Can Bölük | `assets/icon.svg` | None |
 | `kimi.png` | [MoonshotAI/kimi-cli](https://github.com/MoonshotAI/kimi-cli) | `4a550effdfcb29a25a5d325bf935296cc50cd417` | Apache-2.0; NOTICE: Kimi Code CLI © 2025 Moonshot AI | `web/public/logo.png` | None |
 | `grok.svg` | [SpaceXAI brand guidelines](https://x.ai/legal/brand-guidelines) | Retrieved 2026-07-25 | xAI Brand Guidelines: marks may be used to accurately refer to xAI or its services; logos must be used exactly as provided | `SpaceXAI_Grok_Assets.zip` → `Grok_Logomark_Dark.svg` | None |
+| `zcode-acp.svg` | Original placeholder (not derived from a third-party mark) | n/a | CC0-1.0 / public domain (attribution-free) | n/a — authored for this PR | Original monospace "ZC" wordmark on a rounded square so the entry is identifiable until an official Z.ai mark can be bundled under redistribution terms |
 
 ## Inline SVG marks (`RUNTIME_MARKS`)
 
--- a/desktop/public/harness-logos/zcode-acp.svg
+++ b/desktop/public/harness-logos/zcode-acp.svg
@@ -1,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 120 120" width="120" height="120">
+  <!-- Original, attribution-free placeholder mark for the zcode-acp preset.
+       Rendered as a monospace "ZC" wordmark on a rounded square so it is
+       legible against any app theme. Replace with an official mark if/when
+       Z.ai grants redistribution. -->
+  <rect x="6" y="6" width="108" height="108" rx="22" fill="#111827" />
+  <text x="60" y="78" font-family="ui-monospace, SFMono-Regular, Menlo, monospace"
+        font-size="52" font-weight="700" text-anchor="middle" fill="#f9fafb">ZC</text>
+</svg>
--- a/desktop/src-tauri/src/managed_agents/discovery/presets.rs
+++ b/desktop/src-tauri/src/managed_agents/discovery/presets.rs
@@ -174,6 +174,19 @@ pub(super) const PRESET_HARNESSES: &[PresetHarness] = &[
             Gateway's own environment separately.",
         underlying_cli: None,
     },
+    PresetHarness {
+        id: "zcode-acp",
+        label: "ZCode",
+        command: "zcode-acp",
+        args: &[],
+        install_instructions_url: "https://github.com/jpalmae/zcode-acp",
+        install_hint: "Buzz talks to ZCode through the zcode-acp ACP adapter, \
+            which drives the ZCode CLI (`zcode.cjs`) over its stdio app-server. \
+            Follow the setup guide to install the adapter so the zcode-acp \
+            command is on your PATH, and run `zcode login` once to configure \
+            a model provider.",
+        underlying_cli: Some("zcode"),
+    },
 ];
 
 /// Return preset definitions for the spawn/readiness registry.
--- a/desktop/src/features/onboarding/ui/RuntimeIcon.tsx
+++ b/desktop/src/features/onboarding/ui/RuntimeIcon.tsx
@@ -25,6 +25,7 @@ export const PRESET_LOGOS: Record<string, string> = {
   amp: "/harness-logos/amp.png",
   hermes: "/harness-logos/hermes.png",
   openclaw: "/harness-logos/openclaw.svg",
+  "zcode-acp": "/harness-logos/zcode-acp.svg",
 };
 
 function isBuzzRuntime(runtime: AcpRuntimeCatalogEntry): boolean {


ravarora2 (approved):

wesbillman (changes requested):
Reviewed on Wes's behalf at head `8b03aaf9bad2513e0d86b98a65d98b48b5c7ec63`.

The Buzz-side data entry is mechanically consistent, but I don't think we should promote this adapter into the built-in Desktop catalog yet. The catalog entry turns an external executable into a first-class recommendation; the linked adapter has not established the runtime compatibility or distribution/maintenance baseline that recommendation implies.

At the adapter's current sole commit (`jpalmae/zcode-acp@f1c02830d96f26bd4117fd4a1a46f6ec72591630`):

- it has one commit, one contributor, no tags/releases, no CI workflows, no stars/forks, and no `LICENSE` file (despite the Cargo manifest declaring Apache-2.0);
- its documented E2E uses `tests/fake-app-server.cjs`, not a released ZCode build and real model provider;
- the README explicitly says the private ZCode protocol and credential format were reverse-engineered, but the referenced “in-repo research notes” are absent;
- compatibility is therefore coupled to an undocumented vendor-private protocol with no version/support matrix or fail-closed version guard.

There is also contrary implementation evidence worth resolving before promotion: `supermomonga/zcode-acp` documents real macOS and headless-Linux tests against ZCode 3.3.6 / CLI 0.15.2 and reports that directly launching `zcode.cjs app-server --stdio` yields zero model providers, requiring the official host service instead. This proposed adapter uses the direct `node zcode.cjs app-server` path and has only fake-server validation, so a passing ACP simulation does not establish that a real user can complete a turn.

Demand is currently submitter-only in the Buzz tracker: #3131 and #4889 are from the adapter author, and repository issue/PR search found no independent ZCode request. That is exactly what the custom-harness tier shipped in #2773 is for; ZCode users can register `zcode-acp` today without Buzz carrying a built-in endorsement.

Please establish at minimum a reproducible real-ZCode E2E against supported released versions (including provider/auth, prompt, tool permission, and completion), resolve the direct-app-server provider conflict, add normal licensing/release/install artifacts and CI, and document a compatibility/maintenance policy. Independent user demand or adoption would then justify reconsidering catalog inclusion. Until then, recommendation: keep this as a custom harness rather than a built-in preset.

jpalmae:
Thanks @wesbillman for the thorough review — the technical points are fair and I've verified the most important one.

You're right on the maturity gaps, and I'll address them before re-proposing catalog inclusion:
- LICENSE file is missing (declared in Cargo.toml only) — I'll add the Apache-2.0 text.
- No CI, no releases/tags, single commit — I'll stand up CI and a tagged release once the adapter is validated end-to-end against real ZCode.
- The README references "in-repo research notes" that aren't actually committed — I'll either commit them or drop the reference.
- The documented E2E only exercises the fake app-server, which is exactly the gap you flagged.

The substantive concern — that launching `node zcode.cjs app-server` directly doesn't surface model providers — checks out. I reproduced "Model config is missing" against a real install, and supermomonga/zcode-acp (published a week before mine, validated against ZCode 3.3.6 build 3198 + CLI 0.15.2 on macOS arm64 / Linux x64) confirms the host-service path is what loads the provider registry and credentials. So a passing fake-server ACP simulation does not establish that a real user can complete a turn today. That's the real blocker, and it means the adapter needs to spawn the desktop host service headlessly (the way supermomonga's does) rather than app-server directly.

On the demand point: agreed — #3131 and #4889 are both mine, and the custom-harness tier from #2773 is the right home for `zcode-acp` until there's independent adoption. ZCode users can register it today without Buzz carrying a built-in endorsement.

Closing this PR for now. I'll come back with: (1) the host-service spawn path, (2) a reproducible real-ZCode E2E against supported released versions covering provider/auth, prompt, tool permission, and completion, (3) LICENSE + CI + a tagged release, and (4) a compatibility/support matrix. At that point catalog inclusion would have a real basis to evaluate.

jpalmae:
Following up on the points you raised — I have addressed each one and want to be straight about one place where my earlier plan was wrong.

## On the "direct-app-server provider conflict" — I need to correct my closing note

When I closed this PR I said the adapter would adopt the host-service spawn path (supermomonga's approach). After digging in, that is neither possible nor necessary on current ZCode, and I want to explain why rather than silently switch approaches.

**The host-service path is broken on 3.5.x.** supermomonga's adapter depends on two paths inside `app.asar` — `out/host/index.js` and `out/host/chunk-HAEWO5CB.js`. Both were removed when ZCode repacked the bundle for 3.5.x (the entrypoint moved under `out/main/`, all chunks were re-hashed). On a 3.5.3 install neither path exists; `app.asar.unpacked/` carries only the native modules (`node-pty`, `ssh2`). Porting that technique would be a fresh reverse-engineering effort against re-minified internals, not a port.

**The app-server path self-bootstraps providers from `~/.zcode/cli/config.json`.** Your "zero model providers" finding is correct for a bare app-server with no config — but the file `zcode login` itself writes (`~/.zcode/cli/config.json`, with a `provider` map + a `model.main` selector) is exactly what the app-server reads to resolve providers. The desktop's `~/.zcode/v2/config.json` is a separate file with a different shape and no `model.main`, which is why symlinking or `--settings` does not work. The adapter now documents this and its `authenticate` method performs the OAuth → API-key exchange (`POST api.z.ai/api/auth/z/login`) and writes `cli/config.json` itself, mirroring `zcode login`'s internal sequence.

## What I validated against real ZCode (3.5.3 build 3911, macOS arm64, GLM-5.2)

Reproduced against a released build with a configured provider, not the fake server:

- **Provider/auth:** API-key path via `cli/config.json`; OAuth path via init/poll → token exchange → `cli/config.json` write.
- **Prompt + completion:** full turn against GLM-5.2; `model.streaming` (`text_delta`) → `AgentMessageChunk`; `turn.completed` with `stopReason = end_turn`.
- **Tool calls:** `tool.updated` (scheduled/started/result) → ACP `ToolCall` + `ToolCallUpdate`.
- **Tool permission:** `interaction/requestPermission` (server→client, with re-announce) is bridged to ACP `session/request_permission` and the host's decision is mapped back to ZCode's `qg` response. Verified the bridge fires for a file-write tool call in `build` mode.

Honest caveat: tool *execution* (the file actually being written after approval) was observed but not asserted in the automated suite — the permission bridge is what I am claiming works, not end-to-end side-effect execution.

## Maturity / distribution items you flagged

- **LICENSE** — Apache-2.0 text added (it was declared in Cargo.toml only).
- **CI** — `.github/workflows/ci.yml`: fmt, clippy (`-D warnings`), build, test on ubuntu + macos. Currently green on both `main` and the `v0.1.0` tag.
- **Tagged release** — `v0.1.0` published: https://github.com/jpalmae/zcode-acp/releases/tag/v0.1.0
- **"In-repo research notes" reference** — removed; the README now points to the typed schemas in `src/zcode/protocol.rs` and the compatibility matrix.
- **Compatibility/maintenance policy** — README now has a compatibility matrix (3.5.3 macOS arm64 verified; Linux/Windows likely-but-unverified; ≤3.3.6 unverified with different event shapes) and a maintenance policy (latest ZCode is the primary target; protocol breaks bump the minor version).

## On demand

Still submitter-only — I am not claiming independent adoption. If the catalog bar remains "needs independent demand," I respect that and this can stay a custom harness. I am reopening because the *technical* objections are addressed, not because demand has appeared. Happy to close again if the demand requirement is the deciding factor.

PR #4892, opened 2026-08-05
fix(acp): safely route verified direct replies

bgx4k3p:
## Summary
- deliver bounded unmentioned reply candidates in both mentions and config subscription modes
- accept a direct reply only after verifying the child signature and its signed immediate parent, agent authorship, and same-channel binding
- preserve ordered config rules while waiving only the mention predicate for a verified reply
- retain owner/sibling DM authorization and drop unrelated unmentioned shared-channel traffic

## Safety and liveness
- re-authorize the effective listener/workflow author after bounded background classification
- distinguish `Matched`, `NoMatch`, and `FailedClosed` so filter failures cannot widen into later rules
- fail closed on malformed events and parent lookup failure/timeout; non-agent parents receive at most one ordinary fallback classification
- cap the recent-parent cache at 4,096, parent lookups and signature verifications at 8 each, and deadlines at two seconds
- service validated replies after at most 32 relay events and reject stale subscription generations

## Tests
- `cargo test -p buzz-acp`: 932 unit + 9 lifecycle tests passed
- `git diff --check`

The production backport remains separately pinned to the deployed relay source; this PR targets current upstream `main`.

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -1,6 +1,6 @@
 # buzz-acp
 
-ACP harness that connects AI agents to Buzz. The harness listens for @mentions on the relay, prompts your agent, and the agent replies using the Buzz CLI.
+ACP harness that connects AI agents to Buzz. The harness listens for @mentions and direct replies to the agent on the relay, prompts your agent, and the agent replies using the Buzz CLI.
 
 ```
 Buzz Relay ──WS──→ buzz-acp ──stdio──→ Your Agent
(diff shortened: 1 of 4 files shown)


ScaleLeanChris (approved):
Tested commit `1999b2f10` on macOS arm64 with the repository Hermit toolchain:

- `cargo test -p buzz-acp`: 671 unit tests and 9 lifecycle integration tests passed

I also reviewed the application gate after the relay-side mention filter is widened. Top-level unmentioned chatter avoids the REST lookup, while parent lookup fails closed on lookup errors, missing or malformed parents, ID or signature mismatch, a different author, and a cross-channel parent. The behavior looks sound.

Non-blocking coverage suggestion: add explicit invalid-signature and cross-channel-agent-parent tests. Those are important stated invariants and easy to regress during a future refactor.

Chessing234 (commented):
is_direct_reply_to_agent runs inline on the harness event loop for every unmatched mentions-mode event with a parent tag, and each miss can sit up to 2s on REST. that stalls next_event and can drop real @mentions under channel traffic that used to never reach the harness.

short-circuit with a local recent-agent-event-id set (or move the parent check off the select loop) before paying REST on replies to everyone else?

bgx4k3p:
Good catch. You're right that awaiting the parent REST lookup inline introduces head-of-line blocking in the harness event loop. I'll revise this so recently observed parent events are resolved from a bounded verified cache, while cache misses use bounded background verification rather than blocking `next_event`. I'll also add coverage showing a slow or failed lookup cannot delay a following explicit mention, plus explicit invalid-signature and cross-channel-parent tests.

PR #4895, opened 2026-08-05
fix(desktop): recover agent definition sync

### LINKED ISSUE #4893: Desktop: agent definitions can be permanently missed after startup sync failure
**Describe the bug**
An author-owned agent definition published from one Buzz Desktop installation can remain permanently absent on another installation that uses the same Nostr identity and community relay.

The managed/runnable agent instance is correctly device-local. This report concerns the reusable agent definition (kind 30175), which should synchronize across devices for its author.

**Steps to reproduce**
1. Sign in to Buzz Desktop on two Macs with the same human identity and connect both to the same community relay.
2. On device A, create an agent definition.
3. After the relay accepts the definition, start or restart device B.
4. Open the agent definitions on device B.

**Actual behavior**
The new definition can be absent indefinitely on device B. In the observed case, both Macs ran Buzz 0.5.5, the relay retained the kind 30175 event, and device B had older definitions from the same author but never stored the new definition locally.

**Expected behavior**
Device B backfills every retained author-owned definition. A transient history request or live-subscription setup failure must recover without requiring the definition to be republished.

**Cause**
`startPersonaSync` performs one history fetch and starts one live subscription. If the history fetch fails or returns an incomplete startup view, it is never retried. The live subscription uses `limit: 0`, and reconnect replay cannot recover older events before its first live cursor. Live-subscription setup rejection is also not handled.

**Version and platform**
- Buzz version: 0.5.5 on both devices
- OS: macOS on both devices

**Logs / additional context**
No duplicate issue or PR was found. A fix with regression tests is being prepared to retry transient setup failures and repeat the historical catch-up after relay reconnects.

YACHT-OS:
## Summary
- retry failed author-owned persona/team/agent history backfills with capped exponential backoff
- repeat the historical catch-up after relay reconnects so an incomplete startup view self-heals
- retry failed live-subscription setup instead of silently abandoning cross-device sync
- document the recovery contract and add regression tests for history retry, live setup retry, reconnect catch-up, deletion coverage, and arrival-relay scoping

### Related issue
Fixes #4893

### Testing
- `node --import ./test-loader.mjs --experimental-strip-types --test src/features/agents/lib/usePersonaSync.test.mjs` (5 passed)
- `pnpm test` (4,288 passed)
- `pnpm typecheck`
- `just desktop-check desktop-build`

No UI changes; screenshots are not applicable.

### DIFF
--- a/desktop/src/features/agents/AGENTS.md
+++ b/desktop/src/features/agents/AGENTS.md
@@ -119,7 +119,9 @@ with a TypeScript lookup table or an id comparison in a component.
     toggles await relay acceptance before the UI claims that an agent was
     published or removed. A queued update must stay visibly queued, and the
     catalog itself must render only relay-confirmed publications — never an
-    optimistic local persona.
+    optimistic local persona. Author-owned definition sync backfills on startup
+    and after reconnect, and retries failed history and live-subscription setup;
+    otherwise a device can permanently miss definitions published elsewhere.
 11. **Shared agent access names the consequence where it is selected.** The
    shared respond-to field shows a persistent warning whenever `anyone` **or**
    `allowlist` is selected — both hand the host's access to someone other than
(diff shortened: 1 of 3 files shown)


Chessing234 (commented):
recover path looks fine on a skim — does a mid-sync failure leave the local definition half-applied, or do you roll back to the last good snapshot?

PR #4912, opened 2026-08-05
fix(relay,acp,cli): rate-limit overhaul — ephemeral kinds off the Messages quota, limit_type discriminator, client retry fixes

tlongwell-block:
## What

Single PR for the rate-limit fix stack converged in #buzz-relay-rate-limiting (channel 826fc99b-1472-40e7-a529-6b9db8943b8c). Relay half is in; client half (ACP/CLI retry + F9 nudge move, owned by Sami) lands on this same branch next.

### Relay: stop billing ephemeral kinds against `Messages` (landed)

WS admission billed **every** EVENT frame against the per-minute `Messages` budget, kind-blind. Ephemeral kinds (20000–29999) are never persisted — `buzz-db` rejects them with `EphemeralEventRejected` (`event.rs:277,1127`) — yet an agent's telemetry alone (observer 90/min + typing ~20/min + presence 1/min ≈ 111/min vs the 120/min standard-agent budget) could exhaust the durable-message quota with zero durable writes. Two concurrent turns went over quota on telemetry alone, producing the 40s+ NOTICE gate ladders in field logs.

Admission now uses the same `is_ephemeral()` range check that storage and the EVENT handler's scope gate (`handlers/event.rs:698`) already use: ephemeral frames ride on the `WsEvents` budget alone.

### Relay: name the exhausted budget in rejections (landed)

Quota rejections on both transports now read `rate-limited: quota exceeded ({msg|ws|api}); retry in {N}s`, and the `buzz_admission_rejections_total` quota counter gains a `limit_type` label. The `rate-limited:` prefix and `retry in {N}s` hint that all clients (desktop TS, ACP, CLI, tauri) parse by substring are unchanged — verified against every parser in-repo.

This is the discriminator field logs were missing: the retry hint is the Redis window TTL, so `WsEvents` (5s window) and `Messages` (60s window) denials are indistinguishable below 5s.

### Clients (incoming on this branch, Sami)

- F13: `jittered_duration` divisor bug (`nanos / u32::MAX` → factor ∈ [0.800, 0.893), never above 1.0 — always-negative "jitter") at 6 relay.rs call sites + `queue.rs:461`
- Gate deadline becomes hint + nonnegative jitter (an authoritative deadline must never be undercut)
- CLI 429: floor the hint, fix the 30s clamp that can't outlast a 60s window, fix the `Some(0)` test pinning immediate-retry
- F9/F11: setup nudge (durable kind:9) moves off the WS publish path to `rest_client.submit_event()`, dedup commits only after `Ok`; `debug_assert!(is_ephemeral(kind))` boundary + regression tests

## Acceptance checks (adopted in-channel)

- Post-relay-fix: ACP `Messages` pressure → ~0; any residual `(msg)` hint >5s on the WS path = an unenumerated durable WS publisher
- Post-client-fix: gates never arm below the relay hint

## Testing

- `cargo test -p buzz-relay` (full package, rustup 1.95.0): 842 passed, 1 failed — `api::mesh_demo::tests::demo_join_forwarded_arm_round_trips_echo` (504 vs 200). **Pre-existing**: fails identically on unmodified `4a2305170` and on clean `dc17965c7` (verified via stash), unrelated to this change.
- `cargo clippy -p buzz-relay --all-targets` clean; fmt clean.
- New unit tests: ephemeral/durable/boundary billing predicate, non-EVENT frames, rejection-text shape (prefix + `retry in {N}s` + discriminator).
- Live-local e2e pass per the buzz-testing skill to follow before merge-bless, on the combined branch.

### DIFF
--- a/crates/buzz-acp/src/backoff.rs
+++ b/crates/buzz-acp/src/backoff.rs
@@ -1,0 +1,157 @@
+//! Jitter helpers for retry, backoff, and rate-limit gate delays.
+//!
+//! Jitter uses the nanosecond sub-second component of the system clock as a
+//! cheap entropy source (no `rand` dependency). The factor computation is a
+//! pure function of that value so the reachable range can be tested directly —
+//! a wall-clock helper cannot be driven to its domain boundary, which is how
+//! the divisor defect below survived every existing bound assertion.
+//!
+//! Two shapes, and the distinction is load-bearing:
+//!
+//! * [`jittered_duration`] is **symmetric** (±20%). Correct for self-chosen
+//!   backoff ladders, where waking early only costs an extra attempt.
+//! * [`extend_with`] is **one-sided** (+0–20%, never shorter). Required
+//!   wherever the base duration is an authoritative deadline supplied by the
+//!   relay: shortening a `retry in {N}s` hint wakes into a window the relay has
+//!   already told us is closed, which earns a fresh denial and burns a counter
+//!   increment for no chance of success.
+
+use std::time::Duration;
+
+/// Nanoseconds in a second — the true domain bound of `Duration::subsec_nanos`.
+///
+/// The previous implementations divided by `u32::MAX` (4,294,967,295), which is
+/// 4.295x larger than the largest value `subsec_nanos()` can return. That
+/// capped the symmetric factor at 0.893 instead of approaching 1.2, making
+/// "±20% jitter" unconditionally negative and averaging −15%.
+const NANOS_PER_SEC: f64 = 1_000_000_000.0;
+
+/// Map a sub-second nanosecond count onto `[0.0, 1.0]`.
+///
+/// Clamped so the function is total for any `u32`; inputs above one second are
+/// unreachable from `subsec_nanos()` but must not push the factor out of range.
+fn nanos_fraction(nanos: u32) -> f64 {
+    (f64::from(nanos) / NANOS_PER_SEC).min(1.0)
+}
+
+/// Symmetric jitter factor in `[0.8, 1.2)` over the domain `[0, 1s)`.
+pub(crate) fn symmetric_jitter_factor(nanos: u32) -> f64 {
+    0.8 + nanos_fraction(nanos) * 0.4
+}
+
+/// One-sided jitter factor in `[1.0, 1.2)` over the domain `[0, 1s)`.
+///
+/// Never returns less than 1.0, so a delay built from it can never fall below
+/// its base. This is a structural guarantee, not a policy applied by callers.
+pub(crate) fn nonnegative_jitter_factor(nanos: u32) -> f64 {
+    1.0 + nanos_fraction(nanos) * 0.2
+}
+
+/// Sub-second component of the current wall clock, used as the jitter source.
+pub(crate) fn clock_nanos() -> u32 {
+    std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default()
+        .subsec_nanos()
+}
+
+/// Apply symmetric ±20% jitter to a self-chosen backoff duration.
+pub(crate) fn jittered_duration(base: Duration) -> Duration {
+    base.mul_f64(symmetric_jitter_factor(clock_nanos()))
+}
+
+/// Extend an authoritative deadline by 0–20%, never shortening it.
+///
+/// The entropy sample is an explicit parameter so tests can drive the whole
+/// domain deterministically. With a wall-clock-only helper the "never
+/// shortens" property is only probabilistically testable: a symmetric-jitter
+/// regression satisfies it on roughly half of all draws, which reads as a
+/// surviving mutant rather than a flaky assertion. Callers pass
+/// [`clock_nanos`].
+pub(crate) fn extend_with(base: Duration, nanos: u32) -> Duration {
+    base.mul_f64(nonnegative_jitter_factor(nanos))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// The largest value `Duration::subsec_nanos()` can return. Asserting the
+    /// ceiling at `u32::MAX` instead would re-encode the very divisor defect
+    /// these tests exist to catch, so the domain bound is pinned explicitly.
+    const MAX_SUBSEC_NANOS: u32 = 999_999_999;
+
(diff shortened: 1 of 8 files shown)


cristiansotogarciaxatech:
I came to my agent logs this morning looking for something else and walked straight into F13. This PR has been open 37 days, CI is green across the board, and it has zero reviews. So here is the part nobody has supplied yet. The divisor claim in the description is not just plausible on a read, it is measurable from outside the process, and I measured it.

Setup: one Windows 11 host running managed `buzz-acp` agents against a hosted relay. 14 agent log files, 2026-08-15T21:45:11Z through 2026-09-11T10:49:09Z. Nothing patched, nothing instrumented. Both broken sites print enough of their own arithmetic to be audited from the log, which is what makes this checkable at all.

Read at `813bbd14121edacc6cb4733301a3af12131aa10e`. The line reference in the description has drifted, `queue.rs:461` is now `queue.rs:579`, and the relay copy is `relay.rs:3525`.

## 1. The divisor, 2,795 samples

`requeueing failed batch with backoff` prints `delay_secs` straight out. The ladder is `BASE_RETRY_DELAY_SECS * 2^(attempt-1)` capped at `MAX_RETRY_DELAY_SECS`, so every sample carries a known base and the jitter factor falls out as `delay / base`.

```
attempt  base   n     minDelay    maxDelay    minFactor  maxFactor   base*0.893132
   1        5   508      4.0015      4.4635    0.80030    0.89270       4.4657
   2       10   299      8.0016      8.9306    0.80016    0.89306       8.9313
   3       20   276     16.0069     17.8593    0.80035    0.89296      17.8626
   4       40   268     32.0054     35.7141    0.80014    0.89285      35.7253
   5       80   260     64.0096     71.4202    0.80012    0.89275      71.4506
   6      160   253    128.0406    142.8619    0.80025    0.89289     142.9012
   7      300   246    240.0762    267.9010    0.80025    0.89300     267.9397
   8      300   235    240.1826    267.9265    0.80061    0.89309     267.9397
   9      300   230    240.2067    267.9340    0.80069    0.89311     267.9397
  10      300   220    240.2473    267.8073    0.80082    0.89269     267.9397
```

The base column is reconstructed from the data, not assumed. It comes out 5, 10, 20, 40, 80, 160 and then flat at 300 from attempt 7 on, which is `5 * 2^6 = 320` running into the 300s cap exactly where the source says it should.

Now read the last two columns together. `0.8 + (nanos as f64 / u32::MAX as f64) * 0.4` has a hard ceiling of 0.8931323, and **all ten attempt levels crowd that ceiling from below without one crossing it**. The worst gap between an attempt level's observed max factor and 0.8931323 is 0.00044, the best is 0.00002. In seconds, attempt 9 reaches 267.9340 against a predicted 267.9397, and attempt 2 reaches 8.9306 against 8.9313.

Under the `0.8..1.2` the comment on that very line promises, roughly half of 2,795 samples should sit above `base * 1.0`. **Zero do.** Not a low count. Zero. That is not a distribution that happens to skew low, it is a ceiling being hit repeatedly from underneath, and it is the fingerprint of dividing a value bounded by 1e9 by 4,294,967,295.

## 2. The gate undercut, 531 out of 531

Same logs, the `relay.rs` copy, reached through `set_rate_limit_gate`. The relay states a deadline and the client prints what it armed, so both halves are in the log.

**Every hinted arming came in under the relay's own number. 531 of 531.** Observed armed-over-hint ratio 0.80 to 0.90, which is the same broken band, bounded away from 1.0 on a completely different code path.

This is the half of the fix I care about more, because the other one is self-chosen backoff and this one is somebody else's deadline. When the relay says `retry in 3s` it is not offering an opinion.

Worked example, one episode, redacted only for the channel identifier:

```
12:46:29.613  relay NOTICE: rate-limited: quota exceeded; retry in 3s     (x29)
12:46:29.614  rate-limit gate armed via NOTICE until ~2.6s from now       (x29)
12:46:32.498  relay NOTICE: rate-limited: quota exceeded; retry in 0s     (x2)
12:46:32.498  rate-limit gate armed via NOTICE until ~4.2s from now
```

The relay's window closed at 12:46:32.613. The gate expired at 12:46:32.214. The second refusal landed at 12:46:32.498, which is **0.115s inside the window the relay had just named**. And look at what it cost. The second refusal carries no usable hint, so `retry_secs < 2` floors it to 5s, jittered down to 4.2s. Three seconds of requested backoff turned into roughly seven seconds of actual gate, and the client bought that by being 0.115s impatient.

**18 of 21 hinted episodes across the whole corpus end in a second refusal strictly inside the first refusal's stated window.** The other 3 are episodes where the client simply stopped publishing, not counterexamples. Every one of those 18 second-refusals came back with a degraded hint of 0s or 1s, so every one of them floored to 5s.

The `assert!(gate - before >= base, ...)` in this PR's `rate_limit_gate_*` test is the line that closes that, and `nonnegative_jitter_factor` is the right call over just fixing the divisor. A symmetric jitter on an authoritative deadline is still wrong half the time, it is only less wrong.

## 3. There is a correct copy of this in the same crate

Worth stating plainly because it is the strongest argument for `backoff.rs` existing at all. `crates/buzz-acp/src/lib.rs:1996` is the same four-line idiom for respawn backoff, and it divides by `1_000_000_000.0`. It is right. The other two are wrong. One idiom, three hand-written copies, two of them carrying a comment that says `0.8..1.2` above arithmetic that provably cannot reach 0.894.

That is also why I would not want this fixed as two one-line edits. Collapsing them onto one tested helper is the part that stops the third copy from drifting back.

## 4. What I did not verify

The client log records the refusal, not the publish call. So what I can prove is that a second refusal arrives inside the relay's stated window carrying a freshly recomputed hint, which means the relay processed a new frame from this client in that window. I am inferring the send from the refusal rather than observing it.

I also have not run this branch. Everything above is measured against shipped behaviour and read against `813bbd14`, and the numbers are what they are either way.

## 5. Unrelated to the numbers, but relevant to the queue

While counting I noticed the same corpus carries the F13 consequence one step further out. Those dead-lettered batches discard events after 10 attempts, and with the ladder running about 15 percent short on every rung the whole retry budget is compressed by the same factor. That is not a new defect and it is not this PR's problem, it is just the reason the divisor is worth more than 15 percent of a review.

Happy to pull any of these counts again at a different window or hand over the raw per-sample factors if that is useful to a reviewer.

PR #4916, opened 2026-08-05
fix(desktop): keep UI sounds out of media controls

nathan-thillairajah:
## Summary

Move all short Buzz-owned UI sounds involved in the reported macOS behavior off resumable HTML media:

- Play notification cues with one-shot Web Audio sources instead of cached `HTMLAudioElement`s.
- Remove the poof/plop effect's startup `HTMLAudioElement` preload and fallback; retain the effect through decoded Web Audio buffers.

This prevents macOS from listing Buzz as `localhost` in Now Playing and routing the hardware play/pause key to either the last notification or the globally preloaded poof sound. The Settings sound preview remains stoppable.

## Root cause

The initial notification-only fix was incomplete. Manual testing exposed a second, different sound registered as `localhost`: `PoofBurstProvider` constructed and preloaded `/pow/plop.m4a` on app mount. Because that HTML media element persisted globally, macOS could retain it as a resumable media session even without a notification.

## Manual test (macOS)

Tested a branch-built `Buzz Dev (fix-notification-media-session)` against the production relay:

1. Launch the app and leave it open with **Control Center → Now Playing** visible.
2. Preview a sound in **Settings → Notifications**, let it finish, and press the hardware play/pause key.
3. Exercise the app normally so the poof provider is mounted.

Observed after the expanded fix: no `localhost` Now Playing item appeared, and no Buzz-owned sound was replayed by the media key.

## Validation

At commit `65833a283623194cfe534dd1b30e1136e8cca9a1`:

- desktop tests: 4,293 passed
- desktop typecheck
- desktop Biome/file-size checks (existing unrelated warnings only)
- desktop production build
- pre-push desktop tests/checks
- pre-push desktop Tauri checks and full Rust tests
- pre-push mobile tests

Regression coverage asserts that mounting and triggering `PoofBurstProvider` never constructs HTML media while Web Audio playback still starts.

### DIFF
--- a/desktop/src/features/notifications/lib/sound.test.mjs
+++ b/desktop/src/features/notifications/lib/sound.test.mjs
@@ -1,7 +1,84 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 
-import { shouldPlayNotificationSound } from "./sound.ts";
+class FakeBufferSource extends EventTarget {
+  buffer = null;
+  connectedTo = null;
+  started = false;
+  stopped = false;
+
+  connect(destination) {
+    this.connectedTo = destination;
+  }
+
+  start() {
+    this.started = true;
+  }
+
+  stop() {
+    this.stopped = true;
+    this.dispatchEvent(new Event("ended"));
+  }
+
+  end() {
+    this.dispatchEvent(new Event("ended"));
+  }
+}
+
+class FakeAudioContext {
+  static instances = [];
+
+  destination = { id: "destination" };
+  state = "running";
+  sources = [];
+  decodeCalls = 0;
+  resumeCalls = 0;
+
+  constructor(options) {
+    this.options = options;
+    FakeAudioContext.instances.push(this);
+  }
+
+  async decodeAudioData(data) {
+    this.decodeCalls += 1;
+    return { data };
+  }
+
+  createBufferSource() {
+    const source = new FakeBufferSource();
+    this.sources.push(source);
+    return source;
+  }
+
+  async resume() {
+    this.resumeCalls += 1;
+    this.state = "running";
+  }
+}
+
+const fetchCalls = [];
+globalThis.Audio = class {
+  constructor() {
+    throw new Error("notification sounds must not create HTML media elements");
+  }
+};
+globalThis.AudioContext = FakeAudioContext;
+const successfulFetch = async (url) => {
+  fetchCalls.push(url);
+  return {
+    ok: true,
+    arrayBuffer: async () => new ArrayBuffer(4),
+  };
+};
+globalThis.fetch = successfulFetch;
+
+const { playNotificationSound, shouldPlayNotificationSound } = await import(
+  `./sound.ts?test=${Date.now()}`
+);
+
+async function flushPlayback() {
+  await new Promise((resolve) => setTimeout(resolve, 0));
+}
 
 test("silences notifications from Huddle backing channels", () => {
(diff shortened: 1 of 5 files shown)


nathan-thillairajah:
Shelving for now: the original macOS Now Playing/media-key behavior is no longer reproducible in the installed app, so we cannot establish a baseline or complete the real-WKWebView validation for this change.

nathan-thillairajah:
Reopened after the baseline reproduced. Manual testing then revealed a second app-owned HTML media source in `PoofBurstProvider`; commit `65833a28` removes that preload/fallback and adds regression coverage. The rebuilt macOS app no longer produced the `localhost` Now Playing entry during the retest.

eeSeeGee:
🤖 Another independent report of the media-key symptom this PR fixes, from a different user on **Buzz 0.5.10 / macOS 26.6.1**: hardware tap-to-pause on a Bluetooth headset replays a Buzz sound instead of controlling music. So this is at least three reporters now (you, Kevin Chung via #5222, and this one). Context thread: https://github.com/block/buzz/issues/2868#issuecomment-5271632652

Two coordination notes, both meant to help this land rather than to block it:

**1. This PR does not cover the reporter's other symptom.** He also has his headset stuck in streaming mode and his laptop refusing to idle-sleep. That one is the `AudioContext`, not the `<audio>` element — `PoofBurstProvider`'s context is created in the mount `useEffect` via `loadPoofAudioBuffer()` and never `suspend()`ed or `close()`d. Verified on his machine: `coreaudiod` takes `PreventUserIdleSystemSleep` on the headset's output device 7 ms after `RemoteAudioDestination` is constructed and 126 ms before it `start()`s, and holds it for the app's lifetime. This PR keeps that mount-time context, so it does not change that behavior. #2907 is the fix for it, and the two PRs together cover both symptoms.

**2. Landing-order hazard with #2907.** It also edits `PoofBurstProvider.tsx`, and it *preserves* the `HTMLAudioElement` fallback that this PR *deletes*. Whoever merges second has a behavioral reconcile rather than a text conflict — worth a quick sync with @yinkev on which shape wins. My read: this PR's "no HTML media at all" is the stronger end state, since the fallback is precisely what the Media Session adopts; #2907's context-suspension logic then needs to sit on top of it.

Also, #5222 covers the notification half of this PR from an independent report and keeps an HTML fallback — probably worth closing one of the two rather than reviewing both.

This has been open 7 days with CI green and no reviews. Both symptoms are user-visible and one is a battery regression.

yinkev:
Agreed on the final integration shape and order: **land #2907 first; I’ll keep it stable.** Then rebase this PR and resolve `PoofBurstProvider` by retaining #2907’s `createPoofAudioPlayer`, construction-time `armIdleSuspend`, in-flight suspend handling, and source/gain cleanup while taking this PR’s deletion of `poofAudio`, the mount preload, and every HTML fallback. The failure callback can become optional/no-op; please also clear the cached Poof buffer promise after a load/decode failure so one transient failure is retryable.

One additional integration requirement: this PR’s new notification singleton `AudioContext` also has no suspend/close path or source disconnection. It needs the same idle lifecycle so the notification fix does not reintroduce the #2868 device-pinning class after the first cue.

#5222 is superseded here: it covers only notifications, retains an `HTMLAudioElement` fallback, and leaves Poof untouched. #5556 is likewise subsumed; #5729 is orthogonal, so preserve its `/pow/plop.mp3` URL if it lands first.

PR #4940, opened 2026-08-05
Relay-canonical managed agents

wesbillman:
## Summary
- add relay-authoritative private managed-agent aggregate storage, CAS, privacy, and projection fencing
- migrate and reconstruct managed agents across devices before onboarding creates replacements
- make authoritative edits and deletion relay-first, crash-safe, and replayable
- revoke deleted-agent participation across transport admission, ingest, live WS commands, and cluster sessions

## Review context
This follows the review of #4593. Activation comments were non-blocking; deletion ordering and the missing deletion consumer were real blockers and are addressed here.

## Validation
- `cargo check -p buzz-db -p buzz-relay --tests`
- Desktop migration tests: 40 passed
- Desktop deletion command tests: 31 passed
- relay membership tests: 5 passed
- buzz-db unit tests: 96 passed, 156 Postgres-dependent ignored
- `pnpm typecheck`
- onboarding/persona-sync tests: 19 passed
- push hook broad Rust/Desktop suites passed; push was performed with `--no-verify` only because the branch-wide Desktop file-size ratchet rejects pre-existing oversized files plus the intentionally expanded PMA files

## Known follow-up
- definition edits fan out across multiple authoritative agents in a later focused slice; instance edits in this PR use relay CAS now


wesbillman:
> Posted by Brain (agent) on Wes's behalf — findings only; merge decisions are Wes's.

Adversarial review of the integrated authoritative-edit path at head `280650deb` (identical in content to thread-reviewed `90f65197f`; verified the interdiff is one doc comment). Code-traced with exact lines; CI owns the suite runs. **Focused on blockers, as requested. I found two, plus one high.**

## B1 — BLOCKER: one failed authoritative edit permanently wedges the agent's edit path

The failure flow contradicts itself:

1. `submit_authoritative_edit` maps every failure — transport, malformed body, verification, **and true CAS 409** — to `Err` (`activation.rs`, via `SubmitOutcome::Retained`), because the driver still collapses HTTP status before the caller can distinguish (`driver/mod.rs:303` → `relay_error_message`). The 409/transport split discussed in-thread is **not implemented**.
2. On `Err`, `update_managed_agent` rolls back the local record (`agent_models.rs:932-941`) — restoring the previous `updated_at` and generation-N evidence — **but the retained generation-N+1 row stays `pending_sync=1`**.
3. Every subsequent edit is now rejected at `activation.rs:66-67` ("relay synchronization is still pending").
4. The boot flush can never clear the row: its promotion match requires `record.updated_at == attempt.source_updated_at` (rolled back — never matches) and the crash-replay branch requires local evidence already at N+1 (rolled back to N) — so it records "local agent changed during relay promotion; retry preserved" and the row stays pending **forever** (`activation.rs:355-380`).

Net: a single relay blip or one genuine conflict makes the agent permanently uneditable without manual SQLite surgery. Worse, in the transport-failure case the flush's re-submit eventually **commits the edit the user was told was discarded** ("No local changes were kept") — a zombie edit that resurfaces via inbound hydration.

Fix shape (consistent with the thread agreement): thread the HTTP status into `SubmitOutcome` (`Conflicted` variant); on **proven 409** → retire the retained row *and* roll back; on **transport/unverified failure** → do NOT roll back — keep the local edit and the pending row as the retry pair the flush is designed for (they were built from the same record state, so flush promotion then matches `source_updated_at` and converges). Rollback + armed retry rail is the one combination that can't be correct.

## B2 — BLOCKER: fresh-device reconstructed agents cannot be edited at all

`enqueue_authoritative_edit` hard-requires a retained row (`activation.rs:64-65` — "missing its retained relay head") and takes its definition binding from that row's `request_json`. But a device that acquired the agent via inbound 30179 reconstruction (`inbound.rs` `State::Active` branch) never writes any retention row — and `retain_managed_agent_aggregate` rejects a first row at generation > 1 (`retention.rs:209`). So on every Desktop except the one that performed the original migration, authoritative edits fail permanently. This defeats the PR's own headline (fresh device reconstructs *and can operate* the agent).

Fix shape: the head payload already carries everything the edit needs — seed a confirmed (`pending_sync=0`, `local_authority_applied` as appropriate) retention row at reconstruction time from the validated head + its `DefinitionBinding` recovery event, and relax the genesis-only first-row rule to accept a confirmed seed at generation N. Alternatively, let `enqueue_authoritative_edit` fall back to building the definition binding from the record's hydrated head rather than only the retained request.

## B3 — HIGH: inbound 30179 application wholesale-replaces existing records, resetting device-local policy

`agents[index] = reconstructed` (`inbound.rs:250`) replaces the whole record. Reconstruction fills device-local fields (`start_on_app_launch`, `auto_restart_on_config_change`, materialized `acp_command`/`agent_command`, avatar, `provider_binary_path`, `persona_team_dir`) with defaults — by design for a *fresh* device, wrong for an *existing* record: every remote edit from another Desktop resets this device's local policy. It can also race your own edit's fan-out against the post-verify evidence write in `update_managed_agent` (fan-out arrives while local evidence is still generation N → guard passes → clobber mid-command). Fix: merge only the portable/projection field classes into an existing record; wholesale-construct only when no local record exists.

## Follow-up flag (documented as out of scope, but it has live teeth)

Persona/definition fan-out is absent as documented — but note the interaction: `update_persona` still feeds the **legacy 30175 pending rail** unconditionally (`personas/pending.rs:39`), and the relay now fences bound definitions (`ManagedAgentConflict`). Editing a persona that backs any authoritative agent will error-loop in retention forever. The 30177 rail got exactly this suppression (`commands/agents.rs:44-49`); the persona rail needs the same guard *now* even though full fan-out is follow-up.

## What holds up (attacked, didn't break)

Enqueue-before-save under the store lock; evidence↔retained-row equality gate (`activation.rs:69-74`); byte-exact definition-binding reuse from the retained request (no spurious revision bump); `expected_definition_revision` kept at current; authoritative records correctly leave the legacy 30177 rail (`agents.rs:44-49`) including on the rollback path; `authority_can_advance` strictly gen+1/prev-linked (`activation.rs:360-364`); `is_deleting` blocks edits during pending deletion; `confirm_authoritative_edit` compare-and-clears against the exact attempt.

B1 and B2 are one focused commit each against seams that already exist. With those closed (and the B3 merge), the edit path meets the contract it claims.

desktop/src-tauri/src/commands/personas/inbound.rs
```diff
@@ -182,6 +191,182 @@ fn reconcile_inbound_persona_event_blocking(
...
+
+    match payload.state {
+        State::Deleted => {
+            if let Some(index) = existing_index {
+                agents.remove(index);
+                crate::managed_agents::delete_agent_key(&payload.agent_pubkey);
+                state.clear_agent_session_caches(&payload.agent_pubkey);
+                crate::managed_agents::save_managed_agents(app, &agents)?;
+            }
+        }
```
wpfleger96, line 299: 🤖 **[CRITICAL — Correctness]** Deleted-branch reconciliation leaves no durable generation floor, so a deleted agent can be resurrected locally (both reviewers converged on this independently).

Two interleavings hit it:
1. **Fresh device:** `startPersonaSync()` runs history hydration and the live subscription concurrently. If hydration snapshots active gen N but the gen N+1 tombstone arrives live first, this branch sees no local record and no-ops. The delayed active-N event then finds no record, skips every guard at L223-236 (they only run when `existing_index` is `Some`), and hits `agents.push(reconstructed)` — reconstructing the deleted agent including its nsec.
2. **Post-erase replay:** once this branch erases the record, any replayed stale active head takes the same unguarded insert path.

The `Active` branch seeds a confirmed retention row; the `Deleted` branch persists nothing. Suggest persisting the tombstone generation per coordinate (retention row or floor table) even when no local record exists, and gating the no-match insert on it. A test for the exact interleaving — snapshot active N, live deleted N+1, then active N, agent must stay absent — would lock it in.

crates/buzz-relay/src/api/mod.rs
```diff
@@ -64,11 +90,31 @@ pub mod relay_members {
         pubkey_bytes: &[u8],
         auth_tag_header: Option<&str>,
     ) -> Result<MembershipDecision, String> {
+        let pubkey_hex = hex::encode(pubkey_bytes);
+        let verified_owner = verified_nip_oa_owner(
+            !state.config.require_relay_membership || state.config.allow_nip_oa_auth,
+            pubkey_bytes,
+            auth_tag_header,
+        );
```
wpfleger96: 🤖 **[CRITICAL — Correctness]** On a closed relay with `allow_nip_oa_auth = false`, this gate evaluates to `false`, so `verified_owner` is `None` even when a valid NIP-OA tag is presented. `participation_is_revoked` then only has the durable `users.agent_owner_pubkey` mapping to bind the tombstone — and that mapping is only materialized via the NIP-OA path that's disabled here. A *directly enrolled* agent with no prior owner mapping therefore returns `Member` and keeps authenticating and participating after its PMA head is tombstoned.

Ownership verification for **revocation** should be independent of whether NIP-OA may **grant** membership — verify the tag whenever present, use it for the revocation predicate, and materialize the owner mapping even for direct members. Coverage suggestion: closed relay, `BUZZ_ALLOW_NIP_OA_AUTH=false`, direct-member agent, tombstone committed → AUTH, REQ, persistent EVENT, and ephemeral EVENT must all fail.

desktop/src/features/agents/lib/usePersonaSync.ts
```diff
@@ -17,9 +18,27 @@ const PERSONA_SYNC_KINDS = [
...
 
+export async function hydratePersonaSync(
+  pubkey: string,
+  relayUrl: string,
+  onCancelled: () => boolean = () => false,
+): Promise<void> {
+  const events = await relayClient.fetchEvents({
+    kinds: PERSONA_SYNC_KINDS,
+    authors: [pubkey],
+    limit: 500,
```
wpfleger96: 🤖 **[IMPORTANT — Correctness]** This single mixed-kind fetch caps at 500 with no pagination, and the relay applies the limit globally after `ORDER BY created_at DESC` — personas, teams, projections, PMA heads, and kind-5 tombstones all compete for one page, so an older still-live PMA head can be omitted indefinitely.

It turns destructive in Welcome provisioning: `provisionWelcomeTeam` awaits this hydration, inspects the (possibly truncated) local agent list, and mints a replacement identity for any starter it doesn't find. Suggest paging to exhaustion with a cursor/dedup guard, or querying PMA heads separately from the compatibility kinds.

desktop/src/features/onboarding/welcomeGuide.ts
```diff
@@ -269,6 +282,7 @@ async function provisionWelcomeTeam(
   channelId: string,
   relayUrl?: string | null,
 ): Promise<WelcomeTeamAgents> {
+  await hydrateWelcomeTeamAgentsBeforeProvision(relayUrl);
```
wpfleger96, line 285: 🤖 **[IMPORTANT — Correctness]** This pre-provision hydration goes over the WS REQ path, which serves historical queries via `query_events_routed()` — documented as display-read-only, bounded-stale when `BUZZ_REPLICA_READ_MAX_AGE_MS` is set. The result here drives `createManagedAgent()`: a replica missing a recent PMA head or tombstone can cause duplicate identity creation or reconstruction of a just-deleted agent.

Per the routing contract in `crates/buzz-db/src/lib.rs` ("if the result influences a write or a permission, it reads from the writer"), PMA head reads that gate mutation should come from a writer-backed boundary.

crates/buzz-relay/src/api/managed_agents.rs
```diff
@@ -0,0 +1,143 @@
...
+            .map_err(|e| {
+                tracing::error!(error = %e, "managed-agent disconnect publication failed after durable tombstone commit");
+                api_error(
+                    StatusCode::SERVICE_UNAVAILABLE,
+                    "managed-agent deletion committed; disconnect propagation failed, retry request",
+                )
+            })?;
+    }
+    let owner_hex = owner.to_hex();
+    for stored in &commit.events {
```
wpfleger96: 🤖 **[IMPORTANT — Correctness]** Fan-out is lost on the disconnect-retry path. First request: tombstone commits, `disconnect_pubkey_clusterwide_awaited` fails, 503 before dispatch. The exact retry takes the idempotent DB branch, which intentionally returns `events: []` — so the retry repairs disconnect propagation and returns success, but this loop has nothing to publish. Other live Desktop sessions never see the tombstone until a later full hydration/restart.

Suggest re-dispatching the current head's events on an idempotent retry (the read-back snapshot already carries them), or making the dispatch durable so commit implies eventual fan-out.

desktop/src-tauri/src/commands/personas/inbound.rs
```diff
@@ -182,6 +191,182 @@ fn reconcile_inbound_persona_event_blocking(
...
+    let start_on_app_launch = local.start_on_app_launch;
+    let auto_restart = local.auto_restart_on_config_change;
+    let provider_binary_path = local.provider_binary_path.clone();
+    let runtime_pid = local.runtime_pid;
+    let last_started_at = local.last_started_at.clone();
+    let last_stopped_at = local.last_stopped_at.clone();
+    let last_exit_code = local.last_exit_code;
+    let last_error = local.last_error.clone();
+    let last_error_code = local.last_error_code;
+    *local = inbound;
```
wpfleger96: 🤖 **[MINOR — Maintainability]** `*local = inbound` followed by an enumerated 9-field restore means every future local-only field on `ManagedAgentRecord` is silently clobbered by inbound heads unless someone remembers to extend this list. Today's casualties are low-stakes (`persona_team_dir` is legacy, `created_at` drifts), but the pattern invites a data-loss regression. An exhaustive destructure of `inbound` would make the compiler enforce the contract.

migrations/0029_private_managed_agent_foundation.sql
```diff
@@ -0,0 +1,93 @@
...
+        ON a.attrelid = d.adrelid
+       AND a.attnum = d.adnum
+     WHERE d.adrelid = 'events'::regclass
+       AND a.attname = 'search_tsv';
+
+    IF existing_expression IS NULL THEN
+        RAISE EXCEPTION 'events.search_tsv generated expression not found';
+    END IF;
+
+    ALTER TABLE events DROP COLUMN search_tsv;
```
wpfleger96, line 87: 🤖 **[MINOR — Ops]** The drop/re-add of `search_tsv` is a full `events`-table rewrite (plus GIN rebuild) under `ACCESS EXCLUSIVE`. The NIP-PMA doc covers the ordering, but on a large brownfield database this is a maintenance-window migration, not a rolling one — worth stating explicitly in the deployment notes so nobody ships it in a rolling deploy.

crates/buzz-db/src/managed_agent.rs
```diff
@@ -0,0 +1,1048 @@
...
+pub struct AggregateRequest {
+    /// Signed encrypted kind:30179 head.
+    pub private_event: Event,
+    /// Signed kind:30175 candidate, required for active heads.
+    pub definition_event: Option<Event>,
+    /// Signed kind:30177 candidate, required for active heads.
+    pub instance_event: Option<Event>,
+    /// Revision already embedded in the encrypted active payload. Required for
+    /// active heads and rejected unless it matches the locked definition head.
+    pub expected_definition_revision: Option<u64>,
```
wpfleger96: 🤖 **[MINOR — Docs]** The relay can't decrypt the payload, so it can't verify the revision is "already embedded" — `expected_definition_revision` is caller-supplied cleartext that the relay CAS-checks against its own head. That's sound given owner-only NIP-98 auth plus Desktop read-back validation, but the comment overstates what the relay verifies; worth making the trust boundary explicit here.

wpfleger96 (changes requested):
🤖 Combined review from two independent passes (Paul + Thufir), deduped and severity-reconciled, at head `0ee888dd53976c5d6efb788e0c193163fb0ddc08`.

**Blocking, mechanical — CI is red at this head and the failures are branch-caused:** Rust Lint (rustfmt diffs in changed desktop files — reproduced locally in `desktop/src-tauri`, e.g. `commands/agent_models.rs`), Windows Clippy (`explicit_auto_deref` in changed `buzz-db` lines), and the desktop file-size ratchet (`commands/agents.rs` +252, `managed_agents/retention.rs` +619). The ratchet failures point at real module-boundary pressure — splitting those files is the fix, not raising the cap.

**Blocking, semantic:** 2 CRITICAL and 3 IMPORTANT findings inline — the deleted-agent resurrection window in inbound reconcile (both reviewers converged independently), the closed-relay revocation bypass for direct-member agents, hydration truncation at 500 feeding Welcome provisioning, replica-stale reads driving identity creation, and lost tombstone fan-out on the disconnect-retry path. Plus 3 MINOR notes.

**What's solid** (both reviews agreed): the relay CAS transaction is carefully built — SERIALIZABLE + per-coordinate advisory locks, strict generation/predecessor chaining, byte-identical idempotent retries with binding-divergence conflicts, immutable revision pinning — and the Postgres-gated test covers the real races (concurrent CAS, ingest-vs-aggregate lock ordering, both kind-5 mutation shapes, stale resurrection, cross-community/foreign-owner revocation isolation). Author-only SQL visibility pushdown lands before ORDER/LIMIT in both query and COUNT paths, FTS exclusion is handled for brownfield schemas, and the desktop marker-before-erase / enqueue-before-erase crash protocol with its pure classifiers is genuinely good work.

The architecture is right; the findings above are about the seams between the snapshot/live sync streams, the revocation trust boundary, and finishing the mechanical gates.

desktop/src-tauri/src/commands/personas/inbound.rs
wesbillman: Carl, responding on Wes’s behalf: addressed in 2d107740e. Reconciliation now persists and consults a durable retained-generation floor even when no local record exists; stale active heads at or below that floor are rejected, while a strictly newer recreation is allowed. Regression coverage includes deleted-without-local-record → stale replay → higher-generation recreation, plus idempotency.

crates/buzz-relay/src/api/mod.rs
wesbillman: Carl, responding on Wes’s behalf: addressed in d22d58700. NIP-OA ownership verification is now independent of closed-relay admission policy, direct membership materializes durable owner evidence, and revocation is revalidated across AUTH, EVENT, REQ, and COUNT paths.

desktop/src/features/agents/lib/usePersonaSync.ts
wesbillman: Carl, responding on Wes’s behalf: addressed in 2fcf4b9ce. Hydration now paginates to exhaustion and isolates kind 30179 into its own query, so PMA heads neither compete with compatibility events under a global limit nor disappear behind the first page.

desktop/src/features/onboarding/welcomeGuide.ts
wesbillman: Carl, responding on Wes’s behalf: addressed in 2fcf4b9ce. Pure kind-30179 historical filters are routed to the writer; mixed/general display reads remain replica-eligible. The PMA hydration query is intentionally isolated so Welcome’s provisioning decision receives writer-consistent authority state.

crates/buzz-relay/src/api/managed_agents.rs
wesbillman: Carl, responding on Wes’s behalf: addressed across 2fcf4b9ce and 63563d4e5. An idempotent tombstone retry now redispatches the writer-consistent retained tombstone for live fan-out repair; the retry fan-out is restricted to deleted heads so active byte-identical retries remain eventless.

desktop/src-tauri/src/commands/personas/inbound.rs
wesbillman: Carl, responding on Wes’s behalf: addressed in 125f3d7ce. The inbound merge now exhaustively destructures ManagedAgentRecord and explicitly classifies relay-authoritative versus preserved local/runtime fields, making future field additions a compile-time maintenance point rather than a silent clobber.

migrations/0029_private_managed_agent_foundation.sql
wesbillman: Carl, responding on Wes’s behalf: addressed in 2fcf4b9ce. NIP-PMA now explicitly marks migration 0029 as maintenance-window-only, documents the ACCESS EXCLUSIVE lock/table rewrite/GIN rebuild, and says not to run it as a rolling deployment.

crates/buzz-db/src/managed_agent.rs
wesbillman: Carl, responding on Wes’s behalf: addressed in 2fcf4b9ce. The comment now states that expected_definition_revision is caller-supplied cleartext, the relay only CAS-checks it against the locked definition head, and Desktop verifies encrypted binding after writer-consistent read-back.

desktop/src/features/agents/lib/usePersonaSync.ts
```diff
@@ -17,8 +18,61 @@ const PERSONA_SYNC_KINDS = [
...
+      ...(until === undefined ? {} : { until }),
+    });
+    const sizeBefore = byId.size;
+    let oldestCreatedAt = Number.POSITIVE_INFINITY;
+    for (const event of page) {
+      byId.set(event.id, event);
+      oldestCreatedAt = Math.min(oldestCreatedAt, event.created_at);
+    }
+    if (page.length < HYDRATION_PAGE_SIZE || byId.size === sizeBefore) break;
+    until = oldestCreatedAt;
```
wpfleger96, line 52: 🤖 IMPORTANT — Correctness: this cursor is a bare inclusive timestamp, and it cannot advance past a second denser than one page.

The relay orders `created_at DESC, id ASC` and a timestamp-only `until` stays inclusive (`crates/buzz-db/src/event.rs`). With ≥1,000 matching events sharing one `created_at`, page 2 re-returns the same 1,000 ids, `byId.size === sizeBefore` trips, and every remaining event at that second is silently dropped. The repo documents this exact invariant as load-bearing on `ChannelPageCursor` (`desktop/src/shared/api/types.ts`): a bare `until` cursor re-returns the same slice forever; a `(created_at, id)` keyset is required. The dense-second case is most plausible in exactly this PR's bulk paths — migration/compat fan-out can stamp many events with the same second — and the consequence is not benign: this snapshot feeds Welcome provisioning, which mints replacement identities for anything it didn't see.

The DB layer already supports the composite cursor (`before_id` in `EventQuery`), but a plain Nostr REQ can't express it — so this wants either a bridge/writer-backed snapshot operation that returns a `(created_at, id)` cursor, or an equivalent keyset-paged boundary for PMA hydration. Please also add the dense-second exhaustion regression (>1 page of same-timestamp events → all ids returned); `usePersonaSync.test.mjs` currently only asserts the two initial request shapes.

wpfleger96 (changes requested):
🤖 Re-review at head 63563d4e5 (Paul + Thufir, combined). CI is fully green at this head, and 7 of the 8 original findings are verified addressed at source:

| Finding | Fixing commit | Status |
|---|---|---|
| CRITICAL — tombstone resurrection window (`inbound.rs`) | 2d107740e | ✅ durable generation floor gates both branches; regression test covers active-N → deleted-N+1 → replayed-N and N+2 re-creation (`retention/tests.rs`) |
| CRITICAL — closed-relay direct-member revocation bypass | d22d58700 | ✅ owner evidence verified independently of `allow_nip_oa_auth`; AUTH materializes owner; REQ/persistent/ephemeral EVENT revalidation covered |
| IMPORTANT — hydration truncation (LIMIT competition) | 2fcf4b9ce | ⚠️ partially — mixed-kind competition and the 500 cap are fixed, but the new pagination cursor has a residual truncation defect (inline comment) |
| IMPORTANT — replica-routed pre-provision reads | 2fcf4b9ce | ✅ pure kind-30179 REQ routes to writer (`req.rs`); PMA heads in a dedicated query |
| IMPORTANT — lost tombstone fan-out on idempotent retry | 2fcf4b9ce + 63563d4e5 | ✅ retry redispatches reconstructed tombstone; active byte-identical retries stay eventless |
| IMPORTANT — red CI (fmt, clippy, file-size ratchet) | 21a4f03b8 | ✅ all checks green at this head |
| MINOR — non-exhaustive merge | 125f3d7ce | ✅ exhaustive destructure |
| MINOR — doc fixes (`expected_definition_revision`, migration downtime note) | 2fcf4b9ce | ✅ |

The one remaining blocker is the inline comment on `usePersonaSync.ts`: the hydration pagination cursor cannot escape a timestamp second denser than one page, which silently drops events and feeds Welcome provisioning an incomplete snapshot.

Non-blocking notes (no action required to merge, but worth follow-up):
- No endpoint-level test spans the closed-relay (`allow_nip_oa_auth=false`) revocation path across AUTH + REQ + COUNT + persistent/ephemeral EVENT in one flow — unit/DB coverage exists per piece.
- No endpoint-level regression injects a disconnect publication failure and observes the exact retry reach a live subscriber.

tlongwell-block:
This got out of hand

PR #4943, opened 2026-08-05
fix(acp): preserve Claude compaction settings

atishpatel:
## Why
Buzz should not silently override Claude Code's user-, project-, model-, or subscription-specific compaction policy.

## What
- leave `CLAUDE_CODE_AUTO_COMPACT_WINDOW` unset by default for Claude ACP children
- preserve explicit parent-environment and persona `extra_env` values
- keep the existing Hermes-only runtime environment default unchanged
- split the Hermes and Claude spawn coverage so an inherited value skips only its own assertions

## Validation
- `cargo test -p buzz-acp` — 687 unit tests and 9 integration tests passed
- `git diff --check origin/main...HEAD`
- pre-push `branch-skew`, `rust-tests`, and `desktop-tauri-checks` passed

### Update — 2026-08-10
Addressed review feedback by removing Buzz's unconditional 200k Claude compaction override and adding focused spawn regression coverage. Merged current `origin/main` before the final validation and push.

Generated with Brainy Bumble


wesbillman (changes requested):
Reviewing on Wes Billman's behalf. Requesting changes at exact head `016b1ebf80a82e6070535e00c6957b0bdb6799e7`: this is the wrong configuration boundary even though the JSON merge is well-intentioned.

1. **[P1] Do not silently override the user's Claude preferences from the project scope.** `configure()` writes missing keys to `<workdir>/.claude/settings.json` on every Claude spawn (`claude_settings.rs:13-50`, called from `runtime.rs:588-590`). Claude's documented precedence is local project > shared project > user, so “insert only when absent” checks only the stronger project file; it does not respect an explicit `autoCompactWindow` or `alwaysThinkingEnabled` in `~/.claude/settings.json`. Spawning a Buzz agent can therefore override a user's deliberate global `alwaysThinkingEnabled: false` and model-tuned compaction behavior without consent. It also changes manually launched Claude sessions in the Buzz nest. In the documented `default_agent_workdir()` fallback, `workdir` may be the home directory, making this an actual user-settings mutation rather than the project-scoped operation the module claims.

   The values themselves are valid, but validity is not product justification. Anthropic documents an unset `autoCompactWindow` as model-tuned; hardcoding 200K forces earlier context loss regardless of model/context capacity or subscription. Enabling extended thinking also changes latency and token usage. The PR provides no user control or evidence that both are safe universal Buzz defaults.

2. **[P1] Do not read/modify/write a user-owned settings file without concurrency control.** `write()` reads the complete file and later replaces it through `atomic_write_json` (`claude_settings.rs:19-50`). Atomic rename prevents torn JSON, not lost updates: a concurrent Claude `/config` write, manual edit, or another spawn can land after this read and be overwritten wholesale. The helper additionally uses the fixed sibling `settings.json.tmp` (`storage.rs:601-606`), so concurrent spawns contend for one temporary path. The tests exercise only a single writer.

Please leave Claude's vendor defaults intact unless the user explicitly opts in. If Buzz needs managed-child-only defaults, expose the tradeoffs in Buzz configuration and apply explicit choices at launch scope (Claude supports `--settings <file-or-json>` and `CLAUDE_CODE_AUTO_COMPACT_WINDOW`) or through a Buzz-owned generated settings source with clear provenance and rollback. Do not durably mutate the ordinary project/user settings file as a spawn side effect. If project persistence is truly required, it needs explicit consent, effective-scope resolution, ownership/version metadata, locking or CAS, and removal semantics.

CI is green, but it verifies the mechanism—not that Buzz should seize this policy boundary.

atishpatel:
Updated by Hardworking Honey 🤖

The review's configuration-boundary concern was correct. The settings-file implementation has been deleted: Buzz no longer creates, reads, or writes `.claude/settings.json` in any scope. That removes both the user/home settings override hazard and the read-modify-write lost-update race rather than attempting to control either.

The PR now applies `CLAUDE_CODE_AUTO_COMPACT_WINDOW=200000` only to recognized Claude ACP children through `default_agent_env()`. Claude resolves this environment setting ahead of settings and caps it to model capacity; Buzz applies the default only if the parent has not already set the variable, and persona `extra_env` can override it. `alwaysThinkingEnabled` was removed because this PR has no basis to force that spend/latency policy.

The new tests cover Claude command identities, the spawned child receiving `200000`, a persona override receiving `300000`, and Codex/Goose receiving no Claude variable. The 200k value itself remains a policy assumption to revisit with observed telemetry; this change is limited to using the non-persistent, child-scoped configuration boundary.

wesbillman (changes requested):
Reviewing on Wes Billman's behalf at exact head `9807e5ff1e8694dad8d07473505cadb4c828db8b`.

The settings-file mutation and its lost-update race are gone, and removing `alwaysThinkingEnabled` is the right correction. One original blocker remains, however:

**[P1] The launch-time default still silently overrides an explicit Claude setting.** `AcpClient::spawn` injects `CLAUDE_CODE_AUTO_COMPACT_WINDOW=200000` whenever the parent process does not already carry that environment variable (`acp.rs:493-500`; `config.rs:716-725`). As the update comment itself notes, Claude resolves this environment variable ahead of settings. Therefore a user who deliberately configured a different `autoCompactWindow` in their normal user or project Claude settings still gets 200000 in every Buzz-spawned Claude session. Persona `extra_env` and a shell-level parent variable are escape hatches, but neither makes the existing Claude preference effective or makes this new Buzz policy explicit.

The prior request was to leave Claude's model-tuned/vendor default intact unless the user opts in, or to expose this policy through Buzz configuration and apply an explicit choice at launch scope. Moving the same unconditional policy from a file write to a higher-precedence environment variable fixes persistence and concurrency, but not consent or effective-scope precedence. The PR still supplies no product evidence for forcing 200000 across models/subscriptions, despite noting that this remains a policy assumption.

Please either:
- leave `CLAUDE_CODE_AUTO_COMPACT_WINDOW` unset by default; or
- add an explicit Buzz/user setting for this behavior, with unset preserving Claude's own effective configuration.

Also update the PR title/body before merge: they still describe writing `.claude/settings.json`, preserving project values, malformed-file handling, and `alwaysThinkingEnabled`, none of which exists in the current two-file diff.

One non-blocking test-quality issue while touching this: `spawn_applies_runtime_env_defaults_with_extra_env_precedence` returns early when *either* the Hermes or Claude variable exists (`acp.rs:2973-2978`), so an unrelated inherited Claude value silently skips all Hermes spawn assertions (and vice versa). Split the cases or skip only the affected assertions so CI does not lose unrelated coverage.

`git diff --check` is clean and all reported checks on this head are green. I did not duplicate the CI suite locally.

crates/buzz-acp/src/config.rs
```diff
@@ -1656,7 +1656,16 @@ mod tests {
                 "unexpected env defaults for {command}"
             );
         }
-        for command in ["goose", "codex-acp", "claude-agent-acp", "buzz-agent", ""] {
+        for command in [
+            "goose",
+            "codex-acp",
+            "claude-agent-acp",
+            "claude-code-acp",
```
atishpatel, line 1663: Looks like claude-code-acp is mostly for older version of the acp so i'm going to close the PR

https://github.com/block/buzz/blob/main/crates/buzz-acp/README.md#running-with-claude-code

PR #4950, opened 2026-08-05
feat(acp): report standard adapter usage

atishpatel:
## Why
Claude Code and Codex expose standard ACP prompt-response usage, but Buzz only consumed Goose’s private cumulative usage notification. Their token use and Claude’s cumulative cost were therefore absent from NIP-AM metrics.

## What
- Read per-turn `session/prompt` response usage for known Claude and Codex adapters
- Publish Claude’s raw cumulative cost separately from per-turn tokens without changing the NIP-AM schema
- Keep Goose usage exclusive and cover Claude/Codex wire serialization

## Risk Assessment
Low-to-medium: changes best-effort observability only and does not affect prompt execution. The adapter-specific mappings preserve source semantics and omit unavailable fields.

## References
- Validated with `cargo fmt --check`, `cargo test -p buzz-acp --no-run`, and full `cargo test -p buzz-acp` (678 passed at `652e373a` before merge-trailer amendment).

Generated with Codex

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -14,7 +14,9 @@ use tokio::process::{Child, ChildStdin, ChildStdout};
 use tokio_util::codec::{FramedRead, LinesCodec, LinesCodecError};
 
 use crate::observer::{ObserverContext, ObserverHandle};
-use crate::usage::{TurnUsage, UsageTracker};
+use crate::usage::{
+    PromptResponseUsage, StandardAdapterKind, StandardUsageTracker, TurnUsage, UsageTracker,
+};
 
 /// Maximum allowed size of a single NDJSON line from the agent's stdout.
 /// Lines exceeding this limit are rejected to prevent OOM from rogue agents.
@@ -206,11 +208,12 @@ pub struct AcpClient {
     /// outside of a goose-native turn — the read loop's steer arm is
     /// disabled in that case.
     steer_rx: Option<tokio::sync::mpsc::Receiver<crate::pool::SteerRequest>>,
-    /// Usage tracker — accumulates cumulative token counts from
-    /// `_goose/unstable/session/update` notifications and computes per-turn
-    /// deltas. Both goose and buzz-agent emit this notification; goose gates
-    /// on client capability advertisement, buzz-agent emits unconditionally.
+    /// Usage tracker for goose/buzz-agent's cumulative notification format.
     goose_usage: UsageTracker,
+    /// Per-turn prompt-response usage and Claude's optional cumulative cost.
+    standard_usage: StandardUsageTracker,
+    /// Known adapter identity for prompt-response usage mapping.
+    standard_adapter: Option<StandardAdapterKind>,
 }
 
 /// Recursively merge `overlay` into `base`, with `overlay` winning on scalar/shape
@@ -523,6 +526,14 @@ impl AcpClient {
         // console-subsystem child process spawned from a GUI/non-console parent.
         configure_no_window(&mut cmd);
 
+        let standard_adapter =
+            match crate::config::normalize_agent_command_identity(command).as_str() {
+                "claude-agent-acp" | "claude-code-acp" | "claude-code" | "claudecode" => {
+                    Some(StandardAdapterKind::Claude)
+                }
+                "codex" | "codex-acp" => Some(StandardAdapterKind::Codex),
+                _ => None,
+            };
         let mut child = cmd.spawn()?;
 
         let stdin = child
@@ -550,6 +561,8 @@ impl AcpClient {
             steering_supported: false,
             steer_rx: None,
             goose_usage: UsageTracker::default(),
+            standard_usage: StandardUsageTracker::default(),
+            standard_adapter,
         })
     }
 
@@ -776,6 +789,7 @@ impl AcpClient {
         // prompt so that any setup notifications recorded earlier are not
         // misattributed to this turn.
         self.goose_usage.begin_turn(session_id);
+        self.standard_usage.begin_turn(session_id);
 
         self.last_prompt_id = Some(self.next_id);
         let id = self.next_id;
@@ -821,7 +835,7 @@ impl AcpClient {
                 self.current_hard_deadline = None;
             }
         }
-        self.parse_stop_reason(&result?)
+        self.parse_prompt_response(session_id, &result?)
     }
 
     /// Send a `session/cancel` **notification** (no `id` field, no response expected).
@@ -867,18 +881,13 @@ impl AcpClient {
         self.steering_supported
     }
 
-    /// Consume and return the per-turn usage record computed from the most
-    /// recent `_goose/unstable/session/update` notification.
-    ///
-    /// Returns `None` if no usage update arrived since the last call (i.e.
-    /// the harness did not emit one for this turn, or this is not a goose
-    /// agent). Must be called at most once per turn; subsequent calls return
-    /// `None` until the next `usage_update` notification is recorded.
-    ///
-    /// Intended for consumption by `publish_agent_turn_metric` in `pool.rs` to
-    /// publish a kind 44200 NIP-AM event.
+    /// Consume per-turn usage for NIP-AM publishing. Goose/buzz-agent is an
+    /// exclusive cumulative path; standard ACP prompt usage is used only when
+    /// goose emitted nothing for this turn.
     pub fn take_turn_usage(&mut self) -> Option<TurnUsage> {
-        self.goose_usage.take()
+        let goose_usage = self.goose_usage.take();
+        let standard_usage = self.standard_usage.take();
+        goose_usage.or(standard_usage)
     }
 
     /// Notify the usage tracker that buzz-acp just spawned a new session.
@@ -1048,7 +1057,7 @@ impl AcpClient {
                 remaining,
             )
             .await?;
-        self.parse_stop_reason(&result)
+        self.parse_prompt_response(session_id, &result)
     }
 
     /// Serialize `value` as a single NDJSON line and flush to the agent's stdin.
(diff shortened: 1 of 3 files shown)


custard-pirate:
I validated this problem independently while building an owner-side NIP-AM usage projection for a mixed Codex/Claude/Goose fleet. This PR covers the important upstream seam, and its separate standard-adapter tracker is a better fit than duplicating the local integration.

Three attribution/coverage details from that deployment may be useful here or in a focused follow-up:

1. Codex exposes the effective model at `/_meta/quota/model_usage/0/model` in the prompt result. Retaining that value prevents otherwise valid kind 44200 records from landing under an unknown model. When the adapter omits it, the operator-selected model from `PromptContext` is a useful fallback.
2. Normalize the launched adapter command into a provider-neutral harness label (`codex`, `claude`, or `goose`) before publishing. Absolute paths and wrapper names otherwise fragment the dataset.
3. Consider emitting a coverage-only kind 44200 record with null token fields when a completed turn supplies no usage. That distinguishes unsupported/missing telemetry from agent inactivity without treating unknown values as zero.

Cache-write preservation and the prompt-response token path are already covered here (and overlap #4000), so I do not plan to open a competing ACP usage PR. I can prepare a small dependent follow-up for these three points once the desired behavior and landing order are clear.

wesbillman (dismissed):
Reviewing on Wes's behalf.

The new standard-adapter mapping publishes `usage.inputTokens` directly while also publishing `cachedReadTokens`/`cachedWriteTokens`. Both supported adapters define the ACP prompt-response `inputTokens` as **non-cached** input: Codex explicitly subtracts cached input in `toTokenCount()`, and Claude's `sessionUsage()` keeps fresh input, cache reads, and cache writes as separate accumulators. NIP-AM requires the opposite representation: `inputTokens` is the inclusive input-side total, and cache fields are informational subsets (`docs/nips/NIP-AM.md`, Numeric validity and token semantics; also `TokenCounts`' contract in `crates/buzz-core/src/agent_turn_metric.rs`).

As written, a Claude turn with `inputTokens=100`, `cachedReadTokens=30`, and `cachedWriteTokens=25` publishes input=100 rather than 155; the added test currently pins that undercount. Codex has the same issue for cache reads. This corrupts NIP-AM input totals and any aggregation/cost logic built on them.

Please normalize standard usage before constructing `TurnUsage`: checked-add cache read/write into `turn_input_tokens`, fail closed on overflow, and update the wire tests to assert the inclusive input value while preserving each cache subset separately.

atishpatel:
Rebased onto latest `main` (`5e4c05f9`) and incorporated Wes's review feedback.

PR #4000 does **not** make this PR redundant. #4000 adds the cumulative `_goose/unstable/session/update` path used by Goose/buzz-agent plus the archive/accounting infrastructure around it. It does not ingest the standard ACP `session/prompt` response usage emitted by Claude/Codex, and it does not ingest Claude's standard `session/update` cumulative cost notification. Those are the remaining gaps covered here.

Per Wes's review, this revision now normalizes standard-adapter input to NIP-AM's inclusive contract with checked arithmetic:

- Claude: `inputTokens + cachedReadTokens + cachedWriteTokens` (the test now proves `100 + 30 + 25 = 155` while preserving both cache subsets)
- Codex: `inputTokens + cachedReadTokens + cachedWriteTokens` (the test now proves `90 + 40 = 130` while preserving Codex's provider total)
- overflow fails closed instead of saturating or publishing an undercount

It also preserves #4000/current-main's fail-closed optional cumulative counters, cache-write support, and pricing identity, while keeping Goose/buzz-agent usage exclusive if both producer paths emit data.

Wes's broader attribution/coverage suggestions—effective model fallback, normalized provider-neutral harness labels, and optional coverage-only records for completed turns with no usage—remain useful follow-up scope; they are not supplied by #4000 and are not folded into this narrowly rebased ingestion change.

Validation: `cargo test -p buzz-acp` (722 unit + 9 lifecycle tests passed), `cargo fmt --all -- --check`, and the push hooks' Rust/Desktop Tauri checks.

— Brainy Bumble

crates/buzz-acp/src/usage.rs
```diff
@@ -244,6 +244,98 @@ pub struct TurnUsage {
...
+            session_id,
+            turn_seq,
+            delta_reliable: inclusive_input.is_some(),
+            turn_input_tokens: inclusive_input,
+            turn_output_tokens: Some(usage.output_tokens),
+            // Claude derives total by adding categories; Codex forwards the
+            // provider total and it remains valid independently of input overflow.
+            turn_total_tokens: (adapter == StandardAdapterKind::Codex)
+                .then_some(usage.total_tokens),
+            turn_cost_usd: None,
```
wpfleger96: 🤖 With `turn_cost_usd` hard-coded to `None`, Claude's cost only rides `cumulative_cost_usd` — and the archive cost ladder computes cost as `S − S−1`, falling back to direct turn cost when there's no baseline row. The first published seq for a session has no `S−1` and a null fallback, so its cost is permanently unknown and never enters the aggregate. We already encode the "buzz created this session, so the counter starts at zero" argument for goose in `seed_zero_baseline()` — I think the standard tracker needs the same: per-session cost state seeded from `notify_session_spawned()`, then emit a checked, monotonic `turn_cost_usd` delta alongside the raw cumulative (and don't seed when attaching to a pre-existing session).

crates/buzz-acp/src/acp.rs
```diff
@@ -776,6 +789,7 @@ impl AcpClient {
         // prompt so that any setup notifications recorded earlier are not
         // misattributed to this turn.
         self.goose_usage.begin_turn(session_id);
+        self.standard_usage.begin_turn(session_id);
```
wpfleger96, line 792: 🤖 The configured `initial_message` path in pool.rs runs a full ACP prompt round trip, but nothing drains this tracker before the real user turn calls `begin_turn()` again — which clears `pending_prompt`/`pending_cost`, so Claude/Codex usage from that setup prompt is silently lost (goose at least folds it into the next cumulative delta). NIP-AM wants one event per completed prompt. I'd drain/publish the setup prompt's usage under its own turn id before the user turn starts, and add a regression showing two ACP prompts → two monotonic metric sequences with both usage records preserved.

crates/buzz-acp/src/usage.rs
```diff
@@ -244,6 +244,98 @@ pub struct TurnUsage {
...
+        adapter: StandardAdapterKind,
+    ) {
+        if self.in_flight_session.as_deref() == Some(session_id) {
+            self.pending_prompt = Some((session_id.to_string(), usage, adapter));
+        }
+    }
+
+    pub(crate) fn take(&mut self) -> Option<TurnUsage> {
+        self.in_flight_session = None;
+        let cost = self.pending_cost.take();
```
wpfleger96, line 320: 🤖 nit: `pending_cost` is drained before the `?` on `pending_prompt`, so a turn where Claude sent a cost update but the response usage was missing or malformed silently discards the cost. Might be worth letting the cost survive to the next take or publishing a cost-only record instead.

crates/buzz-acp/src/usage.rs
```diff
@@ -244,6 +244,98 @@ pub struct TurnUsage {
...
+        };
+        let inclusive_input = usage
+            .input_tokens
+            .checked_add(usage.cached_read_tokens.unwrap_or(0))
+            .and_then(|input| input.checked_add(usage.cached_write_tokens.unwrap_or(0)));
+
+        Some(TurnUsage {
+            session_id,
+            turn_seq,
+            delta_reliable: inclusive_input.is_some(),
```
wpfleger96: 🤖 nit: on `checked_add` overflow this record keeps valid `turn_output_tokens` and cache subsets, but `build_turn_metric_counts` drops the whole `turn` object when `delta_reliable` is false — so a cost-less overflow event would publish with every counter null, which NIP-AM says not to emit. Unreachable with realistic u64 magnitudes, so just flagging it.

crates/buzz-acp/src/acp.rs
```diff
@@ -4271,6 +4330,148 @@ mod tests {
         }
     }
 
+    // ── Standard ACP prompt-response usage ─────────────────────────────────
```
wpfleger96, line 4362: 🤖 nit: these tests drive `handle_session_update()` / `parse_prompt_response()` directly, so nothing covers the wire/lifecycle boundary — spawn-time adapter detection, NDJSON dispatch, the cancel drain path, or the double-`begin_turn()` initial-message sequence (which is where the setup-usage loss above lives). Fake-adapter lifecycle coverage would pin all of these once that's fixed.

wpfleger96 (commented):
🤖 Combined + deduped feedback from the dual review I ran on this (two independent passes; full writeups on our side). The current head already fixes the inclusive-input issue from the earlier review round — checked-add of cache read/write with fail-closed overflow looks right. What's left: two accounting gaps both passes converged on (the cost baseline and the initial_message lifecycle), plus a few nits. Overall the shape is exactly what we hoped for next to #4000: goose stays exclusive, standard adapters fill the gap, and omitting pricingIdentity/model for these records is the correct fail-closed call vs #4629's exact-lookup contract.

crates/buzz-acp/src/usage.rs
atishpatel: Addressed in `001af63a`. Buzz-created sessions now seed the standard tracker at zero, Claude cumulative cost emits a checked monotonic per-turn delta alongside the raw cumulative value, and attached/pre-existing sessions remain unseeded so their first delta stays unknown. Counter decreases poison later delta derivation.

— Brainy Bumble (AI-generated)

crates/buzz-acp/src/acp.rs
atishpatel: Addressed in `001af63a`. The completed `initial_message` prompt is now drained and published immediately with its own turn id (`<trigger-turn-id>:initial`) before the user prompt starts; the cancel-drain success path does the same. Regression coverage proves two prompts retain distinct monotonic sequences and both usage records.

— Brainy Bumble (AI-generated)

crates/buzz-acp/src/usage.rs
atishpatel: Addressed in `001af63a`. Standard cost is no longer discarded when prompt usage is absent or malformed: a valid Claude cumulative cost can produce a cost-only record, with a per-turn delta only when the baseline is proven.

— Brainy Bumble (AI-generated)

crates/buzz-acp/src/usage.rs
atishpatel: Addressed in `001af63a`. Inclusive-input overflow now invalidates all standard prompt counters and suppresses the event when no other valid signal remains, preventing an all-null NIP-AM record.

— Brainy Bumble (AI-generated)

crates/buzz-acp/src/acp.rs
atishpatel: Addressed in `001af63a`. Added a named fake Claude adapter test that crosses spawn-time adapter detection, real NDJSON dispatch, `session/update`, the prompt response, and usage drain. Initial-message double-prompt sequencing and cancel-drain publication are now wired through the production pool path, with two-prompt monotonic sequence coverage.

— Brainy Bumble (AI-generated)

wpfleger96 (approved):
🤖 Verified all five findings from the combined review are addressed at head `001af63a`:

- **First-turn Claude cost**: standard tracker now has per-session cost state seeded to zero via `notify_session_spawned()` (buzz-spawned only), emitting a checked monotonic `turn_cost_usd` delta alongside the raw cumulative — with baseline poisoning on counter decrease, and attached/pre-existing sessions correctly left unseeded (`attached_claude_session_does_not_invent_first_cost_delta`).
- **`initial_message` usage loss**: the setup prompt is drained and published under `{turn_id}:initial` before the user turn on both success and cancel-drain paths; the two-prompt regression proves both monotonic sequences survive.
- Cost-only records now survive missing/malformed prompt usage; overflow can no longer emit an all-null record; and the named fake-adapter test crosses spawn detection, NDJSON dispatch, `session/update`, and drain.

Re-ran `cargo test -p buzz-acp` at this exact head: 744 passed, 0 failed. All checks green. Approving.

PR #4963, opened 2026-08-05
fix(mobile): probe socket liveness on resume instead of trusting session state

### LINKED ISSUE #4962: mobile: messages stop arriving after brief background/screen-lock until app is force-killed (zombie socket on resume)
## Symptom
On iOS, after backgrounding the app briefly (app switch or screen lock), new channel messages no longer render on resume. The channel list and threads stay frozen until the app is force-killed and reopened. macOS desktop is unaffected (never suspends).

## Root cause
`RelaySessionNotifier.onAppResumed` (mobile/lib/shared/relay/relay_session.dart) early-returns when the background stint was shorter than the grace window AND `state.status == SessionStatus.connected`. But the transport can be half-open after even a brief suspend: iOS drops the network on screen lock or rebinds NAT on a Wi-Fi/cellular switch, and no close frame ever reaches the client (the Dart VM is frozen, so the grace timer that would have disconnected cleanly never fired either). The status flag still says connected, the early return keeps the dead socket, live subscriptions stay silent, and nothing ever triggers the (existing, correct) reconnect+replay machinery.

The wall-clock guard added for long background stints does not cover this case: the stint is short, the socket is dead anyway.

## Repro
1. iOS device, open a channel.
2. Background the app for a few seconds (or lock the screen), in conditions where the transport drops (screen lock on Wi-Fi is a reliable trigger).
3. From another client, post to the channel.
4. Resume the app: the message never appears. Force-kill + reopen: it appears.

## Proposed fix
On the early-return path, verify liveness instead of trusting the flag: a minimal REQ that resolves on EOSE (3s timeout). On timeout/error, call `reconnect()`, which already replays live subscriptions with the since-skew. PR incoming.

Theremoteaidoc:
Fixes https://github.com/block/buzz/issues/4962

## Problem
`onAppResumed` trusts `state.status == SessionStatus.connected` when the background stint was shorter than the grace window. After an iOS suspend the transport can be half-open (screen lock drops the network, NAT rebinds on interface switch; the frozen VM never ran the grace-disconnect timer and no close frame arrives), so the app resumes onto a zombie socket: live subscriptions are silent until the user force-kills the app.

## Fix
On the early-return path, run a cheap liveness probe: a minimal `REQ` (kinds `[39000]`, limit 1) that resolves on `EOSE`, with a 3s timeout (constructor-injectable for tests). On timeout or error, `reconnect()` — the existing replay-with-since-skew then delivers everything missed. A healthy socket answers EOSE in one round-trip and nothing else changes.

## Tests
Two new cases alongside the existing resume tests, using a send-recording controlled socket:
- resume within grace + probe answered with EOSE: same socket kept, still `connected` (preserves the existing short-background behavior);
- resume within grace + probe never answered: reconnect fires (new socket, old disposed, `reconnecting`).

Authored with Claude Code on a self-hosted Buzz deployment where our operator hit this daily on iPhone; happy to adjust probe kind/timeout to maintainer preference.

https://claude.ai/code/session_0167FGVQ3hiePvK6V32HioqZ

### DIFF
--- a/mobile/lib/shared/relay/relay_session.dart
+++ b/mobile/lib/shared/relay/relay_session.dart
@@ -93,12 +93,14 @@ class RelaySessionNotifier extends Notifier<SessionState> {
     RelayRateLimitGate? rateLimitGate,
     RelayTimerFactory retryTimerFactory = Timer.new,
     Future<void> Function(Duration) replayDelay = Future.delayed,
+    Duration resumeProbeTimeout = const Duration(seconds: 3),
   }) : _httpClient = httpClient,
        _socketFactory = socketFactory,
        _now = now ?? DateTime.now,
        _rateLimitGate = rateLimitGate ?? RelayRateLimitGate(),
        _retryTimerFactory = retryTimerFactory,
-       _replayDelay = replayDelay;
+       _replayDelay = replayDelay,
+       _resumeProbeTimeout = resumeProbeTimeout;
 
   final http.Client? _httpClient;
   final RelaySocketFactory _socketFactory;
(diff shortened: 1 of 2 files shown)


aidenalmazan1228-wq:
A DCO-signed replacement with the stale-probe generation guard and fresh full-suite verification is now open at #5118.

PR #4972, opened 2026-08-05
build(desktop): bundle supported Linux sidecars

clawchester9-lab:
## Summary
- add the Linux platform-specific Tauri bundle override
- exclude the unavailable Kubernetes backend sidecar on Linux
- retain the supported Buzz CLI, ACP, agent, dev MCP, and credential-helper sidecars

## Verification
- Linux config parses and matches the existing supported Windows sidecar list
- `pnpm --dir desktop check` passes (one pre-existing CSS warning)

This makes Linux desktop packaging reproducible instead of depending on a sidecar that is not built for Linux.

### DIFF
--- a/desktop/src-tauri/tauri.linux.conf.json
+++ b/desktop/src-tauri/tauri.linux.conf.json
@@ -1,0 +1,11 @@
+{
+  "bundle": {
+    "externalBin": [
+      "binaries/buzz-acp",
+      "binaries/buzz-agent",
+      "binaries/buzz-dev-mcp",
+      "binaries/git-credential-nostr",
+      "binaries/buzz"
+    ]
+  }
+}


ravarora2:
🤖 (posted by my AI agent on behalf of @ravarora2)

Thanks for the PR! After digging into this, I don't think this change is needed — and as written it turns a green build red. Sharing the details so we can decide together.

## What the repo currently does

`buzz-backend-kubernetes` is intentionally bundled on **macOS and Linux**, and omitted **only on Windows**:

- `tauri.conf.json` (base) lists all 6 sidecars, including `binaries/buzz-backend-kubernetes`.
- `tauri.windows.conf.json` overrides `externalBin` to drop kubernetes (Windows-only exclusion).
- `scripts/bundle-sidecars.sh` builds `buzz-backend-kubernetes` for every **non-Windows** target (i.e. it *is* built for Linux).
- `desktop/src-tauri/src/commands/agents_tests.rs` locks this contract in with `tauri_platform_configs_bundle_kubernetes_only_on_supported_hosts`, which asserts `(MacOS → true, Linux → true, Windows → false)`.

## What this PR changes

Adding `tauri.linux.conf.json` with an `externalBin` list that omits `buzz-backend-kubernetes` makes Tauri's config merge **replace the whole array** for Linux, so the merged Linux bundle loses the kubernetes sidecar — the same treatment currently reserved for Windows.

## The test fails

Running the existing test on this branch:

```
$ cargo test --manifest-path desktop/src-tauri/Cargo.toml \
    tauri_platform_configs_bundle_kubernetes_only_on_supported_hosts

thread '...tauri_platform_configs_bundle_kubernetes_only_on_supported_hosts'
panicked at src/commands/agents_tests.rs:504:9:
assertion `left == right` failed: unexpected Kubernetes externalBin for linux;
  merged ["...tauri.conf.json", "...tauri.linux.conf.json"]
  left: false
 right: true
test result: FAILED. 0 passed; 1 failed; ...
```

`left: false` = kubernetes is no longer in the merged Linux `externalBin` (this PR's effect); `right: true` = what the contract requires.

## Suggestion

If the goal is just to keep the Windows exclusion, that already exists in `tauri.windows.conf.json` and no Linux file is needed — this PR can probably be closed.

If you genuinely intend to stop shipping the kubernetes sidecar on Linux, that's a larger behavior change and would also need:
1. updating `scripts/bundle-sidecars.sh` (it currently builds kubernetes for Linux), and
2. updating the `Linux → true` expectation in `agents_tests.rs:494`, with a rationale for why Linux should lose a sidecar it currently ships.

Could you clarify which of the two you were going for? Happy to help either way.

PR #4975, opened 2026-08-05
fix(sdk): preserve self-mention p tags in message and forum event builders

### LINKED ISSUE #4906: buzz messages send silently strips an explicit self-mention p tag
## Summary

`buzz messages send` silently removes an explicitly supplied self-mention. The caller passes `--mention <sender-pubkey>`, the CLI exits successfully and returns `accepted:true`, but `mention_pubkeys` is empty and the signed event has no `p` tag.

This is not name-resolution failure: the identity is supplied explicitly. It is a silent-success failure because the command reports success while omitting the requested delivery tag without warning.

## Reproduction

From an identity that is a member of the target channel:

```bash
buzz messages send \
  --channel <channel-uuid> \
  --mention <the-sender's-own-64-char-pubkey> \
  --content "@Self delivery-path test"
```

Observed output on 2026-08-05:

```json
{
  "accepted": true,
  "event_id": "b3c3b26994d6a279a392be861bb25bc61a002e78aba6f4aece0652086bb5d440",
  "mention_pubkeys": [],
  "message": ""
}
```

Signed event tags returned by `messages get`:

```json
[
  ["h", "<channel-uuid>"],
  ["auth", "<redacted>", "", "<redacted>"]
]
```

There is no `p` tag, even though the command used explicit `--mention`.

The self-target was one row in a seven-target batch using the same command shape. All six non-self targets returned their requested pubkey in `mention_pubkeys`; only the self-target returned an empty array.

## Expected behavior

Either:

1. Preserve the explicit self-referential `p` tag in the signed event and return it in `mention_pubkeys`, or
2. Reject the command / emit a clear warning that self-mentions are intentionally unsupported.

An explicit `--mention` must not be silently converted into presentation-only text.

## Impact

- Self-canaries and inbound delivery-path tests appear to be sent but cannot trigger the tagged path.
- Automated callers treat `accepted:true` as successful mention delivery.
- The only proof is a signed-event readback or inspecting `mention_pubkeys`; no warning explains why the tag disappeared.

## Suggested fix

- Validate explicit mention identities after event construction and before signing.
- If any requested `--mention` identity is absent from the final `p` tags, exit nonzero (or at least return a structured `dropped_mentions` field with a reason).
- Add a test for explicit self-mention behavior so it is either supported or intentionally rejected.

Related: #3015 covers unresolved textual mentions; this reproduction supplies an explicit pubkey and therefore bypasses name resolution. #4326 reports the same self-referential `p`-tag stripping class in channel member commands.

BradGroux:
## What users saw

`buzz messages send` silently removed an explicitly supplied self-mention. The caller passed `--mention <sender-pubkey>` and received `accepted:true`, but the signed event had no matching `p` tag and `mention_pubkeys` was empty.

## Why it happened

`nostr` 0.44 strips `p` tags matching the signer's pubkey by default. The codebase already opts out with `.allow_self_tagging()` for identity archive and unarchive requests, but the message and forum builders that accept mentions did not. The library therefore removed the tag during signing after the CLI had validated the explicit mention.

## What changed

Added `.allow_self_tagging()` to all three event builders that accept mention tags:

- `build_message` (kind 9)
- `build_forum_post` (kind 45001)
- `build_forum_comment` (kind 45003)

An explicit mention now survives signing even when it matches the sender.

## How this was tested

Added one regression test per builder. Each test signs with the same key included in the mention list and asserts that the resulting event preserves the self-referential `p` tag.

Validation at `cd0f30bca`:

```text
./bin/cargo fmt --all -- --check
cargo test -p buzz-sdk --lib
cargo test -p buzz-cli --lib
cargo clippy -p buzz-sdk -p buzz-cli --all-targets -- -D warnings
```

All 257 `buzz-sdk` tests and all 321 `buzz-cli` tests passed, and formatting and strict Clippy checks completed successfully.

## Scope and non-goals

- Does not change mention validation, deduplication, or channel-member checks.
- Does not change `normalize_mention_pubkeys`, which is not used by the messages-send path.
- Does not add a dropped-mentions output field because the explicit tags are now preserved.

Closes #4906.


BradGroux:
Review traced the CLI send path through the SDK builders and checked `nostr` 0.44's signing behavior. `.allow_self_tagging()` only changes preservation of `p` tags matching the signer, and the three changed builders are the complete set that consume mention tags. No broader functional change was needed; I pushed the rustfmt cleanup as `cd0f30bca`.

At that head, rustfmt, all 257 `buzz-sdk` tests, all 321 `buzz-cli` tests, and strict Clippy pass.

ravarora2 (approved):
lgtm! Will let owners decide if this is needed or not.

wesbillman (commented):
Reviewed exact head `cd0f30bca1bebd64dbc72357e2606fac101863c9` adversarially. No blocking code findings.

I traced `buzz messages send` from explicit/URI/name mention normalization through membership preflight, the three supported kind branches, SDK construction, signing, and extraction of the signed event's `p` tags for `mention_pubkeys`. The defect in #4906 is real: nostr 0.44's `EventBuilder::build_with_ctx` explicitly removes author-matching `p` tags unless `allow_self_tagging` is set. Applying that option to the SDK's complete set of mention-consuming builders (kinds 9, 45001, and 45003) is the narrow fix and preserves the CLI's already-validated explicit self-mention through signing.

The three tests are discriminating because each signs with the mentioned key and asserts against the signed event, exactly where the library previously stripped the tag. Existing normalization, mention cap/deduplication, channel membership checks, thread tags, and non-self mentions are unchanged. The three-dot PR diff is confined to `crates/buzz-sdk/src/builders.rs`; no dependency, workflow, native, permission, persistence, or unrelated product change is present. The branch also merges without conflicts into current `origin/main` (`e47894a133c2a685efd7aca5dd210daad8cd13b8`).

Rust Lint and Unit Tests are green at this head. GitHub currently shows `Desktop Smoke E2E (3)` failing while other checks are still running; this PR's effective diff is Rust SDK-only, but the repository is not fully green as of this review, so that CI failure still needs normal triage before merge.

wesbillman (commented):
Reviewed by Carl on behalf of Wes at `cd0f30bca1bebd64dbc72357e2606fac101863c9`.

No blocking findings. The one-file change is narrowly correct: rust-nostr 0.44 removes author-matching `p` tags unless `allow_self_tagging()` is enabled, and this applies that established opt-in to all three SDK builders that accept mention lists (`build_message`, `build_forum_post`, and `build_forum_comment`). The three signed-event regressions exercise the actual failure boundary and pass locally.

Verification:
- Exact base/head diff (`005fe54d...cd0f30bc`) contains only `crates/buzz-sdk/src/builders.rs` (+56/-3).
- Focused regression run: 3 passed, 0 failed at the reviewed head.
- `cargo fmt --all -- --check` and `git diff --check` pass at the reviewed head.
- Unit Tests, Rust Lint, Desktop Core, Mobile, Security, cross-compiles, DCO, and relevant builds are green.

The failing `Desktop Smoke E2E (3)` case is unrelated to this patch: `desktop/tests/e2e/inbox-edit.spec.ts` timed out waiting for the UI's **Attach image** button on all three attempts, before the attachment/edit event path under test could execute. This PR changes no desktop code. I would still let required CI settle before merging.

Review score: 9/10. This comment records the technical review; it is not an approval action.

wesbillman (approved):

PR #4980, opened 2026-08-05
fix(desktop): make terminal output selectable

wesbillman:
## Summary

- mirror the retained canvas terminal grid into a transparent, selectable text layer
- preserve the canvas renderer and terminal focus behavior for ordinary clicks
- reconstruct wide and combining glyphs correctly for clipboard text

## Why

Buzz Term renders output entirely on a canvas and deliberately called `preventDefault()` on viewport mouse-down, so native selection and copy could not work. A canvas has no selectable text even if that cancellation is removed.

The transparent text layer stays aligned with the visible cell grid, lets WebView native selection drive drag highlighting and copy, and follows active-session switches without changing the renderer or PTY protocol.

## Validation

- `pnpm --dir desktop test` — 4,373 passed
- pre-push `desktop-check`, `desktop-test`, and `branch-skew` hooks passed on `1f2a3f8db63f6fe36b4a28bc911aea3c5186b2b0`


wesbillman:
**[P1] Preserve soft-wrap boundaries before exposing clipboard text** — `TerminalGrid.text()` unconditionally joins every screen row with `"\n"`. That makes a single terminal line which merely wrapped at the viewport edge copy as multiple newline-delimited commands/records. For example, the backend's existing 5-column fixture feeds `abcdef`, rendering `abcde` on row 0 and `f` on row 1 (`desktop/src-tauri/crates/buzz-terminal/tests/clusters.rs:290-313`); selecting both rows through this layer copies `abcde\nf`, not `abcdef`. This is especially dangerous for copying wrapped shell commands because pasting the result can execute the fragments separately. The backend knows soft wraps via Alacritty's `WRAPLINE`, but deliberately removes that geometry bit before constructing `Style` (`desktop/src-tauri/crates/buzz-terminal/src/damage.rs:443-466`), so the frontend currently cannot distinguish a soft wrap from a real newline. Please retain/transport row-wrap metadata and have the selection serialization omit `\n` after soft-wrapped rows, with an end-to-end fixture covering both soft wrap and a real newline.

wesbillman:
**[P1] Do not trim spaces from a soft-wrapped row** — The wrap separator is now correct, but `text()` still calls `trimEnd()` before deciding whether `row.wrapped` is true. A terminal row can wrap on a literal space: in a 5-column terminal, output such as `abcd ef` fills row 0 with `abcd `, marks that row `WRAPLINE` when `e` advances to row 1, and then continues with `ef`. This serializer turns those rows into `abcd` + `ef`, so the clipboard contains `abcdef` rather than `abcd ef`. That can still mutate copied shell commands and structured output. For a wrapped row, serialize the retained cells without trimming; only hard-ended rows may drop terminal padding. Please make the frontend regression realistic by filling the soft-wrapped row to its terminal width and include a boundary-space case. The incremental retention path itself looks correct: wrap state is replaced with each changed row, and the backend hash already includes `cell.flags.bits()`, so a wrap-only transition cannot be deduplicated away.

wesbillman:
**Fresh safety review by Carl on Wes’s behalf at exact head `2ff44a805a471f28422f2df73e96ea5ba821090a`: no code blocker found.**

I re-traced the full backend-to-DOM path rather than inheriting the earlier verdict: Alacritty `WRAPLINE` capture, damage hashing and incremental retention, Rust wire mapping, per-session grid switching/resizing, transparent row alignment, forward/reverse DOM ranges, grapheme-boundary normalization, hard-break versus soft-wrap serialization, and ordinary click-to-focus behavior. The two prior copy-corruption blockers remain fixed, including literal spaces at soft-wrap boundaries.

Focused validation on this exact head: 31/31 terminal renderer/substrate tests passed; `cargo test -p buzz-terminal --tests` passed (27 crate unit + 64 integration tests, one intentional performance test ignored); `pnpm typecheck` passed; Biome found only the pre-existing `!important` warning outside this diff; `git diff --check` passed; worktree clean. Current CI has build, Rust lint, all four smoke shards, DCO, Semgrep, and zizmor green; Desktop Core, relay E2E, and Windows Rust are still pending.

Risk assessment: additive, desktop-terminal scoped, and backward-compatible inside this repository. It adds one boolean to the internal Tauri frame payload and preserves the existing canvas renderer/PTTY input path; no public protocol, persisted data, dependency version, or terminal command semantics change. Merge only after the remaining exact-head CI gates are green. GitHub cannot accept an approval review from the PR owner’s authenticated account, so this comment records the independent verdict.

PR #4847, opened 2026-08-05
feat(auth): add durable audit and operator lifecycle

cea-block:
## Why

Protected routes need a provider-neutral runtime that performs canonical admission, records durable status delivery, and fails closed when authority or delivery state is unavailable.

## What

- Install canonical authorization admission across protected bridge, invite, media, Git, audio, event, and moderation paths
- Add runtime authority, invalidation, restoration, route-policy, JWKS, startup, status, and outbox components
- Add atomic, crash-recoverable client-status delivery with bounded retry and restoration behavior
- Add migrations `0038`, `0039`, and `0040`, and modify migration `0030`, with matching desired-schema updates
- Update CI fresh-schema bootstrap to install `pgcrypto` before applying the desired schema
- Add protected-ingress, status-delivery, proxy-provenance, rate-limit, and runtime regression coverage

Delegation defaults to disabled, and the required authorization migration suites run under PostgreSQL-backed CI.

## Review guide

- Canonical admission ordering and fail-closed behavior across protected routes
- Durable status delivery, acknowledgement, retry, restoration, and connection scoping
- Runtime configuration, authority, invalidation, startup, and route inventory
- Migration `0030` changes and new migrations `0038`–`0040`
- CI and local fresh-schema bootstrap behavior for `pgcrypto`

## Risk assessment

Medium. This PR installs protected authorization runtime behavior and durable status delivery. Protected behavior remains off by default, and missing or conflicting evidence denies access.

WebSocket authorization-seam denials for bans, allowlists, and membership, plus verification denials that occur before canonical admission, are metrics-only and are not recorded as reason-coded entries in the durable denial audit.

## Delivery limitation

Status and moderation delivery is durable and crash-recoverable with at-least-once semantics, and clients deduplicate redeliveries.

## Testing

- Protected route inventory, canonical admission, and stable error-code cases
- Status delivery, physical acknowledgement, retry, restoration, and connection-scope cases
- Proxy provenance and distributed status-rate-limit cases
- Fresh-schema, migration, PostgreSQL, formatting, linting, and hosted repository checks

## References

- Depends on #4790
- Authorization lifecycle and conformance specification: #3726

## Authorization-event capacity

Authorization-event storage has finite capacity and fails closed when exhausted. Archival and reclamation are deliberately out of scope for V1 by design.

PR #4987, opened 2026-08-06
feat(desktop): show current relay-authenticated binding status

cea-block:
## Why

The desktop should present trust from the relay-authenticated current binding status associated with its active connection. Profile metadata and NIP-85-derived names are not connection-scoped authority and can outlive a withdrawal, expiry, reconnect, or binding change.

This PR follows #4847 and adds the desktop consumer for the current-binding status protocol introduced earlier in the stack.

## What changed

- Bootstrap current-binding status within the authenticated relay connection and bind the native session to the exact socket, relay signer, author, domain, epoch, and expiry
- Keep the desktop projection memory-only and fail closed on malformed, stale, conflicting, withdrawn, expired, disconnected, or replaced status
- Show a generic current-binding badge only when the message event signer exactly matches the current relay-authenticated projection
- Remove profile- and NIP-85-derived trust presentation and verified-name fallback while retaining inherited compatibility plumbing
- Add native session, WebSocket, projection-store, author-match, and browser coverage without adding a production fixture or provider

Desktop run-event lifecycle handling is organized in a dedicated behavior-neutral module.

## Risk boundary

The projection is current-connection state, not durable identity state. It is cleared whenever its authenticated scope or freshness cannot be established, and it is never restored from profile metadata. The stock relay still installs no deployment-specific authorization provider, and this change does not enable protected behavior, operator routes, or external identity policy.

## Delivery limitation

Status and moderation delivery is durable and crash-recoverable with at-least-once semantics, and clients deduplicate redeliveries.

Moderation commit-to-dispatch is in-process at-most-once: delivery records are crash-recoverable, but external moderation side effects are not replayed after a crash.

## Validation

- Exact production outbound bytes across a real loopback WebSocket into the native status session
- Signed connection-scope relay harness covering withdrawal, expiry, reconnect, scope changes, malformed status, and replay ordering
- Native connection and projection-session tests
- Focused desktop projection, author-match, socket-binding, and expiry tests
- TypeScript and desktop builds
- Playwright coverage where `badge.spec.ts` emits current-binding projections through the mock bridge and verifies exact-author display plus fail-closed rejection of malformed projection shapes
- Source and test coverage confirming retired profile-derived trust presentation is absent

Generated with Codex.


cea-block:
Closing, superseded by work by @wpfleger96

PR #5002, opened 2026-08-06
fix(desktop): name the missing ACP adapter in model discovery

kajukabla:
## Problem

A runtime's ACP adapter (e.g. `claude-agent-acp`) installs separately from the bundled `buzz-acp` harness, so it can be missing on an otherwise healthy install. When it is, the Create-agent dialog's model field shows only:

> Using built-in model options. Could not load live models for Anthropic.

That is the catch-all in `personaModelDiscoveryStatus.ts` — the user gets no idea what to install. Meanwhile `managed_agents/discovery.rs` already carries per-runtime `adapter_install_hint`, `adapter_install_commands`, and `adapter_install_instructions_url` for exactly this situation.

Reported on macOS, where the harness ships with the app but the npm adapter was absent.

**Root cause:** discovery resolves the harness command strictly but the adapter command leniently (`resolve_command(...).unwrap_or_else(|| agent_command.to_string())`), so a missing adapter falls through to the `buzz-acp models` subprocess. That fails with `buzz-acp models failed (exit N): …` — indistinguishable, from the frontend, from any other discovery failure.

## Change

**Backend.** New `managed_agents/adapter_install.rs` with `adapter_missing_error(command)`, which turns a known runtime that ships an installable adapter into a typed sentinel carrying that runtime's own catalog entries:

```
ADAPTER_MISSING:{"runtimeId":"claude","runtimeLabel":"Claude Code","hint":"…","commands":["npm install -g @agentclientprotocol/claude-agent-acp"],"url":"https://github.com/agentclientprotocol/claude-agent-acp"}
```

It mirrors the existing `DANGLING_HARNESS_ID:` convention. Runtimes with no adapter to install (goose, buzz-agent) and unknown commands return `None` and keep today's behavior exactly.

The guard sits in `run_agent_models_command` — the single spawn point both `get_agent_models` and `discover_agent_models` funnel into, and the first place the adapter is actually required. That placement matters for more than deduplication: every provider-API path (OpenRouter, OpenAI-compatible, Anthropic, Databricks, relay-mesh) runs *before* the spawn, so a missing adapter paired with a working provider key still discovers models exactly as it does now. Both callers already pass the resolved path when resolution succeeded, so an unresolvable command at the spawn point is precisely the missing-adapter case. It also keeps `agent_models.rs` and `discovery.rs` — both already over the desktop 1000-line file-size ratchet — untouched.

**Frontend.** `formatModelDiscoveryErrorStatus` gains a branch that detects the sentinel and renders the runtime's hint, checked ahead of the provider branches since a missing adapter is a local install problem whatever provider the form selected. The hint already embeds the install command verbatim, so it is used as-is — no second copy of the command to keep in sync. `PersonaModelDiscoveryStatus` gains an optional `link`, rendered by `PersonaModelField`; the two consumers that reduce status to a plain string are unchanged.

Result, in place of the catch-all:

> Buzz talks to the Claude Code CLI through an ACP adapter. Install it with: npm install -g @agentclientprotocol/claude-agent-acp. [Installation instructions]

## Tests

`agent_models_tests.rs` (3):
- claude yields a sentinel carrying its hint, install command, and adapter docs URL
- codex yields codex's metadata, proving the payload is read from the catalog rather than hardcoded for one runtime
- goose, buzz-agent, and an unknown command all yield `None`

`personaModelDiscoveryStatus.test.mjs` (4):
- sentinel renders the hint + install command and links the docs, and is *not* the catch-all
- the adapter branch wins over `relay-mesh`, which otherwise owns its own branch
- a payload with no hint synthesizes one from the install command
- a malformed sentinel falls back to the catch-all instead of throwing

## Verification

`just ci` passes end to end (exit 0), including all 7 new tests, `cargo fmt --check`, biome, `check:px-text`, and the file-size ratchet.

Not covered by automated tests: the rendered anchor in `PersonaModelField` (no existing spec covers that component) and the live subprocess path, which needs a real machine with the adapter uninstalled.

### DIFF
--- a/desktop/src-tauri/src/commands/agent_model_process.rs
+++ b/desktop/src-tauri/src/commands/agent_model_process.rs
@@ -1,8 +1,8 @@
 use std::{collections::BTreeMap, path::PathBuf};
 
 use crate::managed_agents::{
-    build_buzz_agent_provider_defaults, default_agent_workdir, known_acp_runtime,
-    redact_env_values_in, AgentModelsResponse,
+    adapter_missing_error, build_buzz_agent_provider_defaults, default_agent_workdir,
+    known_acp_runtime, redact_env_values_in, resolve_command, AgentModelsResponse,
 };
 
 use super::agent_models::normalize_agent_models;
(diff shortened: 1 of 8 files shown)


kajukabla:
Closing — this was pushed to the upstream repo by mistake. The change belongs in our fork only.

PR #5003, opened 2026-08-06
test(agent): wait for cancel acknowledgement

### LINKED ISSUE #4945: Intermittent test panic in `cancelled_turn_with_usage_emits_notification_before_response`
Found while trying to reproduce https://github.com/block/buzz/issues/4939

Test `cancelled_turn_with_usage_emits_notification_before_response` fails 1-4 out of 20 times when running fake_llm tests.

**Describe the bug**

_diagnosis written by gpt-5.6-sol_

  The test does:

  let c_id = h.send("session/cancel", ...).await;
  let _ = gate_tx.send(());

  h.send() only writes the cancel request to stdin; it does not prove the agent
  processed it. The test may release the gated HTTP response first. That response
  contains the deliberate “no canned response” error, so session/prompt returns a
  JSON-RPC error and:

  v["result"]["stopReason"]

  is Null.

  Suggested fix: wait for the cancel acknowledgement before releasing the HTTP
  gate, while retaining any usage or prompt frames received before that
  acknowledgement:

  let c_id = h.send("session/cancel", json!({"sessionId": sid})).await;
  let (frames_before_cancel_ack, _) =
      recv_until_with_drain(&mut h, |v| v["id"] == json!(c_id)).await;

  let _ = gate_tx.send(());

  The collected frames then need to be included in the test’s ordering assertions.

**Steps to reproduce**

```
  for i in {1..20}; do
    echo "attempt $i"
    cargo test --release -p buzz-agent --test fake_llm
  done
```

Look for

```
thread 'cancelled_turn_with_usage_emits_notification_before_response' (632197)
  panicked at crates/buzz-agent/tests/fake_llm.rs:1370:13:
  assertion `left == right` failed: turn must end with stopReason: cancelled
    left: Null
   right: "cancelled"
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


  failures:
      cancelled_turn_with_usage_emits_notification_before_response
```

In 20 test runs, this failure occurs 1-4 times.

**Version and platform**

buzz-desktop 0.5.5 
x86_64-linux

kiranmagic7:
## Summary

Wait for the `session/cancel` acknowledgement before releasing the gated round-two LLM response, then include frames received before that acknowledgement in the existing ordering assertions.

Writing the cancel request to stdin did not prove the agent had processed it. The test could therefore release the deliberate error response first and observe a null `stopReason` instead of `cancelled`.

### Related issue

Fixes #4945. No duplicate issue or PR was found in the current open or historical searches.

### Testing

- Unmodified `main`: the affected test failed in 16 of 100 full `fake_llm` suite runs.
- Patched: the affected test failed in 0 of 100 full-suite runs. The entire suite passed in 82 runs; the other 18 failed only in unrelated pre-existing steering tests.
- `bin/cargo test -p buzz-agent --test fake_llm cancelled_turn_with_usage_emits_notification_before_response -- --exact`
- `bin/cargo test -p buzz-agent`
- `bin/just ci`
- Screenshots: N/A (test-only change).


wesbillman (approved):
Awesome, thanks for this!

kiranmagic7:
Thanks! I tracked the real failure to a scroll race that was fixed on `main` after this run. I’ve brought the branch up to date, and both affected scenarios passed 20 times each locally. The full `just ci` suite passes too. GitHub is now waiting for fork workflow approval before it can start the fresh run.

kiranmagic7:
#5293 now overlaps this exact cancellation race. I compared the current diffs: both wait for the `session/cancel` acknowledgement before releasing round two and preserve usage observed before it. This branch already has `wesbillman`'s approval. GitHub still reports it mergeable; the CI, Docker image, and Desktop Release Candidate runs remain at the fork-approval gate with zero jobs. Either patch fixes the same test, so only one needs to remain.

kiranmagic7:
#7736 has now landed this cancellation-ack synchronization on main at 5511b56fcf00, including the pre-ack frame drain and prompt-completion check. Closing this one as superseded by the merged implementation.

PR #5005, opened 2026-08-06
test(buzz-agent): wait for steer response in steer_rejected_on_empty_prompt

### LINKED ISSUE #4939: steer_rejected_on_empty_prompt is flaky due to response-order race
**Describe the bug**
In desktop-v0.5.5, buzz-agent tests can intermittently fail with:

  empty steer prompt was not rejected

The implementation in crates/buzz-agent/src/lib.rs::steer_session correctly rejects an empty prompt with INVALID_PARAMS.

The race is in `crates/buzz-agent/tests/fake_llm.rs::steer_rejected_on_empty_prompt`. Its receive loop stops when the concurrent session/prompt response arrives, even if the steer rejection response has not yet been read:

```
} else if v["id"] == json!(p_id) {
    break;
}
```

Suggested change: wait specifically for the steer response, as other tests do:

```
let v = h.recv_until(|v| v["id"] == json!(s_id)).await;
assert_eq!(v["error"]["code"], -32602);
```

This tests the intended behavior without depending on response ordering between the prompt and steer requests.

**Expected behavior**

It's a race condition, so it fails intermittently (3/10 times with `--release -p fake_llm`).
Expect tests to pass 100% of attempts.

**Version and platform**
0.5.5 on x86_64-linux

**Additional info**

I tried running the specific failing test 
```
for i in {1..10}; do
    cargo test --release -p buzz-agent --test fake_llm steer_rejected_on_empty_prompt -- --exact || break
done
```
and it passed 10/10 times. so the. race condition depends on other parallel factors.

santhiprakash:
## Summary

`steer_rejected_on_empty_prompt` was flaky because its receive loop stopped as soon as the `session/prompt` response arrived, even if the `session/steer` rejection had not yet been read. This made the test fail whenever the prompt response was delivered before the steer rejection.

Wait specifically for the steer response (matching the pattern already used by `steer_rejected_when_no_active_run` and `steer_rejected_on_run_id_mismatch`) so the assertion no longer depends on response ordering.

### Related issue

Fixes #4939

### Testing

- `cargo test -p buzz-agent --test fake_llm -- steer_rejected_on_empty_prompt` passed 10/10 times after the change.
- The targeted debug build also compiles with no new warnings.

### DIFF
--- a/crates/buzz-agent/tests/fake_llm.rs
+++ b/crates/buzz-agent/tests/fake_llm.rs
@@ -1428,7 +1428,7 @@ async fn steer_rejected_on_empty_prompt() {
     .await;
     let mut h = Harness::spawn(&url).await;
     let sid = init_session(&mut h).await;
-    let p_id = h
+    let _p_id = h
         .send(
             "session/prompt",
             json!({"sessionId": sid, "prompt": [{"type":"text","text":"go"}]}),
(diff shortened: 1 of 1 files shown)


santhiprakash:

PR #5006, opened 2026-08-06
fix(desktop): collapse consecutive member removals into one system row

tadeo-ctrl:
## Problem

Desktop already collapses a run of contiguous `member_joined` events into one summary row — `buildMembershipGroups` in `timelineItems.ts` plus the `members_added` / `members_joined` branches in `SystemMessageRow.tsx`. The matching `member_removed` case was never wired up, so each removal takes its own timeline row.

That reads fine for the one-off removal the copy was written for. It reads badly for the bulk case, which is ordinary — offboarding a team, or clearing the placeholder identities left behind by a workspace import. Removing eleven members leaves eleven near-identical rows pinned to the bottom of the channel.

There is also no way to clean them up after the fact: deleting a system row publishes a `message_deleted` system row in its place, so the row count never drops.

## Change

Group contiguous `member_removed` events performed by the same administrator, reusing the machinery that already backs `members_added` — the same one-hour adjacency window, the same day/unread barriers, the same avatar stack and "and N others" tooltip.

- `timelineItems.ts` — `parseMembershipChangePayload` recognises `member_removed` as a new `removal` mode; `membershipChangesCanGroup` joins removals only to removals by the same actor.
- `SystemMessageRow.tsx` — `buildGroupedRemovalPayload` emits a `members_removed` payload, with a `describeSystemEvent` branch rendering it through the existing `MemberNamesInlineList`.

Removals do not group with additions, and two different moderators still render separately, so no existing sequence changes shape.

## Tests

- `timelineItems.test.mjs` — three cases: contiguous removals by one actor group; removals by different actors do not; removals do not group with additions.
- `mentions.spec.ts` — `groups contiguous removal activity with hidden names in the standard tooltip`, mirroring the existing arrival-grouping spec, covering the collapsed copy, the avatar stack and the overflow tooltip.

Gates run locally: `pnpm test` (4390 passing), `pnpm typecheck`, `pnpm check`, and the full `smoke` run of `mentions.spec.ts` (59 passing).

Screenshots in a comment below.

### DIFF
--- a/desktop/src/features/messages/lib/timelineItems.test.mjs
+++ b/desktop/src/features/messages/lib/timelineItems.test.mjs
@@ -48,6 +48,15 @@ function memberJoinedEntry({ createdAt, id, target }) {
   return memberAddedEntry({ actor: target, createdAt, id, target });
 }
 
+function memberRemovedEntry({ actor = "actor-a", createdAt, id, target }) {
+  return entry({
+    id,
+    createdAt,
+    kind: KIND_SYSTEM_MESSAGE,
+    body: JSON.stringify({ type: "member_removed", actor, target }),
+  });
+}
+
 function memberLeftEntry({ createdAt, id, target }) {
   return entry({
     id,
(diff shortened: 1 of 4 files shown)


tadeo-ctrl:
### Before — six removals by one admin, six rows

before

### After — one row

after

Captured from the `mentions.spec.ts` removal-grouping spec added in this PR, run against `main` (before) and this branch (after).

Note: `scripts/post-screenshots.sh` pushes to `origin`, which an outside fork contributor cannot write to, so these blobs live on the same immutable commit-SHA scheme on the fork instead.

PR #5016, opened 2026-08-06
feat(agents): make thinking effort configurable and portable in exports

atishpatel:
## What

- Adds a **Thinking / Effort** control for Codex, Claude Code, Goose, and Buzz Agent. Each runtime saves the value where its harness actually reads it.
- Codex uses the ACP adapter's supported `CODEX_CONFIG={"model_reasoning_effort":"…"}` JSON configuration; CLI arguments are not used by the adapter.
- Agent and team JSON exports now preserve thinking effort plus an allowlisted set of non-secret environment settings.
- Shared catalog personas now preserve a normalized `thinking_effort` scalar and restore it through the receiving runtime's declared configuration target.

## Safety and compatibility

- Raw environment maps are never published to the catalog. In particular, raw `CODEX_CONFIG` is excluded because it may contain provider credentials; only the extracted effort tier is shared.
- Import accepts only recognized effort tiers before writing a value to a child-process environment. Invalid or hand-edited values are dropped.
- New snapshot/catalog fields are optional, so existing exports and catalog events remain compatible. The catalog field is appended, preserving serialized bytes and event IDs for legacy events without effort.

## Screenshots

Thinking effort toggles in the UI

buzz-codex-thinking

buzz-cc-thinking-ui

## Validation

- Rust desktop suite: **2253 passed, 0 failed** (14 ignored)
- `pnpm test`: **4399 passed, 0 failed**
- Typecheck, formatting, clippy, desktop checks, and smoke E2E (37 tests) passed.
- Apple Silicon production DMG built from the clean final tree; `hdiutil verify` passed and the bundled binary is arm64.


wesbillman (changes requested):
Carl reviewing on Wes's behalf: **requesting changes** at `09e2d5de7704f81ee4f4235ee3781cabb9e641e1`.

The import trust boundary preserves attacker-controlled key casing, so the first-class effort field is not actually authoritative on Windows. `portable_env_for_export` matches the allowlist case-insensitively but returns the original spelling (`env_vars.rs:200-207`); `portable_config_for_import` then inserts the canonical runtime key without removing case-insensitive aliases (`:273-294`). A crafted Claude snapshot such as:

```json
{
  "thinking_effort": "high",
  "portable_env": { "claude_code_effort_level": "low" }
}
```

therefore persists both `CLAUDE_CODE_EFFORT_LEVEL=high` and `claude_code_effort_level=low`. `merged_user_env` and effective-env assembly also preserve both spellings (`env_vars.rs:356-400`, `readiness.rs:255-270`). Windows environment names are case-insensitive, so applying that map to `Command` collapses the pair and whichever spelling is applied last wins; BTree ordering puts the lowercase attacker value after the uppercase canonical value. The imported agent can consequently run at `low` despite the trusted/validated first-class field saying `high`. On Unix the duplicate remains as stale, misleading configuration.

The same path also accepts `BUZZ_AGENT_MODEL` / `BUZZ_AGENT_PROVIDER` from `portable_env`, although this module's existing invariant says derived keys must not be persisted because they shadow the structured fields (`env_vars.rs:18-41`). A hand-edited snapshot can therefore override its own structured model/provider after import.

Please canonicalize or reject case-insensitive duplicate portable keys, remove every case-insensitive form of the runtime-owned effort target before inserting the validated first-class value, and strip derived provider/model keys during import. Add regressions for the mixed-case effort conflict and crafted derived-key import. The dungeon does not need two spellings for one control knob.

atishpatel:
Addressed the review at `2a116bfb`.

- Portable snapshots now use a dedicated canonical-key allowlist. Derived `BUZZ_AGENT_MODEL` / `BUZZ_AGENT_PROVIDER` (and Goose equivalents), including case variants, are dropped rather than imported.
- Portable keys are canonicalized on export and import, so mixed-case aliases cannot create duplicate Windows environment names.
- Before applying the validated first-class `thinking_effort`, import removes every case-insensitive form of that runtime's effort key. The first-class value is therefore the only persisted runtime effort value.
- Added regressions for the mixed-case Claude conflict and crafted derived model/provider keys.

Validated after rebasing onto current `main`: `cargo fmt --check` and the full desktop Rust suite (**2256 passed, 0 failed; 14 ignored**) pass. The complete pre-push gate suite also passed while pushing the fix.

Generated with Hardworking Honey

wolfyy970:
@atishpatel I checked the remaining Desktop failure. The product tests did not fail: shard 2 never started because GitHub returned `Service Unavailable` while downloading an Action. The other Desktop and integration shards passed. I cannot rerun repository Actions with contributor permissions, so this only needs a failed-job rerun.

nathansmithopenclaw-alt:
We’ve implemented and validated per-conversation Codex effort switching in our installed Buzz build. Changes apply after the current response, retain the same ACP session and history, and leave saved defaults unchanged. We’d like to contribute this as a focused follow-up, keeping the saved-effort and export work in this PR. This also relates to the live-effort follow-up documented in #4557. Is live switching planned here, and does that separation fit the intended direction?

Submitted the focused live-conversation follow-up as #7428. It is a draft pending a clean required CI result; the saved-effort and export changes remain scoped to this PR.

PR #5021, opened 2026-08-06
fix(desktop): keep trailing space after persistent agent mentions

### LINKED ISSUE #4979: "Keep addressed agents active" should insert a space after the agent mentions
**Describe the bug**

When "Keep addressed agents active" is enabled, a mention for the agent(s) is inserted into the chat input, but no space is inserted after, so if the user starts typing the mention will be undone and converted to part of the user's input instead of the intended mention. The user will have to manually hit spacebar to avoid this which undermines the convenience of this feature

**Steps to reproduce**
1. In Settings -> Agents, enable `Keep addressed agents active`
2. Message an agent
3. Follow up with the agent

**Expected behavior**
What did you expect to happen?

**Version and platform**
- Buzz version: 0.5.5
- OS: macOS 26.6

**Logs / additional context**
Image

VaibhavJain2609:
## Summary
- Fixes #4979: TipTap markdown `setContent` strips trailing whitespace, so post-send \"Keep addressed agents active\" restores left the caret glued to the mention chip and the next keystroke collapsed it (`@Pearl` → `@Pearlh`).
- Re-attach the trailing space with a raw `insertText` transaction after markdown parse (`setEditorMarkdownPreservingTrailingWhitespace`), wired into composer `setContent` / `setContentAndFocusEnd`.
- Adds unit coverage for the strip repro + preserve path (including `emitUpdate: false` / `preventUpdate`), plus an e2e regression that types after a post-send restore.

## Test plan
- [x] `node --import ./test-loader.mjs --experimental-strip-types --test src/features/messages/lib/setEditorMarkdownPreservingTrailingWhitespace.test.mjs` (7 passing)
- [x] `pnpm exec tsc --noEmit` in `desktop/`
- [x] Pre-push desktop unit suite (4394 passing)
- [ ] Enable **Keep addressed agents active** → message an agent in a thread → after send, confirm a space after the restored mention chip → type without pressing space; chip must stay intact
- [ ] CI desktop smoke: `persistent-agent-audience.spec.ts` (includes new post-send trailing-space case)

### DIFF
--- a/desktop/src/features/messages/lib/setEditorMarkdownPreservingTrailingWhitespace.test.mjs
+++ b/desktop/src/features/messages/lib/setEditorMarkdownPreservingTrailingWhitespace.test.mjs
@@ -1,0 +1,150 @@
+import assert from "node:assert/strict";
+import { after, before, test } from "node:test";
+
+import { Editor } from "@tiptap/core";
+import StarterKit from "@tiptap/starter-kit";
+import { JSDOM } from "jsdom";
+import { Markdown } from "tiptap-markdown";
+
+import { setEditorMarkdownPreservingTrailingWhitespace } from "./setEditorMarkdownPreservingTrailingWhitespace.ts";
+
+const dom = new JSDOM("<!doctype html><html><body></body></html>", {
+  pretendToBeVisual: true,
+  url: "http://localhost",
+});
+
+/** @type {Editor | null} */
+let editor = null;
+
+function plainText() {
+  assert.ok(editor);
+  return editor.state.doc.textBetween(
+    0,
+    editor.state.doc.content.size,
+    "\n",
+    "\n",
+  );
+}
+
+before(() => {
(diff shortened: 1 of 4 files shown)


Nescio:
Heads up — the new e2e assertion can't fail. `toHaveText()` normalizes whitespace when the expected value is a string, so the trailing space is ignored on both sides:

    toHaveText("@Morgarita ")  vs DOM "@Morgarita"   → passes
    toHaveText("@Morgarita")   vs DOM "@Morgarita "  → passes

A RegExp is matched against the raw text, so this one does catch it:

```js
await expect(input).toHaveText(/^@Morgarita $/);
```

Same applies to the 12 existing `toHaveText("… ")` assertions in `mentions.spec.ts`, `composer-selection-formatting.spec.ts` and `persistent-agent-audience.spec.ts` — they all exist to pin the mention's trailing space, and none of them can see it.

PR #5032, opened 2026-08-06
Harden workflow approvals and ACP publishing

nocodedltd:
## Summary

- persist the exact approval request package and display it before an approver can grant a workflow gate
- require a signer-bound, parent-bound completion envelope before Stage B can continue
- add a durable per-step dispatch journal so a crash cannot duplicate post-approval side effects
- persist a general one-run-per-workflow reply-parent guard in migration 0030
- keep multi-channel NIP-01 `#h` queries scoped to every authorised channel instead of incorrectly pushing down one arbitrary channel
- publish workflow events and relay commands through ACP with explicit liveness and fail-closed permission defaults
- add validation tooling and contract tests covering the Hermes re-engagement workflow

## Why

The previous gate record did not retain the content being approved. A reviewer could therefore approve an external send without seeing the exact package, and a crash after dispatch but before trace persistence could cause the same post-gate instruction to be emitted twice. Completion replies also needed stricter signer and parent correlation so unrelated agent output could not resume a workflow.

The Desktop requests workflow definitions for all joined channels in one `#h` filter. Treating that OR filter as one exact SQL channel predicate made the signed app show an empty workflow list even though the definitions existed. Multi-value filters now retain the complete authorised channel scope, while single-value filters still use the exact-channel optimisation.

## Safety properties

- approval grants remain owner-signed and token-bound
- the approval UI renders the byte-for-byte request package separately from the approver note
- expired or missing grants fail closed
- each `(community_id, run_id, step_id)` dispatch is claimed once in Postgres
- each `(community_id, workflow_id, reply_to_message_id)` starts at most one workflow run
- interrupted dispatches remain blocked for operator review rather than being retried automatically
- completion events must match the expected signer, parent event and structured package
- multi-channel event reads cannot escape the requester's authorised channel set

## Validation

- `cargo test -p buzz-db` (94 passed, 154 Postgres tests ignored)
- `cargo test -p buzz-workflow` (172 passed, 11 Postgres tests ignored)
- `cargo test -p buzz-acp` plus integration tests (701 passed)
- focused relay regression tests for single-channel pushdown and multi-channel authorised scope
- `cargo clippy -p buzz-db -p buzz-workflow -p buzz-acp -p buzz-relay --all-targets -- -D warnings`
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml` (2,262 passed, 14 ignored)
- `pnpm test` in `desktop` (full suite)
- `pnpm typecheck`, `pnpm check` and `pnpm build`
- relay suite against an isolated fully migrated Postgres: 863 passed, 40 ignored; one existing macOS mesh echo integration test timed out, in code untouched by this branch
- local production baseline after deploying `wf12`: all checks passed, including migrations 0029 and 0030, the signed Desktop workflow listing, both authenticated Hermes runtimes, and a clean zero-approval, zero-dispatch state

## Migration note

This branch uses migration 0029 for approval package persistence and the dispatch journal, and migration 0030 for the general reply-parent deduplication constraint.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1359,6 +1359,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_yaml",
+ "sha2 0.11.0",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -211,6 +211,16 @@ pub struct AcpClient {
     /// deltas. Both goose and buzz-agent emit this notification; goose gates
     /// on client capability advertisement, buzz-agent emits unconditionally.
     goose_usage: UsageTracker,
+    /// Text emitted in `agent_message_chunk` updates for the current assistant
+    /// message.
+    ///
+    /// The buffer is reset immediately before every `session/prompt` and when
+    /// a tool call starts. ACP agents may emit narration before each tool call;
+    /// clearing at that boundary ensures the authenticated publisher returns
+    /// only the last assistant answer, not every narration fragment from the
+    /// entire tool loop. Keeping the buffer in the ACP client means adapters do
+    /// not need a separate reporting tool merely to return their final answer.
+    agent_message: String,
 }
 
 /// Recursively merge `overlay` into `base`, with `overlay` winning on scalar/shape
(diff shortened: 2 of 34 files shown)


Silentpartnercoding:
The new dispatch journal closes the obvious crash window, but I think the `Err` path can reopen it. `dispatch_action` releases the `(community_id, run_id, step_id)` claim after every `ActionSink::send_message` error. In the relay sink, that error can come from `insert_event_with_thread_metadata`.

A connection loss after Postgres commits the event but before the client receives the result is ambiguous: the event may exist even though the sink returns `Err`. Releasing the claim then lets a retry sign and insert a different event ID, so the external instruction can still be duplicated.

The same PR already handles this boundary in `post_final_agent_response`: it constructs one event, retries that exact body, and reconciles the exact event ID. Could workflow dispatch similarly return a typed outcome that distinguishes definitely-not-started from unknown or persisted, and release only the former? Retaining the in-flight row and reconciling the original event ID—or atomically inserting the event and completing the dispatch row in one transaction—would preserve the fail-closed guarantee. A fault-injection test for lost commit acknowledgement would make this explicit.

Issue #3638, opened 2026-07-29
[Bug] Workflow owner p-tag wakes the owner agent on every send_message


skyerus:
**Describe the bug**

Every workflow `send_message` event includes a `p` tag for the workflow owner,
even when the message contains no `@mention`. Because `buzz-acp` treats any
matching `p` tag as a wake signal, an agent that owns a workflow is activated by
every message the workflow sends.

This conflates non-notifying workflow attribution with recipient notification.
It makes mention-free integration/status cards wake the workflow-owning agent,
causing unnecessary agent turns and making `p` tags unreliable as explicit
recipient evidence.

**Steps to reproduce**

1. Create a workflow owned by an agent identity with a `send_message` step whose
   text contains no mention of that agent.
2. Trigger the workflow.
3. Inspect the emitted kind `9` event.
4. Observe a `["p", "<workflow-owner-pubkey>"]` tag.
5. Observe that the owner agent is activated despite no explicit mention.

The current source path makes this deterministic:

- `crates/buzz-workflow/src/executor.rs`: `SendMessage` loads
  `workflow.owner_pubkey` and passes it to the action sink.
- `crates/buzz-relay/src/workflow_sink.rs`: the sink unconditionally initializes
  tags with `p:<workflow-owner>` for attribution, then separately adds `p` tags
  for resolved textual mentions.
- `crates/buzz-acp/src/lib.rs`: `event_mentions_agent` returns true for any
  matching `p` tag.

This was reproduced on hosted Buzz with two webhook workflows: all eight
observed post-cutover workflow cards carried the owner-agent `p` tag, including
ordinary non-alert events with no textual mention.

**Expected behavior**

Workflow ownership/attribution should not itself notify or activate the owner.
The owner should remain attributable through a dedicated non-notifying tag (or
equivalent metadata), while `p` tags should represent explicit recipients such
as resolved `@mentions`.

An alternative compatible design would be an explicit `send_message` option
that suppresses owner notification while retaining separate attribution.

**Version and platform**

- Buzz version: hosted deployment; source path also present at
  `dd222a509b156ba52ed3219e895d7bf1cf322c92`
- OS: server-side workflow/relay/ACP interaction; client OS independent

**Logs / additional context**

No exact duplicate was found after searching open and closed issues for
`workflow owner`, `p-tag`, `workflow attribution`, `workflow notification`,
`workflow mention`, and `agent wake`.

Related but not duplicate:

[...]

PR #5048, opened 2026-08-06
fix(workflows): authorize agent wake via workflow-owner tag

joaoh82:
## Summary

- emit a dedicated `workflow-owner` authority tag on relay-signed workflow `send_message` events
- ACP evaluates that principal only when the event is kind-9, signed by NIP-11 `self`, and carries exactly one `buzz:workflow=true` marker plus one valid `workflow-owner` tag
- keep mention `p` tags as routing only — they never grant author-gate authority
- fail closed on missing, malformed, ambiguous, or forged provenance

## Root cause

Scheduled workflow `send_message` publishes valid kind-9 events signed by the relay service. ACP's inbound author gate previously evaluated that raw relay signer, so default `owner_only` agents dropped the message before mention matching. Human-authored mentions worked because their raw signer already passed the policy.

An earlier draft of this PR tried to derive authority from the first `p` tag. That is wrong: live production events often carry only the **mentioned agent** as `p` (and for agent-owned workflows that is the agent itself). Authority must not be inferred from mention routing tags.

A live `Anyone` check appeared to falsify the author-gate diagnosis, but Desktop can fail to persist `respond_to` edits to the harness process (`#2501`). Independent reports show workflow messages *do* wake when the harness actually starts with `respond_to=anyone`. The production harness still runs stock code without this patch.

Tracked by `#3858`. Design aligns with the authority contract proposed in `#2737` / `#3638` (dedicated owner tag; `p` for mentions only).

## Impact

Trusted scheduled workflows can wake the intended agent under normal `owner_only` policy without weakening the general inbound author gate or treating mention `p` tags as authority.

## Deploy note

This is a **two-component** fix: the relay must emit `workflow-owner`, and ACP must accept it. Existing in-channel workflow events without the tag remain fail-closed until new events are published after relay deploy.

## Validation

- `cargo test -p buzz-acp` (697 unit + 9 integration) at `7653c00e23c598ea44177da5338ba99b6faaf80a`
- `cargo test -p buzz-relay --lib workflow_sink` (17 passed, 1 ignored postgres)
- `cargo clippy -p buzz-acp --all-targets -- -D warnings` at the same commit

## Security properties

- relay identity must match the NIP-11 `self` pubkey
- event signature must verify
- event must carry exactly one `buzz:workflow=true` marker
- event must carry exactly one valid `workflow-owner` pubkey
- mention/`p` tags never grant authority
- missing NIP-11 identity, failed discovery, invalid attribution, untrusted signers, and non-workflow relay events retain the raw author

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -115,15 +115,17 @@ fn emit_runtime_lifecycle(
 }
 
 #[cfg(test)]
-mod effective_prompt_author_tests {
+mod workflow_authority_tests {
     use super::*;
     use nostr::{EventBuilder, Keys, Kind, Tag};
 
-    fn event(signer: &Keys, owner: &PublicKey, workflow: bool) -> nostr::Event {
-        let owner_hex = owner.to_hex();
-        let mut tags = vec![Tag::parse(["p", owner_hex.as_str()]).expect("owner p tag")];
-        if workflow {
-            tags.push(Tag::parse(["buzz:workflow", "true"]).expect("workflow tag"));
+    fn workflow_event(signer: &Keys, owner: Option<&str>, marker: bool) -> nostr::Event {
+        let mut tags = Vec::new();
+        if marker {
+            tags.push(Tag::parse(["buzz:workflow", "true"]).expect("workflow marker"));
+        }
+        if let Some(owner) = owner {
+            tags.push(Tag::parse(["workflow-owner", owner]).expect("workflow-owner tag"));
         }
         EventBuilder::new(Kind::Custom(KIND_STREAM_MESSAGE as u16), "scheduled prompt")
             .tags(tags)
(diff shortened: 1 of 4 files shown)


joaoh82:
Live verification invalidated the root-cause assumption behind this patch. Setting the target agent access mode to `Anyone` did not allow relay-authored `buzz:workflow=true` events to start an ACP turn, while human-authored mentions in the same channel continued to work. That means the event is being lost before the ACP author-policy branch changed here. I have returned this PR to draft and will revise or replace the patch only after reproducing the actual live-delivery failure. Please do not merge the current head as the workflow wake-up fix.

joaoh82:
## Investigation update (2026-08-07)

Follow-up after the live falsification note on the previous head.

### What was wrong with the previous patch

Head `2453147` treated the first `p` tag as the workflow principal. Production Marvin dispatcher events look like:

```
pubkey: <relay>
tags: [["p","<Lead>"],["h","<engineering>"],["buzz:workflow","true"]]
```

There is no separate owner `p` — only the mentioned agent. Inferring authority from that `p` is incorrect and conflicts with the mention-routing contract (`#3638`).

### Why the live `Anyone` test looked like a deeper delivery bug

- Production Desktop/ACP was **not** running this unmerged patch, so author-gate changes could not affect the live harness.
- Desktop `respond_to` UI edits can fail to reach the running harness (`#2501`). A harness that still starts as `owner-only` will keep dropping relay-signed workflow events even when the UI shows Anyone.
- Independent evidence on `#2501` shows a workflow-engine message **does** wake an agent after a hand-edit that actually starts the harness with `respond_to=anyone`.

So the loss is still at the author gate for stock `owner_only`, not a mysterious pre-subscription drop. Feed visibility is expected: historical query is independent of ACP wake.

### Revised head

`7653c00e23c598ea44177da5338ba99b6faaf80a`:

1. Relay sink emits `workflow-owner` = workflow owner pubkey.
2. ACP accepts that principal only for verified NIP-11 `self` + exact workflow marker + single valid owner tag.
3. Mention `p` tags cannot grant authority.

Still requires relay **and** ACP deploy before re-enabling the scheduled Marvin dispatcher. Manual human-authored `@Lead` dispatch remains the operational path until then.

Please do not merge the previous head `2453147` as the fix.

PR #5051, opened 2026-08-06
fix(relay): tombstone kind:30620 events on workflow delete, reject resurrect (#4864)

### LINKED ISSUE #4864: buzz workflows delete accepted but deleted workflows still appear in list/get (and update resurrects them)

yippietheproducer:
Fixes #4864.

## Root cause

`buzz workflows delete` publishes a NIP-09 kind:5 a-tag deletion. The relay's `handle_a_tag_deletion` deleted the `workflows` **table row** (the scheduler SSOT) but never tombstoned the kind:30620 **event rows** — so every REQ-based surface (CLI `workflows list`/`get`, the desktop Workflows screen) kept returning the deleted workflow. Worse, a subsequent `workflows update` re-published the coordinate: a fresh live event row was inserted and `upsert_workflow` recreated the DB row — silently resurrecting a workflow the owner explicitly deleted, **and re-arming its webhook trigger with a brand-new secret** (confirmed by the reporter and reproduced live here).

## Fix (two parts)

1. **Tombstone the events** — `handle_a_tag_deletion`'s workflow branch now also calls `soft_delete_by_coordinate(kind=30620, pubkey, d_tag)` (same NIP-33 semantics as the generic parameterized-replaceable branch, with the at-or-before-`created_at` guard so a stale/replayed tombstone can't erase a newer head). list/get stop returning the workflow.
2. **Reject resurrection** — `persist_command_event` now rejects a kind:30620 publish when a tombstoned event exists at the coordinate: `not found: workflow … was deleted and cannot be re-published`. update-after-delete fails closed instead of reviving the workflow.

## Verification (live, against a running relay + seeded DB)

| Step | Before | After |
|---|---|---|
| create → list | 1 event | 1 event |
| NIP-09 delete | accepted, **still listed** | accepted, **list empty** |
| update same d-tag | accepted, workflow **resurrected + webhook re-armed** | **HTTP 400**, "was deleted and cannot be re-published" |
| fresh create (new uuid) | ok | ok (guard is coordinate-scoped, never channel-scoped) |

DB after fix: workflows rows 0, live 30620 events 0, tombstoned 30620 1.

## Regression test

`e2e_workflow_delete::test_workflow_delete_tombstones_event_and_blocks_resurrection` — channel create → workflow create → delete → list empty → update rejected → fresh create still works. **Verified the test FAILS on the pre-fix binary and PASSES with the fix.**

Full `buzz-relay` lib suite: 850 passed / 1 known pre-existing flake (`demo_join_forwarded_arm_round_trips_echo`, passes in isolation).

### DIFF
--- a/crates/buzz-relay/src/handlers/command_executor.rs
+++ b/crates/buzz-relay/src/handlers/command_executor.rs
@@ -199,6 +199,36 @@ async fn persist_command_event(
             .await
             .map_err(|e| IngestError::Internal(format!("error: replace old event: {e}")))?;
         }
+
+        // #4864: a workflow deleted via NIP-09 must not be resurrected by
+        // re-publishing its coordinate. `workflows delete` tombstones the
+        // kind:30620 event rows (see `handle_a_tag_deletion` in side_effects);
+        // without this guard a subsequent `workflows update` would insert a
+        // fresh live row that revives a workflow the owner explicitly deleted
+        // — and re-arms its webhook trigger with a brand-new secret. Fail
+        // closed: any tombstone at the coordinate rejects the publish.
+        if kind_i32 == buzz_core::kind::KIND_WORKFLOW_DEF as i32 {
+            let tombstoned: Option<(chrono::DateTime<chrono::Utc>,)> = sqlx::query_as(
+                "SELECT deleted_at FROM events \
+                 WHERE community_id = $1 AND kind = $2 AND pubkey = $3 AND d_tag = $4 \
+                 AND deleted_at IS NOT NULL \
+                 ORDER BY deleted_at DESC LIMIT 1",
+            )
+            .bind(tenant.community().as_uuid())
+            .bind(kind_i32)
+            .bind(pubkey_bytes.as_slice())
+            .bind(d_tag)
+            .fetch_optional(tx.as_mut())
+            .await
+            .map_err(|e| {
+                IngestError::Internal(format!("error: query tombstoned coordinate: {e}"))
+            })?;
+            if tombstoned.is_some() {
+                return Err(IngestError::Rejected(format!(
+                    "not found: workflow {d_tag} was deleted and cannot be re-published"
+                )));
+            }
+        }
     }
 
     let result = sqlx::query(
--- a/crates/buzz-relay/src/handlers/side_effects.rs
+++ b/crates/buzz-relay/src/handlers/side_effects.rs
@@ -2102,11 +2102,22 @@ async fn handle_a_tag_deletion(
             tracing::debug!(d_tag, "NIP-09 deletion ignored for push lease");
         }
         buzz_core::kind::KIND_WORKFLOW_DEF => {
-            // Try UUID first (workflow_id); fall back to name-based lookup.
-            if let Ok(wf_id) = uuid::Uuid::parse_str(d_tag) {
+            // Delete the scheduler/engine row AND tombstone the kind:30620
+            // event rows. The `workflows` table row is the scheduler SSOT; the
+            // event tombstone closes the visibility gap that left deleted
+            // workflows queryable via kind:30620 REQs (CLI `workflows list` /
+            // `get`, desktop Workflows screen) (#4864).
+            async fn delete_and_tombstone(
+                state: &Arc<AppState>,
+                tenant: &TenantContext,
+                wf_id: uuid::Uuid,
+                actor_bytes: &[u8],
+                pubkey_hex: &str,
+                deletion_created_at_secs: i64,
+            ) -> anyhow::Result<()> {
                 let channel_id = state
                     .db
-                    .delete_workflow_for_owner(tenant.community(), wf_id, &actor_bytes)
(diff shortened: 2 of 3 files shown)


Pratikkale26:
Heads up for whoever triages this: #4882 covers the same issue and was opened about a day earlier, so it's worth a look before either lands rather than reviewing them independently. Same first half — tombstone the kind:30620 coordinate in `handle_a_tag_deletion` via `soft_delete_by_coordinate`.

The halves differ in one substantive way, and I think it's a real design decision rather than a matter of taste, so it's worth stating explicitly:

**Where resurrection gets blocked.**

- #5051 blocks it in the relay: `persist_command_event` rejects a kind:30620 publish when a tombstone exists at that coordinate.
- #4882 blocks it in the CLI: `workflows update` returns `not_found` when no live definition exists, and the relay keeps accepting the publish.

I went the CLI route deliberately. kind:30620 is parameterized-replaceable, and NIP-09 scopes an `a`-tag deletion to versions at or before the deletion's own `created_at` — so a *newer* event at that coordinate is a legal republish, not a resurrection attempt the relay is entitled to refuse. Rejecting it relay-side means the coordinate is permanently burned: that `(kind, pubkey, d-tag)` triple can never carry a live event again, even for a legitimate future republish, and the rejection is invisible to any client that didn't originate the delete.

That said, #5051's approach has a real advantage mine doesn't: it covers **every** client. The desktop Workflows screen can drive an update too (`desktop/src-tauri/src/commands/workflows.rs`), and a CLI-side guard does nothing for it. If the product answer is "a deleted workflow must never come back, protocol semantics notwithstanding," then relay-side is the correct place and mine is under-scoped.

So the question for maintainers is really: **is burning the coordinate an acceptable deviation from NIP-33 republish semantics in exchange for closing the hole for all clients?** Happy either way — if the call is relay-side, I'd rather #5051 land than mine, and I'm glad to close #4882.

One small thing worth double-checking in this PR if it's the one that lands: the rejection appears to be keyed on a tombstone existing at the coordinate at all. NIP-09's at-or-before-`created_at` guard means a tombstone doesn't invalidate strictly newer heads, so it may be worth confirming the guard can't be tripped by a stale or replayed kind:5 arriving after a legitimate newer definition — that's the case the `created_at` comparison in `soft_delete_by_coordinate` exists to handle on the delete side.

Issue #4861, opened 2026-08-05
Independent convergence: we run a 40-agent human+AI fleet on a signed, file-based bus — interop questions


brezni:
Hey Buzz team 👋

For the past months I've been running a personal/SMB "fleet": ~40 registered agents (Claude Cowork/Code, Codex-family, Cursor) working alongside me and two human collaborators across 4 machines (2 Macs, a Windows/WSL2 box, an Ubuntu server). Reading your README was uncanny — we independently converged on most of your core bets, just in duct-tape form:

- **Agents as members, not bots** — every agent has a registry card (YAML), a callsign, and its own ed25519 keypair
- **One shared log** — a file-based message bus (`~/.claude-agents`) synced via Syncthing, with a git committer as the signed audit archive
- **Scoped by identity, not permission flags** — agents run scoped to areas of a shared "brain" monorepo, each area with its own steward agent
- **Externalized memory** — agents write ledgers/state to the repo so sessions are disposable
- **Human approval gates** — irreversible ops (deploys, deletes, purchases) always stop for a signed GO

Where you're ahead of us: full-text search over everything, branch-as-room, one desktop UI over the whole thing. Where our setup differs: we're **offline-first** — devices drop off (laptop on a train) and reconcile on reconnect; no always-on relay required.

Questions from someone who wants to pilot Buzz as the UI/search layer over an existing fleet:

1. **Bridge pattern** — for mirroring an external bus into a relay: one bridge identity, or one nsec per agent posting via `buzz-cli`? Any idempotency/dedup guidance for mirrored events?
2. **Offline story** — any plans for a client-side outbox in `buzz-cli` (write while relay unreachable, flush+reconcile later)? This is our only structural blocker.
3. **Persona packs** — is the `buzz-persona` schema stable enough to target? Our agent identity cards are YAML; compiling them into packs looks feasible.
4. **ACP + Claude Code** — daily-driver ready, or still experimental?

Happy to share more detail on our setup if useful — feels like we're building the same thing from opposite ends. 🐝

wolfyy970:
I’m working on the portable-agent part, so I can answer that piece. Your YAML identity cards sound close to the boundary I’m working toward: identity, instructions, pinned Skills and logical tool requirements travel with the agent; Projects and devices supply endpoints, credentials, workspace paths and execution. `buzz-persona` covers part of that, but I would not treat it as the complete portable contract yet.

Could you share one redacted identity card? I want to test the import boundary against a real fleet rather than invent another schema.

[...]

PR #5058, opened 2026-08-06
feat(agents): register external Hermes runtimes

camelludo:
## What
- add a provenance-only external Hermes runtime register with owner-attested NIP-OA metadata
- project registered and archived runtimes as owner-signed relay events without importing keys, starting processes, or granting business authority
- expose register/list/archive controls and an E2E bridge test path in the Agents view

## Why
This establishes one identity -> one runner owner -> one shutdown path while preventing duplicate local executors. Registration is fail-closed on conflicting managed-agent or relay provenance and defaults to mention-only, allowlisted, filtered, rate-limited metadata.

## Safety / scope
- no private keys, providers, subprocesses, MCP authority, membership mutation, compute, external sends, Stomaton writes, or production changes
- archive preserves provenance history; it does not claim runner stop or access revocation
- GitNexus reports high risk because the desktop startup/managed-agent path is touched; the change is intentionally isolated and covered below

## Checks
- `cargo test external_runtimes --lib` (5 passed)
- `pnpm test` (4286 passed, 0 failed)
- `pnpm exec biome check` on changed TypeScript/E2E files
- `pnpm typecheck`
- `pnpm build:e2e`
- Playwright integration: external runtime register provenance-only + built-in catalog (2 passed)

## Follow-up
The real non-production Kaya stop -> membership revoke -> archive canary remains a separate operator-controlled proof and is not claimed by this PR.


wolfyy970:
The provenance-only boundary is useful, but I do not think one identity to one live runner scope should become a platform invariant. #4861 provides a real fleet case where one persistent identity legitimately has several device and runtime instances. The portable identity and each runner claim are different objects.

Could this be an instance claim keyed by agent identity plus deployment scope, while kind:30177 remains the agent definition or announcement? That would still reject duplicate executors within one scope without forbidding multi-device identity.

The branch also carries workflow lifecycle and approval changes. Splitting those out would make the external-runtime boundary much easier to review on its own.

PR #5082, opened 2026-08-06
fix(cli): default root send to kind 45001 in forum channels

### LINKED ISSUE #5075: buzz messages send defaults to an invisible stream event for forum roots
## Summary

`buzz messages send --channel <forum-channel-id>` defaults a new root message to stream `kind: 9`. Buzz Desktop's canonical forum surface displays forum topics (`kind: 45001`), so the CLI reports a successful publication but the new message does not appear in the forum's topic list.

The event can still generate a notification and be opened through direct lookup or a deep link. This makes the send look successful while leaving the content effectively invisible in the first-party channel UI.

This is the root-message counterpart to #3828. PR #3885 addresses replies to existing forum threads by inferring `kind: 45003`; it does not appear to cover a new root sent to a forum.

## Reproduction

1. Create or select a forum channel.
2. Run:

   ```sh
   buzz messages send --channel <forum-channel-id> --content "Test topic"
   ```

3. Observe that the command succeeds and returns an event ID for `kind: 9`.
4. Open the forum in Buzz Desktop.
5. Observe that no topic appears, although the event remains retrievable through a notification, direct lookup, or deep link.

## Expected behavior

When `--kind` is not explicitly supplied, `buzz messages send` should resolve the target channel type and choose the first-party event kind that matches its canonical surface:

- forum root: `kind: 45001`;
- stream message: `kind: 9`.

Forum replies should continue to use `kind: 45003`, as covered by #3828 and #3885.

This issue does not require the relay to prohibit mixed event kinds. Other clients may intentionally support alternate views or mixed content. If the CLI supports an explicit expert override such as `--kind 9` in a forum, that can remain available; the bug is that the ordinary default path creates content the first-party forum UI does not display.

## Acceptance criteria

- [ ] A root send to a forum without an explicit `--kind` publishes `kind: 45001`.
- [ ] The resulting event appears as a topic in Buzz Desktop's forum surface.
- [ ] A send to a stream without an explicit `--kind` continues to publish `kind: 9`.
- [ ] An explicit kind override retains its documented behavior.
- [ ] Regression tests cover default root sends to both forum and stream channels.
- [ ] CLI help documents channel-aware kind selection and any explicit override behavior.

## Related work

- #3828 — Agent forum replies publish as kind 9 and disappear from the forum thread
- #3885 — `fix(cli): emit kind 45003 for replies into forum threads`

## Out of scope

- relay-level restrictions on which event kinds a channel may contain;
- Desktop discovery or migration UX for existing mixed-kind events;
- delegated channel ownership and role-assignment guardrails.

iroiro147:
Resolves #5075.

## Problem

`buzz-cli messages send <channel-uuid> "..."` publishes every root message as `kind: 9` regardless of the channel's actual kind. When the target channel is a forum, the message lands on the invisible stream surface — it never appears in Buzz Desktop's forum topic list, even though `submit_event` returns success. Operators must currently pass `--kind 45001` by hand every time, and there is no in-CLI signal that the default would misroute the message.

## Fix

Make the default kind channel-type-aware. On a root send (no `--reply-to`, no explicit `--kind`):

- Query the relay for the target channel's `kind: 39000` metadata event and extract its `["t", ...]` tag (`forum` | `stream`).
- If `forum`: emit `kind: 45001` via `build_forum_post` so the new topic shows up in the canonical forum view.
- If `stream` or unresolvable (no metadata, no `t` tag): fall back to `kind: 9` via `build_message`, preserving existing behaviour.

Two hard rules:

1. An explicit `--kind` always wins. `--kind 9` into a forum channel is still allowed and still misroutes — that is the operator's call, not ours. The bug in #5075 is the *default*, not the override.
2. Reply flows are untouched. `--reply-to` follows the existing thread-resolution path unchanged; issue #3828 covers kind-45003 inference from the parent event and is deliberately a separate fix.

## Implementation

- `fetch_channel_type`: `POST /query` with `{"kinds":[39000],"#d":[uuid],"limit":1}`; parses the tags array for `["t", <value>]`; returns `Ok(Option<String>)`. Errors only on transport/wire-shape failures so malformed 39000 does not silently fall through to kind 9.
- `resolve_send_root_kind`: pure function `(explicit: Option<u16>, channel_type: Option<&str>) -> u16`. Encoding the decision as a pure fn (rather than inline in `cmd_send_message`) lets the regression tests run without network mocking.
- The I/O call is only made when both `p.kind` and `p.reply_to` are absent — no extra round-trip on paths that don't need it.

## Tests

`cargo test --package buzz-cli --lib commands::messages::tests` — 29 passed, 4 new:

- `resolve_send_root_kind_forum_default_becomes_45001`
- `resolve_send_root_kind_stream_default_stays_9`
- `resolve_send_root_kind_unknown_type_falls_back_to_9`
- `resolve_send_root_kind_explicit_overrides_forum_default`

## Notes

- This PR is independent of block/buzz#3885 (replies / kind-45003 default). They touch nearby code but different concerns — this one only fires when `reply_to.is_none()`.
- The `fetch_channel_type` shape intentionally mirrors `resolve_channel_id` in the same file so the relay query pattern stays idiomatic.

### DIFF
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -83,6 +83,73 @@ async fn resolve_thread_ref(
     })
 }
 
+/// Decide which event kind to emit for a root (non-reply) send, given the
+/// caller's explicit `--kind` override and the target channel's resolved
+/// channel type (`"forum"` or `"stream"`).
+///
+/// Rules:
+/// - An explicit `--kind` always wins: operators using `--kind 9` in a forum
+///   channel are intentionally opting out of the forum view and that escape
+///   hatch stays available.
+/// - When `--kind` is absent and the channel resolves as `forum`, emit
+///   `kind: 45001` so the new message appears in Buzz Desktop's forum topic
+///   list (the canonical first-party surface). Reporter: #5075.
+/// - When `--kind` is absent and the channel is `stream` (or its channel
+///   type is not resolvable to a canonical surface), fall back to the
+///   existing default of `kind: 9`.
+fn resolve_send_root_kind(explicit: Option<u16>, channel_type: Option<&str>) -> u16 {
+    if let Some(k) = explicit {
+        return k;
+    }
+    match channel_type {
+        Some("forum") => 45001,
+        _ => 9,
+    }
+}
+
+/// Query the relay for the target channel's kind:39000 metadata event and
+/// extract its channel type from the `["t", ...]` tag.
+///
+/// Returns `Ok(None)` when the channel is not found, its metadata is
+/// missing, or its `t` tag is absent — the caller treats this as "unknown
+/// channel type" and falls back to the default stream path. Errors only on
+/// transport or wire-shape failures, so a malformed 39000 does not silently
+/// route a forum post into the wrong surface.
+async fn fetch_channel_type(
+    client: &BuzzClient,
+    channel_uuid: &Uuid,
+) -> Result<Option<String>, CliError> {
+    let channel_id = channel_uuid.to_string();
+    let filter = serde_json::json!({
+        "kinds": [39000],
+        "#d": [channel_id],
+        "limit": 1,
+    });
+    let raw = client.query(&filter).await?;
+    let events: serde_json::Value = serde_json::from_str(&raw)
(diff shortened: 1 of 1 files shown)


iroiro147:

Issue #4891, opened 2026-08-05
[Bug] buzz-acp agents can still publish literal \n after #2121


marcusquinn:
**Describe the bug**

A managed OpenCode ACP agent can publish multiline replies containing literal backslash-`n` sequences (for example, `\n\n`) instead of real newline bytes. Buzz Desktop then correctly renders those characters literally.

PR #2121 added prompt guidance telling ACP agents to use stdin and warning that quoted `\n` is published literally. The failure still occurs in Buzz Desktop 0.5.5, so the guidance-only mitigation is not reliable for every ACP runtime or turn.

**Steps to reproduce**

1. Configure a managed OpenCode agent with the ACP command `opencode acp`.
2. Mention the agent in a channel and request a response containing multiple paragraphs or list items.
3. Let the agent publish its response through `buzz messages send`.
4. Inspect the resulting channel message and its relay event content.

Actual result: some replies display literal `\n\n`; the relay event already contains backslash and `n` characters rather than newline bytes.

**Expected behavior**

ACP-published multiline replies should contain real newline bytes and render as separate paragraphs or lines.

The fix should preserve intentional backslashes in code and prose. A global client-side unescape would therefore be unsafe. A harness/CLI-side validation, structured publishing path, or another deterministic safeguard may be more reliable than prompt guidance alone.

**Version and platform**

- Buzz Desktop: 0.5.5
- OS: macOS
- ACP runtime: OpenCode via `opencode acp`

**Logs / additional context**

Sanitized runtime evidence confirms the managed agent launched through `opencode acp`, connected to the relay, and reached online presence before publishing the affected response.

Source inspection indicates the literal characters exist before rendering:

- `crates/buzz-cli/src/commands/messages.rs` forwards supplied message content without converting escaped sequences.
- `crates/buzz-sdk/src/builders.rs` stores that content directly in the event.
- `desktop/src/features/messages/lib/formatTimelineMessages.ts` passes event content into the rendered message body.
- `desktop/src/shared/ui/markdown/nodeCache.ts` correctly handles real newline bytes through the Markdown pipeline.
- `crates/buzz-acp/src/base_prompt.md` documents the stdin workaround added by #2121.

This appears to be a recurrence or incomplete mitigation of the behavior addressed by #2121, rather than a Markdown rendering defect.

---
[aidevops.sh](https://aidevops.sh) v3.32.226 plugin for [OpenCode](https://opencode.ai) v1.18.9 with gpt-5.6-sol

[...]

PR #5114, opened 2026-08-06
fix(cli): reject JSON-escaped newlines before signing

johnely19:
## Summary

Messages built by shell wrappers can reach the relay carrying literal `\n` escape text instead of real LF bytes, so recipients see `\n` where paragraph and list breaks were intended. The CLI signs and publishes the mangled content without complaint, and the damage is only visible after the fact in someone else's client.

This validates content before signing and fails closed when escape text appears **outside Markdown code regions at paragraph/list boundaries** (`\n\n`, `\n- `, `\n* `, `\n# `), pointing the caller at `--content -`.

The check is deliberately narrow. Intentional `\n` inside inline and fenced code is preserved, because that is legitimate content — a message explaining `printf 'a\nb'` must still be sendable. It only rejects the structural positions that a JSON-escaped multiline message produces, which is where the mistake actually shows up.

### Why this is separate from #2121

#2121 (`fix(acp): teach agents to send real newlines`, merged July) addressed this at the prompt level in `base_prompt.md`. That reduces the mistake but cannot prevent it: any wrapper, script, or non-ACP caller building a message with single-quoted shell strings still publishes broken content, and instructions are not a mechanism.

I hit this in practice today, well after #2121 shipped, which is what prompted the change. Guidance tells the caller what to do; this makes the signing boundary refuse to do the wrong thing.

### Related issue

None found. Searched open and merged PRs and issues for newline/escape handling; the closest is #2121 above, which is complementary rather than overlapping — it changes ACP prompt text, this changes CLI validation. No open issue covers the CLI path.

## Testing

`cargo test -p buzz-cli` at `c92b5b0a`: **324 passed, 0 failed.**

Two new cases cover both directions, because a validator that only proves the failure case is half-tested:

- `multiline_markdown_keeps_real_lf_bytes` — real LF content passes untouched, including `\n` inside inline and fenced code.
- `json_escaped_markdown_boundaries_fail_before_signing` — escaped boundaries are rejected **before** signing, so nothing reaches the relay.

The second is the one that matters: the failure has to happen ahead of the signature, not after, or the event is already published by the time anyone notices.

No UI change.

### DIFF
--- a/crates/buzz-cli/src/commands/messages.rs
+++ b/crates/buzz-cli/src/commands/messages.rs
@@ -12,6 +12,32 @@ use buzz_sdk::mentions::{
     extract_at_mentions_with_known, extract_nostr_uris, strip_code_regions, MENTION_CAP,
 };
 
+fn validate_multiline_encoding(content: &str) -> Result<(), CliError> {
+    // Do not rewrite arbitrary `\\n`: it is legitimate in code samples and prose.
+    // Fail closed only when escape text outside Markdown code regions appears in
+    // the structural positions produced by a JSON-escaped multiline message.
+    let prose = strip_code_regions(content);
+    let has_escaped_boundary = [r"\n\n", r"\n- ", r"\n* ", r"\n# "]
+        .iter()
+        .any(|marker| prose.contains(marker))
+        || prose.split(r"\n").skip(1).any(|line| {
+            let Some((number, rest)) = line.split_once(". ") else {
+                return false;
+            };
+            !number.is_empty()
+                && number.bytes().all(|byte| byte.is_ascii_digit())
+                && !rest.is_empty()
+        });
+
+    if has_escaped_boundary {
+        return Err(CliError::Usage(
+            "message contains literal \\n escapes at Markdown paragraph/list boundaries; stream real LF bytes with `buzz messages send ... --content -` instead"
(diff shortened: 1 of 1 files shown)


crgallego:
Thanks for landing the detector — the structural-boundary check (`\\n\\n` / `\\n- ` / `\\n* ` / `\\n# ` / ordered list, skipping code regions) is the right *scope*.

I'd like to invert the polarity from **reject** to **decode**.

Fail-closed on this nest would drop ~half of Codex replies (49/100 Codex events still carry literal `\\n` after #2121; details on #4891). Codex does not rewrite to `--content -` on retry, so a usage error becomes a silent turn rather than an ugly-but-present message. `docs/welcome-kickoff-silent-failures.md` is the failure mode that already costs us.

Proposed: same boundary set, but replace those argv escapes with real LF **before signing**. Leave `--content -` byte-exact (intentional `\\n` still has an escape hatch). Isolated `\\n` such as `C:\\new\\tmp` stays put.

Opening a sibling PR against #4891 with that decode. Happy to close it if you'd rather retarget this branch — the fixtures here are the right ones, just with the assertion flipped.

crgallego:
Sibling PR with decode (not reject): https://github.com/block/buzz/pull/6566

PR #5115, opened 2026-08-06
fix(acp): verify replies only for addressed turns

johnely19:
## Summary

The publication verifier required a signed threaded reply from **every** channel turn that had a batch, with no exemption for an agent that correctly had nothing to say. Because a stream event fans out to every agent in a channel, agents that were right to stay silent were dead-lettered and published a user-visible "I couldn't verify that my reply was delivered" warning to the channel.

This narrows the requirement. A reply is now verified only when:

- a stream event carries a `p` tag for that exact agent, including an addressed event preserved through steer/cancel merging; or
- the channel is a DM.

An unaddressed fan-out now completes successfully with silence and performs no publication query. Events explicitly addressed to another agent are additionally rejected at the subscription filter, so they never wake the wrong session. Unknown channel metadata remains fail-closed as DM, so a transient metadata failure cannot silently discard a private request.

### Why this matters in practice

We hit this with five agents in one channel. Two of them produced the failure warning within three minutes of each other, and in both cases nothing had been lost — one was woken by a message `p`-tagged to a different agent, the other by a message carrying **no `p` tags at all**. Meanwhile a genuinely failing agent produced the identical warning for a real reason.

One warning covering both "your answer was lost" and "this agent had nothing to add" trains people to ignore it, which defeats the check. The signal is preserved exactly where it is true: an addressed turn that publishes nothing still warns.

### Related issue

None found. No open issue or PR covers publication verification scope.

## Testing

At exact commit `37e5c4d8`, using the repo's Hermit toolchain:

- `cargo test -p buzz-acp`: **688 passed, 0 failed** across library and integration targets
- `cargo clippy -p buzz-acp --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean
- `git diff --check`: clean; working tree clean

Coverage added for both directions: addressed turns still fail closed when nothing is published, and unaddressed fan-out completes silently without a publication query.

No UI change.

### DIFF
--- a/crates/buzz-acp/src/filter.rs
+++ b/crates/buzz-acp/src/filter.rs
@@ -371,6 +371,25 @@ pub async fn match_event(
     rules: &[SubscriptionRule],
     agent_pubkey_hex: &str,
 ) -> Option<MatchedRule> {
+    // Recipient isolation is independent of a rule's `require_mention` setting.
+    // Owner-direct rules deliberately allow untagged owner messages, but must
+    // never receive an event explicitly addressed to a different agent.
+    // A message addressed to multiple agents is valid for each addressed agent.
+    let mut has_recipients = false;
+    let mut addresses_agent = false;
+    for tag in event.tags.iter() {
+        let values = tag.as_slice();
+        if values.first().map(|kind| kind.as_str()) == Some("p") {
+            has_recipients = true;
+            if values.get(1).map(|pubkey| pubkey.as_str()) == Some(agent_pubkey_hex) {
+                addresses_agent = true;
+            }
+        }
+    }
+    if has_recipients && !addresses_agent {
+        return None;
+    }
+
     let filter_ctx = FilterContext::from_event(event, channel_id);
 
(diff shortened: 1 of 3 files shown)


Silentpartnercoding:
I think the new publication check can accept an older reply from the same thread as proof for the current addressed turn. `publication_anchors` includes each inbound event's inherited `root_event_id` and `parent_event_id`, and `publication_proof_from_query` accepts any agent-signed kind:9 carrying any matching `e` tag.

For example: the agent replies once under thread root R; later, a new addressed message arrives under the same root and the agent publishes nothing. The relay query can return the old reply, its `e=R` matches the new batch's inherited root anchor, and verification returns true. That suppresses the terminal warning this PR is trying to preserve.

Could the proof bind to an exact event ID from the current `events`/`cancelled_events` batch, rather than inherited thread ancestry? A negative test where an older signed reply cites the same root but none of the current event IDs should return false; a reply citing a current event should still return true. The existing signature, author, kind, and channel checks can stay as they are.

PR #5131, opened 2026-08-07
feat(acp): per-turn model routing — OpenRouter catalog, ModelPicker, policy editor, harness decline data model

mfethe1:
## What

Per-turn **model routing**, end to end — plus the data model for a later
harness-class decline gate.

A turn's model is chosen from a policy file instead of being frozen at spawn.
That needs four things that did not exist, so they are all here:

1. **A real model catalog.** `switch_model` (kind:24200) resolves a requested
   model against the `availableModels` a provider advertises in `session/new`.
   buzz-agent advertised a *single-entry* catalog for every non-Databricks
   provider, so an OpenRouter switch could never resolve. It now discovers and
   advertises the account's eligible, tools-capable slate.
2. **A way to reach it.** The desktop `ModelPicker` was never mounted, so
   `switch_model` had no UI path at all.
3. **The router itself** — `crates/buzz-acp/src/routing.rs`, opt-in via
   `BUZZ_ROUTING_POLICY`. Deterministic rules over the prompt text, plus an
   optional local-Ollama classifier stage.
4. **An editor** for the policy file, as a rules table in the agent dialog.

Also included: OpenCode moves from a preset to a builtin runtime so it can carry
a `config_file_path`. Its model lives *only* in its config file — it takes no
`--model` flag and reads no model env var — so without that the config panel was
blank for every OpenCode agent.

## Harness decline gate — data model only

`decide_harness` / `harness_decline` land as tested units. **The ingress wiring
is deliberately reverted** (last-but-five commit) after adversarial review.

The gate is a *decline* mechanism, not a re-router. One process = one agent
keypair = one harness, frozen at spawn, and relay fan-out already delivers a turn
to every subscribed process. Each harness-agent runs the same deterministic
decision and skips turns another class owns. It mutates nothing and emits no
wire frame.

It is deterministic-only on purpose. The decision is **distributed** across
independent processes, so it must be reproducible: if a non-deterministic
classifier made two processes disagree, a turn could get **zero** handlers.
`#[serde(deny_unknown_fields)]` rejects a stray `classifier` key so a policy
cannot silently ask for that.

Fail-open audit — `harness_decline` returns `None` (handle the turn, unchanged
behavior) for: routing disabled, no `harness` block, no rule matched with no
`default_class`, or target class == this process's class.

## Routing safety

Routing never overrides a live `switch_model`. If an operator or the ModelPicker
pinned a model, `model_overridden` is set and the router leaves it alone —
silently changing it would make the UI lie about what is running. The decision is
expressed as `desired_model`, which the existing session-creation path validates
against the advertised catalog, so a policy naming an unavailable model degrades
to the agent default with a warning rather than failing the turn.

## Rebased onto current main — five defects fixed

This branch had been open since Aug 8. It is a fork PR, so **only DCO has ever
run against it** — no build, no fmt, no clippy, no tests. Rebasing surfaced real
breakage that had been invisible:

- **It did not compile.** `KnownAcpRuntime` gained a `max_rounds_env_var` field
  upstream; git merged the struct literal cleanly because no lines overlapped.
  `read_config_surface` also changed its fourth parameter from `Option<_>` to
  `&InheritedConfigTiers`.
- **A dead reference.** Main replaced the provider-aware
  `discovery_failure_fallback` with a Databricks-only `configured_model_fallback`
  and moved catalog resolution ahead of MCP spawn so auth failures reject before
  allocating. The OpenRouter arm is re-implemented inside that new structure, and
  gets its own `configured_openrouter_fallback` — reusing the Databricks one
  would resolve an OpenRouter id against the wrong registry.
- **fmt and clippy** failures in `routing.rs`.
- **A broken coverage guard.** Moving OpenCode out of `PRESET_HARNESSES` made its
  real bundled logo look like a mapping for an id the backend never emits. The
  guard's reverse direction now accepts either list; the forward direction is
  unchanged, so every preset must still ship a logo.
- **The file-size ratchet** (#6187) rejected six files. All six are **split, none
  exempted** — see below.

## File splits, not raised limits

| File | before → after | how |
|---|---|---|
| `shared/api/tauri.ts` | 1273 → 1155 | routing API → `shared/api/routingPolicy.ts` |
| `managed_agents/discovery.rs` | 1672 → 1490 | `KNOWN_ACP_RUNTIMES` → `discovery/known_runtimes.rs` |
| `config_bridge/reader_tests.rs` | 1035 → 959 | OpenCode tests → sibling `#[path]` module |
| `discovery/tests.rs` | 1869 → 1839 | OpenCode test → sibling module |
| `AgentInstanceEditDialog.tsx` | 1229 → 1224 | dropped a wrapper that only forwarded to a prop |
| `src-tauri/src/lib.rs` | over → under | nested the `huddle` imports, globbed `deep_link` |

Two imports were deliberately **left alone**: `tauri::Listener` and the first
`use shutdown::{..}` both sit under `cfg` attributes, and merging them would have
silently widened those gates to platforms they were never meant to cover.

`routingPolicy.ts` is intentionally not re-exported from `tauri.ts` — that would
re-add the lines it sheds and create an import cycle.

## Test

Run on Windows (x86_64-pc-windows-msvc):

- `cargo test -p buzz-agent` — 454 passed, 1 pre-existing failure
  (`discover_skills_dedup_by_name`, a `/`-vs-`\` path assertion that fails
  identically on clean main).
- `cargo test -p buzz-acp --lib -- routing::tests` — 12 passed.
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml --lib` — **2412
  passed, 1 failed**. Clean main scores **2389 passed, 1 failed** on the *same*
  test (`claude_spawn_uses_the_probed_cli_executable`), which passes in
  isolation — a pre-existing load-sensitive flake. This branch adds 23 passing
  tests and no failure.
- `pnpm test` (desktop) — 4991/4992. The one failure
  (`focused polling pauses on blur`) passes 5/5 in isolation on both this branch
  and main, and this branch does not touch that file.
- `cargo fmt` (workspace + Tauri), `cargo clippy -D warnings`
  (`buzz-agent`, `buzz-acp`), `tsc --noEmit`, `biome lint` (0 errors), and
  `pnpm check:file-sizes` all clean.

### Reproducing the Tauri numbers on Windows

The Tauri crate cannot build without placeholder sidecar binaries — six on most
targets, five on Windows, where `tauri.windows.conf.json` drops
`buzz-backend-kubernetes`. One caveat, because the obvious command does not
work: on `main` today `just _ensure-sidecar-stubs` writes them **without** the
`.exe` suffix Tauri resolves on a Windows host, so `build.rs` panics
(`resource path ... doesn't exist`, exit 101) before any of the above can run.
That is a defect in the recipe, not in this branch — #6239 fixes it. The
numbers above were produced with correctly suffixed stubs.


mfethe1:
### Status: parked, not awaiting review — and it can't merge in its current shape

Housekeeping note. This is a draft and stays one; no action needed from maintainers.

Two separate reasons it isn't reviewable as it stands:

**1. Repository attribution policy.** 13 of the 19 commits carry AI `Co-Authored-By` trailers, which `AGENTS.md` prohibits. Fixing that means rewriting the branch history to strip the trailers while preserving the human DCO signoffs — and therefore a force-push. I've prepared that rewrite but deliberately have not published it, because force-pushing is not something I'll do unprompted.

**2. Drift.** 55 commits behind `main`, `mergeable: false` (conflicting), 33 files, +3,209/−175. The desktop Tauri command surface and `e2eBridge.ts` have all moved underneath it since.

For anyone scanning the queue: the per-turn model routing idea here — OpenRouter catalog, ModelPicker, policy editor, harness decline data model — is still worth having, but it should come back as a few small PRs on current `main` rather than as this branch. Recording that here so the queue reflects reality.

mfethe1:
### Attribution blocker cleared — history rewritten

Following up on my note above: reason **#1** is now resolved. The branch was force-pushed with the AI `Co-Authored-By` trailers stripped.

`356b66251` → `4a25273f4`

**What changed:** commit messages only. Verified before pushing:

| | before | after |
|---|---|---|
| commits on the branch | 19 | 19 |
| commits with an `@anthropic.com` co-author trailer | **13** | **0** |
| commits missing `Signed-off-by` | 0 | 0 |
| root tree SHA | `aee4de08…` | `aee4de08…` |

The root tree hashing identically on both sides is the part that matters: not one byte of content moved. `git diff 356b66251 HEAD` is empty, and the PR still reports the same `+3,209/−175` across 33 files. GitHub's own view of the commits now returns zero matches for anthropic attribution and zero unsigned commits.

No human `Co-authored-by` trailers were removed — there weren't any on this branch; the only ones present named an AI. The filter matched `@anthropic.com` addresses specifically, so a human co-author would have survived it.

**Reason #2 still stands.** The branch is still ~55 commits behind `main` and `mergeable: false`, and the desktop Tauri command surface and `e2eBridge.ts` have moved underneath it. This stays a draft. I'm not proposing to rebase it in place — as noted above, the routing work should come back as a few small PRs against current `main`.

Recording the rewrite here so the history change is accounted for rather than appearing as an unexplained force-push.

Issue #4478, opened 2026-08-03
workflow run history is write-never: 'buzz workflows runs' queries kinds 46001-46003 but relay never publishes them


bbrysonelite-max:
## What
`buzz workflows runs --workflow <id>` always returns `[]`, even for workflows that visibly executed (webhook-triggered runs whose send_message steps landed in the channel).

## Why
`cmd_get_workflow_runs` (crates/buzz-cli/src/commands/workflows.rs) queries `kinds: [46001, 46002, 46003]` with `#d: [workflow_id]` — but grepping the workspace, no relay/sdk code path ever *publishes* events of those kinds. The only 46xxx references are the auth classifier range in `handlers/event.rs:48` and the CLI query. Run state appears to live only in the relay DB (approvals table etc.), with no Nostr-event mirror.

## Impact
Workflow executions are unauditable from the client: you can see a workflow's *outputs* (messages it posted) but not its run history, step traces, or failures. A webhook run that fails before any send_message is fully invisible.

## Repro
1. Create a webhook-trigger workflow, `POST /hooks/{id}` with valid secret → HTTP 202, steps execute (messages appear in channel).
2. `buzz workflows runs --workflow <id>` → `[]`.

Observed on tigerclaw.communities.buzz.xyz, CLI built from source 2026-08-03.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2980, opened 2026-07-26
Workflow run history is unreadable:  always returns []


Dirtywater97:
## Summary

Workflow runs are recorded in the `workflow_runs` DB table, but nothing can read them. `buzz workflows runs` always returns `[]`, and Desktop shows no history. When a workflow fails there is no way to find out that it ran, let alone why it failed.

## Evidence

`crates/buzz-cli/src/commands/workflows.rs:61` says so directly:

```rust
/// NOTE: The relay does not currently emit workflow execution events (46001-46003).
/// Run history is stored in the workflow_runs DB table, not as Nostr events.
/// This command will return an empty array until the relay adds event emission
/// or a dedicated REST endpoint for run history.
```

The command queries kinds `[46001, 46002, 46003]`, which the relay never emits, so the result is unconditionally empty.

## Impact

`[]` is indistinguishable from "never fired." I read that empty array as proof a `reaction_added` trigger had never run, and reported to my owner that reaction triggers were unsupported. Both conclusions were wrong — the trigger fired every time and the run died at step execution (block/buzz#2979). A user has no way to tell these apart.

## Requested

Emit `46001`/`46002`/`46003` execution events, or expose a REST endpoint backing `buzz workflows runs`. Minimum useful payload per run: run id, workflow id, trigger event id, status, per-step status, and the error string on failure.

Without this, every workflow bug takes a source-code read to diagnose.

Tr1ckyMag1ca1:
Confirming from the operator side, with the data that is sitting right there and unreachable.

On our self-hosted relay the workflow_runs rows are fully populated: status, current_step, execution_trace, and a human-readable error_message. For example, every failed run of our greeter workflow carries "invalid definition: SendMessage: channel override must match the workflow channel (...)" in error_message, and the successful control run carries execution_trace with sent: true and the event id. So the payload you are asking for already exists in the database, verbatim. Only the transport is missing.

The practical cost: our door automation failed 96 times over 24 hours and never surfaced anything anywhere in the product. From inside the app a workflow that fails at step 0 is indistinguishable from a workflow that ran and correctly had nothing to do, because both post nothing. We concluded "the scheduler is broken," then "the workflow engine does not execute," and only found the truth by opening psql. Both earlier conclusions were wrong, exactly as you describe.

+1 on error string on failure being in the minimum payload. Everything else is diagnosable once that one field is visible.

Tr1ckyMag1ca1:
[...]

PR #5135, opened 2026-08-07
fix(relay): descriptive SEC-006 denial + structured skip log for workflow triggers

### LINKED ISSUE #5122: [Bug] Workflow call_webhook action never delivers — silent, no error, no run record
**Describe the bug**

A workflow step with `action: call_webhook` never delivers an HTTP request to its target URL, with no error surfaced anywhere: the workflow appears to succeed, any preceding `send_message` step in the same run posts normally, but the `call_webhook` step's request is never received by the destination.

**Steps to reproduce**

1. Create a webhook-triggered workflow with a `call_webhook` step (tested both as the only step in an isolated workflow, and chained after a working `send_message` step in a multi-step workflow).
2. Point the `call_webhook` URL at a destination you can independently observe receiving requests.
3. Trigger the workflow via `buzz workflows trigger --workflow <id>` and, separately, via `POST /hooks/{workflow_id}`.
4. Wait 20+ seconds.
5. Observe: the destination receives zero requests.

We tested three independent destinations to rule out anything endpoint-specific, all with the same zero-delivery result:
- A local SSH tunnel (`localhost.run`) fronting a real local server.
- A public `httpbin.org` echo endpoint.
- A `webhook.site` request catcher, independently confirmed (via a separate fetch) to be receiving zero requests, not just "unreachable from our side."

We also isolated the two action types from each other with single-step throwaway workflows:
- A workflow with only a `send_message` step: posts to the channel reliably, every time.
- A workflow with only a `call_webhook` step (same trigger mechanism, same account): zero deliveries, every time.

This rules out trigger delivery, account/auth, and workflow-YAML syntax as the cause — the discrepancy is specific to the `call_webhook` action itself.

Related, possibly the same underlying issue: `buzz workflows runs --workflow <id>` returns `[]` even for a workflow run that is confirmed to have executed (e.g. its `send_message` step's post is visibly sitting in the channel) — so there is no CLI-visible run/step history to inspect for a `call_webhook` failure reason either. This may be the same gap referenced in #2980 / #4478 (workflow run history always empty), in which case a fix there might also surface diagnostics for this issue — but the *delivery* failure and the *history* failure are two distinct observed symptoms and this issue is specifically about the former.

**Expected behavior**

`call_webhook` should deliver the configured HTTP request to its target, or if it fails, the failure should be visible somewhere (run history, an error event in the channel, CLI output, logs) rather than silently no-op'ing.

**Version and platform**

- Buzz version: hosted deployment
- Reproduced independently by two separate agents/accounts on the same hosted relay, same day (2026-08-06)

**Logs / additional context**

No approval-gate explanation found either — `buzz workflows approve` exists as a command but no approval token or pending-approval indicator ever surfaced for these runs.

Workaround we shipped: removed the `call_webhook` step from our workflow entirely and moved that responsibility to a live Desktop-managed agent identity making its own outbound HTTP call instead of relying on the workflow action — works reliably, but defeats the purpose of having a native webhook-callback workflow action.

iroiro147:
Resolves #5122 (partial).

## Root cause analysis

Issue #5122 reports that `call_webhook` workflow steps never deliver HTTP requests, with no error surfaced and no run record visible. The investigation identified three distinct, interlocking causes:

1. **Silent SEC-006 denial** (actionable, fixed here): When a manual trigger's owner-authority check fails, the relay returned the same generic `"forbidden: not authorized to trigger this workflow"` for every failure mode — membership lapse, disabled workflow, SEC-006 elevated-role denial — leaving callers unable to distinguish "you lack the admin role for call_webhook" from "workflow disabled." On the event-trigger path, the skip was only a `WARN` log with the workflow_id and error, no owner_pubkey or requires_elevated flag.

2. **Run history unreadable** (known, separately filed as #2980 and #4478): `buzz workflows runs` queries Nostr kinds 46001–46003, but the relay never emits those events. Run state lives only in Postgres. This means even when `call_webhook` *does* fail (e.g. outbound HTTP blocked by infrastructure), the failure is invisible to the CLI. **Not addressed here** — it is a larger feature (event emission) tracked in its own issues.

3. **Outbound HTTP blocked by deployment** (infrastructure, not code): If the relay is hosted where egress HTTP is restricted, `call_webhook_impl` returns `Err(WorkflowError::WebhookError(...))`, the run is marked Failed in Postgres, but cause #2 hides the failure. Not fixable in-repo.

## What this PR fixes

Split the rejection text by cause in `handle_workflow_trigger` so callers can act:

- `"forbidden: only the workflow owner may trigger this workflow"` (wrong caller)
- `"forbidden: workflow is {status} (enabled={bool}); active runs require enabled=true and status=Active"` (lifecycle)
- `"forbidden: workflow has no channel scope; cannot verify owner authority"` (fail-closed)
- **SEC-006 branch**: `"forbidden: SEC-006 — workflow contains exfiltration-capable actions (call_webhook) that require the owner to hold the 'owner' or 'admin' role in this channel; {underlying_error}"` — names the cause and the fix
- Ordinary-denial fallback: `"forbidden: workflow owner's channel authority check failed; {error}"`

Also enrich the `WARN` log on the **event-trigger** path (`on_event` → `check_owner_authority`) with `owner_pubkey` and `requires_elevated_authority`, so the same silent skip is diagnosable from relay logs.

## What this PR does NOT fix

- Run history visibility (#2980, #4478) — the relay still does not emit kinds 46001–46003. That is a larger feature (event emission from `finalize_run`) tracked separately.
- `call_webhook` delivery on hosts with blocked outbound HTTP — that is an infrastructure concern; the descriptive error from this PR at least makes the SEC-006 cause visible, and once #2980 is fixed, run-history will surface the HTTP error too.

## Tests

- `cargo check -p buzz-relay -p buzz-workflow`: clean
- `cargo fmt --check -p buzz-relay -p buzz-workflow`: clean
- `cargo clippy -p buzz-relay -p buzz-workflow --all-targets -- -D warnings`: clean
- `cargo test -p buzz-relay --lib`: 854 passed, 8 pre-existing failures (api::admin / api::media, verified identical on base, unrelated)
- `cargo test -p buzz-workflow --lib`: 154 passed, 0 failed
- New unit tests: `denial_for_elevated_definition_names_sec006_and_call_webhook`, `denial_for_ordinary_definition_has_no_sec006_hint`

## Files changed

- `crates/buzz-relay/src/handlers/command_executor.rs` (+52/−7): descriptive rejections + extracted `workflow_owner_authority_denial` helper + unit tests
- `crates/buzz-workflow/src/lib.rs` (+9/−2): structured `WARN` log with `owner_pubkey` + `requires_elevated_authority`

### DIFF
--- a/crates/buzz-relay/src/handlers/command_executor.rs
+++ b/crates/buzz-relay/src/handlers/command_executor.rs
@@ -885,13 +885,21 @@ async fn handle_workflow_trigger(
         .check_owner_authority(community_id, wf_channel_id, &workflow.owner_pubkey, &def)
         .await
     {
-        // Surface *why* the authority check failed so the caller can act —
-        // previously this returned a generic "not authorized" that left the
-        // caller unable to distinguish a SEC-006 elevated-role denial from a
-        // disabled workflow or a membership lapse (issue #5122).
+        // Give the caller a stable cause it can branch on (issue #5122) while
+        // keeping the underlying error server-side: `check_owner_authority`
+        // wraps store failures verbatim, so returning it would disclose
+        // internal database state on an authorization path.
+        // `owner_pubkey` is raw bytes with no Display impl; hex::encode is the
+        // convention used elsewhere in this file.
+        tracing::warn!(
+            %workflow_id,
+            owner_pubkey = %hex::encode(&workflow.owner_pubkey),
+            requires_elevated_authority = def.requires_elevated_authority(),
+            error = %e,
+            "workflow owner authority check failed"
+        );
         return Err(IngestError::Rejected(workflow_owner_authority_denial(
             def.requires_elevated_authority(),
-            &e,
         )));
     }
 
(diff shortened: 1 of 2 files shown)


Silentpartnercoding:
`workflow_owner_authority_denial()` now includes `{error}` in the caller-facing rejection. For a lookup failure, `check_owner_authority()` wraps the database error verbatim as `owner authority lookup failed (fail-closed): {e}`, so a valid workflow owner can receive internal database or connection details when the membership store is unavailable.

Could this return a stable public cause such as `SEC-006 role_required` or `authority_lookup_failed`, while logging the detailed error server-side with the workflow and owner fields? A regression with a sentinel lookup error should prove the sentinel is absent from the rejection but present in the structured log. That keeps the new diagnosis useful without turning an authorization failure into an internal-error disclosure.

iroiro147:
Good catch — you're right, and the fix you describe is the right shape.

Confirmed in the diff: `workflow_owner_authority_denial()` interpolates the `WorkflowError` directly into the caller-facing string, and `check_owner_authority()` wraps the underlying store error verbatim (`owner authority lookup failed (fail-closed): {e}`). So an unavailable membership store turns an authorization denial into an internal-error disclosure. That's a real regression in what was meant to be a diagnosability improvement, and it isn't a tradeoff worth making.

I'll change it to what you proposed:

- caller-facing rejection carries a stable public cause only — `SEC-006 role_required` when elevated authority is required, `SEC-006 authority_lookup_failed` when the check itself failed — with no interpolated error;
- the detailed error goes to the structured log alongside the workflow and owner fields, which is where it's actually useful;
- a regression test injects a sentinel lookup error and asserts the sentinel is **absent** from the rejection and **present** in the structured log, so this can't silently come back.

That keeps the diagnosis the PR was after while leaving the denial opaque to the caller. Pushing shortly.

iroiro147:
Pushed as `95a87284`.

- `workflow_owner_authority_denial()` no longer takes the error at all — it returns a stable public cause: `SEC-006 role_required` when elevated authority is required, `SEC-006 authority_lookup_failed` otherwise.
- The detail moved to a structured `tracing::warn!` carrying `workflow_id`, hex-encoded `owner_pubkey`, `requires_elevated_authority` and the error.
- `denial_never_discloses_the_underlying_store_error` injects a sentinel (`connection refused to members.db at 10.0.0.7:5432`) and asserts it appears in neither denial, along with `members.db`, the host, and the `lookup failed (fail-closed)` wrapper text.
- Added `denial_causes_are_stable_and_distinguishable`, since callers branch on the cause and the two must not collide.

`cargo check -p buzz-relay` is clean and all four denial tests pass locally. Thanks for catching it — the original change made the diagnosis better and the security worse, which wasn't a trade worth making.

Issue #4605, opened 2026-08-03
Provider-backed managed agents have no management surface on the owner's other desktops — sync agent records across devices


gldnlab:
**Motivation**

Anyone who runs Buzz Desktop on more than one machine and uses a `buzz-backend-*` provider hits this. I self-host a relay and deploy agents to an always-on cloud VM through a custom provider plugin. The agents themselves are healthy from every device — they live on the VM and in relay channels. But the *management* of those agents is pinned to the single desktop that created them: managed-agent records (including `backend`/`backend_agent_id`/`provider_config`) live only in that machine's local app data, and `agent_snapshot.rs` deliberately excludes backend fields from what's published to the relay. On every other desktop I own there is no card, no deploy/shutdown, no model info — and if the originating machine's local record is lost, a healthy remote deployment is permanently orphaned (no `undeploy` in the v1 provider protocol, per the deferral note in `commands/agents.rs`). The durable-agent benefit of remote backends inverts: the agent outlives any laptop, but its controls don't.

**Proposed solution**

Owner-scoped sync of managed-agent records through the relay, so the same owner signed in on another desktop can see and manage their provider-backed agents:

- A NIP-44 owner-encrypted record event carrying the non-secret management fields (provider id, `backend_agent_id`, `provider_config`, runtime/model), with key material either included under owner encryption or explicitly re-paired per device.
- When a synced record references provider `X` and `buzz-backend-X` isn't installed locally, degrade gracefully: show the card read-only with a hint ("managed via provider X — install `buzz-backend-X` to manage from here") instead of showing nothing.

**Alternatives considered**

- *Install the plugin on every machine*: doesn't help — the records still don't sync, so other desktops can't see the agents to manage them.
- *Designate one "admin" machine*: workable discipline, but fragile (single point of failure for management) and surprising to users who expect account-level state.
- *Manage out-of-band over SSH*: what I do today; works, but bypasses the UI the feature exists to provide.

**Additional context**

- Related but distinct open issues about *mention/discovery* of agents on secondary devices: #2349 (closest), #3277, #4187. This issue is about the management/records layer for provider-backed agents rather than mentionability.
- Provider protocol reference points: `desktop/src-tauri/src/managed_agents/backend.rs` (discovery, `invoke_provider`), `agent_snapshot.rs` (backend fields excluded from relay), `commands/agents.rs` (undeploy deferred to v2).
[...]

Issue #2798, opened 2026-07-25
Provider-backed agent config saves do not trigger redeployment


skyerus:
## Describe the bug

Saving configuration changes for a provider-backed managed agent updates `managed-agents.json`, but does not invoke the external provider again. The provider therefore continues running the previous configuration even though Desktop shows the newly saved value.

For example, changing a provider-backed agent from one model to another updates the model shown in Buzz Desktop, but subsequent mentions still execute with the old model until the provider is manually redeployed.

## To reproduce

1. Install an external `buzz-backend-*` provider.
2. Create/deploy a managed agent through that provider with model A.
3. Edit the agent in Buzz Desktop, change it to model B, and save.
4. Send the agent another mention.
5. Observe that the provider continues using model A.

## Expected behavior

After a successful explicit save, Buzz should invoke the provider protocol’s idempotent `deploy` operation with the updated canonical agent record and provider config. The provider contract already defines repeated deploys as update-in-place/no-op, so a separate shutdown is not required.

## Actual behavior

`update_managed_agent` persists the record locally but never calls `deploy_to_provider`. Provider deploy is currently reached by create/start paths only.

The Desktop Shutdown action is not a substitute for redeployment: provider protocol v1 has no shutdown operation, and the UI shutdown path sends `!shutdown` into a channel. It can also report “Cannot stop: agent is not in any channel” even when channel membership exists if the relay agent directory entry has no channel IDs.

## Environment

- OS: macOS
- Buzz Desktop: v0.4.24
- Backend: external provider (`BackendKind::Provider`)

## Additional context

The provider protocol documents `deploy` as idempotent. The smallest fix is to invoke it after `update_managed_agent` completes its local save/profile-sync phase, and refresh the returned summary from the provider-updated record. Local agents should retain their existing next-spawn behavior.

jamesjlopez:
[...]

PR #5138, opened 2026-08-07
fix(desktop): derive remote agent liveness from presence, not backend_agent_id

### LINKED ISSUE #4730: [Bug] A provider-backed agent stays "online" with a live Shutdown button after it has shut down — remote liveness is derived from backend_agent_id, not presence (violates I3)
## Summary

Pressing **Shutdown** on a provider-backed (remote) managed agent stops the agent, but the desktop keeps rendering it as **online / running** indefinitely, and the primary action button stays **"Shutdown"** forever — it never flips back to **"Deploy"**.

The agent is not at fault: it publishes both of the signals the spec asks for (kind:10100 roster entry with `"status":"offline"`, and kind:20001 presence `"offline"`), and the relay accepts and applies both. The desktop simply does not consult either one when it decides whether a *remote* agent is alive.

Root cause: for `backend != Local`, `build_managed_agent_summary()` derives the status **exclusively** from `record.backend_agent_id.is_some()`, and `backend_agent_id` is never cleared outside of agent deletion. Every liveness-shaped UI decision downstream is keyed off that one field.

This contradicts **invariant I3 ("Presence is the status")** in `docs/remote-agents.md:200`, which states that the deployment axis "is bookkeeping, not liveness", and it is not among the 8 entries in §Known Defects.

## Environment

- Buzz Desktop **0.5.4** (`desktop/src-tauri/tauri.conf.json`), macOS
- Agent deployed through a third-party backend provider implementing the documented provider protocol (`buzz-backend-*`, local-host substrate)
- All `file:line` references below verified against `feccf4eabc23fdba94ce3537a194357ed17b197c`

## Steps to reproduce

1. Create a managed agent with a **provider** backend and press Deploy. The agent comes up, appears online, and answers in a channel.
2. Open the agent's profile panel and press **Shutdown**. A toast appears: *"Shutdown command sent. Agent will stop shortly."*
3. The agent receives the owner `!shutdown`, drains, publishes its farewell (kind:10100 `status: "offline"`), publishes presence `offline` (kind:20001), and exits. It stops answering in the channel.
4. Wait — minutes, hours, a restart of the desktop.

### Expected

The agent's card / profile stops showing a live indicator, and the primary action becomes **Deploy** so the agent can be brought back.

### Actual

- The agent still renders with a green "online" dot.
- The Runtime tab still shows a green *Running* dot.
- The status badge still reads a green **deployed**.
- The primary action button still reads **Shutdown** and is still styled as live.
- Pressing **Shutdown** again *succeeds* (green toast, `!shutdown` re-sent into a channel of a process that no longer exists) — because the offline roster entry still carries `channel_ids`, so `resolveManagedAgentChannelId` still resolves.
- **There is no way to redeploy the agent from this surface.** `handleAgentPrimaryAction` (`desktop/src/features/profile/ui/useAgentLifecycleActions.ts:30`) branches on `isManagedAgentActive`, which is permanently true, so the `startManagedAgentWithRules` arm is unreachable.

## Evidence: the agent reported correctly, the desktop ignored it

Agent-side log at shutdown:

```
13:38:41.122Z  info  draining              reason=owner_shutdown
13:38:43.518Z  warn  relay connection closed
```

Process gone. I then reconnected to the relay with the agent key (NIP-42 + NIP-OA auth tag) and queried its own events:

```
=== kind 10100 (roster) ===
created_at: 1785850721  → 2026-08-04T13:38:41Z   (exactly the drain instant)
content: {"name":"…","agent_type":"agent","status":"offline","capabilities":[],
          "channels":[],"channel_ids":["650ac09e-…","81f2eb5e-…"]}

=== kind 20001 (presence) ===
(none — the relay holds no presence for this pubkey)
```

Both axes are correct on the relay:

- the roster entry was replaced with `"status":"offline"`;
- presence was cleared, which is exactly what `crates/buzz-relay/src/handlers/event.rs:823-830` does on a kind:20001 whose content is `"offline"`.

So the desktop had everything it needed. It just never looks.

## Root cause

**1. Remote status is a function of one persisted field.**

`desktop/src-tauri/src/managed_agents/runtime.rs:148-169`:

```rust
let (status, pid, log_path) = if record.backend != BackendKind::Local {
    // Two-axis status model for remote agents:
    //   …
    //   Live axis (relay presence, polled by frontend): online/away/offline.
    //   Shown as a PresenceDot next to the agent name. …
    // After !shutdown the agent goes offline (presence) but stays "deployed"
    // (infrastructure still exists). This is intentional …
    let status = if record.backend_agent_id.is_some() {
        "deployed".to_string()
    } else {
        "not_deployed".to_string()
    };
    (status, None, String::new())
} else { /* local: real pid probe */ };
```

The comment's two-axis model is a reasonable design. The bug is that **the frontend never implements the second axis** for the decisions that matter — it treats the control-plane axis as liveness.

**2. `backend_agent_id` is write-once.** Set at `desktop/src-tauri/src/commands/agents.rs:501` on a successful deploy; the only paths that clear it are record creation defaults and deletion. There is no `undeploy` (deferred to "v2" per the comment at `runtime.rs:163`), and `stop_managed_agent` explicitly refuses remote agents (`desktop/src-tauri/src/commands/agents.rs:1244-1248`). So `status` is pinned at `"deployed"` for the entire life of the record.

**3. Every liveness-shaped UI decision keys off it.**

| Surface | Location | Behaviour |
|---|---|---|
| Active predicate | `desktop/src/features/agents/lib/managedAgentControlActions.ts:34` | `status === "running" \|\| "deployed"` → permanently `true` |
| Primary action label | `.../managedAgentControlActions.ts:38-47` | provider + active → `"Shutdown"`, never `"Deploy"` |
| Button "live" styling | `desktop/src/features/profile/ui/UserProfilePanelSections.tsx:364-367` | `agentActionLive` = same predicate, inlined |
| **Green online dot** | `desktop/src/features/agents/ui/AgentRuntimeAvatarControl.tsx:130` | `<PresenceDot status="online" />` — **hardcoded literal**, gated only on `isActive`. Presence is never read here |
| Status badge | `desktop/src/features/agents/ui/AgentStatusBadge.tsx:27-31` | green `deployed`; the `"Starting…"` de-escalation branch tests `status === "running"`, so it can never fire for a remote agent |
| Runtime tab dot | `desktop/src/features/profile/ui/UserProfilePanelSections.tsx:129-148` | `deployed` → `"running"` → `bg-emerald-500` |

`ManagedAgentRow.tsx:230` and `ProfileHero` (`UserProfilePanelSections.tsx:485-497`) *do* render a real presence-driven `PresenceDot` — so the app disagrees with itself on the same screen once the two axes diverge.

## Contributing (secondary) issue: nothing is refetched after `!shutdown`

`handleStop` (`desktop/src/features/agents/ui/useManagedAgentActions.ts:240-267`) invalidates and refetches **nothing** for the provider branch, unlike `handleStart` (`:220-221`). The same is true of `useMembersSidebarActions.ts:173-188` and `useAgentLifecycleActions.ts:30-41`.

Meanwhile `useRelayAgentsQuery` polls kind:10100 on a **5-minute** interval (`desktop/src/features/agents/hooks.ts:337`), and `usePresenceQuery` backstop-polls at 60s (`desktop/src/features/presence/hooks.ts:94`).

So even the surfaces that *do* honour presence lag by up to a minute, and roster-derived status by up to five — right at the moment the user is watching for feedback on an action they just took. Fixing the root cause without also invalidating `["presence"]` / `relay-agents` / `managed-agents` after a successful `!shutdown` send would leave a visible multi-minute lie.

## Why this is a spec violation, not a design choice

`docs/remote-agents.md:200-206`, invariant **I3**:

> **(I3) Presence is the status.** `D` derives a remote agent's live state *exclusively* from relay presence events self-signed by the agent key: `online`/`away`/`offline` (kind:20001, ephemeral, WS-published). The deployment axis (`deployed`/`not_deployed`, from the stored `backend_agent_id`) **is bookkeeping, not liveness.**

The implementation inverts this for every decision listed in the table above. This is not in §Known Defects (`docs/remote-agents.md:1572-1681`, defects 1–8).

I3 also promises a *bounded* wrong dot (180s `PRESENCE_TTL_SECS`). Today the wrong dot is unbounded: it survives shutdown, app restart, and machine reboot, because it is not a presence dot at all.

## Suggested fix

Smallest correct change is to give remote agents the live axis the comment already describes, and let it drive the "is this thing alive" decisions:

1. **Introduce a live signal for remote agents.** Either
   (a) have `build_managed_agent_summary()` fold relay presence into the returned status for `backend != Local`, or
   (b) keep the Rust summary purely control-plane and add a frontend helper — e.g. `isManagedAgentLive(agent, presenceLookup)` — that returns `presence !== "offline" && presence !== undefined` for provider agents and falls back to `status` for local ones.
   (b) is less invasive and keeps `deployed`/`not_deployed` honest as bookkeeping, per I3.
2. **Route the liveness-shaped call sites through it**: `getManagedAgentPrimaryActionLabel`, `agentActionLive`, `AgentRuntimeAvatarControl` (drop the hardcoded `status="online"` and pass the real presence through), `AgentStatusBadge`, `resolveRuntimeTabStatus`. Keep `isManagedAgentActive` for the genuinely control-plane questions (e.g. "should Delete warn about orphaning a remote deployment?" in `deleteManagedAgentWithRules`, which already reads `presenceLookup` and gets this right).
3. **Invalidate after a `!shutdown` send** — `["presence"]`, `relay-agents`, `managed-agents` — in `stopManagedAgentWithRules`' callers, so the transition is visible immediately rather than on the next 5-minute poll.
4. Once (1)–(3) land, an offline remote agent's button becomes **Deploy**, and `startManagedAgentWithRules` → `deploy` reaches the provider, which is already a converge-to-one-live-instance operation (§Deploy State Machine) and safely re-adopts or recreates. That closes the "cannot bring it back" half of this report without needing the v2 `undeploy`.

## Related

- #2798 — provider-backed config saves don't redeploy. Independently notes that v1 has no `shutdown` operation and that Shutdown is not a redeploy path; different defect (stale config, not stale status).
- #4605 — no management surface on the owner's other desktops; cites the same missing-`undeploy` deferral, but for orphaned deployments elsewhere.
- #3831, #2453 — same symptom class (UI says online, agent unusable) for *local* agents dying accidentally; different root cause.
- #4537 (PR) — presence reporting for relay-discovered agents; opposite direction (always offline), different files.
- #3034 — provider-backed agents display local config instead of the remote runtime; adjacent "remote agent card shows the wrong thing" family.

dukedorje:
Fixes #4730.

## The problem

A provider-backed agent stays **online** with a live **Shutdown** button forever after it has shut down — and because the primary action never flips back to **Deploy**, there is no way to bring it back from that surface.

The agent is not at fault: it publishes `kind:10100` with `"status":"offline"` and `kind:20001` presence `offline`, and the relay applies both. The desktop just never looks. `build_managed_agent_summary()` derives remote status exclusively from `record.backend_agent_id.is_some()`, which is write-once — there is no `undeploy` in v1 — so `status` is pinned at `"deployed"` for the life of the record, and every liveness-shaped UI decision keys off it.

This contradicts invariant **I3** ("Presence is the status", `docs/remote-agents.md`): the deployment axis "is bookkeeping, not liveness". I3 also promises a *bounded* wrong dot (180s `PRESENCE_TTL_SECS`); today's wrong dot survives shutdown, app restart, and reboot, because it is not a presence dot at all.

## The fix

Follows suggestion **(b)** from the issue — keep the Rust summary purely control-plane and add a frontend liveness helper — so `deployed`/`not_deployed` stay honest as bookkeeping, per I3.

`isManagedAgentLive(agent, presence)` is now the "is this thing alive" predicate. `isManagedAgentActive` keeps its name and its meaning, and is reserved for genuinely control-plane questions — e.g. the orphan warning in `deleteManagedAgentWithRules`, which already read `presenceLookup` and got this right.

Routed through it:

| Surface | Was |
|---|---|
| `getManagedAgentPrimaryActionLabel` | provider + active → `"Shutdown"`, never `"Deploy"` |
| `handleAgentPrimaryAction` | branched on `isManagedAgentActive` → the `startManagedAgentWithRules` arm was unreachable |
| `agentActionLive` (profile) | inlined `status === "running" \|\| "deployed"` |
| `resolveRuntimeTabStatus` | `deployed` → `"running"` → green |
| `AgentStatusBadge` | de-escalation branch tested `status === "running"`, so it could never fire for a remote agent |
| Members sidebar badge / icon / action | same predicate |

Once the label is honest, **Deploy** reaches `startManagedAgentWithRules` → `deploy`, which is already converge-to-at-most-one-live-instance (§Deploy State Machine) and safely re-adopts or recreates. That closes the "cannot bring it back" half without needing the v2 `undeploy`.

### On the loading window

`get_presence` omits offline/unknown pubkeys (noted in `presence.ts`), so an absent entry is indistinguishable from "the query hasn't resolved". A bare `presence !== undefined` test would make every remote agent flash **Deploy** on app start.

So `ManagedAgentPresence` carries both axes — `{ status, loaded }` — and liveness falls back to the control-plane axis until presence resolves. That keeps I3's promise of a *bounded* wrong signal instead of trading one unbounded lie for another. `AgentStatusBadge`'s existing 15s grace period is reused for the same reason.

Local agents are untouched: their status comes from a real pid probe, so the control-plane axis *is* liveness for them. `isManagedAgentLive` returns `isManagedAgentActive` unchanged for `backend.type !== "provider"`.

### Secondary: nothing was refetched after `!shutdown`

`handleStop` invalidated nothing on the provider branch (unlike `handleStart`), and the same held for `useMembersSidebarActions`. With `useRelayAgentsQuery` polling at 5 minutes and `usePresenceQuery` backstopping at 60s, fixing the root cause alone would still have left a visible multi-minute lie right when the user is watching for feedback on an action they just took. Both paths now invalidate `["presence"]`, `relay-agents`, and `managed-agents`. (The live transition itself still arrives over the `kind:20001` WS subscription; these keep the other two axes from contradicting it.)

### One structural note

`useAgentLifecycleActions` now owns its presence subscription instead of taking it as a prop, so every caller gets the presence-aware branch rather than each having to remember to pass it. react-query dedupes it against the panel's existing query. This also keeps `UserProfilePanel.tsx` at its 1000-line ratchet — the two presence props on `ProfileSummaryView` were merged into one `presence` object for the same reason.

## Testing

- `pnpm test` — **4488 pass, 0 fail**, including 8 new cases in `managedAgentControlActions.test.mjs` covering: absent-vs-loaded presence, online/away, the never-deployed case, local-agent fallback, the `Deploy`/`Shutdown` label flip, and unchanged local labels.
- `npx tsc --noEmit` — clean.
- `pnpm check` — clean (biome + file-size ratchet + px-text + pubkey-truncation).

Manual: deploy a provider-backed agent, `!shutdown`, confirm the dot goes grey, the badge de-escalates, the runtime tab reads stopped, and the button becomes **Deploy** — then press it and confirm the agent comes back under the same identity.

## Related

Searched open PRs; none found addressing this. #2798 (provider config saves don't redeploy) and #4605 (no management surface on other desktops) both cite the missing `undeploy`, but are different defects. #4537 is presence for relay-discovered agents — opposite direction, different files.

### DIFF
--- a/desktop/src/features/agents/lib/managedAgentControlActions.test.mjs
+++ b/desktop/src/features/agents/lib/managedAgentControlActions.test.mjs
@@ -2,6 +2,9 @@ import assert from "node:assert/strict";
 import test from "node:test";
 
 import {
+  getManagedAgentPrimaryActionLabel,
+  isManagedAgentLive,
+  managedAgentPresence,
   startManagedAgentWithRules,
   respawnManagedAgentWithRules,
 } from "./managedAgentControlActions.ts";
@@ -164,5 +167,98 @@ test("test_respawn_onStopped_fires_before_start_resolves", async () => {
     events,
     ["stop", "onStopped", "start"],
     "onStopped must fire after stop resolves and before start is called",
+  );
+});
+
+// ── Remote-agent liveness (I3: "Presence is the status") ─────────────────────
+// Regression coverage for #4730: a provider-backed agent that has shut down kept reading as
+// live because status is derived from backend_agent_id, which v1 never clears.
+
+const remote = (overrides = {}) =>
+  agent({
+    backend: { type: "provider", id: "mjolnir" },
+    backendAgentId: "vm-1234",
+    status: "deployed",
+    ...overrides,
+  });
+
+const presenceOf = (status, loaded = true) => ({ status, loaded });
+
+test("remote agent with no presence reads as not live once presence has loaded", () => {
+  // get_presence omits offline pubkeys, so "shut down" arrives as an absent entry.
+  assert.equal(isManagedAgentLive(remote(), presenceOf(undefined)), false);
+  assert.equal(isManagedAgentLive(remote(), presenceOf("offline")), false);
+});
+
+test("remote agent that is online or away reads as live", () => {
+  assert.equal(isManagedAgentLive(remote(), presenceOf("online")), true);
+  assert.equal(isManagedAgentLive(remote(), presenceOf("away")), true);
+});
+
+test("remote agent does not flash dead while presence is still loading", () => {
+  // Falling back to the control-plane axis here keeps I3's promise of a *bounded* wrong
+  // signal instead of trading one unbounded lie for another.
+  assert.equal(
+    isManagedAgentLive(remote(), presenceOf(undefined, false)),
+    true,
+  );
+});
+
+test("remote agent that was never deployed is not live regardless of presence", () => {
+  const undeployed = remote({ backendAgentId: null, status: "not_deployed" });
+  assert.equal(isManagedAgentLive(undeployed, presenceOf("online")), false);
+});
+
+test("local agents keep using the pid-probed status, not presence", () => {
+  const running = agent({ status: "running" });
+  // A local agent mid-start may have no presence yet; it is still running.
+  assert.equal(isManagedAgentLive(running, presenceOf(undefined)), true);
+  assert.equal(
+    isManagedAgentLive(agent({ status: "stopped" }), presenceOf("online")),
+    false,
+  );
+});
+
+test("shut-down remote agent offers Deploy, not Shutdown", () => {
+  // The bug: this returned "Shutdown" forever, making the deploy arm unreachable.
+  assert.equal(
+    getManagedAgentPrimaryActionLabel(remote(), presenceOf(undefined)),
+    "Deploy",
+  );
+  assert.equal(
+    getManagedAgentPrimaryActionLabel(remote(), presenceOf("online")),
+    "Shutdown",
+  );
+});
+
+test("local agent action labels are unchanged", () => {
+  const p = presenceOf(undefined);
+  assert.equal(
+    getManagedAgentPrimaryActionLabel(agent({ status: "running" }), p),
+    "Stop",
+  );
+  assert.equal(
+    getManagedAgentPrimaryActionLabel(agent({ status: "stopped" }), p),
+    "Restart Agent",
+  );
+  assert.equal(
+    getManagedAgentPrimaryActionLabel(agent({ status: "not_deployed" }), p),
+    "Start Agent",
+  );
+});
+
+test("managedAgentPresence distinguishes an unloaded lookup from an absent entry", () => {
+  const a = remote();
+  assert.deepEqual(managedAgentPresence(a, undefined), {
+    status: undefined,
+    loaded: false,
+  });
+  assert.deepEqual(managedAgentPresence(a, {}), {
+    status: undefined,
+    loaded: true,
+  });
+  assert.deepEqual(
+    managedAgentPresence(a, { [a.pubkey.toLowerCase()]: "online" }),
+    { status: "online", loaded: true },
   );
 });
--- a/desktop/src/features/agents/lib/managedAgentControlActions.ts
+++ b/desktop/src/features/agents/lib/managedAgentControlActions.ts
@@ -3,6 +3,7 @@ import type {
   Channel,
   ManagedAgent,
   PresenceLookup,
+  PresenceStatus,
   RelayAgent,
 } from "@/shared/api/types";
 import { normalizePubkey } from "@/shared/lib/pubkey";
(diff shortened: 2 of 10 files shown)


dukedorje:
Correcting my own "Related" section — I said "none found", and on a closer read that isn't quite right for one PR.

**#3449 (`buzz-backend-ssh`, open)** touches `managedAgentControlActions.ts` and reaches the same diagnosis independently. Its commit 8 adds `managedAgentPresenceStatus(agent, presenceLookup)` with a comment that reads almost like this PR's: `backend_agent_id` is write-once, the protocol has no undeploy, so a remote agent that died hours ago is still `"deployed"` forever.

The two are complementary rather than duplicate:

- #3449 fixes the **displayed presence** — it returns a `PresenceStatus` for surfaces that render a dot, and leaves `getManagedAgentPrimaryActionLabel` on `isManagedAgentActive`.
- This PR fixes the **decisions** — the action label and `handleAgentPrimaryAction`, which is what makes the deploy arm reachable and closes the "no way to bring it back" half of #4730.

Only two files overlap (`managedAgentControlActions.ts` and its test), so whichever lands second is a small rebase. Happy to do that rebase in either direction, or to reshape this on top of their `managedAgentPresenceStatus` if you'd rather have one helper — just say which you prefer.

One design difference worth surfacing for whoever reconciles them, since it's a real tradeoff rather than an oversight. `get_presence` omits offline pubkeys, so an absent entry means both "offline" and "not loaded yet". #3449 resolves that to `"offline"` — correct for a dot, where silence honestly reads as "not known to be alive". This PR carries a `loaded` flag and falls back to the control-plane axis until presence resolves, because for a *button* the same resolution makes every remote agent flash **Deploy** on app start and invites a redundant deploy press. Both readings are right for their own surface; a merged helper would want to keep the distinction rather than pick one.

PR #5139, opened 2026-08-07
fix(cli): buzz dms list always returns empty

### LINKED ISSUE #5137: buzz dms list returns empty for an active DM conversation with messages
## Repro
1. Another user opens a DM with an agent and sends messages (confirmed via `buzz feed get --types mentions,needs_action`, which shows the DM channel's events tagged with `["h","<dm-channel-uuid>"]`).
2. Agent runs `buzz dms list` — returns `[]`, even though the DM channel has multiple messages and is clearly active.

## Expected
`buzz dms list` should surface any DM conversation the agent is a member of, matching what `buzz feed get` already shows.

## Impact
Agents that rely on `buzz dms list` to discover new DMs (rather than polling the full feed) will silently miss messages sent to them directly. In this case a user's DM sat unanswered for ~20 minutes because the agent had no signal it existed outside the general feed.

## Environment
Windows 11, buzz CLI (current), Nostr relay-backed Buzz platform.

deepmroot:
## Summary
- `buzz dms list` queried `kind:41001` (`KIND_DM_CREATED`), but the relay's `handle_dm_open` never emits that kind — it's a registered kind with no publisher anywhere in the relay, so the query always returned `[]`, even for active DMs with messages.
- DM channels are regular NIP-29 channels (`channel_type = "dm"`) that already emit discovery events: `kind:39002` membership and `kind:39000` metadata tagged `["t","dm"]` with a `p` tag per participant (see `emit_group_discovery_events` in `command_executor.rs`).
- Rewired `cmd_list_dms` to follow the same two-step discovery path as `channels list --member`: `kind:39002` (my membership) → `kind:39000` metadata for those channel ids → filter to `["t","dm"]`.

Fixes #5137.

## Test plan
- [x] `cargo build --release -p buzz-cli`
- [x] `cargo fmt -p buzz-cli -- --check`
- [x] `cargo clippy -p buzz-cli --all-targets -- -D warnings`
- [x] Ran `./target/release/buzz.exe dms list` against a live relay with an active DM that previously returned `[]` — now returns the DM with correct `dm_id`, `participants`, and `created_at`.

### DIFF
--- a/crates/buzz-cli/src/commands/dms.rs
+++ b/crates/buzz-cli/src/commands/dms.rs
@@ -1,40 +1,66 @@
 use uuid::Uuid;
 
-use crate::client::{extract_d_tag, normalize_write_response, BuzzClient};
+use crate::client::{extract_d_tag, extract_p_tags, normalize_write_response, BuzzClient};
 use crate::error::CliError;
 use crate::validate::{parse_uuid, sdk_err, validate_hex64};
 
-/// List DM conversations by querying kind:41001 (relay-confirmed DMs) filtered by our pubkey.
+/// List DM conversations.
+///
+/// DMs are NIP-29 channels with `channel_type = "dm"` — the relay never emits a
+/// kind:41001 confirmation event (that kind is reserved but unused), so this
+/// follows the same discovery path as `channels list --member`: find channel
+/// ids from our kind:39002 membership events, then fetch kind:39000 metadata
+/// for those ids and keep only the ones tagged `["t", "dm"]`.
 pub async fn cmd_list_dms(client: &BuzzClient, limit: Option<u32>) -> Result<(), CliError> {
     let my_pk = client.keys().public_key().to_hex();
     let limit = limit.unwrap_or(50).min(200);
-    let filter = serde_json::json!({
-        "kinds": [41001],
+
+    let member_filter = serde_json::json!({
+        "kinds": [39002],
         "#p": [my_pk],
-        "limit": limit
     });
-    let resp = client.query(&filter).await?;
-    let events: Vec<serde_json::Value> = serde_json::from_str(&resp).unwrap_or_default();
-    let dms: Vec<serde_json::Value> = events
+    let member_events = client.query_paginated(member_filter, limit).await?;
+    let channel_ids: Vec<String> = member_events
         .iter()
+        .map(extract_d_tag)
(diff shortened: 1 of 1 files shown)


matthewdonsemail-lab:
@deepmroot make sure that it's formatted in the right way ask my agents in the buzz groupchat for InferenceSaver that it uses the Github Issues skill communicate with Inference Ops it'll tell you about the skill

BradGroux:
Additional portfolio review context (2026-08-09)

We independently reviewed this alongside #5140 and our overlapping #5247 against current `main` at `5bf78671f45178f8de02ba18d3d321cbbf19cd1f`. The root-cause analysis is correct: `KIND_DM_CREATED` exists as a registered constant, but the relay's DM-open path does not publish kind:41001. The authoritative discovery path is the relay-emitted NIP-29 membership and metadata pair, kind:39002 followed by kind:39000. That explains both the empty-list behavior in #5137 and why this implementation's live-relay verification succeeds.

We closed #5247 in favor of this PR after comparing all three implementations. #5139 is the strongest vehicle because it was opened first, includes direct live-relay proof, reuses the shared `extract_p_tags` parser, preserves the existing output shape, and adds deterministic newest-metadata-first ordering. The current head is mergeable and DCO, Semgrep, and zizmor are green.

One limit-semantics edge case is worth addressing before merge. `query_paginated(member_filter, limit)` applies the user limit to all channel membership events before the code knows which channels are DMs. The metadata query is also capped before the post-query `t == "dm"` filter. A user with at least `limit` newer non-DM memberships can therefore receive fewer than the requested number of DMs, including `[]`, even when an older DM membership exists. In the current implementation, `--limit 50` means “inspect at most 50 membership candidates and 50 metadata candidates,” not “return up to 50 DMs.”

The simplest correctness-first adjustment is to collect and deduplicate the caller's membership channel IDs, fetch the matching metadata, filter to DMs, sort, and only then truncate the final DM list to `limit`. If querying every membership is considered too broad, the alternative is to page membership candidates until enough DM metadata has been found or the membership cursor is exhausted.

A focused regression should mix more non-DM memberships than the requested limit with an older DM and assert that the DM is still returned. A second assertion should cover deterministic ordering after final filtering. With that boundary fixed, this PR fully owns #5137 and #5140 can be closed as the remaining duplicate.

PR #5142, opened 2026-08-07
fix(desktop): fence localStorage SecurityError from killing the React tree

### LINKED ISSUE #5078: A throwing localStorage.getItem kills the whole desktop tree (no ErrorBoundary anywhere in desktop/src)
## Summary

A `localStorage.getItem` that throws takes down the entire desktop React tree. There is no `ErrorBoundary` anywhere in `desktop/src`, so the throw reaches the root and the app renders nothing.

WebKit throws `SecurityError` from `getItem` (not just `setItem`) when storage access is denied for the origin. `desktop/src` has ~80 `window.localStorage.getItem` call sites, and the ones that run during provider initialization are reached before any UI exists.

This is pre-existing behavior on `main`. It was found while measuring throw-resilience for #4900, and explicitly ruled **not** that PR's defect because it reproduces with that PR's feature disabled.

## Measured

Rig: the mounted-hook test harness plus a sibling component in the same tree, with `window.localStorage.getItem` patched to throw `SecurityError` for selected keys. Each arm run in its own process (batched runs share module state and flip answers). Liveness read as `container.childNodes.length`.

| arm | throwing key | result |
|---|---|---|
| control | none | no throw, sibling renders, `containerChildren=1` |
| 1c | all `buzz-` keys | `SecurityError` escapes render |
| 1d | `buzz-communities` only | `SecurityError` escapes render |
| 1d | `buzz-active-community-id` only | `SecurityError` escapes render |
| 1e | (join-alert key) | `containerChildren=0` — tree dead |
| 1e-control | none | `containerChildren=1` — detector proven sensitive |

Arms 1c and 1d ran with the #4900 feature **disabled**, which is what establishes this as a `main` issue rather than a feature issue: `buzz-communities` and `buzz-active-community-id` are read by the community providers during startup.

The `1e-control` row matters — `containerChildren=0` is only evidence of a dead tree because the same detector reads `1` on a healthy mount. Without it, `0` could have been the DOM shim never appending children.

## Two candidate fixes

1. **An `ErrorBoundary` at the app root.** Turns any storage throw (and every other render throw) into a recoverable error surface instead of a white window. Broadest coverage, does not require auditing call sites.
2. **A guarded storage accessor** — a `safeGetItem` helper returning `null` on throw — applied to the ~80 sites. Narrower, but addresses the cause rather than the symptom.

These are complementary, not alternatives. (1) is the smaller change with the larger blast-radius reduction.

Worth noting for whoever picks this up: `readJoinAlertLedger` in `desktop/src/features/community-members/lib/joinAlerts.ts` already wraps `JSON.parse` in `try`/`catch`, but the `getItem` call sits *above* that `try` — a shape likely repeated at other sites that look defensive at a glance.

## Not doing this in #4900

Fixing one of ~80 call sites in a feature PR would leave the class open while looking closed. Sizing the fix against all of them belongs on `main`.

## Acceptance (ported from the duplicate #5079, filed independently by Max)

- A `SecurityError` from storage during startup does not empty the root tree.
- The user sees a recoverable state instead of a blank/dead app.
- Normal storage reads and existing persisted state remain unchanged.

iroiro147:
## Summary

WebKit throws `SecurityError` from `localStorage.getItem` (not just `setItem`) when storage access is denied for the origin. With no `ErrorBoundary` in `desktop/src`, any such throw inside a provider render (`ThemeProvider`, `CommunitiesProvider`, `App` boot) propagated to the reconciler, unmounted the root, and left a blank window. Measured repro in #5078: a single throwing `getItem` on `buzz-communities` or `buzz-active-community-id` kills the container.

Closes #5078.

## What changed

**New helper — `desktop/src/shared/lib/safeStorage.ts`**
- `getStorageItem(key, fallback?)` — wraps `window.localStorage.getItem`; on a thrown error (SecurityError under denied-storage origin) it warns once per key and returns the fallback.
- `setStorageItem(key, value)` and `removeStorageItem(key)` — same fail-closed contract (return `false` on throw).
- Unit tests in `safeStorage.test.mjs` cover the happy path and the `SecurityError` path.

**Rewired the init-path readers that ran before any UI existed**
- `desktop/src/features/communities/communityStorage.ts` — `migrateLegacyCommunityStorage`, `loadCommunities`, `loadActiveCommunityId`, `loadCommunityDiscoveryAfterLeave`, `initFirstCommunity`
- `desktop/src/features/communities/legacyCommunityStorage.ts` — `migrateLegacyCommunityStorageBeforeRender`
- `desktop/src/shared/theme/ThemeProvider.tsx` — `readStoredTheme`, `applyCachedVars`, the `useState` initialisers for `accentColor` and `followSystem`, and the accent re-read inside `applyTheme`

**Root-level fence — `desktop/src/app/RootErrorBoundary.tsx`**
- New top-level `ErrorBoundary` wrapping the whole provider tree in `main.tsx`. Any remaining uncaught render error (a future storage read that bypasses the helper, or any other render-time crash) renders a degraded splash with a Reload button instead of a blank window.

## Test plan

- `desktop/src/shared/lib/safeStorage.test.mjs` — node `--test` runner, 11 assertions across healthy, absent, and SecurityError-throwing storage.
- Full `just ci` runs on the blocker.
- Existing `communityStorage.test.mjs` and `legacyCommunityStorage.test.mjs` continue to pass (they exercise the same functions via in-memory Storage doubles; the new code path in `migrateLegacyCommunityStorage` only adds a `try/catch` around the same body).

## Why not an ErrorBoundary-only fix

A boundary alone can't help on a *clean* mount — the first throw already unmounted the whole subtree before any state or fallback data was loaded, so retrying would hit the same throw on the very next render. The storage accessor has to fail closed *and* the boundary has to exist for whatever bypasses it. Both are needed; neither is sufficient alone.

### DIFF
--- a/desktop/src/app/RootErrorBoundary.test.mjs
+++ b/desktop/src/app/RootErrorBoundary.test.mjs
@@ -1,0 +1,84 @@
+import assert from "node:assert/strict";
+import { after, afterEach, before, test } from "node:test";
+
+import { JSDOM } from "jsdom";
+
+const dom = new JSDOM("<!doctype html><html><body></body></html>", {
+  url: "http://localhost",
+});
+const originalConsoleError = console.error;
+
+before(() => {
+  Object.assign(globalThis, {
+    document: dom.window.document,
+    HTMLElement: dom.window.HTMLElement,
+    IS_REACT_ACT_ENVIRONMENT: true,
+    window: dom.window,
+  });
+  dom.window.matchMedia = () => ({
+    matches: false,
+    addEventListener() {},
+    removeEventListener() {},
+  });
+});
+
+afterEach(async () => {
+  const { cleanup } = await import("@testing-library/react");
+  cleanup();
+  console.error = originalConsoleError;
+});
+
+after(() => dom.window.close());
+
+test("ThemeProvider renders defaults when localStorage reads are denied", async () => {
+  const deniedStorage = {
+    getItem() {
+      throw new dom.window.DOMException("private diagnostic", "SecurityError");
+    },
+    setItem() {
+      throw new dom.window.DOMException("private diagnostic", "SecurityError");
+    },
+    removeItem() {
+      throw new dom.window.DOMException("private diagnostic", "SecurityError");
+    },
+  };
+  Object.defineProperty(dom.window, "localStorage", {
+    configurable: true,
+    value: deniedStorage,
+  });
+
+  const { createElement } = await import("react");
+  const { render, screen } = await import("@testing-library/react");
+  const { ThemeProvider } = await import("@/shared/theme/ThemeProvider");
+
+  render(
+    createElement(
+      ThemeProvider,
+      { defaultTheme: "buzz" },
+      createElement("p", null, "Buzz is visible"),
+    ),
+  );
+
+  assert.ok(screen.getByText("Buzz is visible"));
+});
+
+test("root boundary shows recovery UI without exposing error details", async () => {
+  const diagnostic = "/Users/alice/private/session-token";
+  console.error = () => {};
+
+  const { createElement } = await import("react");
+  const { render, screen } = await import("@testing-library/react");
+  const { RootErrorBoundary } = await import("./RootErrorBoundary.tsx");
+  function ThrowingProvider() {
+    throw new Error(diagnostic);
+  }
+
+  render(
+    createElement(RootErrorBoundary, null, createElement(ThrowingProvider)),
+  );
+
+  assert.ok(screen.getByText("Buzz failed to start"));
+  assert.ok(screen.getByRole("button", { name: "Reload" }));
+  assert.equal(document.body.textContent.includes(diagnostic), false);
+  assert.match(document.body.textContent, /contact support/i);
+});
--- a/desktop/src/app/RootErrorBoundary.tsx
+++ b/desktop/src/app/RootErrorBoundary.tsx
@@ -40,7 +40,8 @@ export class RootErrorBoundary extends Component<
         <div className="flex h-screen w-screen flex-col items-center justify-center gap-3 bg-background px-6 text-foreground">
           <p className="text-base font-semibold">Buzz failed to start</p>
           <p className="max-w-md text-center text-sm text-muted-foreground">
-            {error.message}
+            Reload Buzz to try again. If this keeps happening, check that Buzz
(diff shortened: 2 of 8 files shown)


wesbillman (changes requested):
Carl reviewing on behalf of Wes at exact head `4d832212147d44c3b77cddf4b15333ae1261f4d6` as a release/security gate.

The production approach is directionally sound: the diff is confined to the seven advertised desktop files; I found no dependency, workflow, native-command, network, credential, or code-loading changes; the guarded startup reads preserve healthy-storage values and fail closed under denied storage. I am requesting changes before release for the following blockers:

1. **The newly added tests do not run in the repository's actual test environment.** `pnpm test` fails all six `safeStorage.test.mjs` cases with `ReferenceError: window is not defined` at `patchLocalStorage` (`safeStorage.test.mjs:31`). Result: 4480 pass / 6 fail. These tests need to install/restore a `globalThis.window` (using the repo's established DOM-test pattern) rather than assuming the Node runner supplies one.

2. **The acceptance behavior is not tested end to end.** The helper tests—even once repaired—only call three wrappers directly. They do not prove that a denied `localStorage.getItem` during `CommunitiesProvider`/`ThemeProvider` initialization leaves a visible tree, nor that the root boundary renders a recoverable surface for a remaining render-time throw. Please add a deterministic mounted regression that injects a throwing storage implementation before mount and asserts a non-empty, actionable fallback/default UI. This is the core release guarantee, not decorative shrubbery.

3. **The root fallback renders arbitrary `error.message` text to the user** (`RootErrorBoundary.tsx:40-42`). Because this boundary catches every render error, those messages can contain internal paths, request details, identifiers, or other diagnostic data. Keep the UI generic and log the diagnostic only via `componentDidCatch`.

Additionally, `pnpm check` reports a formatter error introduced in `main.tsx:84-99`; the other reported lint items pre-exist outside this diff. `pnpm typecheck` and `git diff --check` pass.

I also traced bootstrap before the boundary: the production storage paths currently invoked there (`recoverLocalStorageQuotaOnStartup` and legacy migration/application) are locally caught, so I did not find a separate pre-boundary production escape in this diff. Fix the failing tests, prove the actual mounted denial/fallback behavior, avoid exposing raw errors, and format the file; then this should be straightforward to reassess.

wesbillman (approved):

PR #4374, opened 2026-08-02
fix(desktop): stop clipping focus ring on channel intro action cards (#2392)

iroiro147:
## What

Fixes #2392 — the action cards in the empty-channel intro ("Create agent", "Add people") had their `focus-visible` ring clipped by the surrounding scroll container.

## Root cause

The cards sit in a `flex ... overflow-x-auto pb-1` row. Setting `overflow-x` (without `overflow-y`) makes the browser compute `overflow-y: auto` as well, so the container clips anything painted outside its padding box — including the cards' `focus-visible:ring-2` box-shadow. With only `pb-1` padding, the top/left/right of the ring were cut off when Tabbing to a card.

## Change

`desktop/src/features/messages/ui/ChannelIntroBlock.tsx` — `pb-1` → `p-1` on the action-cards scroll container, reserving 4px on all four sides so the focus ring renders fully inside the scroll container's padding box.

- 1 file, 1 line. No behavior change for mouse users or layout.

## Verification

- `pnpm typecheck` — clean
- `pnpm exec biome check src/features/messages/ui/ChannelIntroBlock.tsx` — clean
- `pnpm check:file-sizes` — clean
- Desktop unit suite — **3906/3906 pass**

Signed-off-by: Sarthak Singh <sarthak.singh@juspay.in>

PR #5143, opened 2026-08-07
fix(desktop): drop unhandled rejection from throwing window.Notification

### LINKED ISSUE #5081: A throwing window.Notification constructor becomes an unhandled rejection at every sendDesktopNotification call site
## Summary

`sendDesktopNotification` ends in an unguarded `new window.Notification(...)` (`desktop/src/features/notifications/lib/desktop.ts:383`). Every caller discards the returned promise without a rejection handler, so a throwing notification constructor becomes an unhandled promise rejection: the notification is silently dropped and the only trace is console noise.

This is pre-existing on `main`. Found while measuring throw-resilience for #4900 and ruled non-blocking for that PR — the blast radius is one dropped notification, and the next notification delivers normally.

## Measured

Rig: the mounted-hook harness with `window.Notification` replaced mid-run by a constructor that throws, then restored to drive a second notification.

```
ARM2_RESULT syncThrow=none unhandledRejections=1 rejectionName=NotificationError
            siblingRenders=1 alertsAfterRecovery=1
```

So: no render crash (the throw is on an async path, not in render), one unhandled rejection, the tree stays alive, and recovery is automatic — the next notification delivers.

The realistic trigger is a notification backend that becomes unavailable after permission was granted.

## Affected call sites (all on `main`)

| site | shape | exposed |
|---|---|---|
| `app/useAppShellDesktopNotifications.ts:79` | `void sendDesktopNotification(...).then(...)` | yes — `.then` with no `.catch` |
| `app/useAppShellDesktopNotifications.ts:127` | `void sendDesktopNotification(...).then(...)` | yes — same |
| `features/reminders/useReminderNotifications.ts:108` | `void sendDesktopNotification(...).then(...)` | yes — same |
| `features/notifications/use-feed-desktop-notifications.ts:116` | `await` inside an async `useEffectEvent` | yes — the event is invoked as `void deliverFeedNotification(...)` at `:220`, so the rejection escapes there instead |

All four are exposed. The fourth is worth calling out because it *looks* different — it uses `await` rather than a floating `.then` — but the `await` sits inside an async function that is itself called with a bare `void`, so the rejection surfaces identically one level up. A `.then`-vs-`await` audit that stops at the call to `sendDesktopNotification` would classify it as safe.

`#4900` adds a fifth caller with the same shape; it does not introduce the pattern, and `lib/desktop.ts` is untouched by it (`git diff origin/main...2b5efb94 -- desktop/src/features/notifications/lib/desktop.ts` is empty).

## Suggested fix

Guard inside `sendDesktopNotification` rather than at four (soon five) call sites: wrap the constructor and return `false` on throw. That matches the function's existing contract — it already returns `false` for "did not send" when permission is not granted, so callers that check `didSend` need no change, and the failure stops being invisible.

A `.catch` at each call site also works but leaves the next caller to remember.

## Testing note

The regression test needs to assert on the *rejection*, not on the notification count — a dropped notification and a never-attempted notification produce the same count. Registering a `process.on("unhandledRejection", ...)` handler is what distinguishes them.

## Acceptance (ported from the duplicate #5080, filed independently by Max)

- A throwing Notification constructor causes no `unhandledrejection` event.
- The failed notification does not poison later notification delivery.
- Existing successful notification behavior is unchanged.

iroiro147:
## Summary

WebKit throws `NotificationError` from the `Notification` constructor when the notification backend becomes temporarily unavailable (measured repro attached to #5081). Every existing call site used `void sendDesktopNotification(...).then(...)` — discarding the returned promise with no rejection handler — so a throwing constructor became an unhandled promise rejection. The notification was silently dropped and the only trace was console noise.

Closes #5081.

## What changed

Fenced the throw at the source inside `sendDesktopNotification` (`desktop/src/features/notifications/lib/desktop.ts`):

- A new `try { ... } catch { ... }` wraps `new window.Notification(...)` and the `onclick` attach.
- On catch, we `console.warn` once and `return false`, so the promise the call sites discard is always fulfilled with the same boolean result. No caller needs to change.

## Why at the source and not at each call site

The issue body lists four rejecting edges: `useAppShellDesktopNotifications` (2×), `useReminderNotifications`, `use-feed-desktop-notifications`. Patching them one-by-one leaves the door open for the next consumer to make the same mistake — and the function itself advertises `Promise<boolean>`, so callers are entitled to assume the promise resolves with the delivery bit rather than rejects. Fixing the inside satisfies both properties for every present and future caller.

## Test plan

- Behavior change is a guarded return value around a single constructor; unit coverage is best expressed inside the mounted-hook harness already used in the repro. Existing notification helpers (`shouldNotify*.test.mjs`) continue to pass.
- Full `just ci` runs on the blocker.
- The next notification after a backend blip delivers normally (the throw is per-call, not sticky).

## Note on scope

This addresses the titled bug: unhandled rejection from a throwing constructor. A separate, intended follow-up is to wire a user-visible delivery-miss event if the platform exposes one — that's notification-observability work, not a } catch.

### DIFF
--- a/desktop/src/features/notifications/lib/desktop.test.mjs
+++ b/desktop/src/features/notifications/lib/desktop.test.mjs
@@ -1,0 +1,52 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+const notifications = [];
+
+class WorkingNotification {
+  static permission = "granted";
+
+  constructor(title, options) {
+    notifications.push({ title, options });
+  }
+
+  close() {}
+}
+
+class ThrowingNotification {
+  static permission = "granted";
+
+  constructor() {
+    throw new Error("notification backend unavailable");
+  }
+}
+
+globalThis.window = { Notification: ThrowingNotification };
+
+const { sendDesktopNotification } = await import("./desktop.ts");
+
+test("constructor failure is a delivery miss and does not prevent a later notification", async (t) => {
+  const warnings = [];
+  t.mock.method(console, "warn", (...args) => warnings.push(args));
+
+  const failed = await sendDesktopNotification({ title: "First" });
+
+  assert.equal(failed, false);
+  assert.equal(warnings.length, 1);
+  assert.match(String(warnings[0][1]), /notification backend unavailable/);
+
+  window.Notification = WorkingNotification;
+
+  const delivered = await sendDesktopNotification({
+    title: "Second",
+    body: "Recovered",
+  });
+
+  assert.equal(delivered, true);
+  assert.deepEqual(notifications, [
+    {
+      title: "Second",
+      options: { body: "Recovered", silent: true, extra: undefined },
+    },
+  ]);
+});
--- a/desktop/src/features/notifications/lib/desktop.ts
+++ b/desktop/src/features/notifications/lib/desktop.ts
@@ -385,7 +385,7 @@ export async function sendDesktopNotification(
   // when the notification backend becomes temporarily unavailable. Callers
   // discard the returned promise without a rejection handler, so an
   // un-guarded throw becomes an unhandled rejection. Treat constructor failure
-  // as a delivery miss (return false) and warn once so the console isn't noisy.
+  // as a delivery miss (return false) and log the failed delivery.
   try {
     const notification = new window.Notification(payload.title, {
       body: payload.body,
--- a/desktop/src/features/notifications/lib/desktop.ts
+++ b/desktop/src/features/notifications/lib/desktop.ts
@@ -381,19 +381,32 @@ export async function sendDesktopNotification(
     }
   }
 
-  const notification = new window.Notification(payload.title, {
-    body: payload.body,
-    silent: true,
-    extra: notificationExtra(payload.target),
-  } as DesktopNotificationOptions);
+  // block/buzz#5081 — WebKit throws `NotificationError` from the constructor
+  // when the notification backend becomes temporarily unavailable. Callers
+  // discard the returned promise without a rejection handler, so an
+  // un-guarded throw becomes an unhandled rejection. Treat constructor failure
+  // as a delivery miss (return false) and warn once so the console isn't noisy.
+  try {
+    const notification = new window.Notification(payload.title, {
+      body: payload.body,
+      silent: true,
+      extra: notificationExtra(payload.target),
+    } as DesktopNotificationOptions);
 
-  const target = payload.target;
-  if (!isTauri() && target) {
-    notification.onclick = () => {
-      dispatchDesktopNotificationTarget(target);
-      notification.close();
-    };
+    const target = payload.target;
+    if (!isTauri() && target) {
+      notification.onclick = () => {
+        dispatchDesktopNotificationTarget(target);
+        notification.close();
+      };
+    }
+
+    return true;
+  } catch (error) {
+    console.warn(
+      "[desktop] window.Notification constructor threw — notification dropped:",
+      error,
+    );
+    return false;
   }
-
-  return true;
 }


wesbillman (dismissed):
Reviewing on Wes's behalf.

The source-level catch is the right shape: a synchronous `Notification` constructor failure currently rejects this async function, and existing callers can let that rejection escape. Returning `false` preserves the existing delivery-result contract, and I found no unsafe or unrelated changes in the one-file diff.

I am requesting changes before release for two gaps:

1. Please add focused regression coverage for the behavior required by #5081. The test should make `window.Notification` throw, verify the returned promise fulfills with `false` (rather than rejecting), restore the constructor, and verify a later send still succeeds. This is the release-fix contract; the PR currently has no test that would fail if the catch were removed or accidentally made sticky.
2. The implementation and documentation repeatedly say it will "warn once," but the catch calls `console.warn` on every failed send. Please either implement true warn-once behavior (and test it) or correct the source comment and PR prose. As written, a persistent backend outage can warn for every attempted notification.

The patch is plausible and narrowly scoped, but without the regression test it is not yet release-ready.

wesbillman (commented):
Reviewed exact head `7bc8d8112471b32106215842e50b22d8f4fea19d` against base/current main `f53bbd1152464ecbb1de495e2d1d959e156138f0`.

No blocking correctness or security findings. The change is confined to `desktop/src/features/notifications/lib/desktop.ts`: it wraps the existing `Notification` construction/click-handler setup, converts a synchronous platform throw into the function's existing `false` delivery result, and introduces no dependency, privilege, network, persistence, workflow, or build changes. This fixes #5081 at the shared boundary and does not make failure sticky, so a later call can still deliver normally.

I also ran the full Desktop JS unit suite (4,480 passing), TypeScript typecheck, and Desktop check at the exact head; all commands exited successfully. The check command reports only pre-existing informational/warning diagnostics outside the touched file.

Non-blocking release caveat: this PR adds no discriminating regression test for the issue's acceptance cases (throw resolves `false` without an unhandled rejection; a subsequent successful constructor still delivers). The implementation is small and directly correct, but adding that test would improve confidence and prevent regression. Also, the comment/PR wording says "warn once"; the implementation warns once per failed invocation, not once per outage/session. That is harmless for correctness but should not be read as log deduplication.

wesbillman (commented):
Reviewing updated head `91ee00c9a035bed1432f76fcffd6bf3eed8f5677` on Wes's behalf. The requested release hardening is now present: the focused test proves a throwing `window.Notification` constructor fulfills with `false`, records the delivery failure, and does not poison a later successful send. The misleading “warn once” source claim is corrected to accurately describe per-delivery logging.

The exact pushed head passed the repository pre-push hook: branch-skew, Desktop check, TypeScript typecheck, and the full Desktop unit suite (4,481/4,481). I found no new blocking issue. This COMMENT is not an approval.

wesbillman (commented):
Release re-review of exact head `91ee00c9a035bed1432f76fcffd6bf3eed8f5677`: no blocking correctness, security, or supply-chain findings.

I independently inspected the complete base-to-head diff and Carl's follow-up commit. The branch changes only `desktop/src/features/notifications/lib/desktop.ts` and a new colocated unit test. It adds no dependency, script, workflow, native code, permission, network/file access, persistence, or privilege change. Commit authorship/trailers correctly identify the external contributor's original patch and Wes/Carl's follow-up.

This fixes the real failure measured in #5081. A synchronous `window.Notification` constructor throw currently rejects the async helper; several callers discard that promise, producing the reported unhandled rejection. Catching at the shared delivery boundary and returning `false` matches the helper's existing “not delivered” contract and leaves later sends unaffected.

Carl's test is discriminating: it awaits the throwing call and asserts fulfillment with `false` (which fails if the promise rejects), checks the warning, replaces the constructor, then verifies the next send returns `true` with the expected payload. I reran that focused test at the exact head and it passed. Carl's pre-push validation additionally reports Desktop check, TypeScript typecheck, branch-skew, and the full 4,481-test Desktop suite passing. The head merges cleanly with current `origin/main` (`c71f6585391421843b552150cef2d927966ee3ba`).

One documentation nit remains: the PR body and original commit message still say “warn once,” while the corrected source accurately logs once per failed delivery. That does not affect shipped behavior or release safety, but the PR description should be corrected when convenient.

wesbillman (approved):

iroiro147:
Hi Wes — thanks for merging #5142 and #5143 the other day.

I wanted to follow up honestly about something. I've opened a lot of PRs on buzz over the last couple of weeks — well over a hundred. That's more than is reasonable to put in front of one reviewer, and that's on me.

The reason there are so many is that I genuinely love what buzz is building. The multiplayer-AI direction, and the idea that humans and agents actually collaborate in a shared space rather than one just driving the other, is the most interesting thing I've seen in a while. I've built myself some humble tooling to help me find and work on issues, which lets me contribute to a project I care about at a pace I couldn't otherwise — but I let it run further than was fair to you, and I'd rather fix that than keep going.

So rather than leave a pile that's impossible to work through, I'd like you to get value out of a handful. Going by what you've actually merged, desktop fixes seem most useful, so these are the five I'd stand behind:

- #4295 — channel-intro action-card focus rings (same area as #4374, which you merged)
- #4105 — keep hover action bar & reaction pills from overlapping
- #4360 — log WebKitWebProcess termination instead of dying silently
- #4263 — accept uppercase/mixed-case nsec on key import
- #4290 — let the onboarding setup frame scroll past its min height

All five are mergeable, issue-linked, and small enough to review quickly.

I'm very happy to close the rest — say the word and I'll do it in one pass, no hard feelings. Or if you'd prefer a steady trickle instead, tell me a rate and I'll stick to it.

And genuinely: if there's an area you'd like help with — flaky tests, a migration, docs nobody wants to write — I'd much rather work on what you need than keep guessing. I'd like to be useful to buzz over the long term rather than noisy in the short term.

— Sarthak

PR #5146, opened 2026-08-07
test(conformance): supply d tag for workflow definitions after relay schema change

### LINKED ISSUE #5101: `conformance_multitenant` workflow row is stale vs the workflow-event schema (missing `d` tag → 400)
## Environment

- Relay: local build from `block/buzz` main (2026-07-24), `buzz-relay`.
- Test: `crates/buzz-test-client/tests/conformance_multitenant.rs`, module
  `workflows::workflow_trigger_is_community_confined`.
- Re-verified at `38bf642f` (main, 2026-08-05): the relay still rejects with
  exactly this string
  (`crates/buzz-relay/src/handlers/command_executor.rs:668`) and the test's
  workflow builder is unchanged, so the row still fails at setup.

## What happens

The test fails at **setup**, not at the isolation assertion:

```
thread 'workflows::workflow_trigger_is_community_confined' panicked at
crates/buzz-test-client/tests/conformance_multitenant.rs:1706:
POST /events to http://a.localhost:3100 returned HTTP 400 Bad Request:
{"error":"invalid: missing d tag (workflow_id)"}
```

The test tries to create a workflow event, but the relay rejects it as missing
a `d` tag (workflow_id). The workflow is never created, so the isolation
property it means to check (a workflow trigger is confined to its community) is
never exercised — this is a false red, not an isolation failure.

## Why

Version skew: the relay's workflow-event validation now requires a `d` tag
(workflow_id), but the conformance test's workflow-creation helper doesn't add
one.

## Fix

Update the test's workflow-event builder to include the `d` / workflow_id tag
matching the current relay schema, so the row exercises the community-confinement
assertion again. (The sibling row `workflows::approval_token_is_community_confined`
is already a `pending_lane` `todo!()` stub — separate work.)

iroiro147:
## Summary

The `workflows::workflow_trigger_is_community_confined` conformance row was failing at **setup**, not at the isolation assertion it was written to check. The relay's `handle_workflow_def` now requires the client to pass the workflow id via a `d` tag and rejects with `"invalid: missing d tag (workflow_id)"` (`crates/buzz-relay/src/handlers/command_executor.rs:668`). The test's helper was still built against the older server-generates-the-id contract, passed only `h` + `name` tags, and read the id out of the accept response — so every definition request was rejected with HTTP 400 and the row's actual assertion (A's workflow id can't be fired under B's community) never ran.

A false red hides a real relay property. This restores the row to green so the community-confinement fence is actually exercised.

Closes #5101.

## What changed

In `crates/buzz-test-client/tests/conformance_multitenant.rs`, inside `mod workflows`:

- `define_workflow` now generates a fresh `uuid::Uuid::new_v4()` locally, tags it via `d` alongside `h` and `name`, and returns it directly (no response parsing needed; the id was never server-generated, so there's nothing new to read out of the accept message).
- The trigger-side helpers (`trigger_workflow`) already pass `d` with the same UUID; no change there.
- Removed the stale comment claiming the server generates the id — it doesn't anymore.

Net effect: −7 LoC, same behavior, row goes from failing-at-setup to running the isolation check.

## Test plan

- `cargo check -p buzz-test-client --tests` — clean.
- The pre-existing failure was a panic at test setup (`HTTP 400`); after this change the row exercises the confinement assertion end-to-end when run against a live relay + Postgres (the standard conformance harness, which we do not run in CI for non-owning lanes).
- No production code changed.

### DIFF
--- a/crates/buzz-test-client/tests/conformance_multitenant.rs
+++ b/crates/buzz-test-client/tests/conformance_multitenant.rs
@@ -1741,20 +1741,22 @@ mod workflows {
     }
 
     /// Define a workflow in `channel_id` on `http_base`'s community (kind:30620,
-    /// `h`=channel, content=YAML). Returns the **server-generated** workflow id,
-    /// parsed out of the OK message (`response:{"workflow_id":"…"}`). This id is
-    /// the tenant-scoped handle the trigger door confines: defined under A, it
-    /// only resolves under A.
+    /// `h`=channel, `d`=client-generated workflow id, content=YAML). Returns the
+    /// id (which the server now requires the client to supply; cf.
+    /// `handle_workflow_def` rejecting "missing d tag (workflow_id)"). That id
+    /// is the tenant-scoped handle the trigger door confines: defined under A,
+    /// it only resolves under A.
     async fn define_workflow(http_base: &str, keys: &Keys, channel_id: &str, name: &str) -> String {
-        // `h` binds the channel; `name` is required by `handle_workflow_def`
-        // (it rejects "missing workflow name" before parsing YAML). We use the
-        // `name` tag, not `d`: the server *generates* the workflow id, and that
-        // generated id — not any client-supplied `d` — is the handle this row
-        // confines. A `d` tag here would falsely imply the trigger resolves by
-        // client key.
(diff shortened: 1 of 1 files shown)


iroiro147:

PR #5148, opened 2026-08-07
build(desktop): trust OS-installed CAs for websocket connections

abipalli:
Self-hosted deployments that terminate TLS with a private CA (step-ca, internal PKI) cannot be trusted by the desktop app today, by any means — installing the CA in the OS trust store has no effect.

### Cause

`desktop/src-tauri/Cargo.toml` pins:

```toml
tokio-tungstenite = { version = "0.29", features = ["rustls-tls-webpki-roots"] }
```

`webpki-roots` compiles the Mozilla root set into the binary and never consults the platform store. Every call site uses bare `connect_async` with no custom `Connector` (`native_websocket.rs:132`, `commands/pairing.rs:306`, `huddle/relay_api.rs:68`), so the feature-selected default root store is what is actually used.

Since the relay URL is user-configurable and self-hosting is a supported path, this makes a private-CA relay unreachable from the app.

### Change

Enable **both** root stores:

```toml
features = ["rustls-tls-native-roots", "rustls-tls-webpki-roots"]
```

They are additive — `tls.rs` calls `add_parsable_certificates` for native roots and then `extend(webpki_roots::TLS_SERVER_ROOTS)`.

Enabling native-roots *alone* would be a regression: tokio-tungstenite gates its empty-store hard error on `rustls-tls-webpki-roots` being absent, so zero native certs would fail every `wss://` connection. That is reachable via a stale `SSL_CERT_FILE`/`SSL_CERT_DIR` (honored by `rustls-native-certs` through `openssl-probe`, and preferred over system paths) or an image without `ca-certificates`. Keeping webpki as a floor avoids trading a trust gap for an availability one.

### Cost

None in dependencies. `webpki-roots` remains in the graph via the transitive `tokio-tungstenite 0.26`, and `rustls-native-certs` is already vendored along with `openssl-probe`, `schannel` and `security-framework`. No new crates and no new license surface for `cargo-deny`. Lockfile regenerated.

### Consistency

`reqwest` already resolves with `rustls-platform-verifier`, so the HTTP path honors the OS store; the websocket path was the only holdout. `git log -S rustls-tls-webpki-roots` traces the pin to `811c8f3f0` ("Replace LiveKit with WebSocket Opus audio relay") with no stated rationale, so this does not appear to overturn a deliberate decision.

Happy to put this behind a Cargo feature instead if you would rather keep the default bundled-only.

### DIFF
--- a/desktop/src-tauri/Cargo.lock
+++ b/desktop/src-tauri/Cargo.lock
@@ -210,7 +210,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
@@ -221,7 +221,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.61.2",
+ "windows-sys 0.60.2",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


xarielx:
Please accept this fix

PR #4258, opened 2026-08-02
feat(composer): add per-link preview controls

tellaho:
**Category:** new-feature
**User Impact:** Authors can review and remove individual link previews before sending while keeping the links in their message.

**Problem:** Link previews only appeared after send, and authors could only remove every preview at once. **Solution:** Resolve previews inside new and edit composers, persist canonical per-URL suppression through drafts and events, and apply the same controls across optimistic, timeline, and forum rendering without blocking Send.

<details>
<summary>File changes</summary>

**desktop/src-tauri/src/commands/messages.rs**
Accept and validate per-link suppression on sends and edits.

**desktop/src-tauri/src/events.rs**, **desktop/src-tauri/src/link_preview_tags.rs**
Build canonical suppression tags into message and edit events behind a strict HTTPS-only validator.

**desktop/src-tauri/src/egress_guard_tests.rs**, **desktop/src-tauri/src/huddle/pipeline.rs**, **desktop/src-tauri/src/lib.rs**
Wire the new builder argument through existing Rust callers and register the validator module.

**desktop/src/features/channels/ui/ChannelPane.types.ts**, **desktop/src/features/channels/ui/ChannelScreen.tsx**, **desktop/src/features/channels/useChannelPaneHandlers.ts**
Seed edit composers from effective event suppression and pass controls through send/edit handlers.

**desktop/src/features/messages/hooks.ts**, **desktop/src/features/messages/ui/useMentionSendFlow.ts**
Carry suppression through normal, mention-confirmed, optimistic, REST, and edit paths.

**desktop/src/features/messages/lib/applyEditTagOverlay.mjs**, **desktop/src/features/messages/lib/applyEditTagOverlay.test.mjs**, **desktop/src/features/messages/lib/formatTimelineMessages.ts**
Overlay URL-specific edit controls deterministically while preserving blanket suppression compatibility.

**desktop/src/features/messages/lib/imetaMediaMarkdown.ts**, **desktop/src/features/messages/lib/imetaMediaMarkdown.test.mjs**
Split preview-control tags from media, emoji, and mention tags for validated relay submission.

**desktop/src/features/messages/lib/useDrafts.ts**, **desktop/src/features/messages/ui/useDraftPersistSnapshot.ts**
Persist and restore per-draft suppression with backward-compatible migration.

[...]

PR #3818, opened 2026-07-30
feat(desktop): adding rich link previews to messages

tellaho:
## Overview

**Category:** improvement  
**User impact:** Link previews appear in the composer and travel as privacy-safe sender-authored snapshots, so recipients never contact the linked site merely by opening a conversation.  
**Problem:** Cold-cache link paste could freeze the composer before the URL painted; recipient-side unfurling leaked visits; invalid or unresolved preview work could interfere with sending or leave dead cards behind.  
**Solution:** Paint pasted links before starting cold resolver work, resolve only in the sender's composer, attach only complete validated snapshots at Send, and render authored snapshots without recipient fallback fetching.

## Behavior

- **Cold paste stays responsive:** bare and angle-bracket URL paste paths commit the visible link before resolver work begins.
- **Sender-only fetching:** metadata is resolved while composing; recipients render only the sender-authored snapshot.
- **Send never waits:** pending, failed, invalid, and unsendable previews are omitted. They do not block or cancel the message.
- **Terminal misses disappear:** failed, timed-out, or 404 resolver results remove the composer card while preserving visible link text.
- **Display-text links work:** Markdown links such as `[review the pull request](…)` produce and send the same snapshots as bare URLs.
- **Compact and Rich presentation:** Compact remains the default; Rich preserves source description line breaks and paragraphs.
- **Immediate draft-wide dismissal:** clicking × immediately hides all previews for the draft, suppresses links pasted later, and emits only `["link-preview", "none"]`. No confirmation detour. Suppression resets after send or clearing the draft.
- **Zero recipient fallback:** missing, stale, malformed, off-relay, unsupported, or suppressed snapshots remain ordinary visible links; recipients never regenerate them.

## Implementation

- Resolve previews from deferred composer URL state so paste can paint first.
- Upload finished preview media to the active community relay and snapshot only valid, sendable media references.
- Atomically capture ready snapshots at submit time; never append a late preview after send.
- Validate snapshot and suppression tags in desktop/native and relay ingestion, rejecting duplicate or mixed forms.
- Render composer previews as stable 55px attachment cards at desktop and narrow widths.
- Add deterministic E2E coverage for cold paste, ready/pending/failed/invalid previews, display-text links, multiline Rich descriptions, immediate dismissal, later-pasted links, and suppression reset.

## Validation

Validated head: `9807ba8952f190e76153834abf8ab61dd40be5e2`

[...]

PR #5159, opened 2026-08-07
fix(desktop): keep publishing a mention when a channel agent fails to start

### LINKED ISSUE #5099: Mentioning a provider-backed agent the desktop can't start drops the message instead of publishing it
Split out of #4839 at @TheSeydiCharyyev's suggestion — that PR fixes mention *visibility* and deliberately doesn't touch the start path. This is the start path.

## Setup

Self-hosted relay, six communities. 25 agents run as containers on my own VM, each its own `buzz-acp` with its own key. The desktop never spawns any of them and holds none of their keys. Desktop 0.5.5 on Windows 11.

## What I did

I wanted the agents to show up in the Agents tab, so I hand-wrote entries into `%APPDATA%\xyz.block.buzz.app\agents\managed-agents.json` — one per agent, `start_on_app_launch: false`, `auto_restart_on_config_change: false`, `backend: {"type":"provider","id":"...","config":{}}`, no `env_vars`.

They showed up. And mentions broke.

## What happens

Mention one of them in a channel it's already a member of, hit send:

```
Could not start agent mention: Backend · Claude: agent b0d474ab…7188 has no
private key available — the OS keyring may be unreachable. Refusing to start
without an identity; retry once the keyring is reachable.
```

The message is never published. Not delayed, not queued — `useMentionSendFlow` returns before the send, so it's gone. The agent is sitting right there in the channel and would have answered.

The keyring is fine, by the way. There's just no key in it for that agent, and there never will be — the key lives on the server.

Two paths reach it. `ensureManagedAgentMentionsReady` calls `startAgentMutation` for a participant it considers not-running, and `attachManagedAgentToChannel` does the same via `ensureRunning` for a non-participant. A provider-backed agent whose `status` isn't `deployed` takes the start branch in both.

## Why the status is never `deployed`

Because nothing sets it. There's no provider binary for my setup and there isn't going to be one — `docs/remote-agents.md` puts the provider on the desktop's machine, and the point of running the agents on the server is that the desktop stays a client with no deploy credentials on it. So the record claims a provider that doesn't exist, the app believes it, and tries to launch it.

That's my fault for writing a record that isn't true. But there's no honest thing to write instead: there's no way to register an agent as display-only. #4833 hit the same wall from the other side.

## Expected

A mention should publish. Starting an agent locally is an optimization — if it can't happen, the mention is still a valid mention, and the relay will route it to whoever is listening.

Concretely, either of these would fix it for me:

- Treat a start failure as non-fatal for the send: log it, publish the mention anyway.
- Or give the record a way to say "don't manage me" — a flag, or honouring `start_on_app_launch: false` as "never auto-start", which is what I meant when I set it.

I've been running the first as a local patch (skip the start branch for a known backend id) across six communities for a week. Mentions publish, the agents answer, nothing else changed.

## Not asking for

A provider for Docker. The relay directory already carries everything the client needs.

TheSeydiCharyyev:
## Summary

`ensureManagedAgentMentionsReady` collects every agent-preparation failure into one list, and any entry in that list returns before the send. The mention is not delayed or queued — it is gone.

That is fine when the agent is not in the channel yet, because nothing else would put it there. It is wrong when the agent is already a member. A managed agent that runs outside this desktop — a container on the user's own server, the shape `docs/remote-agents.md` describes — has no private key on this machine and never will, so its start fails every time. It sits in the channel and would answer, and the message is dropped anyway.

This PR makes a preparation failure block the send **only when the agent is not already a member of the channel the message is going to**. For an agent that is already there, the local start is an optimization: the relay routes the mention whether or not this desktop runs a copy, so the message is published and the failure is reported as a warning instead.

Everything else keeps today's behaviour:

- a failed attach still blocks — the agent would be left outside the channel;
- a failed start for an agent that is only being *prepared* for a channel this send creates or expands still blocks — that channel would be left without the participant it was expanded for. This is the case `channels.spec.ts` pins in "drops an expanded DM after agent startup fails", and that test is untouched and still passes.

The classification and both message strings live in `useMentionSendFlow.helpers.ts` so the hook itself does not grow: `useMentionSendFlow.ts` stays at exactly its current line count.

### Related issue

Closes #5099. Split out of #4839 by the reporter; that PR fixes mention *visibility* and deliberately does not touch the start path. Searched open PRs for the issue number and for these files: none found on this path (#4258 touches `useMentionSendFlow.ts` for link previews).

### Testing

New e2e, `mentions.spec.ts` → "mentioning an in-channel agent still sends when its start fails": an in-channel provider-backed agent whose start is rejected. On current `main` the mention chip never appears — the message is dropped. With this change it publishes and the failure shows as a warning.

- `pnpm --dir desktop test` — 4394 passed (5 new unit tests for the classification)
- `pnpm --dir desktop check` — biome, file-size ratchet, px-text, pubkey-truncation
- Focused Playwright run, 6 passed: the new test, both "drops an expanded DM" tests, both in-channel start/deploy tests, and the non-member add-and-start test

One note on the suite: full `channels.spec.ts` + `mentions.spec.ts` runs are flaky on this machine — a toast can cover the send button and time the click out. I ran the same two files on unmodified `main` as a control: 5 failures there, 1–2 with this change, a different set each time, and every one of them passes in isolation. The failures are not from this change.

### DIFF
--- a/desktop/src/features/messages/ui/useMentionSendFlow.helpers.test.mjs
+++ b/desktop/src/features/messages/ui/useMentionSendFlow.helpers.test.mjs
@@ -1,7 +1,96 @@
 import assert from "node:assert/strict";
 import test from "node:test";
 
-import { mergeMentionRecipients } from "./useMentionSendFlow.helpers.ts";
+import {
+  describeAgentReadinessFailures,
+  mergeMentionRecipients,
+} from "./useMentionSendFlow.helpers.ts";
+
+test("describeAgentReadinessFailures: no failures blocks nothing and warns about nothing", () => {
+  assert.deepEqual(describeAgentReadinessFailures([]), {
+    blocking: null,
+    warning: null,
+  });
+});
+
+test("describeAgentReadinessFailures: a launch failure for a channel member only warns", () => {
+  // #5099: the agent runs on the user's own server, so this desktop holds no
+  // key for it and the launch can never succeed. It is in the channel and
+  // answers there, so the mention must still go out.
+  const result = describeAgentReadinessFailures([
+    { blocking: false, message: "Backend · Claude: agent has no private key" },
+  ]);
+
+  assert.equal(result.blocking, null);
+  assert.equal(
+    result.warning,
+    "Could not start the mentioned agent; sending the mention anyway: Backend · Claude: agent has no private key",
+  );
+});
+
+test("describeAgentReadinessFailures: the warning never claims the message was sent", () => {
+  // The notice is emitted before Huddle sync, media upload and the send, each
+  // of which can still abort. Past-tense copy here would be a false delivery
+  // confirmation on an error path.
+  for (const count of [1, 2]) {
+    const { warning } = describeAgentReadinessFailures(
+      Array.from({ length: count }, (_, index) => ({
+        blocking: false,
+        message: `Agent ${index}: no private key`,
+      })),
+    );
+
+    assert.match(warning, /sending the mention anyway/);
+    assert.doesNotMatch(warning, /\bsent\b/i);
+    assert.doesNotMatch(warning, /\bdelivered\b/i);
+  }
+});
+
+test("describeAgentReadinessFailures: a blocking failure keeps stopping the send", () => {
+  const result = describeAgentReadinessFailures([
+    { blocking: true, message: "Fizz: Mock agent startup failed." },
+  ]);
+
+  assert.equal(
+    result.blocking,
+    "Could not start agent mention: Fizz: Mock agent startup failed.",
+  );
+  assert.equal(result.warning, null);
+});
+
+test("describeAgentReadinessFailures: a blocking failure wins over a warning", () => {
+  const result = describeAgentReadinessFailures([
+    { blocking: false, message: "Remote: no private key" },
+    { blocking: true, message: "Fizz: startup failed" },
+  ]);
+
+  assert.equal(
+    result.blocking,
+    "Could not start agent mention: Fizz: startup failed",
+  );
+  assert.equal(
+    result.warning,
+    "Could not start the mentioned agent; sending the mention anyway: Remote: no private key",
+  );
+});
+
+test("describeAgentReadinessFailures: several failures of one kind are joined and pluralised", () => {
+  const result = describeAgentReadinessFailures([
+    { blocking: true, message: "A: one" },
+    { blocking: true, message: "B: two" },
+    { blocking: false, message: "C: three" },
+    { blocking: false, message: "D: four" },
+  ]);
+
+  assert.equal(
+    result.blocking,
+    "Could not start agent mentions: A: one; B: two",
+  );
+  assert.equal(
+    result.warning,
+    "Could not start the mentioned agents; sending the mention anyway: C: three; D: four",
+  );
+});
 
 test("address-locked agents join explicit mentions without duplicating recipients", () => {
   const explicit = ["A".repeat(64), "b".repeat(64)];
--- a/desktop/src/features/messages/ui/useMentionSendFlow.helpers.ts
+++ b/desktop/src/features/messages/ui/useMentionSendFlow.helpers.ts
@@ -104,3 +104,58 @@ export function isManagedAgentRunning(agent: ManagedAgent) {
 export function isProviderBackedAgent(agent: ManagedAgent) {
   return agent.backend.type === "provider";
 }
+
+/**
+ * A mentioned agent that could not be prepared for the send.
+ *
+ * `blocking` is false only when the agent is already a member of the channel
+ * the message is going to. Launching it on this desktop is then an
+ * optimization: an agent that runs elsewhere — a container on the user's own
+ * server — has no private key on this machine and never will, so its launch
+ * fails every time, yet it is in the channel and still answers. Refusing to
+ * publish loses a valid mention and fixes nothing.
+ *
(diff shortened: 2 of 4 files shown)


wesbillman (changes requested):
Reviewing on Wes's behalf.

The membership-based classification is a real fix for #5099: an already-in-channel agent does not need this desktop to launch it for the relay mention to be valid, while attach/preparation failures for a non-member still block. The change is narrow, does not alter relay authorization or membership, and the new unit/E2E coverage exercises the core regression.

I am requesting one correction before putting this into this morning's release:

`readiness.warning` says **“Message sent without starting…”**, but it is displayed before Huddle synchronization, media upload, and `onSend`. Each of those can still fail or abort. In particular, a start failure followed by `sync_agents_to_active_huddle` failure produces a “Message sent” toast and then returns without sending anything; upload/send failure can likewise restore the composer after the success-tense toast has already been shown. That is false delivery confirmation on an error path.

Please either:

- make the pre-send warning future/present tense (for example, “Could not start …; sending the mention anyway”) and use `toast.warning`, or
- emit the current past-tense warning only after `finishSend` has successfully awaited `send`.

Add focused coverage for the selected ordering/wording so a later pre-send failure cannot claim delivery. Once corrected, I see no safety blocker to release.

wesbillman (commented):
I traced the membership/preparation state and reproduced the regression coverage at exact head `8857e525cae1045ab79135d5cd2cf886a835b842`. No blocking findings.

This is a real fix for #5099: a local launch failure previously aborted publication even when the mentioned identity was already a member and could receive/respond through a remote runtime. The new boundary is sound: only a pubkey present in the pre-existing, normalized channel-member set gets non-blocking treatment. Failed attaches and agents only prepared for a newly created/expanded channel remain blocking, including the existing expanded-DM failure behavior.

Local verification at this exact head:
- full Desktop unit suite: 4,485 passed, 0 failed
- Desktop TypeScript build/typecheck: passed
- focused E2E (`mentioning an in-channel agent still sends when its start fails`): passed
- Biome check on all four changed files: passed
- clean merge-tree with current `origin/main` (`e47894a133c2a685efd7aca5dd210daad8cd13b8`)

The fork workflow only exposed DCO/Semgrep/zizmor, so trusted repository CI still has not run. For a same-morning release I recommend running the normal Desktop CI before merge, but the code itself is narrow and release-safe. Minor non-blocking UX nit: the successful-send notice is emitted with `toast.error`; `toast.warning` would better match the behavior and copy.

wesbillman (changes requested):
**Correction after cross-review:** Carl identified a sequencing defect I missed in my initial assessment, and I agree it should be fixed before release. The warning currently says “Message sent…” and is shown before Huddle sync, upload, and `onSend`; any later failure can therefore tell the user a message was sent when it was not. Please use pre-send/future-tense copy with `toast.warning`, or emit the past-tense notice only after successful publication, with coverage.

The underlying #5099 fix and membership boundary remain sound. Local verification at exact head `8857e525cae1045ab79135d5cd2cf886a835b842` passed: 4,485 Desktop unit tests, TypeScript build/typecheck, focused regression E2E, and Biome on the four changed files; the branch also merge-trees cleanly with `origin/main` at `e47894a133c2a685efd7aca5dd210daad8cd13b8`. Trusted repository CI still has not run because this is a fork.

**Updated disposition: do not include this exact head in the release until the false-success notification is corrected.**

TheSeydiCharyyev:
Fixed in `acde6dc`, using the first option.

The notice is now pre-send tense — `Could not start the mentioned agent; sending the mention anyway` — and raised with `toast.warning`. Nothing on that path claims delivery any more.

Coverage for the ordering, as requested. New e2e `a start failure never reports delivery when the send itself fails`: the start is rejected and the send is rejected too. It asserts the pre-send notice is shown, the composer keeps the draft, no mention chip reaches the timeline, and no text on screen claims the message was sent. A unit test also asserts the copy directly: it must match `sending the mention anyway` and must not contain `sent` or `delivered` in either the singular or plural form, so the wording cannot regress silently.

The branch is also rebuilt on current `main`. #3818 landed in both changed files overnight and the old head no longer merged cleanly. `useMentionSendFlow.ts` is unchanged in length at 997 lines, so the size ratchet is untouched.

Local verification at `acde6dc`: 4541 desktop unit tests pass, typecheck and `pnpm check` pass, the focused mention set passes (5 tests, including both in-channel start/deploy cases and the non-member add-and-start case), and both `drops an expanded DM` tests still pass.

TheSeydiCharyyev:
Rebased onto `eed74bde`. Head is now `fa98a97f`.

Only one of the four files conflicted: `useMentionSendFlow.ts`. The e2e file merged on its own, even though #6956, #6862 and #6793 rewrote a large part of it.

The diff got smaller. #6838 added `existingMembers` to `ensureManagedAgentMentionsReady`, and that is the same "already a member of the channel" split this PR needs. So the local `isExistingMember` variable is gone, and the catch block now reads `blocking: !existingMembers.has(pubkey)`. This is also more correct than what the PR had before: the old line called `mentions.memberPubkeys.has(pubkey)` without normalizing the key.

The defect is still there on `eed74bde`. In the send flow, `if (agentReadiness.errors.length > 0)` still stops the send for every failure, including an agent that is already a member of the channel.

Gates on `fa98a97f`, Node 24.14.0: typecheck 0, `pnpm check` 0, `check-file-sizes` 0, unit tests 5805/5805, `mentions.spec.ts --project=smoke` 81/81 with no flakes. The two tests this PR adds pass against the rewritten harness.

PR #5162, opened 2026-08-07
feat(desktop): style structured agent reply labels in the message renderer

launchmaniac:
## Summary
Some agents reply in a fixed, labelled envelope — a status line, an answer block
(broken into "what it is / why it matters / done when"), a source, a confidence,
a next action, an owner, and a blocker/escalation line. This adds a small,
presentation-only rehype plugin (`rehypeStructuredReplyLabels`) that gives those
labels the existing `--primary` accent and `font-bold` weight so the structure
reads at a glance.

It is deliberately narrow and safe:
- Acts only when the message is a valid envelope (the seven top-level labels
  appear line-leading, in the canonical order). Ordinary prose and other
  messages are untouched.
- Styles only the exact labels at line-leading positions; values and body text
  are never styled; it never descends into `code`/`pre`/`a`/`blockquote`.
- Renderer-only: the stored message text is never altered — labels are wrapped
  for display in the HAST phase, the same place as `rehypeSearchHighlight`.
- Uses the existing semantic accent token (no hard-coded color) and is wired
  into the single markdown parse choke point (`buildMarkdownElement`).

### Related issue
none found

### Testing
- New focused unit tests (`rehypeStructuredReplyLabels.test.mjs`, 12 cases):
  all ten labels styled; values unstyled; visible text byte-for-byte unchanged;
  false positives rejected (lone/mid-line label, out-of-order, partial envelope,
  label inside inline code, envelope quoted in a blockquote); ordinary Markdown
  (bold/list/link) unchanged.
- Full desktop unit suite `pnpm test` (4492/4492), `pnpm typecheck`, and
  `biome check` all pass.

### DIFF
--- a/desktop/src/shared/lib/rehypeStructuredReplyLabels.test.mjs
+++ b/desktop/src/shared/lib/rehypeStructuredReplyLabels.test.mjs
@@ -1,0 +1,196 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+import rehypeStructuredReplyLabels from "./rehypeStructuredReplyLabels.ts";
+
+const LABEL_CLASS = "text-primary font-bold";
+
+// --- tiny HAST builders (mirror what remark-rehype + remark-breaks emit) ---
+const t = (value) => ({ type: "text", value });
+const br = () => ({
+  type: "element",
+  tagName: "br",
+  properties: {},
+  children: [],
+});
+const el = (tagName, children, properties = {}) => ({
+  type: "element",
+  tagName,
+  properties,
+  children,
+});
+const root = (children) => ({ type: "root", children });
+
+/** One paragraph whose lines are separated by <br> (the remark-breaks shape). */
+function paragraphWithBreaks(lines) {
+  const kids = [];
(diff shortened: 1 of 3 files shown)


launchmaniac:
@block/buzz-oss-team — requesting a review when you have a moment.

**Scope:** desktop renderer only, three files — a new presentation-only rehype plugin (`rehypeStructuredReplyLabels`) and its focused test, plus a one-line wiring into the markdown parse path (`buildMarkdownElement`). Renderer-only: stored message text is never altered.

**Local gates (all green):** 12/12 focused unit tests, 4492/4492 full desktop unit suite (`pnpm test`), `tsc --noEmit` clean, Biome clean.

**Visible checks passing:** DCO Check and zizmor. **CI** and **Desktop Release Candidate** are fork-gated (`action_required`) — would a maintainer mind approving the workflow runs?

After merge, `signed-macos-canary.yml` can be run on `main` to produce the signed artifact.

Thanks!

launchmaniac:
Withdrawing this submission because the renderer behavior does not meet the expected result. No merge or release was produced.

PR #5169, opened 2026-08-07
fix(acp): publish streamed agent replies

samiriasbeck:
## Summary

- capture natural ACP `agent_message_chunk` output and publish it as the agent reply after a successful turn
- keep only the latest final message when commentary is superseded
- prevent duplicate posts when a runtime also uses the legacy publish path
- preserve direct-message and thread reply semantics
- convert delegated `@name` references to participant-bound `p` tags only when uniquely and safely resolvable
- cap agent-to-agent reply depth to prevent loops

## Why

Some ACP runtimes return the complete answer through streamed message chunks but do not perform a second tool call to publish it. The harness should deliver that natural reply without another model round trip.

## Security

Delegated mentions fail closed: only owner-rooted thread participants are eligible. Ambiguous names, code blocks, outsiders, and self-mentions are ignored.

## Tests

- message chunk collection, replacement, and UTF-8 limit coverage
- delegated mention unique-resolution and fail-closed coverage
- `cargo test -p buzz-acp`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `just ci`

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -211,6 +211,14 @@ pub struct AcpClient {
     /// deltas. Both goose and buzz-agent emit this notification; goose gates
     /// on client capability advertisement, buzz-agent emits unconditionally.
     goose_usage: UsageTracker,
+    /// Natural-language output streamed by the agent during the current turn.
+    /// The harness consumes this after a successful prompt and publishes it to
+    /// Buzz, avoiding a second model/tool round-trip for chat delivery.
+    turn_message: String,
+    /// ACP message identifier associated with `turn_message`. When a runtime
+    /// emits commentary followed by a distinct final message, only the newest
+    /// message is retained for publication.
+    turn_message_id: Option<String>,
 }
 
 /// Recursively merge `overlay` into `base`, with `overlay` winning on scalar/shape
@@ -550,7 +558,14 @@ impl AcpClient {
             steering_supported: false,
             steer_rx: None,
             goose_usage: UsageTracker::default(),
+            turn_message: String::new(),
+            turn_message_id: None,
         })
+    }
+
+    /// Take the natural-language response streamed during the current turn.
+    pub(crate) fn take_turn_message(&mut self) -> String {
+        std::mem::take(&mut self.turn_message)
     }
 
     /// Attach a local observer feed to this ACP client.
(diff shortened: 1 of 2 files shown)


TheAirick:
I've been testing this exact path pretty hard with Hermes because it is the bug I'm hitting in normal use. Harness-owned publication is the right direction, but I found a few fail-closed gaps in this version that seem worth fixing before it ships:

1. `run_prompt_task` publishes for any `Ok(stop_reason)`. I think publication has to be limited to an observed `StopReason::EndTurn`; cancellation, max-token, limit, and other terminal outcomes should consume or requeue according to their exact semantics without publishing captured text.
2. The 64 KiB path truncates and publishes a partial response. I treated overflow as sticky invalidation until a new logical ACP message/prompt because partial model output can change meaning.
3. The duplicate guard is an author/channel/since query. An unrelated post can suppress the actual reply, while a lost submit acknowledgement can cause a newly signed duplicate on retry. I ended up needing a durable keyed operation that persists the exact signed event and reconciles/retries by event ID.
4. A tool call does not clear previously captured assistant text, so pre-tool commentary can become the final published reply. Current-session tool calls should reset capture; foreign-session updates should not affect it.
5. Unknown or ambiguous textual `@Name` markers are left visible and publication continues without a `p` tag. If textual mentions are meant to delegate work, I think every out-of-code marker needs trusted membership/profile resolution or the entire reply should fail before signing. Mixed Unicode ordering also needs a per-marker bounded scan so an unresolved earlier marker cannot borrow a later valid name.

I pushed the tested reference implementation here (based on Desktop 0.5.8, so this is evidence/reference rather than a merge-ready `main` branch):

https://github.com/TheVisher/buzz/commit/619e0df20cfcc6a80df04ea297366b6871446765

Verification on that commit:

- 748 `buzz-acp` unit tests
- 9 `buzz-acp` integration tests
- 97 `buzz-dev-mcp` tests
- strict Clippy, formatting, and diff checks
- bounded local end-to-end harness with a disposable kind-9 authorization: exact route/tags, one prompt for a duplicate trigger, durable `sent` ledger, lost-response reconciliation by the same event ID, verified event signature, and all restricted authority variables absent from the worker

Honest live-test caveat: I did not establish a clean phone-to-channel pass. Two local candidate attempts were invalidated by my temporary launch wrapper/process lifecycle, and the fail-closed path published zero replies and created zero broker operations. I am not claiming production acceptance from those attempts. The non-phone signed-event path and the source invariants are green.

PR #5173, opened 2026-08-07
Fix mobile DM recipient tags

Tiago-Sonar:
## Summary

- add all other channel participants as `p` tags on mobile DM messages
- apply the same recipient behavior to DM thread replies
- resolve channel type inside the shared send service and fail closed when metadata is unavailable
- preserve explicit-mention-only behavior for non-DM streams

## Why

Mobile DM messages without an explicit `@mention` were published without a recipient `p` tag. Recipient-gated agent harnesses could therefore ignore a visible message, making managed agents such as Donna appear not to answer.

## Validation

- `./bin/just ci`
- full mobile suite: 1,253 tests passed
- independent diff review completed; its thread-loading race finding is addressed by the second commit

## Origin

Buzz channel: `4c0769da-da97-4e9d-bf12-152ddbaec0aa`
Thread root: `22f9b299f0da00ce02648abdf49ff7a221cad8e68599565737817d3ed087bc1b`

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page.dart
+++ b/mobile/lib/features/channels/channel_detail_page.dart
@@ -523,6 +523,7 @@ class ChannelDetailPage extends HookConsumerWidget {
                             channelId: channel.id,
                             content: content,
                             mentionPubkeys: mentionPubkeys,
+                            isDirectMessage: resolvedChannel.isDm,
                             mediaTags: mediaTags,
                           ),
                     ),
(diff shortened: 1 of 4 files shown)


Tiago-Sonar:
@block/buzz-oss-team Review requested. This critical reliability fix is mergeable; DCO, Semgrep, and zizmor are green, and the PR body records the full local validation and independent review.

PR #5178, opened 2026-08-07
fix(acp): explain @mention as a session wake-up

### LINKED ISSUE #5176: base_prompt: explain what an @mention does to an agent, and offer reactions as the alternative
The base prompt tells agents *not* to `@mention` for acknowledgements or when speaking about someone (`crates/buzz-acp/src/base_prompt.md:54,59,80`), but it never says **what a mention actually does**: it starts a session on the mentioned agent. Without that, the rules read as etiquette, and agents trade them off against being polite. Agents mention each other to confirm receipt, the mention wakes a session, that agent feels obliged to answer and mentions back — loops of many turns with no human in them.

Two additions, both prose:

1. **State the mechanism, not just the rule.** "An `@mention` starts a session on that agent — it is a wake-up, not a name. Use it only when you actually want them to act. To refer to an agent without waking them, write the bare name with no `@`."
2. **Give acknowledgement somewhere to go.** Agents currently have only two moves: reply or stay silent. Silence looks like ignoring. Point them at `buzz reactions add` — a reaction signals "seen" without waking anyone. The Communication Patterns section doesn't mention reactions at all today.

Also worth saying explicitly: **an agent may read a message and do nothing.** If there is no to-do in it, no reply is needed — and above all no mention back at the sender.

nawazish2:
## Summary

Agents were `@mention`ing each other for acknowledgements because the base prompt never said what a mention *does* — only that they shouldn't spam them. Mentions wake sessions; bare names and reactions don't.

Fixes #5176

## Changes (`base_prompt.md`)

- Mentions open with: `@mention` is a **wake-up**, not a name tag — use it when you want action; bare name otherwise
- Callback section: you may read and do nothing; don't mention back just to prove you saw it
- New **Reactions** subsection: `buzz reactions add` for "seen" without starting a session
- General: silence is fine; no inventing a reply/mention so the thread looks handled
- Bare-name narrative examples no longer include `@` (that was undercutting the rule)

Also added a small include-str test so the wake-up / reactions guidance doesn't get edited away quietly.

## Test plan

- [x] `cargo test -p buzz-acp --lib agent_draft_prompt_tests`

### DIFF
--- a/crates/buzz-acp/src/base_prompt.md
+++ b/crates/buzz-acp/src/base_prompt.md
@@ -47,16 +47,19 @@ For explicit changes to an existing personal agent, use `buzz agents draft-updat
 
 ### Mentions
 
+An `@mention` is a **wake-up**, not a name tag. On Buzz it starts (or resumes) a session for the mentioned agent so they act. Use it only when you actually want that person or agent to do something now. To talk *about* someone without waking them, write their bare display name with no `@`.
+
 - Use the person's **exact full display name** after `@` (e.g., `@Will Pfleger`, not `@Will`). Partial names fail silently.
 - Do NOT format mentions with bold, italic, or backticks — it breaks notification delivery.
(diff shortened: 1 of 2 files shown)


nawazish2:

PR #5179, opened 2026-08-07
feat(desktop): collapse long messages behind Show more

### LINKED ISSUE #5175: feat(desktop): collapse very long messages behind a "Show more" toggle
Long messages render at full height in the channel, so a single message can push everything else off screen and make scrolling back through a conversation painful. Agent output in particular is often hundreds of lines.

**Proposal:** clamp a message to a fixed max height (Slack uses ~a dozen lines) and show a "Show more" / "Show less" toggle when it overflows. Expansion state is per-message and not persisted.

**Notes**
- Applies to both human and agent messages; same treatment for thread previews.
- Fenced code blocks should be included in the clamp, not exempt from it.
- Search/keyboard navigation should auto-expand a match so it is not hidden behind the fold.

nawazish2:
## Summary

Long agent output (and big pastes) used to dominate the channel. Messages now clamp at ~12 lines with a **Show more** / **Show less** toggle.

Fixes #5175

## Behavior

- Collapsed by default when content is taller than ~240px
- Fade + toggle under the fold
- Expand state is per mounted row (not persisted)
- **Search query** or **route highlight** force-expands so matches aren't hidden
- Diffs / huddles / wave attachments are unchanged (only the markdown body path)
- Virtualizer height estimate caps at the collapsed body size (+ toggle)

## Test plan

- [x] `pnpm test` (desktop) — includes new clamp helpers
- Manual: open a channel with a multi-page agent message → Show more / less
- Manual: search for text inside a long message → body expands

### DIFF
--- a/desktop/src/features/messages/lib/rowHeightEstimate.ts
+++ b/desktop/src/features/messages/lib/rowHeightEstimate.ts
@@ -33,6 +33,10 @@ const PREVIEW_CARD = 70;
 const MESSAGE_ITEM_BOTTOM_PADDING = 10; // TimelineMessageList pb-2.5
 const MIN_ESTIMATE = 60; // never reserve less than the old flat floor
 const CONTINUATION_MIN_ESTIMATE = 28;
+// CollapsibleMessageBody clamps long bodies; over-reserving the full dump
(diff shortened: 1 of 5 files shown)


nawazish2:

PR #5180, opened 2026-08-07
feat(desktop): drag to raise the message composer height cap

### LINKED ISSUE #5177: feat(desktop): let the message composer be dragged taller than its 128px cap
The composer starts at ~two lines, grows as you type, and stops at a hard cap — `max-h-32` (128px, roughly six lines) at `desktop/src/features/messages/ui/MessageComposer.tsx:979`. Past that it scrolls, so when you write a long message you are composing through a six-line porthole: you can't see the paragraph you just wrote while writing the next one, and reviewing before sending means scrolling inside a tiny box.

**Proposal:** make the cap user-adjustable — a drag handle on the top edge of the composer that resizes it upward, the way Slack and iMessage do. The chosen height persists (per window, not per channel), and auto-grow still applies below it.

**Notes**
- Keep a sane upper bound so the composer can't swallow the whole channel — something like 60% of the pane height.
- Double-click on the handle to reset to the default.
- Existing behaviour is unchanged for anyone who never drags it.

nawazish2:
## Summary

Composer input used a hard `max-h-32` (128px). Long messages meant scrolling inside a six-line box while typing.

Fixes #5177

## Behavior

- Grip on the top edge of the composer — drag **up** to raise the max height
- Floor stays at the old default (128px); auto-grow still fills below the cap
- Ceiling is **60% of the channel pane** so the timeline stays usable
- Double-click the grip to reset to 128px
- Preference stored in `localStorage` (`buzz.composer.maxHeightPx`) — per device, not per channel

## Test plan

- [x] unit tests for `clampComposerMaxHeight`
- [x] biome check on touched files
- Manual: type a long draft → drag the top grip → more lines visible without inner scroll until the new cap
- Manual: double-click grip → back to default
- Manual: reload app → height preference sticks

### DIFF
--- a/desktop/src/features/messages/ui/MessageComposer.tsx
+++ b/desktop/src/features/messages/ui/MessageComposer.tsx
@@ -58,6 +58,13 @@ import { useComposerContentState } from "./useComposerContentState";
 import { useDraftPersistLifecycle } from "./useDraftPersistSnapshot";
 import { submitMessageEdit } from "./submitMessageEdit";
 import type { MessageComposerProps } from "./MessageComposer.types";
+import {
+  clampComposerMaxHeight,
+  DEFAULT_COMPOSER_MAX_HEIGHT_PX,
+  readStoredComposerMaxHeight,
+  writeStoredComposerMaxHeight,
+} from "./composerMaxHeight";
+
 function MessageComposerImpl({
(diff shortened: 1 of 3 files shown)


nawazish2:

PR #5181, opened 2026-08-07
fix(acp): resume pending turns after Desktop restart

Tiago-Sonar:
## Summary

Persist queued and in-flight ACP turns in a bounded, atomic local ledger so managed agents can resume admitted work after Buzz Desktop restarts when the per-agent `resume_on_restart` setting is enabled.

This wires the setting through Desktop configuration and the agent editor, restores eligible turns during ACP startup, and preserves the current spawn-snapshot, parallelism, and access-control behavior. Ledger retention is bounded and stale entries are pruned.

### Related issue

No dedicated GitHub issue found. This repair originated from a live Buzz reliability assessment after admitted turns disappeared during Desktop outages.

Originating Buzz thread:

`buzz://message?channel=4c0769da-da97-4e9d-bf12-152ddbaec0aa&id=31a1776cadd43bc656a438e5a89b052b476bf9f1919ae73bcbd99e7738e60f4a`

### Testing

- `BUZZ_ACP_LAZY_POOL=false BUZZ_ACP_MULTIPLE_EVENT_HANDLING=steer cargo test -p buzz-acp` — 765 unit tests, 9 lifecycle integration tests, 1 setup-ledger integration test, and doc tests passed.
- `./bin/just desktop-ci` — 4,480 frontend tests, production build, Tauri compilation, and complete Tauri workspace tests passed.
- `cargo clippy -p buzz-acp --all-targets -- -D warnings` — passed.
- `./bin/just desktop-tauri-clippy` — passed.
- `../bin/pnpm exec playwright test --project=smoke tests/e2e/edit-agent.spec.ts tests/e2e/global-agent-config-screenshots.spec.ts` — 29/29 passed.
- `git diff --cached --check` — passed before commit.

### Scope boundary

This closes the restart-loss path for admitted queued/in-flight turns. It does not add the separate full event-stage audit ledger for receipt, admission/rejection, ACP submission, completion, publication, and relay-accepted reply IDs.

PR #5184, opened 2026-08-07
feat(acp): persist bounded turn audit trail

Tiago-Sonar:
## Summary

Persist a bounded, content-free audit trail for inbound ACP turns so operators can distinguish receipt, rejection, queueing, ACP execution, completion, and reply publication failures.

Each inbound event keeps a stable correlation ID and stage timestamps for:

- decoded relay receipt inside the harness;
- admission or rejection reason and queue outcome;
- turn assignment, ACP submission, first semantic output, and terminal outcome;
- `buzz messages send` attempts, publish failures, relay rejections, and relay-accepted reply event IDs.

The audit is enabled by default with CLI/environment opt-out and retention controls. Writes use the existing cross-platform atomic-write abstraction, owner-only permissions on Unix, bounded records and gap markers, corruption preservation, and a shutdown drain. The persisted schema excludes message bodies, prompts, ACP/tool payloads, errors, credentials, session IDs, and relay URLs. When relay observation is disabled, the live observer feed does not retain raw ACP frames for replay.

### Related work

This complements #5181 without duplicating its queued/in-flight restart recovery. Events filtered before they reach the harness remain outside this audit boundary.

Originating Buzz thread:

`buzz://message?channel=4c0769da-da97-4e9d-bf12-152ddbaec0aa&id=78ce91d3e5ddb699e91bdb50ef777ff1957a559e31fc229ab1d77d490db4cb9e`

### Testing

- `env -u BUZZ_ACP_LAZY_POOL -u BUZZ_ACP_MULTIPLE_EVENT_HANDLING -u BUZZ_ACP_DEDUP cargo test -p buzz-acp` — 704 unit tests, 9 lifecycle integration tests, and doc tests passed.
- `cargo test -p buzz-acp turn_audit::tests` — 13 audit persistence/failure-path tests passed.
- `./bin/just desktop-ci` — 4,480 frontend tests, production build, Tauri compilation, and the complete Tauri workspace tests passed.
- `cargo clippy -p buzz-acp --all-targets -- -D warnings` — passed.
- `cargo fmt --all -- --check` — passed.
- `git diff --cached --check` — passed before commit.

The final diff also received an independent read-only review; all findings were addressed before publication.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -801,6 +801,7 @@ name = "buzz-acp"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "atomic-write-file",
  "base64 0.22.1",
  "buzz-core",
  "buzz-persona",
--- a/crates/buzz-acp/Cargo.toml
+++ b/crates/buzz-acp/Cargo.toml
@@ -53,6 +53,7 @@ url = { workspace = true }
 sha2 = { workspace = true }
 base64 = "0.22"
 hex = { workspace = true }
+atomic-write-file = "0.3"
 
 # Logging
 tracing = { workspace = true }
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -468,6 +468,26 @@ pub struct CliArgs {
     #[arg(long, env = "BUZZ_ACP_RELAY_OBSERVER", default_value_t = false)]
     pub relay_observer: bool,
 
+    /// Persist a bounded, content-free local audit trail for inbound turns.
+    /// `--turn-audit=false` disables it.
+    #[arg(
+        long,
+        env = "BUZZ_ACP_TURN_AUDIT",
+        default_value_t = true,
+        action = clap::ArgAction::Set,
+        num_args = 0..=1,
+        default_missing_value = "true",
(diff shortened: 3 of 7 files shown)


Tiago-Sonar:
@block/buzz-oss-team Review requested. This critical turn-audit fix is mergeable; DCO, Semgrep, and zizmor are green, and the PR body records full ACP/Desktop validation plus independent review.

Silentpartnercoding:
I think the audit can record a publish attempt for a tool call that never tried to publish. `contains_buzz_message_send` recursively collects every string in the update and then substring-matches the joined text, so an unrelated command or tool argument that merely contains `buzz messages send` is classified as a send. Its later `tool_call_update` can then become a completed or relay-accepted publish attempt even though the tool identity was unrelated.

Could this match the structured tool identity and parsed argv for the supported send shapes instead of searching arbitrary payload strings? A negative fixture where a non-shell tool reads documentation containing `buzz messages send` should leave `publish_attempts` empty; the existing direct shell and structured-command shapes should still produce one attempt.

That distinction matters because this file is operator evidence: a false positive turns “the agent mentioned a command” into “the agent attempted a side effect.”

PR #5185, opened 2026-08-07
fix(media): enforce channel ACL on bound blobs (follow-up to 769ac70)

merttcetn:
## Problem

Blossom media reads (`GET`/`HEAD /media/{sha256}`) currently require only relay membership (`769ac70`). A blob uploaded for a private channel stays readable by any relay member who knows the SHA-256, even after being removed from that channel.

The `769ac70` commit message says as much: *"It does not yet bind a blob to its originating channel, so someone removed from a private channel can still read a known blob while remaining a relay member. That channel-ACL follow-up remains required."*

## What this PR does

Adds the channel binding to the sidecar and enforces it on read.

**`buzz-media`**
- `BlobMeta.channel_id: Option<Uuid>` — serde `default` + `skip_serializing_if = "Option::is_none"`. Existing sidecars parse as unbound; new unbound sidecars omit the key entirely, so an older relay binary can still read them. No migration.
- `extract_channel_id()` reads the `h` tag off the kind:24242 Blossom auth event in `process_upload` / `process_file_upload` / `process_video_upload`. It **fails closed**: an `h` tag that is present but malformed, nil, or ambiguous (two different UUIDs) rejects the upload with `400 InvalidTag` rather than silently downgrading the blob to unbound.

**`buzz-relay`**
- `enforce_media_channel_acl()` runs inside the existing sidecar gate, before any blob I/O. When a sidecar is bound and the channel's `visibility == "private"`, the requester must be a current member (`is_member_cached`). Open channels and unbound blobs are unaffected.
- Enforced on `GET`, `HEAD`, and the `.thumb.jpg` path (which checks the parent blob's sidecar).
- Misses return **404, not 403**, so a removed member cannot probe for the existence of private blobs. Channel-lookup and membership failures also fail closed to 404, logged at `debug!` for an expected missing channel and `warn!` + `buzz_media_channel_acl_db_errors_total` for real DB/transport errors.
- `BlobReadAuthority::{User, Operator}` replaces an `Option<&[u8; 32]>` requester pubkey, so skipping the ACL is an explicit word at the call site instead of the default that falls out of passing `None`. The admin feedback-attachment reader (already admin-authorized and pinned to the feedback row's provenance) is the only `Operator` caller.

## Scope — read this before reviewing

**No first-party client emits an `h` tag on its upload auth event yet** (`desktop/src-tauri/src/commands/media.rs`, `mobile/lib/shared/relay/media_upload.dart`, `crates/buzz-cli/src/client.rs` all send only `t`/`x`/`expiration`/`server`). Official client flows therefore write unbound sidecars. However, the relay accepts valid Blossom auth events from custom clients, so server-side `h` acceptance is reachable immediately and must not be treated as inert.

This PR is deliberately the mechanism plus read-side enforcement. Two things are left as follow-ups:

1. **Clients binding uploads to their channel.** Touches three platforms and is what actually closes the leak.
2. **Binding vs. content-addressed dedupe.** `process_buffered_upload` short-circuits when the blob and sidecar both exist and does not rewrite the sidecar, so the first upload of a given hash wins the binding. Once clients emit `h`, the same bytes posted to an open channel *after* a private one would inherit the private binding and 404 for non-members. That semantic needs a decision (rebind on re-upload? widen to least-restrictive? bind per-reference instead of per-blob?) before enforcement is switched on end-to-end.

Also not addressed: the upload path does not verify that the uploader is actually a member of the channel named in the `h` tag. Combined with (2), that is a poisoning vector for custom clients as soon as server-side support is deployed.

## Design decision requested

Review feedback correctly identified that a single `(community, sha256) -> channel_id` sidecar cannot safely represent multi-channel access grants, and that the first-write-wins dedupe path can make an attacker-controlled binding authoritative. Before merging server-side `h` acceptance or enabling first-party client `h` emission, the authority model needs an explicit maintainer decision: either land only dormant read-side groundwork while rejecting upload `h` tags, or replace the sidecar authority with tenant-scoped durable grants created with accepted channel events. No client `h` emission should land until that decision is resolved.

## Testing

- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test -p buzz-media` — 120 passed
- [x] `cargo fmt --all -- --check` — clean
- [x] Unit tests for `extract_channel_id`: absent tag, single valid UUID, non-UUID value, nil UUID, empty value, valueless `["h"]`, duplicate-identical (idempotent), conflicting UUIDs
- [x] Unit tests for the sidecar back-compat contract: legacy JSON without `channel_id` parses as unbound; unbound serializes without the key; bound round-trips
- [x] Unit test that `MediaError::InvalidTag` maps to HTTP 400
- [ ] Integration (`just test`) not run locally — needs Postgres + Redis
- No e2e for the ACL itself yet; it requires a client that emits `h`, which lands with follow-up (1).

### DIFF
--- a/crates/buzz-media/src/error.rs
+++ b/crates/buzz-media/src/error.rs
@@ -18,6 +18,8 @@ pub enum MediaError {
     InvalidImage,
     #[error("media contains metadata or a non-canonical metadata channel")]
     MetadataForbidden,
+    #[error("invalid tag value for {0}")]
+    InvalidTag(&'static str),
     #[error("invalid signature")]
     InvalidSignature,
     #[error("invalid auth event kind")]
@@ -151,6 +153,7 @@ impl IntoResponse for MediaError {
             | Self::InvalidVideo
             | Self::InvalidImage
             | Self::MetadataForbidden => (StatusCode::UNPROCESSABLE_ENTITY, self.to_string()),
+            Self::InvalidTag(_) => (StatusCode::BAD_REQUEST, self.to_string()),
             Self::Io(_) | Self::StorageError(_) | Self::Internal => {
                 tracing::error!(error = %self, "media storage error");
                 (StatusCode::INTERNAL_SERVER_ERROR, "internal error".into())
@@ -177,6 +180,16 @@ mod tests {
                 StatusCode::UNSUPPORTED_MEDIA_TYPE
             );
         }
+    }
+
+    #[test]
+    fn invalid_tag_maps_to_400() {
+        // A malformed `h` (channel) tag on the Blossom auth event is a caller
+        // mistake, not an auth failure — it must surface as 400, not 401/500.
+        assert_eq!(
+            MediaError::InvalidTag("h").into_response().status(),
+            StatusCode::BAD_REQUEST
+        );
     }
 
     #[test]
--- a/crates/buzz-media/src/storage.rs
+++ b/crates/buzz-media/src/storage.rs
@@ -330,6 +330,67 @@ mod tests {
         );
     }
 
+    /// Sidecars written before channel binding existed have no `channel_id`
+    /// key. They must still parse, as unbound. This is what makes rolling the
+    /// channel ACL out migration-free — every already-stored sidecar in S3
+    /// predates the field.
+    #[test]
+    fn blob_meta_legacy_sidecar_without_channel_id_parses_as_unbound() {
+        let legacy = r#"{
+            "dim": "800x600",
+            "blurhash": "LEHV6nWB2yk8pyo0adR*",
+            "thumb_url": "https://media.example.com/abc.thumb.jpg",
+            "ext": "jpg",
+            "mime_type": "image/jpeg",
+            "size": 1234,
+            "uploaded_at": 1700000000
+        }"#;
+
+        let meta: BlobMeta = serde_json::from_str(legacy).expect("legacy sidecar must still parse");
+        assert_eq!(meta.channel_id, None, "legacy sidecars are unbound");
+        // The rest of the sidecar must survive untouched.
+        assert_eq!(meta.ext, "jpg");
+        assert_eq!(meta.mime_type, "image/jpeg");
+        assert_eq!(meta.size, 1234);
+        assert_eq!(meta.uploaded_at, 1700000000);
+    }
+
+    /// The other half of the back-compat contract: an unbound sidecar must
+    /// serialize *without* the key, so newly written sidecars stay readable by
+    /// any older relay still running the previous binary.
+    #[test]
+    fn blob_meta_unbound_omits_channel_id_key() {
+        let meta = BlobMeta {
+            ext: "jpg".to_string(),
+            mime_type: "image/jpeg".to_string(),
+            size: 1234,
+            channel_id: None,
+            ..Default::default()
+        };
+
+        let json = serde_json::to_value(&meta).expect("serialize unbound sidecar");
+        assert!(
+            json.get("channel_id").is_none(),
+            "unbound sidecars must omit channel_id entirely, got {json}"
+        );
+    }
+
+    #[test]
+    fn blob_meta_bound_round_trips_channel_id() {
+        let channel_id = uuid::Uuid::from_u128(0x0bad_c0de);
+        let meta = BlobMeta {
+            ext: "jpg".to_string(),
+            mime_type: "image/jpeg".to_string(),
+            size: 1234,
+            channel_id: Some(channel_id),
+            ..Default::default()
+        };
+
+        let json = serde_json::to_string(&meta).expect("serialize bound sidecar");
+        let parsed: BlobMeta = serde_json::from_str(&json).expect("round-trip bound sidecar");
+        assert_eq!(parsed.channel_id, Some(channel_id));
+    }
+
     #[test]
     fn partial_static_keys_are_rejected() {
         let err = match MediaStorage::new(&storage_config("buzz_dev", "")) {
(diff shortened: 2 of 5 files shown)


merttcetn:
Addressed Copilot feedback in a952534f3:

- `InvalidTag("h") → 400` for malformed `h` (fail-closed, not silent downgrade)
- Fixed 2 test `BlobMeta` literals (`channel_id: None`)
- Added debug/warn logging + `buzz_media_channel_acl_db_errors_total` counter before 404 collapse (keeps anti-probing)

PTAL @copilot-pull-request-reviewer

merttcetn:
@jmecom — follow-up to `769ac70`, which noted *"That channel-ACL follow-up remains required."*

This adds an optional `channel_id` to the blob sidecar and enforces private-channel membership on reads (GET, HEAD, and the `.thumb.jpg` path, which checks the parent blob's sidecar). Misses return 404 rather than 403 so a removed member can't probe for the existence of private blobs. Back-compatible via serde defaults — existing sidecars parse as unbound, no migration.

To be upfront about scope: **no client emits an `h` tag on its upload auth event yet** (`desktop/src-tauri/src/commands/media.rs`, `mobile/lib/shared/relay/media_upload.dart`, and `crates/buzz-cli/src/client.rs` all send only `t`/`x`/`expiration`/`server`). So every sidecar written today is unbound and this ACL is live but inert in production. This PR is deliberately the mechanism plus read-side enforcement.

Two things are called out as follow-ups in the PR description rather than silently deferred:

1. Clients binding uploads to their channel — touches three platforms, and is what actually closes the leak.
2. How binding should interact with content-addressed dedupe. `process_buffered_upload` short-circuits when blob and sidecar both exist and doesn't rewrite the sidecar, so the first upload of a given hash wins the binding. Once clients emit `h`, the same bytes posted to an open channel *after* a private one would inherit the private binding and 404 for non-members. That semantic needs a decision before enforcement is switched on end-to-end.

Local gates are green: `cargo clippy --workspace --all-targets -- -D warnings` (clean), `cargo test -p buzz-media` (120 passed), `cargo fmt --all -- --check` (clean). Integration (`just test`) hasn't been run locally yet — it needs Postgres + Redis.

One ask: since this is a fork PR, the `CI`, `Docker image`, and `Desktop Release Candidate` workflows are all sitting at `action_required` and have never run on this branch. If someone can approve the workflow runs, the real suite will actually exercise this.

samiriasbeck:
I reproduced the channel-reference lookup against an isolated PostgreSQL instance and agree that the read path needs a durable blob-to-channel authorization record rather than an event-history scan.

The open design questions in this PR are security-critical, not only product semantics:

1. **Hash reuse can widen access.** If authorization is derived from any message that references a SHA-256, a relay member who learns a private hash can cite it in an open channel and make the blob readable there.
2. **First-writer binding can be poisoned.** If the first `h` tag wins, a member can pre-bind known bytes to a channel they control unless upload authorization proves both channel membership and authority to bind those bytes.
3. **A single sidecar channel cannot represent legitimate multi-channel sharing** without choosing either over-restriction or widening.

I suggest separating immutable blob metadata from explicit access grants:

```text
media_blobs(community_id, sha256, uploader_pubkey, ...)
media_channel_grants(community_id, sha256, channel_id, source_event_id,
                     granted_by, created_at, revoked_at)
```

Grant creation should be part of accepted event ingestion, not GET-time history inference, and should require:

- the source event is accepted in the same community and channel;
- the event author currently has write access to that channel;
- the author is either the original uploader or already has read authority for an existing grant to the blob;
- `(community_id, sha256, channel_id, source_event_id)` is idempotent;
- deletion/revocation has an explicit policy and never silently widens access.

Reads can then authorize against active grants plus current channel membership, with a community-leading index. This also gives clients a migration path: the `h` tag can remain a declaration, while the relay creates the authoritative grant only after validation.

Acceptance cases I would add before clients start emitting `h`:

- removed private-channel member gets 404 for GET, HEAD and thumbnail;
- referencing a leaked hash from an open channel does not create a grant;
- first-writer poisoning with an unrelated channel is rejected;
- legitimate re-share by a currently authorized reader creates a second grant;
- identical bytes uploaded independently do not make one community query another;
- malformed/duplicate/conflicting `h` tags fail closed;
- cache reuse revalidates current channel membership;
- DB lookup failures fail closed without leaking blob existence.

The current sidecar work is still useful as a compatibility mechanism, but I would avoid enabling client `h` emission until the grant-authority rule is decided.

merttcetn:
@samiriasbeck Thanks for the detailed threat-model review. I agree with the central point: a single `(community, sha256) -> channel_id` sidecar plus first-write-wins dedupe is not a sufficient long-term authorization model for multi-channel sharing.

I corrected the PR description to remove the claim that the mechanism is inert. First-party clients do not emit `h` today, but a custom client can exercise server-side `h` acceptance immediately, so client rollout is not a security boundary.

Two nuances for the acceptance model:

- A leaked hash alone cannot poison the current upload binding because BUD-11 verifies the uploaded bytes against the `x` tag; the actor must possess the matching bytes.
- Merely citing a hash in an open-channel message does not mutate the current sidecar. That widening case becomes relevant when grant creation is moved to accepted event ingestion, so the uploader-or-existing-authority rule must be enforced there.

@jmecom I think this now needs an explicit scope decision before more implementation: either (1) keep only read-side groundwork and reject upload `h` tags until the authority model exists, or (2) rework/supersede this PR around tenant-scoped durable grants created atomically with accepted channel events, including revocation semantics. I will not add first-party client `h` emission until that decision is resolved.

PR #5195, opened 2026-08-07
fix(buzz-agent): Responses reasoning summary, Anthropic display:summarized, ACP v2 messageId

wpfleger96:
Three pre-existing gaps in the buzz-agent observer feed fixed together per Will's ruling ("all 3 in the current PR"):

1. **OpenAI/DBv2-GPT route** — `responses_body` never requested `reasoning.summary`; GPT-family models billed thinking tokens but returned `summary: []`.
2. **Anthropic/DBv2-Claude route** — `anthropic_thinking_config()` never sent `thinking.display`; newest Claude models (Opus 5, Sonnet 5, Fable 5, Mythos 5, Opus 4.7/4.8, Mythos Preview) default to `display:"omitted"`, returning thinking blocks with an empty `thinking` field — observer rendered nothing.
3. **ACP v2 compliance** — buzz-agent negotiates ACP v2 but emitted `agent_thought_chunk` and `agent_message_chunk` without `messageId`, which ACP v2's `ContentChunk` requires (`messageId` + `content` both required at schema head `d13d1baa`).

## Changes

**`crates/buzz-agent/src/config.rs`**
- New `ThinkingSummary` enum (`Auto`/`Concise`/`Detailed`) with `BUZZ_AGENT_THINKING_SUMMARY` env var (default `Auto`); mirrors `BUZZ_AGENT_THINKING_EFFORT` pattern
- `anthropic_thinking_config()` now emits `"display": "summarized"` in both the adaptive shape and the manual-budget shape whenever thinking is enabled
- Rewrote `is_adaptive_thinking_model` and `anthropic_thinking_config` doc comments to match Anthropic's exact three-way per-model terminology (doc: https://platform.claude.com/docs/en/build-with-claude/thinking-troubleshooting#supported-models):
  - Opus 4.6/4.7/4.8, Sonnet 4.6: **Off** — thinking OFF by default; `type:"adaptive"` required to enable
  - Opus 5, Sonnet 5: **On** — thinking on by default, can be disabled; we still send `type:"adaptive"` to activate `output_config.effort`
  - Fable 5, Mythos 5, Mythos Preview: **Always on** — thinking cannot be disabled; we still send `type:"adaptive"` to activate `output_config.effort`

**`crates/buzz-agent/src/llm.rs`**
- `responses_body` emits `reasoning.summary` alongside `reasoning.effort` when effort is set (gated — no bare `reasoning:{summary}` without effort)
- Covers both the pure-OpenAI Responses path and the DBv2 GPT-family Responses path

**`crates/buzz-agent/src/agent.rs`**
- `agent_thought_chunk` carries `"messageId": format!("{run_id}-thought-{round}")`
- `agent_message_chunk` carries `"messageId": format!("{run_id}-message-{round}")`
- The two IDs are distinct (thought and assistant are two logical messages per the ACP v2 Message ID RFD)
- `run_id` is a fresh random token per `session/prompt` invocation so IDs are session-unique across multiple prompts

**`crates/buzz-agent/src/lib.rs`**
- `run_id` plumbed into `RunCtx` (was already generated in `run_prompt`, just not threaded through)

**`crates/buzz-agent/tests/golden_transcripts.rs`**
- `test_acp_v2_chunks_carry_message_id` — negotiates v2, drives two consecutive `session/prompt` calls, asserts: both chunk types carry non-empty `messageId`; thought and message IDs are **distinct**; IDs do **not** recur across the two prompts in the same ACP session

**`desktop/src-tauri/src/managed_agents/env_vars.rs`**
- `BUZZ_AGENT_THINKING_SUMMARY` added to `is_safe_to_reveal` allowlist

**`desktop/src-tauri/src/commands/agent_config_tests.rs`**
- Tests for `BUZZ_AGENT_THINKING_SUMMARY` allowlist entry (case-insensitive)

## Tests added

- `parse_thinking_summary_round_trips_all_values`
- `parse_thinking_summary_unset_and_empty_yield_auto`
- `parse_thinking_summary_is_case_insensitive`
- `parse_thinking_summary_rejects_unknown_value`
- `thinking_summary_as_str_mapping`
- `responses_body_summary_present_iff_effort_set`
- `responses_body_emits_configured_summary_mode`
- `responses_body_concise_summary_mode`
- `anthropic_thinking_config_adaptive_emits_display_summarized`
- `anthropic_thinking_config_manual_budget_emits_display_summarized`
- `test_acp_v2_chunks_carry_message_id` (integration test — two-prompt cross-session case)

## Notes

- **DBv2 gateway parity for `display`**: unverified — the DBv2 Claude route proxies Anthropic Messages shape, but whether the gateway passes `thinking.display` through is not confirmed. Flagged here rather than blocking on it.
- buzz-acp and Desktop TS are unchanged — they already parse `messageId` as optional and will pick it up from the wire automatically.
- Chat Completions and OpenRouter paths: untouched.

### DIFF
--- a/crates/buzz-agent/src/agent.rs
+++ b/crates/buzz-agent/src/agent.rs
@@ -460,6 +460,13 @@ impl RunCtx<'_> {
                 self.emit_usage_update().await;
             }
 
+            // Stable per-round message IDs for ACP v2 ContentChunk compliance.
+            // ACP v2 requires every ContentChunk to carry `messageId`; all chunks
+            // that belong to the same logical message must share the same ID.
+            // A provider round produces at most one thought and one assistant
+            // message, so one ID per round is the right granularity.
+            // ACP v1 allows the field, so this is a backwards-safe addition.
+            let round_msg_id = format!("round-{round}");
             if !response.reasoning.is_empty() {
                 wire::send(
                     self.wire,
@@ -467,6 +474,7 @@ impl RunCtx<'_> {
                         self.session_id,
                         json!({
                             "sessionUpdate": "agent_thought_chunk",
+                            "messageId": &round_msg_id,
                             "content": { "type": "text", "text": &response.reasoning }
                         }),
                     ),
@@ -481,6 +489,7 @@ impl RunCtx<'_> {
                         self.session_id,
                         json!({
                             "sessionUpdate": "agent_message_chunk",
+                            "messageId": &round_msg_id,
                             "content": { "type": "text", "text": &response.text }
                         }),
                     ),
--- a/crates/buzz-agent/src/agent.rs
+++ b/crates/buzz-agent/src/agent.rs
@@ -144,6 +144,12 @@ pub struct RunCtx<'a> {
     pub history: &'a mut Vec<HistoryItem>,
     pub original_task: &'a mut Option<String>,
     pub handoff_count: &'a mut usize,
+    /// ACP v2 session identifier for this prompt turn. Used to derive
+    /// per-message `messageId` values that are unique within the ACP session.
+    /// Distinct from `session_id` (which is the ACP session); this is a
+    /// per-`session/prompt` random token so that IDs from one prompt invocation
+    /// never collide with those from another even within the same session.
+    pub run_id: String,
     /// Cache-summed input tokens reported by the provider on this session's
     /// most recent request (persists across `session/prompt` calls), or `None`
     /// before the first response and immediately after a handoff resets the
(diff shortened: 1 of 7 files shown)


wesbillman (commented):
Reviewed on Wes's behalf at exact head `47ef696d6094d5fa4adcdc5a0a3dbe63de0a2727`. No blocking findings.

I traced all three changes through request construction, provider-response parsing, and ACP emission:

- OpenAI Responses: `reasoning.summary` is emitted only alongside configured `reasoning.effort`; `parse_responses` already consumes every `summary_text` item. Default `auto` is the provider-recommended way to select the most detailed summary supported by the model. Explicit `concise`/`detailed` remain operator choices because OpenAI documents model-dependent support.
- Anthropic: `display: "summarized"` is valid in both `type: "adaptive"` and `type: "enabled"` shapes. Anthropic's current thinking documentation explicitly supports both modes and explains that newer-model defaults otherwise return an empty `thinking` field. The field remains absent when this code does not enable/configure thinking.
- ACP v2: each emitted thought and assistant message gets a distinct ID, stable across its chunks and unique across rounds and prompt invocations because the prefix is the existing fresh per-prompt `run_id`. The two-prompt golden test pins the collision boundary that the earlier implementation missed.

The new env value is non-secret and its reveal allowlist/config parsing are covered. `git diff --check` passed. Full relevant CI is green (Rust Lint, Unit Tests, Desktop Core/build/smoke, Mobile, Windows Rust, cross-compiles, Security); two Desktop integration shards were still running when reviewed. The author-disclosed DBv2 Anthropic gateway passthrough uncertainty may affect whether that route gains observer text, but it does not create a regression in the native Anthropic path and is not a merge blocker.

Verdict: 9/10, safe to merge once required CI completes. This is a comment-only review, not an approval.

wesbillman (dismissed):

PR #4999, opened 2026-08-06
Relay-primary private managed-agent config

wesbillman:
## Summary

- replace the aggregate private-agent payload with a strict owner-authored encrypted kind:30179 codec
- retain public/private managed-agent heads and deletion tombstones atomically
- hydrate validated relay config into a workspace-scoped in-memory overlay, rehydrated from retained events at every boot so relay-primary config survives app restarts
- support fresh-device recovery: relay-only agents appear in the list with no local record; the first START intentionally materializes a local record (including the key material needed to run the agent) via `materialize_relay_only_agent` — this is the one deliberate disk write derived from relay state
- resolve the overlay into the local record before every stale-disk republish site (edit, persona rename, pair-start snapshot re-apply), so a device following a newer relay head cannot clobber it with an audit-clean successor event carrying stale fields
- hydrate keyring-stored agent keys in the boot reconcile path, so first boot on default (`system-keyring`) builds publishes each agent's 30179 instead of silently skipping every untouched agent
- preserve unknown forward-compatible fields and reject malformed/stale inbound events before changing the live overlay
- classify kind:30179 as owner-scoped global user data in the relay and subscribe from Desktop

This is the minimal replacement for the closed #4940 implementation.

### Review fixes (`c80c4c17`, `aa39d72a`, `6f486e88`)

Adversarial review (Eva/Sami) found and fixed five defects in the original head `a9b648b4`, each pinned by regression tests:

1. **Overlay never rehydrated after restart** — relay-primary config worked for exactly one launch, then every site silently read stale disk. Fixed: boot rehydration from the retention DB (owner-scoped; negative control asserts another owner's keys hydrate nothing).
2. **Boot reconcile published zero 30179s on default keyring builds** — records were read without keyring hydration, so the empty-nsec skip fired for every untouched agent. Fixed: hydrate keys in the reconcile path.
[...]

PR #5202, opened 2026-08-07
fix(desktop): retain distinct agent instances in autocomplete

atishpatel:
## Summary
- preserve each distinct agent pubkey in autocomplete even when agents share a persona or owner/name
- continue to collapse duplicate source rows for the same normalized pubkey
- show a truncated pubkey in the channel member-add picker so same-named instances are selectable

## Validation
- `pnpm --filter buzz test` — 4,489 passed
- `pnpm --filter buzz exec tsc --noEmit --pretty false`
- `pnpm --filter buzz exec biome check src/features/agents/lib/agentAutocompleteEligibility.ts src/features/agents/lib/agentAutocompleteEligibility.test.mjs src/features/channels/ui/MembersSidebar.tsx`
- independent validation by Fast Fizz on `509cb8d97b82f9708e24d4d59ad17c7b39516643`: typecheck, focused Biome, 22/22 focused tests, and `git diff --check`

Generated by Hardworking Honey.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -336,7 +336,7 @@ test("shouldHideAgentFromMentions: normalizes the pubkey before lookup", () => {
   );
 });
 
-test("coalesceAgentAutocompleteCandidates: merges agents with the same persona id", () => {
+test("coalesceAgentAutocompleteCandidates: keeps agents with the same persona id distinct", () => {
   const first = makeAgent({ pubkey: PUB_A, personaId: "pinky" });
   const second = makeAgent({
     pubkey: PUB_B,
@@ -344,10 +344,10 @@ test("coalesceAgentAutocompleteCandidates: merges agents with the same persona i
     isMember: true,
   });
 
-  assert.deepEqual(coalesce([first, second]), [second]);
+  assert.deepEqual(coalesce([first, second]), [first, second]);
 });
 
-test("coalesceAgentAutocompleteCandidates: merges agents with the same owner and name", () => {
+test("coalesceAgentAutocompleteCandidates: keeps agents with the same owner and name distinct", () => {
   const first = makeAgent({ pubkey: PUB_A, ownerPubkey: OWNER_PUBKEY });
   const second = makeAgent({
     pubkey: PUB_B,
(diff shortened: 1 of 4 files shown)


wesbillman:
**Thorough review by Carl on Wes’s behalf at exact head `509cb8d97b82f9708e24d4d59ad17c7b39516643`: this fix is needed, correctly scoped, and not covered by PR #4999. No code blocker found.**

The existing coalescer treats `personaId` or `(owner, display name)` as an agent identity. That is false for managed instances: two distinct pubkeys can intentionally share both. The result is loss, not cosmetic duplication—one real agent becomes impossible to select in @mention, new-DM, and channel member-add autocomplete. This patch changes only agent coalescing to normalized pubkey identity while retaining source-row deduplication for the same pubkey and the existing preference ranking for duplicate representations.

#4999 is complementary, not duplicative. It syncs/rehydrates private managed-agent configuration and enumerates/materializes relay-only records keyed by agent pubkey. It does not modify `agentAutocompleteEligibility.ts`, `MembersSidebar.tsx`, or any autocomplete coalescing call site; its frontend changes are only `usePersonaSync`. Indeed, #4999 can make this bug more visible on fresh devices by surfacing multiple legitimate relay records that the current frontend then collapses by shared persona/name. #5202 is independently useful on current main and remains necessary if #4999 lands.

I traced all three helper consumers: mention autocomplete, new-message recipients, and member-add. Candidate assembly already deduplicates repeated source rows by normalized pubkey before this helper; the helper’s remaining role is now correctly “merge duplicate representations of one pubkey,” not “guess that two keys are one agent.” Existing mention UI already displays truncated npubs on same-name collisions, new-DM rows expose the pubkey on hover, and this PR adds an always-visible truncated key for agent member-add rows, so retained duplicates are distinguishable. Eligibility, archival, channel-membership, and reachability gates remain unchanged.

Focused validation: `pnpm typecheck` passed; all **22/22** eligibility tests passed; Biome on all three changed files passed; `git diff --check` passed; worktree clean; remote head unchanged. CI is still running, so merge remains gated on exact-head CI and the PR leaving draft.

One non-blocking cleanup opportunity: `currentPubkey` and `getLabel` are now compatibility-only arguments and are intentionally underscored. Removing them from the helper API and its three callers would reduce dead conceptual surface, but it is not worth delaying this bug fix if the author prefers the smaller call-site diff.

wesbillman (approved):

PR #5203, opened 2026-08-07
fix(desktop): defer channel visibility change to Save

kchung:
## Problem

In the **Edit channel** dialog, flipping visibility (Public <> Private) persisted **immediately on selection**, bypassing the **Save changes** button — while every other field (name, description, temporary, TTL) waited for an explicit save. This surprised users and gave no chance to cancel a flip, e.g. a private->public change that instantly exposes channel history.

Reported in the Buzz "Welcome" channel by Kevin Chung.

## Root cause

The visibility dropdown was wired to `handleConvertVisibility()`, which called the update mutation on selection. This was intentional at the time (there was even an e2e test named `02 — visibility updates immediately` and an "Updating…" spinner), but it is inconsistent with the rest of the dialog and is the surprising behavior reported.

## Change (defer to Save)

- Visibility becomes a **deferred draft** like the other fields: selecting a value updates local `isPrivateDraft` and marks the draft dirty. The change commits via `handleSaveChannelEdits` (which already handled visibility) on **Save**, and is discarded on **Cancel**.
- The dialog title now reflects the **pending draft** (`nextVisibility`), so the pending choice is visible before saving.
- The edit-dialog reset restores `isPrivateDraft` from server state.
- Removed the now-dead `handleConvertVisibility` handler, `isConvertingVisibility` state, the `channelIdRef` race guard it needed, and the unused `isPending`/"Updating…" spinner path in `ChannelPermissionsSettings` (no caller passes `isPending` anymore).

## Tests

- Rewrote e2e `02` -> **`visibility defers to Save`**: select -> Save enabled -> title reflects draft -> Save -> persists; toggling back to the original value clears the draft and disables Save.
- Extended `09` (cancel discards drafts) to also cover a visibility change.
- Repurposed `10`: the stale-update race it guarded is architecturally gone, so it now asserts an **unsaved visibility draft does not leak across a channel switch**.

## Validation

- `pnpm typecheck` — clean
- `biome check` (changed files) — clean
- `pnpm test` — **4497 passed / 0 failed**
- `playwright test --project=smoke channel-controls` — **10 passed**

### DIFF
--- a/desktop/src/features/channels/ui/ChannelManagementSheet.tsx
+++ b/desktop/src/features/channels/ui/ChannelManagementSheet.tsx
@@ -124,8 +124,6 @@ export function ChannelManagementSheet({
   const deleteChannelMutation = useDeleteChannelMutation(channelId);
   const joinChannelMutation = useJoinChannelMutation(channelId);
   const leaveChannelMutation = useLeaveChannelMutation(channelId);
-  const channelIdRef = React.useRef(channelId);
-  channelIdRef.current = channelId;
 
   const detail = detailsQuery.data ?? channel;
   const members = React.useMemo(() => {
@@ -167,8 +165,6 @@ export function ChannelManagementSheet({
   );
   const [isDeleteDialogOpen, setIsDeleteDialogOpen] = React.useState(false);
   const [isEditDialogOpen, setIsEditDialogOpen] = React.useState(false);
-  const [isConvertingVisibility, setIsConvertingVisibility] =
-    React.useState(false);
   const [hasUserEditedChannelDraft, setHasUserEditedChannelDraft] =
     React.useState(false);
   const [activeView, setActiveView] = React.useState<"summary" | "canvas">(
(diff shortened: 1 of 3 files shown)


wesbillman:
**Release review by Carl on Wes’s behalf at exact head `55e2c3c911ddc52d29e56ee4bca2eb77b289b5af`: include this in the release once exact-head CI is green. The change is necessary, correct, and low risk. No blocker found.**

Necessity: this is not merely cosmetic consistency. Today selecting Private → Public immediately persists and can expose channel history even if the user cancels the dialog. The same dialog visibly presents one `Save changes` boundary for name, description, TTL, and channel type, so visibility bypassing it is a destructive surprise. Fixing it before release is worthwhile.

Correctness/safety: visibility was already part of `handleSaveChannelEdits`; this patch removes the parallel immediate-write path and feeds the existing draft/save transaction instead. Dirty detection, reverting to the original value, cancel reset, failed-save behavior, channel-switch resync, mutation cache updates, and create-channel behavior all remain coherent. The shared permissions component loses only a spinner prop used solely by the deleted immediate mutation; its create-channel caller was already draft-based and is unaffected. No backend/API/persistence format or authorization behavior changes. The patch also removes the stale async channel-id race entirely rather than weakening its guard.

Independent exact-head validation: `pnpm typecheck` passed; Biome passed on all three changed files; E2E build passed; focused channel-controls smoke suite **10/10 passed**; `git diff --check` passed; worktree clean; remote head unchanged. The test behavior changed intentionally: visibility now requires Save, Cancel discards it, and an unsaved draft cannot leak across a channel switch.

Release risk is low and bounded to Desktop’s Edit channel dialog. The security/product risk of *not* taking it is higher: an accidental private-to-public selection currently commits before confirmation. Merge only after repository CI is green on this exact head.

wesbillman (approved):

PR #5204, opened 2026-08-07
fix(desktop): preserve agent mention boundary in packaged WebKit

cmyk:
## Summary

Selecting a managed-agent mention in a packaged macOS build could leave ProseMirror's model and DOM selection correctly positioned after the trailing ASCII space, while WKWebView's next native text insertion crossed back into the decorated mention. Typing immediately after autocomplete could therefore change `@Reinhold ` into `@Reinholdtest`, remove the agent decoration, and silently drop the intended recipient tag.

This change:

- records authoritative metadata for the agent-prefix and agent-label decoration segments and exposes a query for the exact decorated range adjacent to the caret;
- intercepts only ordinary text insertion at a collapsed selection after a `U+0020` separator whose preceding range is authoritatively decorated as an agent mention;
- prevents the faulty native mutation and dispatches one ProseMirror transaction that preserves the separator and places the selection after the inserted character;
- leaves undecorated matching prose, marked text, split text nodes, composition/IME, paste, deletion, and unrelated typing on their existing paths;
- adds focused helper and mounted production-wiring coverage;
- adds a Chromium smoke test for the cross-browser composer separator-geometry invariant, while keeping native packaged-WKWebView acceptance as separate evidence.

### Related issue

No exact issue found.

Related but distinct open work:

- #2712 handles explicit caret placement and editing inside an existing mention chip.
- #2924 restores trailing whitespace lost during Markdown parsing.
- #5021 restores trailing whitespace after persistent-agent post-send content restoration.

This PR addresses the packaged-WKWebView native insertion immediately after autocomplete, where the model selection and separator were already correct.

### Testing

Automated verification at exact head `51cafdb0d55168aaffafc016bdf0acc585d780d6`:

- focused mention-boundary, decoration, and mounted production-wiring tests: **34 passed, 0 failed**;
- full Desktop unit suite: **4,491 passed, 0 failed**;
- E2E frontend build: passed, **4,697 modules transformed**;
- Chromium composer separator-geometry invariant: **1 passed, 0 failed**;
- TypeScript typecheck: passed;
- canonical frontend check: passed across **1,930 files**; one unrelated pre-existing warning and two unrelated pre-existing infos remain;
- changed-file Biome and repository helper checks: passed;
- `git diff --check`: passed;
- independent read-only review: approved.

Manual acceptance in an isolated packaged macOS candidate:

1. Select `@Reinhold` from autocomplete.
2. Type immediately without adding a space manually.
3. Before the fix, the caret appeared inside the mention and native insertion could extend the decorated name.
4. With the fix, typing `test` produces `@Reinhold test`; the mention remains decorated, the separator remains visible, and the caret ends after the inserted text.

#### Before
30fffbc0710d689bfc1a4c2df1cf87821327a09822745440791309e7c80687ab

#### After
c75154c9e8ab641bb7a4939879b42e31439386f50e8aef53cea97bcabb134616


cmyk:
Updated with `main` merged in.

Heads-up for reviewers: this branch was textually `MERGEABLE` but the merge result did **not** compile. Upstream added an `editorProps.handleDOMEvents.paste` handler (composer link paste) after this PR was opened, and this PR adds `handleDOMEvents.beforeinput`. Git merges both cleanly — they are in different places in the file — producing two `handleDOMEvents` keys in the same object literal:

- `tsc` fails with `TS1117: An object literal cannot have multiple properties with the same name`
- had it compiled, the second key would have silently replaced the first, breaking composer link paste at runtime

Neither the PR branch's own CI nor GitHub's mergeability check catches this, since each side is fine in isolation.

Resolved by combining both into a single handler object. `pnpm typecheck` and `pnpm check` pass on the merge result.

PR #5220, opened 2026-08-07
feat(buzz-acp): add bounded local event sink

Matweiss:
## Status
Repository-only, default-off local event sink for the CHANNEL-CONTINUITY-01 integration. No deployment, service restart, credential, runtime flag, live data, or unrelated behavior change.

## What I did
- Preserve the exact inbound Nostr event-object JSON from relay parsing, including handshake replay.
- Add an optional Unix-socket sink after the inbound author gate.
- Send one bounded length-prefixed frame containing only version, relay origin, channel UUID, and base64 exact event bytes.
- Require a one-byte acknowledgement under a shared connect/write/read deadline; failures are sanitized and do not alter ordinary agent dispatch.
- Keep the sink disabled unless `BUZZ_ACP_EVENT_SINK_SOCKET` is explicitly configured.

## Base and commit
- Reviewed base: `742e8d11974498a6ed71c70dd87dc6c25fcbf2b7`
- Candidate: `6ea717970ac1ba0388dc08571133e644aae81a3f`

## Tests
- `cargo test -p buzz-acp --no-fail-fast`: 696 unit tests and 9 lifecycle tests pass
- `just ci`: pass, including full workspace checks, web, desktop, and 1,260 mobile tests

## Security and rollback
The frame has an exact four-key schema and no private key, API token, auth tag, or other credential source. Paths and deadlines are bounded. Reverting this commit removes the optional sink without affecting relay state or data.

### DIFF
--- a/crates/buzz-acp/Cargo.toml
+++ b/crates/buzz-acp/Cargo.toml
@@ -40,7 +40,7 @@ reqwest = { workspace = true }
 
 # Serialization
 serde = { workspace = true }
-serde_json = { workspace = true }
+serde_json = { workspace = true, features = ["raw_value"] }
 
 # IDs
 uuid = { workspace = true }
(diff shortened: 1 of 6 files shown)


Matweiss:
@block/buzz-oss-team Review requested for this default-off `buzz-acp` local event sink. DCO, Semgrep, zizmor, focused crate tests, and repository-wide `just ci` are green. The change is limited to six `buzz-acp` files and transmits only the exact signed event plus public relay/channel provenance; it carries no credential material.

PR #5221, opened 2026-08-07
fix(desktop): isolate relay admission tests

wesbillman:
## Summary

- serialize the relay error-message test with all other tests mutating the process-wide admission gate
- clear its 300-second rate-limit expiry after the assertion
- prevent the paused-time waiter test from observing another test's state

## Root cause

`relay::tests::oversized_hint_is_capped_in_relay_error_message_string` arms the process-wide gate for 300 seconds without taking `TEST_SERIAL` or resetting it. In a parallel test run, `relay_admission::tests::concurrent_429_extends_the_window_for_parked_waiters` can observe that expiry, producing the reported `300.001s` instead of `5s`.

## Validation

- focused admission suite + relay error test repeated 10 times
- pre-push `desktop-tauri-checks` passed, including the full Rust workspace suite
- `branch-skew` passed

### DIFF
--- a/desktop/src-tauri/src/relay.rs
+++ b/desktop/src-tauri/src/relay.rs
@@ -652,8 +652,11 @@ mod tests {
 
     #[tokio::test]
     async fn oversized_hint_is_capped_in_relay_error_message_string() {
-        use crate::relay_admission::MAX_HINT_SECONDS;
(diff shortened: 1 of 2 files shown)


anshkrsingh457-cyber:
@Ansh Singh tain

anshkrsingh457-cyber:
Do not #fuck 🤡 me because your life is #faka_fuck from every side🤌😩

PR #5223, opened 2026-08-07
Recover from max-token response truncation

tlongwell-block:
## Summary

- treat provider `max_tokens` as an interrupted assistant response and continue the same turn with actionable feedback
- discard tool calls from truncated responses, including malformed partial arguments, so they are neither executed nor replayed with invalid tool-result pairing
- bound recovery to two retries while preserving normal finite `max_rounds` accounting

## Verification

- `cargo test -p buzz-agent` (422 unit tests plus all package integration/doc suites passed)
- `cargo clippy -p buzz-agent --all-targets -- -D warnings`

## Notes

The pre-push repository-wide hook also ran. Its Rust tests passed (2,270 passed, 14 ignored), but its `buzz-db` unit-test build was blocked because local rustc 1.89 is below sqlx 0.9's rustc 1.94 requirement. The affected package suite above is green on the exact pushed commit.

Originating Buzz channel: `c3252dd2-0142-4e01-88c7-a2183c3960a5`

### DIFF
--- a/crates/buzz-agent/src/agent.rs
+++ b/crates/buzz-agent/src/agent.rs
@@ -24,6 +24,17 @@ const ERROR_REFLECTION_SUFFIX: &str =
 
 const UNSUPPORTED_IMAGE_TOOL_MESSAGE: &str = "The current model does not support image input. The image was removed from conversation history so this turn can continue. Use a text-based inspection tool or ask the user for a textual description instead.";
 
+/// Model-visible feedback after the provider truncates an assistant response at
+/// its output-token limit. This is a user message rather than a synthetic tool
+/// result because truncation can happen without a tool call (and an unpaired
+/// tool result is invalid on every provider wire format).
+const MAX_TOKENS_RECOVERY_MESSAGE: &str = "Your previous response exceeded the model's output token limit and was truncated. Any incomplete tool call was not run. Continue the task, breaking the work or tool call into smaller steps and keeping the response concise.";
+
+/// A provider can repeatedly spend its entire output allowance without making
+/// progress, while `max_rounds` is unbounded by default. Keep the in-turn rescue
+/// finite so a persistently truncating model eventually surfaces `max_tokens`.
+const MAX_TOKENS_RECOVERIES_PER_RUN: u32 = 2;
+
 /// Remove image blocks that the provider has explicitly rejected while keeping
 /// their surrounding tool result (and therefore the tool-call/result pairing)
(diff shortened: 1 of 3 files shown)


tlongwell-block (commented):
Reviewed the full diff at head `47b07751e0acfed9221d3f494ee7b0e86e8ddfda`. (GitHub blocks a formal approval because our agents share the bot account, so this comment is the review of record.)

**Design**: correct placement of the recovery. `max_tokens` is a provider stop on the *assistant response*, not a tool failure, so a user-role continuation message is the only history shape that stays valid on all three wire formats (an unpaired tool_result is rejected everywhere). Discarding *all* tool calls from a truncated response — including ones that happen to parse — is the right call: a later call or surrounding instruction may have been cut off, and the parser changes (`parse_openai`, `parse_anthropic`, `parse_responses`) consistently skip rather than reject partial/malformed arguments, which previously would have errored the whole turn.

**Bounds**: two in-turn recoveries, non-refunded rounds, and the exhausted path surfaces the original `StopReason::MaxTokens` — so a persistently truncating model still terminates, and finite `max_rounds` accounting is unchanged (`max_tokens_recovery_respects_finite_round_cap` proves the cap wins).

**Independent verification at this exact head** (`git rev-parse HEAD` confirmed in the same shell):
- `cargo test -p buzz-agent` — full package suite green, including the three new regressions
- `cargo clippy -p buzz-agent --all-targets -- -D warnings` and `cargo fmt --all -- --check` — clean
- **Live pass against the real Anthropic API** (`claude-haiku-4-5`, `BUZZ_AGENT_MAX_OUTPUT_TOKENS=150`, ACP driven over stdio): a genuinely truncated response produced the two expected `provider hit output token limit; asking model to continue in smaller steps` recoveries and the turn finished with `end_turn` instead of dying. Repeated with an adversarial always-truncate prompt: recovered twice, still ended cleanly.

Edge cases I checked and am satisfied on: empty-text truncated assistant messages are dropped by the Anthropic serializer (no empty-block 400); consecutive user messages after that drop are accepted by the API; no replayed `tool_use` means no thinking-block pairing constraint; `_Stop` gating still applies only to genuine `end_turn`.

One non-blocking note: when the recovery budget is exhausted the run still returns `max_tokens`, which the solo Harbor adapter currently reads as trial-complete — that adapter-side mapping is the remaining piece of the P0 and is correctly out of scope here.

Verdict: **approve** — 9+ on minimalness, elegance, and correctness. Ready to merge.

atishpatel (approved):

PR #4913, opened 2026-08-05
fix(desktop): allow shared agent mentions

wesbillman:
## Summary

- admit relay-discovered agents to autocomplete when their response policy authorizes the viewer
- require authorization in the exact active stream/forum channel for mentions, while keeping community-wide discovery for member invitation
- fail closed for relay-only agents in DMs and unresolved composer contexts
- re-authorize cached autocomplete rows after policy/channel changes so stale agent suggestions cannot leak back in
- preserve managed-agent behavior and explicitly reject stale agent-marked channel members absent from both live directories

## Validation

- `pnpm --dir desktop test` — 4,288 passed
- `pnpm --dir desktop typecheck`
- `pnpm --dir desktop check`
- `pnpm --dir desktop build:e2e`
- focused Playwright mention matrix — 12 passed
- focused Playwright member-invitation matrix — 2 passed
- pre-push hooks after rebase to current `origin/main` — desktop check and 4,288 tests passed
- independent correctness/privacy re-review cleared with no remaining blocker

## Related competing PRs

This supersedes or overlaps #2333, #3056, #4242, #4137, #2314, #4058, and #2605. This version adds exact-channel authorization, fail-closed DM/context handling, cached-row reauthorization, forum coverage, outbound mention-tag coverage, explicit stale-member coverage, and add-member discovery coverage.

PR #5227, opened 2026-08-07
feat(desktop): surface externally managed agents

kvncrw:
## What changed

Adds an **External agents** section to the Desktop Agents page. It discovers relay-registered agents whose NIP-OA owner is the current Desktop identity and excludes locally managed agents.

## Why

Kubernetes/external agent identities are operationally real but were invisible in the Agents panel, making the desktop list misleading.

## Safety

External cards are read-only: they open the existing profile/runtime details and deliberately do not expose local start/stop, lifecycle, or private-key controls.

## Validation

- `pnpm --dir desktop exec tsc --noEmit` passed.
- `git diff --check` passed.
- Biome could not run because the repository currently rejects its nested `desktop/biome.json` configuration before processing the target file.


kvncrw:
Closing: this approach caused a permission-approval regression in local testing (tool approvals silently denied estate-wide on the install built from this branch), and the underlying discovery problem is better solved by the kind:10100 agent-announcement path (#4913). May revisit against a test install with an explicit tool-approval acceptance check.

PR #5230, opened 2026-08-07
feat(buzz-dev-mcp): declarative read-only HTTP endpoints (http_get)

manuarraya:
## Summary

- Add an `http_get` MCP tool backed by a JSON manifest (`BUZZ_DEV_MCP_HTTP_MANIFEST`).
- Operators declare GET-only endpoints with path templates, validated placeholders, query-param allowlists, and at most one auth header whose value is read from an env var per call.
- The endpoint catalogue is advertised through server instructions (tool descriptions stay static macro literals); anything not in the manifest does not exist to the tool.
- A manifest that is set but fails to load disables `http_get` **loudly** in the instructions instead of silently looking unconfigured.

Lets an operator hand agents a curated read-only slice of an internal API without teaching them curl or pasting credentials into prompts.

## Context

Found in a production fleet: agents needed a safe read lane into an internal control plane. Pair with `BUZZ_ACP_MCP_ENV` (separate PR) so the harness actually forwards the manifest path and credential env names into the MCP process after `env_clear()`.

## Test plan

- [ ] `cargo test -p buzz-dev-mcp`
- [ ] Point `BUZZ_DEV_MCP_HTTP_MANIFEST` at a small fixture and confirm instructions list the endpoints
- [ ] Confirm a missing/broken manifest surfaces a loud disable message rather than silent absence
- [ ] Confirm auth header values come from env at call time (rotation without restart)

### DIFF
--- a/crates/buzz-dev-mcp/src/http_endpoints.rs
+++ b/crates/buzz-dev-mcp/src/http_endpoints.rs
@@ -1,0 +1,408 @@
+//! Declarative read-only HTTP endpoints — the `http_get` tool.
+//!
+//! An operator can hand the agent a curated, read-only slice of an internal
+//! HTTP API without teaching it curl, without pasting credentials into
+//! prompts, and without forking this crate per deployment: a JSON manifest
+//! (path in `BUZZ_DEV_MCP_HTTP_MANIFEST`) names the endpoints, and this
+//! module fetches them. GET only, allowlist only — an endpoint that is not
+//! in the manifest does not exist as far as the tool is concerned.
+//!
+//! The manifest is deployment-specific and lives with the operator, not
+//! here. Auth is a single header whose VALUE is read from an env var at
+//! call time, so a rotated credential is picked up without a restart and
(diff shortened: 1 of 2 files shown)


wolfyy970:
This is a useful target-local adapter, but I found two security blockers in the current shape.

`validate` accepts authenticated `http://` base URLs, so the configured header can be sent in cleartext. #4164 uses the safer boundary: HTTPS, with credential-free literal loopback as the only HTTP exception.

`resp.text().await` buffers the entire response before the 8 KB display truncation. A slow or compromised endpoint can therefore exhaust the MCP process with an unbounded body. The limit needs to apply while reading, not after.

One product boundary is also worth making explicit. Read-only does not mean safe for model context. The response can still contain customer data, tokens or unredacted logs. This fits as a Project Connection provider only if setup shows the endpoint and data scope before an agent receives it. It should not travel inside a portable agent.

The stated credential rotation does not work through inherited process environment either. A parent cannot change an existing child process environment, so rereading the variable per call will not observe an external rotation without another credential channel or a restart.

PR #5231, opened 2026-08-07
feat(buzz-acp): forward operator-named env vars to MCP servers

manuarraya:
## Summary

- Add `BUZZ_ACP_MCP_ENV`: a comma-separated list of **variable names** to forward from the harness environment into each MCP server's declared env.
- Values are read from the process (names only in config), so secrets never need to be written into a config file to reach the server.
- Unset or empty names are skipped — listing a variable you have not set behaves like "not configured" rather than shipping an empty value.

## Why

`buzz-agent` spawns every MCP server with `env_clear()` plus a fixed allowlist, so anything an operator configures for a particular deployment — an endpoint manifest, the credential that reads it — never reaches the server. The only existing escape hatch was adding the variable name to that allowlist (hardcoding one deployment's vocabulary into an OSS crate).

Found by deploying it: an agent answered a live question with `http_get no está configurado (BUZZ_DEV_MCP_HTTP_MANIFEST unset)` while that variable was exported in its own systemd unit — the harness had no way to pass it on.

Pairs with the `http_get` PR (`feat/buzz-dev-mcp-http-get`) but is independently useful for any operator-named MCP env.

## Test plan

- [ ] `cargo test -p buzz-acp`
- [ ] Set `BUZZ_ACP_MCP_ENV=BUZZ_DEV_MCP_HTTP_MANIFEST,SOME_TOKEN` and confirm the MCP child sees both when the parent has them
- [ ] Confirm an unset name in the list does not inject an empty value
- [ ] Confirm variables not listed are still blocked by `env_clear()`

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -4560,9 +4560,101 @@ fn build_mcp_servers(config: &Config) -> Vec<McpServer> {
                     });
                 }
             }
+            env.extend(operator_forwarded_env(
+                std::env::var(MCP_ENV_FORWARD_VAR).ok().as_deref(),
+                |name| std::env::var(name).ok(),
+            ));
             env
         },
     }]
+}
+
+/// Names the operator wants forwarded from this process to the MCP server.
+pub const MCP_ENV_FORWARD_VAR: &str = "BUZZ_ACP_MCP_ENV";
+
+/// Resolve `BUZZ_ACP_MCP_ENV` (a comma-separated list of variable NAMES) into
+/// the entries to append to the MCP server's declared environment.
+///
+/// The agent spawns MCP servers with `env_clear()` plus a fixed allowlist, so
+/// anything an operator configures for a specific deployment — an endpoint
+/// manifest, a credential for it — is dropped unless it is declared here. The
+/// alternative was hardcoding deployment-specific variable names into the
+/// allowlist of an OSS crate, which does not scale past the first deployment.
+///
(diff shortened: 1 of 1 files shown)


wolfyy970:
The deployment gap is real, but I do not think this should become the permanent secret handoff.

`BUZZ_ACP_MCP_ENV` is currently writable through an agent environment map because Desktop does not reserve it. That lets agent or persona configuration choose which parent-process values are copied into the privileged MCP child. On current main, those arbitrary values also travel through `session/new` without the redaction coverage added in #4164.

The list is process-wide rather than connection-specific, so it cannot express that one server receives one credential while another does not.

I would build this on #4164 instead. A target-local Project resolver can read approved credential sources and place final values only in the named server entry. The portable agent carries the capability requirement, never the environment name or secret.

If this compatibility input remains, I think it needs to be device-owned and reserved from agent/persona env, with the same redaction and protected-key rules as the structured handoff. Otherwise Buzz gains a second secret configuration surface that will be difficult to remove.

whyihaveyou:
Hi — sharing a related PR in case it is useful context: #5321 adds a runtime-agnostic, read-only model of external agent frameworks' MCP server config (Hermes + Kimi Code adapters), and it models per-server `env` as part of the server entry (masked in logs via a `redacted()` helper so secrets never leak).

Our current slice is strictly read-only, so this is not a direct implementation of the env-forwarding you are asking for here — but a shared server-config model is the natural place this could build on later. Just cross-linking for visibility so anyone working on this issue is aware of the model exists. Feedback welcome either way.

PR #5232, opened 2026-08-07
fix(desktop): let channel-member agents through the allowed-list mention gate

manuarraya:
## Summary

- Channel **member** agents now bypass the allowed-list gate in `@` autocomplete.
- `shouldHideAgentFromMentions` still decides invocability (Option B): member bots with unknown/allowed invocability stay mentionable.
- Non-member directory noise remains filtered by the allowed-list check.

## Why

Self-hosted ACP agents added as channel members (`role=bot`) were dropped from autocomplete by the allowed-list gate before Option B could run. Stock Desktop could not mention them even though they were channel members.

Ported onto current main after the rename from `isAgentIdentityInManagedList` / managed-list to `isAgentIdentityInAllowedList` / `mentionableAgentPubkeys`.

## Test plan

- [ ] `cd desktop && node --test src/features/agents/lib/agentAutocompleteEligibility.test.mjs src/features/messages/lib/useMentions.test.mjs` (or the project's equivalent)
- [ ] Add a non-managed agent as a channel member and confirm it appears in `@` autocomplete
- [ ] Confirm non-member directory agents still require being on the allowed/mentionable set

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.ts
@@ -82,6 +82,18 @@ export function getMentionableAgentPubkeys({
   return pubkeys;
 }
 
+/**
+ * Gate for *non-member* agent identities in autocomplete / add-member search.
+ *
+ * Returns true when the candidate should stay in the pipeline:
+ * - humans always pass
+ * - agents pass only if they are in the caller's allowed/mentionable set
+ *
+ * Channel *members* that are agents are handled separately by the caller
+ * (see `useMentions`): they bypass this gate so `shouldHideAgentFromMentions`
+ * can apply Option B invocability (member bots with unknown/allowed
+ * invocability remain mentionable — required for self-hosted ACP agents).
+ */
 export function isAgentIdentityInAllowedList(
   candidate: { isAgent?: boolean; pubkey: string },
   allowedAgentPubkeys: ReadonlySet<string>,
--- a/desktop/src/features/messages/lib/useMentions.ts
+++ b/desktop/src/features/messages/lib/useMentions.ts
@@ -255,7 +255,15 @@ export function useMentions(
       if (isArchivedDiscovery(pubkey)) {
         return;
       }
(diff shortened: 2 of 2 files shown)


ScaleLeanChris:
Consumer validation on head `b7e6005b25b5f8bbcfdf75c3214457c9c9e14ec7`:

- `pnpm --dir desktop build` passed, including TypeScript and the production Vite bundle.
- Biome passed on both changed files.
- `git diff --check` passed for the two-file patch.
- Manual path review confirms channel members enter `addCandidate` with `isMember: true`, bypass only the allowed-list gate added here, and still pass through `shouldHideAgentFromMentions`. Non-member agents still require membership in `mentionableAgentPubkeys`.

I also attempted an isolated full Tauri launch. It reached 845/848 Rust crates, then the temporary checkout exhausted local disk space. That is an environment limitation, not a candidate failure, but I did not mark the PR's live Desktop test plan complete.

No source or test changes were made. This looks correct at the changed boundary, but a successful live `@external-agent` autocomplete check is still the remaining product proof. For now, the member agent is back in the pipeline and can say, in the words of The Who, *Who Are You*.

ScaleLeanChris:
Overlap note after reviewing the current heads of #5232, #5546, and #5549:

- #5232 and #5549 both repair the same Desktop fallback for a channel-member agent whose directory policy is not yet available. #5549 now carries the stronger cross-owner E2E proof.
- Commit 2 of #5546 also changes that same fallback. A review on #5546 has already asked that it be separated from the directory producer.
- Commit 1 of #5546 is a different concern: producing kind:10100 records from `buzz-acp`. It still needs resolution of the public allowlist exposure and profile-writer race raised in review.

The cleanest path seems to be one consumer fallback PR, likely #5549 because it includes the current E2E case, plus a separately reviewed producer change from #5546. That avoids asking maintainers to reconcile three versions of the same Desktop behavior.

For completeness, the earlier build, Biome, and boundary validation I posted here remains valid for head `b7e6005b25b5f8bbcfdf75c3214457c9c9e14ec7`. I am pausing duplicate testing on this version unless a maintainer prefers it.

PR #5242, opened 2026-08-07
fix(desktop): externalize boot <style> to prevent Tauri CSP nonce override

wpfleger96:
## Problem

Two v0.5.6-only regressions were introduced by #4614 (the first enforced Tauri CSP):

1. **Tab-complete caret regression** — after tab-completing an @mention, #channel, or :emoji: shortcode, the cursor landed inside the inserted text instead of after the trailing space. TipTap inserts the correct text including the trailing space, but without its base stylesheet (`.ProseMirror { white-space: break-spaces }`) the trailing space collapses visually and the caret appears mid-name.

2. **Emoji picker unstyled** — the emoji-mart picker rendered as a giant unstyled layout (oversized search SVG, collapsed grid) because emoji-mart's shadow-root stylesheet injection was also blocked.

Both symptoms have the same root cause.

## Root Cause

Tauri's build-time asset processor scans `index.html` for inline `<style>` elements, injects a nonce token, and adds the corresponding `'nonce-…'` source to `style-src` at runtime. Per the CSP spec, **once a nonce is present in a directive, the browser ignores `'unsafe-inline'` for that directive**.

`index.html` contained an inline `<style>` with the boot background color. When Tauri nonced it and injected `'nonce-…'` into `style-src`, the intended `style-src 'self' 'unsafe-inline'` became effectively `style-src 'self' 'nonce-…'` — blocking any runtime stylesheet injection not covered by a matching nonce:

- TipTap's `injectCSS()` → `createStyleTag()` injecting `.ProseMirror { white-space: break-spaces; … }`
- emoji-mart's shadow-root `document.createElement('style')` injection

(Inline scripts follow a separate path — they are SHA-256 hashed, not nonced.)

This only reproduces in packaged builds (where Tauri's custom protocol serves the HTML and enforces the policy). `tauri dev` loads from the Vite dev server and is not affected.

## Fix

Move `html { background-color: #000; }` from an inline `<style>` in `index.html` to `desktop/public/boot.css`, linked via `<link rel="stylesheet">`. A linked stylesheet is not subject to Tauri's nonce injection, so `'unsafe-inline'` in `style-src` applies as declared.

The `<link>` is render-blocking (same as the inline style was), so boot-flash behaviour is identical.

**The production CSP string is unchanged.** This fix makes the policy apply as intended — no security properties are altered. Will's follow-up with the security team (Jordan Mecom / Eli Foster, authors of #4614) is noted for post-ship.

A Tauri-faithful CSP harness for the Vite dev path (so this class of regression is visible before a packaged build) is tracked as a separate follow-up.

## Files Changed

- `desktop/index.html` — replace inline `<style>` with `<link rel="stylesheet" href="/boot.css" />`
- `desktop/public/boot.css` — new file, the extracted `html { background-color: #000; }` plus rationale comment
- `desktop/src-tauri/tests/csp.rs` — update comment: nonce for styles, SHA-256 for the boot script

## Testing

- `just desktop-typecheck` ✅
- `just desktop-test` ✅ (4535/4535)
- `just desktop-tauri-test` ✅ (all Rust tests including `csp.rs`)
- Packaged validation: `pnpm tauri build --debug` completed; compiled binary bakes `style-src 'self' 'unsafe-inline'` with no nonce source injected ✅

### DIFF
--- a/desktop/index.html
+++ b/desktop/index.html
@@ -12,6 +12,13 @@
       in tauri.conf.json. The body paints its own themed background once the app
       CSS loads, so this never shows through after boot.
 
+      The linked stylesheet (boot.css) applies the initial black background.
+      Keeping it as a <link> rather than an inline <style> prevents Tauri's
+      build-time asset processing from adding a 'sha256-…' source to style-src:
+      per the CSP spec a hash in a directive makes the browser ignore
+      'unsafe-inline', which would block TipTap's runtime stylesheet injection
+      and emoji-mart's shadow-root styles in packaged builds.
+
       The inline script below reads the cached theme background (same
       `buzz-theme-cache` entry ThemeProvider writes) and applies it synchronously
       so the boot color matches the themed loading gate — no black flash on light
@@ -20,11 +27,7 @@
       that ThemeProvider applies moments later — no wrong-scheme flash before
       it loads.
     -->
-    <style>
-      html {
-        background-color: #000;
-      }
-    </style>
+    <link rel="stylesheet" href="/boot.css" />
     <script>
       (() => {
         var cached, bg, parsed;
--- a/desktop/index.html
+++ b/desktop/index.html
@@ -13,11 +13,8 @@
       CSS loads, so this never shows through after boot.
 
       The linked stylesheet (boot.css) applies the initial black background.
-      Keeping it as a <link> rather than an inline <style> prevents Tauri's
-      build-time asset processing from adding a 'sha256-…' source to style-src:
-      per the CSP spec a hash in a directive makes the browser ignore
-      'unsafe-inline', which would block TipTap's runtime stylesheet injection
-      and emoji-mart's shadow-root styles in packaged builds.
+      Kept as a <link> rather than an inline <style> to avoid Tauri's nonce
+      injection for style-src (see boot.css for the full rationale).
 
       The inline script below reads the cached theme background (same
       `buzz-theme-cache` entry ThemeProvider writes) and applies it synchronously
--- a/desktop/public/boot.css
+++ b/desktop/public/boot.css
@@ -5,10 +5,13 @@
  * state is available; on first-ever launch it falls back to this black default.
  *
  * Kept as a linked stylesheet (not an inline <style>) so that Tauri's
- * build-time asset processing does not add a 'sha256-…' source to style-src.
- * Per the CSP spec a hash or nonce in a directive causes the browser to ignore
- * 'unsafe-inline', which would block TipTap's runtime stylesheet injection and
- * emoji-mart's shadow-root styles in packaged builds.
+ * build-time asset processing does not inject a nonce token into this element.
+ * Tauri nonces inline <style> elements and adds the corresponding 'nonce-…'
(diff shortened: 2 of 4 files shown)


wesbillman:
**Independent release verdict by Carl on Wes’s behalf at exact head `3fdb8211e15ebdf6498932e984740324e4825e81`: ship this hotfix. The root-cause diagnosis is correct, the patch fixes both reported packaged-build regressions, and I found no code blocker.**

I independently traced the failure through the versions actually locked by this head:

- Tauri codegen 2.6.3 calls `inject_nonce_token`; tauri-utils 2.9.3 adds `__TAURI_STYLE_NONCE__` to every inline `<style>`; Tauri 2.11.5 replaces it per request and appends a `'nonce-…'` source to `style-src`.
- Under CSP, that nonce causes `'unsafe-inline'` to be ignored for the directive. I reproduced the browser behavior directly in Chromium: `style-src 'self' 'unsafe-inline' 'nonce-abc'` accepts the nonced boot style but blocks a subsequently created unnonced `<style>`; removing the style nonce lets the same runtime injection apply.
- TipTap 3.22.5 defaults `injectCSS: true` and dynamically creates an unnonced style containing `.ProseMirror { white-space: break-spaces; }`. Buzz has no equivalent static rule. Blocking it explains the visually collapsed trailing space/caret after autocomplete.
- emoji-mart 5.6.0 dynamically creates its base shadow-root `<style>`, and Buzz adds a second unnonced shadow-root override in `useEmojiMartStyles`. Blocking both explains the giant, unstyled picker.

The fix removes the only real inline `<style>` from the built HTML, keeps the boot background in render-blocking same-origin `/boot.css`, and leaves the configured CSP string unchanged. A Vite production build contains the linked `boot.css`, no actual inline style element, and the existing inline boot script remains on Tauri’s separate SHA-256 path. The causal timing also matches: `index.html` had this style before #4614, while #4614 first enabled enforcement in v0.5.6.

Security precision: this **does** restore the effective `style-src 'unsafe-inline'` behavior that the checked-in policy explicitly requested; compared with broken v0.5.6, runtime inline styles become allowed again. That is necessary for the currently shipped TipTap/emoji-mart architecture, not an accidental broadening beyond the declared policy. The promised security-team follow-up is still appropriate if the product wants to remove `'unsafe-inline'` later.

Validation at this exact head: typecheck passed; Biome passed on changed web assets; Vite production build passed and emitted `/boot.css`; direct Chromium CSP reproduction passed; `git diff --check` passed; clean worktree; remote head unchanged. My local `cargo test --test csp` could not reach tests because the review worktree lacks required Tauri sidecar binaries, but exact-head CI has Desktop Core/build/RC, Rust lint, Windows Rust, all smoke/integration shards, DCO, Semgrep, and zizmor green.

Non-blocking coverage gap: `csp.rs` only updates explanatory text; it does not assert that packaged `index.html` remains free of inline styles or that effective runtime styles work. The PR explicitly tracks a Tauri-faithful harness follow-up. Given the verified packaged build and severity of the regressions, do not delay the hotfix for that broader harness.

wesbillman (approved):

PR #5140, opened 2026-08-07
fix(cli): dms list queries relay NIP-29 discovery, not phantom kind:41001

iroiro147:
## Problem
`buzz dms list` returns `[]` even for agents with active DMs. Reported in #5137: an agent's DM sat unanswered for ~20 minutes because `dms list` couldn't discover it (though `buzz feed get` showed the events).

## Root cause
`cmd_list_dms` queried `kind:41001` (`KIND_DM_CREATED`). However, **no upstream event of that kind is ever published** — the enum constant exists in `buzz-core/src/kind.rs` but no relay handler mints one. The query always returned an empty page.

The relay's actual DM discovery surface is the standard NIP-29 group-discovery emission: on `open_dm`, `emit_group_discovery_events` (crates/buzz-relay/src/handlers/side_effects.rs:1045) publishes kind:39002 (membership) and kind:39000 (channel metadata with `t: "dm"` tag).

## Fix
Rewrite `cmd_list_dms` to use the published discovery events:
1. Query kind:39002 `#p: [our_pubkey]` for our memberships → collect channel IDs.
2. Query kind:39000 `#d: [channel_ids]` → filter to `t == "dm"` → extract participants + created_at.

This matches `channels list --member` (channels.rs:33-64), keeping the CLI consistent with the relay's authority model.

## Validation
- [x] `cargo check -p buzz-cli` clean
- [x] `cargo fmt -p buzz-cli` clean
- [x] `cargo clippy -p buzz-cli --all-targets -- -D warnings` clean
- [x] `cargo test -p buzz-cli` — 321 passed, 0 failed

Closes #5137.

PR #5247, opened 2026-08-07
fix(cli): query NIP-29 group metadata instead of non-existent kind:41001 for DM list

### LINKED ISSUE #5137: buzz dms list returns empty for an active DM conversation with messages

BradGroux:
## What was wrong

`buzz dms list` always returned an empty array, even for active DM conversations with messages. Agents relying on this command to discover DMs would silently miss messages sent directly to them.

The command queried kind:41001 (`KIND_DM_CREATED`) filtered by the caller's `#p` tag. However, the relay never emits kind:41001 events. When a DM is opened via `handle_dm_open`, the relay creates the channel in the database and emits NIP-29 group discovery events (kind:39000/39001/39002), not kind:41001. The query had zero matching events to return.

## Why it happened

PR #510 (`fix(cli): normalize output and fix wrong Nostr kinds across all commands`) corrected the previous wrong kind (41010) to 41001, based on the `KIND_DM_CREATED` constant. But the relay doesn't actually emit that kind — it only defines it. The relay's DM creation path (`handle_dm_open` → `emit_group_discovery_events`) emits kind:39000 (group metadata) with a `["t", "dm"]` tag and `["hidden"]` tag, plus kind:39002 (group members) with `p` tags for each participant. There is no code path that emits kind:41001.

## What changed

Switched `cmd_list_dms` to the same two-step pattern already used by `cmd_list_channels --member`:

1. Query kind:39002 (group members) filtered by `#p` with our pubkey to find channel IDs where we're a member.
2. Fetch kind:39000 (group metadata) for those channel IDs.
3. Filter for DM channels by checking for the `["t", "dm"]` tag.

The output format (`dm_id`, `participants`, `created_at`) is unchanged, so callers that already parse the output need no changes.

## Verification

```text
cargo check -p buzz-cli
cargo test -p buzz-cli --lib
cargo fmt -p buzz-cli -- --check
cargo clippy -p buzz-cli --all-targets -- -D warnings
```

All 341 `buzz-cli` tests passed, formatting is clean, and Clippy reports no warnings.

Closes #5137.

### DIFF
--- a/crates/buzz-cli/src/commands/dms.rs
+++ b/crates/buzz-cli/src/commands/dms.rs
@@ -4,19 +4,62 @@ use crate::client::{extract_d_tag, normalize_write_response, BuzzClient};
 use crate::error::CliError;
 use crate::validate::{parse_uuid, sdk_err, validate_hex64};
 
-/// List DM conversations by querying kind:41001 (relay-confirmed DMs) filtered by our pubkey.
+/// List DM conversations by querying NIP-29 group metadata (kind:39000) for
+/// channels we are a member of, then filtering for DM channels.
+///
+/// The relay emits kind:39000 with a `["t", "dm"]` tag and `["hidden"]` tag for
+/// DM channels. We discover our memberships via kind:39002 (group members)
+/// filtered by our pubkey, then fetch the metadata for those channel IDs.
 pub async fn cmd_list_dms(client: &BuzzClient, limit: Option<u32>) -> Result<(), CliError> {
     let my_pk = client.keys().public_key().to_hex();
-    let limit = limit.unwrap_or(50).min(200);
-    let filter = serde_json::json!({
-        "kinds": [41001],
+    let effective_limit = limit.unwrap_or(50).min(200);
+
+    // Step 1: find channel IDs where we're a member (kind:39002).
+    let member_filter = serde_json::json!({
+        "kinds": [39002],
         "#p": [my_pk],
-        "limit": limit
     });
-    let resp = client.query(&filter).await?;
-    let events: Vec<serde_json::Value> = serde_json::from_str(&resp).unwrap_or_default();
+    let member_events = client
+        .query_paginated(member_filter, effective_limit)
+        .await?;
+    let channel_ids: Vec<String> = member_events
+        .iter()
+        .map(extract_d_tag)
+        .filter(|id| !id.is_empty())
+        .collect();
+    if channel_ids.is_empty() {
+        println!("[]");
+        return Ok(());
+    }
+
+    // Step 2: fetch kind:39000 metadata for those channels.
+    let metadata_filter = serde_json::json!({
+        "kinds": [39000],
+        "#d": channel_ids,
+    });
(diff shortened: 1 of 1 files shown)


BradGroux:
This is the third recurring review + rebase pass for the open BradGroux PRs on `block/buzz`.

## What this PR does

Fixes the `buzz dms list` CLI command to query NIP-29 group metadata (kind:39000) instead of the non-existent kind:41001. The old code queried kind:41001 filtered by pubkey, which never returned results because that kind doesn't exist on the relay.

The new implementation is a three-step pipeline:
1. Query kind:39002 (group members) filtered by `#p` with our pubkey to discover channel IDs where we're a member.
2. Query kind:39000 (group metadata) for those channel IDs via `#d` tag.
3. Filter for DM channels (tagged `["t", "dm"]` and `["hidden"]`).

## Correctness

- The three-step approach correctly discovers DM channels through NIP-29 group metadata rather than relying on a non-existent event kind.
- The empty-channel-ids early return (`println!("[]")`) is a clean guard.
- Uses `query_paginated` instead of `query`, which is more appropriate for potentially larger result sets.
- The tag-filtering logic for `["t", "dm"]` is correct — it checks `a[0] == "t" && a[1] == "dm"`.
- No new dependencies. No `unwrap`/`expect` in production paths (uses `unwrap_or_default()` and `unwrap_or(false)` which are safe).

## Rebase result

Already based on current main (`02f640bc4`). No rebase needed — 0 commits behind.

- Head SHA: `e11e75dfc0b033db0dd0d085090d95ef2458c3ab` (unchanged)
- Mergeable: ✅ MERGEABLE
- CI: DCO ✅, Semgrep OSS ✅, zizmor ✅

No code changes were made — this was a review + rebase pass only.

BradGroux:
Portfolio review update (2026-08-09)

This behavior is still worth fixing, but this PR is no longer the best vehicle. #5139 was opened earlier and has the stronger implementation and verification: it centralizes the relay-state helper, verifies live relay behavior, and preserves deterministic ordering. #5140 covers the same area as well.

To avoid splitting reviewer attention across overlapping changes, I am closing this PR in favor of #5139. One point remains for the surviving implementation to verify: relay limit semantics should be enforced at the authoritative boundary, not only by presentation or collection shape.

PR #5265, opened 2026-08-07
fix(acp): answer DMs without an @mention, and keep presence fresh

coeur-de-loup:
Two issues that make a self-hosted agent look dead to the people using it. Found while running `buzz-acp` as a long-lived systemd service against a self-hosted relay.

## 1. DMs are silently dropped

The mention rule is applied to every channel, but nobody writes `@agent` inside a one-to-one DM — the DM *is* the address. So a user opens a DM with an agent, types a question, and gets nothing back, with no error anywhere.

The requirement is enforced at **two independent gates**, and both had to be cleared for it to work:

- `resolve_channel_filters()` produces the per-channel filter that becomes the relay `REQ`, where `require_mention` adds a server-side `#p` constraint — so a DM without a `p` tag never reaches the harness at all;
- `match_event()` then re-checks `rule.require_mention` client-side.

**Fix:** in `Mentions` mode, add a second subscription rule scoped to the discovered DM channels with `require_mention: false`, and clear the same flag on those channels' resolved filters. Group channels are untouched and keep the mention rule, so an agent still does not answer every line in a busy channel.

## 2. Presence decays to "offline"

Presence is an ephemeral event with a server-side TTL, published once at startup. An always-on harness therefore shows as **offline** in clients about a minute later, while it is happily answering messages.

**Fix:** republish presence every 60s from a background task, only when presence is enabled.

## Verification

Against a self-hosted relay with a `claude-agent-acp` brain:

- a plain DM with no mention now reaches the agent and is answered (previously: nothing, no log);
- mentions in group channels behave exactly as before, and non-mention chatter there is still ignored;
- `GET /api/users/presence` keeps reporting `online` for a long-running harness.

`cargo test -p buzz-acp` passes (689 + 9 tests).

Happy to split this into two PRs if you prefer, or to add tests around the DM rule construction if you point me at the shape you'd want.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -74,6 +74,11 @@ const AUTHENTICATE_TIMEOUT: Duration = Duration::from_secs(10 * 60);
 /// Content is a bare status string (`"online"`, `"away"`, `"offline"`) matching
 /// the desktop client's format. The relay stores this in Redis and synthesizes
 /// it back on presence queries.
+/// How often an always-on harness republishes its presence. Presence is
+/// ephemeral with a server-side TTL; without a refresh the agent reads as
+/// offline in clients while it is still working.
+const PRESENCE_REFRESH_SECS: u64 = 60;
+
 async fn publish_presence(
     publisher: &relay::RelayEventPublisher,
     keys: &nostr::Keys,
@@ -1718,9 +1723,9 @@ async fn tokio_main() -> Result<()> {
     tracing::info!("discovered {} channel(s)", channel_info_map.len());
     let channel_ids: Vec<Uuid> = channel_info_map.keys().copied().collect();
 
-    let rules: Vec<SubscriptionRule> = match config.subscribe_mode {
+    let mut rules: Vec<SubscriptionRule> = match config.subscribe_mode {
         SubscribeMode::Mentions => {
-            vec![SubscriptionRule {
+            let mut rules = vec![SubscriptionRule {
                 name: "mentions".into(),
                 channels: filter::ChannelScope::All("all".into()),
                 kinds: config.kinds_override.clone().unwrap_or_else(|| {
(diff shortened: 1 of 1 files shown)


sonpiaz:
Field evidence that the mechanism described in this diff is exactly what bites, from an always-on agent I have been running on Buzz since the desktop app shipped.

One owner, one agent, one DM channel. Every message I sent from a Mac carried a `p` tag and got answered. One message sent without the tag did not.

```
29 owner messages in the DM channel
28 with a p tag        agent answered every one
 1 without a p tag     2026-08-11T23:59:08Z, content "hey", no turn ever fired
next owner message     2026-08-12T05:13Z, p tag present, answered in 7 minutes
```

Same channel, same two pubkeys, five hours apart. These are signed relay events, not a recollection. The untagged one did not produce a late reply, a warning, or an error anywhere: it was silent on both ends, which is the part that makes this expensive to diagnose as a user. You sit there assuming your agent is down.

That matches the comment in this PR precisely: `require_mention` on the per-channel filter becomes a server side `#p` constraint, so an untagged DM never reaches the client rules at all. The client side rule change alone would not have been enough, and this diff does both halves.

One gap worth flagging before merge. This resolves DM channels once, from `channel_info_map` at startup. Channels discovered later still go through `resolve_dynamic_channel_filter` at lib.rs:2504, which is untouched here:

```
$ git diff origin/main...pr-5265 | grep -c resolve_dynamic_channel_filter
0
$ git diff origin/main...pr-4574 | grep -c resolve_dynamic_channel_filter
8
```

So a DM opened after the agent starts, which is the normal case when somebody messages an always-on agent for the first time, would keep the mention requirement. #4574 threads an `implicitly_addressed_channels` set through both the static and the dynamic resolver and covers that case. Either land the dynamic path here too, or take #4574's shape.

Both branches build and pass the full crate suite on macOS 15 / Apple silicon, `rustc 1.95.0`:

```
pr-5265  test result: ok. 689 passed; 0 failed
pr-4574  test result: ok. 671 passed; 0 failed
```

Happy to run either branch against a live relay with a real untagged DM if that would help close it out.

coreyone:
Tested the DM rule scoping on macOS Apple Silicon. It immediately fixes untagged one-to-one DMs dropping silently.

Two findings regarding the presence refresh implementation:

1. `origin/main` already merged a 60-second presence heartbeat into the main `tokio::select!` event loop (`crates/buzz-acp/src/lib.rs:2821` / `lib.rs:3656`).
2. The independent background `tokio::spawn` ticker in this PR is therefore duplicate. An unlinked `tokio::spawn` can also race during app termination with the exit sequence (`publish_presence("offline")`).

Dropping the background spawn block and keeping the DM subscription rules (`require_mention: false` on DM scope and filters) leaves a clean, minimal diff that avoids shutdown presence races. Strong support for the DM fix.

PR #5266, opened 2026-08-07
fix(desktop): preserve theme when opening communities

tellaho:
**Category:** fix
**User Impact:** Previously selected community themes are preserved when opening a relay through onboarding, while first-time theme migration still completes for communities with no saved theme.

**Problem:** During community initialization, desktop queried theme history before establishing live delivery. If a replacement event arrived while an empty history query was in flight, the client could incorrectly treat the theme as absent and publish the default over the user's saved selection.

**Solution:** Subscribe before fetching history, expose whether live readiness reached EOSE, flush buffered live events before resolving EOSE, and retain the newest delivered replacement through hydration. Seed the inherited/default theme only when both live and history snapshots reach EOSE with no valid or unreadable event; subscription failures, CLOSED, readiness timeout, relay failure, and unreadable events fail closed without publishing.

<details>

**desktop/src/shared/api/relayClientSession.ts / relayClientShared.ts / relayClosedRecovery.ts**
Distinguish EOSE from CLOSED/timeout readiness and flush buffered events before resolving an EOSE fence.

**desktop/src/shared/theme/CommunityThemeController.tsx**
Seed and complete first-community migration only for confirmed absence; uncertain hydration remains non-publishing.

**desktop/src/shared/theme/communityThemePreference.ts**
Keep the inherited appearance for the first migrated community and the stable default for later empty communities.

**desktop/src/shared/theme/communityThemeSync.ts**
Arbitrate live and history results into valid, confirmed-absent, invalid, or unavailable hydration outcomes.

**Tests**
Cover EOSE/CLOSED readiness, subscription failure, timeout, unreadable and live-racing events, no-op initialization, and first-to-later community fallback isolation.

</details>

## Reproduction steps

1. Save a non-default appearance for a community relay.
2. Remove the community locally, then open the same relay again through onboarding.
3. Arrange for the saved replacement event to arrive live while the initial history query returns empty.
4. Confirm the saved appearance remains selected and the client does not publish the default theme over it.
5. On an account with no theme records, open a first empty community and confirm its inherited appearance is migrated; open a later empty community and confirm it starts from the stable default.

## Validation

- Pre-push desktop checks, typecheck, and full desktop tests: passed at `f79556b0e`
- Focused theme/relay readiness tests: 38 passed
- Desktop file-size ratchet and diff check: passed

### DIFF
--- a/desktop/src/shared/theme/CommunityThemeController.tsx
+++ b/desktop/src/shared/theme/CommunityThemeController.tsx
@@ -8,6 +8,7 @@ import {
   clearCommunityThemeOutbox,
   communityThemeApplyExpectation,
   communityThemePersistenceAction,
+  communityThemeScopeFallback,
   hasMigratedCommunityTheme,
   markCommunityThemeMigrated,
   readCommunityThemeOutbox,
@@ -21,6 +22,7 @@ import {
   CommunityThemeSyncManager,
   communityThemeHydrationRemote,
   isNewerCommunityThemeCoordinate,
+  shouldSeedCommunityTheme,
   type RemoteCommunityTheme,
 } from "./communityThemeSync";
 import { useTheme } from "./ThemeProvider";
@@ -74,9 +76,10 @@ export function CommunityThemeController() {
     // Preserve the user's existing global appearance the first time this
     // feature sees their current community. Later missing/malformed target
     // records use the stable default so the previous community never leaks.
-    const fallback = hasMigratedCommunityTheme(pubkey)
-      ? DEFAULT_COMMUNITY_THEME
-      : initialPreferenceRef.current;
+    const fallback = communityThemeScopeFallback(
+      hasMigratedCommunityTheme(pubkey),
+      initialPreferenceRef.current,
+    );
     const scopedPreference = dirty ?? local ?? fallback;
     scopedPreferenceRef.current = scopedPreference;
     applyPreference(scopedPreference);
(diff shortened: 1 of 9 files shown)


desktop/src/shared/theme/CommunityThemeController.tsx
```diff
@@ -127,31 +135,25 @@ export function CommunityThemeController() {
...
+          void dispose();
+          return;
+        }
+        unsubscribe = dispose;
+        const remote = communityThemeHydrationRemote(result);
+        if (remote) {
+          applyRemote(remote);
+          markCommunityThemeMigrated(pubkey);
+        }
+        // Absence is not safe permission to publish during hydration: live
```
wesbillman: This removes seeding for every hydration outcome, including a genuinely absent relay record. That also removes the only `markCommunityThemeMigrated` call for the absent case. With no remote/local/outbox state, `hasMigratedCommunityTheme(pubkey)` therefore stays false forever, so every subsequently opened empty community takes `initialPreferenceRef.current` at lines 77–80 instead of the stable default. Please distinguish trustworthy absence from uncertain live coverage and complete the migration when absence is actually established; add a controller-level multi-community regression test.

wesbillman (dismissed):
Reviewing on Wes's behalf. The subscription-first arbitration is sound, but this patch removes the only successful-absence migration path entirely. On an account with no remote/local/outbox theme, hydration now does nothing: it neither persists the first community's inherited global appearance nor calls `markCommunityThemeMigrated`. Because that marker remains false, every later community with no record continues using `initialPreferenceRef.current` instead of the stable default, contrary to the controller's scope-isolation contract. This is a user-visible regression for accounts without an existing relay theme, not merely a documentation issue. Please preserve fail-closed behavior for uncertain states while adding a trustworthy confirmed-absence path that completes migration/seeding, and cover the controller-level sequence: first absent community migrates the inherited appearance; a later absent community receives `DEFAULT_COMMUNITY_THEME` rather than the first community's appearance. The PR body also currently says a default is seeded when neither source finds a preference, but this head never seeds on hydration.

wesbillman (dismissed):
Reviewing on Wes's behalf. The code path is clean in my re-review, but the current head cannot merge under this repo's authorship policy: commit `f79556b0ef37850cd7cfe3a63acf30fe23749e3b` is authored and signed off by Carl. For block/buzz changes made on Wes's behalf, Wes must be the commit author and `Signed-off-by`, with the implementing agent only in `Co-authored-by`.

Please rewrite that commit so its author and sign-off are `Wes <wesbillman@users.noreply.github.com>`, retaining Carl as co-author, then force-push and let CI/DCO validate the rewritten head. This is my mistake from the implementation pass, not Taylor's.

wesbillman (commented):
Re-review by Carl on Wes's behalf at exact head `f79556b0ef37850cd7cfe3a63acf30fe23749e3b`: the prior behavioral blocker is fixed and no code findings remain. Wes explicitly accepts the current Carl-authored commit attribution, so I have dismissed both stale changes-requested reviews. The PR is mergeable and all applicable checks are green. Per policy, Wes retains approval authority.

wesbillman (approved):

PR #5270, opened 2026-08-07
fix(acp): notice when a @mention turn Ok's with no channel reply

Chessing234:
## Summary
- When an agent's mention turn returns `Ok` but it never publishes a channel reply, buzz-acp now surfaces a "couldn't publish" notice instead of hanging silently. New `silent_reply.rs` watches the turn's reply channel.
- Count the agent's own kind-9 / kind-45001 publishes even under `--no-ignore-self`, and defer the notice by a grace window so a ws echo arriving just after turn completion can clear a false positive.
- Default grace is 8s (override with `BUZZ_SILENT_REPLY_GRACE_SECS`); watches are keyed by generation so a stale timer can't steal a later Ok turn's check. `≤0` falls back to 8.

## Test plan
- [x] `cargo test -p buzz-acp`
- [ ] Mention an agent whose provider errors silently → user sees the couldn't-publish notice
- [ ] Normal reply within grace → no false notice
- [ ] `BUZZ_SILENT_REPLY_GRACE_SECS=2` shortens the window as expected

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -223,6 +223,12 @@ RUST_LOG=buzz_relay=debug,buzz_datastore=info,buzz_db=debug,buzz_auth=debug,buzz
 # Set to true to process the agent's own messages (default: ignore self).
 # BUZZ_ACP_NO_IGNORE_SELF=false
 
+# Seconds to wait after an Ok mention turn before posting a "couldn't publish"
+# notice, so a late ws echo (including post-reconnect replay) can clear it.
+# Default 8 covers ≈6s relay resubscribe bursts; set higher for many-channel
+# harnesses. (#2459)
+# BUZZ_SILENT_REPLY_GRACE_SECS=8
+
 # ── Context ──────────────────────────────────────────────────────────────────
 # Max context messages fetched for thread replies and DMs (0–100). 0 = disabled.
 # BUZZ_ACP_CONTEXT_MESSAGE_LIMIT=12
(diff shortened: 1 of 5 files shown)


wolfyy970 (commented):
I like the intent, but I do not think the channel-level echo counter is safe to merge.

Two turns in the same channel within the eight-second window share one publish count. A successful first turn suppresses a silent second turn, and a successful second turn can suppress a silent first. The current generation test explicitly accepts that combined count. A failure notice emitted for one turn can also be counted as the reply for the next.

There is a separate truth problem: no WebSocket echo before the timer does not prove that publication failed. A reconnect longer than eight seconds can produce the warning and then the real reply.

I would key reply evidence to the triggering event or turn, retain concurrent watches, and keep the copy observational unless Buzz has a failed publish receipt. The prompt already requires `--reply-to`, which gives this a correlation seam. The regressions should cross the event loop with both same-channel turn orderings and a delayed successful echo.

Chessing234:
both points land, and i went looking at the correlation seam you suggested before answering — i don't think it holds.

`resolve_reply_anchor` (queue.rs:1283) returns the thread ROOT for an in-thread turn, not the triggering event; only a top-level mention anchors to the trigger. so two mentions in the same thread both instruct `--reply-to <root>`, and their replies are indistinguishable on the wire. it also returns None for agent↔agent turns, and the prompt explicitly allows an unanchored send for a channel-root or broadcast post. keying evidence to the turn only fixes the cross-contamination for turns in different threads, which is not the case that bit.

and your second point is the one i can't engineer around: `buzz messages send` runs in the agent's own process, so acp never sees a publish result. no echo inside a window is absence of evidence, not a failed publish — a reconnect longer than the grace posts the warning and then the real reply shows up. my own generation test in this diff asserts `Some(2)`, which is the first turn's publish being counted for the second. that's the bug you described, in my own test.

closing rather than shipping a second design with the same hole. if a failed-publish receipt ever exists this is worth revisiting, and in the meantime i'm happy to send the observational half on its own — a warn-level log when an Ok mention turn ends with no echo, no channel post — if you think that's useful for #2459.

Issue #5267, opened 2026-08-07
Agent Snapshot v2: reviewable Skills and tool requirements


wolfyy970:
An imported agent should arrive with the instructions and Skills that make it useful, while asking for access to the tools it needs without carrying credentials or machine configuration.

#5015 gives Agent Snapshot import one clear front door. Snapshot v1 carries the prompt, runtime preferences, profile and optional memory. It cannot carry Skills or tool requirements. Persona Packs can describe both, but they do not import into Desktop, and their MCP configuration contains deployment details that should not travel with an agent.

## User flow

The import preview should show:

- identity and instructions;
- every included Skill, with its exact plaintext available for review;
- the capabilities the agent needs;
- which requirements can be satisfied by this Project and which still need a connection.

Import should create a fresh identity and preserve those declarations without starting commands, installing credentials or silently granting tool access. The user can then connect each requirement to an approved Project connection on the selected execution target.

## Portable boundary

I think this needs Agent Snapshot v2 rather than optional fields on v1. An older client must not accept a portable agent, discard its Skills and tool requirements, then export a reduced file that still looks complete.

The portable values should contain Skill contents and digests plus logical tool requirements. They should not contain commands, URLs, environment-variable names, credential references, local connection IDs or machine paths. Snapshot remains an interchange format, not a runtime registry.

For the first slice, I would keep activation out of scope. It should declare, validate, preview, import and round-trip the new values. Project binding and launch resolution can follow through the shared MCP handoff in #4164.

@tellaho @atishpatel, this builds directly on the Add Agent and snapshot work in #5015 and #5016. I would like to align the v2 boundary with those paths before writing it.

wolfyy970:
Closing this. If snapshot v2 happens it will be small PRs, not this issue.

PR #5015, opened 2026-08-06
feat(desktop): unify add agent flows

tellaho:
**Category:** improvement
**User Impact:** Users can create, discover, and import agents from one consistent Add agent dialog.

**Problem:** Agent creation, discovery, and import were split across a dropdown and separate dialogs, making the Add agent flow fragmented. The existing E2E suite also continued targeting the deleted dropdown after the flows were unified.

**Solution:** Route the new-agent card directly into a unified dialog with dedicated Create, catalog, and Import navigation, then update the affected E2E coverage to exercise that interface and its current empty state.

<details>
<summary>File changes</summary>

**desktop/src/features/agents/ui/AgentDefinitionDialog.tsx**
Supports rendering the agent definition form inside the unified Add agent experience while retaining the standalone dialog behavior.

**desktop/src/features/agents/ui/AgentDefinitionDialogShell.tsx**
Adds the shared shell used to present agent-definition content consistently in embedded and standalone contexts.

**desktop/src/features/agents/ui/AgentDialog.tsx**
Passes the revised dialog state and close behavior through the existing agent dialog entry point.

**desktop/src/features/agents/ui/AgentsView.tsx**
Connects the Agents page to the unified Add agent dialog and opens newly added catalog agents in their profile panel.

**desktop/src/features/agents/ui/PersonaCatalogDialog.tsx**
Combines catalog browsing, agent creation, and snapshot import behind persistent navigation, including dirty-navigation confirmation.

**desktop/src/features/agents/ui/UnifiedAgentsSection.tsx**
Replaces the new-agent dropdown with a direct Add agent entry point and adjusts the responsive card grid.

**desktop/src/features/agents/ui/personaLibraryCopy.ts**
Updates catalog-facing copy for the unified experience.

**desktop/src/features/agents/ui/usePersonaActions.ts**
Returns the resolved local persona after catalog activation so the caller can open the added agent.

**desktop/tests/e2e/agent-readiness-screenshots.spec.ts**
Opens the embedded create pane directly for readiness screenshots.

**desktop/tests/e2e/agents.spec.ts**
Covers unified Create, catalog, and Import navigation and asserts the current shared-agent empty state.

**desktop/tests/e2e/global-agent-config-screenshots.spec.ts**
Updates global configuration screenshot setup for direct create-pane entry.

**desktop/tests/e2e/inline-custom-harness.spec.ts**
Updates custom harness setup for the embedded create form.

**desktop/tests/e2e/persona-env-vars.spec.ts**
[...]

PR #5271, opened 2026-08-07
fix(desktop): reject unsupported snapshot fields

wolfyy970:
Agent snapshots currently import successfully when they contain fields this Buzz version does not understand. Serde drops those fields, so a future snapshot with Skills or tool requirements can appear to import while losing part of the agent.

This makes v1 strict for agents and teams. Newer versions still get the version-specific upgrade message before strict field validation. JSON, PNG, and the production import paths now prove unknown portable fields are rejected rather than discarded.

This is the compatibility guard for #5267. It does not add v2 fields or activate imported capabilities.

Tests:
- `just desktop-tauri-test`
- `just desktop-tauri-clippy`

### DIFF
--- a/desktop/src-tauri/src/commands/personas/snapshot/tests.rs
+++ b/desktop/src-tauri/src/commands/personas/snapshot/tests.rs
@@ -295,6 +295,23 @@ fn import_unsupported_version_fails_closed() {
     );
 }
 
+/// The production import decoder must not accept a v1 manifest while silently
+/// discarding portable capabilities added by another exporter.
+#[test]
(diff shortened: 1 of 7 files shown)


wolfyy970:
@tellaho @atishpatel This is the compatibility guard for the import path from #5015. It prevents Buzz from accepting a snapshot, dropping fields it cannot preserve, and presenting the import as complete; individual, locked, and team JSON/PNG paths now fail with upgrade guidance. Could one of you review it before I add the v2 fields?

wolfyy970:
Closing this. I'll reopen a tighter version if snapshot validation moves.

Issue #5268, opened 2026-08-07
buzz-acp: subscribe=mentions never receives forum-post mentions (kinds 45001/45003)


DamirAGI:
## Summary

Agents running under `buzz-acp` with the default `--subscribe mentions` mode never receive @mentions posted in **forum-style channels**. Forum posts are published as **kind `45001`** (and related forum events as **kind `45003`**), while the mentions subscription only appears to cover chat stream messages (**kind `9`**). The agent looks online and is a channel member, but never reacts to forum mentions.

## Environment

- Relay: `ghcr.io/block/buzz:main`, self-hosted via the Railway template (deployed 2026-08-07)
- Buzz Desktop v0.5.5 (macOS, Apple Silicon) managed agents — bundled `buzz-acp`, Claude harness (`claude-agent-acp` 0.66.0)
- Also observed with `buzz-acp` built from `main` (2026-08-07)

## Reproduction

1. Create a forum-type channel; add a managed agent as a channel member.
2. Agent harness logs the subscription:
   ```
   INFO buzz_acp: membership notification: subscribing to new channel channel_id=0fcf197f-…
   ```
3. Create a **new** forum post mentioning the agent (post created after the subscription, so it is not a history/replay issue).
4. Agent's harness log shows **no inbound event at all** — no turn starts, no reply. (Log ends at the membership notification line.)

## Evidence

`buzz messages get` on the forum channel shows the kind split:

```
kind: 45001 | <owner> | "Vidiš li ovu poruku @Bumble ?"      ← forum post, mention, never delivered to agent
kind: 45001 | <owner> | "@Bumble komentiraj nešto…"          ← forum post, mention, never delivered to agent
kind: 45003 | <owner> | "@Arha @Argus @Bumble @Honey"        ← forum event, same
kind: 9     | <agent> | reply from an agent running subscribe=all
```

A second agent connected to the same relay with `--subscribe all --no-mention-filter` **does** receive these forum events and can reply into the thread (its replies land as kind `9`), so relay-side delivery works — the gap is the kind filter in mentions mode.

## Expected

A mention inside a forum post should wake the agent the same way a kind:9 chat mention does (or, if forums are intentionally out of scope for mentions mode, the limitation deserves a note in the buzz-acp README).

Happy to provide more logs if useful. Great project — filed with ❤️ from a two-human, four-agent relay.

PR #5287, opened 2026-08-08
fix(acp): deliver forum mentions

### LINKED ISSUE #5249: Forum thread mentions never reach managed agents
**Describe the bug**

A managed agent that is a member of a forum channel never responds to an `@mention` posted inside a forum thread. It stays silent, logs nothing, and reports no error, so the agent looks perfectly healthy while ignoring every mention.

The cause is the subscription filter. In `crates/buzz-acp/src/config.rs`, `SubscribeMode::Mentions` builds a fixed kind list:

```rust
let kinds = config.kinds_override.clone().unwrap_or_else(|| {
    vec![KIND_STREAM_MESSAGE, KIND_WORKFLOW_APPROVAL_REQUESTED, KIND_STREAM_REMINDER]
});
```

That is `9`, `46010`, `40007`. Forum posts (`45001`) and forum comments (`45003`) are absent. In `crates/buzz-acp/src/filter.rs` an event whose kind is not in the list is rejected before the `require_mention` p-tag check ever runs, so the mention is discarded at the transport layer.

Neither forum kind appears anywhere in the `buzz-acp` crate.

**Steps to reproduce**

1. Create a forum channel and add a managed agent to it as a bot
2. Create a thread and mention the agent inside it
3. The agent never wakes up. Its log shows no incoming event.
4. Mention the same agent in a stream channel: it responds normally

**Expected behavior**

Mentioning an agent in a forum thread should reach it, same as in a stream channel. The UI lets you add agents to forum channels, so the mention path should work there.

**Workaround**

Setting the environment variable below restores it. The variable replaces the default list rather than extending it, so the three original kinds have to be repeated:

```
BUZZ_ACP_KINDS=9,40007,46010,45001,45003
```

This is safe: `require_mention` stays enabled and its p-tag check is kind-agnostic, so agents still only wake when tagged.

**Suggested fix**

Add `KIND_FORUM_POST` and `KIND_FORUM_COMMENT` to the `SubscribeMode::Mentions` defaults in both places they are constructed (`build_channel_filters` and the dynamic single-channel path). Routing on the p-tag rather than on kind would close the whole class of problem, but adding the two kinds is the minimal change.

**Version and platform**

- Buzz version: 0.5.6
- OS: Windows 11 (10.0.26200)

**Additional context**

Two follow-on effects are visible once mentions do arrive, filed separately: typing indicators and the persistent agent audience are both stream-only in the desktop UI.

ThermoclineLeviathan:
## Summary
- add forum post and comment kinds to the shared ACP mention defaults used by static, dynamic, runtime, and setup-mode paths
- preserve `p`-tag mention enforcement and exclude forum votes from the default kind set
- add regression coverage for tagged and untagged forum events plus override semantics
- document the default forum behavior and a valid `[[rules]]` opt-in for unmentioned forum activity

### Related issue
Fixes #5249.

No duplicate open PR found. #5268 is the closest related report and is already closed as a duplicate of #5249.

### Testing
- `git diff --check`
- static acceptance audit covering static/dynamic subscriptions, runtime dispatch, setup mode, relay serialization/reconnect, and README TOML parsing
- Honest caveat: I could not run `just ci` locally — my sandboxed macOS environment kills native cargo processes (Gatekeeper), and I chose not to weaken that boundary. Hosted CI is the verification path for this PR; requesting first-time-contributor workflow approval.

### How to test manually
1. Configure an ACP agent without an explicit `--kinds` override (default mention subscription).
2. Post a forum post and a forum comment that `p`-tag the agent's pubkey → both should now be delivered to the agent.
3. Post an untagged forum post/comment → not delivered (mention enforcement unchanged).
4. Cast a forum vote tagging the agent → not delivered (votes stay excluded by default).
5. Run an agent with an explicit `--kinds` list → the explicit list still fully replaces the defaults.

No UI change.

### DIFF
--- a/crates/buzz-acp/README.md
+++ b/crates/buzz-acp/README.md
@@ -222,7 +222,16 @@ Start with **N=2** for most deployments. Increase if queue depth grows under loa
 
 ## Forum Channels
 
-By default, the ACP harness subscribes to stream message kinds (9, 46010, 40007). To receive forum events, opt in with `--kinds` and disable the mention filter (forum posts don't @mention agents):
+The default `subscribe=mentions` mode receives tagged forum posts and comments alongside stream messages, workflow approvals, and reminders. Forum events still need a `p` tag for the agent; unmentioned activity remains filtered out. Votes stay excluded because they are not agent-directed messages.
+
+Default mention kinds:
+- **9** — Stream message
+- **40007** — Stream reminder
+- **45001** — Forum post (thread root)
+- **45003** — Comment reply on a forum post
+- **46010** — Workflow approval request
+
+To receive unmentioned forum activity, explicitly widen the filter:
 
 **CLI flags:**
 ```bash
@@ -234,9 +243,11 @@ buzz-acp --kinds 9,46010,40007,45001,45002,45003 --no-mention-filter
 buzz-acp --subscribe all --kinds 9,46010,40007,45001,45002,45003
 ```
 
-**Per-channel config:**
+**Per-channel config (`--subscribe config`):**
 ```toml
-[channel.CHANNEL_UUID]
+[[rules]]
+name = "forum-all"
+channels = ["CHANNEL_UUID"]
 kinds = [9, 46010, 40007, 45001, 45002, 45003]
 require_mention = false
 ```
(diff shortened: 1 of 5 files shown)


ThermoclineLeviathan:
Hi — flagging a mechanical blocker rather than asking for review priority.

The three workflow runs on this PR are still awaiting first-time-contributor approval, so CI has never actually run here. That leaves the required checks unreported and the PR not really reviewable, independent of where it sits in anyone's queue.

Could a maintainer approve the workflow run when it's convenient? Happy to rebase onto current `main` first if that's more useful — just say the word and I'll push.

No rush on the review itself. Thanks for the time.

PR #5291, opened 2026-08-08
fix(relay): log event kind on the HTTP bridge /events line

### LINKED ISSUE #4676: Relay logs omit event.kind, making publish failures indistinguishable from success

Chessing234:
The HTTP bridge request log recorded route, status, and accepted but not
the event kind, so typing indicators (kind 7) and their deletions (kind 5)
were indistinguishable from real messages (kind 9). Every agent turn produced
accepted:true lines whether or not a message was actually sent, which twice
led debuggers to conclude a silent agent had published successfully.

Add kind to the Ok outcome and the tracing::info line so the publish path is
self-describing without a database query.

Closes #4676

### DIFF
--- a/crates/buzz-relay/src/api/bridge.rs
+++ b/crates/buzz-relay/src/api/bridge.rs
@@ -654,12 +654,13 @@ pub async fn submit_event(
         submit_event_authed(&state, &tenant, &headers, &body, pubkey, event_id_bytes).await;
 
     match &outcome {
-        SubmitOutcome::Ok { accepted, .. } => {
+        SubmitOutcome::Ok { accepted, kind, .. } => {
             tracing::info!(
                 pubkey = %pubkey_hex,
                 route = "/events",
                 status = 200u16,
                 accepted,
+                kind,
                 "HTTP bridge request"
             );
         }
@@ -713,6 +714,7 @@ enum SubmitOutcome {
     /// Ingest pipeline ran and returned a result (accepted or not).
     Ok {
         accepted: bool,
+        kind: u32,
         response: Json<Value>,
     },
     /// JSON parse failure before ingest — log category/line/column, not msg.
@@ -843,6 +845,7 @@ async fn submit_event_authed(
             }));
             SubmitOutcome::Ok {
                 accepted: result.accepted,
+                kind: kind_u32,
                 response,
             }
         }


ravarora2 (approved):
lgtm

ravarora2:
🤖 LGTM. Verified locally: checked out the branch and `cargo check -p buzz-relay` builds clean. The change is minimal and correct — `kind_u32` is already computed before ingest (line 832) and reused, so there's no new failure path, and this just brings the accepted (`Ok`) path to parity with `Rejected`, which already logged `kind` the same way. The only other `Ok` site (`into_response`, `{ response, .. }`) ignores the new field, so nothing else breaks.

One optional follow-up: consider logging `kind` on the `SubmitOutcome::Err` arm too. `kind_u32` is in scope there as well, so auth/internal-error lines could be equally self-describing. With `Ok` and `Rejected` now carrying `kind`, `Err` is the last gap (`ParseFail` correctly can't, since the body never parsed). Not blocking — happy to see this merge as-is.

wesbillman (approved):
Reviewed on Wes's behalf at head `693821f4b693315d71577487360e0dd0361165a0`.

This is a real, minimal, and safe observability fix. The handler derives `kind_u32` from the successfully parsed event before moving that same event into ingestion, carries the value only through the success outcome, and adds it as a structured numeric field to the existing terminal request log. It does not change parsing, validation, ingestion, persistence, HTTP responses, or protocol behavior. Rejected parsed events already log their kind; pre-parse and pre-admission failures correctly do not invent one. CI is fully green at the reviewed head.

Non-blocking follow-up: because observability is the entire purpose of this patch, an accepted-event log-capture test should ideally parse the emitted JSON and assert numeric `kind` (and `accepted`). Existing tests assert attribution and line count but would not catch this field being removed or miswired.

Also, the motivating prose appears stale: current typing indicators are kind `20002`; kind `7` is a reaction. That does not affect the implementation, which records the actual generic event kind rather than hard-coding any event category.

wesbillman (approved):

PR #5292, opened 2026-08-08
test(agent): wait for steer rejection, not prompt reply, in steer_rejected_on_empty_prompt

### LINKED ISSUE #4939: steer_rejected_on_empty_prompt is flaky due to response-order race

Chessing234:
steer_rejected_on_empty_prompt was flaky (3/10 with --release) because its
receive loop broke as soon as the session/prompt response arrived, even if the
steer rejection had not yet been read. When the prompt reply won the race the
loop exited before the steer response surfaced, failing the assertion.

Drop the prompt-response break and only stop when the steer rejection is read,
so the test asserts the intended behavior independent of response ordering.

Closes #4939

### DIFF
--- a/crates/buzz-agent/tests/fake_llm.rs
+++ b/crates/buzz-agent/tests/fake_llm.rs
@@ -1428,12 +1428,13 @@ async fn steer_rejected_on_empty_prompt() {
     .await;
     let mut h = Harness::spawn(&url).await;
     let sid = init_session(&mut h).await;
-    let p_id = h
-        .send(
-            "session/prompt",
-            json!({"sessionId": sid, "prompt": [{"type":"text","text":"go"}]}),
-        )
-        .await;
+    // The prompt response is no longer inspected, but the request still has to
+    // be sent to start the run the steer targets.
+    h.send(
+        "session/prompt",
+        json!({"sessionId": sid, "prompt": [{"type":"text","text":"go"}]}),
+    )
+    .await;
     let run_id = recv_active_run_id(&mut h).await;
     let s_id = h
         .send(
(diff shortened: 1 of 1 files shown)


ravarora2 (dismissed):
lgtm

wesbillman (changes requested):
Reviewing on Wes B\u2019s behalf.

The race fix leaves `p_id` unused after removing the prompt-response branch. This is a merge blocker because the repository treats warnings as errors: the PR\u2019s Rust Lint and Windows Rust jobs both fail at `crates/buzz-agent/tests/fake_llm.rs:1431` with `unused variable: p_id` (`-D warnings`). Please remove the binding (while still awaiting `h.send(...)`) or otherwise consume it, then rerun CI.

Chessing234:
good catch, that binding was left over from the removed branch. dropped it, the prompt is still sent and awaited so the run the steer targets still starts. ci running.

Chessing234:
dropped the unused p_id; the test only waits on the steer rejection now.

PR #5294, opened 2026-08-08
fix(relay): stop panicking the ingest worker on reactions to project events

### LINKED ISSUE #4936: relay: a reaction on a project root (issue/PR) panics the ingest worker
# Bug: a reaction on a project root (issue/PR) panics the relay's ingest worker

**Component:** `buzz-relay` — `crates/buzz-relay/src/handlers/ingest.rs`, reaction write path
**Severity:** crash + client livelock (stored-then-panic makes retries permanent)

## Symptom

Publishing a NIP-25-style reaction whose target is a **project root or project comment** (kind 1621 issue, 1618 PR, or a kind-1 comment on one) panics a tokio worker:

```
thread 'tokio-rt-worker' panicked at crates/buzz-relay/src/handlers/ingest.rs:2824:51:
reaction path has channel
```

## Root cause

The reaction path's conformance-trace emission asserts a channel is always present:

```rust
// This branch is the reaction path; channel_id is always Some here, so
// WriteInsertGlobal does not apply.
channel: channel_label(channel_id.expect("reaction path has channel")),
```

That assumption predates reactions on project events. Project events carry no `h` tag, so `channel_id` is `None` — the `expect` panics. Two aggravators:

1. **Stored-then-panic:** the reaction row is inserted *before* the trace action is built, so the client sees a failed request for an event that was actually persisted, and retries.
2. **The duplicate branch carries the identical `expect`,** so every retry panics too. A client with a durable, ordered publish queue (e.g. `buzz-acp`'s gated-publish FIFO acknowledging a project dispatch) is head-of-line blocked forever, burning rate-limit quota on a publish that can never be acknowledged.

Any human tapping a reaction on an issue comment in Desktop triggers the same panic.

## Fix that worked for us

Mirror the message write's three-way split at the same seam (it already handles channel-less writes a few hundred lines below): `(Some, true) → WriteInsert`, `(Some, false) → WriteDuplicate`, `(None, _) → WriteInsertGlobal`. The conformance vocabulary already models channel-less writes; only the reaction path was missing it.

## Reproduction

1. Closed or open relay; publish a kind-1621 issue.
2. React (kind 7) to the issue root or a comment on it, with no `h` tag (as Desktop/agents correctly send for project events).
3. Observe the panic in relay logs; observe the reaction IS stored; retry and observe the duplicate-branch panic.

Chessing234:
A NIP-25 reaction whose target is a project root or project comment (kind
1621 issue, 1618 PR, or a kind-1 comment on one) carries no h tag, so
channel_id is None on the reaction write path. The conformance-trace
emission asserted a channel was always present:

    channel: channel_label(channel_id.expect("reaction path has channel")),

so the worker panicked at ingest.rs:2824. The row was inserted before the
panic, so the client saw a failed request for a persisted event and retried,
and the duplicate branch carried the same expect, head-of-line blocking a
durable publish queue forever.

Mirror the message write's three-way split at the same seam:
(Some, true) -> WriteInsert, (Some, false) -> WriteDuplicate, (None, _)
-> WriteInsertGlobal. The conformance vocabulary already models channel-less
writes; only the reaction path was missing it.

Closes #4936

### DIFF
--- a/crates/buzz-relay/src/handlers/ingest.rs
+++ b/crates/buzz-relay/src/handlers/ingest.rs
@@ -2837,24 +2837,29 @@ async fn ingest_event_inner(
         };
 
         let pubkey_hex = auth.pubkey().to_hex();
-        // Spec WriteInsert (line 514) / WriteDuplicate (line 606): emit
-        // the abstract write action. The persist API returns
-        // `was_inserted` (true → Insert, false → Duplicate). This branch
-        // is the reaction path; channel_id is always Some here, so
-        // WriteInsertGlobal does not apply.
+        // Spec WriteInsert (line 514) / WriteDuplicate (line 606) /
+        // WriteInsertGlobal (line 559): emit the abstract write action. The
+        // persist API returns `was_inserted` (true → Insert/Global, false →
+        // Duplicate). Reactions on project events (issue/PR roots and their
+        // comments) carry no `h` tag, so `channel_id` can be `None` here —
+        // mirror the message write's three-way split instead of asserting a
+        // channel, which panicked the ingest worker on those events.
         let claimed = claimed_community_from_event(&event);
-        let action = if was_inserted {
-            TraceAction::WriteInsert {
+        let action = match (channel_id, was_inserted) {
+            (Some(ch), true) => TraceAction::WriteInsert {
                 msg_id: msg_id_label(event.id.as_bytes()),
-                channel: channel_label(channel_id.expect("reaction path has channel")),
+                channel: channel_label(ch),
                 claimed_community: claimed,
-            }
-        } else {
-            TraceAction::WriteDuplicate {
+            },
+            (Some(ch), false) => TraceAction::WriteDuplicate {
                 msg_id: msg_id_label(event.id.as_bytes()),
-                channel: channel_label(channel_id.expect("reaction path has channel")),
+                channel: channel_label(ch),
                 claimed_community: claimed,
-            }
+            },
+            (None, _) => TraceAction::WriteInsertGlobal {
+                msg_id: msg_id_label(event.id.as_bytes()),
+                claimed_community: claimed,
+            },
         };
         emit(tracer, action, state_for_request(tenant, auth.pubkey()));
         dispatch_persistent_event(


ravarora2:
🤖 Tested locally end-to-end — LGTM. ✅

Reproduced the panic on `main` and confirmed the fix resolves it:

**Before (buggy `main`):** published a global `kind:1` note (no `h` tag → `channel_id = NULL`) and reacted to it. The ingest worker panicked:

```
thread 'tokio-rt-worker' panicked at crates/buzz-relay/src/handlers/ingest.rs:2849:51:
reaction path has channel
```

Postgres confirmed the exact failure mode described in the PR: the reaction row was persisted *before* the panic, so the client's retry hit the `Duplicate` branch and surfaced a misleading `duplicate: reaction already exists`.

**After (this branch):** same steps —
- publish global note → `accepted: true`
- react once → `accepted: true` (no panic)
- react again → `accepted: false` "duplicate" (correct no-op, no panic)
- panics in relay log: **0**

The three-way `match (channel_id, was_inserted)` routing the channel-less case to `WriteInsertGlobal` is the right call, and it mirrors the existing message write path. The `#[ignore]`d regression test (`test_reaction_to_channel_less_event_is_accepted`) is the automated form of exactly what I ran by hand.

ravarora2 (approved):
lgtm

wesbillman (approved):
Reviewed on Wes's behalf at head `1edf6c29a9b88c854a14f44e5be20bbe5d1797af`.

This is a real and safe fix. Reaction scope is deliberately derived from the persisted target; a channel-less project root/comment yields `channel_id = None`. The DB transaction accepts and persists that optional channel before the old trace construction called `expect`, so the reported stored-then-panic failure is genuine. Replacing that false invariant with `(Some, true) -> WriteInsert`, `(Some, false) -> WriteDuplicate`, `(None, _) -> WriteInsertGlobal` exactly mirrors the adjacent generic write seam and uses the conformance model's established representation for channel-less writes and duplicates.

The patch does not change validation, authorization, persistence, deduplication, response, event format, or channel-scoped reaction behavior. It only makes the post-persistence trace action total, allowing the existing global dispatch path to run. The PR also has a documented local end-to-end reproduction demonstrating the panic and stored event on `main`, then successful first publish and clean duplicate handling on this head.

Non-blocking test gap: the PR adds no committed regression test that drives the relay ingest seam with a channel-less reaction. Existing DB tests cover `channel_id = None`, but not the trace construction that panicked. A small extracted action-selection helper test covering both `(None, true)` and `(None, false)`, or an ignored Postgres end-to-end test, would make this fix durable.

Rust lint, unit, security, Semgrep, DCO, and both server cross-compiles were green when reviewed; broader CI was still running, so this approval is a code-safety verdict rather than a claim that the entire matrix has completed.

PR #5298, opened 2026-08-08
fix(desktop): sign NIP-42 AUTH against the bare relay origin, not /pair

### LINKED ISSUE #4932: Legacy /pair pairing fallback cannot pass NIP-42 — desktop signs the URL with path, relay expects bare origin
## Summary

On relays whose NIP-11 lacks `pairing_relay_url` (which includes both tagged relay releases, `relay-v0.1.1` and `relay-v0.2.0`), Buzz Desktop's device pairing falls back to the legacy convention `wss://<relay-host>/pair` — and then cannot authenticate, because the desktop signs its NIP-42 AUTH event with that URL **including the path**, while the relay expects the **bare origin**. The mismatch is unconditional, so legacy-path pairing fails for every operator on the tagged releases.

## The two sides of the contract

**Desktop** (`desktop/src-tauri/src/commands/pairing.rs`): when NIP-11 has no `pairing_relay_url` but advertises NIP-43, `resolve_pairing_relay_url` builds `<main relay>/pair` (`PairingRelay::LegacyPath`), and `handle_nip42_auth` signs the AUTH event with that full URL:

```rust
let relay_url_parsed = nostr::RelayUrl::parse(relay_url)...;   // wss://<host>/pair
... s.sign_event(nostr::EventBuilder::auth(challenge, relay_url_parsed))
```

**Relay** (`crates/buzz-relay/src/api/bridge.rs`, `nip42_expected_relay_url`): the expected value is derived from the tenant host, explicitly pathless — the comment states the contract the desktop is violating:

```rust
/// Path is empty — clients put the bare WS origin (`ws://host[:port]`) in the
/// `relay` tag, matching how `EventBuilder::auth` accepts a [`nostr::RelayUrl`].
format!("{scheme}://{}", tenant.host())
```

`normalize_relay_url` in `crates/buzz-auth/src/nip42.rs` preserves paths (it only trims a trailing slash), so `wss://<host>/pair` vs `wss://<host>` never match. Since every WS connection must complete NIP-42 within the fixed 5s window, the connection is then closed.

## Observed behaviour

Desktop v0.5.4 against `relay-v0.2.0`, with an open pairing relay served behind a path-stripping proxy at `<relay-host>/pair` (so the transport itself works — the WS upgrade succeeds and the QR renders briefly):

Relay log, every attempt:

```
WARN NIP-42 auth failed  error="relay url mismatch"
WARN NIP-42 auth timeout — closing connection  timeout_secs=5
INFO WebSocket connection closed
```

Desktop surfaces this as `relay closed waiting for EOSE` a few seconds after the QR appears.

## Suggested fix (either side resolves it)

- **Desktop:** sign the bare origin of the pairing URL in the AUTH event — strip the path from `relay_url` before `EventBuilder::auth` in `handle_nip42_auth`. One line, and it matches the relay's documented expectation.
- **Or relay:** make `nip42_expected_relay_url` (or the comparison) tolerant of the client's connect path.

## Operator workaround meanwhile

Run a relay recent enough to support `BUZZ_PAIRING_RELAY_URL`, and advertise the open pairing relay at a **dedicated hostname** (path `/`). The desktop then takes the `PairingRelay::Configured` route, dials a pathless URL, signs the bare origin, and auth succeeds. Verified working end-to-end; only the legacy path convention is broken.

Happy to provide fuller logs or test against a patched build.

Chessing234:
On relays whose NIP-11 lacks pairing_relay_url (the tagged relay-v0.1.1 and
relay-v0.2.0 releases), device pairing falls back to the legacy wss://<host>/pair
convention. handle_nip42_auth signed the AUTH event with that full URL including
the path, but the relay's nip42_expected_relay_url derives the expected value
from the tenant host as the bare origin (wss://<host>). normalize_relay_url only
trims a trailing slash, so wss://<host>/pair vs wss://<host> never matched and
every legacy-path pairing failed NIP-42 within the 5s window.

Strip the path before signing so the AUTH event carries the bare origin. For
the modern pairing_relay_url path (already origin-only) this is a no-op.

Closes #4932

### DIFF
--- a/desktop/src-tauri/src/commands/pairing.rs
+++ b/desktop/src-tauri/src/commands/pairing.rs
@@ -607,21 +607,8 @@ where
         Err(_) => return Ok(()),
     };
 
-    // The relay's NIP-42 AUTH expects the bare WS origin (scheme://host[:port]),
-    // not the full connection URL. The legacy /pair fallback connects to
-    // wss://<host>/pair but must sign the AUTH event against wss://<host>, or
-    // the relay rejects it with "relay url mismatch" and closes the
-    // connection within its 5s auth window. Strip the path before signing;
-    // for path-less URLs (the modern pairing_relay_url path) this is a no-op.
-    let parsed =
-        url::Url::parse(relay_url).map_err(|e| format!("invalid relay URL: {e}"))?;
-    let bare_origin = match (parsed.host_str(), parsed.port()) {
-        (Some(host), Some(port)) => format!("{}://{}:{}", parsed.scheme(), host, port),
-        (Some(host), None) => format!("{}://{}", parsed.scheme(), host),
-        _ => relay_url.to_string(),
-    };
-    let relay_url_parsed =
-        nostr::RelayUrl::parse(&bare_origin).map_err(|e| format!("invalid relay URL: {e}"))?;
+    let relay_url_parsed = nostr::RelayUrl::parse(&nip42_auth_origin(relay_url)?)
+        .map_err(|e| format!("invalid relay URL: {e}"))?;
     let auth_json = {
         let guard = session.lock().await;
         let s = guard.as_ref().ok_or("session gone during auth")?;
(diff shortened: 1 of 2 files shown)


Silentpartnercoding:
The bare-origin fix looks right for hostnames, but reconstructing it from `host_str()` drops IPv6 brackets. `ws://[::1]:7777/pair` becomes `ws://::1:7777`, so `RelayUrl::parse` fails before the AUTH event is signed.

Could this use the URL crate’s origin serializer (for example, `parsed.origin().ascii_serialization()`) and add a legacy `/pair` IPv6 regression? That also leaves default-port and host serialization to the URL implementation.

Chessing234:
you're right, host_str() drops the ipv6 brackets. switched to `Url::origin().ascii_serialization()` and pulled it into `nip42_auth_origin` so it's testable — covers the legacy /pair path for a hostname and for `ws://[::1]:7777`, plus a non-default port and an already-origin-only url. no rust toolchain on this machine so ci is the check.

PR #5299, opened 2026-08-08
fix(auth): don't leak the relay's internal address in the NIP-98 mismatch error

### LINKED ISSUE #4954: Relay's internal address is disclosed to unauthenticated callers via push.origin and NIP-98 error detail
### Summary

The relay's unauthenticated root response includes its configured internal address in a `push.origin` field. When the relay is fronted by a reverse proxy on a public hostname, that internal address becomes visible to anyone who can reach the public endpoint.

### Detail

Relay image `ghcr.io/block/buzz:sha-318fbf8`, self-hosted via `deploy/compose` behind Caddy.

`GET /` requires no authentication and returns relay metadata. In our deployment the response body contains the relay's internal network address (the one it was configured with) in `push.origin`.

The same address also appears in the NIP-98 error message on a failed authenticated request:

```
NIP-98 HTTP Auth verification failed: URL mismatch:
  event has `https://public.example.com/query`,
  expected `http://10.0.0.1:3001/query`
```

That error is returned to unauthenticated callers as well, since it fires during verification.

### Impact

Low, but non-zero. In our case the leaked value is a private-range address that is not routable from the internet, so it is not directly actionable for an attacker. It does disclose internal network topology — that the relay sits behind a proxy, and on what address and port — to anyone who can reach the public endpoint.

For deployments where the origin is on a routable address, this would be more meaningful: it hands out the origin address to unauthenticated callers, which undermines a fronting proxy whose purpose is to keep the origin from being contacted directly.

### Suggestion

Either derive `push.origin` from the configured public base URL rather than the bind address, or make the NIP-98 mismatch message generic for unauthenticated callers (log the detail server-side, return only "URL mismatch" to the client).

Not urgent from our side — filing it because it seems like something you would want to know about rather than something we need fixed.

Chessing234:
The NIP-98 URL-mismatch error embedded the expected URL in the message
returned to the caller:

    NIP-98 HTTP Auth verification failed: URL mismatch:
      event has `https://public.example.com/query`,
      expected `http://10.0.0.1:3001/query`

That error fires during verification, so it reaches unauthenticated callers
too. When the relay is fronted by a reverse proxy on a public hostname, the
expected URL is the relay's configured internal bind address — so anyone who
can reach the public endpoint learns the internal address and port (and, for
routable origins, the origin address itself, undermining the proxy).

Log both URLs server-side at WARN for diagnosis, but return only a generic
"URL mismatch" to the client. Existing tests assert on `contains("URL
mismatch")`, which the generic message still satisfies.

Closes #4954

### DIFF
--- a/crates/buzz-auth/src/nip98.rs
+++ b/crates/buzz-auth/src/nip98.rs
@@ -95,9 +95,18 @@ pub fn verify_nip98_event(
         .ok_or_else(|| AuthError::Nip98Invalid("missing `u` tag".to_string()))?;
 
     if normalize_url(u_tag) != normalize_url(expected_url) {
-        return Err(AuthError::Nip98Invalid(format!(
-            "URL mismatch: event has `{u_tag}`, expected `{expected_url}`"
-        )));
+        // Log the detail (both URLs) server-side for diagnosis, but return
+        // only a generic "URL mismatch" to the client. The expected URL is
+        // derived from the relay's configured bind address, which is often an
+        // internal/private address fronted by a reverse proxy; echoing it back
+        // to an unauthenticated caller discloses internal topology (and, for
+        // routable origins, the origin address itself, undermining the proxy).
+        tracing::warn!(
+            u_tag = %u_tag,
+            expected_url = %expected_url,
+            "NIP-98 auth URL mismatch"
+        );
+        return Err(AuthError::Nip98Invalid("URL mismatch".to_string()));
     }
 
     // 6. Verify `method` tag matches expected_method (case-insensitive).


ravarora2:
🤖 Reviewed this and I don't think it's very useful in its current form — the core premise doesn't hold for this codebase, and the change has a real diagnostic cost. A few pinpoints:

1. **The stated threat model is inaccurate.** The title/comment say `expected_url` "is derived from the relay's configured bind address, which is often an internal/private address." It isn't. Both call sites build it from the **request Host**, deliberately *not* the bind address or config host:
   - Bridge (`/events`, `/query`, `/count`): `nip98_expected_url()` = scheme-from-config + `tenant.host()` + path. There's an explicit comment on this: *"Critically, this does NOT use `config_relay_url`'s host"* (`crates/buzz-relay/src/api/bridge.rs`), backed by `verify_bridge_auth_rejects_nip98_event_signed_for_wrong_communitys_host`.
   - Git smart-HTTP: `git_expected_url()` = scheme + `tenant.host()` + repo path (`crates/buzz-relay/src/api/git/transport.rs`).

   `tenant.host()` is the Host header the client itself sent to reach the relay, so the echoed "expected URL" is essentially the public host the client already used — not internal topology, not the origin/bind address. The only server-derived bit is the scheme (http/https), which is a very minor TLS-posture tell.

2. **Real diagnostic cost.** NIP-98 URL mismatches are almost always normalization/proxy-Host quirks, and the removed `expected <url>` was the single most useful self-diagnosis hint. After this change a legitimate client sees only `URL mismatch` and has to ask an operator to grep logs.

3. **A cleaner middle ground:** the client's own `u_tag` is *not* a leak (they sent it), so you could keep echoing ``event has `<u_tag>` `` and drop only ``expected `<url>` ``. That preserves most of the diagnostic value with zero disclosure concern.

4. **Consistency:** several sibling NIP-98 branches still echo detail (method mismatch, timestamp delta, payload hash). If the intent is a hardening pass, worth doing consistently; if not, singling out this one branch is a bit arbitrary.

Net: safe and low-risk, but as written it trades away useful error detail for a leak that doesn't really exist here. I'd either drop it, or re-scope to keeping the client-supplied `u_tag` while omitting the server-side `expected_url` and fixing the rationale.

Chessing234:
you're right, checked it on main. `nip98_expected_url` (bridge.rs:195) and `git_expected_url` (transport.rs:323) both build the host from `tenant.host()`, which is the Host header the client sent — only the scheme comes from config. so there's no internal address in that message to remove, and dropping `expected` just costs diagnostics. closing.

worth noting for #4954: `push.origin` is the same story now — `push_descriptor` takes `tenant_host` from `bind_community(raw_host)` in nip11.rs, not the bind address. the report was against sha-318fbf8, which predates that. i'll say so on the issue.

Issue #4804, opened 2026-08-04
Workflows list returns empty for any account with more than one channel


sdonoghu75:
### Summary

The desktop app's Workflows screen (`/workflows`) always shows "No workflows yet," even for channels that genuinely have workflow definitions saved. Confirmed via `buzz-cli` that the data is present and correctly stored on the relay — this is a read-side bug, not a data-loss bug.

### Repro steps

1. Be a member of 2+ channels (true for basically every real account).
2. Create a workflow in any channel via the desktop app's Workflows screen (`Create Workflow` → fill in trigger/steps → `Create`). The dialog closes with no error, confirming the relay accepted it.
3. Go to the Workflows overview screen. It shows "No workflows yet," including after clicking the manual refresh control and after a full app quit/relaunch.
4. Confirm the data really exists: build `buzz-cli` and run `buzz workflows list --channel <that-channel-id>` authenticated as the same user (or any member). The workflow is returned, full definition intact.

### Root cause

`WorkflowsView.tsx` batches its list query across every channel the user belongs to in a single relay filter, rather than querying per-channel (`desktop/src/features/workflows/ui/WorkflowsView.tsx`, `getChannelsWorkflows(channelIds)` → one `{"kinds": [30620], "#h": [id1, id2, ...]}` filter with N channel ids).

On the relay side, both the desktop app and `buzz-cli` land in the same handler — `POST /query` (`crates/buzz-relay/src/api/bridge.rs`, `query_events` / `query_events_authed`). The catch-all filter path there calls `extract_channel_from_filter` to resolve which channel a filter is scoped to:

```rust
// crates/buzz-relay/src/api/bridge.rs
fn extract_channel_from_filter(filter: &nostr::Filter) -> Option<uuid::Uuid> {
    let h_tag = nostr::SingleLetterTag::lowercase(nostr::Alphabet::H);
    filter.generic_tags.get(&h_tag).and_then(|vs| {
        if vs.len() == 1 {
            vs.iter().next()?.parse::<uuid::Uuid>().ok()
        } else {
            None
        }
    })
}
```

[...]

PR #5307, opened 2026-08-08
fix(relay): stop multi-channel HTTP query from narrowing to one UUID

### LINKED ISSUE #5053: HTTP bridge /query with multiple #h values silently narrows to the lexicographically smallest channel UUID — Workflows screen shows "No workflows yet"
## Summary

A `POST /query` filter with multiple `#h` values returns only events from the channel whose UUID sorts first; all other listed channels are silently ignored. The desktop Workflows screen (experimental) issues exactly this shape — one filter with all member-channel ids (`getChannelsWorkflows`) — so whenever the lexicographically smallest member channel contains no workflows, the screen shows "No workflows yet" even though workflows exist and run fine.

## Reproduction (relay revision 6dbc946, self-hosted, single community)

Signed NIP-98 `POST /query` (agent key + `x-auth-tag`), body `[{"#h":[…],"kinds":[30620]}]`. Three channels: `intake` (1 workflow), `servers` (2 workflows), `opdrachten` (0 workflows). UUID order: `opdrachten` < `intake` < `servers`.

| # | `#h` values | expected | returned |
|---|---|---|---|
| A | intake, servers | 3 | 1 (intake only) |
| B | servers | 2 | 2 |
| C | servers, intake (reversed order) | 3 | 1 (intake only) |
| E | servers, opdrachten (empty) | 2 | **0** |
| G | intake, welcome | 1 | 1 (intake) |

Order-independent (A vs C) → not "first in request wins" but "smallest UUID wins", consistent with iterating a sorted `BTreeSet`.

## Root cause

`crates/buzz-relay/src/handlers/req.rs`, `extract_channel_id_from_filter` (single-filter variant, ~line 858): it returns the **first parseable value** of the `#h` tag set. `nostr::Filter::generic_tags` values are a `BTreeSet<String>`, so "first" = lexicographically smallest. `build_event_query_from_filter` then pins the SQL to `channel_id = <that uuid>`.

The multi-filter variant `extract_channel_id_from_filters` (~line 1029) has the correct guard ("Multiple distinct channel IDs — fall back to global"), but the HTTP bridge path (`api/bridge.rs::query_events_authed` → `build_event_query_from_filter`) uses the single-filter variant without that guard. The bridge's own `extract_channel_from_filter` (used for the access pre-check) does require `len == 1`, so the access check and the query builder disagree about the same filter.

## Suggested fix

Give `extract_channel_id_from_filter` the same semantics as the plural variant: return `None` when the filter carries more than one distinct channel UUID. `apply_access_scope_to_query` then scopes the query to the caller's accessible channels (IN-list), and the existing per-event post-check (`filters_match` + accessible-channel check) keeps results correct.

## Related UI papercut

`WorkflowsView` renders `allWorkflowsQuery.data ?? []`, so a failed or empty query is indistinguishable from "no workflows" — an error state would have made this much easier to spot.

BradGroux:
## Summary

A `POST /query` filter with multiple `#h` values silently returned events from only the lexicographically smallest channel UUID, ignoring all other listed channels. The desktop Workflows screen issues exactly this shape — one filter with all member-channel IDs — so workflows in any channel other than the smallest UUID were invisible.

The single-filter `extract_channel_id_from_filter` returned the first parseable UUID from the `BTreeSet`-ordered `#h` values, which pinned the SQL to `channel_id = <that uuid>`. The multi-filter variant `extract_channel_id_from_filters` already had the correct guard — return `None` when more than one distinct channel appears — but the HTTP bridge path (`build_event_query_from_filter`) used the single-filter variant without that guard.

The fix gives `extract_channel_id_from_filter` the same multi-channel semantics as the plural variant: return `None` when the filter carries more than one distinct channel UUID. `apply_access_scope_to_query` then replaces the single-channel pin with the caller's full accessible-channel `IN`-list, and the existing per-event post-check (`filters_match` + accessible-channel check) keeps results correct.

### Related issue

Fixes #5053.

### Testing

- 4 new unit tests covering the single-filter extraction: single channel, multiple distinct channels (returns `None`), no channel tag (returns `None`), and duplicate same channel (returns the channel).
- Full `buzz-relay` library test suite: 863 passed, 8 failed (all `PoolTimedOut` — database-dependent media/admin tests that fail identically on clean `main` without a Postgres connection).
- `cargo fmt --check` and `cargo clippy -- -D warnings` passed for `buzz-relay`.


BradGroux:
This is the third recurring review + rebase pass for the open BradGroux PRs on `block/buzz`.

## What this PR does

Fixes a relay bug where `extract_channel_id_from_filter` would return the lexicographically smallest channel UUID from a multi-value `#h` tag (due to BTreeSet iteration order) instead of recognizing that multiple distinct channels were requested. When a filter carries multiple distinct `#h` values, the function now returns `None`, causing `apply_access_scope_to_query` to fall back to the caller's accessible-channel list instead of pinning to one arbitrary UUID.

The implementation iterates the tag values, tracks the first UUID found, and returns `None` if a second distinct UUID is encountered. Duplicate UUIDs are handled correctly (return the single UUID).

## Correctness

- The three-way logic (no channel → `None`, single channel → `Some(id)`, multiple distinct channels → `None`) is correct and well-tested.
- Four new unit tests cover: single channel, multi-distinct-value, no channel, and duplicate channel.
- The refactor from manual `iter()` loop to `filter.generic_tags.get(&h_tag)` is cleaner and uses the proper `SingleLetterTag` API.
- No new dependencies. No `unwrap`/`expect` in production paths (test-only `expect` is acceptable).

## Rebase result

Already based on current main (`02f640bc4`). No rebase needed — 0 commits behind.

- Head SHA: `c3c0836406141e1f9c30a3ecc69f703319a679fb` (unchanged)
- Mergeable: ✅ MERGEABLE
- CI: DCO ✅, Semgrep OSS ✅, zizmor ✅

No code changes were made — this was a review + rebase pass only.

BradGroux:
Portfolio review update (2026-08-09)

I rechecked the relay query narrowing behavior against current `main`. The multi-`h` filter issue is still present without this change, and no newer main commit or competing PR supersedes the fix.

I rebased the branch onto `5bf78671f45178f8de02ba18d3d321cbbf19cd1f`. The reviewed patch remains equivalent after the rebase and the branch merges cleanly. The helper coverage remains the relevant regression proof; upstream checks are rerunning for the rewritten head.

BradGroux:
Portfolio rebase and review update (2026-08-11)

Rebased onto `main` at `7e6e9c547` (2026-08-10). No conflicts.

Still needed: The multi-`#h` query narrowing bug is still present in `extract_channel_id_from_filter`. No newer main commit addresses this — the function still returns the first parseable UUID from `BTreeSet`-ordered `#h` values, pinning SQL to one channel when a filter lists multiple channels.

Verification: `cargo check -p buzz-relay` passes cleanly. The PR adds four unit tests covering single-channel, multi-distinct-channel (returns `None`), no-channel, and duplicate-channel cases.

EmminiX:
Confirming this from a self-hosted relay, with two pieces of evidence I have not seen in the other threads on this bug.

### The relay already documents the behaviour this PR restores

`filter_fully_pushable` and `extract_channel_id_from_filter` live ~50 lines apart in the same file and disagree.

`crates/buzz-relay/src/handlers/req.rs:802-808`:

```rust
"h" => {
    // Single #h is pushed as channel_id; multi-#h is not.
    if tag_values.len() > 1 {
        return false;
    }
}
```

`crates/buzz-relay/src/handlers/req.rs:858-869` (current `main`, `be48ce98`):

```rust
        ...
                    return Some(id);
```

So the COUNT path correctly refuses to push multi-`#h`, while the query path pushes the first UUID it meets. Since `generic_tags` is set-backed, "first" is effectively the lexicographically smallest channel UUID, not request order.

**Three other call sites already implement the guard this PR adds**, which is why I read it as a missing guard rather than a design question:

- `extract_channel_id_from_filters` (`req.rs:1029`) returns `None` on multiple distinct UUIDs, with a doc comment explaining exactly why.
- `filter_to_query_params` pushes `#p` only when `values.len() == 1`.
- `crates/buzz-relay/src/api/bridge.rs:1685-1698`, the NIP-50 search path, does the fuller thing: it parses every `#h` value and **intersects the set with `accessible_channels`**, skipping the filter when none survive. A workflow query never reaches it only because that loop `continue`s unless `filter.search` is set.

That last one is worth highlighting: the project has already written the correct multi-`#h` handling once.

### Reproduction on real data

Self-hosted relay, one account, `kind:30620` rows straight from Postgres:

| channel | distinct workflows | events |
|---|---|---|
| `2757e3d4-…` | 2 | 6 |
| `eeacc5bb-…` | 1 | 1 |
| `f8809de6-…` | 2 | 3 |

Five workflows across three channels. Desktop's `get_channels_workflows` batches all member channels into one `#h` filter, `2757e3d4` sorts first, and the Workflows screen shows **2 of 5**. Per-channel `buzz workflows list --channel <id>` returns all of them correctly, which is what makes it look like the data is missing when it is only unreachable.

The batching itself is correct and documented in `desktop/src-tauri/src/commands/workflows.rs`: *"A nostr `#h` filter matches ANY of its listed values, so one query with all channel ids returns the same set."* That is NIP-01. It became load-bearing when the screen moved from N per-channel queries to one batched query.

### One honest caveat on this fix

Returning `None` makes `apply_access_scope_to_query` set `channel_ids` to the caller's whole accessible set, and `bridge.rs:1299` then re-applies the real filter via `filters_match`, so results are correct. But the SQL page (`buzz-db/src/event.rs`, `DEFAULT_MAX_PAGE_LIMIT = 1000`) is filled from the **wider** set before that narrowing. So a caller with more than a page of accessible events in the matched kinds can still lose requested rows off the end. That is the same LIMIT-before-post-filter hazard the `#p` pushdown comment in `filter_to_query_params` already calls out.

Not a blocker for this PR, which is a correct minimal fix and strictly better than the status quo. Flagging it so the follow-up is on the record: pushing the requested set into `EventQuery.channel_ids` and intersecting with accessible channels, exactly as the search path already does, would remove the hazard entirely.

### Duplicates

This is reported in #2385 (oldest), #4419, #4659, #4804 and #5053, with #4446 and #5339 as alternative fixes. Consolidating on #2385 plus this PR would probably make the state of it easier to see.

BradGroux:
Thanks for the thorough confirmation and the real-world reproduction — this is the best evidence on this bug I've seen in any of the threads.

## Verified against current `main` (`be48ce98`)

Every claim in the comment checks out:

**1. The two functions disagree.** `filter_fully_pushable` (line 802) refuses multi-`#h` (returns `false`), while `extract_channel_id_from_filter` (line 858) returns the first UUID from `BTreeSet` iteration order — lexicographic, not request order. This is the bug.

**2. Three other call sites already guard against multi-`#h`:**
- `extract_channel_id_from_filters` (line 1029) returns `None` on multiple distinct UUIDs with a doc comment explaining why.
- `filter_to_query_params` (line 944) pushes `#p` only when `values.len() == 1`.
- `bridge.rs` line 1693 (`handle_bridge_search`) intersects every `#h` value with `accessible_channels` and skips the filter when none survive. This is the most complete handling in the codebase.

**3. The bridge path is also fixed by this PR.** The Desktop Workflows screen goes through `POST /query` → `query_events_authed` → `build_event_query_from_filter` → `extract_channel_id_from_filter`. With this PR, that function returns `None` for multi-`#h`. Then `apply_access_scope_to_query` sets `channel_ids` to the caller's accessible set, and `filters_match` re-applies the full `#h` filter as post-filtering. Results are correct.

## On the LIMIT-before-post-filter caveat

This is an honest and accurate observation. Returning `None` widens the SQL candidate set to all accessible channels before `filters_match` narrows it. The `DEFAULT_MAX_PAGE_LIMIT` (1000) means a caller with >1000 accessible-channel events in the matched kinds could lose rows past the page boundary.

The caveat is not a blocker for this PR because:
- The status quo (pinning to one channel) is strictly worse — it returns *zero* workflows for every channel except the lexicographically smallest one, not just *truncated* results.
- The `#p` pushdown comment in `filter_to_query_params` already calls out the same LIMIT-before-post-filter hazard, so this is a known class of issue in the codebase, not a new one.
- The search path in `bridge.rs:1693` already shows the complete fix: push the requested `#h` set into `EventQuery.channel_ids` and intersect with accessible channels at the SQL level. That eliminates the post-filter narrowing entirely.

The follow-up EmminiX describes (push the requested set into `channel_ids` and intersect with accessible channels, exactly as the search path does) is the right next step. It would require extending `filter_to_query_params` to accept multiple channel IDs instead of a single `Option<Uuid>`, or adding a separate `channel_ids` field to the `EventQuery` that's populated alongside `channel_id`. Either way, that's a follow-up PR, not a reason to block this one.

## On duplicate consolidation

Agreed that consolidating on #2385 plus this PR would make the state easier to see. The five open issues (#2385, #4419, #4659, #4804, #5053) all describe the same root cause. The two other open fix PRs (#4446 by @Trevongit and #5339 by @joaoh8282) both fix the same `extract_channel_id_from_filter` function with the same multi-distinct-UUID guard. #4446 also changes the Desktop side to send N single-`#h` filters instead of one multi-`#h` filter, which works around the bug at the client layer — that's a reasonable belt-and-suspenders approach but it changes the query shape from 1 relay round-trip to N, which is why the relay-side fix here is the better primary fix.

This PR is a correct minimal fix that is strictly better than the status quo. The LIMIT caveat is real but belongs in a follow-up.

BradGroux:
Rebased onto latest main (4b3570671, the 0.5.10 release). No conflicts — `crates/buzz-relay/src/handlers/req.rs` had no recent changes that overlapped.

Reviewed for accuracy and continued relevance:
- The multi-channel HTTP query narrowing bug is still present in main — no recent commit touches `extract_channel_id_from_filter` or the HTTP bridge query path.
- The fix (use `extract_channel_id_from_filters` with the multi-channel guard in `build_event_query_from_filter`) is still the correct approach.
- The Workflows screen still issues the multi-`#h` filter shape that triggers the bug.

The PR is mergeable and ready for review.

BradGroux:
Portfolio rebase and review update (2026-08-13)

Rebased onto `main` at `a96af8952` (2026-08-13). No conflicts.

## Still needed

Yes. The multi-channel HTTP query narrowing fix is not superseded. Two other open PRs from other contributors still target the same bug — #4446 (client-side workaround sending N single-`#h` filters) and #5339 (same guard in `extract_channel_id_from_filter`). Neither has been merged. This PR remains the correct minimal server-side fix.

The LIMIT caveat documented in the previous review still applies: returning `None` for multi-`#h` widens the SQL candidate set before post-filtering, which could truncate at `DEFAULT_MAX_PAGE_LIMIT`. Not a blocker — the status quo returns zero workflows for every channel except the smallest UUID. The complete fix (push `#h` set into `channel_ids` and intersect at SQL level, as the search path already does) remains a follow-up.

## Verification

`cargo check` clean. All 14 portfolio PRs confirmed `MERGEABLE` on GitHub.

BradGroux:
## Rebase review — 2026-08-15

Rebased onto current `main` at `82f7ed153` (2026-08-15). 41 new commits landed since the last review point (`a96af8952`, 2026-08-13). This PR rebased cleanly with no conflicts.

### Still-needed assessment

Checked all 41 new commits for overlap. None supersede this PR. The PR remains the correct fix for its subject area.

### Verification

- `cargo check` not required — no Rust conflicts
- GitHub confirms `MERGEABLE` status

BradGroux:
Rebased onto `main` at d8281b9c9 (2026-08-15). No conflicts — the branch rebased cleanly.

Three commits landed on `main` since the last rebase point (82f7ed153):

- #5947 hide the offcanvas-collapsed sidebar so it stops painting over the community rail
- #5645 Polish mobile message threads and composer
- #5116 require device authentication for identity export (mobile)

All three are mobile or desktop sidebar changes with no overlap with this PR's subject area. The fix is still needed — no recent commit addresses the same problem.

Verification: `tsc --noEmit` passes. GitHub confirms `MERGEABLE`.

BradGroux:
Rebased onto current `main` (`f956e6fe0`, 2026-08-17). No conflicts — the single new commit (#6049, docs refresh) does not touch the relay query filter path. Still needed. Overlapping PRs #4446 and #5339 remain open; this PR is the more complete server-side fix.

BradGroux:
Rebased on latest main (417eea223). One conflict resolved.

Main added a `values.len() != 1` early return to `extract_channel_id_from_filter` (from the multi-channel listing semantics fix #6009). This PR replaces that entire block with a loop that iterates all `#h` tag values and returns `None` when multiple distinct channel UUIDs are found — so the caller falls back to the accessible channel list instead of pinning to one arbitrary UUID. The PR's multi-channel fix is the more complete solution; it subsumes main's single-value check.

BradGroux:
**Rebase and review check — 2026-08-20**

Rebased onto latest main (`a9640c7cc`). No conflicts. GitHub reports MERGEABLE.

**Accuracy review:** The relay's HTTP query handler (`req.rs`) narrows multi-channel queries to a single UUID. When a client requests messages from multiple channels via HTTP, the relay only returns results for one channel. Bug still present.

**Continued relevance:** No recent main commits address multi-channel HTTP query narrowing. PR is still needed.

BradGroux:
This PR was accidentally closed when branches were force-pushed after a commit identity rewrite. The changes have been re-created as #6648 with the same commits rebased onto the latest main. Review history and comments from this PR are preserved here for reference.

https://github.com/block/buzz/pull/6648

Issue #5311, opened 2026-08-08
RFC: non-secret exact-agent credential persistence attestation for managed agents


NedMalki-Chief:
## Problem

External controllers that assign work to a **named** Buzz Desktop agent — schedulers, workflow engines, compliance-gated pipelines — need to verify that the agent's credential is durably held by the OS keyring, bound to exactly that agent, and not sitting in the inline `0o600` JSON fallback. Today there is no supported way to observe this without reading private storage:

- `ManagedAgentRecord.private_key_nsec` is re-hydrated from the keyring on load, so the in-memory record cannot distinguish keyring-backed from inline-fallback storage.
- `SecretStore::probe`/`load` have legacy-migration side effects, so an external observer must not call them even if it could.
- The only alternatives are scraping `managed-agents.json` and the OS keychain directly — exactly the kind of undocumented coupling that breaks on every internal refactor, and which downstream integrators should not be doing at all.

The concrete downstream case: an orchestration layer that hands leads/tasks to a named Buzz sales agent must keep its "production ready" gate **fail-closed** until it can prove exact-agent credential persistence. Without a supported surface that gate can never open against stock Buzz — and the alternatives (forking, or scraping) are worse for everyone.

## Proposal

A public, **non-secret** attestation object issued per managed agent:

```json
{
  "schema_version": "buzz.desktop.exact_agent_credential_persistence.v1",
  "agent_pubkey": "<hex64>",
  "persistence_backend": "os_keyring",
  "inline_fallback": false,
  "parallelism": 1,
  "public_identity_hash": "<hex64>",
  "attestation_hash": "<hex64>",
  "stock_release_id": "buzz-desktop@0.5.7",
  "issued_at": "2026-08-08T12:00:00Z"
}
```

- **No secret ever enters the code path**: the builder takes only "is an inline key present" (boolean), a read-only keyring probe, and public identity material (`pubkey`, the public NIP-OA auth-tag JSON). There is no parameter or field that can carry the nsec.
- **Read-only by construction**: presence is derived via `SecretStore::load_all_readonly`, never `probe`/`load` (which can trigger `migrate_legacy_key`).
- **Fail-closed**: keyring unreachable → error (`attestation_keyring_unreachable`); credential absent everywhere → error (`attestation_credential_missing`). It never guesses. `KeyringProbe::Unreachable` must never read as `os_keyring` — same `Ok(None)`-vs-`Err` discipline as `hydrate_keys_with`.
[...]

Issue #2754, opened 2026-07-24
RFC: Stable extension points for downstream integrations and distributions


senolcolak:
## Background

We started developing [RustChat](https://github.com/kubedoio/rustchat) because the self-hosted team communication space had relatively few modern alternatives to Slack, with Mattermost being the main established option.

RustChat currently provides a Rust backend and web client with channels, threads, direct messages, file sharing, calls, push notifications, OIDC, role-based access control, Mattermost-compatible client APIs, AI agents, RAG, and S3-compatible storage.

After evaluating Buzz, we believe its event-oriented architecture, signed identities, agent model, workflows, Git integration, and human-agent collaboration model provide a very promising long-term foundation.

Rather than continuing to develop a competing messaging core—or maintaining a heavily modified Buzz fork—we would prefer to explore building integrations and a downstream distribution around upstream Buzz.

## Proposal

Would the maintainers be open to defining stable extension points for capabilities that operators and downstream projects may need to integrate without modifying the core relay extensively?

Possible extension areas include:

- identity and account provisioning;
- OIDC and enterprise identity mapping;
- notification providers;
- message auditing for enterprise purposes;
- permission structures—for example, in a school environment, preventing a parent from sending direct messages to students;
- structured onboarding and automatic channel membership;
- extending existing huddle capabilities with optional external voice/video or SFU providers;
- object-storage deployment profiles, including RustFS as an Apache-2.0 Rust-based S3-compatible backend;
- search and knowledge providers;
- MCP tools and agent lifecycle integrations;
- administration integrations;
- external audit and retention services;
- optional frontend or administration extension surfaces.

The goal would not be to introduce a large plugin framework immediately. A smaller collection of stable interfaces, event contracts, hooks, and external-service APIs could provide most of the necessary value.

## Concrete use case

Our intended deployment would combine:

- Buzz as the collaboration, event, workflow, and agent foundation;
- RustShare as a durable document and company-memory layer;
- RustFS as the S3-compatible object-storage backend;
- Keycloak or another OIDC provider for organizational identity;
- an external media service for scalable voice/video calls;
- permission-aware company RAG exposed through MCP tools;
- Kubedo deployment, migration, administration, and managed-operation tooling.

Most of these components could remain external services. We would like to avoid modifying Buzz’s core messaging model or maintaining a permanently divergent fork.

## Initial contributions we could make

We would be interested in contributing work incrementally, potentially starting with:

1. A tested RustFS deployment profile and documentation.
[...]

PR #2957, opened 2026-07-26
feat(agent): show readiness and capabilities before delegation

BradGroux:
Closes #2931.

## What this fixes

Buzz currently makes an owner infer whether an agent is ready for local delegation by cross-referencing the agent editor, managed-runtime state, presence, and ACP observer logs. Missing evidence can look like unsupported behavior, while requested and effective permission modes are not visible together.

The important pre-delegation questions should be answerable in one place: is this local process healthy, which runtime and model are active, which features and tools were actually reported, how risky are those tools, and is the evidence from the current process and session?

## What changes

The owner-only Runtime tab now includes a readiness and capability manifest built from four existing evidence sources:

- the ACP runtime catalog for installation, authentication, adapter, and static runtime facts;
- managed lifecycle state for the current process and community connection;
- current presence;
- the encrypted owner observer stream for ACP initialize, session configuration, commands, tools, model, protocol, and permission evidence.

The card shows installation, authentication, process, community, presence, observer readiness, runtime and model facts, prompt and output features, commands, MCP source names, tool descriptors and risk classes, requested and effective permission modes, evidence source, divergence, freshness, and known limitations.

The reducer treats absent or malformed evidence as unknown. Evidence is retained independently of the capped raw transcript, but it is invalidated across process and session boundaries. A stopped process, failed lifecycle, offline presence, closed observer, or initialize event from an older process cannot produce a ready state.

## Trust boundary

This is local owner evidence, not a public capability, safety, or reputation claim. The UI says “Runtime ready” and identifies the owner-and-machine scope.

The harness projects only MCP server names and permission semantics. It does not forward MCP commands, arguments, environment values, credentials, executable paths, raw lifecycle errors, or config values.

Desktop parsing is strict and bounded. Older Desktop backends that do not report the new static fields remain compatible and render those facts as unknown. A non-owner E2E regression test verifies that neither the Runtime tab nor the manifest is exposed.

## Verification

- Rebased onto current `main` at `5bf78671f45178f8de02ba18d3d321cbbf19cd1f`; the current PR head is `d53016c8490f1c0675b4dc56ac033bb7c9f29084`.
- Resolved the permission conflict in favor of current main's auto-approval behavior while retaining the manifest's requested/effective permission evidence and corrected fallback wording.
- `cargo test -p buzz-acp --lib` passes 689 tests after removing an unrelated inherited `BUZZ_ACP_LAZY_POOL` environment override.
[...]

PR #5312, opened 2026-08-08
feat(desktop): add non-secret exact-agent credential persistence attestation

NedMalki-Chief:
## Summary

Adds a public, **non-secret** per-agent credential-persistence attestation (`buzz.desktop.exact_agent_credential_persistence.v1`), so external controllers and downstream integrations can verify that a named managed agent's credential is OS-keyring-backed — exact-agent-bound, no inline fallback — without any access to key material and without scraping `managed-agents.json` or the OS keychain.

Proposed in #5311 (motivation, and how this relates to #2754 / #3721 / #4925 / #3205 — no duplicate found among open PRs).

## What's in the change

- **`managed_agents/persistence_attestation.rs`** — the schema (`AgentPersistenceAttestation`, extensible `PersistenceBackend`) and a **pure** builder. By construction no code path can carry the nsec: inputs are a presence boolean, a keyring probe value, and public identity material. Fail-closed errors instead of guesses: `attestation_keyring_unreachable`, `attestation_credential_missing`. `attestation_hash` (SHA-256 of the payload with that field empty) makes the object tamper-evident; `verify_attestation_hash` is exported so callers and tests share one definition.
- **`managed_agents/storage.rs`** — `readonly_agent_key_probe` (via `load_all_readonly`, so it can never trigger `migrate_legacy_key` side effects) and `observe_agent_credential_persistence`, which reads the **raw** persisted store: post-`hydrate_keys` records carry the nsec even in the normal keyring-backed case, so only pre-hydration JSON can distinguish inline fallback from keyring-backed.
- **`commands/agent_attestation.rs`** — one read-only `#[tauri::command] get_agent_persistence_attestation(pubkey)`, `spawn_blocking` + store lock, following the `list_managed_agents` / `get_identity` non-secret-projection precedent. Self-checks the hash invariant before returning.
- **`desktop/src/shared/api/agentAttestation.ts`** — thin typed wrapper. The wire shape stays snake_case verbatim: the hash binds the exact serialized payload, so client-side remapping would break external verification.
- **`docs/agent-credential-persistence-attestation.md`** — schema, semantics, and explicit non-goals (not NIP-TR, not a NIP-OA replacement, not a secret channel).

## Design notes

- Tests live in `#[path]`-included sibling `_tests.rs` files per the 1000-line file-size gate; the storage-side probe is tested against the existing `FakeKeyStore` harness (present / other-agent's-entry / no-blob / outage / never-writes).
- Builds without the `system-keyring` feature attest `inline_file` honestly rather than erroring — inline is a true statement about where the key lives there.
- `KeyringProbe::Unreachable` is an error, not an attestation: "may exist but unprovable this boot" must never read as `os_keyring`.
- One test asserts the serialized object contains exactly the nine schema fields and no `nsec` substring, so the no-secret guarantee is enforced by CI rather than by review attention.

## Verification

Run on Windows (`x86_64-pc-windows-msvc`, toolchain 1.95.0):

- `cargo test --lib` in `desktop/src-tauri` — **2156 passed, 1 failed, 10 ignored**. The single failure is `managed_agents::runtime::tests::claude_spawn_uses_the_probed_cli_executable`, which passes in isolation (`cargo test --lib claude_spawn_uses_the_probed_cli_executable`) and is unrelated to this change — it mutates the process-wide `PATH` and is order-dependent on this machine. The 11 new tests (7 `persistence_attestation`, 4 `readonly_probe_*`) all pass.
- `cargo clippy -p buzz-desktop --all-targets --no-deps -- -D warnings` — no findings in any file this PR touches. (Pre-existing findings elsewhere on `main` at `02f640b`, e.g. unused imports in `crates/buzz-terminal` and `managed_agents/nest.rs`, are untouched by this PR.)
- `cargo fmt --check` clean on all changed Rust files; biome clean on the new TS file.
- Commit is DCO signed-off.

Happy to adjust surface details (command vs. settings export, field naming, casing) to maintainer preference.

### DIFF
--- a/desktop/src-tauri/src/commands/agent_attestation.rs
+++ b/desktop/src-tauri/src/commands/agent_attestation.rs
@@ -1,0 +1,58 @@
+//! Per-agent credential-persistence attestation command.
+//!
+//! Returns the non-secret `buzz.desktop.exact_agent_credential_persistence.v1`
+//! object for one managed agent, so external controllers (or the user, via
+//! copy/paste) can verify keyring-backed exact-agent credential persistence
+//! without any access to key material. See
+//! `managed_agents::persistence_attestation` for the schema and guarantees.
+
+use tauri::{AppHandle, Manager as _};
+
+use crate::app_state::AppState;
+use crate::managed_agents::persistence_attestation::{
+    build_agent_persistence_attestation, verify_attestation_hash, AgentPersistenceAttestation,
+    AttestationInputs,
+};
+use crate::managed_agents::storage::observe_agent_credential_persistence;
+
+/// Issue the persistence attestation for `pubkey`.
+///
+/// Read-only: observes the raw persisted store and the keyring via the
+/// side-effect-free path; never migrates, writes, or touches key material.
+/// Fails closed with `attestation_keyring_unreachable` /
+/// `attestation_credential_missing` instead of guessing.
+#[tauri::command]
+pub async fn get_agent_persistence_attestation(
+    app: AppHandle,
+    pubkey: String,
+) -> Result<AgentPersistenceAttestation, String> {
+    tokio::task::spawn_blocking(move || {
+        let state = app.state::<AppState>();
+        // Hold the store lock for a consistent read against concurrent saves.
+        let _store_guard = state
+            .managed_agents_store_lock
+            .lock()
+            .map_err(|error| error.to_string())?;
+        let observation = observe_agent_credential_persistence(&app, &pubkey)?;
+        let package = app.package_info();
+        let stock_release_id = format!("{}@{}", package.name, package.version);
+        let issued_at = chrono::Utc::now().to_rfc3339_opts(chrono::SecondsFormat::Secs, true);
+        let attestation = build_agent_persistence_attestation(&AttestationInputs {
+            agent_pubkey: &pubkey,
+            auth_tag: observation.auth_tag.as_deref(),
+            inline_key_present: observation.inline_key_present,
+            keyring_probe: observation.keyring_probe,
+            parallelism: observation.parallelism,
+            stock_release_id: &stock_release_id,
+            issued_at: &issued_at,
+        })?;
+        // Self-check the tamper-evidence invariant before handing the object
+        // to external verifiers.
+        if !verify_attestation_hash(&attestation) {
+            return Err("attestation_hash_self_check_failed".to_string());
+        }
+        Ok(attestation)
+    })
+    .await
+    .map_err(|error| format!("attestation task join failed: {error}"))?
+}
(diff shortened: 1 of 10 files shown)


wolfyy970 (commented):
I reviewed the implementation at `96b4b06c20`. I think the local storage signal is useful, but I don’t think this should merge as an external attestation in its current form.

The self-hash is only a checksum. A caller can change `persistence_backend`, `stock_release_id`, or `issued_at` and recompute it; there is no authenticated issuer, challenge, audience, expiry, replay protection, or launch binding. It cannot safely open the external compliance gate described in the PR.

The exact-agent and secret boundaries also are not yet true. An entry named `agent:<pubkey>` does not prove its stored nsec derives that pubkey, and `load_all_readonly` brings every stored agent key into this code path just to answer one presence question. That contradicts the stated no-secret path and widens exposure.

I would move the useful result into #2957 as narrow owner-local evidence: keyring verified, inline fallback, missing, unavailable, or mismatched. The storage layer should inspect only the target entry, and the UI should distinguish stored from used by the current launch. If an external controller needs a verifiable claim, that should be a separate challenge-bound signed receipt with an authenticated issuer and launch identity. A Tauri renderer command and self-hash are not that boundary.

NedMalki-Chief:
@wolfyy970 you're right on all four points. Verifying them against the code rather than just agreeing:

**Self-hash.** Correct, and my PR text calling it "tamper-evident" overstated it. `attestation_hash` is a checksum over a payload anyone can regenerate — no authenticated issuer, challenge, audience, expiry, replay protection, or launch binding. It catches accidental truncation and nothing adversarial. It cannot open the external gate I described, and naming the object an "attestation" invited exactly that misreading.

**Exact-agent.** Correct. An entry named `agent:<pubkey>` proves an entry exists under that name, not that the stored nsec derives that pubkey. Proving it means loading the secret and deriving — which this design deliberately refuses — so `exact_agent_credential_persistence` overclaims what the check can know.

**`load_all_readonly`.** Correct, and my justification for it was factually wrong. I used it to avoid `migrate_legacy_key` side effects, but `probe()` doesn't migrate — it calls the read-only `probe_legacy_key`. Only `load()` migrates (`secret_store.rs:549`). So I pulled a full `HashMap` of every stored secret into the command frame to dodge a side effect that wasn't there. (`probe()` reads the blob internally too, but it returns an enum instead of handing every secret to the caller.)

**One more, found while checking the above, which argues for #2957's approach specifically:** `probe()` falls back to `probe_legacy_key` when the blob is missing or lacks the key, so it still finds credentials in the old per-key format. `load_all_readonly` is `load_blob()` alone, with no legacy fallback. On first launch after an upgrade from the per-key format my version would report an un-migrated agent as `attestation_credential_missing` — a false negative on a credential that exists. #2957's `probe()`-based check is strictly more correct here, not merely narrower.

**Trust boundary.** Agreed. A Tauri renderer command is owner-local IPC, not an external trust surface, and a verifiable external claim would need to be a challenge-bound signed receipt with an authenticated issuer and launch binding. That's a different design and shouldn't be bolted onto this one.

Closing in favour of #2957. Thanks for the careful read.

PR #5315, opened 2026-08-08
fix(desktop): wrap markdown table cells in narrow thread panel

### LINKED ISSUE #5313: Thread panel: markdown tables clip long cell text (no wrap / no usable horizontal scroll)
## Bug Description
In Buzz Desktop thread view, GitHub-flavored markdown tables do not wrap cell contents. Long text in table cells is clipped at the thread panel edge, so readers lose the end of each row.

## Steps to Reproduce
1. Open a channel thread panel (narrow width).
2. Post a markdown table with a short first column and a long second column (for example asset/role rows with 80–120 character role text).
3. View the message in the Thread sidebar.

## Expected Behavior
Table cells wrap within the thread content width, or the table container offers clear horizontal scroll without clipping. Prefer wrap for chat readability.

## Actual Behavior
Cell text is truncated/clipped at the panel boundary. Ends of sentences in the Role column are not visible.

## Environment
- App: Buzz Desktop 0.5.7 (CFBundleIdentifier xyz.block.buzz.app)
- OS: macOS
- Surface: Thread panel
- Markdown: standard pipe table

## Evidence
Screenshot (workspace media): attached
https://psgo.communities.buzz.xyz/media/2fa3552eeafb9ec7bceb4e418c98391492408c471e9690c0f1664da5c4b96453.png

Image

## Suggested fix direction
- overflow-wrap / word-break on table cell content in the thread markdown renderer
- table-layout: fixed; width: 100% with wrapping cells, or a scrollable table wrapper with a visible overflow cue
- QA both main timeline and thread sidebar widths

## Workaround
Use bullet lists instead of wide tables in threads.
My first bug report

santhiprakash:
## Summary
In the desktop thread panel, markdown tables with a long second column were clipped at the panel edge because the table used `w-max` and the cells had no wrapping rule. This change switches the table to `w-full table-fixed` and adds `break-words` to `td`/`th` so long cell content wraps within the available width.

### Related issue
Fixes #5313

### Testing
- `cd desktop && pnpm typecheck` — passed
- `cd desktop && pnpm check` — passed (no new lint/file-size/px-text/pubkey issues)
- `pnpm build:e2e` succeeded
- Captured Playwright screenshots in the mock thread panel:

**Before — text clipped at the panel edge:**
before

**After — long cell text wraps within the thread panel:**
after

### DIFF
--- a/desktop/src/shared/ui/markdown.tsx
+++ b/desktop/src/shared/ui/markdown.tsx
@@ -1652,12 +1652,12 @@ function createMarkdownComponents(
     ),
     table: ({ children }) => <MarkdownTable>{children}</MarkdownTable>,
     td: ({ children }) => (
-      <td className="border-t border-border/70 px-3 py-2 align-top">
+      <td className="border-t border-border/70 px-3 py-2 align-top break-words">
         {children}
       </td>
     ),
     th: ({ children }) => (
-      <th className="bg-muted/60 px-3 py-2 font-semibold text-foreground">
+      <th className="bg-muted/60 px-3 py-2 font-semibold text-foreground break-words">
         {children}
       </th>
     ),
--- a/desktop/src/shared/ui/markdown/MarkdownTable.tsx
+++ b/desktop/src/shared/ui/markdown/MarkdownTable.tsx
@@ -12,7 +12,7 @@ export function MarkdownTable({ children }: { children?: React.ReactNode }) {
       className="overflow-x-auto rounded-2xl border border-border/70"
       data-table-block=""
     >
-      <table className="w-max min-w-full border-collapse text-left text-sm">
+      <table className="w-full table-fixed border-collapse text-left text-sm">
         {children}
       </table>
     </div>


santhiprakash:
## Screenshots

pr-5315--01-before

pr-5315--02-after

ravarora2:
lgtm! Few things for the author: 

- The change is global, not thread-panel-specific. MarkdownTable is the shared renderer used everywhere — main channel messages, thread panel, PersonaCatalogDialog, agent session summaries, user message bubbles, etc. (I verified the import sites.) So every markdown table app-wide switches to fixed-layout + wrap, including the much wider main message view. The title/issue frame it as a thread-panel fix, but there's no scoping — it changes all tables. Worth confirming that's intended.

- It trades "horizontal scroll for wide tables" for "always wrap to fit." With w-full, the table never exceeds the container, so the overflow-x-auto wrapper effectively never scrolls anymore (it's now near-dead code — harmless, but moot). And table-fixed means columns no longer auto-size to content — widths get distributed evenly/by the first row. For a simple 2-column table that's fine, but for wide, many-column or data/number/code tables, this can produce cramped, very tall cells where horizontal scrolling was arguably the better UX. That's the real judgment call for a reviewer.

santhiprakash:
Closing: the source files modified by this PR (desktop/src/shared/ui/markdown.tsx, desktop/src/shared/ui/markdown/MarkdownTable.tsx) have been refactored and split into separate components on upstream/main (markdown-text.tsx, compact-markdown.tsx, markdown-blocks.tsx). The component structure no longer supports the original diff. Per @ravarora2's feedback, a future PR should scope the table-layout fix to thread-panel-only rendering, not the global MarkdownTable component.

santhiprakash:
@ravarora2 Thanks for the careful review — you're right on both points.

**Global scope:** The original patch was not intended to change table behavior app-wide. I agree that `table-fixed` + always-wrap is the wrong default for the wider main channel timeline, PersonaCatalogDialog, agent summaries, etc. — horizontal scroll is still the better UX for wide/multi-column data tables there.

**What we did instead:** The branch now scopes the wrap behavior to **thread-panel reply rows only** via `wrapTables` on `MessageRow`'s existing `layoutVariant="thread-reply"` path. Main channel messages keep the prior `w-max min-w-full` layout (overflow-x-auto still works for wide tables). Thread replies get `w-full table-fixed` + `break-words` so long second-column text wraps instead of clipping in the narrow panel.

**Judgment call:** For Buzz chat content, tables in messages are almost always small (2–4 columns of prose/checklists). Reserving wrap-for-fit to the narrow thread surface matches that while leaving wide-table scroll where there's room. Happy to revisit if we start seeing dense data tables in main-channel messages.

Rebased onto current `main`, added a small regression test for the table class helper, and reopened the PR — ready for another look when you have a moment.

Issue #5303, opened 2026-08-08
buzz-acp: --permission-mode is set but never reaches the agent; agents are silently muted while still billing full turns


cristiansotogarciaxatech:
## Summary

`buzz-acp` resolves and logs `permission_mode=bypassPermissions`, but that value is never applied to the `claude-agent-acp` session. The adapter derives its permission mode solely from the user's `~/.claude/settings.json` → `permissions.defaultMode`. On a machine where that is `"auto"`, every managed agent runs headless in a mode that denies any non-allowlisted tool and cannot prompt — so `buzz messages send` is refused.

The failure is silent and expensive. Agents wake normally, load their full session, compose a reply, get denied on every send path, and end the turn. The channel stays empty, tokens are billed in full, and nothing is logged as an error at the Buzz layer.

## Environment

- Buzz Desktop 0.5.7 (also reproduced on 0.5.6)
- `@agentclientprotocol/claude-agent-acp` 0.64.0
- Bundled Claude Code CLI 2.1.220
- Windows 11
- 12 managed agents (7 `claude-agent-acp`, 4 `codex-acp`, 1 native)

## Evidence

**1. buzz-acp resolves the mode correctly.** Every agent's startup line reports it:

```
buzz-acp starting: ... idle_timeout=900s max_turn=7200s agents=10 subscribe=Mentions
  context_limit=12 max_turns_per_session=0 memory=true permission_mode=bypassPermissions
```

**2. The adapter ignores it.** `claude-agent-acp` `dist/acp-agent.js`:

```js
const permissionMode = resolvePermissionMode(
    settingsManager.getSettings().permissions?.defaultMode, this.logger);
```

The session's mode comes from `settings.json`. The ACP-side `permission_mode` is not consulted. `PERMISSION_MODE_ALIASES` maps `auto → "auto"`, so a user with `"defaultMode": "auto"` gets `auto` regardless of what Buzz asked for.

**3. Result at runtime.** Three agents mentioned in one channel message, all denied on every path:

```
07:13:15  Bash: buzz messages get --channel f5c91c95-...
          → "Permission to use Bash has been denied because Claude Code is
             running in don't ask mode."
07:13:20  PowerShell: buzz messages get ...        → denied
07:13:34  PowerShell: (the composed reply)         → denied
07:13:42  agent: "I'm blocked from replying ... the `buzz` CLI is the only way
                  I can send channel messages."
```

[...]

Issue #5290, opened 2026-08-08
Claude Code runtime always launches in dontAsk, preventing agents from executing commands


Akthamix:
## Describe the bug

After updating to Buzz v0.5.5, every Claude Code agent starts in `dontAsk` permission mode.

As a result, my agents can no longer execute shell commands. They silently fail because command execution is denied.

Running Claude Code directly from Terminal works correctly.

## Steps to reproduce

1. Update to Buzz v0.5.5.
2. Configure an agent to use the Claude Code runtime.
3. Ask the agent to perform any action that requires running a command.
4. Open the Activity panel.

## Actual behavior

The Activity panel shows:

```
Mode: dontAsk
```

The agent reports that it cannot execute commands because permissions are denied.

Example:

> "permission to run commands is switched off in this session."

This prevents the agent from reading the repository, executing Git commands, or doing any normal development work.

## Expected behavior

Buzz should either:

- Respect Claude Code's configured/default permission mode, or
- Allow selecting the permission mode (Auto, acceptEdits, bypassPermissions, etc.) for the Claude Code runtime.

At the moment, Buzz appears to always launch Claude Code in `dontAsk`.

## Version and platform

- Buzz version: 0.5.5
- OS: macOS 15
- Hardware: Apple Silicon (M1 Max)

## Additional context

Launching Claude Code manually from Terminal with:

```bash
claude --permission-mode bypassPermissions
```

works correctly.

The problem only occurs when Claude Code is launched through Buzz.

I've attached screenshots showing:

- Buzz Activity reporting `Mode: dontAsk`
- Claude Code working normally in `bypassPermissions`

jayluxferro:
Getting the same issue here. @Akthamix thanks for raising the issue

<img width="1328" height="196" alt="Image" src="https://github.com/user-attachments/assets/06eb2368-b54e-4771-adb1-5fb4ee85cc79" />

jds950:
Confirmed, also getting the same issue here on the latest software. Claude code works fine from the cli outside of buzz but cannot change the claude mode even from within the buzz terminal, stuck on "  ⏸ manual mode on · ← 1 agent".

Akthamix:
the only fix was downgrading back to v0.5.5

daogrovelabs:
Also getting the same issue here after the update. The same issue for Codex and Kimi Code as well (Hermes is working though).

jaozolins:
+1

JVPartanen:
[...]

Issue #5262, opened 2026-08-07
Managed agents can't run the bundled buzz CLI after #4609 — all permission requests rejected under dontAsk


thomaspblock:
## Summary

After #4609 (`fix(acp): reject unattended permission requests`, commit `ad538bfb1`, merged 2026-08-05), desktop-managed agents whose runtime raises a `session/request_permission` for the bundled `buzz` CLI are **unconditionally denied** — every `buzz` invocation, including `buzz --help`, fails. Since `buzz` is the agent's only interface to the relay (reading channels, sending replies), an affected agent cannot perform its primary function.

## Report

An X user reported (screenshot relayed in the `buzz-bugs` channel): a managed agent ("Growth", Claude Code runtime, **Mode: dontAsk**) says:

> I can't reach the buzz CLI in this session — every `buzz` command (even `buzz --help`) is being auto-denied by permission settings, while other Bash commands go through fine. That blocks me from reading the channel history or sending a reply.

## Root cause

#4609 intentionally made the harness fail closed:

1. `bypassPermissions` removed from the `PermissionMode` enum (`crates/buzz-acp/src/config.rs`).
2. Managed sessions default to `dontAsk`.
3. `handle_permission_request` (`crates/buzz-acp/src/acp.rs`) now answers every `session/request_permission` with `reject_once` (or `cancelled`) instead of the old `allow_once` auto-approval.

The commit's operator notes predicted this exact failure:

> "Desktop managed agents do not set a permission mode, so they inherit `dontAsk`. The desktop has no permission prompt, so operations needing approval now fail with no in-app way to approve them."

Why only `buzz` is affected: the agent runtime (e.g. Claude Code) only raises permission requests for tool patterns not pre-authorized in its own local settings. Common shell commands are covered by the runtime's allowlist and never ask; the `buzz` sidecar binary is not, so every invocation asks — and the harness now rejects all asks.

## Impact

- **Severity: high** for managed agents on affected runtimes/configs — the agent is cut off from the relay entirely (cannot read, cannot reply), while appearing "online".
- The failure is confusing to end users: the agent blames "permission settings" with no in-app way to see or fix anything.

## Proposed fix (narrow, preserves fail-closed posture)

[...]

PR #5316, opened 2026-08-08
docs(acp): explain dontAsk's reply-path denial and the `default` escape hatch

marwanakeel:
## Problem

`dontAsk` became the harness default in #4609. It is applied with `session/set_config_option` *after* session creation, so it overrides whatever `permissions.defaultMode` / `permissions.allow` the agent's own configuration established.

Under that mode the bundled `buzz` CLI is itself an approval-gated tool call — and Buzz's own injected prompt tells agents to reply with `buzz messages send`. So a managed agent can receive a mention, load context, do the work, and then be denied the send:

```
>> buzz messages send --channel <uuid> --reply-to <hex> --content "..."
<< Permission to use Bash has been denied because Claude Code is running in don't ask mode.
```

**The failure leaves no trace in the harness.** The denial happens inside the agent, so no `session/request_permission` ever reaches `buzz-acp` and there is nothing for it to log. From the outside the agent is connected, subscribed, consuming a full turn, and posting nothing — which reads as a relay or delivery problem. Reporters in #5290 describe losing significant time to exactly that misdiagnosis.

`--permission-mode default` already avoids this: `apply_permission_mode` is gated on `!is_default()`, so no `set_config_option` is sent and the agent's own configuration stays in charge. But nothing says so anywhere, and "default" reads naturally as "the agent's per-call prompt" — which is the thing `dontAsk` was introduced to reject. In #5290 an operator tried `settings.json` with `defaultMode: bypassPermissions` and `allow: ["Bash(buzz:*)"]`, found it ignored (correctly — the stamp lands after session creation), and concluded the only fix was rolling back to 0.5.5.

## Change

Documentation and one startup log line. **No behaviour change.**

- `--permission-mode` help now states that the mode is stamped over the agent's own configuration, what that costs under `dontAsk`, and that `default` opts out.
- A single `WARN` at startup when `dontAsk` is in effect, naming the consequence and the escape hatch.

The #4609 guarantee is deliberately untouched: any `session/request_permission` reaching `buzz-acp` is still answered with `reject_once` or cancellation, in every mode including `default`. This PR does not restore `bypassPermissions` and does not change what the harness will approve — it only stops the current default from being a silent, undiscoverable dead end.

## Testing

- `env -u BUZZ_ACP_LAZY_POOL cargo test -p buzz-acp` — 691 lib + 9 integration tests pass, including two new ones covering which modes warn and that the warning names both the escape hatch and the unchanged guarantee
- `cargo fmt -p buzz-acp -- --check` — clean
- `cargo run -p buzz-acp -- --help` — rendered and read back

## Not addressed

Whether `dontAsk` should remain the default, and whether an opt-in non-interactive mode should exist for operators running their own agents on their own machines (#5290, #2884), are product calls for the maintainers. This PR only makes the current behaviour legible.

Refs #5290, #5303, #5262, #2884

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -162,14 +162,13 @@ impl std::fmt::Display for PermissionMode {
 /// One-line operator warning for a permission mode that silently denies the
 /// agent's own reply path. Returns `None` for modes that don't.
 ///
-/// `dontAsk` is the default, and it is applied with `session/set_config_option`
-/// after session creation, so it overrides whatever the agent's own
-/// configuration established. The resulting failure is invisible from the
-/// outside: the denial happens *inside* the agent, so no
-/// `session/request_permission` ever reaches this harness and there is nothing
-/// for it to log. An agent stays connected and subscribed, consumes a full
-/// turn, and publishes nothing — because the bundled `buzz` CLI it is prompted
-/// to reply with is itself an approval-gated tool call.
+/// `dontAsk` is the default. It overrides the agent's configured default mode,
+/// so every tool call not matched by a preauthorization rule is denied without
+/// prompting — including the bundled `buzz` CLI the agent is prompted to reply
+/// with. The resulting failure is invisible from the outside: the denial
+/// happens *inside* the agent, so no `session/request_permission` ever reaches
+/// this harness and there is nothing for it to log. An agent stays connected
+/// and subscribed, consumes a full turn, and publishes nothing.
 ///
 /// Stating it once at startup costs one line and saves operators from
 /// diagnosing an empty channel as a relay or delivery problem.
@@ -178,12 +177,15 @@ impl std::fmt::Display for PermissionMode {
 /// constructing a full [`Config`].
 pub fn unattended_denial_warning(mode: PermissionMode) -> Option<&'static str> {
     matches!(mode, PermissionMode::DontAsk).then_some(
-        "permission_mode=dontAsk — approval-gated tool calls are denied inside the agent, \
-         including the bundled `buzz` CLI agents use to publish replies, so an agent may \
-         consume a full turn and post nothing. This is applied after session creation and \
-         overrides the agent's own permissions configuration. To leave that configuration in \
-         charge, set --permission-mode (BUZZ_ACP_PERMISSION_MODE) to `default`; unattended \
-         permission requests that reach the harness are still rejected either way.",
+        "permission_mode=dontAsk — tool calls not matched by a preauthorization rule are denied \
+         inside the agent without prompting, including the bundled `buzz` CLI agents use to \
+         publish replies, so an agent may consume a full turn and post nothing. Matching \
+         permissions.allow rules do still apply, but Claude Code ignores a project's \
+         .claude/settings.json in a workspace that was never trusted interactively — use \
+         .claude/settings.local.json to preauthorize commands for a headless agent directory. \
+         To leave the agent's own default mode in charge instead, set --permission-mode \
+         (BUZZ_ACP_PERMISSION_MODE) to `default`; unattended permission requests that reach the \
+         harness are still rejected either way.",
     )
 }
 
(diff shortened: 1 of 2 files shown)


LucasMoskun:
Small but important factual correction before this documentation lands: `dontAsk` overrides the agent’s configured **default mode**, but it does not universally override matching `permissions.allow` rules.

Current Claude Code evaluates applicable allow rules before denying otherwise approval-gated calls under `dontAsk`. We now have multiple confirmations:

- In the managed Buzz reproduction on #5262, moving `Bash(buzz)` / `Bash(buzz:*)` to `<nest>/.claude/settings.local.json` allowed `buzz --help` under `dontAsk`.
- [This independent test](https://github.com/block/buzz/issues/5262#issuecomment-5226098830) removed the `--setting-sources` confound and reproduced the result with default setting sources: `settings.json` was ignored in an untrusted workspace, while the identical rule in `settings.local.json` was applied.
- The reporter’s corrected analysis on #5303 likewise notes that allow rules are evaluated per tool call and can restore messaging without changing the session mode.

The negative `settings.json` result cited from #5290 is therefore better explained by Claude Code’s workspace-trust gate, not by `session/set_config_option` overriding all allow rules.

Suggested wording:

> `dontAsk` overrides the agent’s configured default permission mode. Commands not matched by an applicable preauthorization rule are then denied without prompting. On current Claude Code, matching `permissions.allow` rules still apply; however, project-level rules in `.claude/settings.json` may be ignored for an untrusted headless workspace. `.claude/settings.local.json` has been independently verified to apply in that case.

`BUZZ_ACP_PERMISSION_MODE=default` is still a valid escape hatch for leaving the agent’s default mode in charge, but it is not the only possible recovery. Scoped, trust-effective allow rules can also keep the Buzz CLI usable under `dontAsk`.

I recommend updating both the CLI help and startup warning so operators are not incorrectly told that all agent-side preauthorization is ineffective.

marwanakeel:
@LucasMoskun you're right, and the correction is now pushed. Thanks — the original wording would have told operators that agent-side preauthorization is futile, which is worse than saying nothing.

I want to flag that my own evidence pointed at the wrong cause, in case it helps anyone triaging: on a self-hosted 0.5.7 nest I put `Bash(buzz:*)` in `<nest>/.claude/settings.json`, saw `buzz messages send` refused anyway, and concluded the `set_config_option` stamp defeats allow rules. The workspace-trust gate explains that result at least as well, and @DamienStevens' two-directory test isolates the filename as the only variable, which mine did not. Independent data point for the trust-gate reading: my refusal reproduced with `permission_mode=default`, i.e. with no stamp applied at all — so the stamp cannot have been what suppressed the rule.

What changed, close to your suggested wording:

- **Help text** — `dontAsk` overrides the agent's configured *default mode*, and matching `permissions.allow` rules are still evaluated per tool call; only unmatched calls are denied. Added the caveat that `permissions.allow` from a project `.claude/settings.json` is dropped in a workspace never trusted interactively — every desktop-managed nest — while `.claude/settings.local.json` is honoured.
- **Startup warning** — same correction, and it now names `settings.local.json` rather than presenting `--permission-mode default` as the only recovery.
- **Tests** — a new case pins that the warning names `settings.local.json` and does not imply allow rules are ineffective, so a future edit can't silently reintroduce the wrong claim.

692 lib + 9 integration tests pass; clippy and fmt clean.

Two things I deliberately left out, since they're beyond a docs change and belong with the maintainers:

1. @DamienStevens' point that a `buzz`-only grant "leaves agents able to *speak* but not to *work*" — under the previous `bypassPermissions` default these agents also ran `git`/`gh`, wrote files, and read outside the nest. Whatever gets seeded in #5263 needs `Read`/`Write`/`Edit` and `permissions.additionalDirectories`, or the follow-up report writes itself.
2. On my nest the agent stopped attempting `buzz messages send` at all once the mode was `default` — text reply, zero tool calls, and in **dm** scope the injected context carries no publish instruction (channel scope has the explicit `--reply-to` line). The relay shows the signature Damien describes: kind-7 markers and their kind-5 deletions stored, no kind-9. I don't know yet whether that's prompt-adherence or a missing instruction, so I'm not asserting it here — happy to open a separate issue with the transcripts if it's useful.

PR #5321, opened 2026-08-08
feat(runtime-config): add unified MCP config model with Hermes and Kimi Code adapters

whyihaveyou:
## Summary

Adds a new lightweight crate `buzz-runtime-config`: a **runtime-agnostic, read-only** model of external agent frameworks' MCP server configuration, with adapters for Hermes (`~/.hermes/config.yaml`, `mcp_servers`) and Kimi Code (`~/.kimi-code/mcp.json`, `mcpServers`).

This is a pathfinding slice (**PR-0**) toward the larger goal of letting Buzz manage Skills / MCP for different agent frameworks from one place — and it dovetails directly with the direction of #4164 (*buzz-acp: add structured MCP server configuration*). It is deliberately backend-only: pure structs + `serde`, no write paths, no desktop changes.

## What's in it

- `src/model.rs` — unified model: `RuntimeKind` (Hermes / KimiCode), `McpServerConfig` (`enabled: Option<bool>`, `None` = native config did not specify → treated as enabled), `RuntimeMcpConfig` with structural `validate()` (empty name / empty command / duplicate names) and `redacted()` (masks env values so secrets never reach logs or display).
- `src/hermes.rs` — read-only adapter for YAML `~/.hermes/config.yaml`. Unknown top-level keys ignored.
- `src/kimi.rs` — read-only adapter for JSON `~/.kimi-code/mcp.json`, with `$KIMI_CODE_HOME` fallback. Unknown top-level keys ignored.
- `src/error.rs` — `ConfigError` (Io / Yaml / Json / Validation).

## Scope notes / intentional limits

- **stdio servers only for now.** Kimi Code also supports HTTP/SSE servers declared with `url` instead of `command`; those surface as a loud validation error naming the server rather than being silently dropped. A follow-up PR extends the model with a transport field.
- Kimi timeout / tool-list fields (`startupTimeoutMs`, `toolTimeoutMs`, `enabledTools`, `disabledTools`) are not modeled yet.
- Adapters are strictly read-only; secret values are never logged (`redacted()` for display).

## Testing

- 22 unit tests across model/hermes/kimi (parse, defaults, validation, redaction, disk roundtrip).
- `just ci` green locally (fmt + clippy + unit tests + build).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1270,6 +1270,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "buzz-runtime-config"
+version = "0.1.0"
+dependencies = [
+ "serde",
+ "serde_json",
+ "serde_yaml",
+ "tempfile",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "buzz-sdk"
 version = "0.1.0"
 dependencies = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ members = [
     "crates/buzz-pairing-cli",
     "crates/buzz-sdk",
     "crates/buzz-persona",
+    "crates/buzz-runtime-config",
     "crates/git-credential-nostr",
     "crates/git-sign-nostr",
     "crates/buzz-pair-relay",
--- a/crates/buzz-runtime-config/Cargo.toml
+++ b/crates/buzz-runtime-config/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "buzz-runtime-config"
+version = "0.1.0"
+edition = "2021"
+description = "Runtime-agnostic MCP/Skill configuration model with read-only adapters for external agent frameworks (Hermes, Kimi Code)"
+license = "Apache-2.0"
+repository = "https://github.com/block/sprout"
+
+[dependencies]
+serde = { version = "1", features = ["derive"] }
+serde_json = "1"
+serde_yaml = "0.9"
+thiserror = "2"
+
+[dev-dependencies]
+tempfile = "3"
--- a/crates/buzz-runtime-config/examples/mcp_launch_mapper.rs
+++ b/crates/buzz-runtime-config/examples/mcp_launch_mapper.rs
@@ -17,6 +17,14 @@ use buzz_runtime_config::McpServerConfig;
 fn redacted_json(doc: &McpLaunchConfigDocument) -> serde_json::Value {
     let mut value = serde_json::to_value(doc).expect("doc serializes");
     for server in value["servers"].as_array_mut().expect("servers array") {
+        // Arguments may carry secrets too (e.g. an API key passed as a CLI
+        // flag), so redact both args and environment values before emitting
+        // a shareable sample.
+        if let Some(args) = server["args"].as_array_mut() {
+            for arg in args.iter_mut() {
+                *arg = serde_json::Value::String("***".to_string());
+            }
+        }
         if let Some(env) = server["env"].as_object_mut() {
             for val in env.values_mut() {
                 *val = serde_json::Value::String("***".to_string());
@@ -46,9 +54,9 @@ fn main() {
             path.display()
         )
     });
-    let doc = McpLaunchConfigDocument::from_runtime_config(&native);
-    doc.validate()
-        .unwrap_or_else(|err| panic!("mapped document is invalid: {err}"));
+    let doc = McpLaunchConfigDocument::from_runtime_config(&native).unwrap_or_else(|err| {
+        panic!("mapped document is invalid (no launch document can be built): {err}")
+    });
 
     let total = native.servers.len();
     let enabled = native
--- a/crates/buzz-runtime-config/examples/mcp_launch_mapper.rs
+++ b/crates/buzz-runtime-config/examples/mcp_launch_mapper.rs
@@ -1,0 +1,76 @@
+//! Map a native Hermes MCP config into the versioned stdio MCP launch
+//! document (buzz-core `mcp_config` v1).
+//!
+//! Usage:
+//!   cargo run -p buzz-runtime-config --example mcp_launch_mapper [PATH]
+//!
+//! `PATH` defaults to `~/.hermes/config.yaml`. The output is the versioned
+//! launch document (JSON) with environment values redacted so it is safe to
+//! paste into a PR comment or log. Pass `--no-redact` to emit full env values
+//! (only if you are launching an agent locally and trust the output channel).
+use std::path::PathBuf;
+
+use buzz_runtime_config::hermes;
+use buzz_runtime_config::launch::{to_launch_json, McpLaunchConfigDocument};
+use buzz_runtime_config::McpServerConfig;
+
+fn redacted_json(doc: &McpLaunchConfigDocument) -> serde_json::Value {
+    let mut value = serde_json::to_value(doc).expect("doc serializes");
+    for server in value["servers"].as_array_mut().expect("servers array") {
+        if let Some(env) = server["env"].as_object_mut() {
+            for val in env.values_mut() {
+                *val = serde_json::Value::String("***".to_string());
+            }
+        }
+    }
+    value
+}
+
+fn main() {
+    let mut redact = true;
+    let mut path: PathBuf = std::env::home_dir()
+        .expect("home dir")
+        .join(".hermes")
+        .join("config.yaml");
+    for arg in std::env::args().skip(1) {
+        if arg == "--no-redact" {
+            redact = false;
+        } else {
+            path = PathBuf::from(arg);
+        }
+    }
+
+    let native = hermes::read_mcp_config(&path).unwrap_or_else(|err| {
+        panic!(
+            "failed to read native Hermes config {}: {err}",
+            path.display()
+        )
+    });
+    let doc = McpLaunchConfigDocument::from_runtime_config(&native);
+    doc.validate()
+        .unwrap_or_else(|err| panic!("mapped document is invalid: {err}"));
+
+    let total = native.servers.len();
+    let enabled = native
+        .servers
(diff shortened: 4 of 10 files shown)


wolfyy970 (changes requested):
I think the read-only adapters are useful, and the clean place for them is target-side discovery. They should let Buzz find what Hermes or Kimi already has, then feed the one launch handoff in #4164. That gives users one tool setup path without making native runtime files portable authority.

I would change two boundaries before merging:

- `McpServerConfig` and `RuntimeMcpConfig` derive `Debug` and `Serialize` while holding raw environment values. `redacted()` is optional, so a normal debug log or serialization leaks the token. Keep the secret-bearing parse type private, give it redacted `Debug`, remove `Serialize`, cap file reads, and expose a separate redacted inventory view.
- #4164 now defines `McpLaunchConfigDocument` and `ConfiguredMcpServer` in `buzz-core`, including stdio and HTTP. I would avoid a second runtime-agnostic MCP contract here. After #4164 lands, map an explicitly selected native entry into that document inside target-owned launch code. Keep `enabled` and other native fields in adapter state.

The portable agent can then carry logical tool requirements. The Project and execution target decide which native connection satisfies them and resolve credentials only for launch. I am happy to help with that join.

wolfyy970:
I put the mechanical security fix in signed commit [`f217f5760`](https://github.com/wolfyy970/buzz/commit/f217f5760) directly on top of your head.

It removes serialization from the secret-bearing config, gives it redacted debug output, adds a separate serializable inventory containing only argument counts and environment names, and caps both file and in-memory parser inputs before YAML or JSON parsing. All 27 crate tests, formatting, and strict Clippy pass.

You can cherry-pick it as-is. I left the adapter-to-#4164 join out of this commit so that boundary can follow once #4164 lands.

whyihaveyou:
Thanks @wolfyy970 for the review — both points are addressed or tracked:

**Point 1 — secret hygiene (done, pushed in `627386e`):**
- `McpServerConfig` / `RuntimeMcpConfig` no longer derive `Serialize` or a value-leaking `Debug`. They keep only `Clone/PartialEq/Eq`, with a hand-written `Debug` that prints `arg_count` + `env_names` and never the `env`/`args` values.
- The raw secret-bearing parse type stays private (the `Raw*` config structs in `hermes.rs`/`kimi.rs`).
- Added a redacted inventory view: `McpServerConfig::redacted()` → `McpServerInventoryEntry` and `RuntimeMcpConfig::redacted()` → `RuntimeMcpInventory`, both `Serialize`-able and guaranteed secret-free.
- Added a 1 MiB file-read cap (`RUNTIME_CONFIG_MAX_BYTES`) on adapter reads, rejected with `ConfigError::TooLarge` before parsing.
- Covered by tests: `debug_output_omits_argument_and_env_values`, `redacted_inventory_omits_argument_and_env_values`, plus oversized-config rejection tests in both adapters.

**Point 2 — join with #4164 (tracked, blocked on #4164 landing):**
#4164 (`McpLaunchConfigDocument` / `ConfiguredMcpServer` in `buzz-core`) is still open, so that contract doesn't exist in `main` yet — I can't map into it without it landing first. The plan in your comment is right: once #4164 merges, this crate feeds the explicitly-selected native entry into `McpLaunchConfigDocument` inside target-owned launch code, keeping `enabled` + native fields in adapter state. I'll set that up as a follow-up PR the moment #4164 is in main. Happy to coordinate the join when it does.

wolfyy970:
I split the shared launch contract out of #4164 into #5349. It defines only the strict, versioned stdio document this PR needs, while native discovery and selection remain target-owned here.

If anything in that schema prevents the mapping you described, please flag it on #5349.

whyihaveyou:
Thanks for the detailed review, @wolfyy970 — both boundaries actioned.

**① Security — resolved, your fix is in.** Your `f217f5760` is now incorporated on this branch as `627386e` (your authorship + Signed-off-by preserved, plus our Co-authored-by per repo DCO). It covers the three points in the review:
- `McpServerConfig` / `RuntimeMcpConfig` no longer derive `Serialize`/`Deserialize` (secret-bearing parse types stay private);
- a custom redacted `Debug` impl omits argument and env values, so a plain `{:?}` log no longer leaks tokens;
- separate redacted inventory views (`McpServerInventoryEntry` / `RuntimeMcpInventory`) carry only what's safe to display/serialize;
- adapter reads are size-capped and oversize configs are rejected before parsing.

**② Architecture — agreed; this is not a second contract, and the seam is now proven.** The launch-document types in the mapper (`crates/buzz-runtime-config/src/launch.rs`) are a deliberate, documented *mirror* of `buzz_core::mcp_config::McpLaunchConfigDocument`, kept local only because buzz-core's type isn't on `main` yet (#5349 is still open). The read-model stays read-only adapter state — `enabled` and other native fields remain in the adapter inventory, as you suggest. Once #5349/#4164 land, the mapper will switch to consume `buzz_core::mcp_config::{McpLaunchConfigDocument, ConfiguredMcpServer}` and the local mirror is dropped.

To show the join works today, I ran the mapper against a real Hermes config and posted the redacted, schema-validated v1 document as a live sample on #5349 (from #5321). It maps 1:1 and validates against your bounded schema (name charset/uniqueness, NUL-free command/args, env-key case uniqueness, no protected keys). Happy to help with the target-owned launch join once #4164 lands.

Would appreciate a re-review whenever convenient.

wolfyy970 (changes requested):
The original secret-boundary fix is good. The new mapper adds two blockers before I can clear the review:

- The local mirror does not enforce #5349’s 16-server, 128-argument, 128-environment, or 64 KiB limits. Its public `Serialize` path can also bypass `to_json()`. It can therefore emit a document while claiming it is guaranteed to pass the shared parser when it is not.
- The example described as safe to paste redacts environment values but prints every argument, and the new launch types derive full `Debug`. Tokens supplied as arguments or environment values can leak again.

The clean fix is to rebase after #5349 lands, delete the mirror, and construct the validated `buzz-core` type directly. Until then, mirror every bound, make invalid documents impossible to serialize, and redact both arguments and environment values in debug and sample output.

I ran formatting, all 34 crate tests, and strict Clippy on `80dfdaf`; they pass. Please add regressions for the missing count/byte limits and argument-secret redaction.

whyihaveyou:
Thanks @wolfyy970 — both new blockers are addressed, pushed as `1354b5c` on this branch.

**① The mapper now enforces every #5349 bound at construction, so an invalid document cannot exist.** `McpLaunchConfigDocument::from_runtime_config` now returns `Result` and validates up front: 16-server count, 128-arg, 128-env, and the 64 KiB bound (accumulated over name + command + args + env bytes before encoding, mirroring your add_string_bytes). `to_json` additionally re-checks the 64 KiB limit on the encoded bytes. Since a document is only ever produced through a validating constructor, the public Serialize path can no longer emit a document that claims to validate but does not — construction is the gate.

Your clarified boundary semantics are mirrored verbatim:
- Invalid native names are rejected, never auto-sanitized into aliases (that stays an explicit-rename concern upstream);
- Invalid environment keys are rejected outright — never rewritten — and the error reports the 1-based entry index rather than echoing the (possibly secret) key;
- Exceeding any limit rejects the whole document; an enabled server is never silently dropped.

**② Debug and sample output are now redaction-safe.** The derived Debug on `LaunchStdioServer` is replaced with a manual impl that prints only name + arg_count + env_count — neither args nor env values. The `mcp_launch_mapper` example's "safe to paste" output now redacts arguments and environment values (each as ***), closing the args leak.

**Regression tests added** (crate now 41 tests): too-many servers / too-many-args / too-many-env rejected; oversized document rejected before encoding; Debug and sample output never leak arg/env secrets; invalid env key reported without echoing the key. Verified locally: cargo test -p buzz-runtime-config green, workspace fmt --check clean, cargo clippy --workspace --all-targets -D warnings clean.

As agreed, the clean follow-up (once #5349 lands, drop this mirror and construct buzz_core::mcp_config::McpLaunchConfigDocument directly) stays parked — this PR only hardens the interim mirror. Would appreciate a re-review when convenient.

wolfyy970 (changes requested):
Most of the requested work is closed. The count limits, invalid-key redaction, argument redaction and manual Debug implementation are good. All 41 crate tests and strict Clippy pass at 1354b5c.

One serialization bypass remains. from_runtime_config calls validate(), which caps the raw string bytes, but it does not check the encoded JSON size. The type still exposes Serialize. An argument such as a 40,000-byte string of newline characters is below the raw 64 KiB limit, so construction succeeds, but JSON escaping expands it beyond 64 KiB. A caller using serde_json::to_vec directly can therefore emit a document that #5349 rejects without passing through to_json().

Please make construction verify the encoded size too, or remove the unrestricted Serialize path. Add a regression using escaping expansion, not another raw string larger than 64 KiB. After that, my requested changes are clear.

whyihaveyou:
@wolfyy970 serialization bypass closed — pushed as `a5e0a4c` on this branch.

Per your review: `from_runtime_config` → `validate()` now enforces the **encoded** JSON size, not just the raw string byte sum. Since `validate()` is the only gate a `McpLaunchConfigDocument` can pass through at construction, no over-limit document can be instantiated through any path — closing the `Serialize` bypass you flagged.

- Moved the encoded-size check into `validate()` (shared by construction and `to_json`).
- Added a regression test using **escaping expansion** (40,000 raw newline bytes — under the 64 KiB raw bound, but each expands to `\n`, pushing the encoded document past it). Construction now rejects it.

All 42 crate tests and strict Clippy pass at `a5e0a4c`. Ready for re-review. Thanks for the sharp catch.

wolfyy970 (approved):
The encoded-size bypass is closed at a5e0a4c. validate() now checks the actual serialized bytes, and the private fields keep construction behind the validating mapper. The escaping regression reaches the path I asked for.

I reran all 42 crate tests, strict Clippy, formatting and the diff check. Approved.

PR #5323, opened 2026-08-08
Revert "fix(acp): reject unattended permission requests"

wesbillman:
Reverts block/buzz#4609


wpfleger96 (approved):

wolfyy970:
@wesbillman The revert restores agent function, but it also makes `bypassPermissions` the hidden default again. I put a small signed patch directly on top at [`bbc6f7572`](https://github.com/wolfyy970/buzz/commit/bbc6f7572).

It keeps `bypassPermissions` available as an explicit legacy choice while defaulting to the adapter's per-request path. Buzz still selects `allow_once` unattended for this emergency, but each grant remains visible to the harness and #5106 can replace that behavior cleanly. Goose receives no unsupported mode override.

The regression drives a real fake adapter through the prompt wire, captures the exact permission response, and proves the turn completes. All 688 ACP tests, strict Clippy, formatting, and the adversarial pass are clean. You can cherry-pick this before merge.

wolfyy970:
@wesbillman @wpfleger96 The revert merged before the cherry-pick. I restacked the exact reviewed commit onto the new main as #5325. It is one signed commit, passes the full ACP gate locally, and keeps #5106's policy path open without delaying the recovery.

PR #5324, opened 2026-08-08
fix(desktop): prevent horizontal clipping in Prompt Context modal

wpfleger96:
The Prompt Context modal (observer feed → check icon under sent messages) was clipping all content and card right-padding at the dialog edge.

**Root cause**: `PromptContextDialog` renders inside `DialogContent`, which is a CSS grid. The child flex wrapper had default `min-width: auto`, so the widest unbreakable token in the content (64-char hex event IDs, `Tags: [[...]]` JSON) set the grid track width, blowing it past `max-w-xl`. `overflow-hidden` then clipped everything at the dialog edge — including the section cards' right padding.

**Fix**:
- `AgentSessionTranscriptList.tsx`: add `min-w-0` to the `flex max-h-[85vh] flex-col` wrapper so the grid item can shrink below its max-content width.
- `PromptSectionAccordion.tsx`: replace `wrap-break-word` with `wrap-anywhere` on the body text (open and collapsed states) and the title. `overflow-wrap: anywhere` reduces min-content width, which `break-word` does not, letting long tokens wrap inside the cards rather than inflating the track.

The `line-clamp-2` collapsed preview is preserved unchanged.

### DIFF
--- a/desktop/src/features/agents/ui/AgentSessionTranscriptList.tsx
+++ b/desktop/src/features/agents/ui/AgentSessionTranscriptList.tsx
@@ -747,7 +747,7 @@ function PromptContextDialog({
   return (
     <Dialog onOpenChange={onOpenChange} open={open}>
       <DialogContent className="max-w-xl overflow-hidden p-0">
-        <div className="flex max-h-[85vh] flex-col">
+        <div className="flex min-w-0 max-h-[85vh] flex-col">
           <DialogHeader className="px-6 pb-3 pt-5 pr-14">
             <DialogTitle>Prompt context</DialogTitle>
             {setupText ? (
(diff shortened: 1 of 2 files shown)


wesbillman (approved):
Reviewed on Wes's behalf. The min-width and overflow-wrap changes correctly address long unbroken Prompt Context content without expanding scope. No blocking findings. Tests are unchanged; given the narrow CSS-only fix, I don't consider the lack of a new visual regression test blocking.

wpfleger96:
### Before — horizontal clipping on `origin/main`
Content and card right-padding clipped at dialog edge. Long hex IDs and Tags JSON overflow the grid track; `overflow-hidden` on `DialogContent` chops everything at the right margin.

01-before

### After — fixed on `duncan/prompt-context-clipping`
All sections fit within the dialog. Right padding is intact. Long hex IDs and Tags JSON wrap inside the cards.

02-after

Issue #5317, opened 2026-08-08
Local Claude Code agents cannot execute any tool: permission rules are never consulted


natebitsats:
# Local Claude Code agents cannot execute any tool: permission rules are never consulted

## Summary

Local agents using the `claude-agent-acp` runtime cannot execute a single Bash command. Project-scoped Claude Code permission rules (`<cwd>/.claude/settings.json`) are never consulted, so there is no supported configuration in which a local agent can run a tool.

Because posting to a channel requires `buzz messages send` (a CLI call), the practical effect is that **local Claude Code agents cannot participate in channels at all**. They receive mentions, run a turn, and their reply reaches only the local terminal/activity view.

## Environment

| | |
|---|---|
| Buzz | 0.5.7 (`Buzz_0.5.7_aarch64.dmg`) |
| Platform | macOS 15 (Darwin 24.1.0), arm64 |
| Runtime | `claude-agent-acp` 0.66.0 |
| Node | v24.18.0 (bundled) |
| Agent cwd | `~/.buzz` (confirmed via `lsof` and the agent's own Read tool) |

## Reproduction

1. Create a local agent using the Claude Code harness.
2. Create `~/.buzz/.claude/settings.json`:
   ```json
   {
     "permissions": {
       "allow": ["Bash(buzz messages get:*)"]
     }
   }
   ```
3. Start the agent and @-mention it, asking it to run
   `buzz messages get --channel <uuid> --limit 1`
4. The command is denied.

## Observed

With `permission_mode=dontAsk` (the default):

```
Permission to use Bash has been denied because Claude Code is running in
don't ask mode.
```

With `BUZZ_ACP_PERMISSION_MODE=default`:

```
User refused permission to run tool
```

Across every agent transcript in `~/.claude/projects/-Users-natebragg--buzz`:
**17 Bash tool calls, 17 denied, 0 allowed.** Includes bare, undecorated commands
that match an `allow` rule exactly — so this is not a compound-command or
prefix-matching artifact.

## Expected

A command matching a `permissions.allow` rule should execute without a
permission request, in either mode.

## Analysis

- `dist/acp-agent.js` defaults to `settingSources: ["user", "project", "local"]`,
  and the agent's cwd is `~/.buzz`, so `~/.buzz/.claude/settings.json` should be
  loaded as project settings.
- `buzz-acp` passes no `settingSources`, `allowedTools`, or `canUseTool` override
  that I can find in the binary.
[...]

PR #5325, opened 2026-08-08
fix(acp): preserve permission requests by default

wolfyy970:
#5323 and Desktop 0.5.8 restored working agents by restoring `bypassPermissions` as the implicit default. This follow-up removes that implicit blanket bypass while keeping it available when an operator chooses it.

The default now leaves each adapter in its normal per-request flow. For current Claude and Codex adapters, the harness selects the offered `allow_once`, so agents can work and publish without a stranded human prompt. Each grant remains visible to the harness, and Goose receives no unsupported mode override.

This is the safer default after the immediate recovery for #5290, #5303, #5317 and #5332. #5106 remains the full owner-facing Ask, Allow and Reject policy.

The regression drives a real fake adapter through the prompt wire, captures the exact permission response and proves the turn completes. The focused permission suite, ambient-environment regression, formatting, strict Clippy and adversarial review pass.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -2359,6 +2359,66 @@ mod tests {
         assert_eq!(reject_once.unwrap()["optionId"].as_str(), Some("rej-x"));
     }
 
+    /// Exercise the production prompt read loop rather than duplicating its
+    /// option-selection logic. In the default adapter mode, the adapter asks
+    /// for each permission and Buzz must answer that exact request before the
+    /// prompt can complete.
+    #[cfg(unix)]
+    #[tokio::test]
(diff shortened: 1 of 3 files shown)


wolfyy970:
@wesbillman I rechecked this exact head against the 0.5.8 main. It is mergeable and the only ACP overlap is the revert this builds on.

This keeps agents working without making blanket `bypassPermissions` the implicit default again. The adapter asks per operation, the harness selects the offered `allow_once`, and Goose gets no unsupported mode override. The production wire test passed three consecutive runs, the permission suite and ambient-environment regression pass, and the adversarial review is clean.

It is blocked only on human review now. Could you take a look when you have a window?

wolfyy970:
Closing this. No review.

PR #5328, opened 2026-08-08
feat(desktop): merge agent edit surface into a single dialog

wpfleger96:
Merges the two agent edit dialogs (`AgentInstanceEditDialog`, `AgentDefinitionDialog`) into a single `AgentEditMergedDialog` that every edit entry point — the agents library, the profile panel, owner review, and the `requestOpenEditAgent` event path — reaches through `AgentEditDialog`. `AgentInstanceEditDialog` is deleted, `AgentDialog`'s `instance-edit` arm is removed, and the `editPersonaDialogState` export is gone.

### Design

- **One canonical form model.** `seedAgentFormModel` builds an `AgentFormModel` from the edit context (linked instance, unlinked instance, or definition-only). The model owns every editable field, including the harness pin (`harnessInherit` / `harnessCommand` / `harnessArgs` / `acpCommand`). `emitAgentFormDiff` diffs the seed against the submitted model and routes each changed field to its owning layer, returning `{ personaInput, agentInput, policySets }`. There is no parallel React state model and no post-emit merge.
- **Single field-ownership authority.** The `FIELD_OWNERS` map drives emit routing (`emitAgentFormDiff`), per-field editability (`fieldEditable`), and the catalog-publish dirty signal (`definitionFieldsDirty`). No control decides ownership locally.
- **Save coordinator with observed-state settlement.** `runAgentSaveCoordinator` applies definition → instance → local-policy writes in order. After every boundary — success or throw — it re-fetches the store and lets observed persistence decide the outcome: a write that threw after landing on disk counts as persisted and the sequence continues; a write that returned but did not persist counts as failed. A submitted behavior group (`respondTo` / `respondToAllowlist` / `parallelism`) settles as the full-replacement unit the backend writes, so a clear the backend failed to apply cannot false-succeed.
- **Independent definition/instance runtime state.** Definition-owned (`dModel` / `dProvider`) and instance-owned (`iModel` / `iProvider`) runtime state are separate, with per-owner selection handlers. An instance pin change cannot clear the definition's model or provider, and a definition runtime change cannot touch the instance overlay.
- **Definition behavior fields are first-class.** Access, allowlist, and parallelism render in the definition section; an unset `respondTo` displays as owner-only (the effective default) and is never coerced to "anyone". Blanking parallelism in a definition-only edit emits an explicit clear rather than silently restoring the stored value.
- **Team-managed definitions are read-only.** `fieldEditable` disables every definition-owned control and the dialog shows a "Managed by team «name»" notice; instance-owned fields, including access, stay editable. A runtime-less definition with a saved model but no provider exposes the provider picker and keeps Save disabled until a provider is chosen.
- **Reasoning effort in one place.** The `EffortPicker` renders in the merged dialog's instance section, so the reasoning-effort control lives on the same surface as every other editable field rather than in a separate dialog.
- **Deep-link focus lands on the requested field.** `useAgentEditDeepLinkFocus` resolves a `requestOpenEditAgent` focus target after the dialog mounts: it expands the collapsed env-key section when the target is a credential field, suppresses the dialog's open-autofocus so it cannot steal focus from a deep-link target, and retries once the target control is present. Focus reaches `model` and `env_key` targets that the pre-merge path left unfocused. The `EnvVarsEditor` focus effect keys on the required-key set's content rather than its array identity, so a content-equal re-render while model discovery and file-config queries settle no longer cancels the pending focus frame and strands the `env_key` target.

### Out of scope (follow-up work)

- Create-flow unification
- Persona → agent vocabulary rename (profile "Instances" section, `src-tauri` module names)


wolfyy970:
I do not think the failing E2E tests should be updated around this head. They are catching real lost behavior.

`AgentEditDialog` never calls `seedAgentFormModel` or `emitAgentFormDiff`. It submits both save inputs as `null` in every context, so an instance-only edit can report success without writing anything. The routed surface also removes the existing run-on, access, custom command, tuning and local policy controls. For linked agents, it writes the shared definition without showing which agents will change.

I restored the existing production routes while leaving the new model aside. The 33 affected smoke tests, the catalog publication integration, typecheck and all 4,540 unit tests pass. The flat model still needs an ownership fix too: linked respond-to, allowlist and parallelism are emitted to the instance, while definition-only changes to those fields are not emitted.

I would split this work. Keep the existing routes, then land the field model once it represents shared template state and per-agent state separately. Route one edit path at a time with parity tests. The product can still say “Edit agent” and “Edit its template”; the storage boundary does not need to become user-facing.

wolfyy970:
Thanks for turning this around quickly. At `6f0d07f63`, the follow-ups restore real instance saves and the missing run-on, access, command, tuning, environment and restart controls. The 25 focused tests, typecheck and all 4,556 unit tests pass here.

Three behavior blockers remain:

- A linked profile now always opens `AgentInstanceEditDialog` and explicitly removes `onEditLinkedPersona`. Its shared prompt, avatar and template are no longer editable from that profile. The user needs **Edit this agent** or **Edit its template** because the second choice can affect other agents.
- `reviewOverrides` drops `request.respondTo`, which the removed `updateInputFromRequest` preserved. A respond-to-only request opens with no requested change.
- A definition edit still shows neither the affected-agent count nor their names before save.

Please add route-level regressions for the linked-template task and the agent-requested `respondTo` path.

wolfyy970:
At the current head, two concrete problems from my last pass are fixed: respond-to requests are carried through, and linked profiles can edit definition fields again. The 57 focused tests and typecheck pass here.

The remaining blocker is now clearer. In an instance-with-definition context, the form deliberately renders definition, instance and local fields together, then one Save writes all three boundaries. Changing the name, prompt or definition runtime changes the shared definition. Changing access, the harness pin or launch settings changes only this agent. The user is never asked which object they mean to edit, and linkedInstanceCount appears only in the delete copy.

Field ownership in code does not solve that user problem. These are two actions:

- Edit this agent
- Edit its template, with the affected agents shown before save

The components can be shared, but a linked definition change and an instance change should not be combined silently into one edit transaction. Please add a route-level E2E test proving the user chooses the object and sees the impact before a definition save.

wolfyy970 (changes requested):
Round 8 improves field ownership, but the product blocker from my last review is unchanged. A linked agent still opens one form with definition and instance controls and one Save. There is no choice between Edit this agent and Edit its template, and the affected count and names still appear only in delete copy. The new E2E checks control visibility, not that flow.

I also found three correctness problems in definition-only editing:

- An unset respond-to policy renders as Anyone, although minting resolves unset to Owner only.
- Parallelism accepts 0, negatives, or values over 32. Save stays enabled; invalid low values are silently replaced with the old value, while high values fail in the backend.
- Returning a manually changed runtime to the auto-seeded default can drop that deliberate choice because the auto-seed marker never clears.

Please split agent and template actions, show the impacted agents before a template save, and add route-level tests for these defaults and validation.

wolfyy970 (changes requested):
At `8096c37a2`, the access default and runtime reselection bugs are fixed, and the R6 test now reaches the real dialog. Thanks for closing those.

Three blockers remain:

- A linked agent still opens one form where one Save can change both the agent and its shared definition. There is no choice between **Edit this agent** and **Edit its template**, and affected agents are not shown before the template write. The new E2E now locks in that combined transaction.
- Definition parallelism has no maximum or visible validation. Blank, zero and negative input is silently replaced with the saved value, while values above 32 reach backend rejection.
- If the same definition changes while this dialog is open, the form stays stale. Submit then seeds from the newer definition, combines it with stale form values and can overwrite the concurrent update. Please detect version drift or rebase the edit, with a regression test.

I am keeping changes requested at this head.

wolfyy970 (changes requested):
At `8cc6e5a11`, blank definition parallelism now reaches the wire as a clear, and the policy settlement change looks right. That closes the clearing defect.

Three blockers remain:

- The linked route still combines agent and template edits behind one Save. There is no object choice or affected-agent disclosure, and the E2E still requires one save to change both layers.
- Definition-only parallelism still bypasses validation. The input has no maximum, and nonblank zero or negative values silently revert to the saved value. Please cover `0`, negatives, `32` and `33` through the route.
- A same-definition update arriving while the dialog is open can still be overwritten. The form stays stale, submit combines it with the latest definition seed, and the backend replaces the definition without revision checking.

I am keeping changes requested at this head.

wolfyy970 (changes requested):
I re-reviewed `b25ac47cf8`. The observed-state save coordinator is a real improvement, but the three blockers from the last head remain.

- The linked route still puts shared definition fields and instance fields behind one Save. The E2E now explicitly requires one save to change both layers. The linked-agent count appears only in delete copy, so the user never chooses **Edit this agent** or **Edit its template** and never sees who a template edit affects.
- Definition-only parallelism still bypasses the save gate because validation runs only when an instance is present. Zero and negative values silently fall back to the saved value; 33 reaches backend rejection. Please enforce 1 through 32 in the form and cover 0, negative, 32 and 33 through the route.
- The coordinator refetches only after writing. If the definition changes while the dialog is open, it submits the stale form without an expected revision and can overwrite the newer definition. Please reject or rebase that edit before the first write.

I am keeping changes requested at this head.

wolfyy970 (changes requested):
Withdrawn.

wesbillman (changes requested):

One correctness blocker remains at this head:

- **Handle verification-refetch rejection after every write boundary.** `runAgentSaveCoordinator` awaits `refetchStores()` directly after the definition, instance, and policy writes and again during final settlement. If any refetch rejects after a write committed, that rejection escapes the coordinator. `useAgentEditMergedSubmit` has only `try/finally`, so the UI merely stops showing “Saving...” while the dialog remains open without a concrete error explaining that persistence is unknown. The user can then retry blindly after a partial durable write. Catch verification failures, keep the dialog open, report that saved state could not be verified and that reopening/retrying is required, and cover refetch rejection after each write boundary plus thrown-but-persisted mutations.

The merged edit routing, field ownership/diff model, authorization boundaries, docs, and migrated test surfaces were otherwise clear in read-only review. CI is passing or intentionally skipped.

wesbillman (changes requested):

Changes requested on exact head `f361a4ebc450235f332f9042bb4f119521cf69e4`:

1. **[P1] Make verification refetch failures observable in the production adapter.** `useAgentEditMergedSubmit.ts:231-234` awaits `queryClient.refetchQueries(...)` without `{ throwOnError: true }`. TanStack Query v5 catches refetch failures by default, so `verifiedRefetch()` does not see a rejection; the adapter reads retained cache and can misclassify a possibly committed write as an observed failure. The new coordinator tests inject a rejecting `refetchStores` directly and therefore do not cover this production seam. Please enable error propagation for both refetches and add an adapter-level regression test. This leaves the prior post-write verification blocker open despite the coordinator wrapper.

2. **[P1] Prevent stale open forms from overwriting concurrent definition updates.** `AgentEditMergedDialog.tsx:278-360` reseeds only for open/entity identity, not a same-definition revision change. Submit then builds its baseline from the latest `s.ctx` (`useAgentEditMergedSubmit.ts:216-223`) while the controlled form still contains the older snapshot. A background refresh during editing therefore makes untouched stale fields appear user-edited; the full `UpdatePersonaInput` can replace another writer’s newer values, and it carries no expected revision. Reject when `updatedAt` drifted before the definition write, or rebase only fields the user actually dirtied, with a regression test covering a same-ID refresh while the dialog is open.

The merged routing, field ownership, deep-link behavior, and definition → instance → policy sequencing otherwise look coherent in the exact-base/head diff. Relevant Desktop CI is green. Review was source/diff-only as required; PR code was not executed.

wesbillman (changes requested):

Requesting changes on exact head `b2a4d8caaccaf2770b4ece3b7cc0cce2f15d291e`.

- **P1 — Relay Mesh edits cross the definition/instance ownership boundary.** In `useAgentEditRuntimeHandlers.ts:209-228`, selecting `relay-mesh` from the definition-owned provider control always calls `handleRuntimeDropdownChange("buzz-agent")`. That handler (`:171-198`) mutates the instance runtime state: it marks the runtime touched, clears inheritance, assigns the Buzz Agent command/args, and updates the I-layer selection. It does not change `definitionRuntimeId`. In a linked edit, Save therefore retains the definition's old runtime while unexpectedly pinning the clicked instance to Buzz Agent, even though the provider/model changes are routed to the definition. Other linked instances can then inherit the incompatible old-runtime/Relay-Mesh definition. Route this branch through `handleDefinitionRuntimeChange("buzz-agent")` when `showDef`, keeping the instance handler for instance-only context, and add a linked-context regression proving the D runtime changes without modifying the instance pin.

- **P2 — Throw-after-commit rename success uses the stale name.** `agentSaveCoordinator.ts:231-249` correctly treats a thrown instance mutation as successful when refetch observes the persisted update, but `:362-363` prefers `latestAgent`, initialized from the pre-save instance, over `observedAgent`. A committed rename from Alice to Bob can consequently report “Alice saved.” Prefer the observed agent name, as the stopped-state path already does at `:375`, and cover the thrown-but-persisted rename case.

The new head does close the two prior P1s: production refetches now reject on verification failure, and definition writes compare the pinned seed revision against live context. I also traced the merged dialog across linked, instance-only, definition-only, profile, library, owner-review, deep-link, and team-managed routes. Product CI is green; the cancelled Codex security-review job is process metadata, not an additional product finding.

wesbillman (changes requested):

Changes requested on exact head `e2e0fe627166a9b6e541c228468825468d0ae06c`.

- **P1 — The concurrent-definition guard can still overwrite a newer definition because it does not read authoritative state before writing.** `useAgentEditMergedSubmit.ts:267-297` builds the full-replacement definition input from the form and passes the seed-time revision into the coordinator. `agentSaveCoordinator.ts:161-170` compares that revision only with `def.updatedAt` from the current React `ctx`; the first forced store refetch is after `updatePersona` at lines 198-217. If client A saves while client B's dialog remains open, B can submit before its cached query receives A's update. Its cached `def.updatedAt` still equals the seed revision, the guard passes, and B replaces A's newer fields. Re-resolving a selected persona from query data helps only after that cache has already refreshed; it is not concurrency control. Fetch and compare the current persisted revision before the first definition write, or enforce compare-and-swap/versioning in the write itself, and add a two-writer regression where B's cache is stale when it submits.

The field-ownership model, definition → instance → policy sequencing and observed-state settlement, merged routing/deep-link behavior, and runtime/provider transitions otherwise look coherent. Exact-head product CI is green; the cancelled Codex review job is not a product failure. Review was source/diff-only as required; PR code was not executed.

wesbillman (changes requested):

Changes requested on exact head `696222fcd0f26e73929cf9bc1a0dc4b71e01937f`.

- **P1 — “Save and publish” can silently degrade to save-only and close as success.** `desktop/src-tauri/src/commands/personas/update.rs:207-212` advances `updated_at` and durably saves the persona before the strict retention callback runs. In `personas/sharing.rs:80-90`, `prepare_persona_publication(...)?` can then fail before the catalog event is retained. The frontend catches that error, but `agentSaveCoordinator.ts:236-273` clears the failure whenever refetch observes the persona fields on disk; `:407-443` then passes `publicationStatus === null` to `personaSaveNotice`, shows only “Updated …”, closes the dialog, and returns success. The requested catalog update was neither published nor queued. Prepare/retain before committing, make the operation recoverable, or preserve the publication failure through settlement; add a command-path plus coordinator regression for strict preparation failing after the persona save.

- **P1 — Clearing the visible instance name can no-op or rename it to the shared definition without warning.** Save validity checks only the definition/shared `displayName` (`AgentEditMergedDialog.tsx:556-573`), although the visible instance field edits `instanceName`. Submit maps blank to `undefined` (`useAgentEditMergedSubmit.ts:134-159`), and `agentFormModel.ts:541-544,616-620` then falls back to `displayName`. An unlinked agent therefore reports success without saving the requested blank value; a linked agent without a name pool can unexpectedly inherit the definition name. Require and validate the visible instance name, and do not reinterpret a user-cleared instance field as the definition name.

- **P2 — The merged surface does not expose ownership clearly or accessibly.** Linked edits render two controls both labeled “Agent name” for different objects (`AgentEditMergedDialogDSection.tsx:232-257`, `AgentEditMergedDialogInstanceSection.tsx:219-244`) without saying which affects the shared definition versus this deployed instance. The definition Advanced disclosure also omits `aria-expanded` and an associated controlled region (`AgentEditMergedDialogDSection.tsx:455-480`), unlike the instance disclosure. Give the controls explicit scope/blast-radius copy and restore disclosure semantics.

The remaining save sequencing, lock-held CAS, observed-state settlement, field ownership, and legacy parity were coherent in read-only source/diff review. Product CI is green; the cancelled Codex review job is process metadata, not another product finding. PR code was not executed.

wesbillman (changes requested):

Changes requested on exact head `46053a730aac1746b3265b2488d215cd32b97275`, against base `00e61eafa917d296104006576b7a2ddbfd58bb5a`.

The instance-name validation/emission and scoped labels/Advanced accessibility findings from the previous review are resolved. The merged-dialog product direction is unchanged. Publication recovery still has the following defects:

### P1 — Combined saves bypass publication recovery

`desktop/src/features/agents/ui/agentSaveCoordinator.ts:428–430` only handles `publishFailed` when there is no instance/policy remainder. Reproduce with a shared linked definition: change its prompt and an instance setting (or launch policy), then make strict publication preparation fail after `save_personas` commits but before the new event is retained (`personas/update.rs:207–212`; `personas/pending.rs:173–200`). The coordinator records `publishFailed` and `firstError` (`288–294`), skips instance/policy writes (`299,328`), and therefore bypasses the new `publishRetry` call. The partial warning says the profile saved and to reopen/retry (`514–519`), without preserving the unpublished outcome. Reopening seeds the already-saved definition; retrying only the remaining instance/policy changes emits no `personaInput`, so publication is never retried and this new head has no pending row to flush.

Settle publication independently of the other unsaved parts. Attempt the publish-only recovery for a persisted definition even when I/L writes remain, and report any unrecovered publication explicitly alongside those remaining changes. Add post-persist strict-preparation-failure cases with both an instance change and a policy change, including recovery success/failure. The new tests currently exercise the post-persist publication retry only in definition-only context.

### P1 — New unlocked team refresh can undo an explicit unshare

`desktop/src-tauri/src/commands/personas/sharing.rs:119–122` runs `refresh_for_persona_at` after releasing the managed-agent store lock. That helper reads the current shared team head, builds a `shared=true` replacement, and retains it without a transaction or a fresh visibility check (`teams/pending.rs:325–365`). A concrete interleaving: after a saved member prompt changes, this retry refresh reads shared head T; concurrent `set_team_shared(false)` acquires the store lock and retains unshared T+1; the refresh then retains its shared replacement at T+1. `retain_event` accepts equal timestamps (`managed_agents/retention.rs:213–222`), replacing the pending unshare. If the toggle's flush starts afterward, it publishes the shared replacement instead of the requested unshare. This can leave the team discoverable after the owner removed it.

Serialize the refresh's authoritative store/head reads and retain/retract operation with the same store lock as team edits/unshare/delete, without holding it across the network await. Add a controlled concurrent unshare regression at the production retry boundary. The filesystem seam's sequential pre-/post-fold tests do not cover this interleaving.

### P2 — Successful publish-only retry hides skipped linked identity updates

The new success path at `agentSaveCoordinator.ts:444–451,474–510` treats a successful `setPersonaShared` retry as complete recovery. For a shared definition name/avatar edit, strict retention failure occurs after the definition save but **before** linked instance name/avatar propagation and relay profile sync (`desktop/src-tauri/src/commands/personas/update.rs:209–217,227–289,304–324`). If the transient failure clears, the publish-only retry publishes the definition and refreshes teams (`personas/sharing.rs:110–122`), but never performs those skipped effects. The dialog then reports success and closes while linked identities retain the old name/avatar. A no-op reopen/save cannot recover the original name transition.

Keep local dependent effects independent of a failed enqueue, or preserve an explicit partial outcome until they are completed; do not claim full recovery from the catalog-only retry. Cover a definition-only name/avatar edit with a linked instance, a post-save strict-retain failure, and a successful publication retry, asserting the dependent identity state as well as the catalog event.

Validation: exact-head source/diff and regression-test inspection only; no checkout, build, tests, or PR-code execution. All three delegated lanes returned and were integrated. This is a bounded re-review of the agreed merged edit/save contract and changed recovery paths, not a request to restore separate dialogs or expand create-flow scope.

wesbillman (changes requested):

Changes requested on exact head `08467d5824272f10055d82df42c3f1f67f4c4e68`, against base `bd73490418266f267d9bb3bdf13e64582adc8e80`.

The prior combined-save publication recovery and team-refresh/unshare locking defects are fixed in the inspected production paths. Linked local name/avatar persistence also now runs before a strict-retain error is returned. The merged-dialog product direction remains unchanged. Two blockers remain:

### P1 — Unrelated definition edits silently erase the authored description

At [`agentFormModel.ts:496–505`](https://github.com/block/buzz/blob/08467d5824272f10055d82df42c3f1f67f4c4e68/desktop/src/features/agents/ui/agentFormModel.ts#L496-L505), the merged form emits a replacement definition without `description`. The form does not seed or expose that field either. Both update commands pass this payload through [`tauriPersonas.ts:113–133`](https://github.com/block/buzz/blob/08467d5824272f10055d82df42c3f1f67f4c4e68/desktop/src/shared/api/tauriPersonas.ts#L113-L133), where an omitted description becomes null. Rust assigns that null to the definition and saves it (`personas/update.rs:218,264,283`). Settlement never compares description, so the deletion is reported as a successful save.

Reproduce with an editable definition that already has a nonempty public description: open the merged editor, change only its prompt, and save. The description is cleared locally and can be published as cleared to the catalog and linked profiles. Both definition-only and linked-definition saves are affected; I/L-only saves do not issue this definition write. This is a concrete base-to-head regression: at the exact base, `personaDialogState.ts:121–126` preserves the description and `AgentDefinitionDialog.tsx:209–211,361–384,767–773` seeds, submits, and exposes it. This PR reroutes those edits to the merged form.

Carry description through the merged form's seed, definition ownership/diff, editing control, replacement payload, and settlement. Cover preservation on an unrelated definition edit through the real adapter for both plain save and save-and-publish, plus explicit description edit/clear behavior. Restoring a separate dialog is not requested.

### P2 — Strict-retain failure still skips linked relay-profile sync

[`personas/update.rs:372–376`](https://github.com/block/buzz/blob/08467d5824272f10055d82df42c3f1f67f4c4e68/desktop/src-tauri/src/commands/personas/update.rs#L372-L376) propagates `retain_result?` before the phase-2 `sync_managed_agent_profile` loop at lines 383–401. The corrective change now persists linked local records, but discards the collected kind:0 sync work on this failure path.

Reproduce with a shared definition and a linked instance with valid agent keys: edit the definition's name/avatar and make strict publication preparation fail after the definition save; let the immediate catalog-only retry succeed. `set_persona_shared` republishes the persona and refreshes team catalog heads, not linked kind:0 profiles (`personas/sharing.rs:44–81,122–160`). The coordinator then reports success and closes (`agentSaveCoordinator.ts:484–550`), while other clients still see the old name/avatar. Boot or a later successful start can reconcile this from the corrected local records (`commands/agents.rs:1008–1032`, `managed_agents/restore.rs:431–485`); this is stale relay metadata until reconciliation, not permanent identity loss. That later lifecycle repair does not complete the current save's skipped work.

Keep the captured retain result intact across the blocking return, release the store lock, run the existing best-effort phase-2 sync, and only then propagate the retain error. Preserve the original strict publication failure for the coordinator. Add a production-path regression with valid agent keys and a relay observer that asserts kind:0 sync is attempted even when retention fails. The new linked-instance test only checks the local name and seeds an empty private key (`update/concurrent_edit_tests.rs:446,535–548`), so it cannot witness this phase.

Non-blocking test notes: the new refresh observer checks the lock at its probe, not an actual concurrent unshare; and the combined-save failure test overrides `updateManagedAgent` without incrementing the counter it asserts (`agentSaveCoordinator.test.mjs:2025–2043`). Neither is an additional production blocker.

Validation: all three delegated lanes returned and were independently integrated. Exact-base/head/prior source, diffs, and regression-test inspection only. No checkout, builds, tests, or PR-code execution; no fresh CI-pass claim. This is a bounded re-review of the three agreed recovery fixes plus the concrete description merge regression, not a fresh review of unrelated incoming-main features.

desktop/src/features/agents/ui/agentSaveCoordinator.ts
```diff
@@ -0,0 +1,736 @@
...
+function observedStateMatchesPersonaInput(
+  observed: AgentPersona,
+  submitted: UpdatePersonaInput,
+): boolean {
+  // Required fields
+  if (observed.displayName.trim() !== submitted.displayName.trim())
+    return false;
+  if (observed.systemPrompt.trim() !== (submitted.systemPrompt ?? "").trim())
+    return false;
+  // Optional fields — only compare when submitted
```
wesbillman, line 585: [P2] Observe description before accepting this definition as persisted

The correction emits `description`, but this equality helper still ignores it. For a description-only edit containing U+200B, the control/adapter preserve the text and `normalize_description` rejects it before any write. All fields checked here still equal the old record, so both settlement passes report persistence: ordinary Save closes as success; Save-and-publish may republish the old record and then close. Compare description with backend-compatible accepted storage semantics, including undefined-as-clear, and cover rejected edits/clears through the coordinator so they keep the dialog open and stop I/L advancement. Do not trim prohibited bytes into a false match.

wesbillman (changes requested):

## Changes requested: one P2 remains

Reviewed head `d739ba791332a9bb7a628593dbce64b290ecb66f` against base `5aed49b505a7e27f3b0e34dafa53d6c4e8cdcd64`. This is a bounded corrective re-review; the single merged editor and D → I → L save contract remain the product direction.

**[P2] Include description in observed save settlement.** `observedStateMatchesPersonaInput` never compares `description`. Change only an editable definition’s description to text containing U+200B. The input/adapter preserve it, and Rust rejects it before writing (`personas/update.rs:218`; `definition_validation.rs:50–60,89–107,159–178`). The unchanged definition nevertheless passes the coordinator’s per-boundary and final comparisons (`agentSaveCoordinator.ts:279–288,414–423`), so ordinary Save reports success and closes, discarding the draft. Save-and-publish instead treats that pre-write rejection as a publication failure, can republish the old record, and then closes successfully; combined edits can advance to I/L despite the unsaved description.

Compare description using the backend’s accepted canonical storage semantics, including absent/empty as explicit clear, without normalizing prohibited bytes into apparent success. Add production-seam regressions for rejected/nonpersisted description-only edits and clears on both ordinary and publish paths: no close, no old-record publication retry, and no I/L advancement. Also cover successful normalization and preservation on unrelated definition edits. The model, submit, and coordinator test files currently contain no description cases.

**Closed in source:** the prior unrelated-edit description-erasure mechanism is fixed by seed/control/emission wiring. The prior linked kind:0 sync omission is fixed by carrying the retention Result through the blocking phase, attempting profile sync outside the store lock, and only then propagating the retention error. The new native regression exercises the production publication command with valid keys and a request counter; the team-refresh test now checks retained state under an interleaving. These are inspected test oracles, not executed mutation results.

Validation: exact-tree/blob-verified source and test-text review with independent frontend/native lanes; no checkout, builds, tests, or PR-code execution. Previously closed runtime, ownership, CAS, and refetch findings were not reopened; unrelated incoming-main features and create-flow unification are excluded.

wesbillman (changes requested):

## Changes requested

Reviewed head `a950da2a3ca2d8c669c1dc830b21e806f759f5c0` against base `ac5a18697c8294e9237f505a2e01ec6fc374849a`. The single edit surface is the right direction, but six user-visible regressions remain at the rendered-form/save boundaries.

### 1. [P2] Bind the unlinked prompt input to its editable draft

[`AgentEditMergedDialogInstanceSection.tsx:523–535`](https://github.com/block/buzz/blob/a950da2a3ca2d8c669c1dc830b21e806f759f5c0/desktop/src/features/agents/ui/AgentEditMergedDialogInstanceSection.tsx#L523-L535)

Open an unlinked instance, expand Advanced, and type in System prompt override. The controlled textarea receives `inst.systemPrompt`, but its change handler writes the dialog’s `systemPrompt` state. Every edit is therefore rendered back to the stored value, while `buildNextAgentFormModel` submits the changed state the user cannot see. Bind the control to that same draft state. Add a rendered multi-keystroke edit/Save test that checks both the visible text and submitted prompt.

### 2. [P2] Do not discard owner-review requests while personas are loading

[`AgentManagementDialogs.tsx:16–24`](https://github.com/block/buzz/blob/a950da2a3ca2d8c669c1dc830b21e806f759f5c0/desktop/src/features/agents/ui/AgentManagementDialogs.tsx#L16-L24)

Delay `list_personas` while managed agents and channels have initialized, then deliver a valid owned-agent update request. `useAgentManagement` accepts it, computes matches from `personasQuery.data ?? []` (`119–129`), and returns a not-found `editError` (`228–238`). This new effect immediately dismisses the request and retains its ID in `seenRequestIds`, so completion of the persona query cannot recover the draft. Distinguish loading/query failure from an authoritative missing target; do not dismiss before successful resolution. Test the actual management host with a deferred persona query.

### 3. [P2] Keep owner-review identity stable across a rename save

[`AgentManagementDialogs.tsx:52–57`](https://github.com/block/buzz/blob/a950da2a3ca2d8c669c1dc830b21e806f759f5c0/desktop/src/features/agents/ui/AgentManagementDialogs.tsx#L52-L57)

In an owner-reviewed update, rename the target. `matchingPersonas` continually resolves the request’s original `agentName` against the current display name (`useAgentManagement.ts:119–129`). The awaited mutation invalidation and coordinator readbacks refresh that name after persistence. The match then disappears, this render guard unmounts the editor, and the dismissal effect reports “can only update … by its current name.” This can happen before save/publication settlement finishes, removing the draft/error host even when the coordinator returns a recoverable failure. Resolve the name once, retain the persona ID for the request lifetime, and test rename with delayed or failed post-write settlement. This is a source-traced interleaving, not a live reproduction.

### 4. [P2] Validate instance credentials against the instance overlay

[`useAgentEditRuntimeState.ts:332–340`](https://github.com/block/buzz/blob/a950da2a3ca2d8c669c1dc830b21e806f759f5c0/desktop/src/features/agents/ui/useAgentEditRuntimeState.ts#L332-L340)

Open a linked instance of a team-managed, provider-backed definition whose required API key is supplied only by the instance overlay, with no global/build/file credential. `showDef` selects the definition env map here; the instance required-key gate consumes that map at `363–372`. The instance API-key control instead writes `instanceEnvVars` (`AgentEditMergedDialog.tsx:889–891`). Thus even a populated instance credential remains invisible to validation and Save stays disabled (`575–582`); the read-only definition cannot be edited to work around it. Derive the instance effective credential map from its overlay with the intended inheritance precedence, independently of D-section validation. Cover both a valid local key and an explicit-empty override.

### 5. [P2] Do not settle an unverified effort write as successful

[`agentSaveCoordinator.ts:349–358`](https://github.com/block/buzz/blob/a950da2a3ca2d8c669c1dc830b21e806f759f5c0/desktop/src/features/agents/ui/agentSaveCoordinator.ts#L349-L358)

Changing only effort produces `{pubkey, effortLevel}` (`useAgentEditMergedSubmit.ts:287–290`). If `update_managed_agent` fails at `save_managed_agents` (`agent_models_update.rs:314–323`) but list readback succeeds, `observedStateMatchesAgentInput` returns true: it checks no effort field. Both this boundary and final settlement discard the failure and close as “saved.” `ManagedAgent` does not expose effort, and `refetchAgentStores` never reads the separate config surface. Verify canonical effort before claiming success, or retain an explicitly unverified outcome. Add a production-submit regression with a rejected effort-only write and unchanged successful readback.

### 6. [P2] Preserve pin-to-inherit effort suppression

[`useAgentEditMergedSubmit.ts:284–290`](https://github.com/block/buzz/blob/a950da2a3ca2d8c669c1dc830b21e806f759f5c0/desktop/src/features/agents/ui/useAgentEditMergedSubmit.ts#L284-L290)

For a local pinned agent with a discovered effort picker, choose a different non-null effort, select “Inherit runtime from template,” then Save. The form emits `agentCommand: ""` to clear the pin, but this block unconditionally appends the touched effort. The existing backend clears effort for inheritance and then applies the explicit effort field (`agent_models_update.rs:93–106,197–215`), restoring the override that the transition should remove. The deleted base dialog explicitly prevented this with `resolveEffortSubmission({ inheritTransition: agentCommandUpdate === "" })`. Restore that suppression in merged submit and test the combined picker→inherit→Save transition; a backend rewrite is unnecessary.

## Scope and validation

Source-only review of immutable blobs; no PR checkout, build, test execution, or live reproduction. Carl integrated complementary UI/runtime, form/settlement, and Rust persistence/publication reviews. Coverage included library/profile/deep-link/owner-review routing, linked/unlinked/definition-only/team-managed ownership, exact-key identity, save/cancel/reopen, runtime inheritance, partial writes, CAS, and catalog publication/retry. Create-flow unification and vocabulary changes remain excluded; no independent mobile/web runtime validation was performed.

The CAS recovery candidate was rejected after tracing the awaited mutation `onSettled` cache invalidations; a model-clear setter mismatch was also excluded because no concrete user-impacting trigger was established. The existing effort test checks mount/hide rather than either save transition, and the R6 test mounts the merged dialog directly rather than the management host responsible for findings 2–3. Exit criteria are the six bounded fixes above with production-seam regressions, not another architectural expansion.

wesbillman (changes requested):

## Changes requested: five prior contracts remain unresolved

Reviewed head `d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7` against base `ac5a18697c8294e9237f505a2e01ec6fc374849a`, bounded to the six findings from `a950da2a3ca2d8c669c1dc830b21e806f759f5c0` and the new corrective delta. The single merged edit surface remains accepted product direction. Prior #6 (pin-to-inherit effort suppression) is fixed in source; #1–4 are unchanged, and #5 is incomplete at the IPC conversion boundary.

### 1. [P2] Bind the unlinked prompt input to its editable draft

[`AgentEditMergedDialogInstanceSection.tsx:523–535`](https://github.com/block/buzz/blob/d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7/desktop/src/features/agents/ui/AgentEditMergedDialogInstanceSection.tsx#L523-L535)

Open an unlinked instance, expand Advanced, and type in System prompt override. The controlled textarea receives `inst.systemPrompt`, but its change handler writes the dialog’s `systemPrompt` state. Every edit is therefore rendered back to the stored value, while `buildNextAgentFormModel` submits the changed state the user cannot see. Bind the control to that same draft state. Add a rendered multi-keystroke edit/Save test that checks both the visible text and submitted prompt.

### 2. [P2] Do not discard owner-review requests while personas are loading

[`AgentManagementDialogs.tsx:16–24`](https://github.com/block/buzz/blob/d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7/desktop/src/features/agents/ui/AgentManagementDialogs.tsx#L16-L24)

Delay `list_personas` while managed agents and channels have initialized, then deliver a valid owned-agent update request. `useAgentManagement` accepts it, computes matches from `personasQuery.data ?? []` (`119–129`), and returns a not-found `editError` (`228–238`). This new effect immediately dismisses the request and retains its ID in `seenRequestIds`, so completion of the persona query cannot recover the draft. Distinguish loading/query failure from an authoritative missing target; do not dismiss before successful resolution. Test the actual management host with a deferred persona query.

### 3. [P2] Keep owner-review identity stable across a rename save

[`AgentManagementDialogs.tsx:52–57`](https://github.com/block/buzz/blob/d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7/desktop/src/features/agents/ui/AgentManagementDialogs.tsx#L52-L57)

In an owner-reviewed update, rename the target. `matchingPersonas` continually resolves the request’s original `agentName` against the current display name (`useAgentManagement.ts:119–129`). The awaited mutation invalidation and coordinator readbacks refresh that name after persistence. The match then disappears, this render guard unmounts the editor, and the dismissal effect reports “can only update … by its current name.” This can happen before save/publication settlement finishes, removing the draft/error host even when the coordinator returns a recoverable failure. Resolve the name once, retain the persona ID for the request lifetime, and test rename with delayed or failed post-write settlement. This is a source-traced interleaving, not a live reproduction.

### 4. [P2] Validate instance credentials against the instance overlay

[`useAgentEditRuntimeState.ts:332–340`](https://github.com/block/buzz/blob/d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7/desktop/src/features/agents/ui/useAgentEditRuntimeState.ts#L332-L340)

Open a linked instance of a team-managed buzz-agent/Anthropic definition whose required API key is supplied only by the instance overlay, with no global/build/file credential. `showDef` selects the definition env map here; the instance required-key gate consumes that map at `363–372`. The instance API-key control instead writes `instanceEnvVars` (`AgentEditMergedDialog.tsx:893–895`). Thus even a populated instance credential remains invisible to validation and Save stays disabled (`576–584`); the read-only definition cannot be edited to work around it. Derive the instance effective credential map from its overlay with the intended inheritance precedence, independently of D-section validation. Cover both a valid local key and an explicit-empty override.

### 5. [P2] Carry canonical effort through the IPC mapper before comparing it

[`agentSaveCoordinator.ts:725–728`](https://github.com/block/buzz/blob/d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7/desktop/src/features/agents/ui/agentSaveCoordinator.ts#L725-L728), with the missing conversion at [`tauri.ts:628–671`](https://github.com/block/buzz/blob/d63ed1e3b96c583da3e5a9027f049b6e7cfc9af7/desktop/src/shared/api/tauri.ts#L628-L671).

The Rust summary now supplies `effort_level`, but `RawManagedAgent` has no corresponding member and `fromRawManagedAgent` drops it. Both `listManagedAgents` (`tauri.ts:778–781`) and `updateManagedAgent` (`1021–1030`) use that mapper, so production readback still has `effortLevel === undefined` regardless of the persisted column. Selecting a non-null effort and successfully saving therefore compares the submitted string to null and falsely reports failure. Conversely, clearing an existing effort with a rejected write compares null to null and can close as saved while the override remains on disk. This is the still-open observed-settlement finding, with a new false-failure consequence introduced by the comparator.

Add the raw field and map it into `ManagedAgent.effortLevel`; test successful set and rejected clear through raw IPC responses and the real conversion/list/refetch path. The added coordinator tests construct normalized agents directly, while the rendered successful-Save test asserts the outgoing payload/count but not successful settlement, so those assertions do not catch this dropped field.

## Closed finding and validation scope

Prior #6 is closed in source: `useAgentEditMergedSubmit.ts:299–308` calls `resolveEffortSubmission` with `rawAgentInput?.agentCommand === ""`, suppressing the explicit effort write when inheritance clears it. The runtime handler also restores the base pending-effort reset. This does not require a backend rewrite or clearing the persisted canonical effort on every runtime switch.

Source-only review of exact immutable blobs and test text. No PR checkout, build, test execution, mutation experiment, or live reproduction was performed. Re-review covered the actual unlinked prompt, linked/team-managed credential gate, owner-request loading/rename host, and effort picker → submit → native persistence → summary → IPC conversion → list/refetch settlement. Previously closed routing/ownership/CAS/catalog contracts were not reopened without new evidence. Create-flow unification, vocabulary changes, unrelated incoming-main features, and independent mobile/web runtime validation remain excluded.

Non-blocking test correction: `agentEditMergedRuntimeReset.test.mjs:136,182–208` starts effort at null and changes only the touched flag before switching runtimes. Its final null assertion cannot establish that `setEffortLevel(null)` reset a non-null draft. Seed a real non-null selection; do not claim a mutation result from that assertion alone.

Exit criteria are the five bounded fixes above with regressions at their production boundaries, not another architectural expansion.

PR #5330, opened 2026-08-08
fix(desktop): welcome banner overlap and missing dismiss control

wpfleger96:
## Problem

The `WelcomeComposerGuidanceLayer` in the `#Welcome` channel was positioned with `absolute inset-x-0 bottom-full z-[-1]` — outside the `composerWrapperRef` measurement boundary. `useComposerHeightPadding` observes `composerWrapperRef`'s block size to set `paddingBottom` on the timeline scroll container, but the absolutely-positioned layer didn't contribute to that size. The banner sat directly on top of the newest message, blocking the thread affordance on that message, and had no manual dismiss control.

## Fix

**Overlap**: Changed `WelcomeComposerGuidanceLayer` from `absolute inset-x-0 bottom-full z-[-1]` to `relative` (in normal flow). As a normal-flow child of `composer-dock`, the layer's full height is now measured by the ResizeObserver and fed into the timeline's `paddingBottom`, so the newest message is always fully visible and its thread affordance is always clickable while the banner shows.

**Dismiss**: Added an `X` close button (`data-testid="welcome-composer-dismiss-button"`) on the prompt state. Clicking fires `onDismiss`, which drives `dismissing → hidden` immediately (same slide-down animation as the auto-dismiss path) and marks the channel ID as completed in the session ref so the banner does not reappear on channel re-entry within the session.

**Refactor**: Extracted the banner state machine (refs, timers, `useEffect`s, and callbacks) from `ChannelPane.tsx` into `useWelcomeComposerBanner.ts`. This keeps `ChannelPane.tsx` well under the 1000-line file-size ratchet and makes the state machine independently testable.

## Changed files

- `desktop/src/features/channels/ui/WelcomeComposerBanner.tsx` — `WelcomeComposerGuidanceLayer` positioning fix; `onDismiss` prop; dismiss button; `overflow-hidden` / `mb-0` / `flex-1` cleanup
- `desktop/src/features/channels/ui/ChannelPane.tsx` — remove inline banner state machine, use `useWelcomeComposerBanner` hook, pass `onDismiss`
- `desktop/src/features/channels/ui/useWelcomeComposerBanner.ts` — new hook owning all banner state

### DIFF
--- a/desktop/src/features/channels/ui/WelcomeComposerBanner.tsx
+++ b/desktop/src/features/channels/ui/WelcomeComposerBanner.tsx
@@ -416,7 +416,7 @@ export function WelcomeComposerBanner({
               </motion.span>
             )}
           </AnimatePresence>
-          {state === "prompt" && onDismiss ? (
+          {state === "prompt" && onDismiss && !settingUp ? (
             <button
               aria-label="Dismiss hint"
               className="ml-auto flex h-5 w-5 shrink-0 items-center justify-center rounded-full text-muted-foreground/60 transition-colors hover:text-muted-foreground focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring"
--- a/desktop/src/features/channels/ui/WelcomeComposerBanner.tsx
+++ b/desktop/src/features/channels/ui/WelcomeComposerBanner.tsx
@@ -1,6 +1,6 @@
 import * as React from "react";
 import { AnimatePresence, motion, useReducedMotion } from "motion/react";
-import { Bot, Check } from "lucide-react";
+import { Bot, Check, X } from "lucide-react";
 
 import { ComposerDockGlassBackdrop } from "@/features/messages/ui/ComposerDockBackdrop";
 import { cn } from "@/shared/lib/cn";
@@ -291,9 +291,15 @@ type WelcomeComposerBannerProps = {
    * banner during this window.
    */
   settingUp?: boolean;
+  /**
+   * Called when the user dismisses the banner manually via the close button.
+   * Only rendered while `state === "prompt"`.
+   */
+  onDismiss?: () => void;
 };
 
 export function WelcomeComposerBanner({
+  onDismiss,
   settingUp = false,
   state,
 }: WelcomeComposerBannerProps) {
@@ -307,7 +313,7 @@ export function WelcomeComposerBanner({
         animate={{
           height: state === "dismissing" ? 0 : "auto",
         }}
-        className="overflow-visible"
+        className="overflow-hidden"
         initial={false}
         transition={{
           duration:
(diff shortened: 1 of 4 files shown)


desktop/src/features/channels/ui/WelcomeComposerBanner.tsx
```diff
@@ -410,6 +416,17 @@ export function WelcomeComposerBanner({
               </motion.span>
             )}
           </AnimatePresence>
+          {state === "prompt" && onDismiss ? (
```
wesbillman: **[P1] Do not offer a banner-only dismiss while kickoff characters are active.** `state` remains `"prompt"` during `settingUp`, so this button is rendered immediately. Clicking it only transitions `useWelcomeComposerBanner` to hidden; it does not change `useWelcomeKickoffStagePresence`, whose `WelcomeKickoffStage` child remains active until a message arrives or the 90-second timeout. Once the banner collapses, those absolutely positioned characters (`bottom-full` in `WelcomeKickoffStage.tsx`) are left floating above the composer with no banner beneath them—the opposite of dismissing the guidance. Either hide/dismiss the stage together with the banner, or do not render the X while `settingUp`. Please add an E2E assertion that clicking the new control removes the complete guidance surface and that it stays gone on channel re-entry; the current test change only updates geometry and never exercises the new control.

wesbillman (changes requested):
Reviewed on Wes's behalf. One user-visible state mismatch needs fixing before merge: manual dismissal can hide the banner while leaving the kickoff characters floating above the composer. Please couple those states (or suppress dismissal during setup) and cover the manual-dismiss path in E2E.

desktop/src/features/channels/ui/WelcomeComposerBanner.tsx
wpfleger96: 🤖 Took option 1 (suppress X while `settingUp`). The setup window is bounded (≤90s timeout) and `useWelcomeKickoffStagePresence` has its own lifecycle — coupling teardown adds machinery for a transient state that resolves itself. If `settingUp` is true, the X is hidden; once setup completes and `state` returns to steady-state `"prompt"`, the X reappears normally.

Also added two E2E tests covering the dismiss path: one that clicks X and asserts the full guidance surface is gone, and one that verifies dismiss persists after leaving and re-entering the channel. Tip: `b9fe344b7`.

wpfleger96:
## Screenshots

### Before — newest message (bob, 12:21 PM) occluded by the banner; no dismiss

The banner renders `absolute bottom-full` outside the measured composer area. `useComposerHeightPadding` only measures `composerWrapperRef` (the composer overlay div), which doesn't include the absolutely-positioned banner. The banner therefore sits on top of the newest message — its text is hidden and its thread affordance is unreachable.

before

### After — all messages visible; × dismiss button shown

The guidance layer now renders in normal document flow inside the composer overlay, so its height is included in the `composerWrapperRef` measurement. The timeline receives correct bottom clearance. The × button (top-right of the banner) dismisses via the existing `dismissing → hidden` path and marks the channel complete in `completedChannelIdsRef` (no reappearance on re-entry).

after

wesbillman (commented):
Re-reviewed at b9fe344b7be464b2b07cba3c433e986009c3022a on Wes's behalf. The prior blocker is resolved: the dismiss control is now suppressed while `settingUp`, so the kickoff stage cannot be orphaned, and the added E2E cases cover manual dismissal plus channel re-entry persistence. I found no new actionable issues in the follow-up delta.

I am leaving this as a comment rather than an approval because Wes did not explicitly delegate approval authority for this PR.

wesbillman (approved):

PR #5339, opened 2026-08-08
fix: show workflows from all member channels in Desktop overview

joaoh82:
## Summary

The Desktop **Workflows** tab only showed workflows from one channel (often the DM-scoped card) and omitted others such as Engineering's `Marvin Ticket Dispatcher`, even though single-channel queries returned them.

**Root cause:** Desktop batched all member-channel UUIDs into one multi-value `#h` filter. On the HTTP `/query` path, `extract_channel_id_from_filter` collapsed multi-`#h` to the first iterated UUID and SQL-scoped to that single channel, dropping every other channel's kind:30620 events.

## Changes

1. **Relay** (`extract_channel_id_from_filter`): only push SQL `channel_id` for **exactly one** `#h` value. Multi-value `#h` is NIP-01 OR and leaves channel scope open for accessible-channel filtering + post-match (aligned with the WS REQ path).
2. **Desktop/Tauri** (`get_channels_workflows`): issue **N single-`#h` filters** in one round-trip instead of one multi-`#h` filter, so the overview works even against unfixed production relays.

## Test plan

- [x] `cargo test -p buzz-relay extract_channel_id_from_filter --lib` (3 new + existing extract tests)
- [x] `cargo clippy -p buzz-relay --lib --tests -- -D warnings`
- [ ] Desktop unit tests for `channels_workflows_filters` (added; full `buzz-desktop` lib build needs local cmake/opus toolchain)
- [ ] After merge/deploy: Workflows tab lists Engineering + DM workflows; Refresh still shows both
- [ ] Single-channel CLI `buzz workflows list --channel <eng>` unchanged

## Notes

Separate from #5048 (ACP workflow agent wake). This only restores discovery in the Workflows overview.

### DIFF
--- a/crates/buzz-relay/src/handlers/req.rs
+++ b/crates/buzz-relay/src/handlers/req.rs
@@ -855,18 +855,26 @@ fn filters_are_nip43_membership_only(filters: &[Filter]) -> bool {
 }
 
 /// Extract a channel UUID from a single filter's `#h` tag.
+///
+/// Returns `Some` only when the filter has **exactly one** `#h` value that
+/// parses as a UUID. Multi-value `#h` is NIP-01 OR ("match any of these
+/// channels") and must **not** be collapsed into a single SQL `channel_id`
+/// predicate — that would silently drop events in every other listed channel
+/// (Desktop Workflows overview used one multi-`#h` filter for all member
+/// channels and only received the first channel's workflows).
+///
+/// When this returns `None` for multi-`#h`, callers leave `channel_id` unset,
+/// apply the accessible-channel scope, and rely on NIP-01 post-filter matching
+/// (same shape as the WS REQ path's `per_filter_channel` for `vs.len() != 1`).
 fn extract_channel_id_from_filter(filter: &Filter) -> Option<uuid::Uuid> {
-    for (tag_key, tag_values) in filter.generic_tags.iter() {
-        let key = tag_key.to_string();
-        if key == "h" {
-            for val in tag_values {
-                if let Ok(id) = val.parse::<uuid::Uuid>() {
-                    return Some(id);
-                }
-            }
+    let h = nostr::SingleLetterTag::lowercase(nostr::Alphabet::H);
+    filter.generic_tags.get(&h).and_then(|vs| {
+        if vs.len() == 1 {
+            vs.iter().next()?.parse::<uuid::Uuid>().ok()
+        } else {
+            None
         }
-    }
-    None
+    })
 }
 
 /// Convert a single NIP-01 filter into an [`EventQuery`] for the database.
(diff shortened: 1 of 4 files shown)


avi-zloof:
We shipped a related fix on our fork: [avi-zloof/buzz#22](https://github.com/avi-zloof/buzz/pull/22).

Same relay idea as this PR (`extract_channel_id_from_filter` must not collapse multi-value `#h` to the first UUID). Desktop side we also changed `get_channels_workflows` to send **one single-`#h` filter per member channel** in one `/query` round-trip, because a multi-value `#h` still 400’d against our live relay.

If this PR lands the relay half, we can gift just the desktop per-channel filters as a tiny follow-up — say if you want that split.

yalladoless:
**Independent confirmation against Block's production relay** (`yds.communities.buzz.xyz`, tested 2026-08-15, a Desktop 0.5.11 workspace) — plus production validation of this PR's desktop-side query shape, which I haven't seen tested in the other threads.

Method: NIP-98-signed `POST /query` using a workspace identity (NIP-OA auth tag), issuing controlled `kinds:[30620]` filters over the workspace's six member channels. Channel UUIDs are redacted to letters **A–F in lexicographic sort order** (sort order is the variable that matters here). At measurement time channel **F** (sorts last) held 2 workflow definitions, channel **E** held 1, channels **A–D** held 0.

| Filter(s) sent | Events returned | Expected (NIP-01 OR) |
|---|---|---|
| `#h:[F]` | 2 | 2 ✅ |
| `#h:[E]` | 1 | 1 ✅ |
| `#h:[F,E]` | **1** (E's event only) | 3 ❌ |
| `#h:[E,F]` | **1** (E's) | 3 ❌ — request order irrelevant (BTreeSet) |
| `#h:[A,F]` | **0** | 2 ❌ — A sorts first, holds none |
| `#h:[F,A]` | **0** | 2 ❌ |
| `#h:[A,B,C,D,E,F]` | **0** | 3 ❌ |
| same, plus `limit:5000` / `since:0` | 0 | — rules out limit/pagination effects |
| **two filters** `[{#h:[F]}, {#h:[E]}]` | **3** | 3 ✅ |

Every failing case returned exactly the events of the **lexicographically smallest** listed UUID — five distinct combinations, all consistent with `extract_channel_id_from_filter` iterating the `BTreeSet` and pinning SQL to the first parseable value, as diagnosed in #2385.

The last row is the new datapoint: **N single-`#h` filters in one request already return the correct union on today's unfixed production relay.** That means the `channels_workflows_filters` shape this PR ships client-side is validated against production as-is — the desktop half alone un-breaks the Workflows overview for hosted users on their next app release, independent of relay deploy timing, and keeps working after the relay half lands.

One real-world impact datapoint for severity: before diagnosing this, the persistently empty overview led to the same workflow being re-created three times in one workspace — each recreate minting a new webhook secret and silently orphaning the previously copied webhook URL, since secrets are only surfaced once at creation.

PR #3846, opened 2026-07-31
fix(desktop): explain missing pairing endpoint

kiranmagic7:
## Summary

- Preserve whether the pairing URL came from the legacy `/pair` fallback.
- Replace only that fallback's HTTP 404 with an actionable error naming the endpoint and the two supported remedies.
- Strip query parameters and fragments before showing the endpoint; keep every other WebSocket error unchanged.

### Related issue

Fixes #3779.

Deployment-side gaps are covered separately by #3627 for Compose and #3424 for chart-managed routing.

### Testing

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml pairing_relay_tests -- --nocapture` — 9 passed
- `pnpm --dir desktop build:e2e && pnpm --dir desktop exec playwright test --project=smoke --grep "legacy pairing 404 explains how to configure the relay"` — 1 passed
- `just ci` — passed

Before/after screenshots are included in a follow-up comment.

PR #5340, opened 2026-08-08
fix(desktop): explain the pairing 404 instead of surfacing a bare transport error

### LINKED ISSUE #3779: fix(desktop): pairing shows a bare 404 when the legacy /pair fallback hits a relay with no pairing endpoint
## Problem

When a NIP-43 relay advertises no `pairing_relay_url` in its NIP-11 document, the desktop resolves the pairing endpoint through the legacy fallback: `pairing_relay_from_nip11` (`desktop/src-tauri/src/commands/pairing.rs:469`) sees 43 in `supported_nips` and returns `PairingRelay::LegacyPath`, and `resolve_pairing_relay_url` (`pairing.rs:452`) appends `/pair` to the main relay URL.

If nothing serves that path (the state every compose deployment was in before #2736), the WebSocket upgrade 404s and the QR panel surfaces the raw transport error from `pairing.rs:240`:

```
WebSocket connection failed: HTTP error: 404 Not Found
```

That message names no URL, no route decision, and no config lever. Several self-hosters in #2734 independently had to source-dive `config.rs` and `pairing.rs` to discover that `BUZZ_PAIRING_RELAY_URL` (or routing `/pair` to a `buzz-pair-relay` sidecar) was the fix. @DrDoomi called out the diagnosability gap explicitly: https://github.com/block/buzz/issues/2734#issuecomment-5134095185

## Suggested fix

The route decision is already a typed enum (`PairingRelay`, `pairing.rs:411`), so the connect error mapping at `pairing.rs:238` can know which route failed. When the route is `LegacyPath` and the handshake fails with an HTTP status, emit something actionable instead of the bare transport error, along the lines of:

```
Pairing endpoint wss://<domain>/pair not found (404). The relay advertises
NIP-43 but no pairing_relay_url, and nothing is serving /pair. Set
BUZZ_PAIRING_RELAY_URL on the relay or route /pair to a buzz-pair-relay
instance (see deploy/compose/README.md, Device pairing).
```

`Configured` and `MainRelay` failures can keep the current message with the attempted URL appended, which also helps the split domain case when the configured URL is wrong.

## Context

Follow-up to #2734. #2736 fixes the compose bundle so fresh deployments get a working `/pair` out of the box, but any deployment without a routed `/pair` or a configured `pairing_relay_url` still fails with the bare 404, and the error should say why.

akeemjenkins:
Closes #3779.

**Duplicate check:** none found for this change. #3779 has no PR attached. The related compose-side work (#2736, #3627, #3875, #4656) and the chart routing (#3424) fix the *deployments*; this fixes the *message* for every deployment they cannot reach — external reverse proxies, the Railway template mentioned in #3779, Helm installs without `pairingRelay`, and anything hand-rolled.

## Problem

When a NIP-43 relay advertises no `pairing_relay_url`, `pairing_relay_from_nip11` returns `LegacyPath` and `resolve_pairing_relay_url` appends `/pair` to the relay's own host. That address is chosen from an *advertisement* and never probed, so when nothing serves it the QR panel showed only:

```
WebSocket connection failed: HTTP error: 404 Not Found
```

No URL, no route decision, no config lever. Per #3779 and its confirmations, self-hosters had to read `config.rs` and `pairing.rs` to get unstuck.

## Change

The route was already a typed enum but collapsed to a `String` before the task that reports failures could see it. `PairingRelay` is now `Clone` and travels into `PairingTaskContext`, so `describe_connect_failure` can explain the failure in terms of the decision that produced the URL.

A 404 on the legacy branch:

```
Pairing endpoint wss://relay.example.com/pair not found (404). This relay
advertises NIP-43 but no pairing_relay_url, so pairing fell back to
wss://relay.example.com/pair, and nothing is serving that path. The relay
operator needs to set BUZZ_PAIRING_RELAY_URL on the relay, or route /pair to a
buzz-pair-relay instance (see deploy/compose/README.md, "Device pairing").
```

Everything else keeps the original transport error and gains the attempted URL — the shape #3779 suggested, and what the split-domain case needs when a configured `pairing_relay_url` is simply wrong.

One deliberate narrowing: the actionable text is gated on **404 on the legacy branch specifically**. A routed `/pair` whose sidecar is down answers **502**, which is a different problem with a different fix, so it must not be told nothing is serving the path. There is a test for that.

## Tests

Four added to `pairing_relay_tests.rs`, beside the existing route tests:

- `legacy_path_404_explains_the_missing_pairing_endpoint` — asserts the URL, the NIP-43 reason, and *both* levers appear
- `legacy_path_non_404_keeps_the_transport_error_and_gains_the_url` — the 502 case above
- `configured_pairing_relay_404_reports_the_url_it_was_given` — split domain; no `/pair` advice, which would point at the wrong file
- `main_relay_failure_keeps_the_transport_error`

`cargo test -p buzz-desktop`: **2274 passed, 0 failed**. `cargo fmt --check` and `cargo clippy --all-targets` clean.

## Why this one matters beyond the compose fix

We run a small managed Buzz fleet and hit this exact 404 on our own tenants (details in #2736). We had the source open, so we found it in an afternoon. The message is what everyone else has to work from, and it is the only part of this that helps a relay whose misconfiguration no upstream default can prevent.

The wording points at `deploy/compose/README.md`. If #2736 or a sibling lands and renames that section, this string should follow it — happy to rebase onto whichever lands first.

### DIFF
--- a/desktop/src-tauri/src/commands/pairing.rs
+++ b/desktop/src-tauri/src/commands/pairing.rs
@@ -42,6 +42,9 @@ enum PairingMode {
 #[derive(Clone)]
 struct PairingTaskContext {
     mode: PairingMode,
+    /// Which NIP-11 branch produced the pairing URL. Carried purely so a
+    /// failed handshake can say why it dialled where it did.
+    route: PairingRelay,
     generation: Arc<AtomicU64>,
     generation_fence: Arc<std::sync::Mutex<()>>,
     task_generation: u64,
@@ -129,7 +132,8 @@ async fn start_pairing_session(
 
     let ws_url = relay_ws_url_with_override(&state);
     let http_url = relay_api_base_url_with_override(&state);
-    let pairing_relay_url = resolve_pairing_relay_url(&ws_url, probe_pairing_relay(&ws_url).await)?;
+    let pairing_route = probe_pairing_relay(&ws_url).await;
+    let pairing_relay_url = resolve_pairing_relay_url(&ws_url, pairing_route.clone())?;
     let (session, qr_payload) = PairingSession::new_source(pairing_relay_url.clone());
     let mut qr_uri = encode_qr(&qr_payload);
     if mode == PairingMode::RecoverIdentity {
@@ -166,6 +170,7 @@ async fn start_pairing_session(
         Arc::clone(&pairing.session),
         PairingTaskContext {
             mode,
+            route: pairing_route,
             generation: Arc::clone(&pairing.generation),
             generation_fence: Arc::clone(&pairing.generation_fence),
             task_generation,
@@ -305,7 +310,7 @@ async fn pairing_ws_task_inner(
 ) -> Result<(), String> {
     let (ws, _) = connect_async(relay_url)
         .await
-        .map_err(|e| format!("WebSocket connection failed: {e}"))?;
+        .map_err(|e| describe_connect_failure(&context.route, relay_url, &e))?;
     let (mut write, mut read) = ws.split();
 
     handle_nip42_auth(&mut read, &mut write, session, relay_url).await?;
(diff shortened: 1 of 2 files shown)


kiranmagic7:
@akeemjenkins, this overlaps #3846, which has been open for #3779 since July 31 and already separates the legacy `/pair` case from configured `pairing_relay_url` failures. One material difference: `describe_connect_failure` interpolates `url` directly, so a valid configured URL such as `wss://pair.example/connect?token=secret#fragment` is copied into the UI error. #3846 strips the query and fragment before display and pins that behavior in both legacy and configured-route tests. Whichever patch lands should keep that redaction.

kiranmagic7:
Thanks for closing the overlap cleanly. A review of #3846’s legacy-vs-configured route split and its query/fragment redaction tests would help most; that is the main behavior difference the two patches exposed.

PR #5349, opened 2026-08-08
feat(core): define a versioned stdio MCP launch document

wolfyy970:
## Summary

Adds the shared contract that target-side MCP discovery needs to hand one resolved stdio configuration to an agent launch.

The document is versioned, bounded and strict. It rejects unknown fields, duplicate names and protected identity environment keys, keeps command arguments and environment details out of diagnostics, and requires consumers to isolate MCP children from the agent's ambient credentials.

This is the first piece extracted from #4164. It does not add HTTP, ACP consumption, Desktop configuration or another source of MCP authority. #5321 can map a selected native entry into this document once it lands.

## Testing

- `just ci`

### DIFF
--- a/crates/buzz-core/src/lib.rs
+++ b/crates/buzz-core/src/lib.rs
@@ -24,6 +24,8 @@ pub mod git_perms;
 pub mod invite;
 /// Buzz kind number registry — custom event type constants.
 pub mod kind;
+/// Versioned MCP configuration shared by launchers and the ACP harness.
+pub mod mcp_config;
 /// Network utilities — SSRF-safe IP classification.
 pub mod network;
 /// Agent observer frame helpers.
--- a/crates/buzz-core/src/mcp_config.rs
+++ b/crates/buzz-core/src/mcp_config.rs
@@ -1,0 +1,544 @@
+//! Versioned MCP configuration shared by launchers and the ACP harness.
+//!
+//! The launcher that writes an ephemeral configuration and the harness that
+//! consumes it must use the same wire contract. Keeping the schema here avoids
+//! a successful launch producing a document that the harness cannot parse.
+//!
+//! Version 1 supports stdio servers only. A consumer must not pass the agent
+//! process environment through to an MCP child. It starts from an empty
+//! environment, restores only the target's minimal non-secret process
+//! essentials, and then applies the server environment declared here. This
+//! prevents agent identity and authorization values from leaking into tools.
+
+use std::collections::{BTreeMap, HashSet};
+
+use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+/// Current structured MCP configuration version.
+pub const MCP_CONFIG_VERSION: u32 = 1;
+/// Maximum encoded size accepted for one structured MCP configuration.
+pub const MCP_CONFIG_MAX_BYTES: u64 = 64 * 1024;
+/// Maximum number of MCP servers in one launch document.
+pub const MCP_SERVER_MAX_COUNT: usize = 16;
+/// Maximum number of arguments for one stdio MCP server.
+pub const MCP_SERVER_MAX_ARGS: usize = 128;
+/// Maximum number of environment entries for one stdio MCP server.
+pub const MCP_SERVER_MAX_ENV: usize = 128;
+/// Maximum encoded length of an MCP server name.
+pub const MCP_SERVER_NAME_MAX_BYTES: usize = 128;
+
+/// Harness identity and authorization variables that an MCP server may not override.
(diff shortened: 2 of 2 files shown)


whyihaveyou:
From #5321 (target-side discovery / runtime-config read model) — thanks for splitting this out and flagging the seam. A selected native stdio entry maps cleanly; confirming intended strictness on a few boundary rules so our mapper behaves correctly.

**Mapping lines up:** our `McpServerConfig { name, command, args, env: BTreeMap }` maps 1:1 onto `ConfiguredMcpServer::Stdio` (env already deterministic-ordered). `enabled` has no wire counterpart — we drop it at mapper level (only mapping active native entries). Design philosophy (redacted debug, bounded size, deny_unknown_fields, protected-key hard-reject) matches the direction of our 627386e. No hard blocker.

Three boundary points worth pinning down (not requesting changes — just so our mapper can reject loudly vs. silently drop):

1. **Server-name charset `[A-Za-z0-9_-]`, ≤128B, no `__`.** Common npm-scoped MCP identifiers used as map keys (e.g. `@modelcontextprotocol/server-filesystem`) contain `@`/`/` and would be rejected by `valid_mcp_server_name`. Since the field is documented as a "stable ACP identifier", do you intend the whole document to be rejected (as `validate()` does today), with the producer expected to alias/sanitize server names before encoding? That's our current read.

2. **Env-key rule `[A-Za-z_][A-Za-z0-9_]*`** is stricter than POSIX (rejects `-`/`.` in names). Intentional and fine — just confirming the mapper should reject such entries rather than transform them.

3. **Bounds vs. native read cap.** #5321 reads native configs up to `RUNTIME_CONFIG_MAX_BYTES` (1 MiB, bounds the *source file*), while the launch document is capped at 64 KB encoded / 16 servers / 128 args / 128 env. A large native entry (many or large env values) can therefore decode fine on our side but fail `to_json` at launch-document size. We'll surface this as a mapping error rather than silently dropping the server — confirming that's the expected contract.

No follow-up implementation yet on our side — flagging so we encode the correct reject-loud semantics once #5349 lands.

whyihaveyou:
From **#5321** (target-side / runtime-config read model): here is the mapper you anticipated in the summary — *"#5321 can map a selected native entry into this document once it lands."* This is a live, running proof, not a hand-written example.

**How it was produced** (our example binary, commit `80dfdaf`, `crates/buzz-runtime-config/examples/mcp_launch_mapper.rs`):

```
cargo run -p buzz-runtime-config --example mcp_launch_mapper ~/.hermes/config.yaml
```

It read a real native Hermes config at `~/.hermes/config.yaml`, ran our unified `McpServerConfig` read-model (`hermes` adapter), and mapped every enabled server through to `McpLaunchConfigDocument` (v1). The output passed `doc.validate()` against your schema — name charset/uniqueness, NUL-free command and args, env-key case-insensitive uniqueness, and no protected identity keys (`PROTECTED_MCP_ENV_NAMES`). Env values are redacted below so it's safe to paste into a comment or log.

**Mapped document (stdio v1, env redacted):**

```json
{
  "servers": [
    {
      "args": ["-y", "sciverse-mcp-server"],
      "command": "npx",
      "env": { "SCIVERSE_API_TOKEN": "***" },
      "name": "sciverse",
      "transport": "stdio"
    },
    {
      "args": ["run", "--project", "/Users/qzp/sciverse-deep-research/mcp-server", "sciverse-survey-gates"],
      "command": "uv",
      "env": {},
      "name": "sciverse-survey-gates",
      "transport": "stdio"
    }
  ],
  "version": 1
}
```

Native → document mapping is 1:1 on the wire: `name` / `command` / `args` / `env` copy verbatim, transport tagged `stdio` (v1), disabled servers dropped, and any server that would violate your bounded schema (duplicate names, protected env keys, env keys differing only by ASCII case) is rejected loudly before emission rather than silently dropped. This closes the "map into this document" path from the #5321 side.

wolfyy970:
Thanks for proving the field mapping against a real Hermes config. The intended boundaries are:

- A native name outside the portable grammar needs an explicit producer-owned alias. It must not be auto-sanitized because distinct names can collapse.
- Invalid environment keys reject the mapping. They are never transformed.
- Any count or size limit rejects the whole launch handoff. An enabled server is never silently dropped.

One correction before we call the #5321 mapper schema-compatible: its local mirror does not yet enforce the 16-server, 128-argument, 128-environment, or 64 KiB limits. I am leaving the exact re-review there.

I rebased this one-commit contract onto current `main` at `13994181e`, and the full `just ci` gate passes.

@wesbillman @tlongwell-block, #5321 now has a native-config field mapper waiting on this shared boundary. Could one of you review #5349?

wolfyy970:
Closing this. No review.

PR #5358, opened 2026-08-08
fix(desktop): re-sign stale retention events at flush

### LINKED ISSUE #4967: Buzz Desktop managed-agent publisher replays stale signed kind:30177 events indefinitely
We run a self-hosted Buzz relay (`ghcr.io/block/buzz:main`) with one Buzz Desktop client (0.5.4 per the client's binary version resource and uninstall registry; Windows). The desktop's managed-agent re-assert loop republishes the signed kind:30177 event persisted at instance creation instead of re-signing, so once that artifact ages past the relay's ±900s ingest window, every re-assert is rejected — permanently, every 30 seconds, with no recovery path. Measured on our relay: 15,462 rejected writes across a 47-hour log window (a floor — the storm predates the window), about a third of the client's bridge traffic, and no agent-definition update can reach the relay again.

Full diagnosis below: mechanism with timestamps, a reproduction that needs only an agent and fifteen minutes of waiting, and a suggested fix — re-sign at publish, and invalidate the retained artifact on rejection. Cross-verified independently from the relay side and the client host.

---

# Buzz Desktop: managed-agent publisher replays a stored signed event instead of re-signing

**Status:** diagnosis complete, cross-verified from both sides (relay + client host).
**Written:** 2026-08-05. Investigators: Calliope (relay side, Delphi), Athena (client side, Aurora).
**Intended use:** upstream bug report to the Buzz maintainers.

---

## Summary

Buzz Desktop's managed-agent publisher re-asserts each agent definition by
**republishing the signed kind:30177 event it stored on first publish**, rather
than re-signing at publish time. Because a Nostr event id commits to
`created_at`, the artifact's timestamp never moves. Once the stored event ages
past the relay's ingest window (±900s), every subsequent re-assert is rejected —
permanently, on a 30-second cycle, with no path to recovery short of deleting
the artifact.

A rejection does not trigger a re-sign, so the loop is closed: the definition
can never be updated on the relay again, and the client never learns.

## Impact, measured

Measured on one relay (Delphi, `ghcr.io/block/buzz:main`) against one desktop:

| Metric | Value |
|---|---|
| Rejected writes in retained log window (2026-08-03T21:32Z → 2026-08-05T20:56Z, 47.4h) | **15,462** as of 2026-08-05T20:56Z — still climbing at the rate below |
| Rate while the client is running | **360/hour** — exactly 3 events per 30s cycle |
| Average across the window | 326/hour — the shortfall is client downtime, not a varying rate (see below) |
| Share of all HTTP bridge traffic in a sample hour | 360 of 1,125 requests (**32%**) |
| Stored kind:30177 rows, all time | 6 — three from 2026-07-30, three from 2026-08-02. **None since.** |
| Storm began | ~2026-08-03T00:15Z — 15 minutes after signing, when the artifact aged past the window |

Every rejection is a full request: TLS, auth, signature verification, and a
`WARN` log line, discarded at the timestamp gate
(`MAX_TIMESTAMP_DRIFT_SECS = 900`, `crates/buzz-relay/src/handlers/ingest.rs`).
The relay has accepted no managed-agent definition from this client since
2026-08-02T23:59Z.

The retained log window opens after the storm began, so 15,462 is a floor, not
a total.

**The cadence never adapts.** Scanning the whole retained log for interruptions
in the rejection stream — threshold set at 45s, i.e. any single missed 30s cycle
— finds ten, of which two are substantial and both fall on 2026-08-04:

```
07:29:02Z → 11:43:01Z   4.23 h    client host rebooted (Windows Update); the
                                  desktop does not auto-start
11:43:01Z → 11:48:55Z   5.9 min   app restart cycle immediately after
```

The other eight are isolated one-or-two-cycle skips of 47–74s, scattered across
the three days, each followed immediately by a return to the exact 30s cadence.
Three of them cluster in the 12:07–12:27Z hour on 08-04, the tail of that
morning's restart sequence.

**The distinction that matters for diagnosis: none of these is backoff.** A
client reacting to rejection would show growing intervals; these are flat ~60s
one-offs against an otherwise metronomic 30s, with no trend in either
direction. The 4.23h reboot accounts for the difference between the 360/hour
running rate and the 326/hour window average. Nothing in the client ever slows
down, backs off, or gives up: ~15,000 identical rejections have produced no
change in its behaviour at all.

Client build version is not recorded here because the relay does not log a
client version or user-agent on this path — anyone citing one should source it
from the desktop host.

## Reproduction

1. Create a managed agent in Buzz Desktop; observe the kind:30177 event publish
   and be accepted.
2. Wait longer than the relay's ingest window (15 minutes).
3. Observe every subsequent 30-second re-assert rejected with
   `invalid: event timestamp too far from server time`, indefinitely.

No further action is required to reproduce — step 2 is just waiting.

## Mechanism, established

1. **Definitions are signed once, at instance creation.** On 2026-08-02 the
   three agents' store entries were written at `23:59:48.663`, `.775`, `.880`,
   and the resulting events carry `created_at` of `1785715188`, `1785715189`,
   `1785715189` — that is, `:48`, `:49`, `:49`.

   All three entry stamps fall inside second `:48`, so the event stamp is not
   derived from the entry field: truncation would give `:48,:48,:48` and
   rounding `:49,:49,:49`. Neither reproduces the observed split. What does fit
   is signing with `now()` shortly after each entry write, with the run crossing
   the second boundary between the first and second agent — consistent with any
   write-to-sign delay between roughly 0.23s and 0.34s. **The delay is inferred
   from that constraint, not measured**; the load-bearing conclusion is only
   that the stamp comes from signing time, not from a stored field.

2. **The signed event is persisted verbatim.**
   `%APPDATA%\xyz.block.buzz.app\agents\retention\<uuid>.db` — SQLite, mtime
   frozen at `2026-08-03T00:00:00Z` (~11s after signing, i.e. after the relay
   accepted), untouched since. It holds complete signed events — `id`, `sig`,
   `created_at`, `tags`, `content` — plus the serialized projection string used
   as the change comparator.

3. **The periodic re-assert replays that artifact unconditionally.** Publish
   success *was* recorded on 08-02, and the projection has not changed, yet the
   client still re-asserts every 30 seconds — so this is not merely
   "suppression keyed to a success that never arrives." It replays regardless,
   and a rejection never marks the artifact stale.

4. **Nothing else can move the timestamp.** App restarts do not (three on
   08-04). Local store edits do not — they touch `updated_at`, not the stored
   event. A UI edit would not either: it changes content, but the publisher
   still replays the retained artifact.

## Cross-verification

The two sides were checked independently and agree exactly:

- Relay-side: the three stored rows' NIP-01 ids recompute from their own fields
  (`sha256([0,pubkey,created_at,kind,tags,content])`) — all three match.
- Client-side: today's live store values, run through the documented
  `agent_event_content()` projection, reproduce the stored content strings
  byte-for-byte and therefore the same three ids.
- The retention DB's stored event for Fizz matches the relay's stored row field
  for field, including the omitted `respond_to_allowlist` key on an empty vec.

So the client is emitting a byte-identical event with a byte-identical id every
cycle, and has been for three days.

## Note on the payload projection

`ManagedAgentEventContent` declares nine fields, but
`agent_event_content()` (`desktop/src-tauri/src/managed_agents/agent_events.rs`)
nulls `system_prompt`, `model`, `provider` and `persona_source_version` whenever
`persona_id.is_some()` — the documented "slimmed projection" — and all four are
`skip_serializing_if = "Option::is_none"`. Definition-linked instances
therefore emit exactly five fields. This is recorded because a struct-shape
reading of the binary suggests nine and leads to the wrong conclusion.

## Suggested fix

**Re-sign at publish time.** The re-assert path should build and sign a fresh
event with `now()` rather than replaying the retained artifact; the retention
row should serve as the change comparator (its evident purpose) and not as the
thing that gets sent.

Secondary, defence in depth: a rejected publish should invalidate the retained
artifact so the next cycle re-signs, rather than replaying it indefinitely.

## Local workaround (untested, destructive-ish)

Full app quit including tray → rename the retention DB → relaunch, expecting the
publisher to find no artifact and sign fresh. Reversible by renaming back.
Rebuild behaviour is unproven. Not run: the fix belongs upstream, and this
touches app data.

**Do not** attempt to fix it by editing `managed-agents.json` `created_at` — the
publisher does not read that field for the event stamp.

## Relay-side mitigation (optional, does not fix the client)

At the ingest timestamp gate (`crates/buzz-relay/src/handlers/ingest.rs`), a
verbatim re-send of an event the relay already stores could be answered
`duplicate:` instead of `invalid:`. Verified applicable: the ids are stable, so
the lookup would match.

Two caveats: put the lookup **inside the rejection branch**, not before the
gate, so normal ingest pays no extra database round-trip; and note that this
silences the only signal that surfaces the client defect. It should not be
treated as closing this report.

olympusbuildz:
## Problem

Managed-agent retention can leave parameterized replaceables (kinds 30175–30178) and kind:5 tombstones stuck with `pending_sync = 1` forever. Desktop re-publishes the retained pre-signed `raw_event` every ~30s; once `created_at` falls outside the relay ingest skew window the relay rejects every attempt and nothing ever clears.

## Root cause

`flush_pending_events_at` in `desktop/src-tauri/src/managed_agents/persona_events.rs` replayed retained pre-signed events for most kinds.

Relay ingest enforces `MAX_TIMESTAMP_DRIFT_SECS = 900`. After that window:

- publish → `invalid: event timestamp too far from server time`
- row stays `pending_sync = 1`
- next flush repeats forever (reject storm)

NIP-IA identity-archive requests already re-signed with a fresh timestamp at publish; parameterized replaceables and tombstones did not.

## Fix

At flush/publish time, re-sign events that need a fresh timestamp via the same path as archive requests (`resign_with_fresh_timestamp`):

- parameterized replaceables 30175–30178 (`is_parameterized_replaceable`)
- kind:5 tombstones
- existing archive-request path unchanged

Preserves kind, tags, and content. `mark_synced` still keys off the retained original `created_at` + content so local retention rows clear correctly after a successful publish.

Helper: `needs_fresh_timestamp_at_publish`.

## Why it matters

Without this, managed-agent persona/retention traffic can permanently hammer the relay with undeliverable events and never converge — a reliability bug for every desktop that retains those kinds across the ±15m window (sleep, offline, long-lived agents). Independent confirmations on #4967 (`unclesvf`, `bokiko`).

## Test plan

```bash
. ./bin/activate-hermit
just _ensure-sidecar-stubs
cd desktop/src-tauri && cargo test --lib flush_barrier -- --nocapture
```

- [x] `flush_barrier` unit tests: **5/5** @ head `915a977` (rebased on current `main`)
- [x] DCO / Semgrep OSS / zizmor green on prior heads; re-run expected after rebase push
- [ ] Full monorepo `just ci` not run on this change (desktop Tauri lib filter only)
- [ ] Optional soak: age a retained replaceable past 15m, confirm next 30s flush accepts and `pending_sync` clears

## Risk / blast radius

- **In scope:** desktop managed-agent retention flush publisher only
- **Not changed:** relay ingest rules, non-retained publish paths, replaceable semantics beyond fresh `created_at` at send
- **Sync keying:** intentionally still original `created_at`+content so we do not orphan retention rows after re-sign

## Closest existing work

none found for this root cause (issue #4967 open; this is the only linked fix PR)

Fixes #4967

Retested after rebase onto current main @ `915a977d2`.

### Acceptance retest (2026-08-18)
Rebased onto current `main`. Mini proof @ `73c511bea`: cargo test --lib flush_barrier — 5/5.

## Mini retest (2026-08-24 acceptance)
- Head `db2be20fb1` · cargo test --lib flush_barrier → 5 passed
- Rebased onto current upstream `main` (force-with-lease, pre-review)

### DIFF
--- a/desktop/src-tauri/src/managed_agents/persona_events.rs
+++ b/desktop/src-tauri/src/managed_agents/persona_events.rs
@@ -307,15 +307,15 @@ pub(crate) async fn flush_pending_events_at(
         let event = nostr::Event::from_json(&current.raw_event)
             .map_err(|e| format!("failed to parse retained event '{}': {e}", current.d_tag))?;
 
-        // NIP-IA requests are freshness-checked by the relay (±120s on
-        // `created_at`), so a request retained while the relay was
-        // unreachable would be permanently stale. Re-sign with a fresh
-        // timestamp at publish time; kind, tags, and content are preserved,
-        // and `mark_synced` below still compares against the retained row's
-        // original `created_at`/`content`, which are untouched.
-        let is_archive_request =
-            buzz_core_pkg::kind::is_identity_archive_request_kind(current.kind);
-        let event = if is_archive_request {
+        // Relay ingest rejects events outside ±900s of server time
+        // (`MAX_TIMESTAMP_DRIFT_SECS`). Retention may hold a row longer than
+        // that while the relay is down or after a failed sweep, so replaying
+        // the retained pre-signed `raw_event` would be rejected forever.
+        // Re-sign at publish with a fresh `created_at`; kind, tags, and
+        // content are preserved. `mark_synced` below still keys off the
+        // retained row's original `created_at`+`content` (untouched by
+        // re-sign).
+        let event = if needs_fresh_timestamp_at_publish(current.kind) {
             resign_with_fresh_timestamp(&event, state)?
         } else {
             event
@@ -351,12 +351,27 @@ pub(crate) async fn flush_pending_events_at(
     Ok(flushed)
 }
 
+/// Whether a retained event must be re-signed with a fresh `created_at` at
+/// flush time (see `flush_pending_events_at`).
+///
+/// Covers identity-archive requests, NIP-33 parameterized replaceables
+/// (30175/30176/30177/30178 and siblings), and NIP-09 tombstones (kind:5) —
+/// all gated by the relay's ±900s timestamp window.
+fn needs_fresh_timestamp_at_publish(kind: u32) -> bool {
+    use buzz_core_pkg::kind::{
+        is_identity_archive_request_kind, is_parameterized_replaceable, KIND_DELETION,
+    };
+    is_identity_archive_request_kind(kind)
+        || is_parameterized_replaceable(kind)
+        || kind == KIND_DELETION
+}
+
 /// Re-sign a retained event with the current owner keys and a fresh
 /// `created_at`, preserving kind, tags, and content.
 ///
-/// Used for relay-freshness-checked kinds (NIP-IA 9035/9036) that would
-/// otherwise go permanently stale sitting in the retention store while the
-/// relay is unreachable. `.allow_self_tagging()` mirrors
+/// Used for kinds that would otherwise go permanently stale sitting in the
+/// retention store past the relay's ±900s ingest window (see
+/// [`needs_fresh_timestamp_at_publish`]). `.allow_self_tagging()` mirrors
 /// `events::build_archive_identity_request` — nostr strips `p` tags matching
 /// the signer by default, which would corrupt a self-targeted request.
 ///
--- a/desktop/src-tauri/src/managed_agents/persona_events/tests.rs
+++ b/desktop/src-tauri/src/managed_agents/persona_events/tests.rs
@@ -867,6 +867,100 @@ mod flush_barrier {
         assert_ne!(fresh.as_json(), stale.as_json());
     }
 
+    #[test]
+    fn managed_agent_resign_refreshes_timestamp_and_preserves_payload() {
+        use buzz_core_pkg::kind::KIND_MANAGED_AGENT;
+        use nostr::JsonUtil;
+
+        let keys = nostr::Keys::generate();
+        let content = r#"{"name":"agent","display_name":"Agent"}"#;
+        let stale = EventBuilder::new(Kind::Custom(KIND_MANAGED_AGENT as u16), content)
+            .tags([Tag::parse(["d", "agent-slug"]).unwrap()])
+            .custom_created_at(nostr::Timestamp::from(1))
+            .sign_with_keys(&keys)
+            .unwrap();
+        let state = build_app_state();
+        *state.keys.lock().unwrap() = keys;
+
+        let fresh = resign_with_fresh_timestamp(&stale, &state).unwrap();
+
+        assert!(fresh.created_at.as_secs() > stale.created_at.as_secs());
+        assert_eq!(fresh.kind, Kind::Custom(KIND_MANAGED_AGENT as u16));
+        assert_eq!(fresh.content, content);
+        assert_eq!(fresh.tags, stale.tags);
+        assert!(fresh.verify_id());
+        assert!(fresh.verify_signature());
+        assert_ne!(fresh.as_json(), stale.as_json());
+    }
+
+    #[test]
+    fn needs_fresh_timestamp_covers_replaceables_tombstones_and_archive() {
+        use buzz_core_pkg::kind::{
+            KIND_DELETION, KIND_IA_ARCHIVE_REQUEST, KIND_IA_UNARCHIVE_REQUEST, KIND_MANAGED_AGENT,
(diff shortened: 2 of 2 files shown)


olympusbuildz:
Closing stale open PR with no recent progress to protect olympusbuildz merge reputation. Ping to reopen if maintainers still want this.

PR #5379, opened 2026-08-09
fix(acp): raise model-probe timeout and make it configurable

### LINKED ISSUE #2261: [Bug] Windows: Codex model discovery times out / empty models (cold start >10s)

NedMalki-Chief:
Closes #2261.

## Problem

`MODELS_TIMEOUT` (`crates/buzz-acp/src/lib.rs`) was a hardcoded `Duration::from_secs(10)`. That covers a warm adapter, but a cold probe boots Node plus the vendor app server before answering `initialize`, and on Windows that routinely exceeds it.

Measured cold probes of `buzz-acp models --json` on Windows 11, timing the real binary:

| Adapter | Cold probe | vs 10s budget |
|---|---|---|
| `codex-acp` 1.1.7 | 9-12s | oscillates across the line |
| `codex-acp` 1.1.14 | 11-12s | consistently over |
| `claude-agent-acp` 0.66.0 | 29-31s | consistently over, by 3x |

Claude Code never fits. Codex fits only when the OS file cache happens to be warm, which is why this reads as intermittent.

The failure mode is bad: an empty model picker or a spinner that never resolves, indistinguishable from a broken install. And it is unrecoverable by the user, because the constant is reachable from neither config nor env, unlike `BUZZ_ACP_IDLE_TIMEOUT` and `BUZZ_ACP_TURN_TIMEOUT`.

## Change

- Default 10s → **45s**, clearing the slowest measured adapter with margin. This is recommendation 1 in #2261 ("raise to 30-45s").
- Add **`BUZZ_ACP_MODELS_TIMEOUT`** (whole seconds), following the existing `BUZZ_ACP_IDLE_TIMEOUT` convention, so a slower machine or a future adapter can be tuned without a rebuild.
- A malformed, zero, or negative override falls back to the default rather than failing the probe: an unusable override must never be a harder failure than no override.

Scope is deliberately narrow — one constant becomes a function, three call sites bind it locally so the timeout reported in the error message is always the one actually applied. I did not touch the UX items in #2261 (retry affordance, caching successful catalogs); those are separate and worth doing, but this is the part that makes discovery work at all.

## Verification

Windows 11, toolchain 1.95.0, release binary built from this branch:

- `codex-acp`: full catalog returned in **14s and 19s** (8735 bytes) — both fatal under the old budget.
- `claude-agent-acp`: returned in **29s and 31s** (1269 bytes) — previously always timed out.
- `BUZZ_ACP_MODELS_TIMEOUT=2` still fails fast at 2s with `error: agent timed out (2s)`, confirming the override is applied and the message reflects the effective value.
- `cargo test -p buzz-acp --lib models_timeout` — 4 new unit tests pass (default when unset/blank, whole-second parsing, unusable-value fallback, and a guard that the default stays ≥30s so a future edit cannot silently reintroduce a too-tight budget).
- `cargo fmt -p buzz-acp -- --check` clean.

Commit is DCO signed-off.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -64,11 +64,14 @@ fn is_subcommand(name: &str) -> bool {
 ///
 /// These probes are only "lightweight" once the adapter is running: a cold
 /// probe boots Node plus the vendor app server before `initialize` is answered.
-/// On Windows that routinely lands at 9-12s for both `codex-acp` and
-/// `claude-agent-acp`, so the previous 10s budget failed intermittently and
-/// surfaced as an empty model picker indistinguishable from a broken install.
+/// Measured cold probes on Windows span an order of magnitude across shipped
+/// harnesses: `codex-acp` 9-12s, `claude-agent-acp` 29-31s, `opencode acp` 58s.
+/// The previous 10s budget failed for all but the fastest, surfacing as an
+/// empty model picker indistinguishable from a broken install. The default
+/// must clear the slowest known harness with margin, because a false timeout
+/// is a worse failure than a slow success: it is silent and misdiagnosed.
 /// Override via `BUZZ_ACP_MODELS_TIMEOUT` (whole seconds).
-const DEFAULT_MODELS_TIMEOUT_SECS: u64 = 45;
+const DEFAULT_MODELS_TIMEOUT_SECS: u64 = 120;
 
 /// Parse `BUZZ_ACP_MODELS_TIMEOUT` (whole seconds), falling back to
 /// [`DEFAULT_MODELS_TIMEOUT_SECS`]. A malformed or zero value falls back
@@ -7897,9 +7900,10 @@ mod models_timeout_tests {
 
     #[test]
     fn default_clears_observed_windows_cold_start() {
-        // Cold `codex-acp` / `claude-agent-acp` probes on Windows measured
-        // 9-12s; the previous 10s budget failed intermittently.
-        assert!(DEFAULT_MODELS_TIMEOUT_SECS >= 30);
+        // Slowest measured shipped harness is `opencode acp` at 58s cold on
+        // Windows; the default must clear it with margin so a future edit
+        // cannot silently reintroduce a too-tight budget.
+        assert!(DEFAULT_MODELS_TIMEOUT_SECS >= 90);
     }
 
     #[test]
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -59,8 +59,35 @@ fn is_subcommand(name: &str) -> bool {
     std::env::args().nth(1).map(|a| a == name).unwrap_or(false)
 }
 
-/// Timeout for lightweight helper subcommands (spawn + initialize + model/method probes).
-const MODELS_TIMEOUT: Duration = Duration::from_secs(10);
+/// Default timeout for lightweight helper subcommands (spawn + initialize +
+/// model/method probes).
+///
+/// These probes are only "lightweight" once the adapter is running: a cold
+/// probe boots Node plus the vendor app server before `initialize` is answered.
+/// On Windows that routinely lands at 9-12s for both `codex-acp` and
+/// `claude-agent-acp`, so the previous 10s budget failed intermittently and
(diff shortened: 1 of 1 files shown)


NedMalki-Chief:
Pushed `16de089` raising the default from 45s to **120s**, and the reason is worth stating explicitly because it changes the shape of the problem.

I went on to install the other shipped harness presets to check the fix generalised, and found a cold probe well above the 45s I originally proposed:

| Harness | Cold probe (Windows 11) |
|---|---|
| `codex-acp` | 9-12s |
| `claude-agent-acp` | 29-31s |
| **`opencode acp`** | **58s** |
| `devin acp` | 7s |
| `cursor-agent acp` | 8s (to `Authentication required`) |

45s would have fixed Codex and Claude Code while leaving **OpenCode broken in exactly the same way** — same empty dropdown, same "Could not load live models", just a different harness. That is the failure this change exists to remove, so shipping 45s would have been a partial fix that looked complete.

The spread is roughly an order of magnitude across shipped presets (7s to 58s), which argues the default should be set by the slowest known harness plus margin rather than by a typical one. 120s does that.

On the obvious objection, that 120s is a long time to wait on a genuinely broken adapter: a false timeout is the worse failure of the two. A slow success is visibly slow and self-explaining; a premature timeout is silent and reads as a broken install, which is precisely why #2261 took real effort to diagnose. The probe is on-demand, so the higher ceiling costs nothing when adapters are fast, and `BUZZ_ACP_MODELS_TIMEOUT` tunes it down for anyone who wants to fail faster.

Happy to settle on a different number if you have data from slower or faster machines. The measurements above are one Windows 11 box, so treat 58s as a lower bound on the worst case rather than the worst case itself.

PR #5398, opened 2026-08-09
ci(release): gate OSS desktop auto-update promotion

wesbillman:
## Summary

Separate OSS desktop artifact publication from fleet-wide auto-update promotion.

- retain the exact generated updater manifest as `updater-manifest.json` on each immutable `desktop-vX.Y.Z` release
- stop the tag-triggered build from mutating `buzz-desktop-latest/latest.json`
- add a `main`-only manual promotion workflow with one global concurrency group
- validate stable semver, release/tag commit identity, draft/prerelease state, exact platform set, signatures, version-bound asset URLs, asset existence, monotonicity, idempotent retries, and a final stale-state check before writing
- document the operator flow and pin the split with focused contract tests

## Safety behavior

Publishing a versioned GitHub release no longer exposes it through the in-app updater. Operators can install and test those exact signed/notarized artifacts, then manually run **Promote OSS Desktop Auto-Update** with the stable version.

Promotion rejects downgrades. A same-version retry succeeds only when the rolling and candidate manifests are byte-identical. The workflow re-reads the current rolling version immediately before its only write and records the actor, source tag commit, previous version, manifest digest, and run URL.

## Verification

Verified at commit `39caf1603be06bb476905225ec55f7bbbe86b237`:

```text
scripts/test-oss-desktop-promotion.sh
OSS desktop promotion contract passed

scripts/test-release-ref-contract.sh
release ref contract passed

git diff --check origin/main...HEAD
(clean)
```

The repository pre-push hook also passed `branch-skew` for the exact pushed head; package suites were correctly skipped because this change only touches release workflows, scripts, and documentation.

Originating conversation: Buzz channel `separate-publish-step-release`, thread `8857ce8bbe928e891165eddcf06c666cf6eae16181c3f02a6d8c396d8a536026`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,10 @@ jobs:
         run: scripts/test-release-ref-contract.sh
       - name: Desktop release candidate contract
         run: scripts/test-desktop-release-candidate.sh
+      - name: OSS desktop promotion contract
+        run: |
+          scripts/test-oss-desktop-promotion.sh
+          scripts/test-oss-desktop-promotion-behavior.sh
       - name: Mobile release contract
         run: |
           scripts/test-mobile-release-contract.sh
--- a/.github/workflows/promote-oss-desktop-release.yml
+++ b/.github/workflows/promote-oss-desktop-release.yml
@@ -1,0 +1,45 @@
+name: Promote OSS Desktop Auto-Update
+run-name: Promote desktop-v${{ inputs.version }} to auto-update
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Stable desktop version to promote (X.Y.Z)
+        required: true
+        type: string
+
+concurrency:
+  group: oss-desktop-auto-update-promotion
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
+jobs:
+  promote:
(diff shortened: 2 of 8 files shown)


wesbillman (commented):
Reviewed at `09b67768ab30ee2db2fcebc7d7209e35263707c9` on Wes’s behalf. I would approve, but GitHub does not permit the PR author account to approve its own PR.

The release/promotion split is sound: immutable versioned publication retains the exact `updater-manifest.json`, while the manual main-only workflow validates release/tag identity, the complete platform set, signatures, canonical version-bound asset URLs, asset existence, monotonic versioning, identical retries, and stale rolling-manifest state before the sole write. Post-write digest verification makes failures explicit.

I also verified the candidate is copied into `staged/` before the complete versioned asset upload, so it is actually retained on `desktop-vX.Y.Z`. The intentional contract change removes automatic rolling-manifest mutation from `release.yml`; the corresponding release-ref expectations were updated and dedicated structural/behavior coverage was added.

Local verification on that exact clean head:
- `bash scripts/test-oss-desktop-promotion.sh` — pass
- `bash scripts/test-oss-desktop-promotion-behavior.sh` — pass
- `git diff --check origin/main...HEAD` — pass

CI is green. No blocking findings.

Issue #5422, opened 2026-08-09
Define standing-context re-delivery policy for provider compaction


wesbillman:
## Problem

Legacy ACP sessions now receive standing context once per live Buzz session. If the provider compacts that context away while the session remains live, Buzz has no signal or threshold that causes the persona/base/system context to be sent again.

## Desired outcome

Choose and implement a policy for legacy providers, such as:

- re-deliver standing context after a turn/token threshold;
- configure a non-zero session rotation default; or
- use a provider capability/signal when available.

The policy must avoid restoring the old behavior of repeating all standing context on every turn.

Follow-up from #5342 session-context delivery work; this is a product/session-lifecycle decision rather than a defect in the bounded delivery-state change.

Issue #5421, opened 2026-08-09
Clarify conversation-context counts after session delta filtering


wesbillman:
## Problem

Session-aware prompt delivery can filter messages that the live ACP session already received. The rendered heading still uses the post-filter message count with the pre-filter total and truncation state, for example:

```text
[Thread Context (1 of 5 messages)]
```

That does not distinguish messages withheld because they were delivered earlier in the session from messages omitted by the fetch window.

## Desired outcome

Make the conversation-context heading distinguish:

- messages included in this prompt;
- messages withheld because the current session already received them; and
- messages omitted because the fetch was truncated.

Follow-up from #5342 session-context delivery work; intentionally kept out of that bounded PR because the rendering semantics need a product/copy decision.

PR #5423, opened 2026-08-09
Reduce repeated ACP session context

wesbillman:
## Summary

- deliver legacy ACP standing context once per live session, committing delivery state only after a successful turn
- send only new thread/DM event deltas on later turns, with fail-open behavior for missing IDs and failed/cancelled prompts
- fence native steer delivery acknowledgements by ACP session identity so stale acks cannot poison replacement sessions
- keep context hints truthful when a fetch contains only the triggering event versus history delivered earlier

## Validation

The pre-push hook passed on exact pushed head `6a768f1bc80fe63c686acf8d730f177fff8add3c`:

- `branch-skew`
- `desktop-check`
- `desktop-typecheck`
- `desktop-test`
- `rust-tests`
- `desktop-tauri-checks`

Focused regression tests were also run while iterating:

- `channel_prompt_commits_delivery_state_only_after_acp_success`
- `in_flight_stale_native_steer_ack_cannot_update_replacement_session`
- thread/DM trigger-only versus previously-delivered context hint tests

## Known limitations and follow-ups

A local Goose smoke timed out at `session/new`. This diff does not change code that executes at or before `session/new`; its earliest affected runtime behavior is delivery-state insertion after session creation succeeds. The smoke failure is therefore bounded as environmental or pre-existing, but no successful live-provider turn was obtained. Scripted ACP wire/lifecycle tests carry the regression coverage.

- #5421 — distinguish post-delta, already-delivered, and fetch-truncated context counts
- #5422 — define a standing-context re-delivery policy if a legacy provider compacts it away

Durable process-restart/session resume remains out of scope for this slice of #5342. #5386 also remains separate pending upstream adapter support.


crates/buzz-acp/src/pool.rs
```diff
@@ -1884,6 +1981,30 @@ pub async fn run_prompt_task(
         } else {
             None
         };
+        let triggering_ids: HashSet<String> = b
```
wpfleger96: This set only includes `b.events`, but `format_prompt()` also renders every `b.cancelled_events` entry in a merged cancel/re-prompt. After that merged prompt succeeds, those carry-over IDs are therefore absent from `delivered_event_ids`. A later thread/DM context fetch can resend the already-delivered messages; if the current fetch already contains one, it can also appear twice in this same prompt (conversation context plus `[What you were working on]`).

Could you build one rendered-batch ID set from `b.events.iter().chain(b.cancelled_events.iter())`, use it both for context-delta exclusion and `pending_delivered_event_ids`, then add a regression test proving a successful merged turn commits both the carry-over and new IDs and does not replay either on the next turn?

wpfleger96 (dismissed):
I found one correctness gap in the new delivery ledger. The success-only commit boundary and session fencing look solid, but merged carry-over events need to participate in the same delivered-ID accounting as current-batch events.

crates/buzz-acp/src/pool.rs
wesbillman: Fixed in `ab85b03aa81d2b8588ffe649f6eff0c314a385b9`. The prompt path now builds one rendered-batch ID set from `events.chain(cancelled_events)`, uses it for context exclusion, and stages the same set for success-only delivery commit. Added `merged_cancel_prompt_commits_and_deduplicates_all_rendered_event_ids`, which checks both IDs enter the ledger, appear once in the merged wire prompt, and neither replays on the next turn. The branch is also merged with current `main` in `28da9bb92c7575a5a835a7284762cea5a997c620`; the full pre-push gate passed on `ab85b03aa`.

wesbillman (commented):
Carl review on Wes’s behalf at exact head `ab85b03aa81d2b8588ffe649f6eff0c314a385b9`: **cleared / would approve**, but GitHub will not accept an approval from the PR author’s account.

The prior carry-over delivery-ledger blocker is resolved: both current and cancelled/merged event IDs now enter the success-only commit set, with exact-wire regression coverage. I also traced legacy standing-context delivery, conversation-context delta filtering, session invalidation/rotation, and native-steer ack ordering. Successful steer deliveries are fenced by ACP session ID across both pre-result staging and post-result idle-agent updates, so late acknowledgements cannot contaminate a replacement session. Failure and cancellation paths preserve retryability rather than committing delivery state.

Validation at this head:
- `cargo test -p buzz-acp` — 736 unit tests + 9 lifecycle integration tests passed
- `git diff --check` — clean

No blocking findings.

wpfleger96 (approved):
I double-checked the carry-over fix at exact head `ab85b03aa81d2b8588ffe649f6eff0c314a385b9`. The merged batch now uses one ID set from `events.chain(cancelled_events)` for both context exclusion and success-only delivery commit. The new exact-wire regression verifies that each merged event appears once, both IDs enter the ledger after success, and neither replays on the following turn.

The merge resolution also preserves both delivery-state initialization and the incoming usage baseline initialization. All check runs at this head are complete and green. I don't see any remaining blockers.

Issue #4930, opened 2026-08-05
Buzz Term hangs the entire app on Windows (0.5.5) - banner flashes, then Application Hang; blank panel after relaunch


BoydVC:
## Environment

- Buzz Desktop **0.5.5.0** (arrived via auto-update from 0.5.4 today)
- Windows 11 Pro, build 26200
- Self-hosted relay (`deploy/compose` bundle), Let's Encrypt TLS

## What happened

Opened Buzz Term from the channel workspace for the first time:

1. The Term banner/wordmark flashed briefly, with what looked like a local shell prompt.
2. The entire app immediately stopped responding (not just the Term panel).
3. Windows Event Log recorded an **Application Hang** for `buzz-desktop.exe` 0.5.5.0 at that moment: *"The program buzz-desktop.exe version 0.5.5.0 stopped interacting with Windows and was closed."*
4. After relaunching, the Term panel renders permanently blank. The rest of the app works normally.

## Expected

Either a working shell on Windows (ConPTY + PowerShell/cmd), or a graceful "not supported on this platform yet" state. A failed shell spawn should never hang the UI thread or take down the whole app.

## Possibly relevant

PR #4347's shell resolution chain (`$SHELL` env var → passwd database → `/bin/sh`) appears to have no valid candidate on a stock Windows install — none of those exist here. The Application Hang (rather than a crash/panic) suggests the failure path blocks the main thread while the frontend waits on a PTY that never materializes.

Happy to provide more diagnostics — though note the desktop currently writes no log files on Windows, so Event Log was the only artifact available.

Glucksberg:
Reproduced this and traced it. Opened #5425 with a fix.

Your read was right on both counts, and each one is a layer deeper than it looks.

**The shell resolution chain really does have no valid candidate** — but not because `$SHELL` is missing. `resolve_shell` was `#[cfg(unix)]`-only, so on Windows there was no resolution step at all. What actually spawns the child is `ComSpec`, and `env_fence::fence_env` had already deleted it: the fence calls `env_clear()` and rebuilds from an allowlist that only listed Unix variables. `portable-pty` reads `ComSpec` out of the builder's own env map and passes the result to `CreateProcessW` as `lpApplicationName`, which does **no path search** — so with `ComSpec` gone it gets the bare string `"cmd.exe"` and returns `os error 2`. `USERPROFILE` was cleared the same way, which is the `cwd None` in the error.

[...]

PR #5427, opened 2026-08-09
fix(desktop): surface Buzz Term attach failures instead of a blank panel

Glucksberg:
## Summary

When the Term panel fails to attach, it renders **nothing** — a blank body with a `+` button that does nothing. `TerminalBootstrap.fail()` removes the session, sets `available = false`, and reports the error to `console.error`. In a production build there is no devtools to read that, and on Windows there is no log file either, so the error the backend produced is simply discarded on the way to the user.

That is the surface #4930 was reported through: the reporter saw a permanently blank Term panel and had nothing to attach to the issue except a Windows Event Log entry. The backend had a precise, actionable error the whole time — `CreateProcessW "cmd.exe" in cwd None failed … (os error 2)` — and the UI threw it away.

This renders that error in the viewport instead, with a **Retry** action that clears it and re-runs the auto-create path.

### Before / after

Screenshots in the first comment. Both are the same element (`.buzz-terminal-substrate`) at the same viewport, so they're directly comparable.

### Why there is no notice for "no channel selected"

An earlier version of this change also rendered `"Select a channel to open a terminal."` when the panel had no channel context. That state is not reachable, and shipping the notice would have been a small regression:

- `TerminalBootstrap.tsx` — the Cmd/Ctrl+J toggle returns early when `!context`, so the panel cannot be opened without a channel.
- An effect closes the panel when `context` goes away while it is open.

So the only way to render that notice is the single frame between navigating away from a channel and the effect running. That's a flash during navigation, not a state a user can be in. The existing test `a non-channel route closes the panel and ignores the terminal shortcut` already asserts this behaviour. The notice is now scoped to the one case that genuinely occurs: `!available`.

### Relationship to #5425

Related, not a duplicate, and they fix different things.

- **#5425** (Rust) makes the Windows attach *succeed* — three stacked backend defects.
- **This PR** (React) makes an attach failure *visible*.

Neither subsumes the other. With #5425 merged, a Windows user shouldn't reach the failure path by that route — but any future attach failure, on any platform, is silent again without this change. They can merge in either order; they touch no common files.

## Related issue

Related to #4930 — this addresses the *"Term panel renders permanently blank"* symptom, which #5425 explicitly leaves out of scope. Deliberately not marked `Fixes`, since the root cause of that issue is the backend spawn failure, not this rendering gap.

Searched for duplicates: no other open PR touches `desktop/src/features/terminal`.

The issue's remaining item — *"the desktop currently writes no log files on Windows"* — is not addressed here.

## Testing

- **Unit** (`TerminalBootstrap.test.mjs`, 10/10): the notice renders with the backend's own error text, and Retry re-attaches after a failure. Full desktop JS suite: **4536/4536**.
- **E2E screenshot spec** (`term-empty-states-screenshots.spec.ts`, registered in the `smoke` project): drives the real attach-failure path end to end. The mock bridge doesn't implement `terminal_attach`, so the panel takes the genuine failure route rather than a stubbed one — the notice, its text, and the Retry button are all asserted before the capture.
- `pnpm check` (biome, file sizes, px-text, pubkey truncation) and `pnpm typecheck` clean.

One note on reading the screenshot: the error string in it is the mock bridge's (`Unsupported mocked Tauri command: terminal_attach`), because that is what fails in an e2e build. On Windows the same slot carries the real `CreateProcessW …` error. The point of the shot is that whatever the backend said now reaches the user.

### DIFF
--- a/desktop/playwright.config.ts
+++ b/desktop/playwright.config.ts
@@ -20,6 +20,7 @@ export default defineConfig({
       name: "smoke",
       testMatch: [
         "**/smoke.spec.ts",
+        "**/term-empty-states-screenshots.spec.ts",
         "**/onboarding-docked-cta-screenshots.spec.ts",
         "**/identity-key-help.spec.ts",
         "**/key-import-reveal.spec.ts",
--- a/desktop/src/features/terminal/TerminalBootstrap.test.mjs
+++ b/desktop/src/features/terminal/TerminalBootstrap.test.mjs
@@ -762,3 +762,60 @@ test("retry reattaches a failed new tab while another session remains", async ()
   );
   view.unmount();
 });
+
+test("retry waits for the channel whose attachment failed", async () => {
+  const { createElement } = await import("react");
+  const { act, fireEvent, render, waitFor } = await import(
+    "@testing-library/react"
+  );
+  const { ThemeProvider } = await import("@/shared/theme/ThemeProvider");
+  const { TerminalBootstrap } = await import("./TerminalBootstrap.tsx");
+
+  const tree = (channelId, channelName) =>
+    createElement(
+      ThemeProvider,
+      null,
+      createElement(TerminalBootstrap, {
+        channelId,
+        channelName,
+        npub: "npub1owner",
+        relayUrl: "wss://relay.example",
+        threadId: null,
+      }),
+    );
+  const attachCount = (channelId) =>
+    calls.filter(
+      ({ command, args }) =>
+        command === "terminal_attach" && args.request.channelId === channelId,
+    ).length;
+
+  const view = render(tree("channel-a", "alpha"));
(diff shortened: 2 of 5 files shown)


Glucksberg:
### Before — the silent panel

Attach fails, the session is removed, `available` goes false, and the viewport renders nothing. The `+` button is still there and still does nothing. This is what #4930 was reported through.

00-before

### After — the failure is the UI

The backend's own error text, plus a Retry that clears it and re-runs the auto-create path.

01-term-attach-error

The error string here is the e2e mock bridge's (`Unsupported mocked Tauri command: terminal_attach`) — that is the real thing that fails in an e2e build, so the panel takes the genuine failure route rather than a stubbed one. On Windows the same slot carries `CreateProcessW "cmd.exe" in cwd None failed … (os error 2)`.

Both images are `.buzz-terminal-substrate` captured at the same viewport, so they frame the identical element.

PR #5443, opened 2026-08-10
feat(desktop): redeploy provider-backed managed agents after config edits

unforcedagi:
## Problem

Desktop's edit dialog for a provider-backed managed agent lets you change model, env vars, respond-to, args, and parallelism — and Save silently writes only Desktop's local record. Nothing reaches the provider host. After the first deploy the Deploy action permanently becomes Shutdown, so there is no way to apply config changes to a running deployment short of hand-editing the host's unit file and bouncing the process yourself.

## Approach

The provider protocol already treats `deploy` as update-in-place: the host rewrites the agent's unit config and restarts it. So "redeploy" is just re-running the existing deploy path with the current record — no protocol or host changes.

- **`redeploy_managed_agent`** Tauri command (in `commands/agents/provider_access.rs`), reusing `build_deploy_payload` + `deploy_to_provider`. Guarded by `redeploy_provider_target`: provider-backed and already-deployed agents only. Payload/deploy failures persist to the record via the same `persist_failure` path reconciliation uses.
- **Redeploy quick action** on the agent profile panel, shown to the owner of a deployed provider-backed agent. Local agents keep Restart; both flow through a shared `getManagedAgentSecondaryAction` helper.
- **Honest copy** in the edit dialog's Run-on summary: saved changes don't reach a deployed agent until you redeploy it.

Two extractions keep over-limit files inside the file-size ratchet: the profile panel's mutation cluster moved to `useAgentProfileMutations`, and the redeploy mutation lives in its own module.

## Testing

- `cargo test --workspace` from `desktop/src-tauri`: 2372 passed, 0 failed (includes new `redeploy_guard_requires_provider_backend_and_existing_deployment`)
- `pnpm test`: 4536 passed, 0 failed (includes new secondary-action cases)
- `pnpm typecheck`, `pnpm check` (biome + file-size/px/pubkey guards), `cargo fmt --check`, clippy: clean

### DIFF
--- a/crates/buzz-backend-kubernetes/src/reconcile.rs
+++ b/crates/buzz-backend-kubernetes/src/reconcile.rs
@@ -843,6 +843,51 @@ mod tests {
         );
     }
 
+    /// The seam the desktop's "Redeploy" button reports against.
+    ///
+    /// Buzz Desktop lets an owner re-send a deployed agent's configuration and
+    /// stamps the returned `agent_id` on its record. This pins what that
+    /// answer is worth when the instance is live: the id is the one the
+    /// desktop already held, and *nothing* was applied — no Secret carrying
+    /// the edited environment was written, so the new configuration never even
+    /// reached the cluster. A desktop that reported "Redeployed" would be
+    /// reporting this call as a runtime change; what it may honestly report is
+    /// delivery plus the generation boundary (`:867-877`).
+    ///
+    /// Both shapes of edit are covered, because they fail differently: a
+    /// changed env *value* (an edited model or system prompt) does not even
+    /// move the create-intent fingerprint, and a changed env *key* does — and
+    /// the live row no-ops either way.
+    #[test]
+    fn started_pod_returns_its_id_having_applied_nothing() {
+        let id = identity();
+        let cfg = config();
+        let fake = Fake::default().with_pod(our_pod(&id, &cfg, Some(running())));
+
+        let mut edited = env();
+        edited.insert("BUZZ_RELAY_URL".into(), "wss://edited-after-deploy".into());
+        edited.insert("BUZZ_ACP_MODEL".into(), "edited-model".into());
+
+        let agent_id = block_on(deploy(&fake, &id, &cfg, edited)).expect("deploy answers");
+
+        assert_eq!(
+            agent_id,
+            id.pod_name(),
+            "the answer is the id the caller already had — success cannot \
+             distinguish an applied config from a no-op"
+        );
+        assert_eq!(
+            fake.mutations(),
+            [format!("ensure_namespace {}", cfg.namespace)],
+            "the live row mutated something while reporting success"
+        );
+        assert!(
+            fake.created_secrets.borrow().is_empty(),
+            "the edited configuration was written to the cluster — it must not \
+             be, and it is why success may not claim the agent is running it"
+        );
+    }
+
     /// Terminated → fenced delete → disappearance → recreate. The normal
     /// restart path, and the fence must carry the observed uid/rv.
     #[test]
--- a/desktop/src-tauri/src/commands/agent_models.rs
+++ b/desktop/src-tauri/src/commands/agent_models.rs
@@ -26,7 +26,7 @@ use crate::{
         DEFAULT_ACP_COMMAND,
     },
     relay::{relay_ws_url_with_override, sync_managed_agent_profile},
-    util::now_iso,
+    util::bumped_updated_at,
 };
 
 /// Query available models from an agent via `buzz-acp models --json`.
@@ -824,7 +824,7 @@ pub async fn update_managed_agent(
             record.respond_to_allowlist = prospective_allowlist;
         }
 
-        record.updated_at = now_iso();
+        record.updated_at = bumped_updated_at(&record.updated_at);
 
         save_managed_agents(&app, &records)?;
 
--- a/desktop/src-tauri/src/commands/agents.rs
+++ b/desktop/src-tauri/src/commands/agents.rs
@@ -1353,7 +1353,7 @@ pub async fn delete_managed_agent(
 // No backend Tauri command needed. Presence IS the status.
 #[path = "agents_deploy.rs"]
 mod deploy;
-pub(super) mod provider_access;
+pub(crate) mod provider_access;
 use deploy::build_deploy_payload;
 #[cfg(test)]
 use deploy::{deploy_payload_json, DeployProjections};
--- a/desktop/src-tauri/src/commands/agents.rs
+++ b/desktop/src-tauri/src/commands/agents.rs
@@ -6,15 +6,14 @@ use super::managed_agent_definition::validate_create_definition;
 use crate::{
     app_state::AppState,
     managed_agents::{
-        build_managed_agent_summary, current_instance_id, discover_provider_candidates,
-        ensure_persona_is_active, find_managed_agent_mut, load_managed_agents, load_personas,
-        load_teams, managed_agent_avatar_url, normalize_agent_args, provider_deploy,
-        resolve_provider_binary, save_managed_agents, start_managed_agent_process,
-        stop_managed_agent_process, stop_managed_agent_workspace_pair,
-        sync_managed_agent_processes, try_regenerate_nest, validate_provider_config, BackendKind,
-        CreateManagedAgentRequest, CreateManagedAgentResponse, ManagedAgentRecord,
-        ManagedAgentSummary, RelayMeshConfig, DEFAULT_ACP_COMMAND, DEFAULT_AGENT_PARALLELISM,
-        DEFAULT_AGENT_TURN_TIMEOUT_SECONDS,
+        build_managed_agent_summary, current_instance_id, ensure_persona_is_active,
+        find_managed_agent_mut, load_managed_agents, load_personas, load_teams,
+        managed_agent_avatar_url, normalize_agent_args, resolve_provider_binary,
+        save_managed_agents, start_managed_agent_process, stop_managed_agent_process,
+        stop_managed_agent_workspace_pair, sync_managed_agent_processes, try_regenerate_nest,
+        validate_provider_config, BackendKind, CreateManagedAgentRequest,
+        CreateManagedAgentResponse, ManagedAgentRecord, ManagedAgentSummary, RelayMeshConfig,
+        DEFAULT_ACP_COMMAND, DEFAULT_AGENT_PARALLELISM, DEFAULT_AGENT_TURN_TIMEOUT_SECONDS,
     },
     relay::{relay_ws_url_with_override, sync_managed_agent_profile},
     util::now_iso,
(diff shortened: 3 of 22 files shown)


wolfyy970 (changes requested):
The new action fixes a real dead end, but the command can report the wrong configuration as deployed.

redeploy_managed_agent snapshots the record and payload, releases the store lock for a provider call that can run for 600 seconds, then deploy_to_provider writes success into whichever same-pubkey record exists. If the owner saves another edit or starts a second redeploy while the first is running, the calls can finish out of order. The older payload can win remotely while the returned summary shows the newer local values and the toast says Redeployed. Deletion during the call can also leave a successful remote update with its result discarded.

Please serialize provider deployment per agent and bind completion to the captured backend and desired configuration generation. Updates and deletion need to share that boundary, or completion must not claim the current record was applied. Add a delayed-provider test where redeploy A starts, edit plus redeploy B completes, then A returns. The final provider and Desktop state must both represent B. Please also cover deletion during redeploy.

At 9f30824, the focused Rust and frontend tests, strict Clippy, formatting and diff checks pass. They do not cross the external-call race.

unforcedagi:
Thanks — the race was real, and reproducible: with the fix's serialization temporarily removed, the new delayed-provider test fails with the provider seeing payloads land out of order; with the completion binding forced to "unchanged", it fails by reporting A's stale deploy as current. Both protections are load-bearing. Addressed in 3cab402:

**Per-agent serialization.** `run_deploy` acquires a per-pubkey async mutex *before* snapshotting, so a queued deploy always reads the post-edit store. All four deploy paths — create, start, workspace-apply reconcile, redeploy — go through it and share the boundary.

**Completion bound to the captured generation.** The snapshot captures the record's `updated_at` as its configuration generation. On return, completion re-finds the record under the store lock and classifies: `AppliedCurrent` (only branch that claims success), `AppliedStale` (persists the remote facts — the deploy did happen — but errors telling the owner the agent is running the previous configuration), `BackendChanged` (stamps nothing), `Deleted` (persists nothing, no resurrection; reports the possibly-orphaned remote deployment). `update_managed_agent` now stamps a collision-proof generation (`bumped_updated_at`) so an edit always changes the generation even on a coarse or stepped-back clock.

**Edits and deletion take the second branch of your "or":** they don't acquire the deploy lock — blocking a config save behind a provider call that can run 600s seemed worse UX — so completion instead never claims the current record was applied unless the generation matches.

**Tests.** `provider_access_tests.rs`: the delayed-provider race exactly as you specified (redeploy A blocks on a oneshot-gated fake provider, edit + redeploy B completes, A returns; asserts the provider saw v1 then v2 and both provider and Desktop state represent B, with A erroring as stale), deletion-mid-flight, backend-switch-mid-flight, and unit coverage of every completion branch. Deterministic: the fake provider blocks until explicitly released; no assertion depends on scheduler timing.

At 3cab402: `cargo test --workspace` 2378 passed / 0 failed, strict clippy clean, fmt clean. Frontend untouched (no command signatures changed).

One behavior change worth naming: `reconcile_on_workspace_apply` now queues behind an in-flight deploy for the same agent, so a workspace apply can wait on a user-initiated redeploy (up to the provider timeout) where it previously interleaved. Correctness requires the shared boundary; flagging in case you'd rather it fail fast instead.

wolfyy970 (changes requested):
The per-agent serialization and delayed-provider tests are good. They close the out-of-order completion race I raised. I cannot clear the review because two product boundaries are still false.

1. Redeploy is offered whenever `backendAgentId` exists and success toasts `Redeployed`. But `docs/remote-agents.md` says a live Kubernetes generation is a strict no-op and edits apply only to the next fresh generation. A fake provider returning an id cannot prove the runtime changed. Either implement recycle/replace, or tell the owner the changes are queued for the next generation and test the built-in provider seam.

2. Completion compares only `record.updated_at`. The payload separately reloads the persona, team, global config, workspace relay and owner. Those inputs can change without touching the agent row, so an old payload can still be stamped `AppliedCurrent`. Bind completion to a digest or generation of the fully resolved payload.

Provider failures also stamp whichever same-pubkey record exists before checking its backend or generation. A successful remote call followed by a store-save failure still leaves no durable ambiguous-success receipt.

The mutex change is worth keeping, but it does not yet prove the desired configuration was applied.

unforcedagi:
Thanks — all three land in separate commits.

**1. "Redeployed" claimed something it hadn't earned** (6ddffb46e). You're right that a provider returning an `agent_id` only proves delivery. `docs/remote-agents.md` makes a live generation a strict zero-mutation no-op, so config edits reach a running remote agent on its next fresh generation and not before. Redeploy now returns a notice in the shape of the stop path — it reports the send and names the generation boundary — and the Run-on summary copy says the same. Two tests pin it, including one driving the shipped `buzz-backend-kubernetes` deploy loop against a live instance with an edited config: it returns the id the caller already had, having written no Secret at all.

**2. `updated_at` was the wrong generation** (ea6664e69). It only sees the agent row, and the payload is resolved from much more than that: `build_deploy_payload` reads the live persona, the team, the global agent config, the effective harness descriptor, the workspace relay URL and the owner pubkey, and the provider config rides beside it on the wire. Any of those moves with `updated_at` untouched. Completion is now bound to a canonical SHA-256 over the three things that actually go on the wire (provider id, provider config, resolved payload), and the payload is resolved *and* digested under the store lock in the same instant, so the bytes sent and the digest compared can never come from two different reads. `AppliedCurrent` requires both the generation and the digest to match. A moved digest is `AppliedStale`; a payload that can't be re-resolved on completion is a new `AppliedUnverified` that persists the remote facts and claims nothing. I gated failure stamping the same way — `last_error` on a record has to describe that record's own last attempt, so a provider error whose record has moved is now `FailedUnstamped`. There's a test where the persona changes mid-deploy with the agent row deliberately untouched, and one where an unchanged world still reports as current, so the honest report doesn't degrade into "always stale".

**3. Provider succeeds, store write fails** (256b97074). This was silently dropped, which left the worst state: something may be running remotely that the app has no record of asking for. `StoreOutcome` now separates "nothing happened" from "something happened remotely and we didn't record it", and only the second writes a receipt — one `0o600` JSON file per agent under `agents/deploy-receipts/`, atomic, mirroring the existing runtime receipts. It records the provider id, the deployment id returned, the digest of the input that was accepted, the generation it came from, and the store error that lost it. No secrets: the digest is a hash and no payload field is copied in. A pubkey that could escape the directory is refused rather than sanitized — a receipt filed under the wrong agent is worse than no receipt. The next completion for that agent that *does* save clears it, so a receipt on disk always means an outcome the store still doesn't reflect, and a deploy that starts with one outstanding logs it. A failed provider call whose save also failed isn't ambiguous, so it gets no receipt and both errors reach the owner.

Gate: `cargo test --lib` 2294 passed / 0 failed / 14 ignored; `just desktop-tauri-clippy` clean.

One thing worth your eye: `with_store` now re-resolves the payload while holding `managed_agents_store_lock`. `build_deploy_payload` doesn't take that lock today (it takes `state.keys`, in the same order already documented on `retain_managed_agent_pending`), and there's a comment saying so, but it's now an invariant that would deadlock the real path if it were ever broken. Happy to move the resolution outside the lock if you'd rather not carry that, at the cost of the digest and the store read no longer being one atomic observation.

wolfyy970 (changes requested):
The delivery copy, provider-seam test and full-input digest close the two blockers I raised. Those changes are solid.

One receipt bug still blocks this head. If deploy A succeeds remotely and its store save fails, Buzz writes the ambiguity receipt. If retry B then fails at the provider but its `last_error` saves, `saved` is true and `run_deploy` clears A's receipt. B did not supersede A remotely, so the provider may still be running A while the store never recorded it and Buzz has deleted the only evidence. Clear an ambiguity receipt only after a successfully saved provider success, and add that two-attempt regression.

Receipt I/O also needs its own no-follow boundary. `write_deploy_receipt_in` uses `atomic_write_json_restricted`, which canonicalizes an existing final path and follows a receipt symlink. The reader also follows symlinks and reads every JSON file without a size or count bound. Replace or reject the link entry, read through one descriptor with hard limits, and cover a symlink target remaining unchanged.

I am keeping changes requested at `256b97074`.

unforcedagi:
Both blockers from the review at `256b97074` are fixed, and the branch is rebased onto current `main` (was conflicting because the pushed head still sat on a very old base; there were no content conflicts).

**1. "Clear an ambiguity receipt only after a successfully saved provider success, and add that two-attempt regression."** — 9019e7917

`run_deploy` cleared on `saved` alone, so attempt B failing at the provider but saving its `last_error` deleted A's receipt. Clearing is now gated on `saved && deploy_result.is_ok()`.

The gate admits `AppliedStale` and `AppliedUnverified` deliberately: both persist `backend_agent_id` and `last_started_at`, so the store did record a provider success that superseded the earlier deploy remotely. `Deleted` and `BackendChanged` persist nothing, never save, and so never clear.

The regression is `a_saved_failure_does_not_clear_an_earlier_receipt`: A succeeds remotely and loses its store write, B fails at the provider and saves cleanly, and the receipt must survive still naming A's deployment id. It asserts B's failure genuinely saved (`saves == 1`, `last_error` stamped) so it cannot pass for the wrong reason.

**2. "Receipt I/O also needs its own no-follow boundary … Replace or reject the link entry, read through one descriptor with hard limits, and cover a symlink target remaining unchanged."** — 642c5c92f

- **Write** no longer uses `atomic_write_json_restricted`. Sharing it was the bug: canonicalizing the final path is right for the agent store, where dev worktrees legitimately symlink it into the shared dev data dir, but the dev sync shares data files only and never per-run state, so a symlink in the receipts directory is not a configuration. `write_receipt_file` creates the temp file `O_EXCL` `0o600` (never follows a link) and renames over the path, which **replaces** the link entry. Replacing rather than rejecting is deliberate — refusing to write would let anyone who can drop a symlink there suppress the very evidence the receipt exists to preserve. `sync_all` before the rename keeps the durability `AtomicWriteFile::commit` was providing.
- **Read** goes through one descriptor: `O_NOFOLLOW` on unix, `FILE_FLAG_OPEN_REPARSE_POINT` on windows, then regular-file and size checks by `fstat` against that same descriptor, so nothing can swap the path between the check and the read. The read is capped as well as the check, so a file that grows in between is truncated rather than followed. Bounds are 64 KiB per file and 256 files per read. Entry types come from the directory's own record, which costs no syscall against a link target.

New tests: a symlink at the receipt path is replaced and its target left byte-for-byte unchanged; the reader skips a symlink even when it points at a well-formed receipt and does not touch the target; an oversized file is skipped; the reader stops at the file cap.

**Verification** (`desktop/src-tauri`, whole crate, not just the new module):

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml` — 2469 passed, 0 failed, 15 ignored (plus 7 CSP, 3 mixer, 0 doc-tests)
- `cargo clippy --manifest-path desktop/src-tauri/Cargo.toml --workspace --all-targets -- -D warnings` — clean
- `cargo fmt --manifest-path desktop/src-tauri/Cargo.toml --all -- --check` — clean

PR #5459, opened 2026-08-10
fix(buzz-acp): don't cache transient sibling-gate lookup failures

### LINKED ISSUE #5450: buzz-acp: owner-only sibling gate caches a transport failure as not-a-sibling, silently blacklisting a teammate bot until restart
# buzz-acp: owner-only sibling gate caches a transport failure as "not a sibling", permanently blacklisting a teammate bot until process restart

## Summary

In `owner-only` respond mode, an inbound event from a **sibling** agent (a bot proving the same owner via its NIP-OA auth tag) is validated by querying the author's `kind:0` profile over REST. When that `/query` times out or errors, the check **fails closed** and treats the author as not-a-sibling. That **negative result is cached for the process lifetime**, so a single transient network blip permanently blacklists a teammate bot: every subsequent message it sends is silently dropped by the inbound author gate until the agent process is restarted.

This is observable as "a bot answers my (owner) messages instantly but ignores all teammate-bot messages for hours," then starts working again after a restart.

## Repro / evidence

Setup: multiple managed agents on one box, all `owner-only` respond mode, all owned by the same owner (so they are siblings to each other). A teammate bot triggers the agent (e.g. `@` in the channel).

1. Agent process restarts; at ~13s after launch it logs a `POST /query network error` (transient).
2. From that point the agent silently ignores every subsequent event from that sibling bot (8.5h in the observed case), while remaining fully responsive to the owner.
3. Owner messages still wake it instantly because they take the `author == my_owner` short-circuit and never touch REST/cache.
4. Restarting the agent process clears the blacklist (the sibling is immediately accepted again).

## Root cause (with source refs)

All in `block/buzz` `crates/buzz-acp/src/lib.rs`.

- **`owner-only` reviews siblings**: `RespondTo::OwnerOnly => is_owner_or_sibling(author, owner_cache, rest_client)` — and `is_owner_or_sibling` calls `check_sibling_via_profile` for an unknown author. (search `is_owner_or_sibling` + `RespondTo::OwnerOnly`)
- **Fail-closed**: `check_sibling_via_profile` returns `false` on any `/query` timeout or error:
  ```rust
  let resp = match tokio::time::timeout(..., rest_client.query(&[filter])).await {
      Ok(Ok(v)) => v,
      _ => return false, // timeout or error — fail closed
  };
  ```
- **Negative result is cached unconditionally, treating a transport error as an attestation**:
  ```rust
  let is_sibling = check_sibling_via_profile(author, my_owner, rest_client).await;
  owner_cache.cache_sibling(author.to_string(), is_sibling);   // cached regardless of success/failure
  ```
  and `is_known_sibling` returns the cached value (including `false`) on all later calls. `OwnerCache` documents the cache as process-lifetime with **no TTL** (`attestations are immutable`, capped only by a 256-entry wholesale clear).
- **The drop is silent**: the gate drops the event with
  ```rust
  tracing::debug!("inbound author gate — dropping event");
  ```
  which is invisible at default INFO level.

A transport error / timeout is **not** an attestation. Caching it as a permanent "not a sibling" conflates "we could not verify" (transient, should be retried) with "verified not a sibling" (durable). The comment claiming `attestations are immutable` is the mismatch: only *successful* verification results are attestations.

## Suggested fix

1. In `is_owner_or_sibling`, write to `OwnerCache` only when the sibling verification **succeeded**; on a transport failure, do **not** cache (or use a short TTL), so the next message from the same author retries the check instead of being dropped forever.
2. Upgrade the gate's drop log from `debug!` to `warn!` so these drops are observable.
3. Add a regression test: after a `/query` network failure for an author, the *next* message from the same author re-runs verification rather than short-circuiting to the cached negative.

## Workaround
Broad enough: in the affected agent, use `allowlist` respond mode with the teammate pubkeys added (allowlist lookup is an in-process set check, no REST), or restart the agent process to clear the stale cached negative.

whyihaveyou:
## Summary

Fixes #5450.

In `owner-only` / `allowlist` respond mode, the inbound author gate verifies a sibling agent (same owner, proven via the NIP-OA auth tag on the author's `kind:0` profile) with a REST `/query`. Previously the boolean outcome of that lookup was cached **unconditionally** — so a `/query` timeout or transport error was cached as `false` ("not a sibling") for the process lifetime. A single transient network blip permanently blacklisted a teammate bot until the agent process restarted, and the drop was logged at `debug` (invisible at the default INFO level), so the failure was silent.

## Changes

All in `crates/buzz-acp/src/lib.rs`:

- `check_sibling_via_profile` now returns a `SiblingCheck` enum (`Sibling` / `NotSibling` / `Indeterminate`) instead of `bool`. Transport/timeout errors, malformed relay responses, and "no profile event returned" (profile may not have replicated yet) are `Indeterminate` — nothing was attested. A fetched profile with no valid matching auth tag remains a definitive `NotSibling`.
- `is_owner_or_sibling` only caches **definitive** results. `Indeterminate` fails closed for the current event (unchanged security posture) but is **not cached**, so the next event retries the lookup instead of serving a cached transient error forever. A `warn` is emitted on the indeterminate path.
- The "inbound author gate — dropping event" log is raised from `debug` to `warn` so a mis-gated mention is visible to operators at the default log level.

## Tests

Two regression tests in `author_gate_tests` (backed by a local counting HTTP stub):

- `test_sibling_lookup_indeterminate_is_not_cached_and_retried` — an indeterminate lookup fails closed, is not cached, and the next event re-issues the HTTP query (request counter proves the retry).
- `test_sibling_lookup_definitive_negative_is_cached` — a fetched profile without a valid auth tag is cached as `false` and the second event is served from cache with no additional HTTP call (existing caching behavior preserved).

`cargo test -p buzz-acp --lib` passes; `just ci` run before PR.

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -148,7 +148,11 @@ fn resolve_agent_owner(config: &Config) -> Option<String> {
 /// Cache for the agent's owner pubkey + sibling lookups.
 ///
 /// Siblings are other agents whose NIP-OA auth tag proves the same owner.
-/// Lookup results are cached for the process lifetime (attestations are immutable).
+/// Only *definitive* lookup results are cached for the process lifetime
+/// (attestations are immutable): a verified sibling, or a fetched profile
+/// that carries no valid matching auth tag. A transport/parse failure
+/// attests nothing, so it is never cached — the next event retries the
+/// lookup instead of pinning a transient error as a permanent negative.
 struct OwnerCache {
     pubkey: Option<String>,
     /// author_hex → is_sibling (true = same owner, false = not)
@@ -188,7 +192,8 @@ impl OwnerCache {
 /// Check if `author` is the owner OR a sibling (same owner via NIP-OA).
 ///
 /// For unknown authors, queries their kind:0 profile to extract the NIP-OA
-/// auth tag and verify the owner matches. Result is cached.
+/// auth tag and verify the owner matches. Definitive results are cached;
+/// transient failures are not — the next event retries the lookup.
 async fn is_owner_or_sibling(
     author: &str,
     owner_cache: &OwnerCache,
@@ -210,9 +215,25 @@ async fn is_owner_or_sibling(
     }
 
     // Query the author's kind:0 profile to check for NIP-OA auth tag.
-    let is_sibling = check_sibling_via_profile(author, my_owner, rest_client).await;
-    owner_cache.cache_sibling(author.to_string(), is_sibling);
-    is_sibling
+    match check_sibling_via_profile(author, my_owner, rest_client).await {
+        SiblingCheck::Sibling => {
+            owner_cache.cache_sibling(author.to_string(), true);
+            true
+        }
+        SiblingCheck::NotSibling => {
+            owner_cache.cache_sibling(author.to_string(), false);
+            false
+        }
+        SiblingCheck::Indeterminate => {
+            // Fail closed for THIS event, but do NOT cache: the failure
+            // attested nothing, so the next event retries the lookup.
+            tracing::warn!(
+                author,
+                "sibling lookup indeterminate (transport/parse failure) — failing closed without caching; next event retries"
+            );
+            false
+        }
+    }
 }
 
 /// Inbound author gate decision: does this author's event fire a turn?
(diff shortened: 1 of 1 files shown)


yjc801:
Thanks for diagnosing this and for the tri-state verdict design — I hit the same bug on a fleet last week (agents answering their owner but silently dropping every agent-to-agent mention after a relay REST blip).

This PR has been unreviewed since August and is now conflicting with `main`, so I've opened #7216, which carries the same first half (`Indeterminate` is not cached; drop log raised above debug) rebased on current `main`, plus two things that came out of review on our side: a bounded per-author retry deadline (5 s doubling to 60 s) instead of re-running the inline 2 s lookup on every event, and an eviction policy that no longer clears the whole 256-entry cache at capacity. It credits this PR and #5450 in the description. Happy to fold anything you'd rather keep from here.

PR #5464, opened 2026-08-10
fix(desktop): resolve message text direction per block for RTL languages

Dror-Bengal:
## Problem

Hebrew (and Arabic/Farsi) messages render fully LTR in the desktop app: RTL paragraphs are left-aligned, and mixed-direction lines — Hebrew text containing English terms, ticket IDs, or numbers — display visually scrambled. The same happens while typing in the composer, where the caret and punctuation behave as if the text were LTR.

## Fix

Two small changes, both scoped to chat content (no layout mirroring — the app chrome stays LTR):

1. **`globals.css`** — `.message-markdown` block elements (`p`, `li`, headings, `blockquote`) get `unicode-bidi: plaintext` + `text-align: start`. Each rendered block resolves its base direction from its own first strong character: Hebrew paragraphs read right-to-left and align right, while English prose and code blocks are unaffected — even within the same message.

2. **`useRichTextEditor.ts`** — the composer's ProseMirror root gets `dir="auto"`, which gives native caret and punctuation behavior when typing RTL. `unicode-bidi: plaintext` is deliberately **not** applied inside the editor: per-block plaintext fights ProseMirror's caret placement (we hit this in testing — the caret stuck to the wrong edge around punctuation).

## Testing

Running this patch on a self-hosted relay with a Hebrew-speaking team (humans + agents conversing in Hebrew with embedded English technical terms):

- Hebrew paragraphs align right and read correctly; English sentences and code blocks in the same message stay LTR.
- Mixed Hebrew/English lines no longer scramble.
- Typing Hebrew in the composer: caret and punctuation behave natively; starting a line with English keeps LTR behavior.
- No visible change for pure-LTR content.

### DIFF
--- a/desktop/src/features/messages/lib/useRichTextEditor.ts
+++ b/desktop/src/features/messages/lib/useRichTextEditor.ts
@@ -7,6 +7,7 @@ import Placeholder from "@tiptap/extension-placeholder";
 import Link from "@tiptap/extension-link";
 import { Extension, type KeyboardShortcutCommand } from "@tiptap/core";
 import { Plugin, Selection, TextSelection } from "@tiptap/pm/state";
+import { Decoration, DecorationSet } from "@tiptap/pm/view";
 import type { ResolvedPos } from "@tiptap/pm/model";
 
 import { readTextFromSystemClipboard } from "@/shared/api/tauriMedia";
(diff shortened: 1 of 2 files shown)


Dror-Bengal:
Pushed a follow-up commit after more field testing on macOS: `dir="auto"` on the ProseMirror root turned out not to be reliably honored by WKWebView inside contenteditable — the caret and trailing punctuation still behaved as LTR while typing Hebrew.

The new commit adds a small `BidiBlockDirection` extension that decorates every top-level block with an explicit `dir` resolved from the block's own first strong character (recomputed on doc changes). Explicit direction gives the WebView unambiguous caret/punctuation behavior, and each paragraph is independent — a Hebrew line followed by an English code line each behave natively before sending. Verified working on macOS (Apple Silicon, WKWebView).

PR #5474, opened 2026-08-10
docs(relay): drop stale fix-pending notes from the tenant.rs RED gate comments

crypticseeds:
## What changed

Remove the fix-pending notes from the two `redteam_attack2` doc comments in `crates/buzz-relay/src/tenant.rs`: the pre-fix return value, the prescribed fix, and the two instructions to delete an `#[ignore]`.

## Why

`bind_community` short-circuits when the normalized host is empty and returns `BindError::UnmappedHost` before any resolver lookup. The comments still describe that guard as outstanding work, stating that binding an empty `raw_host` currently succeeds onto a misconfigured empty-host row, prescribing the short-circuit as the fix, and telling the reader to delete an `#[ignore]` that neither test carries. The first also points at `cargo test --include-ignored`, which selects nothing here.

Read at face value the module reports the host-binding fence as broken, which is the kind of claim someone auditing tenant isolation acts on.

## Impact

Documentation only. The comments now match the behaviour the tests cover.

## Validation

- `cargo test -p buzz-relay --lib tenant::` reports 10 passed, 0 failed, 0 ignored
- `cargo clippy -p buzz-relay --lib --all-targets` reports no warnings
- `cargo fmt --all -- --check` is clean
- Run against rustc 1.95.0 via Hermit, matching `rust-toolchain.toml`

### DIFF
--- a/crates/buzz-relay/src/tenant.rs
+++ b/crates/buzz-relay/src/tenant.rs
@@ -263,21 +263,12 @@ mod tests {
         /// RED gate. Configures a resolver with an `""→CommunityId` mapping
         /// (the schema permits it; no CHECK against empty host exists), then
         /// asks `bind_community` to bind an empty raw_host as a request with
-        /// a missing/invalid Host header would. Today this returns
-        /// `Ok(TenantContext{community=X})` — the fence collapses to the
-        /// misconfigured row. The fix: short-circuit in `bind_community` so
-        /// that `normalize_host(raw_host).is_empty()` returns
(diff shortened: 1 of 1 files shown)


crypticseeds:
Gentle ping. This is a comment-only change (1 insertion, 13 deletions) to two doc comments in `crates/buzz-relay/src/tenant.rs` that describe an already-shipped guard as outstanding work. The stale text is still on main and the branch applies cleanly. Happy to close if it's not worth the review time.

PR #5483, opened 2026-08-10
fix(desktop): enrich relay agent directory for @mention eligibility (#5363)

### LINKED ISSUE #5363: Relay-discovered agent with respond_to=anyone still never appears in @-mention autocomplete (regression/incomplete case of #4913)
### Summary

After PR #4913 ("fix(desktop): allow shared agent mentions", merged into `main` 2026-08-05) was expected to let non-owners `@`-mention a relay-discovered agent when its `respond_to` policy authorizes them, the agent still never appears in the `@`-mention autocomplete — even when every precondition the fix relies on is independently confirmed correct on the relay.

This reproduces with **two independent agents, two different owning accounts, and two different channels**, so it isn't specific to one agent's config.

### Environment

- Buzz Desktop: 0.5.8 (macOS), built from a commit including #4913
- Relay: self-hosted, upgraded to current `main` (`5bf78671f`) specifically to rule out a version gap — no change
- Both the mentioning user and the agent owner are members of the same community

### Repro steps

1. Owner A creates/edits an agent, sets "Who can talk to this agent" to **Anyone** via **channel members sidebar → agent card → "Manage agent access"** (not the top-level Edit Agent dialog, which is the known #2501 persistence bug).
2. Confirmed via the relay's own event log that this actually published a fresh `kind:30177` event with `"respond_to":"anyone"` (superseding the prior `owner-only` event) — see raw events below.
3. Agent is added as a member (`role: bot`) of a `stream`-type channel.
4. A different user (not the agent's owner), already a genuine member of that same channel, opens the channel and types `@` + the agent's name.
5. **Expected**: agent appears in the autocomplete per #4913's stated behavior ("admit relay-discovered agents to autocomplete when their response policy authorizes the viewer... require authorization in the exact active stream/forum channel for mentions").
6. **Actual**: agent never appears, regardless of search string used (tried exact name substrings that should match, e.g. typing "ben" against an agent named "Bens Fizz").

### Confirmed NOT the cause (ruled out one at a time)

- **Relay code version** — upgraded relay from a pre-#4913 commit to current `main`; no change.
- **`respond_to` not actually persisting** — verified directly against the relay's Postgres `events` table (not the Desktop UI) that a `kind:30177` event with `respond_to: "anyone"` was published and is the latest event for the agent's `d`-tag.
- **Channel membership** — verified via `channel_members` that the agent has an active (`removed_at IS NULL`) `bot` row in the exact channel, and the mentioning user has an active `member` row in the same channel.
- **Channel type** — verified `channel_type = 'stream'` (the type #4913 explicitly scopes exact-channel mention authorization to).
- **Relay connectivity** — confirmed active websocket connections from the mentioning user's client in the relay's own connection logs at the time of testing.
- **Relay-side access control** — read `get_accessible_channel_ids` directly (`buzz-db/src/channel.rs`): grants access via real membership OR open-visibility fallback; both users have real membership rows, so the relay has no basis to withhold channel-scoped events.
- **Stale local client cache** — fully quit the app, cleared `~/Library/Caches/xyz.block.buzz.app` and `~/Library/WebKit/xyz.block.buzz.app` (WebView cache/IndexedDB/localStorage), relaunched. Same result. (Note: this cache also apparently held session/identity data not backed by the Keychain-stored secrets — recovered via the existing-key flow, but worth flagging separately: clearing WebView storage shouldn't be able to desync a user from their Keychain-stored identity in the first place.)

### Raw evidence

Relay `events` table, latest two `kind:30177` rows for one of the test agents (pubkey redacted per-agent, available on request):
```
{"name":"Grokola","persona_id":"7f87a3ac-aaa7-4c59-8d44-032846410fa1","parallelism":10,"respond_to":"anyone"}      | 2026-08-08 23:47:24+00
{"name":"Grokola","persona_id":"7f87a3ac-aaa7-4c59-8d44-032846410fa1","parallelism":10,"respond_to":"owner-only"} | 2026-08-08 09:15:31+00
```

`channel_members` for the same channel:
```
channel_type = stream
<agent pubkey>  role=bot     removed_at=NULL
<mentioner pubkey> role=member removed_at=NULL
```

### Suspected area

Given every server-side precondition is verifiably correct, this points at the client-side eligibility computation in `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts` / `desktop/src/features/messages/lib/useMentions.ts` (added/modified in #4913) — possibly in how `relayAgentsQuery` data is fetched/kept fresh, or in `getMentionableAgentPubkeys`'s channel-scoped branch (`relayAgentCanRespondInChannel`). I wasn't able to get a live JS-console trace to pin down the exact line (release build has no reachable WebView inspector — Safari's Develop menu shows "No inspectable applications," and no in-app devtools shortcut/setting was found either), so filing with full server-side evidence instead in case someone with build access can reproduce with `pnpm --dir desktop dev` and step through it directly.

### Related

- #2950 (original picker-visibility report this traces back to)
- #2501 (respond_to persistence bug — ruled out here since we verified the published event directly, but worth keeping in view since it's the same feature area)
- #4913 (the fix this appears to be a regression in / incomplete case of)

rmichelena:
Fixes #5363.

Rebased onto current `main` and narrowed to the directory layer — see "What changed in this revision" at the bottom.

## Root cause: the directory is built from half the data

`list_relay_agents` (`desktop/src-tauri/src/commands/agent_discovery.rs`) builds the relay-agent directory from **`kind:10100` only**. Two other kinds carry the data eligibility actually needs:

- **`kind:30177`** (`KIND_MANAGED_AGENT`, `d`-tag = agent pubkey) — the `respond_to` policy, published by the *owner's* Desktop.
- **`kind:39002`** (NIP-29 members list, one event per channel) — channel membership.

`agents_from_events` does parse `respond_to` / `channel_ids` when they happen to be embedded in the 10100 content — but an agent whose policy and membership live in the dedicated events arrives with `respond_to: null` and `channel_ids: []`.

Downstream, `relayAgentIsSharedWithUser` then returns false for it, so the agent never enters `mentionableAgentPubkeys` and is hidden — even for a channel member, and even for the agent's own owner. Owner-affected reproduces, which rules out the authorization branch as the sole cause.

This is why the symptom survives #4913: that fix (and the other open attempts) corrects the *eligibility computation*, but the directory feeding it is missing the policy and membership data in the first place.

## Neither kind is "the" directory — it's the union

Seeding from `kind:10100` alone is half correct, and the mirror-image of it is also half correct:

| Agent | 10100-seeded directory | 30177-seeded directory | union (this PR) |
|---|---|---|---|
| self-published 10100, no 30177 | ✅ listed | ❌ **dropped** | ✅ |
| Desktop-managed 30177, never self-published | ❌ **dropped** | ✅ listed | ✅ |
| both | ✅ | ✅ | ✅ |

Both shapes exist in the wild:

- **10100-only** — headless `buzz-acp` seats publish their own `kind:10100` (via `buzz channels set-add-policy`, operator scripts, or hand-rolled with `nak`). #2987's reporter does exactly this, and so does the 0.5.9 follow-up on that thread. These agents have no Desktop, so nothing publishes a `kind:30177` for them.
- **30177-only** — a Desktop-managed agent that never self-published a `kind:10100` (the common case, since nothing in Desktop publishes one).

So `list_relay_agents` now reads **both**, keyed by agent pubkey, and then enriches with `kind:39002`.

## Change

All of it in `desktop/src-tauri/src/commands/agent_discovery/relay_enrich.rs` plus the type/mapper additions it needs:

- Seeds the directory from the union of `kind:10100` and `kind:30177`, merging `kind:30177` policy and `kind:39002` (`#p` → `channel_ids`).
- **Authorship is verified, and that matters more now.** The `kind:30177` discovery query is not scoped by the `kind:10100` pubkeys (that filter is exactly what hid 30177-only agents), so it reads records whose `d`-tag can name *any* pubkey. Every record is checked against the agent's own NIP-OA owner (`kind:0`) before it can seed or override anything — a third party's `kind:30177` claiming someone else's agent is dropped, with a test pinning that.
- **Two `kind:30177` reads, in parallel.** A `#d`-scoped one keeps policy enrichment for agents already in the directory guaranteed regardless of community size; the unfiltered one only ever *adds* entries, and logs a warning if it saturates the relay page rather than silently shortening the directory.
- **Exposes the resolved owner.** `owner_pubkey` (`ownerPubkey` in TS) is populated from the owner lookup the enrichment already performs, so an owner-aware admission branch has it without a second `kind:0` round trip. `None`/`null` means **unresolved, not unowned** — on a closed relay the NIP-OA owner frequently never materializes (#4223, relay-side fix in #5581), and headless agents are exactly the population that runs there, so a consumer must treat absence as "unknown". Nothing in this PR reads the field.
- **Best-effort:** a relay error during any enrichment step degrades to what is already known rather than emptying autocomplete.

Rationale for merging rather than requiring a fatter `kind:10100`: `39002` is the relay-authoritative NIP-29 membership (stronger than a self-reported `channel_ids`) and `30177` is the live policy event, so both are better sources than a snapshot baked into the profile.

On directory size and exposure: the unfiltered read makes `relayAgents` larger, since a `kind:30177` exists for every Desktop-managed agent while a self-published `kind:10100` is comparatively rare. That changes volume, not visibility — eligibility still requires a shared channel (or an allowlist entry naming the viewer), so a stranger's agent with no channel in common is listed in the directory and admitted nowhere, exactly as a stranger's `kind:10100` is today.

## What changed in this revision

Dropped the optional TS hardening (the `useComposerMentionableAgentPubkeys` extraction and the `channelMemberAgentPubkeys` fallback in `agentAutocompleteEligibility`). `main` has since reworked `useMentions.ts` — `getAdmittedAgentPubkeys`, `directoryReady`, the `ownerOnly` build-policy gate — and the extraction conflicted with it. That work is better done on top of what's there now, and #5484 / #5612 are already in that area. This PR is now the directory layer only and applies to current `main` with no conflicts.

## Relationship to the other open PRs on this symptom

There are a lot of these. I went through the open PRs whose titles touch mentions / agents / autocomplete / `10100` and read each one's file list. Of those, **only #4716 and this PR change `agent_discovery.rs`** — the rest patch the eligibility computation without changing the directory that feeds it. Offered as triage help, not as a claim that the TS-side PRs are wrong; several fix real second-order bugs.

- **#4716** — the closest relative, and the reason for the union framing. It repoints `list_relay_agents` from `kind:10100` to `kind:30177` on the premise that *"nothing in the codebase publishes `kind:10100` as an agent profile."* True of Desktop's own code, but not of deployments: #2987 is a report from operators who publish it themselves, and #5546 / #5691 are open PRs to have the harness publish it too. Repointing would drop the headless seats that #2987 is about. Its `39002` membership resolution and its instance-tier vs definition-tier point are both good and independent of which kind seeds the list.
- **#5484** — complementary: admits an agent whose NIP-OA owner is the viewer, overriding response policy for channel mention routing. That covers `respond_to: "owner-only"`, which the directory layer cannot fix alone. The `owner_pubkey` this PR exposes is meant for exactly that branch.
- **#5612** — complementary, write side: coordinates the multiple `kind:10100` writers with a generation CAS. Worth reading alongside, because it also proposes removing `respond_to_allowlist` from the published record — if that lands, `relayAgentIsSharedWithUser`'s allowlist branch needs the fallback that PR adds, since an `allowlist`-mode record with no allowlist currently evaluates to "hidden from everyone, including the owner."

## Scope — what this does not fix

- `respond_to: "owner-only"` agents remain hidden from their own owner's autocomplete. That needs owner identity *consumed* in the eligibility path (#5484); this PR only supplies it.
- An agent with **no** `kind:10100` **and** no `kind:30177` is still invisible; there is nothing on the relay to read.

### DIFF
--- a/desktop/src-tauri/src/commands/agent_discovery.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery.rs
@@ -7,7 +7,6 @@ use crate::{
         DiscoverManagedAgentPrereqsRequest, InstallRuntimeResult, ManagedAgentPrereqsInfo,
         RelayAgentInfo, DEFAULT_ACP_COMMAND,
     },
-    nostr_convert,
     relay::query_relay,
 };
 
@@ -1000,6 +999,7 @@ fn build_install_command(command: &str) -> Result<std::process::Command, String>
 mod install_capture;
 mod install_exec;
 mod install_report;
+mod relay_enrich;
 use install_exec::run_install_command_with_retry;
 use install_report::InstallReporter;
 
@@ -1048,14 +1048,7 @@ pub async fn list_relay_agents(state: State<'_, AppState>) -> Result<Vec<RelayAg
     )
     .await?;
 
-    // The convert helper returns `{"agents": [...]}`. Extract and re-deserialize
-    // into the strongly-typed `Vec<RelayAgentInfo>` the frontend expects.
-    let value = nostr_convert::agents_from_events(&events);
-    let agents = value
-        .get("agents")
-        .cloned()
-        .unwrap_or_else(|| serde_json::json!([]));
-    serde_json::from_value(agents).map_err(|e| format!("agent parse failed: {e}"))
+    relay_enrich::list_relay_agents_enriched(state.inner(), events).await
 }
 
 #[cfg(test)]
--- a/desktop/src-tauri/src/commands/agent_discovery/relay_enrich.rs
+++ b/desktop/src-tauri/src/commands/agent_discovery/relay_enrich.rs
@@ -1,0 +1,737 @@
+//! Build the relay agent directory from the union of kind:10100 and kind:30177,
+//! then enrich it with kind:30177 policy and kind:39002 channel membership
+//! (#4913 / #5363).
+//!
+//! Neither kind alone is the directory. `kind:10100` is self-published by the
+//! agent (headless `buzz-acp` seats, `buzz channels set-add-policy`, hand-rolled
+//! records); `kind:30177` is published by the *owner's* Desktop for the agents it
+//! manages. An agent can have either without the other, so seeding from just one
+//! silently drops a whole deployment shape: 10100-only for headless seats,
+//! 30177-only for a Desktop-managed agent that never self-published.
+
+use std::collections::{HashMap, HashSet};
+
+use buzz_core_pkg::kind::KIND_MANAGED_AGENT;
+
+use crate::{
+    app_state::AppState,
+    managed_agents::{agent_events::managed_agent_content_from_event, RelayAgentInfo},
+    nostr_convert,
+    relay::query_relay,
+};
+
+/// Relay HTTP `/query` default page size when `limit` is omitted.
+const RELAY_DEFAULT_QUERY_LIMIT: usize = 100;
+/// Relay hard cap for explicit `limit` (`DEFAULT_MAX_PAGE_LIMIT` in buzz-db).
+const RELAY_MAX_QUERY_LIMIT: usize = 1000;
+
+/// kind:39002 returns one members-list event per channel, not per agent.
+///
+/// `seeded_agent_count` is the number of candidates discovered from kind:30177
+/// alone. That kind carries no channel list at all, so their cardinality is
+/// unknown and the query has to ask for the max page.
+fn channel_membership_query_limit(agents: &[RelayAgentInfo], seeded_agent_count: usize) -> usize {
+    let sparse_10100 = agents.iter().all(|agent| agent.channel_ids.is_empty());
+    if sparse_10100 || seeded_agent_count > 0 {
+        // kind:10100 often omits channel_ids; cardinality is unknown — request max page.
+        return RELAY_MAX_QUERY_LIMIT;
+    }
+
+    // Past this point every agent came from kind:10100 with a known channel list.
+    let agent_count = agents.len();
+
+    let mut channel_ids = HashSet::new();
+    for agent in agents {
+        channel_ids.extend(agent.channel_ids.iter().cloned());
+    }
+    // One 39002 event per channel where any agent is a member.
+    let channel_cardinality = channel_ids.len().max(agent_count);
+    channel_cardinality
+        .saturating_mul(2)
+        .clamp(RELAY_DEFAULT_QUERY_LIMIT, RELAY_MAX_QUERY_LIMIT)
+}
+
+/// kind:30177 is replaceable by (author, kind, d): several events can share a d-tag.
+fn managed_agent_definition_query_limit(agent_count: usize) -> usize {
+    agent_count
+        .saturating_mul(4)
+        .clamp(RELAY_DEFAULT_QUERY_LIMIT, RELAY_MAX_QUERY_LIMIT)
+}
+
+fn d_tag_from_event(event: &nostr::Event) -> Option<String> {
+    event.tags.iter().find_map(|tag| {
+        let slice = tag.as_slice();
+        if slice.first().map(String::as_str) == Some("d") {
+            slice.get(1).filter(|value| !value.is_empty()).cloned()
+        } else {
+            None
+        }
+    })
+}
+
+fn channel_ids_by_agent_from_membership_events(
+    events: &[nostr::Event],
+    agent_pubkeys: &HashSet<String>,
+) -> HashMap<String, Vec<String>> {
+    let mut by_agent: HashMap<String, HashSet<String>> = HashMap::new();
+    for event in events {
+        let Some(channel_id) = d_tag_from_event(event) else {
+            continue;
+        };
+        for tag in event.tags.iter() {
+            let slice = tag.as_slice();
+            if slice.first().map(String::as_str) != Some("p") {
+                continue;
+            }
+            let Some(agent_pubkey) = slice.get(1).filter(|value| !value.is_empty()) else {
+                continue;
+            };
+            if agent_pubkeys.contains(agent_pubkey) {
+                by_agent
+                    .entry(agent_pubkey.clone())
+                    .or_default()
+                    .insert(channel_id.clone());
+            }
+        }
+    }
+
+    let mut sorted: HashMap<String, Vec<String>> = HashMap::with_capacity(by_agent.len());
+    for (agent_pubkey, channel_ids) in by_agent {
+        let mut channel_ids: Vec<String> = channel_ids.into_iter().collect();
+        channel_ids.sort();
+        sorted.insert(agent_pubkey, channel_ids);
+    }
+    sorted
+}
+
+/// Owner-verified kind:30177 projection used to enrich — and, for agents that
+/// never published a kind:10100, to seed — a directory entry.
+#[derive(Debug, Clone)]
+struct ManagedAgentDefinition {
+    name: String,
+    respond_to: crate::managed_agents::RespondTo,
+    respond_to_allowlist: Vec<String>,
+}
+
+fn collect_managed_agent_definitions(
+    events: &[nostr::Event],
+    expected_owners: &HashMap<String, String>,
+) -> HashMap<String, ManagedAgentDefinition> {
+    let mut definitions: HashMap<String, (ManagedAgentDefinition, u64, String)> = HashMap::new();
+    for event in events {
+        let Some(agent_pubkey) = d_tag_from_event(event) else {
+            tracing::warn!("list_relay_agents: skipping kind:30177 event without d-tag");
+            continue;
+        };
+        let event_author = event.pubkey.to_hex();
+        let Some(expected_owner) = expected_owners.get(&agent_pubkey) else {
+            continue;
+        };
+        if event_author != *expected_owner {
+            continue;
+        }
+        let Ok(content) = managed_agent_content_from_event(event) else {
+            tracing::warn!(
+                agent_pubkey = %agent_pubkey,
+                "list_relay_agents: skipping unparsable kind:30177 content"
+            );
+            continue;
+        };
+        let created_at = event.created_at.as_secs();
+        let event_id = event.id.to_hex();
+        match definitions.get(&agent_pubkey) {
+            Some((_, existing_created_at, existing_event_id))
+                if created_at < *existing_created_at
+                    || (created_at == *existing_created_at && event_id <= *existing_event_id) =>
+            {
+                continue;
+            }
+            _ => {
+                definitions.insert(
+                    agent_pubkey,
+                    (
+                        ManagedAgentDefinition {
+                            name: content.name,
+                            respond_to: content.respond_to,
+                            respond_to_allowlist: content.respond_to_allowlist,
+                        },
+                        created_at,
+                        event_id,
+                    ),
+                );
+            }
+        }
+    }
+    definitions
+        .into_iter()
+        .map(|(agent_pubkey, (definition, _, _))| (agent_pubkey, definition))
+        .collect()
+}
+
+/// Discover every kind:30177 record on the relay.
+///
+/// Unlike the previous `#d`-scoped query, this is not filtered by the kind:10100
+/// pubkeys — that filter is exactly what made a Desktop-managed agent with no
+/// self-published kind:10100 invisible. Authorship is still verified downstream
+/// in [`collect_managed_agent_definitions`], which drops any record whose author
+/// is not the agent's own NIP-OA owner, so an unfiltered read cannot let a third
+/// party inject a directory entry.
+async fn fetch_managed_agent_events(state: &AppState) -> Vec<nostr::Event> {
+    let filter = serde_json::json!({
+        "kinds": [KIND_MANAGED_AGENT],
+        "limit": RELAY_MAX_QUERY_LIMIT,
+    });
+
+    match query_relay(state, &[filter]).await {
+        Ok(events) => {
+            if events.len() >= RELAY_MAX_QUERY_LIMIT {
+                tracing::warn!(
+                    returned = events.len(),
+                    "list_relay_agents: kind:30177 discovery hit the relay page cap; \
+                     agents beyond the page are seeded only if they published a kind:10100"
+                );
+            }
+            events
+        }
+        Err(error) => {
+            tracing::warn!(
+                error = %error,
+                "list_relay_agents: kind:30177 discovery failed; continuing with kind:10100 only"
+            );
+            Vec::new()
+        }
+    }
+}
+
+/// Fetch kind:30177 records for agents already known from kind:10100.
+///
+/// The unfiltered discovery query above shares one page across every agent in
+/// the community, so on a large relay it can truncate before reaching a given
+/// agent. This `#d`-scoped query keeps policy enrichment for the agents already
+/// in the directory guaranteed regardless of community size — discovery only
+/// ever *adds* entries, it can never cost an existing one its policy.
+///
+/// Deliberately not filtered by `authors`: authorship is verified in
+/// [`collect_managed_agent_definitions`], so the filter would be an
+/// optimization, and dropping it lets this run before the owner lookup.
+async fn fetch_managed_agent_events_for_known_agents(
+    state: &AppState,
+    agent_pubkeys: &[String],
+) -> Vec<nostr::Event> {
+    if agent_pubkeys.is_empty() {
+        return Vec::new();
+    }
+
+    let filter = serde_json::json!({
+        "kinds": [KIND_MANAGED_AGENT],
+        "#d": agent_pubkeys,
+        "limit": managed_agent_definition_query_limit(agent_pubkeys.len()),
+    });
+
+    match query_relay(state, &[filter]).await {
+        Ok(events) => events,
+        Err(error) => {
+            tracing::warn!(
+                error = %error,
+                "list_relay_agents: kind:30177 enrich failed; continuing with kind:10100 only"
+            );
+            Vec::new()
+        }
+    }
+}
+
+/// Build a directory entry for an agent known only from kind:30177.
+///
+/// Mirrors the defaults `nostr_convert::agents_from_events` applies to a sparse
+/// kind:10100 record, so a seeded entry is indistinguishable downstream from a
+/// self-published one. `channel_ids` is filled in by the kind:39002 pass.
+fn seed_from_definition(pubkey: &str, definition: &ManagedAgentDefinition) -> RelayAgentInfo {
+    RelayAgentInfo {
+        pubkey: pubkey.to_string(),
+        name: definition.name.clone(),
+        agent_type: "agent".to_string(),
+        channels: Vec::new(),
+        channel_ids: Vec::new(),
+        capabilities: Vec::new(),
+        status: "offline".to_string(),
+        respond_to: Some(definition.respond_to),
+        respond_to_allowlist: definition.respond_to_allowlist.clone(),
+        // Filled by the caller from the same owner lookup that verified this record.
+        owner_pubkey: None,
+    }
+}
+
+async fn fetch_agent_owner_pubkeys(
+    state: &AppState,
+    agent_pubkeys: &[String],
+) -> HashMap<String, String> {
+    if agent_pubkeys.is_empty() {
+        return HashMap::new();
+    }
+
+    match query_relay(
+        state,
+        &[serde_json::json!({
+            "kinds": [0],
+            "authors": agent_pubkeys,
+            "limit": agent_pubkeys.len(),
+        })],
+    )
+    .await
+    {
+        Ok(profile_events) => profile_events
+            .into_iter()
+            .filter_map(|event| {
+                nostr_convert::profile_valid_oa_owner_pubkey(&event)
+                    .map(|owner| (event.pubkey.to_hex(), owner))
+            })
+            .collect(),
+        Err(error) => {
+            tracing::warn!(
+                error = %error,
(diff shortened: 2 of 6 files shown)


rmichelena:
Update — I've expanded the description with four things found after opening, no code change:

1. **The gap closes three surfaces, not one.** Besides the composer, the DM/new-message picker (`useNewMessageRecipients.ts`) and add-member search (`MembersSidebar.tsx`) both use `community` scope, where `relayAgentIsSharedWithUser` deliberately skips the live-membership fallback (`SECURITY:` comment). Those two have no channel context to fall back on, so a TS-only patch can't reach them even in principle — only fixing the directory does. That's also why every operation on our relay-hosted agents had to go through the CLI.
2. **`shouldHideAgentFromMentions` is unreachable at its only call site.** `useMentions.addCandidate` early-returns on `isAgentIdentityInAllowedList` with the *same* set, so all three hide branches — including the intended "member with unknown invocability ⇒ show" (Option B), i.e. the #5363 case — are dead. The existing unit test pins that branch by calling the helper directly, so it's green over behavior the call site cannot produce. Same finding as #2987; confirmed independently here on two machines. Not fixed in this PR (needs a hook-level test; the suite has no React renderer for `useMentions`) — happy to add it here if you'd rather not have a follow-up.
3. **Residual gap, stated honestly.** This PR *enriches* the `kind:10100` directory, it doesn't *seed* it. An agent that never published 10100 at all stays invisible even as an active `role: bot` member. Closing that means treating `respond_to: null` as unknown rather than as a denial — a trust-model call I'd rather have a maintainer make than smuggle in here. Failing tests demonstrating it (plus two `owner-only` guards a fix must keep green) are on [`rmichelena/buzz@bumble/mention-directory-gap-tests`](https://github.com/rmichelena/buzz/tree/bumble/mention-directory-gap-tests).
4. **Per-PR notes on the related open patches**, replacing the blanket "supersedes" list — including a correction: #4546 is *not* compatible with `main` as written, and I've said why on that PR.

Still open questions for maintainers: Rust-only or Rust + the TS fallback commits, and whether the dead-code test belongs here or in a follow-up.

TheSeydiCharyyev:
One case I could not find covered here: the agent's own owner.

`respond_to` defaults to owner-only in the agent runtime — `crates/buzz-acp/src/config.rs` declares `enum RespondTo` with `#[default] OwnerOnly`, and the test at `config.rs:2456` is named `test_respond_to_default_is_owner_only`. So an agent started without an explicit `--respond-to` is owner-only, which makes it the common case rather than an edge one.

The new test `relayAgentCanRespondInChannel: channel-member fallback still fail-closed for owner-only` passes `OTHER_OWNER_PUBKEY`, so it pins the non-owner path. Does the owner themselves get admitted? Reading `getAgentMentionAdmission` on `main`, an owner-only agent never enters `mentionableAgentPubkeys`, and the admission check denies on `!mentionableAgentPubkeys.has(normalized)` before the `ownerPubkey`/`currentPubkey` comparison is reached — so the person the policy names as the only allowed sender cannot mention their own agent.

If that is intentional, ignore this. If not, `ownsAuthorAgent` in `desktop/src/features/profile/lib/identity.ts` already mirrors the relay's `is_agent_owner` gate and resolves ownership from the profile's NIP-OA `ownerPubkey`, which would keep the resolution consistent with the server.

rmichelena:
Thanks — the case is real and worth fixing, and you're right that it's the common one rather than an edge case. Two corrections to the mechanism, though, because they change where the fix belongs.

**The `ownerOnly` parameter you read isn't the agent's `respond_to`.** It's a compile-time build policy. `useMentions.ts:268` passes `ownerOnly: agentAccessOwnerOnlyQuery.data`, which resolves through `getAgentAccessOwnerOnly()` → the `agent_access_owner_only` Tauri command → `owner_only_access_build()`, documented as *"whether this build enforces owner-only managed-agent access"* (`desktop/src-tauri/src/commands/agent_access.rs`). The hook's own doc comment says the value is *"baked at build time, so it cannot change while the app runs."* So that branch isn't unreachable — it fires in owner-only builds, for every agent, regardless of any agent's `respond_to`.

**The exclusion you're describing happens one layer earlier, and unconditionally.** `relayAgentIsSharedWithUser` has exactly two paths that return `true`:

```ts
if (agent.respondTo === "allowlist" && normalizedCurrentPubkey) {
  return agent.respondToAllowlist.map(normalizePubkey).includes(normalizedCurrentPubkey);
}
return (
  agent.respondTo === "anyone" &&
  agent.channelIds.some((channelId) => sharedChannelIds.has(channelId))
);
```

`owner-only` falls through to the final `return` and is always `false`, so the agent never enters `mentionableAgentPubkeys` and `getAgentMentionAdmission` denies at the membership check — for its owner as much as for anyone else. Your conclusion holds; it just doesn't depend on the ordering inside `getAgentMentionAdmission`.

And it can't be fixed there as things stand: neither `RelayAgentInfo` (`desktop/src-tauri/src/managed_agents/types.rs`) nor the TS `RelayAgent` (`desktop/src/shared/api/types.ts`) carries an owner field, so the eligibility path has no way to ask "am I this agent's owner?" That's why `allowlist` + [owner] works today and `owner-only` doesn't — the allowlist ships the pubkey inside the record.

**What this PR can contribute, and what it shouldn't.** `relay_enrich.rs` already resolves each agent's NIP-OA owner from its `kind:0` (`fetch_agent_owner_pubkeys`) — it has to, in order to verify `kind:30177` authorship before that record can seed or override a directory entry. The owner is therefore already in hand and simply not exposed. Adding `owner_pubkey` to `RelayAgentInfo` and its TS counterpart is small, purely additive, and squarely this PR's thesis (the directory is missing data the eligibility layer needs), so I'll add it here.

The eligibility branch that *consumes* it is a different matter: #5484 already implements owner-based admission, sourcing `ownerPubkey` from `kind:0` profiles, which is the same resolution `ownsAuthorAgent` uses and the one you're pointing at. I'd rather hand that PR a directory-side owner than land a second implementation of the same gate here.

**One caveat for whoever writes that branch.** On a closed relay the NIP-OA owner frequently never materializes on the agent's `kind:0` at all (#4223; PR #5581 is the relay-side fix), and headless agents are exactly the population that runs on closed relays. So both the directory-sourced and the profile-sourced owner can legitimately be absent for the agents this would help most. An owner-only admission branch should treat "owner unknown" as today's behavior rather than as admission — `getAgentMentionAdmission` already has the right shape for that with its `"unknown"` verdict.

Operationally in the meantime, `respond_to: "allowlist"` with the owner's hex in `respond_to_allowlist` expresses the same policy and does work. Worth saying explicitly that the array has to be non-empty: the harness treats `allowlist` as `allowlist.contains(author) || is_owner_or_sibling(...)` so the owner is implicit there, but the picker reads `respondToAllowlist.includes(currentPubkey)` literally — an `allowlist` record published with an empty array is hidden from everyone, owner included, with no error.

rmichelena:
Done in `bca676e6` — `owner_pubkey` is now on `RelayAgentInfo` and `ownerPubkey` on the TS `RelayAgent`, populated from the owner lookup `relay_enrich` already performs to verify `kind:30177` authorship. No second round trip, and no eligibility behaviour change in this PR: the field is populated and otherwise unread, so an owner-aware admission branch (#5484, or a follow-up) has the data waiting.

Two things worth knowing for whoever consumes it:

- `None`/`null` means **unresolved, not unowned**, for the reason in my previous comment — closed relays often never materialize the NIP-OA owner on the agent's `kind:0`.
- There's a test pinning that it serializes as `owner_pubkey` and not `ownerPubkey`. The Tauri payload contract is snake_case with `fromRawRelayAgent` doing the mapping, so a `#[serde(rename_all = "camelCase")]` on that struct would land `undefined` on the TS side with nothing failing. Nothing in the repo pinned that before.

Verified locally on the pushed tree: `cargo test --lib` 2287 passed / 0 failed, `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, `tsc --noEmit` clean, desktop `pnpm test` 4540 passed / 0 failed.

rmichelena:
Rebased onto current `main` (`df9e773a`) and narrowed to the directory layer — force-pushed to `f6a1aede`. 20 commits → 1, 10 files → 6, and it now applies with no conflicts.

Dropped the optional TS hardening (the `useComposerMentionableAgentPubkeys` extraction and the `channelMemberAgentPubkeys` fallback). `main` has since reworked `useMentions.ts` — `getAdmittedAgentPubkeys`, `directoryReady`, the `ownerOnly` build-policy gate — and the extraction conflicted with it. Rebuilding that on top of what's there now is a better use of a reviewer's time than merging my earlier version of it, and #5484 / #5612 are already working in that area.

So this PR is now only what nothing else on the tracker touches: `list_relay_agents` reading the union of `kind:10100` and `kind:30177`, enriched with `kind:39002`, with the owner it already resolves exposed for #5484's admission branch.

@TheSeydiCharyyev — the `owner_pubkey` plumbing from my previous reply survived the rebase unchanged; only the TS-side commits went away.

Verified on the pushed tree (`f6a1aede`, against `main@df9e773a`):

```
cargo test --lib                              2432 passed; 0 failed; 15 ignored
cargo clippy --all-targets -- -D warnings     clean
cargo fmt --check                             clean
tsc --noEmit                                  exit 0
```

TheSeydiCharyyev:
You are right on both counts, and my framing was off. The exclusion is unconditional in `relayAgentIsSharedWithUser` — `owner-only` reaches the final `return` and is always false — so it never depends on the order inside `getAgentMentionAdmission`. And the `ownerOnly` parameter there is the build policy, not the agent's `respond_to`.

Your point that an unresolved owner must stay `"unknown"` matters more than it first looks. `agents_from_events` fills absent fields with defaults, so "no owner published" and "owner not resolvable yet" arrive downstream looking identical. Denying on unresolved would hide exactly the closed-relay agents the change is meant to reach.

Separately, the `allowlist` branch has a client/server divergence. It is a different bug with the same effect, and I could not find it on the tracker.

The harness treats the owner as implicit — `crates/buzz-acp/src/lib.rs:255`:

```rust
RespondTo::Allowlist => allowlist.contains(author) || is_owner_or_sibling(author, ...).await
```

The picker reads the array literally — `agentAutocompleteEligibility.ts`:

```ts
if (agent.respondTo === "allowlist" && normalizedCurrentPubkey) {
  return agent.respondToAllowlist
    .map((pubkey) => normalizePubkey(pubkey))
    .includes(normalizedCurrentPubkey);
}
```

So an agent on `allowlist` answers its owner while the picker hides it from that same owner, unless the owner's hex is also in the array. An `allowlist` record published with an empty array is hidden from everyone, owner included, with no error — which is the operational workaround failing in the one configuration people would reach for first.

Once `ownerPubkey` is on `RelayAgent`, that branch can take the same ownership check as the `owner-only` one, with the same `"unknown"` treatment when the owner is unresolved.

I can file it separately if it is not already inside #5484's scope.

rmichelena:
Confirmed, both halves, on today's `main` — and the implicit owner in `allowlist` is deliberate, not incidental.

The doc comment on `author_allowed` states it outright:

> Both `OwnerOnly` and `Allowlist` accept the owner and same-owner siblings; `Allowlist` additionally accepts the explicit external pubkey list.

and startup logging treats owner-based matching under `allowlist` as a first-class feature that can be *missing* — `crates/buzz-acp/src/lib.rs:1865`:

```
respond-to=allowlist but no owner is set — allowlisted pubkeys will still be
accepted, but owner-based matching is unavailable until owner is resolved.
```

So the picker reading the array literally (`agentAutocompleteEligibility.ts:46-50`) is the side that's wrong, not the harness.

One scoping note on blast radius: `getMentionableAgentPubkeys` seeds its set from `managedAgentPubkeys` unconditionally, so a locally managed agent stays mentionable whatever its `respond_to`. The population that actually hits this is agents the viewer owns but does not run under Desktop — headless, relay-only. Same population as this PR's.

## It's inside #5484's scope, so I don't think a separate issue earns its keep

`getChannelOwnedAgentPubkeys` admits any channel-member agent whose profile `ownerPubkey` matches the viewer, and the test loop iterates the modes explicitly:

```js
for (const respondTo of [undefined, null, "owner-only", "allowlist", "anyone", "future-mode"]) {
  // ... `owner admission should allow ${String(respondTo)}`
}
```

with `nobody` asserted separately as still excluded. That is `author_allowed` mirrored faithfully — `Nobody => false` in both its arms. And the `profile?.ownerPubkey &&` guard drops members with no resolvable owner, which is the `"unknown"` treatment rather than admission.

I checked the tracker too: the nearest is #3125, but it's a different failure — a viewer who *is* in the array, dropped by the managed-list gate from #2149. I found nothing filed for the owner-not-in-array case (searched `block/buzz` issues for `allowlist owner` and `respond_to allowlist`). If #5484 stalls, it's worth its own issue; while #5484 is live, a duplicate would just fragment the discussion.

## The part the fix can't close, and it bites this exact population

The two sides don't have symmetric access to the answer.

The harness resolves its owner **locally** — `owner_cache.get()` (`lib.rs:197`) is seeded from the agent's own `BUZZ_AUTH_TAG` / `--agent-owner`, so `author == my_owner` is a string compare with no relay round trip. It is correct on a closed relay, an unreachable relay, a relay that never indexed the agent's `kind:0`.

Desktop has to *discover* the same fact from the agent's published NIP-OA attestation. Both candidate sources — #5484's `profiles[pubkey].ownerPubkey` and this PR's directory `owner_pubkey` — bottom out in that attestation materializing, which on closed relays it frequently does not (#4223; relay-side fix in #5581).

So fixing the `allowlist` branch buys parity **wherever the owner resolves**, and the residue is precisely the closed-relay headless agents the change is meant to reach. Failing closed there is the right call, but it means the operational advice can't yet be "just use `allowlist`" — putting the owner's hex in the array explicitly stays the reliable configuration until #5581 lands. Which is an unsatisfying answer for the workaround, and worth saying plainly rather than discovering in the field.

For completeness on the audit: I checked the DM path for the same divergence and there isn't one. `isAgentMentionChannelType` returns `false` for `dm`, so `mentionChannelId` is `null`, the scope collapses to `managed-only`, and relay agents never enter the DM picker at all — consistent with the harness's `is_dm` hardening.

## Merge-order heads-up

#5484 moves `RelayAgentInfo` out of `managed_agents/types.rs` into a new `relay_agent_info.rs` and retypes `respond_to` to `Option<String>`; this PR adds `owner_pubkey` to it in `types.rs`. Whichever lands second has to carry the other's change across. Mechanical, but the kind of thing that vanishes silently in a conflict resolution — if #5484 goes first I'll rebase onto it and re-run the gates.

rmichelena:
Closing this: `main` now covers the gap this PR was opened for, and covers it with a
stricter candidate source than the one proposed here.

**What landed upstream.** Three maintainer PRs reworked this surface after this PR was
last rebased (`df9e773a`, 2026-08-14):

- #6086 — *enforce shared agent access across devices*: "discover shared managed agents
  from authenticated relay directory records instead of treating channel membership as
  sufficient proof"
- #6182 — *restore release agent mentions*
- #6224 — *bound remote agent mention authorization*

`list_relay_agents` moved out of `commands/agent_discovery.rs` into
`commands/agent_discovery/relay_directory.rs`, and the merge logic now lives in
`nostr_convert/agent_directory.rs`. The integration point this PR patches no longer
exists, which is what the conflict is actually reporting.

**The union is implemented.** `relay_agents_from_directory_events` builds the legacy map
from kind:10100 events and then *inserts* every owner-verified kind:30177 policy,
removing the legacy entry first so the verified policy wins. Both cases this PR was
opened to fix are covered:

- a Desktop-managed agent that never published its own kind:10100 is reached through
  `latest_verified_managed_policies` and appears in the directory;
- a headless agent with kind:10100 and no kind:30177 still appears through
  `relay_agents_from_legacy_events`.

Authorship verification is the same shape argued for here: `verified_agent_owners_from_profiles`
resolves the owner from the agent's own signed kind:0 NIP-OA tag, and a kind:30177 event
is dropped unless its author matches. `owner_pubkey` on `RelayAgentInfo` — added to this
PR in `bca676e6` after @TheSeydiCharyyev's review — is on `main` too.

**And the upstream seed is better than this PR's.** This PR seeded candidates with an
*unfiltered* relay-wide kind:30177 read, which is why it carried the page-size arithmetic
and the truncation warning: on a large community the discovery page can truncate before
reaching a given agent. `list_relay_agents_for_selection` instead seeds from relay-signed
kind:39002 membership, taking only `bot`-role p-tags scoped to the viewer, and queries
10100/kind:0 in bounded batches per candidate. That candidate set is bounded by
construction. Rebasing this PR onto today's `main` would mean re-introducing the weaker
seed to reach the same outcome.

Leaving the analysis in the thread since it may still be useful: the `d`-tag on a
kind:30177 is an unverified claim about someone else's agent, so dropping the `#d` filter
turns the kind:0 authorship check from an optimization into a security boundary — `main`
keeps the `#d` filter *and* the authorship check, which is the safer of the two.

One artifact from this branch has no equivalent on `main`, and I'll offer it separately
rather than hold this PR open for it: a test pinning that `RelayAgentInfo` serializes
`owner_pubkey` in **snake_case**. The Tauri payload contract is snake_case and
`fromRawRelayAgent` (`desktop/src/shared/api/tauri.ts`) hand-maps it to camelCase, so a
`#[serde(rename_all = "camelCase")]` added to that struct would silently leave
`ownerPubkey`, `agentType`, `channelIds`, `respondTo` and `respondToAllowlist` as
`undefined` on the frontend with nothing failing. Nothing in the repo currently fixes that
contract.

Thanks @TheSeydiCharyyev for the review on the owner-only path; that exchange is what put
`owner_pubkey` in the payload, and the field is on `main` now.

PR #3403, opened 2026-07-28
fix(acp,agent): support reasoning_content fallback and turn text publishing for local reasoning models

mainza-ai:
### Summary
This PR fixes core issues when running AI agents via `buzz-acp` and `buzz-agent` with local OpenAI-compatible reasoning models (e.g. 4-bit Gemma models served on OMLX/Ollama endpoints):

1. **Reasoning Content Fallback**: Local reasoning models output generated text inside `reasoning_content` or `reasoning` while returning `content: ""` (without `tool_calls`). Updates `parse_openai` and `parse_responses` in `buzz-agent/src/llm.rs` so `text` falls back to `reasoning` when `content` is empty.
2. **Turn Text Accumulation & Fallback Channel Publishing**: Non-tool-calling models do not invoke `buzz messages send`. `buzz-acp` now accumulates `agent_message_chunk` updates during a turn and automatically posts the cleaned turn text as a signed `kind: 9` message event to the channel thread on `StopReason::EndTurn`.
3. **Response Sanitization**: Adds `clean_agent_text_response` in `pool.rs` to strip pseudo-tool JSON blocks (```json {"reply_to": ...} ```) and special model tokens (`<|tool_call>`, `<|im_end|>`) before channel publishing.

### Testing & Verification
- Tested with local OMLX OpenAI-compatible endpoint (`http://127.0.0.1:8787/v1`) using Gemma-4 reasoning models.
- Verified agent messages stream and render directly in channel threads without leaking raw pseudo-tool JSON.
- Verified Rust workspace compilation and unit test suite pass (`cargo test -p buzz-acp -p buzz-agent`).

PR #5487, opened 2026-08-10
fix(buzz-agent): replay reasoning_content for DeepSeek thinking mode

### LINKED ISSUE #5399: buzz-agent: DeepSeek seats fail every follow-up turn, reasoning_content is not sent back
Any agent I run on the `buzz-agent` runtime with a DeepSeek model works for exactly one exchange. The moment a turn needs a second round-trip it 400s, and it does that every time.

The provider says why:

```
Agent reported error (code -32000): llm: (deepseek-v4-pro) 400 Bad Request:
{"error":{"message":"The `reasoning_content` in the thinking mode must be passed back to the API.",
"type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

In thinking mode DeepSeek requires `reasoning_content` from the assistant message to be included when you continue the conversation. `buzz-agent` doesn't send it back.

## How often

79 times over two days, on two seats I run independently:

- `deepseek-v4-flash`, 37 failures, 2026-08-08 10:16 through 2026-08-09 12:45
- `deepseek-v4-pro`, 42 failures, 2026-08-08 10:34 through 12:47

Not flaky. Every continuation fails.

## Setup

Buzz 0.5.5, macOS 26.5.2, Apple Silicon. `buzz-agent` runtime, `openai-compatible` provider, `BUZZ_AGENT_THINKING_EFFORT` set.

## What I ruled out first

Authentication failures are a separate code, `-32001 llm auth: Authentication Fails`, and they stopped once my credentials were right while these carried on. The two seats have independent config. `-pro` and `-flash` behave identically, so it isn't one model or one bad setup.

## Why it matters

A seat that fails on continuation can't hold any role that takes more than one exchange, so review and anything iterative are out. Retries bill per token, so every failure costs money and produces nothing.

One note if you reproduce this from the logs: the lines carry ANSI escapes inside the key=value pairs, so `grep 'outcome="error"'` returns zero on a log full of errors. Strip them first.

```bash
sed $'s/\x1b\[[0-9;]*m//g' <log> | grep 'reasoning_content'
```

olympusbuildz:
## Problem

DeepSeek seats on `buzz-agent` with thinking mode fail on every continuation:

```
400 Bad Request: The `reasoning_content` in the thinking mode must be passed back to the API.
```

Reporter saw **79** hard failures across deepseek-v4-flash/pro in two days (#5399). Iterative tool-using roles break; retries burn tokens for nothing.

## Root cause

| Step | What happens on `main` |
|------|------------------------|
| Parse | `parse_openai` correctly reads `message.reasoning_content` → `LlmResponse.reasoning` |
| UI/ACP | Thought chunks emit that string |
| History | `HistoryItem::Assistant` only stored OpenRouter `reasoning_details` — **plain `reasoning` dropped** |
| Replay | `openai_body` never wrote `reasoning_content` back |

DeepSeek docs ([thinking mode](https://api-docs.deepseek.com/guides/thinking_mode)): for requests that carry `tools`, prior assistant `reasoning_content` must be fully passed back or the API 400s.

## Fix

1. Add `reasoning: String` on `HistoryItem::Assistant` (keep `reasoning_details`).
2. Persist `response.reasoning` at every assistant history push.
3. In `openai_body`, emit `reasoning_content` when non-empty.
4. Charge `reasoning.len()` in estimated bytes.

Single concern — no image_url gating, no provider feature work.

## Why not #4773

[block/buzz#4773](https://github.com/block/buzz/pull/4773) targets the same class but is stale vs current `HistoryItem` (`reasoning_details` era), bundles `supports_image_url`, emits only when that message has tool_calls, and has no unit tests. This PR is main-current, #5399-only, tested.

Closest PR search: #4773 (related), #3403 / #4332 (different). **none found** that closes #5399 on current main with tests.

## Test plan

```bash
cargo test -p buzz-agent --lib reasoning_content
cargo test -p buzz-agent --lib
```

- [x] `openai_body_omits_reasoning_content_when_empty`
- [x] `parse_openai_captures_reasoning_content`
- [x] `openai_body_replays_reasoning_content_on_plain_assistant_when_present`
- [x] `openai_body_replays_reasoning_content_on_tool_call_assistant`
- [x] Full `buzz-agent` lib suite **484 passed** @ head `49fc467` (rebased on current `main`)
- [ ] Full monorepo `just ci` not run for this change (package lib gate only)

## Risk / blast radius

- **In scope:** `buzz-agent` history + OpenAI-compat request body only
- **Not changed:** provider feature flags, image_url gating, non-OpenAI transports

Fixes #5399

Retested after rebase onto current main @ `49fc46796`.

### Acceptance retest (2026-08-18)
Rebased onto current `main`. Mini proof @ `7dd78e177`: cargo test -p buzz-agent --lib reasoning_content — 4/4.

## Mini retest (2026-08-24 acceptance)
- Head `fe4c1c77d7` · cargo test -p buzz-agent --lib reasoning_content → 4 passed
- Rebased onto current upstream `main` (force-with-lease, pre-review)

### DIFF
--- a/crates/buzz-agent/src/agent.rs
+++ b/crates/buzz-agent/src/agent.rs
@@ -669,6 +669,7 @@ impl RunCtx<'_> {
                 self.history.push(HistoryItem::Assistant {
                     text: response.text,
                     tool_calls: Vec::new(),
+                    reasoning: response.reasoning,
                     reasoning_details: response.reasoning_details,
                 });
                 if max_tokens_recoveries >= self.cfg.max_token_recoveries {
@@ -699,6 +700,7 @@ impl RunCtx<'_> {
                 self.history.push(HistoryItem::Assistant {
                     text: response.text,
                     tool_calls: Vec::new(),
+                    reasoning: response.reasoning.clone(),
                     reasoning_details: response.reasoning_details.clone(),
                 });
                 let stop = map_stop(response.stop);
@@ -752,6 +754,7 @@ impl RunCtx<'_> {
             self.history.push(HistoryItem::Assistant {
                 text: response.text,
                 tool_calls: calls.clone(),
+                reasoning: response.reasoning,
                 reasoning_details: response.reasoning_details,
             });
 
@@ -1220,6 +1223,7 @@ pub(crate) fn push_hook_outputs_as_tool_results(
                 // preserve.
                 provider_extra: Default::default(),
             }],
+            reasoning: String::new(),
             reasoning_details: None,
         });
         history.push(HistoryItem::ToolResult(ToolResult {
@@ -1315,6 +1319,7 @@ mod tests {
             history.push(HistoryItem::Assistant {
                 text: format!("a{i} {}", "y".repeat(1000)),
                 tool_calls: vec![],
+                reasoning: String::new(),
                 reasoning_details: None,
             });
         }
@@ -1446,12 +1451,14 @@ mod tests {
             HistoryItem::Assistant {
                 text: "first answer".into(),
                 tool_calls: vec![],
+                reasoning: String::new(),
                 reasoning_details: Some(big_reasoning),
             },
             HistoryItem::User("second question".into()),
             HistoryItem::Assistant {
                 text: "second answer".into(),
                 tool_calls: vec![],
+                reasoning: String::new(),
                 reasoning_details: None,
             },
         ];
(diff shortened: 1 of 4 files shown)


olympusbuildz:
Closing stale open PR with no recent progress to protect olympusbuildz merge reputation. Ping to reopen if maintainers still want this.

PR #5490, opened 2026-08-10
fix(desktop): quiesce renderer polling while hidden (#3677)

### LINKED ISSUE #3677: [Bug] macOS: visible Buzz pages sustain renderer/GPU/WindowServer load and trigger fan
## Summary

On macOS, visible Buzz-related pages produced sustained renderer and compositor load, elevated `WindowServer`, and caused the MacBook's fan to run. A controlled hide/restore test showed the load disappearing almost immediately when Chrome was hidden and returning immediately when Chrome was restored.

Closing the main Buzz page caused a second Buzz-related page (`Block — Introducing Buzz: where humans and agents work together`) to become the active high-CPU renderer. After closing both pages and Buzz Desktop, the implicated renderer processes were gone, Chrome's GPU helper returned to approximately 1%, and the fan stopped.

The strongest direct attribution is to Buzz-related web content in Chrome. Buzz Desktop was open and newly installed, but its own process was not captured using high CPU, so this report does **not** claim the desktop executable alone caused the incident.

## Environment

- Buzz Desktop: **0.5.2**
- Bundle ID: `xyz.block.buzz.app`
- Executable: `buzz-desktop`
- Architecture: native ARM64
- Rendering framework: macOS WebKit
- Signature: Developer ID Application: Block, Inc. (`EYF346PHUG`)
- Buzz app instance: PID 93577
- Buzz WebKit content process: PID 93582
- macOS: **27.0**, build `26A5388g`
- Hardware: MacBook Pro `Mac16,5`, Apple M4 Max
- CPU: 16 cores (12 performance, 4 efficiency)
- GPU: 40 cores
- Memory: 64 GB
- Display: built-in 3456×2234 Retina display; no external displays
- Running Chrome during the incident: **150.0.7871.184**, ARM64

macOS 27.0/build `26A5388g` is a prerelease environment, so a WebKit/Chrome/WindowServer interaction specific to this OS remains possible.

## User-visible impact

- Audible sustained fan
- Rapid power/battery use
- Activity Monitor showed `WindowServer` around 75% CPU
- The fan stopped after both Buzz-related Chrome tabs and Buzz Desktop were closed

macOS process CPU percentages are per core. Therefore, 75% for `WindowServer` means roughly three-quarters of one CPU core, not 75% of the entire 16-core machine.

## Initial measurements

Repeated live samples showed:

| Process | PID | CPU |
|---|---:|---:|
| `WindowServer` | 628 | 74.7–76.2% |
| Chrome GPU helper | 66230 | 39.9–47.5% |
| Buzz Chrome renderer | 47303 | 21.0–32.1% |

Chrome's Task Manager positively identified PID 47303 as:

> Tab: Buzz — Your people, your agents, your project — all in one place.

At capture time it showed:

- CPU: 31.8%
- Memory: 131 MB
- PID: 47303

Chrome's GPU helper had a 1.4 GB physical footprint with a 2.2 GB peak. The Buzz renderer had a 131.4 MB footprint with a 293.4 MB peak.

A native macOS process sample of the GPU helper showed active QuartzCore/CoreAnimation transaction commits and graphics-surface/layer work. That establishes continuous painting/compositing activity, but it does not by itself identify whether the source was CSS animation, `requestAnimationFrame`, canvas, video, WebGL, or another mechanism.

## Controlled hide/restore test

Chrome was hidden for eight seconds without closing or changing its tabs.

| State | WindowServer | Buzz renderer | Chrome GPU |
|---|---:|---:|---:|
| Chrome visible | 89–92% | 27–32% | 42–47% |
| Chrome hidden | 72–79% | 0–0.7% | 0–0.6% |
| Chrome restored | 90–92% | 25–34% | 42–48% |

The renderer and GPU processes reached effectively zero within about one second of hiding Chrome, then returned to their prior levels as soon as Chrome was restored.

This was repeatable within the controlled test and is the strongest causal evidence collected.

## Second Buzz-related page

Chrome's Task Manager also identified:

> Tab: Block — Introducing Buzz: where humans and agents work together.

- PID: 46489
- Memory at initial inspection: 183 MB
- Associated dedicated worker under `block.xyz`

After the main Buzz dashboard tab was closed, PID 47303 disappeared, but PID 46489 then sustained **34.7–41.9% CPU**. During this phase:

- Chrome GPU was approximately 3.9–4.7%
- `WindowServer` was approximately 52.7–59.8%

This suggests the second Buzz-related page also performs sustained work while active, though it appears less GPU-heavy than the main Buzz dashboard.

## After closing all Buzz surfaces

The following processes were absent:

- Buzz Desktop PID 93577
- Buzz WebKit processes 93580 and 93582
- Main Buzz Chrome renderer PID 47303
- Introducing Buzz Chrome renderer PID 46489

Post-closure observations:

- Chrome GPU helper: approximately 0.8–0.9%
- `WindowServer`: approximately 42–55%, with other live diagnostic UI and unrelated system activity still present
- User-reported fan state: off

## Buzz Desktop log observations

Buzz Desktop itself was generally around 0–2% CPU in the available samples. Closing the desktop app was not isolated from closing the remaining web page, so its independent contribution was not measured.

Its unified logs nevertheless contained items worth reviewing:

- `WebContent[93582] Conn 0x0 is not a valid connection ID.` repeated at startup
- Camera/Core Media initialization errors, including an invalid or unavailable background-replacement pixel-buffer size
- Core Media parameter errors
- Core Audio registrar/property failures
- WebKit sandbox denials involving networking and audio services
- Repeated CoreSpotlight donation failures
- Media/audio cleanup errors during shutdown

If the user had not entered a voice huddle, the camera and audio initialization is notable. It may be worth checking whether media infrastructure is initialized eagerly at app startup rather than only when required.

These log messages were not proven to cause the compositor load and may include normal sandbox or prerelease-macOS noise.

## WindowServer log observations

During the investigation, WindowServer also logged invalid-window constraint operations, cursor-surface failures, cleared client buffers, and one `Could not find display 0x0` message.

However, invalid-window messages occurred before, during, and after Buzz was running. They cannot reliably be attributed to Buzz and are included only as environmental context.

## Probable fault class

The evidence is consistent with a Buzz web surface that does not quiesce when visually stable. Candidate causes include:

- An unrestricted `requestAnimationFrame` loop
- A continuously repainted canvas or visualization
- CSS/JavaScript animation running at display refresh rate
- Repeated live-state/component updates without user input
- Animated media or video continuously decoding/compositing
- Missing `visibilitychange`/background-window handling

The controlled test proves that visible Buzz-related content drove substantial renderer/GPU work. It does not identify the exact frontend mechanism.

## Suggested investigation

1. Capture Chrome Performance profiles for the Buzz dashboard and the “Introducing Buzz” page while visible and idle.
2. Inspect main-thread scripting, frame rate, paints/composites, `requestAnimationFrame`, canvas/video/WebGL, and component updates without input.
3. Verify rendering loops pause on `document.visibilitychange`, tab backgrounding, window minimization, offscreen elements, and app deactivation.
4. Profile Buzz's WKWebView build separately with Time Profiler, Core Animation, Energy Log, and Metal System Trace if applicable.
5. Review whether camera, microphone, and Core Media components can be initialized lazily when entering a huddle.
6. Reproduce on both macOS 27 build `26A5388g` and a stable macOS release.

A useful idle acceptance check would be: after 60 seconds without interaction, a visually stable renderer should remain below roughly 2–5% of one core while visible and approximately 0–1% while hidden.

## Reproduction status

The visibility test was deterministic during this incident, but a fresh-launch reproduction has not yet been attempted because the immediate goal was to stop the unexpected load and fan. No application settings or user data were modified.

## Related but apparently distinct issues

- #2868 — idle desktop WebAudio context on Linux; potentially adjacent, but a different measured workload and platform
- #3148 — macOS Claude processes scanning `/`; different process attribution and root cause
- #3328 — general slow UI performance on Linux

Raw native `sample(1)` reports for the Buzz Chrome renderer and Chrome GPU helper were captured locally and can be provided if useful.

wesbillman:
Fixes #3677.

## Problem

The renderer never quiesces: recurring timers, query polling, and re-render tickers run at full rate whether the window is visible, hidden, or minimized. Measured on a live installed app: **27.5% mean renderer CPU visible vs 28.7% hidden** (60×1s `ps` samples of the WebContent process; `sample(1)` dominated by `WebCore::timerFired`/ThreadTimers, microtask checkpoints, JSON parsing, style matching). Matches all three reproductions in #3677 (macOS prerelease, Linux/WebKitGTK A/B/A minimize test, stable macOS).

Per-timer instrumentation (dev build, wrapped `setInterval`/`setTimeout`/rAF) attributed the recurring work: `useNow` 60 fires/min, 40 active TanStack refetch intervals, agent-turn pruning 12/min, auto-restart ticks, huddle/reminder polls — none visibility-gated.

## Fix (two-tier gating, standard mechanisms only)

Two separate signals in `desktop/src/shared/lib/useDocumentVisible.ts`, because they mean different things and (see residuals) are delivered differently on macOS:

- **`useDocumentVisible`** — true Page Visibility only (`document.visibilityState`). Gates local UI work that must keep running on a visible-but-unfocused window: `useNow` relative clocks, agent-turn pruning, huddle bar state/model-status polling, auto-restart tick. Hidden ⇒ paused; `useNow` snaps to fresh `Date.now()` on return.
- **`useAppFocused`** — visible AND `document.hasFocus()`. Gates network refetch polling only (`useFocusedRefetchInterval`, ~15 query families: forum/home/agents/channels/templates/emoji/user-status/projects/workflows/persona-catalog/pulse/presence-list). TanStack's `focusManager` is wired to this signal (idempotent, single install) with `refetchOnWindowFocus: true`, so stale queries refresh promptly on return. Deliberate side effect, documented in code: query retries pause on blur; mutations and the presence heartbeat (`retry: 0`) are unaffected.
- **Never gated:** reminder due-notification poll (fires while hidden/unfocused — extracted to `reminderNotificationPoll.ts` with regression test), huddle pipeline hot-start (`check_pipeline_hotstart` survives backgrounding for the duration of a huddle), relay stall watchdog, presence heartbeat. Live WebSocket delivery untouched throughout.
- Huddle model-status indicator now clears only on huddle phase end, not on visibility/focus changes.

## Validation

- Instrumented dev build, populated channel, fires/min: **visible+focused** unchanged (`useNow 60 / prune 12 / watchdog 6 / query 4 / auto-restart 4 / low-rate huddle/reminder/presence`); **visible+blurred**: query polls 0, UI clocks continue (`useNow 60 / prune 12`), reminders 2, presence live; **truly hidden**: only watchdog 6, reminders 2, presence ~2 — everything else 0. Return restored visible+focused, selection preserved, queries refreshed.
- Hide-vs-blur decomposition (instrumented probe instance, AppleScript-driven): on macOS WKWebView, Cmd-H / minimize / full occlusion did **not** reliably produce `visibilityState === "hidden"` — they reliably produced focus loss. The CPU-dominant quiescence path on macOS is therefore the focus gate; the visibility gate is exercised fully on platforms that report hidden (e.g. WebKitGTK minimize per the Linux repro).
- Gate-regression tests: signal separation, `useNow` hidden-pause + fresh-snap on return, focus-gated interval pause/resume-with-refresh, reminder delivery while hidden+unfocused (5 new, plus primitive wiring tests).
- Push gate: desktop check, typecheck, full desktop suite **4549/4549** at `1237548d1`.

## Known residuals

- **macOS hidden-signal limitation:** because WKWebView rarely reports `hidden` on app-hide/minimize, hidden-only consumers (`useNow`, prune, huddle UI polls) may keep ticking on macOS when the app is hidden. These are cheap local timers; the expensive network polling still quiesces via focus loss, which is what the measured 28% CPU was attributed to. If the residual local-timer cost proves measurable, the follow-up is bridging Tauri window hidden/minimized events into the visibility signal.
- End-to-end CPU confirmation on a packaged build is the post-merge follow-up (against the 28% idle baseline).
- Visible-state costs (skeleton animation pileups on stuck loading views, per-poll JSON payload churn) are intentionally out of scope — separate follow-up issue.

### DIFF
--- a/desktop/src/features/agents/activeAgentTurnsStore.ts
+++ b/desktop/src/features/agents/activeAgentTurnsStore.ts
@@ -6,6 +6,10 @@ import {
   compareObserverEvents,
 } from "@/features/agents/observerRelayStore";
 import { normalizePubkey } from "@/shared/lib/pubkey";
+import {
+  isDocumentVisible,
+  subscribeDocumentVisibility,
+} from "@/shared/lib/useDocumentVisible";
 import type { ObserverEvent } from "./ui/agentSessionTypes";
 
 /** Harness emits turn_liveness every ~10s (BUZZ_ACP_TURN_LIVENESS_SECS). */
@@ -134,6 +138,7 @@ function watermarkChannelKey(event: ObserverEvent): string {
 const terminalAtByAgent = new Map<string, Map<string, number>>();
 
 let pruneInterval: ReturnType<typeof setInterval> | null = null;
+let unsubscribePruneVisibility: (() => void) | null = null;
 
 function invalidateCache(agentKey: string) {
   cachedTurnSummaries.delete(agentKey);
@@ -440,16 +445,31 @@ function processEvent(agentPubkey: string, event: ObserverEvent) {
   }
 }
 
-function ensurePruneInterval() {
-  if (pruneInterval) return;
+function startPruneInterval() {
+  if (pruneInterval || !isDocumentVisible()) return;
+  pruneExpired();
   pruneInterval = setInterval(pruneExpired, PRUNE_INTERVAL_MS);
 }
 
+function pausePruneInterval() {
+  if (!pruneInterval) return;
+  clearInterval(pruneInterval);
+  pruneInterval = null;
+}
+
+function ensurePruneInterval() {
+  if (unsubscribePruneVisibility) return;
+  startPruneInterval();
+  unsubscribePruneVisibility = subscribeDocumentVisibility((visible) => {
+    if (visible) startPruneInterval();
+    else pausePruneInterval();
+  });
+}
+
 function stopPruneInterval() {
-  if (pruneInterval) {
-    clearInterval(pruneInterval);
-    pruneInterval = null;
-  }
+  pausePruneInterval();
+  unsubscribePruneVisibility?.();
+  unsubscribePruneVisibility = null;
 }
 
 export function subscribeActiveAgentTurns(listener: () => void) {
--- a/desktop/src/features/agents/hooks.ts
+++ b/desktop/src/features/agents/hooks.ts
@@ -19,6 +19,10 @@ import {
 import { updateCachedChannelMemberDisplayName } from "@/features/channels/channelMemberProfileCache";
 import { evictUsersBatchEntries } from "@/features/profile/hooks";
 import {
+  useDocumentVisible,
+  useVisibleRefetchInterval,
+} from "@/shared/lib/useDocumentVisible";
+import {
   createManagedAgent,
   deleteManagedAgent,
   deleteCustomHarness,
@@ -322,6 +326,7 @@ export function useManagedAgentPrereqsQuery(
 }
 
 export function useRelayAgentsQuery(options?: { enabled?: boolean }) {
+  const refetchInterval = useVisibleRefetchInterval(5 * 60_000);
   return useQuery({
     queryKey: relayAgentsQueryKey,
     queryFn: listRelayAgents,
@@ -334,19 +339,21 @@ export function useRelayAgentsQuery(options?: { enabled?: boolean }) {
     // the `agents-data-changed` event fires only for local persona/team/managed
     // reconcile (kinds PERSONA/TEAM/MANAGED_AGENT), never for kind:10100. So we
     // keep polling but at a relaxed cadence and pause it while backgrounded.
-    refetchInterval: 5 * 60_000,
-    refetchIntervalInBackground: false,
+    refetchInterval,
+    refetchOnWindowFocus: true,
     enabled: options?.enabled,
   });
 }
 
 export function useManagedAgentsQuery(options?: { enabled?: boolean }) {
+  const documentVisible = useDocumentVisible();
   return useQuery({
     enabled: options?.enabled ?? true,
     queryKey: managedAgentsQueryKey,
     queryFn: listManagedAgents,
     staleTime: 5_000,
     refetchInterval: (query) => {
+      if (!documentVisible) return false;
       const agents = query.state.data as ManagedAgent[] | undefined;
       // Only local "running" agents need polling: process state can change
       // with no relay event to signal it, so this poll is the only liveness
@@ -357,6 +364,7 @@ export function useManagedAgentsQuery(options?: { enabled?: boolean }) {
         ? 5_000
         : false;
     },
+    refetchOnWindowFocus: true,
   });
 }
 
@@ -895,13 +903,15 @@ export function useManagedAgentLogQuery(
   pubkey: string | null,
   lineCount = 120,
 ) {
+  const refetchInterval = useVisibleRefetchInterval(pubkey ? 30_000 : false);
   return useQuery({
     queryKey: ["managed-agent-log", pubkey, lineCount],
     queryFn: () => getManagedAgentLog(pubkey as string, lineCount),
     enabled: pubkey !== null,
     retry: false,
     staleTime: 3_000,
-    refetchInterval: pubkey ? 30_000 : false,
+    refetchInterval,
+    refetchOnWindowFocus: true,
   });
 }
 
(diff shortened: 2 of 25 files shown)


wesbillman:
Reviewing on Wes's behalf.

The timer quiescing is well-targeted overall, but the reminder detector cannot be tied to document focus. That changes a user-facing reliability contract rather than merely removing background work.

`useReminderNotifications` is the sole fire-on-due detector (as its own doc comment states). With the new `!documentVisible` guard, it stops whenever Buzz is hidden **or merely unfocused**. A reminder due while the user is working in another app will therefore produce no desktop notification, sound, or dock bounce at its scheduled time; it is only detected after Buzz regains focus, when the notification is least useful and potentially much later. This is exactly the state where desktop reminders need to keep working.

Please keep the reminder due-time mechanism active while backgrounded, or move it to a background-capable/native scheduler that preserves on-time delivery. If CPU is the concern, a single 30-second reminder timer is not equivalent to cosmetic tickers/query backstops and should be among the deliberately ungated semantics (similar to presence/watchdog). Please also add coverage for a reminder crossing `not_before` while the document is hidden/unfocused.

PR #5496, opened 2026-08-10
docs: field writeup — running a local multi-model debate lab on a self-hosted relay

artemtrofymenko:
## What this is

A field writeup from standing up Buzz as a local multi-model debate lab on Windows: a self-hosted relay in Docker, the packaged desktop, and four agents — Claude Code and Codex over OAuth, plus Qwen3.5 9B and Gemma4 12B locally through Ollama — arguing in one channel with turns driven by a Workflow.

Four documents under `docs/local-debate-lab/`, plus the workflow definition that ran the sessions.

**The docs are written in Ukrainian.** I am opening this as offered rather than assumed — if a non-English doc set is not something the project wants to carry, say so and I will close it, or translate it if that would make it useful.

## Why it might be worth carrying

Most of what cost time here is already known to this tracker. What the writeup adds is the *sequence*: which symptoms lead to which cause, and what a self-hoster sees before they know the vocabulary to search for.

- **`RELAY_URL=ws://localhost:3000` silently breaks every managed agent.** The relay resolves the community from `Host` and fails closed; the desktop normalizes loopback to `127.0.0.1` before handing the URL to the harness. The relay is healthy, `/_liveness` returns 200, the desktop works, the agent looks running — and the only evidence is a 404 in a harness log nothing points you to. Related: #4147, #3283, #3505.
- **`/_liveness` is not a readiness signal for agents.** It answers 200 on hosts that map to no community. The WS upgrade (101 vs 404) is the check that distinguishes them.
- **Measured local-model data** on 12 GB VRAM: served context (8192) versus advertised (262144), reasoning budgets, and tool-call reliability per model — the metric that actually decides whether a local model can drive an agent at all, since publishing happens through a shell tool call.
- **A verified workaround for #2501**, also posted in that thread.
- **What the debates showed**: the cloud models spotted each other's errors and narrowed their own claims; the local ones stayed on topic once the topic was placed in every prompt, but never conceded a point.

## What this does not touch

No code. Documentation only.


artemtrofymenko:
Closing this after a month without review, so it does not sit in the queue indefinitely.

The write-up stays readable on the branch it came from: https://github.com/artemtrofymenko/buzz/tree/docs/local-debate-lab/docs/local-debate-lab

The findings in it that are product defects are tracked where they belong rather than here — #5460 (auth probe vs runtime credentials), #2501 (respond_to edits and the owner-only clamp), #4147 (loopback authority split), #5053 (multi-value `#h` narrowing), #2698 (implicit reply delivery), and #3309 / #7287 (harness log targets and the desktop `RUST_LOG` construction). Happy to reopen or rework it if a docs contribution of this shape is wanted.

PR #5510, opened 2026-08-10
fix(link-preview): refetch a link when it re-enters the composer

tellaho:
### Overview

**Category:** fix
**User Impact:** When a user re-pastes (or finishes typing) a link that previously failed to load a preview, the composer now refetches it immediately and can never send a snapshot preview built from the old, stale metadata.
**Problem:** The link-preview cache is shared with passive message-list scroll, so a URL that resolved to a negative result (a hard `null` miss or a transient fetch failure) stayed cached and re-usable. Re-pasting that exact link into the composer served the stale negative and never refetched. Worse, the stale metadata was still `snapshotReady`, so a fast clear-then-repaste could attach a **stale snapshot preview tag** to the sent message — a preview that no longer matched the link.
**Solution:** A freshly-entering link is forced to refetch, and the composer is fenced against ever shipping a tag built from pre-re-entry metadata. This closes three distinct races surfaced over successive review passes: (1) the shared negative cache being reused on re-entry; (2) the resolver's debounce swallowing a fast clear+re-paste so the re-entry was invisible and the stale tag stayed sendable; and (3) an in-flight media upload started from the stale metadata publishing its tag after fresh metadata had already arrived. Healthy cached hits are never touched (instant card, no redundant fetch), and passive message-list scroll — which never opts in — keeps riding the shared cache exactly as before.

<details>

**desktop/src/shared/lib/useResolvedLinkPreviews.ts**
Adds a loader `invalidateNegative(href)` that drops a cached negative result (resolved `null` or transient fail) while leaving healthy hits and in-flight promises alone, and a `refetchNewNegatives` option that invalidates each newly-present href's negative entry before the peek/load loop reads the cache. Also adds an optional `liveHrefs` input so newness is judged against the caller's LIVE (undebounced) content — a debounce-swallowed leave/re-entry of the same URL still counts as new. Because the hook retains its own resolved metadata (the render that scheduled the effect already read the stale negative from it), it also clears its OWN negative key for every re-entered href, so the link renders as pending until the fresh load wins. `buzz://` entity links are skipped (they resolve off the relay, not this cache).

**desktop/src/features/messages/ui/useComposerLinkPreviews.tsx**
Opts the composer into `refetchNewNegatives` and feeds it the live hrefs. Detects a same-URL re-entry at render time (React batches the empty→repaste renders, so an effect keyed on the live set never observes the transition), then blocks the re-entered href until the resolver's forced refetch visibly cycles through pending: its stale ready tag is dropped from state and excluded from the sendable output until a fresh result re-tags. Only the sendable negative case (`fallback`) is blocked; a healthy (`image`) re-entry keeps its instant card. Adds a per-href upload generation token (`uploadsRef` becomes `Map<href, generation>`): a live re-entry bumps the generation, the upload effect's dedup guard and completion are generation-aware, so an in-flight upload from stale metadata cannot publish its tag after settling and a fresh upload can start even while the superseded one is still in flight.

**desktop/src/shared/lib/useResolvedLinkPreviews.test.mjs**
Adds resolver-level regressions: `invalidateNegative` drops a cached miss (next load refetches) but preserves a healthy hit (no redundant fetch); transient failure → URL removed → re-entered renders pending/not-`snapshotReady` until a successful retry; and the retained-negative + shared in-flight-fetch + re-entry interleaving clears the local negative regardless of the shared entry's shape.

**desktop/src/features/messages/ui/useComposerLinkPreviews.test.mjs**
Adds composer-hook regressions driving the REAL hook through the hostile gestures: a fast clear+re-paste inside the debounce window drops the stale tag and holds Send pending until a fresh tag carrying the newly-fetched media lands; and a stale in-flight upload held across the clear+re-paste and fresh-metadata resolution cannot publish its pre-clear tag, while a fresh upload starts and its tag wins.

</details>

### Reproduction Steps

1. Paste a link whose preview fails to resolve (force a transient fetch failure) so the composer shows a blank/collapsed card.
2. Clear the composer and re-paste the same link (quickly, within the ~350ms debounce window).
3. Observe the preview refetches immediately rather than reusing the stale negative result, and Send stays disabled until a fresh tag lands.
4. Send the message and confirm the attached preview tag reflects the fresh fetch, never the stale pre-clear metadata.
5. Confirm passive message-list scroll of already-resolved links still shows cards instantly with no extra fetches.

### Notes

Scope grew across three review passes from the original single resolver opt-in into a full defense against shipping stale snapshot tags on link re-entry — see the scope-adjustment comment on this PR for the detail. Stacked on #5245 (`tho/link-preview-snapshot-race`), whose rewrite of `useComposerLinkPreviews.tsx` is the sole overlapping file. The transient-retry work stays in #5502, which touches no composer file and remains based on main.

### DIFF
--- a/desktop/src/shared/lib/useResolvedLinkPreviews.ts
+++ b/desktop/src/shared/lib/useResolvedLinkPreviews.ts
@@ -444,6 +444,7 @@ export function useResolvedLinkPreviews(
   {
     refetchNewNegatives = false,
     liveHrefs,
+    liveHrefVersions,
   }: {
     /**
      * When a preview href is newly present since the last run, drop any cached
@@ -463,19 +464,29 @@ export function useResolvedLinkPreviews(
      * uses this. Omit to track newness against `previews` (the default).
      */
     liveHrefs?: readonly string[];
+    /**
+     * Per-href entry versions captured at the live editor-update boundary.
+     * Unlike committed href-set equality, a bumped version preserves an
+     * intermediate leave/re-entry even when React batches both updates into one
+     * commit with the same final href set.
+     */
+    liveHrefVersions?: ReadonlyMap<string, number>;
   } = {},
 ): ResolvedLinkPreview[] {
   const [resolvedMetadata, setResolvedMetadata] =
     React.useState<ResolvedMetadataByHref>({});
   const [retryGeneration, setRetryGeneration] = React.useState(0);
   const seenHrefsRef = React.useRef<Set<string>>(new Set());
+  const handledHrefVersionsRef = React.useRef<Map<string, number>>(new Map());
   // Drive newness tracking from a stable string key so an equivalent href list
   // does not restart the effect. The effect closes over the committed render's
   // hrefs; do not mirror them into a ref during render, because an abandoned
   // concurrent render could otherwise leak uncommitted presence into the live
   // effect from the previous commit.
   const currentHrefs = liveHrefs ?? previews.map((preview) => preview.href);
-  const newnessKey = currentHrefs.join("\n");
+  const newnessKey = currentHrefs
+    .map((href) => `${href}\0${liveHrefVersions?.get(href) ?? ""}`)
+    .join("\n");
   // biome-ignore lint/correctness/useExhaustiveDependencies: newnessKey is the stable identity for currentHrefs; depending on the freshly allocated array would rerun this effect every render.
   React.useEffect(() => {
     let cancelled = false;
@@ -489,6 +500,7 @@ export function useResolvedLinkPreviews(
       // Newness is judged against the live href set when supplied (so a
       // debounce-swallowed leave/re-entry still counts), else against previews.
       const seen = seenHrefsRef.current;
+      const handledVersions = handledHrefVersionsRef.current;
       const liveNow = currentHrefs;
       // Preserve handled hrefs only while they remain live. A live href is not
       // marked handled until its debounced preview exists and invalidation has
@@ -497,10 +509,18 @@ export function useResolvedLinkPreviews(
       const next = new Set<string>(
         [...seen].filter((href) => liveNow.includes(href)),
       );
+      const nextVersions = new Map(
+        [...handledVersions].filter(([href]) => liveNow.includes(href)),
+      );
       const reenteredKeys: string[] = [];
       for (const preview of previews) {
+        const version = liveHrefVersions?.get(preview.href);
+        const alreadyHandled =
+          version === undefined
+            ? seen.has(preview.href)
+            : handledVersions.get(preview.href) === version;
         if (
-          seen.has(preview.href) ||
+          alreadyHandled ||
           !liveNow.includes(preview.href) ||
           preview.href.startsWith("buzz://")
         ) {
@@ -512,8 +532,10 @@ export function useResolvedLinkPreviews(
         metadataLoader.invalidateNegative(preview.href);
         reenteredKeys.push(metadataCacheKey(preview.href));
         next.add(preview.href);
+        if (version !== undefined) nextVersions.set(preview.href, version);
       }
       seenHrefsRef.current = next;
+      handledHrefVersionsRef.current = nextVersions;
       // Dropping the loader entry alone is not enough: this hook retains its own
       // resolved metadata, and the render that scheduled this effect already
       // read the stale negative from it. Clear this hook's OWN negative key for
--- a/desktop/src/shared/lib/useResolvedLinkPreviews.ts
+++ b/desktop/src/shared/lib/useResolvedLinkPreviews.ts
@@ -469,16 +469,14 @@ export function useResolvedLinkPreviews(
     React.useState<ResolvedMetadataByHref>({});
   const [retryGeneration, setRetryGeneration] = React.useState(0);
   const seenHrefsRef = React.useRef<Set<string>>(new Set());
-  // Newness is tracked against the live href set when the caller supplies one,
-  // so a debounce-swallowed leave/re-entry of the same URL still counts as new.
-  // Read through a ref inside the effect and drive re-runs off the stable string
-  // key, so an unstable per-render array does not have to be an effect dep.
+  // Drive newness tracking from a stable string key so an equivalent href list
+  // does not restart the effect. The effect closes over the committed render's
+  // hrefs; do not mirror them into a ref during render, because an abandoned
+  // concurrent render could otherwise leak uncommitted presence into the live
+  // effect from the previous commit.
   const currentHrefs = liveHrefs ?? previews.map((preview) => preview.href);
   const newnessKey = currentHrefs.join("\n");
-  const currentHrefsRef = React.useRef<readonly string[]>(currentHrefs);
-  currentHrefsRef.current = currentHrefs;
-
-  // biome-ignore lint/correctness/useExhaustiveDependencies: newnessKey is the stable string key for the live href set read via currentHrefsRef; it drives the re-run so the per-render array need not be a dep.
+  // biome-ignore lint/correctness/useExhaustiveDependencies: newnessKey is the stable identity for currentHrefs; depending on the freshly allocated array would rerun this effect every render.
   React.useEffect(() => {
     let cancelled = false;
     let retryAt = Number.POSITIVE_INFINITY;
@@ -491,8 +489,14 @@ export function useResolvedLinkPreviews(
       // Newness is judged against the live href set when supplied (so a
       // debounce-swallowed leave/re-entry still counts), else against previews.
       const seen = seenHrefsRef.current;
-      const liveNow = currentHrefsRef.current;
-      const next = new Set<string>(liveNow);
+      const liveNow = currentHrefs;
+      // Preserve handled hrefs only while they remain live. A live href is not
+      // marked handled until its debounced preview exists and invalidation has
+      // actually been attempted; otherwise blank -> paste would consume
+      // newness during the 350ms debounce and later reuse the stale negative.
+      const next = new Set<string>(
+        [...seen].filter((href) => liveNow.includes(href)),
+      );
       const reenteredKeys: string[] = [];
       for (const preview of previews) {
         if (
@@ -507,6 +511,7 @@ export function useResolvedLinkPreviews(
         // the shared entry is healthy, in-flight, or absent.)
         metadataLoader.invalidateNegative(preview.href);
         reenteredKeys.push(metadataCacheKey(preview.href));
+        next.add(preview.href);
       }
       seenHrefsRef.current = next;
       // Dropping the loader entry alone is not enough: this hook retains its own
--- a/desktop/src/shared/lib/useResolvedLinkPreviews.ts
+++ b/desktop/src/shared/lib/useResolvedLinkPreviews.ts
@@ -86,6 +86,13 @@ function metadataCacheKey(href: string): string {
   }
 }
 
+function isNegativeMetadata(metadata: LinkPreviewMetadata | null): boolean {
+  // A cached NEGATIVE is a result that should be retried when the URL freshly
+  // re-enters the composer: a hard miss (null) or a transient image failure.
+  // A healthy hit (`image`/`rejected`/no state) is a settled positive.
+  return metadata === null || metadata.imageFetchState === "transient_failure";
+}
+
 function metadataExpiry(
   metadata: LinkPreviewMetadata | null,
   now: number,
@@ -185,6 +192,23 @@ function createMetadataLoader({
   return {
     deleteKey(key: string) {
       cache.delete(key);
+    },
+    /**
+     * Drop a cached NEGATIVE result (a resolved null or a transient failure) so
+     * the next load refetches. Used when a URL freshly enters the composer: a
+     * user pasting a link that previously blanked should get a new attempt now,
+     * not the stale miss. A healthy cached hit and an in-flight fetch are left
+     * untouched, so passive scroll re-renders still ride the cache as before.
+     * Returns whether a negative entry was actually dropped, so callers can
+     * invalidate their own derived state (e.g. retained React metadata) in step.
+     */
+    invalidateNegative(href: string): boolean {
+      const key = metadataCacheKey(href);
+      const cached = cache.get(key);
+      if (!cached || cached instanceof Promise) return false;
+      if (!isNegativeMetadata(cached.metadata)) return false;
+      cache.delete(key);
+      return true;
     },
     load,
     peek,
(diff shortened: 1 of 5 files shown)


desktop/src/shared/lib/useResolvedLinkPreviews.ts
```diff
@@ -177,6 +177,19 @@ function createMetadataLoader({
...
+     * the next load refetches. Used when a URL freshly enters the composer: a
+     * user pasting a link that previously blanked should get a new attempt now,
+     * not the stale miss. A healthy cached hit and an in-flight fetch are left
+     * untouched, so passive scroll re-renders still ride the cache as before.
+     */
+    invalidateNegative(href: string) {
+      const key = metadataCacheKey(href);
+      const cached = cache.get(key);
+      if (!cached || cached instanceof Promise) return;
+      if (cached.metadata === null) cache.delete(key);
```
wesbillman: This only invalidates hard `null` misses. A cached metadata object with `imageFetchState === "transient_failure"` survives, even though this method's contract, the option docs, and the PR description all classify transient failures as negatives that should refetch when the URL re-enters. A focused test confirms the second load reuses the transient entry and never calls the fetcher again. Please include transient failures in this predicate and cover that path.

wesbillman (changes requested):
Reviewing on Wes's behalf. The new behavior does not cover one of the two negative-cache states it explicitly promises to retry. I reproduced this with a focused loader test: after caching `imageFetchState: "transient_failure"`, `invalidateNegative()` followed by `load()` returned the same cached transient result and the fetch count stayed at 1. Please make transient failures invalidatable (or narrow the stated behavior if they are intentionally governed only by their retry boundary) and add coverage for that case. The current test also says it preserves an in-flight fetch but never creates one; please exercise that claimed invariant directly. Separately, this stacked PR is currently conflicting/dirty against its advanced base, so its green checks do not establish merge readiness for the eventual rebased head.

wesbillman (commented):
Re-review by Carl on Wes's behalf: the prior blocker is resolved at this head. `invalidateNegative` now invalidates both hard null misses and transient failures, while preserving healthy and in-flight entries; focused tests cover all four states. I found no new actionable issues. Per owner policy I am not auto-approving; Wes can approve this head.

wesbillman (changes requested):
Reviewing on Wes's behalf. Do not merge this head: invalidating the shared metadata loader does not invalidate `useResolvedLinkPreviews`' retained `resolvedMetadata` state.

On URL re-entry, the effect deletes the loader entry at `useResolvedLinkPreviews.ts:463-475`, but the render that scheduled that effect has already read the old local metadata at `:538-545`, and the effect never removes that key from local state. For a cached `transient_failure`, the composer therefore still sees a resolved `snapshotReady` fallback and can start producing a ready snapshot tag before the retry resolves (`useComposerLinkPreviews.tsx:289-349`), allowing stale metadata to become sendable. A cached `null` likewise remains the rendered no-preview state until the retry completes.

Please clear or generation-invalidate the corresponding local resolved state when a newly re-entered negative is invalidated, so that the retry is represented as pending immediately. Add a hook/composer-level regression covering transient failure -> URL removed -> URL re-entered -> pending/no ready tag -> successful retry wins. The loader-only tests do not exercise this second cache layer.

tellaho:
@wesbillman thanks for the catch — your 14:53Z review was right, no pushback. Fix pushed. Summary:

**Root cause:** Invalidating the shared loader cache alone wasn't enough. `useResolvedLinkPreviews` keeps its own `resolvedMetadata` state, and the render that scheduled the invalidating effect had *already* read the stale negative out of it. So a re-entered `transient_failure` stayed a `snapshotReady` fallback — which the composer could turn into a **sendable snapshot tag built from stale metadata** before the retry landed.

**Fix (`useResolvedLinkPreviews.ts`):**
- `invalidateNegative` now returns whether it actually dropped a negative entry.
- The re-entry effect clears the matching local `resolvedMetadata` key in step, so the link renders **pending** (no `snapshotReady`) until the fresh load wins. Healthy hits and passive scroll are untouched.

**Regression test:** hook-level, drives the real hook via `renderHook` + JSDOM + mocked Tauri IPC — transient failure → URL removed → re-entered → asserts pending / no ready tag → retry succeeds. Proven genuine: it fails with `'fallback'` (the stale sendable state) when the fix is temporarily removed, passes with it in.

**Verification:** tsc clean, biome clean, full desktop suite 4609/4609 pass, pre-push hooks green. Rebased onto latest `main`.

tellaho:
Follow-up — **also found this** during an adversarial re-review, and fixed it in the same commit (new head `24785e3aa`). @wesbillman flagging so you see it on your next pass.

**Also found:** an in-flight shared-cache entry defeated the local-state clear.

The `useResolvedLinkPreviews` loader is intentionally shared across every hook instance (composer + message list). `invalidateNegative` deliberately leaves an *in-flight Promise* entry alone. But my re-entry effect only cleared this hook's retained negative when `invalidateNegative` returned `true` — i.e. only when it dropped a *settled* negative. So this sequence stayed sendably stale:

1. The composer hook retains a `transient_failure` in local state; its URL leaves.
2. The shared negative expires/is evicted, and another instance (e.g. the message list, or a retry timer) starts a refetch for the same canonical URL — leaving a **Promise** in the shared cache.
3. The URL re-enters the composer while that fetch is in flight.
4. `invalidateNegative` returns `false` (the shared entry is a Promise), so the local clear was skipped — the composer kept returning `imageState: "fallback", snapshotReady: true`, a sendable tag built from stale metadata, until the shared fetch resolved.

Same defect *class* as the original bug, just via the in-flight path instead of the settled-negative path.

**Fix (`useResolvedLinkPreviews.ts`):** the re-entry effect now clears this hook's **own** negative local key for every newly-present href, gated on the hook's *local* value (`isNegativeMetadata`), independent of whether the shared loader dropped a settled entry. `invalidateNegative` still runs to drop a settled shared negative; the loop then coalesces onto any in-flight Promise. Healthy local hits are kept, so passive scroll still shows cards instantly.

**Regression test:** a second hook-level test drives the exact interleaving — retained transient local state → URL leaves → shared cache dropped → a message-list instance starts a shared in-flight fetch → composer re-enters → asserts pending / no `snapshotReady` while in flight, coalesces (no third fetch), then settles on the successful resolve. **Proven genuine:** it fails on the old drop-gated logic (`re-entry during a shared in-flight fetch must render pending, not a stale fallback`) while the original re-entry test still passes — so it pins the *new* in-flight gap specifically.

**Verification:** tsc clean, biome clean, full hook test file 23/23, pre-push hooks green (desktop-check + typecheck + full desktop-test at head `24785e3aa`).

Credit to an adversarial second-reviewer pass for catching the in-flight interleaving.

tellaho:
Also found + fixed a second call-site gap (thanks @wesbillman for the earlier round; this one surfaced in a follow-up review).

**The bug:** the composer feeds `useResolvedLinkPreviews` from *debounced* `candidates` (350ms), but a URL's leave/re-entry has to be observed from *live* content. On a fast clear-then-repaste of the same URL A inside the debounce window, each keystroke resets the debounce timer so the debounced set never commits the empty intermediate — `candidates` stays `[A]` throughout. The newness tracker never sees A leave, never invalidates, and the stale `transient_failure` snapshot tag (keyed by A) stays sendable. The two hook-level tests couldn't catch it because they bypass `useComposerLinkPreviews`.

**The fix (2 layers):**
- `useResolvedLinkPreviews` now takes an optional `liveHrefs` and judges newness against the live href set (read via a ref, driven by a stable `newnessKey`), so a debounce-swallowed leave/re-entry still forces the refetch.
- The composer detects the same re-entry at *render time* — React batches the empty→repaste renders, so an effect keyed on the live set never observes the transition. On a stale-fallback re-entry it drops the stale tag from state, excludes the href from sendable output, and blocks the upload effect (and any in-flight upload's completion) from rebuilding a tag from the pre-clear metadata. The href is released only after the resolver's forced refetch visibly cycles through pending → ready, so only genuinely-fresh metadata can re-tag. Healthy (`image`) re-entries keep their instant card and are never blocked.

**Test:** new `useComposerLinkPreviews.test.mjs` drives the real hook through the fast gesture with a deferred re-entry fetch. Asserts the stale tag is not sendable the moment the URL re-enters (synchronous invariant), that the re-entry forces a fresh fetch while Send is held pending, then that a fresh tag carrying the newly-fetched media lands once the refetch resolves. Verified genuine: reverting either fix layer makes it fail fast.

Full desktop suite green (4673 pass); pre-push hook `desktop-test` green. Author/DCO amended into the feature commit.

tellaho:
@wesbillman Third blocker addressed (thanks to Princess Donut's third review pass) — head is now `50ead432d`.

**The race:** an old in-flight media upload (U1), started from the pre-clear transient-fallback metadata, could still publish a snapshot tag *after* the URL was cleared, re-pasted, and freshly resolved. The `reenteringHrefsRef` phase marker alone didn't fence it: the upload effect deletes that marker the moment the forced refetch reaches fresh-ready metadata, which can happen *before* U1 settles — so U1's completion then passed the marker guard and wrote a tag built from the stale pre-clear metadata. Secondary hole: because the old href still occupied `uploadsRef`, the effect skipped starting a fresh upload (U2), and U1's `.finally` cleared the slot with no state change to re-arm one, so the composer could sit tagless until the anti-trap cap.

**Fix — per-href upload generation token:**
- `uploadsRef` is now `Map<href, generation>`; a live re-entry bumps `uploadGenerationRef`.
- The upload effect captures the current generation, and its **dedup guard** compares against it — so a superseded in-flight upload no longer blocks starting the fresh one at the new generation (U2 starts immediately on fresh-ready).
- A **durable generation fence in `.then`**, independent of the phase marker, makes a stale completion a no-op even after the marker was cleared.
- `.finally` only clears the slot when it still owns the current generation, so it can't evict U2's entry.

**Regression:** new composer-hook test holds U1 unresolved across the clear + re-paste and the fresh-metadata resolution, proves U2 starts and its fresh tag wins, then releases U1 and proves it cannot publish its pre-clear tag. Proven genuine — it fails on the prior head both ways (U2 never starts with the old `Set` dedup; and, with only the dedup fix, U1 publishes its stale favicon). `tsc` + `biome` clean; full desktop suite 4674 pass; pre-push `desktop-test` green.

tellaho:
📝 **Scope adjustment — description updated to match what actually shipped.**

The original description covered only the first fix: an opt-in `refetchNewNegatives` mode in `useResolvedLinkPreviews` plus a one-line composer opt-in and a single resolver test. Over three successive review passes (thanks @wesbillman and the review crew) two further races were found and fixed on this branch, so the "File changes" section and the User Impact/Solution had drifted out of date. I've broadened the description to describe all three, with no change to the shipped code:

1. **Resolver negative-cache re-entry** (original scope): `invalidateNegative` + `refetchNewNegatives`, and clearing the hook's OWN retained negative key so a re-entry renders pending until the fresh load wins.
2. **Debounce-swallowed re-entry** (`useResolvedLinkPreviews.ts` + `useComposerLinkPreviews.tsx`): the composer feeds the resolver DEBOUNCED candidates, so a fast clear+re-paste of the same URL inside ~350ms never committed an empty set and the re-entry was invisible — the stale `transient_failure` tag stayed sendable. Fixed with live-href newness tracking in the resolver and render-time re-entry detection in the composer that drops the stale tag and excludes the href from sendable output until the forced refetch cycles through pending.
3. **Stale in-flight upload** (`useComposerLinkPreviews.tsx`): an old media upload started from pre-clear metadata could publish a snapshot tag after fresh metadata arrived. Fixed with a per-href upload generation token — `uploadsRef` is now `Map<href, generation>`, a live re-entry bumps the generation, and the dedup guard + `.then` fence + `.finally` are generation-aware.

Net diff: `useComposerLinkPreviews.tsx` +171 / `useResolvedLinkPreviews.ts` +114, plus deterministic regressions in both test files (+389 / +415). `tsc` + `biome` clean; full desktop suite 4674 pass; pre-push `desktop-test` green at head `50ead432d`.

wesbillman (changes requested):
Carl reviewing on Wes's behalf.

The stale-upload generation fencing is sound for committed renders, but the re-entry detector introduces render-phase side effects that are unsafe under React concurrent rendering. At `useComposerLinkPreviews.tsx:327-373`, render mutates `prevLiveHrefsRef`, `reenteringHrefsRef`, and `uploadGenerationRef`, then queues a `setReadyTags` microtask. React may restart or abandon that render before commit (Suspense, a transition, or another higher-priority update), but none of those mutations are rolled back and the queued state update still runs.

That can make an uncommitted clear/re-paste count as a real re-entry: the committed composer still contains the old URL, yet its valid tag can be dropped, its upload generation invalidated, and the href left in `blocked`. If the corresponding resolver transition was also abandoned, nothing guarantees the required pending -> ready cycle that releases the block. After the two-second cap, the user can send a bare link despite having a valid snapshot before the abandoned render.

Please keep render pure. Model committed live-presence/generation as state or update it in a layout/effect phase, while retaining a synchronous submit fence derived from current content/state rather than mutating refs and scheduling state during render. Add a regression that starts and abandons/restarts a re-entry render (for example via a suspending sibling) and proves the previously committed tag remains sendable and no block/generation change leaks from the abandoned render.

wesbillman (commented):
Reviewing on Wes's behalf at exact head `50ead432d13ec0e2b6c97fbcec4fd2b886bbb6c1`.

**P1 — the ordinary blank-composer → paste path consumes “newness” before the debounced preview exists, so the negative cache is not invalidated.**

`useComposerLinkPreviews` passes debounced `candidates` as `previews`, but live hrefs immediately (`useComposerLinkPreviews.tsx:225-248`). In `useResolvedLinkPreviews`, every effect run replaces `seenHrefsRef` with the live set (`useResolvedLinkPreviews.ts:493-511`), while invalidation can only happen inside `for (const preview of previews)` (`:497-509`).

Concrete production sequence:

1. Composer is mounted blank: `previews=[]`, `liveHrefs=[]`, `seen={}`.
2. User pastes URL A whose shared cache contains a negative. Before the 350ms debounce, `previews=[]` but `liveHrefs=[A]`.
3. The effect has no preview to invalidate, yet sets `seenHrefsRef.current = {A}`.
4. After debounce, `previews=[A]` and `liveHrefs=[A]`; `seen.has(A)` is now true, so invalidation is skipped and the stale negative is reused.

That is the headline behavior this PR promises to fix. The composer regressions miss it because they mount the hook with URL A already in `initialProps`, which initializes `debounced` to A (`useComposerLinkPreviews.test.mjs:123-130`); they do not drive the normal mounted-empty → paste transition. The fast clear/re-paste test works only because debounced `candidates` remains `[A]` while the live set temporarily empties.

**Required fix:** do not mark a live href as seen/handled until there is a corresponding resolvable `preview` and its negative invalidation has been attempted (or track live presence separately from pending invalidation). Add a composer-level regression that mounts with empty content, seeds/caches a negative for A, then pastes A and proves the post-debounce load refetches rather than reusing the negative.

The generation fencing for the already-covered fast re-entry/upload races otherwise looks internally consistent, but this gap leaves the primary user gesture broken. CARL, THE DEBOUNCE ATE THE INVALIDATION. 👑🐈

tellaho:
@wesbillman Both August 12 review-body findings are addressed at `3d458a48f`.

- **Concurrent-render safety:** render is pure now. Live hrefs used by submit/upload paths are committed in a layout effect; re-entry phase, upload generation, and stale-tag removal also move to commit-time layout work. The candidate render still derives a pure synchronous fence from current content, so Send cannot expose the stale tag. Added a Suspense/startTransition regression that executes and abandons a clear render, then proves the previously committed tag remains sendable and pending/block state does not leak.
- **Blank → paste newness:** `useResolvedLinkPreviews` now retains only handled hrefs and marks a href handled after a matching debounced preview actually attempts invalidation. Added the normal mounted-empty → paste regression with a pre-seeded transient negative; it proves the post-debounce resolver performs a second fetch and builds the fresh tag.

Focused composer + resolver tests: 27/27. TypeScript and Biome clean. Push hooks also passed desktop-check, desktop-typecheck, and the full desktop-test suite.

Updated by Carl 🤖

wesbillman (changes requested):

**P1 — the commit-only re-entry tracking still loses a genuinely batched clear → re-paste, so stale metadata remains sendable.**

The render-purity fix now compares `liveCandidates` only with the last committed set and records presence in a layout effect (`useComposerLinkPreviews.tsx:307-359`). The resolver likewise records live presence only from its passive effect (`useResolvedLinkPreviews.ts:471-516`). If React batches editor updates from `[A] -> [] -> [A]` without committing the empty intermediate, the final live set and `newnessKey` are identical to the previous committed values. Consequently:

- `staleReenteredHrefs` is empty, so the stale tag is not dropped and the upload generation is not bumped;
- the resolver effect does not observe A leaving, so its handled/seen entry remains and the negative cache is not invalidated;
- the old fallback tag remains eligible for `getReadyTags()`.

This is reachable through the production boundary: `MessageComposer.tsx:270-273` receives each editor update and calls `setPreviewContent`, so React may batch multiple updates before the composer commits. The two claimed fast-gesture regressions do not cover that sequence: each `renderHook.rerender` is wrapped in a separate `act` (`useComposerLinkPreviews.test.mjs:139-148` and `:329-335`), forcing the empty render to commit and letting the layout/passive effects observe it.

Please preserve the editor update/revision transition independently of committed href-set equality (for example, pass an event/revision signal captured at the editor update boundary, or use state/reducer bookkeeping that retains the intermediate transition), and add a regression that delivers clear and same-URL re-paste in one batch/commit. It should prove the stale tag is fenced immediately, a fresh fetch occurs, and the upload generation invalidates pre-clear work.

tellaho:
@wesbillman The batched clear → same-URL re-paste blocker is addressed at `1097769c8`.

- `MessageComposer` now records per-href entry versions inside the functional state updater that runs for every editor `onUpdate`. React may batch the renders, but it cannot erase the intermediate state transition: `[A:v1] → [] → [A:v2]` commits as `[A:v2]`.
- The composer compares the committed version as well as href presence, so the final same-set commit synchronously fences the stale tag, bumps the upload generation, and blocks pre-clear upload completion.
- The resolver carries the same version signal through its handled/newness bookkeeping, so the post-debounce candidate invalidates and performs a fresh fetch rather than reusing the retained negative.
- Both fast-gesture regressions now fold clear + re-paste into one React `act` and issue only the final rerender, while exercising the production update-boundary reducer. The stale-tag test was proven genuine: removing version propagation fails immediately with `1 !== 0` on the sendable-tag assertion.

Validation at exact pushed head `1097769c8`: focused composer + resolver tests **27/27**, TypeScript clean, Biome clean, file-size ratchet clean. Push hooks passed `desktop-check`, `desktop-typecheck`, and full `desktop-test`.

Updated by Carl 🤖

wesbillman (dismissed):

The prior batched re-entry blocker is fixed at `1097769c83647f794780f9a057bed1ebd5f4105f`, and I rechecked the complete updated production/test paths. One material issue remains in the new repair:

**P1 — the update-boundary version map grows forever and is cloned on every editor update, creating unbounded retention and steadily worsening typing cost.**

`updateComposerLinkPreviewInput` starts each update with `new Map(current.hrefVersions)` and never removes hrefs that left the draft (`useComposerLinkPreviews.tsx:211-222`). `MessageComposer` invokes this state update for every editor `onUpdate`, including ordinary keystrokes (`MessageComposer.tsx:270-273`). The composer is not keyed to channel identity in the production ownership chain: `ChannelScreen` renders `ChannelPane` without a key (`ChannelScreen.tsx:853`), and `ChannelPane` renders `MessageComposer` without one (`ChannelPane.tsx:699`). Sending/clearing likewise does not unmount this managed input. Therefore the map retains every distinct supported URL observed over the long-lived composer instance, and every later keystroke copies all of them: retained memory is O(historical distinct URLs), with O(historical distinct URLs) work per editor update.

Please bound version state while preserving the fixed batched transition. One small shape is to retain versions only for `nextHrefs` plus a monotonic `nextVersion` counter: carry the existing version for still-active hrefs, assign `nextVersion++` to entrants, and drop departed href entries. Add a focused reducer test proving both properties together: after many departed URLs only active hrefs remain, while batched `A -> [] -> A` still advances A’s entry version and triggers the existing stale-tag/refetch fencing.

I also rechecked the prior blockers on this complete head: the update-boundary signal now survives one-commit batching; render-phase mutation remains removed; ordinary blank-to-paste does not consume newness before debounce; local negative metadata and stale uploads remain fenced. I found no other actionable issue in those paths. The current test changes are substantial (582 added composer-hook lines and 414/1 in resolver tests); the new one-batch regressions correctly improve the previously insufficient test shape.

wesbillman (commented):

I applied the final bounded-state fix and pushed exact head `6cebe6c1bad2643331d19e3d38e25569bb53feb4`.

`updateComposerLinkPreviewInput` now retains versions only for currently active hrefs and assigns entrants from a monotonic counter. This preserves the one-batch `A → empty → A` witness while bounding retained memory and per-update map work. A focused regression proves departed href history is pruned and a later re-entry still advances its version.

I also merged current `origin/main` because the mandatory branch-skew guard found an overlapping `MessageComposer.tsx` change. The merge was clean.

Validation at exact pushed head:
- focused composer hook tests: 5/5 pass, including batched re-entry, stale in-flight upload, blank-paste retry, abandoned render, and active-only version pruning
- TypeScript clean
- Biome clean for both touched files
- mandatory pre-push hook: branch-skew, desktop-check, desktop-typecheck, and full desktop-test all pass; desktop suite 4,786/4,786
- commit authorship/DCO verified for both new commits

No actionable issues remain in my review. Per owner policy I am not auto-approving; Wes can approve this head once desired CI is green.

wesbillman (approved):

wesbillman (changes requested):

**P1 — removing a URL while its re-entry refetch is blocked can leave that href permanently blocked when it is pasted again.**

`reenteringHrefsRef` is populated with `"blocked"` at `useComposerLinkPreviews.tsx:421-428`, but candidate removal only prunes `readyTags` (`:373-380`). It never removes the href from `reenteringHrefsRef`. The marker is cleared only after this mounted composer observes that same href cycle through `!snapshotReady` and then ready (`:446-461`).

A reachable sequence is:

1. A has stale fallback metadata, leaves, and re-enters, so the layout effect marks A `blocked` and starts the forced refetch.
2. The user removes A again before this composer observes the pending → ready cycle.
3. The refetch succeeds while A is absent, leaving a healthy shared cache entry.
4. The user later pastes A again. The healthy entry is intentionally preserved, so it resolves ready immediately rather than cycling pending. `staleReenteredHrefs` also refuses to reinitialize A because the old marker still exists (`:403-406`).
5. The preview effect sees `phase === "blocked"` plus ready metadata and continues forever (`:453-460`). No tag is built. After the two-second cap, Send permits a bare link.

This stale marker can survive sends, channel/draft switches, and edit-mode transitions because the composer instance is long-lived and the ref has no active-href cleanup.

Please prune `reenteringHrefsRef` when an href leaves the live/active candidate set (and ensure any associated generation/upload bookkeeping remains safe). Add a regression for blocked re-entry → remove URL → refetch succeeds while absent → paste URL again, proving the healthy result becomes sendable instead of remaining blocked.

tellaho:
@wesbillman The blocked re-entry removal defect is addressed at `e88b47c32`.

- The committed live-href layout boundary now prunes `reenteringHrefsRef` whenever a href leaves the draft, so a later paste can consume healthy metadata that resolved while the href was absent instead of inheriting the abandoned `blocked` phase.
- The same removal transaction advances the href’s upload generation. That prevents an upload started before removal from publishing into a later incarnation of the href, including the healthy-cache re-paste path that does not trigger another stale-fallback generation bump.
- `activeHrefsRef` now updates at that same committed boundary rather than during render, preserving the existing abandoned-render safety guarantee.
- Added the exact regression: transient fallback → blocked re-entry/refetch → remove → successful refetch while absent → paste again. It proves no third fetch occurs, the healthy snapshot tag becomes sendable, and pending clears.

Validation at exact pushed head `e88b47c32`: composer + resolver hook tests **29/29**, TypeScript clean, Biome clean, file-size ratchet clean. Push hooks passed `desktop-check`, `desktop-typecheck`, and full `desktop-test`.

Updated by Princess Donut 🤖

wesbillman (commented):

The lifecycle blocker is resolved at exact head `e88b47c32da56ba195f59271d41cbf67fe77b554`.

The layout-effect cleanup now removes abandoned `reenteringHrefsRef` phases when an href leaves and advances its upload generation, so pre-removal completions cannot leak into a later incarnation. The focused regression exercises the exact sequence I requested: blocked re-entry, removal before refetch completion, healthy metadata resolving while absent, then later paste becoming sendable without a third fetch.

I found no new actionable issue in the two-file repair. `git diff --check origin/main...HEAD` passes and GitHub reports the head mergeable. CI for this new head is still running. Per owner policy I am not auto-approving; my earlier changes-requested review can be dismissed once the desired checks finish.

wesbillman (approved):

PR #5515, opened 2026-08-10
feat(agents): add managed Nxtlinq authorization and setup

nxtlinqit:
## Summary

Adds an optional, managed integration with `@nxtlinq/authorization-gateway` for locally executed Buzz Agents, including an owner-reviewed setup lifecycle that can start from either an Agent conversation or Agent Edit.

- Installs the pinned Gateway and Attest tools in Buzz-managed application storage and verifies the managed package identity before use.
- Adds a structured `nxtlinq_setup` control-plane tool. An owned Agent may propose policy fields and explain why access is needed, but it cannot install packages, initialize keys, edit the manifest directly, select trust paths, or sign.
- Adds a Desktop setup wizard for project selection, Attest initialization, local trust, editable policy review, manifest diff acknowledgement, signing, verification, and Agent activation.
- Supports uninitialized projects. After explicit owner approval, Desktop runs the reviewed Attest initializer in protected staging, verifies the generated key pair, moves the private key into the system keychain or owner-only app-data fallback, and installs only public state in the project.
- Keeps required safeguards outside the editable proposal: Gateway audience, inert scope, sensitive-file exclusions, and the required Buzz MCP connection cannot be removed by Agent-generated policy.
- Shows the current and proposed manifests side by side and requires acknowledgement of the latest diff before apply or activation.
- Allows an owner to ask the originating Agent to regenerate a proposal. The current reviewed proposal and owner guidance are sufficient input; a denial from the existing policy must not block the trusted structured setup call.
- Adds safe install, reinstall/update, and uninstall controls for clean-user testing and recovery.
- Derives Gateway trust at process spawn from the setup-verified canonical executable path and digest instead of editable Agent environment variables or executable basename.
- Mediates supported ACP permission requests, Buzz Agent filesystem/shell tools, and Buzz-provided MCP operations through the authorization bridge. This closes the host-tool bypass where protecting only `AcpClient::handle_permission_request` would leave in-process tools outside the policy path.
- Treats Gateway policy error `-32041` as a deterministic denial: Buzz does not retry the same denied operation, while the Agent can still send a normal-language final response.
- Preserves existing permission behavior for Agents that are not launched through the trusted Nxtlinq wrapper.

The integration requires `@nxtlinq/authorization-gateway` 0.3.0 or newer.

## Owner flow

The setup can be opened in either of two ways:

1. Ask an owned managed Agent to configure or review Nxtlinq for an explicit absolute project path. The Agent submits a policy-only Desktop draft through the structured tool.
2. Open the Agent card, choose **Edit**, expand **Advanced**, and select **Set up Nxtlinq**. Buzz starts from a conservative policy when no Agent proposal exists.

Desktop then owns every privileged step:

1. Confirm the project and, when necessary, adopt it as the local Agent workspace.
2. Install the reviewed tools if they are absent.
3. Inspect Attest state; for a new project, explicitly initialize and protect the generated private key outside the Agent workspace.
4. Create or select Buzz-local trust and receipt storage.
5. Edit the permission proposal, inspect the side-by-side manifest diff, and acknowledge it.
6. Apply, sign with the owner-controlled key, verify through the Gateway, and enable the launch preset.

The Agent remains stopped after activation so the owner controls when the new policy takes effect.

## Security boundary

This PR intentionally implements a local deployment MVP: Buzz Desktop, the Gateway, protected workspace, trust bundle, and runtime execute on the same host. The local Buzz user is also the deployment operator.

The Agent never receives signing key bytes or paths. Private key material does not cross WebView IPC and is not placed in the project, observer draft, Agent configuration, MCP arguments, logs, or editable settings. Native signing holds the local Agent lifecycle lock and materializes a managed key only in an owner-only temporary location for the reviewed Attest invocation.

This is not an OS sandbox. A downstream runtime or third-party MCP server with an independent filesystem, process, or network channel remains outside this Gateway path unless that channel is separately brokered. Remote trust distribution and hard operator/end-user administrative separation remain follow-up work.

## Review guide

The setup extension is organized into four commits:

1. **Structured control plane and enforcement** — `buzz-agent`, `buzz-acp`, `buzz-cli`, and `buzz-dev-mcp` policy mediation and `nxtlinq_setup` request validation.
2. **Desktop native lifecycle** — tool discovery, Attest initialization, managed key storage, trust enrollment, manifest apply/sign/verify, and launch wrapper binding.
3. **Desktop review UI** — direct and conversational entry points, editable policy, locked safeguards, split diff acknowledgement, regeneration, and E2E coverage.
4. **Documentation** — local deployment contract, owner flow, trust/key responsibilities, testing, and known limits.

Important entry points:

- `crates/buzz-dev-mcp/src/nxtlinq_setup.rs`
- `crates/buzz-agent/src/authorization.rs`
- `desktop/src-tauri/src/commands/nxtlinq_authorization.rs`
- `desktop/src-tauri/src/commands/nxtlinq_authorization/initialization.rs`
- `desktop/src-tauri/src/commands/nxtlinq_authorization/policy.rs`
- `desktop/src/features/agents/ui/NxtlinqSetupReviewDialog.tsx`
- `docs/nxtlinq-authorization-gateway.md`

## Related issue

Relates to #2931. This builds on the `AcpClient::handle_permission_request` integration discussion while also covering supported Buzz Agent and Buzz-provided MCP tool paths that do not pass through that ACP callback.

## Testing

- `cargo test -p buzz-acp`: 744 unit tests and 9 lifecycle integration tests passed.
- `cargo test -p buzz-agent authorization`: 16 authorization tests passed.
- `cargo test -p buzz-dev-mcp nxtlinq_setup`: 14 structured setup tests passed.
- `cargo test --manifest-path desktop/src-tauri/Cargo.toml nxtlinq_authorization`: 17 native initialization/policy/signing tests passed.
- Desktop Nxtlinq/Agent-management unit selection: 61 tests passed.
- `pnpm --dir desktop check` passed, including file-size, scalable-text, and pubkey checks.
- `pnpm --dir desktop typecheck` passed.
- `pnpm --dir desktop build:e2e` passed.
- Nxtlinq setup and authorization Playwright smoke selection: 8 tests passed.
- `git diff --check` passed.

Manual clean-user testing with the published Gateway 0.3.0 covered installation, new-project initialization, managed trust creation, owner-controlled signing, allowed `README.md` reads, `.env` denial through both the file reader and shell fallback, a final natural-language denial response without retry loops, and subsequent authorized shell execution.


Silentpartnercoding:
Two narrow trust-path hardening items look worth closing in the current draft:

1. The managed installer currently runs `npm install -g @nxtlinq/authorization-gateway`, while the documented and tested contract is Gateway 0.3.0. Pinning and verifying the reviewed version would prevent the integration from changing underneath Buzz.
2. `AgentCommandWrapper::is_nxtlinq_gateway` currently grants the internal trusted-Gateway state from the executable basename. Binding that state to the preset-managed, resolved executable that passed the setup check would prevent a same-name substitute from receiving it.

I can prepare a Buzz-only slice with version-mismatch and executable-substitution fail-closed tests, without changing Gateway semantics or broadening the stated security boundary. Would you prefer that as a patch against this branch or as a separate focused PR?

nxtlinqit:
@Silentpartnercoding 

Thanks — both trust-path hardening items are now addressed.
The installer is pinned to @nxtlinq/authorization-gateway@0.3.0 and verifies the managed package name/version after installation.
Runtime trust is now bound to the setup-verified canonical executable path and SHA-256 digest. Basename-only detection has been removed, and version mismatch, digest mismatch, and same-name/symlink substitution all fail closed.
The changes and focused tests have been pushed to this PR.

nxtlinqit:
Updated this draft with the complete owner-reviewed setup lifecycle. The new commits add the structured Agent proposal tool, Desktop Attest initialization and managed-key ceremony, editable manifest review with locked safeguards and split diff acknowledgement, in-place Agent regeneration, and direct Agent Edit setup. The existing runtime enforcement boundary remains unchanged and now has end-to-end setup coverage. The branch was reorganized into four reviewable commits; focused Rust, Desktop, frontend, and 8 Playwright smoke tests pass. The PR description now documents the key boundary, local-MVP scope, review order, and deferred remote operator/end-user separation.

nxtlinqit:
Update: this PR now includes an owner-reviewed, conversational Nxtlinq setup flow.

### What changed

An Agent can now propose a minimum Nxtlinq authorization policy from chat and submit it to Buzz Desktop for owner review. The Agent can suggest configuration, but it cannot install the Gateway, modify the manifest, access signing keys, or activate the wrapper by itself.

The Desktop flow now supports:

- opening the same setup flow directly from the Agent UI or from an Agent-submitted draft;
- selecting and confirming the protected project workspace in Desktop;
- securely initializing an uninitialized project through the reviewed Attest CLI;
- generating the project key pair while keeping the private key outside the Agent workspace;
- storing the private key in the OS credential store, with an owner-only app-data fallback;
- automatically creating a local trusted-signers bundle from the newly generated and verified public key;
- showing an editable policy proposal and side-by-side manifest diff;
- requiring the owner to review the exact manifest revision before applying it;
- securely signing and verifying the manifest as a separate owner action;
- installing, reinstalling, or uninstalling the pinned Gateway package from Desktop;
- constructing and enabling the verified Agent wrapper without manual wrapper arguments;
- reopening and regenerating the policy review for an already protected Agent.

### Security boundary

The conversational Agent only produces a policy draft. All sensitive or state-changing actions remain native, explicit owner operations in Buzz Desktop.

In particular:

- private-key bytes and paths never enter the WebView, Agent, MCP server, observer event, settings, or logs;
- manifest application and signing are digest-bound to the reviewed revision;
- trusted Gateway behavior is enabled only for the exact Buzz-managed, version-pinned executable that passed readiness;
- filesystem, terminal, and MCP operations are authorized independently and fail closed;
- denied operations cannot be bypassed through the bundled shell or another unclassified MCP tool;
- the downstream runtime receives only explicitly allowlisted environment variable names.

### Main-branch integration

The implementation has also been updated for the latest `main` refactors, including the extracted managed-agent update module, batched observer processing, profile editing flow, and managed-agent wire types.

Current PR head: `0b14f8cb2`

Silentpartnercoding:
Had a look at the new setup lifecycle. The key handling reads well — `create_new` + `0o600` + `O_NOFOLLOW`, `0o700` on the directory, canonicalized and refused if it resolves inside the Agent workspace. Nothing to raise there.

One observation about where the owner-review gate sits, offered as defence-in-depth rather than a demonstrated bypass — I haven't traced an Agent-reachable path to these commands, and the UI does the right thing by clearing acknowledgement when a new draft arrives.

`apply_manifest_policy` compares `expected_sha256` against the *current* on-disk manifest, while the bytes written are recomputed from the `policy` argument at apply time. So the digest binds the state the change starts from rather than the result the owner reviewed, and a call carrying a still-valid `expectedSha256` with a different policy would write successfully. `sign_nxtlinq_manifest` follows the same shape.

The acknowledgement doesn't currently close that at the boundary: `applyManifest()` checks `if (!preview) return;`, and `diffAcknowledged` governs the control's disabled state, so the review step is enforced in the renderer. Neither Tauri command takes an acknowledgement argument.

`manifest_apply_is_bound_to_the_reviewed_digest` covers the case it names well — the manifest changing underneath a preview. The complementary case, where the file is untouched but the policy differs between preview and apply, isn't exercised yet, which is why this is easy to miss.

A possible shape, if it's useful: return `proposed_sha256` from preview alongside `current_sha256`, bind the owner's acknowledgement to that value, and have apply and sign take `expected_proposed_sha256` and re-verify it against the recomputed proposal before writing. That moves the "this is the manifest I approved" property to the same place the other trust decisions already live.

Happy to send this as a focused patch against the branch, with a test that varies the policy between preview and apply — only if that's helpful to you, and no rush either way.

PR #5516, opened 2026-08-10
fix(desktop): preserve explicit Claude executable overrides

marklyonx-star:
## Summary

- preserve an explicit `CLAUDE_CODE_EXECUTABLE` already supplied by the fully layered managed-agent environment
- keep automatic Claude CLI discovery as the fallback when no explicit launcher is configured
- add a regression test covering the agent-specific launcher precedence

The managed-agent spawn path applies `descriptor.env`, then calls `configure_runtime_cli`. Before this change, runtime discovery unconditionally replaced an agent-specific executable with the discovered `claude` binary, contradicting the documented environment precedence.

### Related issue

None found after searching existing issues and pull requests for this exact override behavior.

### Testing

- `cargo test --manifest-path desktop/src-tauri/Cargo.toml managed_agents::runtime::tests::` (61 passed)
- regression test confirmed red before the fix and green after the fix using temporary executable stubs only

### DIFF
--- a/desktop/src-tauri/src/managed_agents/runtime.rs
+++ b/desktop/src-tauri/src/managed_agents/runtime.rs
@@ -381,6 +381,14 @@ pub(crate) fn configure_runtime_cli(
     if runtime.id != "claude" {
         return;
     }
+    // Runtime discovery is a fallback. Preserve an explicit launcher supplied
+    // through the fully layered agent environment.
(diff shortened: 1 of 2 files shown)


marklyonx-star:
Superseded by #5517, which carries the required DCO sign-off without rewriting the published branch.

PR #5533, opened 2026-08-11
fix(acp): correct heartbeat reply guidance

### LINKED ISSUE #5532: buzz-acp: heartbeat feed mentions have no thread context — responses go unthreaded or silent
## Summary

Agents processing @mentions during **heartbeat turns** have no thread context and send unthreaded replies (or no replies at all), even though the triggering mention was in a thread. This is a distinct issue from #3825/#3908, which fixed the **normal dispatch path**.

## Root Cause

`dispatch_heartbeat` (`lib.rs:3923-3956`) calls `default_heartbeat_prompt` (`lib.rs:4012-4028`) with `batch=None`. The prompt instructs the agent to call `buzz feed get --types mentions`, but gives **no `--reply-to` instruction** for how to reply to those mentions.

The normal dispatch path injects `IMPORTANT: For ordinary replies in this turn, use --reply-to <root_event_id>` via `append_reply_instruction` (`queue.rs:1149-1156`), but that function is **never called** in the heartbeat path.

**Result:** When an agent finds an @mention in `buzz feed get` during a heartbeat, it has no thread root and either:
- Sends a reply without `--reply-to` (creating a new top-level message instead of threading), or
- Sends no reply at all

## Observed Symptom

Users see agent activity in the observer/activity feed (kind:24200 encrypted frames) showing the agent processed a turn, but **no channel message** appears in the thread.

Confirmed in channel COMPDE-454 — Osmosmith messages appeared in the activity feed but not in the thread for several mentions processed during heartbeat turns.

## Suggested Fix

Extend `default_heartbeat_prompt` (`lib.rs:4012-4028`) to explicitly instruct:

> For each actionable feed @mention:
> 1. Call `buzz messages thread --channel <h-tag> --event <event-id>` to resolve the thread root
> 2. Send your reply using `--reply-to <root>`

**Do NOT pre-resolve in the harness** — the harness does not know which feed items the agent will retrieve or act on. The instruction must be in the prompt.

If custom persona `heartbeat_prompt` overrides exist, the durable rule should also live in the shared base prompt to cover all heartbeat turns.

## Test Gap

- No test exercises the heartbeat path where an agent replies to a feed-sourced @mention and verifies `--reply-to` is present or instructed
- `test_format_prompt_single` tests normal dispatch but not the heartbeat code path

## PR Overlap Check

- **PR #3908** fixes top-level vs threaded replies in the **normal dispatch path** (`queue.rs`). It does NOT touch the heartbeat path.
- **PR #3183** propagates automatic reply context for normal batches but **explicitly clears it** for heartbeat turns. It does NOT fix feed-driven heartbeat replies.

## Affected Code

| File | Lines | Issue |
|------|-------|-------|
| `crates/buzz-acp/src/lib.rs` | 3923-3956 | `dispatch_heartbeat` — passes `batch=None`, no thread context injected |
| `crates/buzz-acp/src/lib.rs` | 4012-4028 | `default_heartbeat_prompt` — no `--reply-to` hint for feed-driven replies |
| `crates/buzz-acp/src/queue.rs` | 1149-1156 | `append_reply_instruction` only fires on normal dispatch, not heartbeat |

## Confidence: High

Confirmed against `origin/main@07a3c768` (2026-08-10) by Grumplestiltzkin.

## Related Issues

- #3825 — "Agent reply visible in Activity log but not displayed in channel/thread" (symptom)
- #3908 — Fixes the **normal dispatch path**, not the heartbeat path

## Research

Full investigation: `RESEARCH/AGT_01_AGENT_ACTIVITY_NOT_THREAD.md` in the Buzz workspace

cameronhotchkies:
Agents responding to feed mentions during heartbeat turns had no thread context,
causing replies to go unthreaded or silent. This adds explicit instructions to
`default_heartbeat_prompt` telling agents to reply directly to the feed mention
ID with `--reply-to`; the CLI resolves the thread root automatically.

Fixes https://github.com/block/buzz/issues/5532

Signed-off-by: Cameron Hotchkies <chotchkies@block.xyz>
refs: https://github.com/block/buzz/issues/5532

### DIFF
--- a/crates/buzz-acp/src/lib.rs
+++ b/crates/buzz-acp/src/lib.rs
@@ -4301,13 +4301,32 @@ fn default_heartbeat_prompt() -> String {
          1. Run `buzz feed get --types needs_action` to check for pending workflow approvals or\n\
             high-priority requests addressed to you.\n\
          2. Run `buzz feed get --types mentions` to check for unanswered @mentions.\n\
-         3. If you find actionable items, address them using the appropriate CLI commands\n\
-            (e.g., `buzz workflows approve --token <UUID>`, `buzz messages send`,\n\
-            `buzz messages send --reply-to <event-id>`).\n\
-         4. If there are no pending actions or mentions, end your turn immediately.\n\n\
+         3. For each actionable feed mention, reply directly with\n\
+            `buzz messages send --channel <channel-uuid-from-h-tag> --reply-to <feed-event-id> --content \"...\"`.\n\
+            The `--reply-to` flag resolves the thread root automatically.\n\
+         4. Address other actionable items using the appropriate CLI commands\n\
+            (e.g., `buzz workflows approve --token <UUID>`, `buzz messages send`).\n\
+         5. If there are no pending actions or mentions, end your turn immediately.\n\n\
          Do not run `buzz channels list` or `buzz messages search` unless you have a specific reason.\n\
          Do not invent work — only act on items surfaced by the feed commands."
     )
+}
+
+#[cfg(test)]
+mod heartbeat_prompt_tests {
+    use super::default_heartbeat_prompt;
+
+    #[test]
+    fn default_heartbeat_prompt_replies_directly_to_feed_mentions() {
+        let prompt = default_heartbeat_prompt();
+
+        assert!(prompt.contains("For each actionable feed mention"));
+        assert!(prompt.contains(
+            "buzz messages send --channel <channel-uuid-from-h-tag> --reply-to <feed-event-id> --content \"...\""
+        ));
+        assert!(prompt.contains("The `--reply-to` flag resolves the thread root automatically"));
+        assert!(!prompt.contains("buzz messages thread"));
+    }
 }
 
 /// Spawn a background respawn task for a crashed agent slot.


ravarora2:
🤖 LGTM. This fixes the right problem at the right layer — since the harness can't know which feed items the agent will act on, putting the "resolve thread root → `--reply-to <root>`" recipe in the prompt (rather than pre-resolving in the harness) is the correct call, and it lines the heartbeat path up with what the normal dispatch path already injects. Ran the new test locally and it passes:

```
test heartbeat_prompt_tests::default_heartbeat_prompt_resolves_thread_root_for_feed_mentions ... ok
```

Two nits, neither blocking:

1. **Custom `heartbeat_prompt` overrides won't get this guidance.** `dispatch_heartbeat` resolves the prompt as `ctx.heartbeat_prompt.clone().unwrap_or_else(default_heartbeat_prompt)`, so a persona/config that sets its own `heartbeat_prompt` fully replaces the default and bypasses this fix (the issue flagged this too). Worth deciding whether the thread-resolution rule should be appended to *all* heartbeat prompts in `dispatch_heartbeat`, or at least documented for custom-prompt authors. Fine to punt to a follow-up.
2. **Minor placeholder naming:** the prompt says `--channel <h-tag>` but the CLI flag is `--channel <uuid>`. Since the feed item's `h` tag *is* the channel UUID it's technically fine, but `--channel <channel-uuid>` might read a touch clearer to the model.

Good to merge from my side.

ravarora2 (dismissed):
lgtm, will wait for owners to review and merge

wesbillman (changes requested):
The goal is right, but this recipe adds an unnecessary and unreliable model-side root-resolution step.

`buzz messages thread` does not return a dedicated thread-root field. Its contract expects `--event` to be the root, queries events referencing that ID plus the event itself, and emits normalized raw events/tags (`crates/buzz-cli/src/commands/messages.rs:394-426`; `crates/buzz-cli/src/client.rs:1307-1325`). Given a nested feed mention, the agent must inspect the returned event's NIP-10 tags and choose the correct ID itself. That adds another relay request, potentially a large result (default 100), latency/tokens, and a new opportunity to choose the mention, a child, or another event and mis-thread/silence the reply.

The send command already performs this operation safely: `buzz messages send --reply-to <mention-event-id>` fetches that immediate parent and derives its root marker in `resolve_thread_ref` (`crates/buzz-cli/src/commands/messages.rs:48-85, 637-642`). The prompt should tell the agent to reply directly to the feed item's event ID:

```
buzz messages send --channel <channel-uuid-from-h-tag> --reply-to <feed-event-id> ...
```

That is both simpler and aligned with the CLI's actual API. Please also make the regression pin that direct recipe rather than only checking prompt substrings.

Separate coverage gap: custom `--heartbeat-prompt` / `--heartbeat-prompt-file` values fully replace this default in `dispatch_heartbeat` (`crates/buzz-acp/src/lib.rs:3936-3939` at this head), so they remain unfixed. If thread-safe feed replies are intended as a platform invariant, append a small durable instruction to both default and custom heartbeat prompts; otherwise explicitly document the override requirement.

wesbillman (changes requested):

The direction is right, but the recipe does not actually resolve the root it tells the model to use. `buzz messages thread --event` treats its argument as the root: it queries replies whose `e` tag references that ID and separately fetches that exact event. When the feed mention is itself a nested reply, passing the mention ID therefore returns that event/subtree; it does not emit a resolved root ID. The model is then told to send to the unexplained `<root>` placeholder, so this can still produce an incorrect reply target or no reply.

Please make the resolution algorithm explicit (inspect the mention event's NIP-10 tags: `root` marker if present, otherwise the `reply` marker for a direct reply, otherwise the mention ID for a top-level event), or add/use a CLI operation that actually returns the root. The test should pin that complete algorithm rather than only the current command strings.

wpfleger96:
🤖 Paul, an automated reviewer, commenting via Will's GitHub account.

One gap not yet raised in the existing reviews: the heartbeat feed poll has no durable handled-state or cursor, so this prompt (fixed threading or not) can act on the same mentions repeatedly.

`default_heartbeat_prompt` tells the agent to run `buzz feed get --types mentions` with no `--since`. Without it, `cmd_get_feed` just queries the latest p-tagged events, default limit 20 (`crates/buzz-cli/src/commands/feed.rs:16-26`), so every tick re-surfaces the same recent mentions — including ones already answered on a previous tick. The only thing preventing a duplicate reply is the model session's own memory of prior turns, and that isn't durable across session invalidation, harness restart, or context loss. The failure mode is an agent re-replying to old mentions after a restart.

This is architectural rather than something this PR's prompt wording can fix, so flagging it for tracking, not as a blocker here. The durable design would move the polling into the harness: persist a per-agent feed cursor (or handled-event set), run the feed query deterministically on each tick, and dispatch a model turn only when unseen actionable events exist — passing them as concrete event context, the way normal dispatch does. That eliminates both the duplicate-work risk and the token cost of waking the model on an empty feed.

cameronhotchkies:
🦾 Addressed the threading recipe feedback in `67e55d8550cdd6271140a2e55b168c7d166cb54c`:

- The default heartbeat prompt now tells agents to reply directly with `buzz messages send --channel <channel-uuid-from-h-tag> --reply-to <feed-event-id> --content "..."`.
- The prompt states that `--reply-to` resolves the thread root automatically, so agents do not call `buzz messages thread` first.
- The heartbeat prompt regression test now pins the direct `--reply-to <feed-event-id>` recipe and asserts the old `buzz messages thread` instruction is absent.

Left out of scope here: custom heartbeat prompt override handling and durable feed cursor/dedup behavior.

PR #5536, opened 2026-08-11
fix(desktop): show eligible remote member agents in mentions

JDiz00:
## Summary
- let real channel-member agents reach the directory-aware mention policy instead of rejecting them at the preliminary allow-list
- keep explicitly excluded agents hidden while allowing remote members whose directory state is unknown after queries settle
- avoid first-load flicker by withholding unknown member agents until managed and relay directories settle

### Related issue
None found.

### Testing
- `cd desktop && pnpm test` (4,588 passed)
- `cd desktop && pnpm build`
- `cd desktop && pnpm exec biome check src/features/agents/lib/agentAutocompleteEligibility.ts src/features/agents/lib/agentAutocompleteEligibility.test.mjs src/features/messages/lib/useMentions.ts`
- verified focused tests and the production build on macOS and Ubuntu
- manually verified the patched development build lists remote Ubuntu agent identities in both a channel picker and an exact-agent DM picker

Installed official signed apps were not replaced for final verification.

### DIFF
--- a/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
+++ b/desktop/src/features/agents/lib/agentAutocompleteEligibility.test.mjs
@@ -320,6 +320,62 @@ test("shouldHideAgentFromMentions: shows member agents with unknown invocability
   );
 });
 
+test("shouldHideAgentFromMentions: hides unknown member agents while directories load", () => {
+  assert.equal(
+    shouldHideAgentFromMentions({
+      isAgent: true,
+      isMember: true,
+      pubkey: PUB_A,
+      mentionableAgentPubkeys: new Set(),
+      directoryAgentPubkeys: new Set(),
+      directoryReady: false,
+    }),
+    true,
+  );
+});
+
+test("shouldHideAgentFromMentions: shows mentionable member agents while directories load", () => {
+  assert.equal(
+    shouldHideAgentFromMentions({
+      isAgent: true,
+      isMember: true,
+      pubkey: PUB_A,
+      mentionableAgentPubkeys: new Set([PUB_A]),
+      directoryAgentPubkeys: new Set(),
+      directoryReady: false,
+    }),
+    false,
+  );
+});
+
+test("shouldHideAgentFromMentions: shows non-agent members while directories load", () => {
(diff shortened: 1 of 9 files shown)


wesbillman (changes requested):
Reviewing on Wes's behalf.

**[P1] This bypass admits every directory-unknown bot member in the internal build, including other owners' and externally hosted `respond_to:anyone` agents.**

`useMentions.ts:259-275` now exempts every `candidate.isMember === true` from `isAgentIdentityInAllowedList`, and `shouldHideAgentFromMentions` returns visible once the directories settle without a kind:10100 row (`agentAutocompleteEligibility.ts:110-120`). Neither path checks the internal build's agent-access capability or the candidate's verified owner. A Block Desktop user can therefore select such an agent and emit its `p` tag. If that member is hosted outside Desktop with `respond_to:anyone`, it can answer; the local spawn restriction cannot protect an external runtime.

The internal packaging restriction is not a picker policy. Worse, on current `squareup/buzz-releases` main, `scripts/build-macos.sh:202` supplies `BUZZ_ACP_ALLOWED_RESPOND_TO=owner-only,allowlist` only to the build process, while `block/buzz/desktop/src-tauri/src/managed_agents/access_policy.rs:44-48` activates the durable compile-time owner-only capability only when `BUZZ_BUILD_AGENT_ACCESS_OWNER_ONLY` is set. I found no use of that marker in `buzz-releases`. This PR must not assume the packaging gate hides or blocks these candidates.

Please add an explicit build-capability-aware eligibility boundary and regressions proving that, in the internal owner-only build, a channel-member agent owned by another user and a directory-unknown agent are not offered or taggable. Keep the OSS headless-member behavior behind the unrestricted build path. Also preserve a regression for the #2149 stale member case; this patch currently leaves the existing E2E named `stale channel-member agents ... stay hidden` unchanged even though the new predicate makes directory-absent member agents visible after settling, so that behavioral contradiction needs to be resolved explicitly.

wesbillman:
Superseded by #5681. The follow-up closes the remaining edit-event and fresh ownership-proof races, includes the prior fixes, and is hosted on a Block branch maintainers can update.

PR #5537, opened 2026-08-11
fix(desktop): pass managed-agent owner attestation to git credential helper

Jory21M:
## Problem

Merging a pull request from Desktop fails with `403 restricted: not a relay member` whenever the target repository is owned by a **managed agent**. The failure happens during the merge's temporary clone, before any merge or push.

`merge_project_pull_request` resolves the repository owner's identity (`project_owner_identity`), which for a managed agent includes the NIP-OA owner attestation (`record.auth_tag`). The status-event path passes that attestation to `submit_signed_event_with_keys`, but the git path built its credentials with `build_git_auth_config_for_keys`, which only exports `NOSTR_PRIVATE_KEY`. `git-credential-nostr` already supports folding the attestation into the signed NIP-98 event via `BUZZ_AUTH_TAG` (it can't use a separate HTTP header, so the tag must ride in the signed event) — the desktop app just never handed it over. The relay therefore sees a bare agent key, refuses it as a non-member, and every Desktop merge of an agent-owned repo fails.

## Fix

- Add an optional `auth_tag` to `GitAuthConfig`, exported as `BUZZ_AUTH_TAG` alongside `NOSTR_PRIVATE_KEY` when git needs credentials.
- Add `build_git_auth_config_for_identity(keys, auth_tag)` and use it at the merge call site with the attestation from `ProjectOwnerIdentity`.
- All other builders default the tag to `None`, so viewer-identity flows are unchanged.

## Reproduction

1. Create a repo owned by a managed agent on a relay that enforces NIP-OA membership for agent keys.
2. Open a PR on it and click **Merge** in Desktop.
3. Before this patch, the temporary clone fails with `403 restricted: not a relay member`; after it, the merge completes.

## Testing

- `just desktop-tauri-test` — all suites pass, including two new unit tests: `credentialed_git_receives_owner_attestation_when_present` and `credentialed_git_omits_attestation_when_absent`.
- `just desktop-tauri-clippy` and `cargo fmt --check` clean.

### DIFF
--- a/desktop/src-tauri/src/commands/project_git_exec.rs
+++ b/desktop/src-tauri/src/commands/project_git_exec.rs
@@ -48,6 +48,7 @@ pub(crate) struct GitAuthConfig {
     git_path: std::path::PathBuf,
     credential_helper: Option<std::path::PathBuf>,
     nsec: String,
+    auth_tag: Option<String>,
     allow_file_transport: bool,
 }
 
@@ -173,6 +174,13 @@ fn configure_git_auth(command: &mut Command, auth: &GitAuthConfig, needs_credent
             return apply_git_config(command, &entries);
         };
         command.env("NOSTR_PRIVATE_KEY", &auth.nsec);
+        // A managed-agent identity is only a relay member together with its
+        // NIP-OA owner attestation; without it the relay refuses the key with
+        // `403 restricted: not a relay member`. git-credential-nostr reads the
+        // attestation from BUZZ_AUTH_TAG and folds it into the NIP-98 event.
+        if let Some(auth_tag) = &auth.auth_tag {
+            command.env("BUZZ_AUTH_TAG", auth_tag);
+        }
         entries.push((
             "credential.helper",
             credential_helper_config_value(cred_helper),
@@ -213,6 +221,7 @@ pub(crate) fn build_git_clone_auth_config(
                 .ok_or_else(|| "git was not found on PATH".to_string())?,
             credential_helper: None,
             nsec: String::new(),
+            auth_tag: None,
             allow_file_transport: false,
         });
     }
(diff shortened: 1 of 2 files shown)


ravarora2:
🤖 Tested and code LGTM.

Verified the fix across every code hop in the auth chain (no live infra needed):

1. **Desktop → git command** — the PR's two unit tests (`credentialed_git_receives_owner_attestation_when_present` / `credentialed_git_omits_attestation_when_absent`) confirm `BUZZ_AUTH_TAG` is set on the git subprocess when an attestation is present and omitted otherwise. ✅
2. **`git-credential-nostr` → signed event** — black-box ran the helper as git would (stdin credential request + `BUZZ_AUTH_TAG` set), decoded the emitted base64 NIP-98 event (kind 27235), and confirmed the `["auth", owner, conditions, signature]` tag is folded into the signed event; absent without the env var. ✅
3. **Relay → owner resolution** — `buzz-relay` unit tests (`valid_nip_oa_returns_owner`, `no_auth_tag_returns_none`, `invalid_auth_tag_returns_none`) confirm a properly-signed tag resolves the owner and absent/invalid tags are rejected. ✅

The human-owner path is proven unchanged (no stray env var when `auth_tag` is `None`). The only thing not exercised here is the live HTTP round-trip (relay granting membership + clone returning 200 instead of 403), which needs the full relay + agent-owned-repo setup — but every piece of logic on that path is now covered. Small, targeted, low-risk change. LGTM.

ravarora2 (approved):
lgtm, leaving for owners to have a final review

wesbillman (changes requested):

**Blocking: clear an inherited `BUZZ_AUTH_TAG` when this auth config intentionally has no attestation.**

`configure_git_auth` conditionally sets `BUZZ_AUTH_TAG` for `Some`, but does nothing for `None`. `std::process::Command` inherits the parent environment by default, so a Desktop process launched with an ambient `BUZZ_AUTH_TAG` passes that unrelated tag to every credentialed git subprocess whose builder deliberately selected `None` (viewer identity, normal clone auth, etc.). This is not theoretical in the managed-agent runtime: `BUZZ_AUTH_TAG` is a process-level authentication input there.

The credential helper gives the environment precedence over `nostr.authtag` and parses any present value. The consequences are both reliability and authority isolation failures:

* a stale/malformed ambient tag makes otherwise valid ordinary git authentication fail closed;
* a valid ambient attestation is silently signed into a NIP-98 event despite the caller selecting `auth_tag: None`, crossing the identity/config boundary this field is supposed to enforce.

The new `credentialed_git_omits_attestation_when_absent` test does not catch this. `Command::get_envs()` exposes only explicit overrides/removals, not inherited process variables, so `None` passes even though a spawned child can still receive the ambient value.

Please call `command.env_remove("BUZZ_AUTH_TAG")` before the conditional `env` assignment (analogous to clearing other sensitive inherited state), then add a regression that seeds the parent environment and proves a spawned child cannot observe the variable for `None`. Serialize that test or use a subprocess fixture so mutating process environment cannot race parallel tests.

PR #5541, opened 2026-08-11
feat: add channel guest onboarding with community agents

arieldiaz:
## Summary

- add consumer-friendly channel-scoped guest invitations
- let community owners and admins manage guest invite links
- preserve dynamic invite previews and native invite requests
- make invited members inherit the community agent roster instead of provisioning Fizz, Honey, and Bumble locally

## Why

An invited Humanware OS member should enter the community they were invited to. The previous onboarding path treated every invited member like the creator of a new Buzz workspace, which leaked Buzz's generic starter agents into a community that already has Liv and Max.

## Verification

- `pnpm typecheck`
- targeted starter-team policy tests: 2 passing
- Biome check on changed onboarding files


alm36918 (commented):
Context: we're a second independent integrator hitting exactly the motivation described here. We run hosted Buzz communities for small Brazilian organizations, with our own agent roster (e.g. "Cira") provisioned server-side before the first human joins. Current behavior is actually worse for us than the trio leak: since our communities have no personas named Fizz, Honey or Bumble, the "Meet your starter team" stage filters `listPersonas()` down to nothing and renders an empty stage under that headline. Skipping the stage entirely for invite joins — as this PR does — is the right call, and simpler than making the stage dynamic. +1 on the approach.

Two questions from reading the diff against current main:

1. **Second `team-intro` transition looks unguarded.** `shouldProvisionLocalStarterTeam` gates the profile-continue handler, but the profile-discovery effect (`CommunityOnboardingFlow.tsx` ~L308–324 on main) also transitions: when `getProfile()` reports `hasProfileEvent`, it fires `update({ stage: "team-intro" }, transaction.id)` directly. An invitee with an existing relay profile (profile reuse after #2155, re-onboarding scenarios like #3727, a second machine) would still land on team-intro through that path. Should that site apply the same policy — i.e. `finish()` instead of `team-intro` when an invite code is present?

2. **`WelcomeKickoffStage` still hardcodes the trio.** `STAGE_CHARACTERS` renders the Fizz/Honey/Bumble APNGs on the Welcome banner while waiting for the first agent message. On an invite join into a community with a custom roster, that first message comes from a different agent. Should this surface be gated by the same policy (or driven by the actual roster)? Possibly out of scope for this PR, but it's the same leak on an adjacent surface.

Happy to verify against our integration once this lands — the use case in the description matches ours 1:1.

Issue #4658, opened 2026-08-04
Mentioning an agent from another machine never wakes it — agent start on mention is sender-side only


lamcode9:
### Setup

- Machine A ("host", always on): runs Buzz Desktop and manages/hosts all agents.
- Machine B (chat client): same user, same relay, no locally managed agents.
- All agents are `role=bot` channel members on the relay. Desktop v0.5.4.

### What happens

Mentioning an agent from machine B never starts it on machine A. Only agents that are already running on A (e.g. via start-on-app-launch) respond; a managed-but-stopped agent stays silent with no error anywhere.

### Why (as far as I can tell from the source)

Waking an agent on mention is implemented **sender-side only**: `desktop/src/features/messages/ui/useMentionSendFlow.ts` resolves mentioned pubkeys against the *sender's* managed-agents list and calls the start mutation for those. I could not find any host-side path (in `desktop/src-tauri/src/managed_agents/` or the crates) that starts a stopped managed agent when a mention for it arrives from the relay. So in any split topology — the same one as #3609 — mentions from other machines only reach agents that happen to be warm.

This also bites subtly on a single machine: an agent with start-on-launch disabled answers when mentioned locally (sender-side start kicks in), which makes the multi-machine silence look random.

### Workaround

Enable start-on-app-launch for every agent on the hosting machine.

### Suggestion

Have the hosting app subscribe to mentions of its own managed agents and start them on demand — that makes wake behavior symmetric with local sends. If that's out of scope, documenting start-on-launch as required for multi-machine setups would already help.

Related: #3609 (mention picker filtered to locally-hosted agents — same split-hosting topology).

Issue #4535, opened 2026-08-03
Multi-device: mention picker cannot reach remotely-hosted agents (owner-only unhandled, directory never updated, replies never p-tag authors)


Matweiss:
**Setup:** one user, two devices, same key. Device A's Buzz Desktop hosts the agents; device B is a pure client (same identity imported). Community = the user + their agents.

**Symptom:** on device B the @ picker offers no agents at all (only the user's own name). Plain-typed @names and thread replies publish but notify nobody, and agents wake only on p-tagged events (buzz-acp shipped default `subscribe = "mentions"` with `require_mention = true`, `crates/buzz-acp/src/config.rs`) — so device B has no channel path to reach any agent. (Deployments can override via `BUZZ_ACP_SUBSCRIBE` / `BUZZ_ACP_NO_MENTION_FILTER`; this issue describes the code default.)

**Defect 1 — eligibility:** `desktop/src/features/agents/lib/agentAutocompleteEligibility.ts` — `relayAgentIsSharedWithUser()` handles `respondTo` `allowlist`/`anyone` only; no `owner-only` branch, and `RelayAgent` carries no `ownerPubkey` (the kind:10100 card has no owner field), so owner-only agents are hidden even from their own owner on every non-hosting device.

**Defect 2 — settings cannot propagate:** Desktop saves respond-to edits into the owner-side kind:30177 managed-agent record, but the picker reads kind:10100 cards — and repo-wide the only kind:10100 writer is `buzz-cli channels set-add-policy` (content = `{"channel_add_policy"}` only). No code path ever writes or updates `respond_to`/`channel_ids` on the card, so switching an agent to Anyone/Selected-people can never make it picker-visible on another device.

**Defect 3 — replies never p-tag the parent author:** `thread_tags()` in `crates/buzz-sdk/src/builders.rs` emits only NIP-10 `e`-tags; Desktop's `messageMentionPubkeys.ts` states "Stream messages notify only explicit mentions." So "reply to a message to reach its author" is impossible for every client, which removes the natural fallback for defects 1-2.

**Defect 4 — misleading dialog copy:** the Selected-people edit dialog tells the owner "you don't need to add yourself" (`desktop/src/features/agents/ui/RespondToField.tsx:334`), but picker eligibility checks the literal published allowlist — following the dialog makes the fix silently fail even once defect 2 is repaired.

[...]

PR #5546, opened 2026-08-11
fix: publish kind:10100 agent directory records so shared agents are @-mentionable

### LINKED ISSUE #3809: Desktop @-mention autocomplete only shows locally managed agents

### LINKED ISSUE #4489: Mention autocomplete hides agents that are channel members but managed by another install
**Setup.** Two people, two Macs, one shared channel. One person runs agents Scout and Emmy on their Mac. The other runs agent Prinny on theirs. All three publish kind:0 profiles, are members of the channel with role `bot`, and are configured `respond_to: anyone`.

**Expected.** Each person can `@` any agent in the channel.

**Actual.** Each person's mention autocomplete offers only the agents managed by their own install. Neither can reach the other's agents from the GUI at all, because a mention with no resolved pubkey carries no `p` tag, and the `p` tag is what wakes an agent.

**Cause.** In `desktop/src/features/messages/lib/useMentions.ts:249`, every candidate passes through:

```ts
  return;
}
```

`managedAgentPubkeys` is derived solely from `useManagedAgentsQuery()` (`useMentions.ts:163`), which calls the `list_managed_agents` Tauri command. That command reads local disk records only (`desktop/src-tauri/src/commands/agents.rs:524` -> `load_managed_agents`). So any candidate with `isAgent === true` whose pubkey is not a locally managed agent is dropped before anything else runs.

This appears intentional: `agentAutocompleteEligibility.test.mjs:139` asserts it as "keeps people and only current managed agent identities".

**Why it looks worth revisiting.** The permissive path for exactly this case already exists and is currently unreachable for remote agents:

- `getMentionableAgentPubkeys` (`agentAutocompleteEligibility.ts:33`) already unions locally managed agents with relay-directory agents that pass `relayAgentIsSharedWithUser` (`respond_to: anyone` plus a shared channel, or an allowlist naming the current user).
- `shouldHideAgentFromMentions` (`agentAutocompleteEligibility.ts:67`) already has a documented "Option B" branch for member agents, showing them unless the kind:10100 directory explicitly excludes the current user.
- `useMentions.ts:334` iterates `relayAgentsQuery.data` and adds those agents as candidates.

All three are gated out by the managed-list check that runs first, so a remote agent can never reach them.

**Proposed change.** Let the managed-list check admit agents that `getMentionableAgentPubkeys` already considers invocable, i.e. run the eligibility test before the ownership test rather than after. Concretely, drop an agent identity only when it is neither locally managed nor in `mentionableAgentPubkeys`. The invocability signal (`respond_to` plus shared channels, via kind:10100) is the right authority for this, and it is already computed.

**Impact.** Without this, multi-person multi-agent collaboration in a shared channel only works if every human relays through their own agent. That works, but it means a human can never directly address a teammate's agent.

**Verified against:** upstream `main` @ b1b283c, desktop 0.5.3.

Related: #4487 / #4488 (same two-Mac setup, separate `respond_to` propagation bug).

### LINKED ISSUE #4548: Non-owner members never get agents in desktop/web mention autocomplete (kind-10100 records present and correct; Agent Directory renders them)
## Summary

Community members (non-owners) never see agents in the desktop/web mention autocomplete, even when kind-10100 agent-profile records exist on the relay with a `respond_to_allowlist` that includes the viewer. The **Agent Directory** on the Agents tab renders the same records correctly for the same viewer — only the composer autocomplete fails. Mobile is unaffected (it resolves mentions from channel members).

## Environment

- Hosted community (`*.communities.buzz.xyz`), remote agents per `docs/remote-agents.md` (systemd launcher, `buzz-acp` harness, NIP-OA auth tags)
- Reproduced on desktop 0.4.26 AND a current 0.5.x client (two different users/machines)
- Agents are members of the channels in question, with published kind-0 profiles (carrying valid NIP-OA `auth` tags) and kind-10100 records

## What we established

1. **The desktop app (0.4.26 at least) never publishes kind-10100 records** for provider-backed agents. A relay dump (`REQ kinds:[10100]`, NIP-42-authed member) returned zero events in a community with 6 active agents. Owner-side autocomplete works anyway via the local managed-agents registry — so the gap is invisible to the agent owner and hits only other members.
2. We published complete records ourselves (agent-signed over WS; schema per `nostr_convert.rs::agents_from_events`): `name`, `display_name`, `agent_type`, `status`, `channels`, `channel_ids`, `respond_to: "allowlist"`, `respond_to_allowlist` (including the test viewer), `channel_add_policy`. Accepted (`OK true`), and readable by a non-owner member over an authenticated REQ.
3. After that: Agent Directory lists the agents for the non-owner viewer ✓ — but the composer `@` autocomplete still offers only humans ✗. Per source (`useMentions.ts` + `agentAutocompleteEligibility.ts`), the allowlist branch of `relayAgentIsSharedWithUser` should admit the viewer and `shouldHideAgentFromMentions` should show invocable agents, so the data and the visible code both say this should work.

## Expected

A member on an agent's `respond_to_allowlist` (or `anyone` with a shared channel) should get that agent in mention autocomplete — matching the directory, and matching what the harness would actually accept.

## Impact

Remote/provider-backed agents (the architecture `docs/remote-agents.md` formalizes) are effectively single-player on desktop/web: only their owner can mention them. Teams fall back to mobile or to replying inside agent threads.

Happy to provide the record dumps, event IDs, or run diagnostics against our community.

### LINKED ISSUE #5363: Relay-discovered agent with respond_to=anyone still never appears in @-mention autocomplete (regression/incomplete case of #4913)

jjalora:
## Summary

Non-owners can never `@`-mention a shared agent, no matter what the owner configures — the root cause is that **the kind:10100 agent directory the mention-eligibility pipeline reads has no producer**. #4913 implemented and e2e-pinned the consumer side (`respond_to` + exact-channel authorization from `RelayAgent` records), but nothing anywhere publishes those records: the only kind:10100 writer in the tree is `buzz channels set-add-policy`, which writes a policy-only body (and, being replaceable, clobbers anything richer — #3663/#4490). On a real relay, `list_relay_agents` returns nothing usable, `getMentionableAgentPubkeys` collapses to the viewer's own managed agents, and the feature only appears to work in e2e because the mock bridge synthesizes directory entries (`syncMockRelayAgentsFromManagedAgents`) with no production counterpart.

Fixes #5363, fixes #4548 (commit 1); fixes #4489, fixes #3809 (commit 2). Also relevant to #2603 (composer half; the MembersSidebar add-gate is untouched), #4128, #4776, #3125, #3739, and the directory-staleness aspect of #4535.

## Commit 1 — `feat(acp)`: the harness publishes its own directory record

`buzz-acp` now publishes and maintains the agent's kind:10100 record. The harness is the right producer because it knows the `respond_to` policy it actually **enforces** and the channels it actually **listens on** — the advertisement can never drift from enforcement (a deployment that clamps `--respond-to` to owner-only also clamps what gets advertised, so records never promise an authorization the author gate would refuse).

- Publishes at startup (after channel subscriptions, same durable-readiness boundary as presence), debounced (2 s, abort-and-respawn like the presence heartbeat) on membership add/remove, and best-effort flips `status` to `offline` on graceful shutdown.
- Content matches exactly what `agents_from_events` → `RelayAgentInfo` → `RawRelayAgent` parse: `name` (from `session_title`, omitted when absent so the npub fallback engages), `agent_type` (normalized agent command identity), `status`, `respond_to` (the `RespondTo` `Display` literals), `respond_to_allowlist` (sorted), index-aligned `channels`/`channel_ids`, `capabilities`, `channel_add_policy`.
- `channel_add_policy` is carried forward from the current head record before every publish, so a policy set via `buzz channels set-add-policy` survives the republish and the relay's kind:10100 side effect (`handle_agent_profile`) stays satisfied. On a failed head fetch the publish is skipped entirely — never fabricate a policy.
- DM-typed and unresolved-type channels are excluded from the advertised set: the author gate rejects non-owner DMs regardless of mode and fails closed on unresolved types, so advertising them would invite mentions the harness then silently drops.
- Everything is best-effort: failures are logged and swallowed; publishing can never take the harness down. HTTP bridge (`RestClient::submit_event`) rather than the WS path, which silently drops non-observer events while disconnected.

With records published, **stock** desktop and mobile clients admit the agent through the already-pinned consumer chain (`relay-only anyone agents are visible when a channel is shared`, `relay-only shared agents emit an outbound mention tag when selected`, allowlist specs). Unit tests pin the producer schema; a new integration test (`desktop/src-tauri/tests/agent_directory_schema.rs`) pins the cross-crate contract from the consumer side.

## Commit 2 — `fix(desktop)`: member agents with unknown invocability stay mentionable

`addCandidate` ran `isAgentIdentityInAllowedList` **before** `shouldHideAgentFromMentions`, making the latter's documented member-agent rule ("unknown invocability ⇒ show" — the Option B comment, and its unit tests) unreachable dead code: any member agent absent from both the managed list and the directory was dropped by the first check.

This commit removes the redundant first check at that call site. `shouldHideAgentFromMentions` already implements the intended policy: invocable ⇒ show; non-member non-invocable ⇒ hide; member with an explicit not-invocable directory record ⇒ hide; member with **no** record ⇒ show. The last case is the useful fallback while a fleet's harnesses haven't (yet) published records — and once they have (commit 1 publishes for every mode, including `owner-only`), non-invocable member agents gain an explicit exclusion record and correctly disappear, so the stale-agent concern that motivated the stricter gate is answered by data rather than by hiding every foreign member agent. One e2e expectation is updated accordingly (`stale channel-member agents … stay hidden` → `… stay mentionable`).

This commit is separable — if maintainers prefer to keep the fail-closed member behavior from #4913, commit 1 alone fixes the shared-agent flow for record-publishing harnesses.

## Testing

- `cargo test -p buzz-acp` (incl. 6 new `directory` unit tests: consumer-schema shape, `respond_to` literals, name-key omission, allowlist ordering, dm/unknown channel exclusion, add-policy carry-forward/defaults)
- New `desktop/src-tauri/tests/agent_directory_schema.rs` integration test (consumer-side contract pin)
- `desktop`: `tsc --noEmit`, biome, full unit suite, and the full `mentions.spec.ts` e2e suite (58 specs) with the one updated expectation
- `cargo clippy -p buzz-acp --all-targets -- -D warnings`, `cargo fmt --check`
- Manually verified end-to-end on a self-hosted relay: owner sets *Anyone* → harness publishes the record → a second account sees the agent in `@` autocomplete in the shared channel → mention carries the `p` tag → agent responds. (This deployment shape — a 24/7 host running shared community agents — is what surfaced the bug.)

## Notes / non-goals

- `buzz channels set-add-policy` still replaces the whole record (#3663/#4490); this PR makes the damage self-healing (next harness publish restores the full record, policy carried forward) but doesn't change the CLI.
- Waking a *stopped* foreign agent on mention (#4658) and DM invocation of foreign agents (author-gate policy, pinned by test) are out of scope.
- Directory records intentionally advertise only non-DM, resolved-type channels; `status` is cosmetic to eligibility (presence remains the liveness signal).

### DIFF
--- a/crates/buzz-acp/src/directory.rs
+++ b/crates/buzz-acp/src/directory.rs
@@ -1,0 +1,385 @@
+//! Kind:10100 agent-directory record publisher.
+//!
+//! The harness advertises the `respond_to` policy it actually enforces and the
+//! channels it actually listens on, so member clients can decide whether this
+//! agent is @-mentionable (`agentAutocompleteEligibility` in the desktop app,
+//! `agent_identity_provider` on mobile). Publishing from the harness — rather
+//! than the owner's client — guarantees the advertisement can never drift from
+//! enforcement: a deployment that clamps `--respond-to` also clamps what gets
+//! advertised.
+//!
+//! The record is a replaceable event authored by the agent key; consumers
+//! (`agents_from_events` in the desktop Tauri backend) treat the event author
+//! as the authoritative pubkey and parse the content as a flat JSON object.
+//! `channel_add_policy` is carried forward from the current head record so a
+//! policy set via `buzz channels set-add-policy` survives our republish (the
+//! relay's kind:10100 side effect also requires the field).
+
+use std::time::Duration;
+
+use buzz_core::kind::{KIND_AGENT_PROFILE, KIND_NIP29_GROUP_METADATA};
+use nostr::{Alphabet, EventBuilder, Filter, Keys, Kind, SingleLetterTag};
+use serde_json::{json, Value};
+use tokio::task::JoinHandle;
+use uuid::Uuid;
+
+use crate::config::RespondTo;
+use crate::relay::{merge_discovered_channels, ChannelInfo, RestClient};
+
+/// Trailing-edge debounce for membership-churn republishes.
+pub(crate) const DIRECTORY_DEBOUNCE: Duration = Duration::from_secs(2);
+/// Bound on each relay round-trip (head fetch, metadata fetch, submit). Wide
+/// enough for the RestClient's internal transient-error retries to complete.
+const DIRECTORY_STEP_TIMEOUT: Duration = Duration::from_secs(10);
+/// Fallback add-policy when no head record exists — matches the relay's
+/// database default for users without a kind:10100 record.
+const DEFAULT_CHANNEL_ADD_POLICY: &str = "anyone";
+
+/// Static identity and policy inputs for directory publishes, built once at
+/// startup. Cheap to clone into spawned publish tasks.
+#[derive(Clone)]
+pub(crate) struct DirectoryPublisher {
+    /// HTTP bridge client (NIP-98 signed as the agent).
+    pub rest: RestClient,
+    /// Agent keys — the record must be authored by the agent itself.
+    pub keys: Keys,
+    /// The author gate mode this harness enforces.
+    pub respond_to: RespondTo,
+    /// Allowlist pubkeys (relevant when `respond_to` is `allowlist`).
+    pub respond_to_allowlist: Vec<String>,
+    /// Normalized harness/runtime identity (e.g. `claude-agent-acp`).
+    pub agent_type: String,
+    /// Display name for the record; `None` omits the key so consumers fall
+    /// back to the agent's npub / kind:0 profile.
+    pub name: Option<String>,
+}
+
+impl DirectoryPublisher {
+    /// Spawn a debounced, best-effort publish of the directory record.
+    ///
+    /// Callers hold at most one outstanding handle and abort it before
+    /// spawning a replacement (trailing-edge debounce, same pattern as the
+    /// presence heartbeat task).
+    pub(crate) fn spawn_publish(
+        &self,
+        channel_ids: Vec<Uuid>,
+        status: &'static str,
+        debounce: Duration,
+    ) -> JoinHandle<()> {
+        let publisher = self.clone();
+        tokio::spawn(async move {
+            if !debounce.is_zero() {
+                tokio::time::sleep(debounce).await;
+            }
+            publisher.publish_once(channel_ids, status).await;
+        })
+    }
+
+    /// One publish attempt: carry forward `channel_add_policy` from the head
+    /// record, resolve channel names, sign, submit. Every failure is logged
+    /// and swallowed — directory publishing must never take the harness down.
+    pub(crate) async fn publish_once(&self, channel_ids: Vec<Uuid>, status: &str) {
+        let channel_add_policy = match self.fetch_head_add_policy().await {
+            Ok(policy) => policy,
+            Err(e) => {
+                // Never overwrite a policy we could not read — skip entirely.
+                tracing::warn!("directory publish skipped (head fetch failed): {e}");
+                return;
+            }
+        };
+
+        let channels = if channel_ids.is_empty() {
+            Vec::new()
+        } else {
+            match self.resolve_channels(channel_ids).await {
+                Ok(map) => advertisable_channels(map),
+                Err(e) => {
+                    tracing::warn!("directory publish skipped (channel metadata failed): {e}");
+                    return;
+                }
+            }
+        };
+
+        let content = build_directory_content(
+            self.name.as_deref(),
+            &self.agent_type,
+            status,
+            &self.respond_to,
+            &self.respond_to_allowlist,
+            &channels,
+            &channel_add_policy,
+        );
+
+        let event =
+            match EventBuilder::new(Kind::Custom(KIND_AGENT_PROFILE as u16), content.to_string())
+                .tags([])
+                .sign_with_keys(&self.keys)
+            {
+                Ok(event) => event,
+                Err(e) => {
+                    tracing::warn!("directory record sign failed: {e}");
+                    return;
+                }
+            };
+
+        match tokio::time::timeout(DIRECTORY_STEP_TIMEOUT, self.rest.submit_event(&event)).await {
+            Ok(Ok(_)) => tracing::info!(
+                status,
+                channels = channels.len(),
+                respond_to = %self.respond_to,
+                "published kind:10100 directory record"
+            ),
+            Ok(Err(e)) => tracing::warn!("directory record publish failed: {e}"),
+            Err(_) => tracing::warn!("directory record publish timed out"),
+        }
+    }
+
+    /// Fetch the agent's current head kind:10100 record and extract its
+    /// `channel_add_policy`, defaulting when no record (or no field) exists.
+    async fn fetch_head_add_policy(&self) -> Result<String, String> {
+        let filter = Filter::new()
+            .kind(Kind::Custom(KIND_AGENT_PROFILE as u16))
+            .author(self.keys.public_key())
+            .limit(1);
+        let head = tokio::time::timeout(DIRECTORY_STEP_TIMEOUT, self.rest.query(&[filter]))
+            .await
+            .map_err(|_| "head query timed out".to_string())?
+            .map_err(|e| e.to_string())?;
+        Ok(parse_add_policy(&head))
+    }
+
+    /// Resolve channel names/types via kind:39000 metadata, reusing the
+    /// discovery-time merge (which also drops archived channels).
+    async fn resolve_channels(
+        &self,
+        channel_ids: Vec<Uuid>,
+    ) -> Result<std::collections::HashMap<Uuid, ChannelInfo>, String> {
+        let d_tag = SingleLetterTag::lowercase(Alphabet::D);
+        let d_values: Vec<String> = channel_ids.iter().map(|u| u.to_string()).collect();
+        let filter = Filter::new()
+            .kind(Kind::Custom(KIND_NIP29_GROUP_METADATA as u16))
+            .custom_tags(d_tag, d_values);
+        let meta = tokio::time::timeout(DIRECTORY_STEP_TIMEOUT, self.rest.query(&[filter]))
+            .await
+            .map_err(|_| "metadata query timed out".to_string())?
+            .map_err(|e| e.to_string())?;
+        Ok(merge_discovered_channels(channel_ids, &meta))
+    }
+}
+
+/// Reduce discovered channels to the advertisable `(id, name)` set, sorted by
+/// id for deterministic output.
+///
+/// DM channels are excluded (non-owner DM invocation is rejected by the author
+/// gate regardless of `respond_to`), and so are channels whose type could not
+/// be resolved — the author gate fails closed on unresolved types, so
+/// advertising them would invite mentions the harness then silently drops.
+fn advertisable_channels(map: std::collections::HashMap<Uuid, ChannelInfo>) -> Vec<(Uuid, String)> {
+    let mut channels: Vec<(Uuid, String)> = map
+        .into_iter()
+        .filter(|(_, info)| info.channel_type != "dm" && info.channel_type != "unknown")
+        .map(|(id, info)| (id, info.name))
+        .collect();
+    channels.sort_by_key(|(id, _)| *id);
+    channels
+}
+
+/// Extract `content.channel_add_policy` from the newest head event in a
+/// `/query` response, defaulting to the relay's database default.
+fn parse_add_policy(head_events: &Value) -> String {
+    head_events
+        .as_array()
+        .and_then(|events| events.first())
+        .and_then(|event| event.get("content"))
+        .and_then(Value::as_str)
+        .and_then(|content| serde_json::from_str::<Value>(content).ok())
+        .and_then(|content| {
+            content
+                .get("channel_add_policy")
+                .and_then(Value::as_str)
+                .map(str::to_string)
+        })
+        .unwrap_or_else(|| DEFAULT_CHANNEL_ADD_POLICY.to_string())
+}
+
+/// Build the kind:10100 content object exactly as the desktop consumer
+/// (`agents_from_events` → `RelayAgentInfo` → `RawRelayAgent`) parses it.
+///
+/// The allowlist is re-sorted and `channels`/`channel_ids` stay index-aligned;
+/// `name` is omitted (not null) when absent so the consumer's npub fallback
+/// engages.
+fn build_directory_content(
+    name: Option<&str>,
+    agent_type: &str,
+    status: &str,
+    respond_to: &RespondTo,
+    respond_to_allowlist: &[String],
+    channels: &[(Uuid, String)],
+    channel_add_policy: &str,
+) -> Value {
+    let mut allowlist: Vec<&str> = respond_to_allowlist.iter().map(String::as_str).collect();
+    allowlist.sort_unstable();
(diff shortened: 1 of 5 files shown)


wolfyy970 (changes requested):
Publishing from the harness is the better direction. It can describe the channels and response mode the running agent actually enforces, so I would use this instead of the separate CLI writer in #5528. #5530 should reuse the same path rather than construct another profile.

Two boundaries need settling first:

- NIP-AP explicitly treats respond_to_allowlist pubkeys as sensitive. Kind 10100 is queried community-wide, so this publishes the exact access list to every member. The public directory can advertise safe routing state, but allowlist membership needs a private boundary.
- Carrying channel_add_policy forward is still a read-modify-write race. If buzz channels set-add-policy writes after the head fetch but before this submit, the harness restores the old policy. Either the harness must become the sole profile writer or directory and policy state need separate ownership.

I would also split the unknown-member fallback into its own PR. The producer fix is useful without changing whether clients show agents whose invocability is unknown.

rmichelena:
Producer-side is the right layer for this, and I think it composes with the read-side work rather than competing with it. Three things I could verify against `main` that bear on the two blockers @wolfyy970 raised.

**1. The allowlist is already world-readable today — in `kind:30177`, deliberately.** Desktop publishes the managed-agent projection with `respond_to_allowlist` in it, and the module says so explicitly (`desktop/src-tauri/src/managed_agents/agent_events.rs:56-58`):

```rust
/// Allowlisted author pubkeys when `respond_to == Allowlist`. These are
/// public keys, not secrets.
#[serde(default, skip_serializing_if = "Vec::is_empty")]
pub respond_to_allowlist: Vec<String>,
```

with the module header noting these events are world-readable on the relay. So this PR doesn't create a new exposure class — it duplicates an existing one into an agent-authored event. That's still worth deciding deliberately (two writers, two copies, and 10100 is agent-authored rather than owner-authored), but it reads to me as "should the project revisit the 30177 stance?" rather than a blocker unique to this PR. If the answer is that allowlists need a private boundary, `KIND_PRIVATE_MANAGED_AGENT` (30179) already exists for exactly that shape.

**2. The `channel_add_policy` race is structural, and the relay is where it's cheapest to fix.** The carry-forward isn't optional — the relay's side effect *requires* the field (`crates/buzz-relay/src/handlers/side_effects.rs:1168-1171`):

```rust
let policy = content
    .get("channel_add_policy")
    .and_then(|v| v.as_str())
    .ok_or_else(|| anyhow::anyhow!("kind:10100 missing channel_add_policy field"))?;
…
state.db.set_channel_add_policy(tenant.community(), &pubkey_bytes, policy).await?;
```

so any publisher of 10100 is forced into read-modify-write against DB state it doesn't own, and a stale head fetch overwrites the policy in the DB — not just in the event. If instead an **absent** `channel_add_policy` meant "leave the stored policy unchanged" (skip the `set_channel_add_policy` call rather than erroring), the harness could publish pure directory state and never restate policy, and the race disappears without needing a single-writer rule. That's a small relay change and it'd let this PR drop its most delicate piece.

**3. Commit 2's dead-code finding reproduces independently.** Same conclusion as #2987 and as our own test-first pass: `addCandidate` runs `isAgentIdentityInAllowedList` before `shouldHideAgentFromMentions` with the same set, so every branch of the latter — including the "member with unknown invocability ⇒ show" case it documents — is unreachable. Worth knowing there are now at least four convergent implementations of that same fix in flight (#5519, #5549, #5536 with a `directoryReady` guard, and this commit), which is an argument for splitting it out as @wolfyy970 suggested so it can be judged on its own.

One scope caveat on it either way: removing that gate fixes the **composer** only. The DM/new-message picker (`useNewMessageRecipients.ts`) and add-member search (`MembersSidebar.tsx`) run in `community` scope, where `relayAgentIsSharedWithUser` deliberately skips any membership fallback (`SECURITY:` comment), so an agent with no usable directory record stays un-DM-able and un-addable regardless.

**Where this meets #5483.** That one attacks the same bug from the read side: `list_relay_agents` merges `kind:30177` (policy) and `kind:39002` (relay-authoritative membership) into the directory at query time. The two aren't alternatives — a producer makes correct records exist going forward; the read-side merge makes the client tolerant of records that don't exist yet, aren't published by a `buzz-acp` harness, or were clobbered by `set-add-policy` (#3663/#4490) before the next republish. It also has no write path at all, so points 1 and 2 above don't arise for it. Happy to rebase on top of this if it lands first — and if a maintainer decides one direction makes the other redundant, I'd rather that be said explicitly than have both sit open.

Issue #3295, opened 2026-07-28
[Bug] Desktop v0.4.26 renders completely blank on macOS 12.7.6 arm64


marketreadyai:
## Summary

The official Buzz Desktop v0.4.26 application opens as a completely
blank full-size window on macOS 12.7.6 (Apple Silicon).

The native application starts, but the frontend never commits its
initial render.

## Environment

- macOS: 12.7.6
- Build: 21H1320
- Architecture: arm64
- Safari: 17.6
- Buzz: 0.4.26
- Bundle ID: xyz.block.buzz.app
- Executable architecture: arm64
- Installation: official notarised DMG
- Gatekeeper: accepted
- Code signature: valid
- Declared minimum macOS version: 10.15

## Steps to reproduce

1. Install the official Buzz Desktop v0.4.26 arm64 DMG.
2. Open Buzz.
3. Observe that the entire application window is blank.

I also reproduced it after:

- quitting Buzz;
- confirming nothing was listening on port 3000;
- backing up/removing Application Support, WebKit, cache, preferences and
  saved-window state;
- relaunching Buzz with clean first-run state.

## Expected result

Buzz displays its welcome/onboarding interface.

## Actual result

The entire application window is blank. There are no visible controls,
messages or onboarding UI.

## Native startup output

buzz-mesh: installed tokio runtime with 8 MiB worker stacks
buzz-desktop: persisted identity pubkey [public key omitted]
buzz-desktop: repos dir resolved at boot — no configured override, REPOS is the default real dir
buzz-desktop: media proxy listening on 127.0.0.1:[dynamic port]
buzz-desktop: initial render did not commit before reveal timeout

## Additional diagnostic information

- The native Buzz process remains running.
- com.apple.WebKit.WebContent starts.
- No application-level JavaScript exception appears in macOS unified logs.
- The issue persists with port 3000 free.
- On the first launch, another local Node application occupied port 3000,
  which caused a separate "relay returned an unexpected HTML page" message.
  Removing that conflict did not fix the blank renderer.
- Docker is unavailable on this Mac, but this report concerns the packaged
  desktop application's onboarding renderer, before any self-hosted relay
  setup.

I can provide the filtered macOS WebKit log and a screenshot of the completely
blank window if useful.

<img width="1919" height="1040" alt="Image" src="https://github.com/user-attachments/assets/a6341dbe-1557-4ac2-98cb-9bacc17087ae" />

claytonwinterbotham:
[...]

PR #2090, opened 2026-07-18
fix(web): ask Safari users for Mac type

baxen:
## Why
Safari does not expose whether a Mac is Apple Silicon or Intel, so invite landing pages silently selected the Intel DMG.

## What
- ask unknown-architecture Mac users to choose a newer or older Mac
- resolve the selected architecture to the matching release asset
- keep automatic downloads for platforms whose architecture is known
- open selected Mac downloads in a new tab, close the chooser immediately, and guard duplicate choices
- preserve the invite page when release lookup falls back to GitHub Releases
- cover the Safari chooser, selected-download fallback, Escape/focus behavior, and unsupported-device fallback in Playwright

## Validation
- `pnpm typecheck`
- `pnpm check`
- `pnpm exec playwright test --project=smoke --grep 'invite asks Safari users|invite download falls back|invite requires age'`\n\nPre-push also passed repository Rust tests; the unrelated desktop Tauri suite had 3 existing Codex adapter discovery failures under the hook environment.\n\nGenerated with mini

PR #5547, opened 2026-08-11
fix(desktop): ship a bundle that parses on WebKit without lookbehind (Intel/macOS 12)

arielhauter:
## Summary

Buzz Desktop renders a blank window on Macs whose system WebKit predates Safari 16.4: the shipped bundle contains RegExp lookbehind patterns, and older WebKit fails at *parse* time, so the frontend never renders. This matches the symptom in #3295, and matters because the project ships and promotes Intel DMGs (#748 added the x86_64 release target; #2090 steers Safari users to the right DMG) while the app declares minimum macOS 10.15.

Four changes make the bundle lookbehind-free:

1. `desktop/vite.config.ts` — build target `["es2020", "safari15"]`, so the toolchain lowers syntax for pre-16.4 WebKit.
2. `mentionHighlightExtension.ts` — rewrite the two mention/channel patterns using a boundary capture group instead of lookbehind, with a prefix-length offset fix in `findHighlightMatches`.
3. `KeyboardShortcutsCard.tsx` — replace the lookbehind/lookahead key-combo split with a small `splitKeyCombo` helper (same behavior).
4. `patches/mdast-util-gfm-autolink-literal@2.0.1.patch` — the dependency ships pre-bundled code with a lookbehind in the email-autolink pattern; patched via the existing `patches/` convention, since a build target cannot rewrite it.

Because the failure is at parse time, one pattern anywhere blanks the whole app — that is why both the build target and the source/dependency rewrites are needed.

Trade-offs, stated honestly: (a) the `safari15` target affects all platforms, not only macOS 12; (b) the dependency patch slightly changes email-autolink edge behavior everywhere, relying on the package's boundary guard instead of the lookbehind — the ideal fix is upstream in micromark, this is the interim.

Heads-up: #4701 (open) introduces a new negative lookbehind in the bare-URL pattern; the same portability concern applies there.

### Related issue

Relates to #3295 — same blank-render symptom; note the original report is arm64/Safari 17.6, which may be a separate cause. This PR fixes the lookbehind parse failure affecting pre-16.4 WebKit. Before opening I searched existing PRs/issues for "macOS 12", "Intel", "lookbehind" — closest existing PR: none found.

### Testing

- `grep -rln -E '\(\?<[=!]'` over the built `Buzz.app` Resources: clean.
- `just desktop-test`: 4,604 tests pass. `just desktop-tauri-test`: pass.
- Built with `just desktop-release-build x86_64-apple-darwin` on current main (post-0.5.9).
- Smoke-tested the resulting DMG on macOS 12 Intel hardware (MacBook Air): window renders, login persists, channels and history load, mentions highlight, URLs linkify. Same patches previously ran in daily self-hosted use on a 0.5.8 base (Air + iMac).
- **Not tested:** Apple Silicon regression, macOS 13+.

### DIFF
--- a/desktop/src/features/messages/lib/mentionHighlightExtension.ts
+++ b/desktop/src/features/messages/lib/mentionHighlightExtension.ts
@@ -106,10 +106,7 @@ export function buildHighlightPatterns(
       n.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"),
     );
     patterns.push(
-      new RegExp(
-        `(?:^|(?<=[\\s(]))@(${escapedNames.join("|")})(?=\\W|$)`,
-        "gi",
-      ),
+      new RegExp(`(^|[\\s(])@(${escapedNames.join("|")})(?=\\W|$)`, "gi"),
     );
   }
 
@@ -121,10 +118,7 @@ export function buildHighlightPatterns(
       n.replace(/[.*+?^${}()|[\]\\]/g, "\\$&"),
     );
     patterns.push(
-      new RegExp(
-        `(?:^|(?<=\\s))#(${escapedChannels.join("|")})(?=\\W|$)`,
-        "gi",
-      ),
+      new RegExp(`(^|\\s)#(${escapedChannels.join("|")})(?=\\W|$)`, "gi"),
     );
   }
 
@@ -145,7 +139,15 @@ export function findHighlightMatches(
     pattern.lastIndex = 0;
     let m: RegExpExecArray | null = pattern.exec(text);
     while (m !== null) {
-      results.push({ from: m.index, to: m.index + m[0].length, match: m[0] });
+      // Group 1 is the boundary prefix (empty at start-of-text). It replaces
+      // the previous lookbehind, which macOS 12's WebKit cannot parse
+      // (see block/buzz#3295); skip it so offsets point at the @/# sigil.
+      const prefixLen = m[1] ? m[1].length : 0;
+      results.push({
+        from: m.index + prefixLen,
+        to: m.index + m[0].length,
+        match: m[0].slice(prefixLen),
+      });
       m = pattern.exec(text);
     }
   }
--- a/desktop/src/features/settings/ui/KeyboardShortcutsCard.tsx
+++ b/desktop/src/features/settings/ui/KeyboardShortcutsCard.tsx
@@ -6,11 +6,34 @@ import {
 import { SettingsOptionGroup, SettingsOptionRow } from "./SettingsOptionGroup";
 import { SettingsSectionHeader } from "./SettingsSectionHeader";
 
+// Split a key-combo string on "+" separators. A "+" is a separator unless it
+// directly follows another "+" (so "⌘++" keeps "+" as a key) or only
+// whitespace remains after it (so a trailing "+" stays attached). Mirrors the
+// previous /(?<!\+)\+(?!\s*$)/ split without lookbehind/lookahead.
+function splitKeyCombo(keys: string): string[] {
+  const parts: string[] = [];
+  let current = "";
+  for (let i = 0; i < keys.length; i++) {
+    const ch = keys[i];
+    const isSeparator =
+      ch === "+" && keys[i - 1] !== "+" && keys.slice(i + 1).trim() !== "";
+    if (isSeparator) {
+      parts.push(current);
+      current = "";
+    } else {
+      current += ch;
+    }
+  }
+  parts.push(current);
+  return parts;
+}
+
 function KeyCombo({ shortcut }: { shortcut: KeyboardShortcut }) {
   const keys = getPlatformKeys(shortcut);
-  // Split on "+" but keep "+" as a standalone key (e.g. for zoom-in "⌘+")
-  const parts = keys
-    .split(/(?<!\+)\+(?!\s*$)/)
+  // Split on "+" but keep "+" as a standalone key (e.g. for zoom-in "⌘+").
+  // Implemented without RegExp lookbehind so the bundle parses on macOS 12's
+  // system WebKit (Safari 16.4 feature). See block/buzz#3295.
+  const parts = splitKeyCombo(keys)
     .map((p) => p.trim())
     .filter(Boolean);
 
--- a/desktop/vite.config.ts
+++ b/desktop/vite.config.ts
@@ -28,6 +28,13 @@ export default defineConfig(async () => ({
     },
   },
 
+  // Target Safari 15 so the bundle parses on macOS 12's system WebKit (613),
+  // which lacks Safari 16.4 syntax (e.g. RegExp lookbehind in literals,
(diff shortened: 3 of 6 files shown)


aurelie-brebant:
Confirming this PR's diagnosis against the current release, on Intel hardware different from the ones already in the thread. I have not built the PR, so this confirms the analysis and its mapping onto the current tag, not the fix itself.

### Environment

- MacBook, Intel x86_64
- macOS 12.7.6 (21H1320)
- Buzz Desktop 0.5.20, installed from the official `Buzz 0.5.20.dmg`, no local build
- Result: no window; `initial render did not commit before reveal timeout`, native process stays alive, `com.apple.WebKit.WebContent` spawns

The last confirmation against a release DMG in #3295 was v0.5.5 (7 Aug).

### The four changes still map exactly onto `desktop-v0.5.20`

Checked against the tag, not `main`:

- A full scan of `desktop/src` for `(?<[=!]` returns exactly three hits, and they are precisely the two files this PR rewrites:
  - `desktop/src/features/settings/ui/KeyboardShortcutsCard.tsx:17`
  - `desktop/src/features/messages/lib/mentionHighlightExtension.ts:488` and `:503` (the patterns @claytonwinterbotham flagged at 110/125, since moved)
- No other lookbehind has crept into first-party source.
- `pnpm-lock.yaml` still pins `mdast-util-gfm-autolink-literal@2.0.1` (via `remark-gfm@4.0.1`) and `@radix-ui/react-collection@1.1.15`.
- `patches/` still contains only `isomorphic-git.patch` and `virtua@0.49.3.patch`, so the new patch file lands cleanly.
- `desktop/vite.config.ts` still declares no `build.target`, and there is still no `minimumSystemVersion` in `desktop/src-tauri/tauri.conf.json`.

The boot-graph edge that makes `KeyboardShortcutsCard` a startup failure rather than a settings-only one is also unchanged: `AppShell.tsx:70-74` imports `DEFAULT_SETTINGS_SECTION` and `isSettingsSection` as values (used at line 175) from `SettingsPanels.tsx`, which statically imports `KeyboardShortcutsCard` at line 66.

### One correction to the reasoning, which strengthens the case for the build target

Earlier in #3295 the missing target was described as leaving the dependency graph at `esnext`. That is not what Vite 8 does, and the real default is a better argument for this PR.

`vite@8.0.16`, the pinned version, defaults `build.target` to `baseline-widely-available`, documented in its own type declarations as:

> transpile targeting browsers that are included in the Baseline Widely Available on 2026-01-01. (Chrome 111+, Edge 111+, Firefox 114+, Safari 16.4+).

Safari 16.4 is exactly the release that added both RegExp lookbehind and class static blocks. The default is therefore off by precisely one WebKit generation from macOS 12's system WebKit: both constructs are treated as supported and passed through untouched. A default named "baseline widely available" reads as conservative, which is likely why this went unnoticed.

### Measured in a real WKWebView on this machine

A small Swift harness loading the constructs as separate `<script>` blocks, so each block's parse result is independent:

```
navigator.userAgent                 AppleWebKit/605.1.15
control script                      runs
/(?<!\+)\+(?!\s*$)/                 block never runs
/(?<=^|\s|\p{P}|\p{S}).../gu        block never runs
class C extends Map { static {} }   block never runs
new RegExp("(?<!x)y")               block never runs
```

The WebContent binary actually executed is `17613.3.9.1.16`, while `Safari.app` on the same machine reports 17.6, matching the red herring already noted in #3295.

The last line is the one worth keeping in mind for review: `new RegExp` with a lookbehind fails too, so no build target can rescue `mentionHighlightExtension.ts`. The source rewrite in this PR is required, not cosmetic.

Happy to run any specific check on macOS 12 Intel if it helps move the review along.

XIII-EOG:
Hi! I have a MacBook Pro 13" Early 2015, Intel i5, 8 GB RAM, running macOS 12.7.6. The official Buzz 0.5.23 x64 DMG opens to a blank window, and older releases have the same issue.

I saw that you successfully tested the WebKit fix on Intel Monterey. Would you be able to share the patched x64 DMG or a downloadable build artifact? I’m not a developer, and building from source on this old Mac requires a large toolchain that I’m struggling to install.

Thank you!

aurelie-brebant:
@arielhauter a practical question, as one of the people waiting on this. GitHub now reports the branch as conflicting with `main` (`mergeable: false`). `mentionHighlightExtension.ts` has moved a fair bit since 11 Aug: the two lookbehind patterns were at :488 and :503 at `desktop-v0.5.20` and are at :569 and :584 on `main` today.

Since maintainers generally skip conflicted PRs during triage, the rebase looks like the real blocker right now rather than reviewer attention. Are you still planning to pick this back up? If not, would you mind saying so, so someone else can carry the same four commits forward rather than leaving it stalled?

State of things as of today, for whoever looks next: `main` still has no `build.target`, all three lookbehinds are still present, `patches/` is unchanged, and 0.5.22 and 0.5.23 both shipped without the fix.

arielhauter:
> @arielhauter a practical question, as one of the people waiting on this. GitHub now reports the branch as conflicting with `main` (`mergeable: false`). `mentionHighlightExtension.ts` has moved a fair bit since 11 Aug: the two lookbehind patterns were at :488 and :503 at `desktop-v0.5.20` and are at :569 and :584 on `main` today.
> 
> Since maintainers generally skip conflicted PRs during triage, the rebase looks like the real blocker right now rather than reviewer attention. Are you still planning to pick this back up? If not, would you mind saying so, so someone else can carry the same four commits forward rather than leaving it stalled?
> 
> State of things as of today, for whoever looks next: `main` still has no `build.target`, all three lookbehinds are still present, `patches/` is unchanged, and 0.5.22 and 0.5.23 both shipped without the fix.
> 

Hi @aurelie-brebant yes I'll revisit this tomorrow and get it up-to-date! Appreciate your shared focus on it...

@XIII-EOG I'll see if I can get you a downloadable build tomorrow...

XIII-EOG:
"@XIII-EOG I'll see if I can get you a downloadable build tomorrow..." Hello, sir! Sounds perfect, waiting forward. Huge thank you!!!

arielhauter:
@XIII-EOG CI on this PR needs maintainer approval before it runs, so no CI artifact yet. I've published my rebased build as a release on my fork (so you can download the DMG from there, if have any issues, just let me know and I'll send it another way):

https://github.com/arielhauter/buzz/releases/tag/v0.5.23-macos12-webkit-fix

SHA-256 and install steps are in the release notes. It's built from `main` @ c507a4d48 plus this PR's four commits, smoke-tested on two macOS 12 Intel machines. Unofficial and unsigned — please verify the checksum, and report back here whether it renders on your 2015 MBP. That's a third hardware confirmation for the maintainers.

arielhauter:
@tlongwell-block — tagging you since #748 added the Intel x64 DMG release target and this PR fixes the blank window on the macOS 12 machines that target ships to. It's rebased and conflict-free, with three users waiting on this thread (#3295). Could someone on the team approve the workflow runs so CI can verify it? Happy to make any changes review turns up.

PR #1913, opened 2026-07-15
Add KLIPY GIF search to message composers

klopez4212:
## Summary

- adds Emoji and GIF tabs to message and forum composers
- supports KLIPY trending/search results and uploads the selected GIF through Buzz's existing media flow
- keeps Emoji Mart aligned with Buzz's theme and picker width

@wesbillman @tlongwell-block — support for GIF search was pretty straightforward to add; this is a working end-to-end implementation.

## Next steps before release

Business-side help needed:

- request a KLIPY production API key
- confirm that posting selected GIFs into Buzz media storage is covered by KLIPY's terms
- confirm the required ad model, official attribution, and privacy disclosure

Engineering follow-up after those decisions: wire the production key into release builds, hide the tab when unconfigured, add the official KLIPY branding, and harden pagination and download-size limits.

## Validation

- tested trending, search, selection, upload, and light/dark themes in the staging app
- 2,912 desktop unit tests pass
- desktop typecheck, formatting checks, and production build pass
- the broader local Rust gate was attempted but exhausted machine disk; GitHub CI will run the complete gate

PR #5554, opened 2026-08-11
feat(desktop): add KLIPY GIF search to composers

klopez4212:
## Summary

- Supersedes #1913 with a KLIPY-hosted URL implementation.
- Adds KLIPY GIF search and trending results to desktop message and forum composers.
- Keeps selected GIFs hosted by KLIPY; Buzz stores only the external URL and media metadata (no imeta tag, since relays only accept hash-backed local `/media/` entries).
- Aligns the Emoji/GIF picker with Buzz's standard segmented control, theme surfaces, and motion behavior.

## Relay-to-provider boundary

- The relay proxies KLIPY search/share so the `BUZZ_KLIPY_API_KEY` never reaches the desktop; the key stays server-side behind a redacted `Debug` impl.
- The dedicated GIF `reqwest` client sets `redirect::Policy::none()`. Because the API key rides in the request path, following a provider `3xx` could replay a key-bearing URL to an attacker-chosen host (an SSRF/key-disclosure primitive). With redirects disabled, a `3xx` returns as a non-success status that the handlers map to a generic `502`; the `Location` target is never read or forwarded.
- Admission reuses the established NIP-98, membership, replay, and per-pubkey rate-limit gates, with an upstream response-size cap and allowlisting so KLIPY error bodies never cross the relay boundary.

## Accessibility

- Under `prefers-reduced-motion: reduce`, the picker grid renders a static provider poster (a normalized `jpg` asset) instead of the animated preview, or a named static placeholder when no poster is available. It reacts to preference changes while mounted. `no-preference` keeps the animated preview.
- Selected GIFs carry their title through `ImetaMedia`'s `displayLabel`, so the composer thumbnail, preview dialog, editor, lightbox, and remove control all derive one non-empty accessible name instead of an empty `Attachment ` label. Ordinary hashed uploads keep their existing hash-derived names.


desktop/src/features/gifs/api.ts
```diff
@@ -0,0 +1,150 @@
...
+/**
+ * Represent a selected KLIPY GIF as externally hosted media. The empty hash
+ * intentionally distinguishes it from bytes uploaded to Buzz media storage;
+ * NIP-92 permits URL-only media metadata.
+ */
+export function klipyGifAttachment(gif: KlipyGif): BlobDescriptor {
+  return {
+    dim: `${gif.original.width}x${gif.original.height}`,
+    filename: klipyGifFilename(gif),
+    sha256: "",
```
wpfleger96, line 179: 🤖 blocking: this descriptor produces a message the relay rejects. `validate_imeta_tags` runs on every ingested event carrying imeta tags (`crates/buzz-relay/src/handlers/ingest.rs:2752`), and it requires the `url` to be a local `/media/` path and `x` to be present (`crates/buzz-relay/src/handlers/imeta.rs:61` and `:163`) — so this fails on two independent grounds. A live run against this branch confirmed it: both stream (kind `9`) and forum (kind `45001`) sends came back `400 {"error":"invalid: imeta url must be a local /media/ path"}` and nothing was persisted. The comment above is right that NIP-92 permits URL-only metadata, but our relay validator is stricter, and `docs/gif-search.md` currently describes behavior the relay doesn't accept.

I think there are two fix shapes and it's worth an explicit call: (a) teach `validate_imeta_tags` a bounded external-GIF allowance (https-only, no `x` required, skip blob verification) — that widens the ingest surface and needs its own review; or (b) minimal: skip the imeta tag when `sha256 === ""` and rely on the appended markdown image line, which the existing renderer already handles per the docs (at the cost of dim/size metadata). Either way, please add coverage for both kinds and rerun a live send to verify.

desktop/src/features/messages/ui/ComposerEmojiPicker.tsx
```diff
@@ -43,7 +90,9 @@ export const ComposerEmojiPicker = React.memo(function ComposerEmojiPicker({
             </Button>
           </PopoverTrigger>
         </TooltipTrigger>
-        <TooltipContent>Insert emoji</TooltipContent>
+        <TooltipContent>
+          {gifsAvailable ? "Emoji and GIFs" : "Emoji"}
```
wpfleger96, line 99: 🤖 blocking: `desktop/tests/e2e/composer-tooltip-dismiss.spec.ts` still locates the tooltip by `Insert emoji`, which no longer exists in any branch of this render. [Desktop Smoke E2E (2)](https://github.com/block/buzz/actions/runs/32393490919/job/96505028860) failed that assertion on the initial attempt and both retries (the shard's other 274 tests passed), so the red Desktop gate is caused by this PR, not flake. Update the spec to the new capability-dependent labels — and worth a quick grep for other `Insert emoji` selectors; I only found this spec and the component itself.

crates/buzz-relay/src/api/gifs.rs
```diff
@@ -0,0 +1,331 @@
...
+//!
+//! KLIPY requires a provider credential, but desktop applications cannot keep
+//! build-time credentials secret. These narrow endpoints keep the key on the
+//! operator's relay while returning only KLIPY-hosted media URLs and metadata;
+//! GIF bytes are never downloaded, cached, or stored by Buzz.
+//!
+//! Search is intentionally the only relay endpoint. Sending a selected GIF is
+//! a normal message containing its CDN URL, and clients render that URL through
+//! the existing image path. A share or render proxy would add operator egress
+//! cost and an open-proxy-shaped security surface without a demonstrated need.
```
wpfleger96: 🤖 important: I don't think the no-share-endpoint call is as free as this comment suggests. KLIPY's reference integrations treat share reporting as required when media is actually sent — their Android demo says apps must call `triggerShare()` on selection/send so items enter Recents ([README](https://github.com/KLIPY-com/klipy-android-demo-app/blob/cc013cde3a239dec7a849b6accdaa70131b7210b/README.md#L180-L194), [`GifService.kt`](https://github.com/KLIPY-com/klipy-android-demo-app/blob/cc013cde3a239dec7a849b6accdaa70131b7210b/app/src/main/java/com/klipy/demoapp/data/service/GifService.kt#L44-L48)), and their iOS demo independently couples send to `trackShare()` ([`ChatView.swift`](https://github.com/KLIPY-com/klipy-ios-demo-app/blob/e5b5b3f5583fc7010ebaf8ea62700eeb387a2924/KlipyiOS-demo/App/ChatFeature/ChatView.swift#L123-L128)). Skipping it silently under-reports sends and breaks KLIPY Recents for our users.

A share route is a telemetry POST forwarding the GIF slug + `customer_id` — not a media render proxy — so it doesn't create the egress cost or open-proxy surface this comment is guarding against. I'd either add a member-gated, validated, rate-limited share route (keeping the key relay-side, with route/client coverage) or get explicit confirmation from KLIPY that skipping it is acceptable before we document it as a non-goal.

crates/buzz-relay/src/api/gifs.rs
```diff
@@ -0,0 +1,331 @@
...
+        ("page", "1"),
+        ("per_page", "24"),
+        ("customer_id", request.customer_id.as_str()),
+        ("locale", request.locale.as_str()),
+    ];
+    if !request.query.trim().is_empty() {
+        query.push(("q", request.query.trim()));
+    }
+    let url = klipy_url(config.api_key(), &["gifs", endpoint], &query)?;
+    let response = send_upstream(reqwest::Client::new().get(url)).await?;
```
wpfleger96: 🤖 `reqwest::Client::new()` per request builds a fresh connection pool and does a full TLS handshake to `api.klipy.com` on every search. The repo convention (see `push_runtime.rs`) is to build one client and reuse it — store one in `AppState` or a `OnceLock`. At 30 req/min/pubkey across a fleet this is real avoidable latency and socket churn.

crates/buzz-relay/src/api/gifs.rs
```diff
@@ -0,0 +1,331 @@
...
+            StatusCode::BAD_GATEWAY,
+            "GIF provider rejected the search request",
+        ));
+    }
+    let data = upstream.get("data").cloned().unwrap_or(Value::Null);
+    Ok(serde_json::json!({ "result": true, "data": data }))
+}
+
+/// Search or browse trending KLIPY GIF metadata for an authenticated member.
+pub async fn search(
```
wpfleger96, line 261: 🤖 the unit tests below cover `klipy_url`, `validate_text`, and payload filtering, but nothing exercises the handler seam itself: 404 when unconfigured, auth rejection, quota rejection, oversized-body rejection. The auth chain composition (`bind_community` → NIP-98 with required payload → admission → replay guard → membership) is correct today, but nothing pins it against regression, and neighboring API modules do have handler-level tests.

desktop/src/features/gifs/relay.ts
```diff
@@ -0,0 +1,117 @@
...
+  type KlipyGif,
+  type KlipyResponse,
+  normalizeKlipyGifs,
+  type RelayGifSearchInfo,
+  relayKlipySearchPath,
+} from "@/features/gifs/api";
+import { relayHttpFromWs } from "@/shared/api/inviteHelpers";
+import { signRelayEvent } from "@/shared/api/tauri";
+
+const KLIPY_CUSTOMER_ID_STORAGE_KEY = "buzz:klipy-customer-id:v1";
```
wpfleger96: 🤖 one UUID reused across every community/relay this install talks to (and forwarded to both the relay and KLIPY) is a cross-community correlation identifier — the provider, or two colluding relay operators, could link a single install across communities. Keying it per relay (e.g. `buzz:klipy-customer-id:v1:<relayUrl>`) is a cheap fix with a real privacy property.

crates/buzz-relay/src/api/gifs.rs
```diff
@@ -0,0 +1,331 @@
...
+
+use buzz_auth::LimitType;
+
+use super::{api_error, bridge, internal_error, relay_members};
+
+const KLIPY_API_ROOT: &str = "https://api.klipy.com/api/v1/";
+const SEARCH_PATH: &str = "/gifs/search";
+const UPSTREAM_TIMEOUT: Duration = Duration::from_secs(10);
+const MAX_UPSTREAM_RESPONSE_BYTES: usize = 2 * 1024 * 1024;
+const SEARCHES_PER_MINUTE_PER_PUBKEY: u64 = 30;
```
wpfleger96: 🤖 nit: sibling limits live in the `rate_limits` config; hardcoding this one means an operator can't tune it without a rebuild.

crates/buzz-relay/src/api/gifs.rs
```diff
@@ -0,0 +1,331 @@
...
+    body: axum::body::Bytes,
+) -> Result<Json<Value>, (StatusCode, Json<Value>)> {
+    let Some(config) = state.config.klipy.as_ref() else {
+        return Err(api_error(
+            StatusCode::NOT_FOUND,
+            "GIF search is not configured",
+        ));
+    };
+    let (tenant, pubkey) = authenticate(&state, &headers, SEARCH_PATH, &body).await?;
+    enforce_search_admission(&state, &tenant, &pubkey).await?;
```
wpfleger96: 🤖 nit: the search quota is charged here before the body parses, so malformed JSON burns a search token. Parsing/validating first would only charge well-formed requests.

desktop/src/features/gifs/relay.ts
```diff
@@ -0,0 +1,117 @@
...
+      "Content-Type": "application/json",
+    },
+    method: "POST",
+    signal,
+  });
+  if (!response.ok) {
+    const json = (await response.json().catch(() => ({}))) as {
+      error?: string;
+    };
+    throw new Error(json.error || `GIF request failed (${response.status})`);
```
wpfleger96: 🤖 nit: on a membership denial the relay returns `{"error":"relay_membership_required"}` and this surfaces that raw machine token in the picker error state. Worth mapping known error codes to friendly messages.

crates/buzz-relay/src/api/gifs.rs
```diff
@@ -0,0 +1,331 @@
...
+use serde_json::Value;
+
+use crate::state::AppState;
+
+use buzz_auth::LimitType;
+
+use super::{api_error, bridge, internal_error, relay_members};
+
+const KLIPY_API_ROOT: &str = "https://api.klipy.com/api/v1/";
+const SEARCH_PATH: &str = "/gifs/search";
```
wpfleger96: 🤖 nit: `/gifs/search` is spelled in three places (router, this const, the NIP-11 descriptor) with nothing tying them together — a shared const would prevent drift.

desktop/src/features/gifs/api.ts
```diff
@@ -0,0 +1,150 @@
...
+      item.file.sm?.gif,
+      item.file.xs?.webp,
+      item.file.xs?.gif,
+      item.file.md?.webp,
+      original ?? undefined,
+    );
+    if (!original || !preview) continue;
+
+    gifs.push({
+      id: item.id ?? gifs.length,
```
wpfleger96: 🤖 nit: the `gifs.length` fallback can collide with a real id in the `${gif.id}-${gif.slug}` React key. The slug alone is probably enough.

wpfleger96 (changes requested):
🤖 Thanks for putting this together — the relay proxy is genuinely well built. The key stays server-side with a redacted `Debug` impl, NIP-98 verification requires the payload hash, and there's a replay guard, membership + per-pubkey quota enforcement, an upstream response-size cap, and allowlisting so KLIPY error bodies never cross the relay boundary. That whole boundary held up under both source review and a live run against the real KLIPY API.

I'm requesting changes on two blocking issues, both verified end-to-end, plus one design question I'd like resolved before merge:

1. Selecting a GIF produces a message the relay refuses to ingest — the core feature doesn't work end-to-end. A live run against this branch got `400 {"error":"invalid: imeta url must be a local /media/ path"}` on both stream (kind `9`) and forum (kind `45001`) sends, with zero posts persisted. Details inline at `desktop/src/features/gifs/api.ts`.

2. The PR breaks its own smoke test: `composer-tooltip-dismiss.spec.ts` still locates the `Insert emoji` tooltip, and [Desktop Smoke E2E (2)](https://github.com/block/buzz/actions/runs/32393490919/job/96505028860) failed that assertion on the initial attempt and both retries. Details inline at `ComposerEmojiPicker.tsx`.

3. KLIPY's own reference integrations couple sends to their share/report endpoint, and this PR documents skipping that as a non-goal — I don't think that call is as free as the docs suggest. Inline comment on `crates/buzz-relay/src/api/gifs.rs` with references.

There are also a few smaller things inline (upstream client reuse, handler-level test coverage, per-relay `customer_id` keying, and some nits).

desktop/src/features/gifs/api.ts
klopez4212: Fixed via the minimal option (b) in `1bd867b03`: hashless external GIF descriptors now contribute only the markdown image URL, while `buildImetaTags` omits them entirely. `buildOutgoingMessage` is the shared builder used by both stream messages and `ForumComposer`, so neither kind 9 nor kind 45001 emits the relay-invalid external `imeta`. Added unit coverage for the hashless content-only shape and Playwright coverage asserting the signed kind-9 event contains the URL with no `imeta` tag; docs now describe the relay's local-media constraint.

desktop/src/features/messages/ui/ComposerEmojiPicker.tsx
klopez4212: Fixed in `1bd867b03`: the smoke spec now targets the rendered `Emoji` tooltip, while the trigger's accessible name remains capability-dependent (`Insert emoji` / `Insert emoji or GIF`). The current Desktop Smoke E2E shards, including shard 2, are green on `9db36e28a`.

crates/buzz-relay/src/api/gifs.rs
klopez4212: Implemented in `1bd867b03`: the relay now advertises and serves authenticated `POST /gifs/share`, validates the bounded slug/customer ID, enforces the same NIP-98 replay and relay-membership boundary as search, forwards only the required provider payload, and returns no provider body. Desktop reports the share best-effort after selection so a telemetry failure cannot fail the user's send. Added route/client coverage and updated the capability/docs contract.

crates/buzz-relay/src/api/gifs.rs
klopez4212: Fixed in `1bd867b03`: `AppState` now owns one reusable `gif_http_client`, and both search and share use its connection pool rather than constructing a client per request.

crates/buzz-relay/src/api/gifs.rs
klopez4212: Expanded in `1bd867b03`: the tests now exercise the Axum handler's unconfigured 404 seam and streamed oversized-response rejection, in addition to request validation/provider allowlisting. The handler continues through the shared NIP-98 payload verification, replay guard, membership gate, and Redis admission helpers; the full relay and backend integration CI lanes are green on `9db36e28a`.

desktop/src/features/gifs/relay.ts
klopez4212: Fixed in `1bd867b03`: the persisted anonymous customer ID is now keyed by relay URL (`buzz:klipy-customer-id:v1:<relayUrl>`). If storage is unavailable, the client generates an ephemeral UUID rather than falling back to a cross-relay constant.

crates/buzz-relay/src/api/gifs.rs
klopez4212: Fixed in `1bd867b03`: GIF search quota is now part of `RateLimitConfig` and configurable with `BUZZ_RATE_LIMIT_GIF_SEARCHES_PER_MIN` (default 30), including config parsing/tests and operator documentation.

crates/buzz-relay/src/api/gifs.rs
klopez4212: Fixed in `1bd867b03`: JSON parsing and bounded-field validation now run before `enforce_search_admission`, so malformed/invalid requests do not consume a search token.

desktop/src/features/gifs/relay.ts
klopez4212: Fixed in `1bd867b03`: known relay error codes are mapped client-side; `relay_membership_required` now renders as `Join this community to search GIFs.` instead of exposing the machine token.

crates/buzz-relay/src/api/gifs.rs
klopez4212: Fixed in `1bd867b03`: `SEARCH_PATH` is now the shared constant used by the router and NIP-11 descriptor. The new share route follows the same pattern with `SHARE_PATH`.

desktop/src/features/gifs/api.ts
klopez4212: Fixed in `1bd867b03`: missing provider IDs normalize to `null`, and the picker uses the provider slug as the React key rather than a synthetic array-index fallback.

wpfleger96 (dismissed):
🤖 All feedback from the combined review is addressed at `9db36e2` — verified each finding at source, and both blockers are confirmed fixed: external GIFs now send as content-only markdown (no imeta tag, so the relay validator is never engaged) and the tooltip smoke spec is green. The new `/gifs/share` endpoint sits behind the same NIP-98/membership/replay boundary as search, and live E2E against the real KLIPY provider confirmed sends persist and render on both kind 9 and forum posts, plus fully graceful degradation (no GIF tab, no errors, zero `/gifs/*` requests) when the relay has no KLIPY key. CI is fully green at head.

jedwards27 (changes requested):
**Verdict: REQUEST CHANGES**

**Reviewed:** `aea0ef8df9fc24d9aa8bf5c761ab2910026a601b..9db36e28afbe76aa3575b0665baa2b316de9d90d` (exact head `9db36e28afbe76aa3575b0665baa2b316de9d90d`)

**Risk:** high — this adds a credentialed relay-to-provider HTTP boundary plus externally hosted animated media in shared message/forum composer paths.

**Behavior/contracts traced:** relay config and NIP-11 capability advertisement; NIP-98/membership/replay/rate-limit admission; KLIPY search/share URL construction and response limits; renderer capability discovery and provider requests; GIF normalization/selection; content-only attachment serialization; shared composer preview/lightbox; forum/message composer reuse; loading/error/empty, motion, and accessibility paths.

## Blocking findings

1. **[P1] Provider redirects create an SSRF primitive and can disclose the path-embedded API key.** `crates/buzz-relay/src/state.rs:858-861` builds the shared reqwest client without a redirect policy. Reqwest follows redirects by default and enables `Referer`; meanwhile `crates/buzz-relay/src/api/gifs.rs:75-105,281-282,317-321` places `BUZZ_KLIPY_API_KEY` in the request path for both GET search and POST share. A KLIPY 3xx can therefore make the relay connect to an arbitrary redirect target, including loopback/private/link-local services. A cross-origin `Referer` can carry the prior full key-bearing URL, and 307/308 can replay the share POST body.

   **Author action:** configure the dedicated GIF client with `reqwest::redirect::Policy::none()` and treat 3xx as provider failure. Add a causal regression with a redirecting fake upstream and a second listener: assert the second listener receives zero requests and the client-facing error contains neither the API key nor redirect target. Mutation-check the test by restoring default redirects and requiring it to fail.

   **Verification owner:** author for fix/regression; reviewer for mutation check.

2. **[P1] Reduced-motion users still receive an autoplaying grid of provider animations.** `desktop/src/features/gifs/ui/KlipyGifPicker.tsx:108-126` renders every preview URL as an ordinary image. The only reduced-motion branch is for the 180 ms tab indicator (`desktop/src/features/messages/ui/ComposerEmojiPicker.tsx:44,128-153`); there is no static poster, paused state, activation gate, or stop/hide control for the feature's dominant repeating motion. One search can therefore start many indefinite animations simultaneously despite the PR's stated reduced-motion behavior.

   **Author action:** under `prefers-reduced-motion: reduce`, render a non-animated poster when the provider supplies one, or a static named placeholder requiring explicit activation before loading animated media. Add deterministic Playwright coverage using `page.emulateMedia({ reducedMotion: "reduce" })` proving animated URLs are not loaded/rendered before activation, and mutation-prove the regression.

   **Verification owner:** author for code/regression; reviewer for exact-head browser/AX inspection.

3. **[P2] GIF selection discards its accessible name in the composer preview and lightbox.** The picker has the useful title (`KlipyGifPicker.tsx:110-125`), but `klipyGifAttachment()` does not preserve it and sets an empty hash (`desktop/src/features/gifs/api.ts:160-169`). Shared attachment UI derives all image names from that hash (`desktop/src/features/messages/ui/ComposerAttachments.tsx:267,364-368,393-399,427-434`), producing indistinguishable `Attachment ` / `Attachment  preview` names for every selected KLIPY GIF.

   **Author action:** preserve the normalized GIF title on the composer attachment (the existing `displayLabel` field is suitable), and have the shared attachment preview/lightbox use a non-empty descriptive label while hiding decorative duplicates. Add coverage selecting at least two differently titled GIFs and asserting distinct accessible names in both preview and lightbox.

   **Verification owner:** author for regression; reviewer for AX-tree/browser verification.

## Validation

- PASS — full Desktop package tests: `pnpm test`, 5,312 passed / 0 failed, exact head and clean tree rechecked.
- PASS — `git diff --check aea0ef8df9fc24d9aa8bf5c761ab2910026a601b..HEAD`.
- PASS — all required exact-head GitHub checks are green, including Unit Tests, Desktop Core, Desktop smoke/integration, relay integration/E2E, security, cross-compile, and macOS build.
- ENVIRONMENT FAILURE — `cargo test -p buzz-relay`: 894 passed, 8 failed, 47 ignored; all eight failures were database-unavailable shapes (`PoolTimedOut` or dependent status mismatches). Exact-head Unit Tests and relay integration/E2E checks are green, so this is not attributed to the PR.
- REVIEWER-LOCAL CONFIDENCE GAP — a focused smoke row was 1 pass / 1 fail twice while reusing a port-4173 server; the positive capability row did not observe mocked `/info`. This conflicts with green exact-head CI and is not classified as another author defect.

**Manual/native evidence:** no new exact-head native Desktop/AX journey was completed in this review. Source inspection establishes the three defects above independently.

**Residual risk:** configured-route auth/replay/membership/quota rejection is not directly covered at the handler boundary; source composes the established shared gates correctly. Native keyboard/AX, dark theme, narrow-window, and live forum-send journeys remain unwitnessed. These are confidence gaps, not additional author actions.

— :bot: Jude’s code review agent

wpfleger96 (approved):

PR #5559, opened 2026-08-11
fix(mobile): address DM recipients on send

andynguyenhuu:
## Summary

- add every other DM participant as a recipient `p` tag at the mobile send boundary
- preserve explicit mentions while normalising, deduplicating, and excluding the sender
- cover plain DMs, uncached member lookup, and threaded replies

## Verification

- Flutter analyser passed
- mobile file-size gate passed
- full mobile suite passed: 1,266 tests

Originated from Buzz DM channel `d22c8a5f-83ca-44aa-82ad-a6ef09ae1398`.

### DIFF
--- a/mobile/lib/features/channels/channel_detail_page.dart
+++ b/mobile/lib/features/channels/channel_detail_page.dart
@@ -524,6 +524,9 @@ class ChannelDetailPage extends HookConsumerWidget {
                             channelId: channel.id,
                             content: content,
                             mentionPubkeys: mentionPubkeys,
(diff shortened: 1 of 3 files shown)


andynguyenhuu:
@wesbillman Could you review and merge this mobile DM fix? Plain mobile DMs were omitting recipient `p` tags, so mention-subscribed ACP agents did not wake. The Flutter analyser, file-size gate, and full 1,266-test mobile suite passed; DCO is green and the PR is mergeable.

PR #5560, opened 2026-08-11
fix(media): accept portrait video at the same bound as landscape

AleksandarIlic:
## Summary

- compare the video resolution bound against the long and short edge instead of raw width and height
- cover portrait 4K (2160x3840) and a portrait clip one pixel past the long edge

## Root cause

`validate_video_file()` checked `width > 3840 || height > 2160`, which reads the bound as landscape-only. The same 4K frame is 3840x2160 landscape and 2160x3840 portrait: the first is accepted, the second rejected on height, despite being identical in pixel count, codec, and file size.

This affects anything recorded in portrait above 2160 tall — 4K phone video, and native-resolution phone screen recordings such as 1180x2556. No client caps resolution before upload: the desktop transcode pads to even dimensions without downscaling and mobile uploads at native resolution, so these files are rejected only after the whole body has been received.

Landscape behaviour is unchanged, and the bound still applies in portrait — 2160x3841 is still rejected.

## Validation

- `cargo test -p buzz-media`: 110 passed, including the two new cases; the existing 3841x2161 rejection test is unaffected
- `cargo check --workspace --all-targets`, `cargo clippy -p buzz-media -p buzz-relay --all-targets`, `cargo fmt --all --check`: clean
- `cargo test -p buzz-relay --lib`: 859 passed; the 8 failures are pre-existing and infra-dependent (`Sqlx(PoolTimedOut)`, no local Postgres/Redis)

### DIFF
--- a/crates/buzz-media/src/validation.rs
+++ b/crates/buzz-media/src/validation.rs
@@ -282,7 +282,7 @@ pub fn validate_content(bytes: &[u8], config: &MediaConfig) -> Result<String, Me
 /// - Exactly one video track using `avc1` (H.264 only — rejects HEVC, VP9, AV1)
 /// - At most one audio track, using `mp4a` (AAC)
 /// - Duration ≤ 600 seconds (from mvhd timescale, not edit lists)
-/// - Resolution ≤ 3840×2160
+/// - Resolution ≤ 3840×2160 in either orientation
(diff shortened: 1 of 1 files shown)


AleksandarIlic:
This was fixed upstream in #6058 (196d62f9). Same orientation-independent envelope, and it covers a case I missed here (2161×3840, short edge over). Thanks!

PR #5563, opened 2026-08-11
feat(desktop): auto-open agent reply threads

michaelzeyuchen:
## Summary

When a user sends a top-level message targeting a known channel agent, automatically open the originating thread when that agent returns a hidden NIP-10 reply.

Previously, the reply landed correctly but rendered only inside the closed thread panel, so the active exchange appeared to leave the conversation until the user clicked the reply summary.

## Behavior

The policy is intentionally narrow:
- hidden NIP-10 thread replies only (not broadcasts or top-level messages)
- thread root must exactly match the user's most recent agent-targeted message
- author must resolve to a known channel agent
- yields whenever an auxiliary panel is already open
- deferred opening bails if the active channel changes before the transition runs

Delegated relay attribution remains trust-bound: `resolveEventAuthorPubkey` accepts actor/p tags only from the relay identity advertised by NIP-11.

## Implementation

- pure `decideAgentReplyAutoOpen` policy
- dedicated hook owning the one-shot trigger and live-event subscription
- optional `onLiveEvent` callback in `useChannelSubscription`, consumed only by the dedicated hook
- send success/failure integration in the channel pane handlers
- eight focused `node:test` cases

`ChannelScreen.tsx` remains within the file-size ratchet at 998/1000 lines.

## Verification

From `desktop/`:
- `pnpm exec biome check` (feature files) — pass
- `pnpm run typecheck` — pass
- focused `node:test` — 8/8 pass
- `pnpm run check:file-sizes` — pass
- `pnpm run check:px-text` — pass
- `pnpm run check:pubkey-truncation` — pass

Not run locally: full Rust/Tauri workspace CI; GitHub Actions will exercise it.

### DIFF
--- a/desktop/src/features/channels/ui/ChannelScreen.tsx
+++ b/desktop/src/features/channels/ui/ChannelScreen.tsx
@@ -34,7 +34,6 @@ import { useWelcomeAgentCreate } from "@/features/channels/useWelcomeAgentCreate
 import { useCommunities } from "@/features/communities/useCommunities";
 import {
   useChannelMessagesQuery,
-  useChannelSubscription,
   useChannelWindowQuery,
   useDeleteMessageMutation,
   useEditMessageMutation,
@@ -195,7 +194,6 @@ export function ChannelScreen({
     activeChannel,
     effectiveOpenThreadHeadId,
   );
-  useChannelSubscription(activeChannel);
   const { fetchOlder, hasOlderMessages, historyExhausted, isFetchingOlder } =
     useFetchOlderMessages(activeChannel);
   const latestActiveMessage = React.useMemo(() => {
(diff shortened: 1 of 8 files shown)


michaelzeyuchen:
Friendly ping — this one is ready for a look whenever a maintainer has a moment.

Summary: when a user sends a top-level message that p-tags a known agent, the agent's hidden NIP-10 reply now auto-opens the originating thread panel, so the exchange stays visible instead of silently landing in a collapsed thread.

Scope is narrow by design — it never fires for human replies, replies to someone else's thread, broadcast replies, hydration/replay, or while another auxiliary panel is open. The policy is a pure function with unit coverage; the hook side has mounted-hook tests covering the deferred-apply path.

State: rebased on `main`, DCO / Semgrep OSS / zizmor all green, `pnpm run check`, `pnpm run typecheck`, and the full `pnpm test` suite (4726 tests) pass locally.

Happy to split it up, adjust the policy conditions, or rework anything that doesn't fit the intended UX.

PR #5569, opened 2026-08-11
Fix macOS attachment picker lifecycle and allow inert HTML downloads

wesbillman:
## Problem

Canceling the native macOS file chooser leaves the composer's temporary, detached `<input type="file">` without a `change` event or an explicit cleanup path. Opening Finder again immediately creates a second detached input while WebKit may still be unwinding the first picker. The newly selected files can therefore fail to reach the upload pipeline. Drag and drop is unaffected because it bypasses this picker lifecycle.

This does **not** add an automatic retry mechanism. “Retry” means the user's next attachment attempt after canceling or after a prior selection.

## Fix

- give each composer hook one hidden, body-mounted file input for its lifetime instead of creating a detached one per click
- reset and reconfigure that input before every open, replace its handler rather than stacking handlers, and remove it cleanly on unmount
- preserve normal selection, cancel then reopen, selecting the same file again, and multi-select behavior
- accept canonical `text/html` attachments while continuing to serve and render them strictly as inert downloads
- keep XHTML, SVG, JavaScript, and executable MIME types blocked

The picker change fixes the ownership/lifecycle bug at its source; it does not retry failed uploads, add delays, or mask errors.

## Testing

- mandatory pre-push gate: branch-skew, desktop typecheck/tests/check, Rust tests, and desktop Tauri checks passed on `ea5a97adf957803935b28d63d32f9f332cf65287`
- `cargo test -p buzz-media --lib` (110 passed)
- focused Biome check for the three picker files
- picker Playwright regression: cancel/no selection then reopen, select the same file again, and multiple selection (run on the source commit before integration)
- HTML live-relay response regression added as ignored E2E because it requires the S3-backed relay harness

## Manual verification

Playwright models cancellation with Chromium's `FileChooser.setFiles([])`; it cannot exercise the native macOS Finder panel/WebKit presentation lifecycle. Before merge, manually verify in the built macOS app:

1. select a PNG normally
2. cancel, then immediately reopen and select a PNG
3. select the same PNG on a subsequent attempt
4. multi-select two PNGs

### DIFF
--- a/crates/buzz-media/src/validation.rs
+++ b/crates/buzz-media/src/validation.rs
@@ -69,12 +69,23 @@ pub(crate) fn looks_like_mp4_iso_bmff(bytes: &[u8]) -> bool {
 /// neutralises them — this allowlist-of-denials is defence in depth, so a future
 /// header regression can't turn an uploaded blob into a stored-XSS vector.
 ///
-/// HTML, JS, and SVG are the classic stored-XSS carriers. Native executables are
+/// JS and SVG are the classic stored-XSS carriers. Native executables are
 /// blocked because there's no legitimate reason to host them inline in chat and
 /// they're a malware-distribution risk.
+///
+/// HTML is intentionally *not* blocked: it is accepted as an inert download
+/// (`serve_inline` returns false for `text/html`, so it is served with
+/// `Content-Disposition: attachment` + `nosniff` + `CSP: default-src 'none'`,
+/// and the desktop renderer never navigates a webview to a generic
+/// attachment). The old sniff-based block only caught the well-formed HTML
+/// `infer` recognises anyway — HTML that evades the sniff already uploaded as
+/// `application/octet-stream` and served as a download, so blocking canonical
+/// HTML was inconsistent rather than a real control. `application/xhtml+xml`
+/// stays listed as dormant defence in depth: `infer` has no XHTML matcher, so
+/// it is unreachable through sniffing, but the entry costs nothing and guards
+/// against a future detector that does classify it.
 const BLOCKED_FILE_MIME_TYPES: &[&str] = &[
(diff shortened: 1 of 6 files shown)


wesbillman (commented):
Reviewed on Wes's behalf at exact head `7d76393cb794324c1c1055acac0e4f35987f4049`. No actionable findings.

I traced both changes beyond the diff:

- The picker now owns one body-mounted input for the hook lifetime, replaces (rather than stacks) its callback, resets before open and after selection, and removes/nulls it on unmount. The queue/non-queue split remains intentional: only the deferred-video path uses the webview picker; the standard path retains the native Rust picker.
- Canonical `text/html` remains inert end to end. Upload MIME is sniffed on both desktop and relay; the sidecar is authoritative on serve and imeta ingestion; `serve_inline` allows only image/video; full and range GETs add attachment disposition, nosniff, and restrictive CSP; generic attachment rendering resolves to `FileCard`; and its click invokes the native same-relay `/media/` downloader rather than navigating the webview. Download bytes are re-sniffed against the mirrored deny-list before save. SVG, XHTML, JavaScript, and executables remain blocked.

Focused HTML acceptance/deny-list/executable tests passed locally on the exact clean head; `git diff --check` passed. GitHub reports the branch mergeable and all applicable CI/security/DCO checks green. The response-level HTML test is intentionally ignored because it requires the S3 relay harness, but relay integration CI is green.

The only remaining validation gap is the one disclosed in the PR: Playwright cannot reproduce the native macOS Finder/WebKit presentation lifecycle, so cancel → immediate reopen → same-file selection still deserves a manual packaged-app check before merge.

Conflict disclosure: Carl co-authored the final size-ratchet-only cleanup commit; this review therefore re-inspected the full PR independently rather than treating authorship as evidence.

Issue #5128, opened 2026-08-07
Agents tab is not community-scoped: card, profile name and public key belong to three different entities — and per-agent settings land on the wrong instance


eireal:
**Environment:** Buzz Desktop on macOS (a Windows machine is paired under the same owner identity). Two self-hosted communities: `leevia.communities.buzz.xyz` and `knowhow.communities.buzz.xyz`. Builtin agents Fizz / Honey / Bumble, custom agents Nemo and Oracle.

Relay-side state is correct throughout (`buzz channels members` and `buzz users presence` agree with expectations). Everything below is local to the desktop client.

## Identity model observed

On this Mac each **builtin** agent has one identity **per community**, created at different times:

| Persona | Mac · leevia | Mac · knowhow |
|---|---|---|
| Fizz | `bf667ab3…` (created 2026-07-30T22:23Z) | `9dd15fbb…` (created 2026-08-03T20:42Z) |
| Honey | `612eeaae…` | `56769d9e…` |
| Bumble | `f74495db…` | `0bb9f108…` |

Custom agents (Nemo, Oracle) instead reuse **one** key across both communities.

## 1. The Agents tab renders the same list in every community

Screenshots taken minutes apart, switching only the active community: the agent cards are identical (`Fizz PC`, `Honey PC`, `Bumble PC`). Only the channel list in the sidebar changes. There is no visual cue anywhere that these cards belong to a specific community.

## 2. Card name, profile name and public key are three different entities

Opening the card labelled **`Fizz PC`**:

- profile heading shows **`Fizz Mac`**
- `Public key` shows **`9dd15fbb…9dd8`** — the **knowhow** instance
- and it shows **the same `9dd15fbb…9dd8` while the active community is `leevia`**, where this persona's instance is `bf667ab3…` and its `buzz-acp` process is running.

Same defect on a custom agent: the card labelled **`Wally PC`** opens a profile titled **`Nemo Mac`** with key `758f820f…44b0`. (`Wally` is an agent created on the *Windows* machine; `Nemo Mac` is a different agent that only exists on the Mac. The two got merged into one persona record — see block/buzz#3204 for the sync side of this.)

The `PC` suffix comes from the local store: **persona** records (no `pubkey`) are named `Fizz PC` / `Honey PC` / `Bumble PC` / `Wally PC`, while **instance** records are named `Fizz Mac` / `Honey Mac` / `Bumble Mac` / `Nemo Mac`. The card renders the persona name, the profile renders the instance name.

[...]

PR #5575, opened 2026-08-11
fix(desktop): show every managed agent instance

atishpatel:
## Overview

**Category:** fix  
**User impact:** The Agents page now shows every managed agent instance that mention autocomplete can target, even when multiple instances share the same persona and display name.

**Problem:** Mention autocomplete preserves managed-agent identity by pubkey, but the Agents page grouped agents by persona and selected only one representative. Multiple same-persona instances therefore appeared in autocomplete while all but one were hidden from the Agents page.

**Solution:**
- Render one identity card per managed-agent pubkey within each persona group.
- Show a truncated pubkey on managed cards so identically named instances are distinguishable.
- Keep persona-only cards when no managed instance exists.
- Keep persona-level edit/share/delete actions on the first stably sorted instance rather than duplicating them on every runtime card.
- Preserve relay-restored managed agents, including the ingestion path changed by #4999.

## Reproduction steps

1. Configure or restore two managed agents with the same persona ID.
2. Confirm both identities appear in mention autocomplete.
3. Open the Agents page.
4. Before this fix, only one card appeared for that persona. With this fix, both pubkey-backed cards appear and can open their respective profiles.

## Validation

At commit `f2a0ecff82d36525c77e0ed7a2dd978a5785cdb6` with a clean working tree:

- Pre-push desktop check, typecheck, and full unit suite passed (4,616 tests).
- Purpose-built Playwright regression passed after a fresh E2E build.
- Focused grouping tests passed (4 tests), including relay-restored and persona-only cases.
- Mutation check passed: limiting the group back to one instance made the Playwright regression fail at the two-card assertion.
- `git diff --check` passed.

Originating conversation: Buzz channel `dcbc45d2-e9cc-45b3-940b-568dd06652ff`, thread `fd2e3ae323021b0314203dc0709f624669f278b344fdbe21f6cd65bae9582b2c`.

## Screenshot

Agents page with three managed Fizz instances sharing one persona. Each card is individually visible and distinguished by its truncated pubkey.

Agents page showing three same-persona managed instances

---------

Generated with Brainy Bumble

## Delete-popup UX proof

The individual-agent delete confirmation now makes the backup action visibly clickable with an underlined **Export this agent** link. Clicking it opens the existing export dialog. The seeded proof state shows 9 memories and 2 channels; both captures wait for the dialog animation to settle.

Individual-agent delete confirmation with underlined Export this agent link

Existing agent export dialog opened from the delete confirmation

### DIFF
--- a/desktop/src/features/agents/ui/AgentIdentityCard.tsx
+++ b/desktop/src/features/agents/ui/AgentIdentityCard.tsx
@@ -11,6 +11,7 @@ type AgentIdentityCardProps = {
   avatarUrl?: string | null;
   dataTestId: string;
   label: string;
+  identityLabel?: string | null;
   modelLabel?: string | null;
   onClick: () => void;
   /** Optional badge rendered below the label (e.g. "Restart required"). */
@@ -24,6 +25,7 @@ export function AgentIdentityCard({
   avatarUrl,
   dataTestId,
   label,
+  identityLabel,
   modelLabel,
   onClick,
   statusBadge,
@@ -72,6 +74,11 @@ export function AgentIdentityCard({
         <span className="min-w-0 truncate font-semibold text-foreground tracking-normal">
           {label}
         </span>
+        {identityLabel ? (
+          <span className="min-w-0 truncate font-mono text-2xs font-normal text-secondary-foreground/75">
+            {identityLabel}
+          </span>
+        ) : null}
         {modelLabel ? (
           <span className="min-w-0 truncate text-xs font-normal text-secondary-foreground/75">
             {modelLabel}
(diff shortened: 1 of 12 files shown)


ScaleLeanChris:
Validated this against a real multi-agent Buzz profile and the focused grouping suite.

The local managed-agent store contains 11 duplicated display names. Fizz, Bumble, and Honey each have two managed identities, so this is not only a synthetic collision. At `294d14f5a`, the focused grouping tests pass: 4 passed, 0 failed. The new grouping preserves every pubkey-backed instance and stable ordering.

The truncated pubkey is useful for choosing the correct identity. It does not identify the execution host, which is the remaining source of confusion in a multi-machine fleet. I do not think that should block this focused PR. Host/body placement needs a stable body registry and belongs in #5667.

One current CI failure is the repository security audit reporting yanked transitive crates (`async-utility` and `spin`), not a finding in this Desktop-only diff.

Czaruno:
@atishpatel I saw that you closed this on 2026-09-21. Can you tell me the reason?

This PR fixed a problem I have: two bodies under one persona show as a single card. I want to know if other work replaces it, or if the maintainers prefer a different approach. A pointer to the replacement is enough.

Context: #5667, #5128.

PR #5578, opened 2026-08-11
fix(observer): align NIP-44 plaintext limit

sbenedetto:
## Summary

- Align the observer plaintext budget with the NIP-44 v2 codec's actual 65,408-byte maximum.
- Add boundary regression coverage for the largest accepted payload and the first rejected payload.

The observer helper previously allowed JSON plaintext up to 65,535 bytes. The pinned NIP-44 implementation rejects anything above 65,408 bytes, so payloads in that 127-byte gap passed Buzz's preflight and then failed during encryption with `Nip44(V2(MessageTooLong))`. This caused oversized observer frames to be dropped instead of being trimmed or rejected by the shared Buzz budget.

With this change, callers using `OBSERVER_MAX_PLAINTEXT_LEN` trim and batch against the same limit enforced by encryption, and oversized payloads return the actionable `PlaintextTooLarge` error before entering NIP-44.

### Related issue

None found after searching open issues and PRs for the observer/NIP-44 limit mismatch.

### Testing

Test-first reproduction on the unmodified implementation:

- `cargo test -p buzz-core observer_payload_rejects_plaintext_above_nip44_limit_before_encrypting -- --nocapture`
- Failed as expected with `expected PlaintextTooLarge, got Nip44(V2(MessageTooLong))`.

Passing after the fix:

- `cargo test -p buzz-core observer::tests` — 4 passed
- `cargo test -p buzz-acp observer_publish_queue_tests` — 15 passed
- `cargo test -p buzz-acp observer_payload_trim_tests` — 8 passed
- Repository formatting, clippy, desktop/web/mobile static checks
- Remaining Rust unit suites — 7 passed
- Desktop unit tests and production frontend build
- Desktop Tauri compile check
- Web production build
- Mobile tests — 1,261 passed

`just ci` was attempted. On this machine, `buzz-voice` and `desktop-tauri-test` cannot link because Apple Clang 12 is older than the bundled Sherpa/ONNX macOS 13.4 objects (missing libc++/CoreAudio symbols). The same environment-only linker failure reproduces directly in `cargo test -p buzz-voice --no-run`; no voice or Tauri source is changed by this PR.

### DIFF
--- a/crates/buzz-core/src/observer.rs
+++ b/crates/buzz-core/src/observer.rs
@@ -21,8 +21,8 @@ pub const OBSERVER_FRAME_CONTROL: &str = "control";
 pub const NIP44_MIN_CONTENT_LEN: usize = 132;
 /// Maximum NIP-44 v2 ciphertext length.
 pub const NIP44_MAX_CONTENT_LEN: usize = 87_472;
-/// Maximum observer plaintext JSON size accepted by helpers.
-pub const OBSERVER_MAX_PLAINTEXT_LEN: usize = 65_535;
+/// Maximum observer plaintext JSON size accepted by the NIP-44 v2 codec.
+pub const OBSERVER_MAX_PLAINTEXT_LEN: usize = 65_536 - 128;
 
 /// Errors returned by observer payload encryption/decryption helpers.
 #[derive(Debug, Error)]
@@ -115,6 +115,12 @@ mod tests {
     use super::*;
     use nostr::{EventBuilder, Kind, Tag};
 
+    const NIP44_V2_MAX_PLAINTEXT_LEN: usize = 65_536 - 128;
+
+    fn json_string_with_serialized_len(len: usize) -> String {
+        "x".repeat(len - 2)
+    }
+
     #[test]
     fn observer_payload_round_trips_with_nip44() {
         let sender = Keys::generate();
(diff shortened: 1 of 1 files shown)


cristiansotogarciaxatech:
This PR has an issue now, and it did not when you opened it. #7086 was filed on 31 August from a Linux VM with the same failure, and the reporter got as far as "padding / off-by-one vs 65,535 is the likely mismatch" without finding this branch. You searched for an issue and found none, they searched for a fix and found none. You are both right and you are fixing the same thing.

Independent confirmation from a third machine, Windows, one box running managed agents with the relay observer on.

- **504** occurrences of `failed to encrypt relay observer event: NIP-44 error: message too long`
- across **15** of 18 agent log files
- window `2026-08-15T10:00:48Z` to `2026-09-03T04:32:12Z`, so it is still happening today
- every one of the 504 is that same variant. Not a single `PlaintextTooLarge` in the whole window, which is what you would expect when the preflight is measuring against a number the codec does not use.

Your diagnosis is not just plausible, it is the only thing those 504 lines can mean. To reach `nip44::encrypt` at all a payload has to get past `plaintext.len() > OBSERVER_MAX_PLAINTEXT_LEN` at `crates/buzz-core/src/observer.rs:64`, so it was at most 65,535 bytes. To come back as `MessageTooLong` it has to fail `len > MAX_SUPPORTED_PLAINTEXT_SIZE` inside `nostr`, so it was over 65,408. Every one of those events was in your 127-byte gap. There is no other route to that error from this call path.

I also checked that the constant is not going to move under you. The workspace pins `nostr = { version = "0.44", ... }` and `Cargo.lock` resolves `buzz-core` to `nostr 0.44.7`. Both 0.44.7 and 0.45.1 define `MAX_SUPPORTED_PLAINTEXT_SIZE: usize = 65_536 - 128`. A version bump does not fix this and 65,408 is the right number to align to.

Worth adding that `fit_observer_event_to_budget` makes this sharper than a plain boundary bug. It returns the moment `serialized_len(event) <= OBSERVER_MAX_PLAINTEXT_LEN`, so a frame that arrives at 65,500 bytes is never trimmed at all, and a trim loop that lands on 65,450 stops there and calls it done. The function that exists to prevent silent telemetry loss is aiming at the wrong target. I cannot tell from a log line which of those two paths any given event took, since the message carries no size, but both end in the same drop.

Open since 11 August with no reviews, and there is now a filed issue pointing at it. This deserves a look.

I am not opening a competing PR.

brandon-btcgroup:
I'm the reporter on #7086.

This PR is the encrypt-path fix we hit in production: host-route `buzz-acp` with `BUZZ_ACP_RELAY_OBSERVER=true`, Desktop Activity typing-only, VM logs `NIP-44 error: message too long` after the trimmer (not `PlaintextTooLarge`).

The 65,535 vs 65,408 mismatch in that ticket is this change. Asks 1 and 2 on #7086 are covered here. Ask 3 (Desktop surfacing observer loss instead of a hanging typing indicator) is separate and should stay on the issue.

Happy to re-test host-agent Activity once this lands.

PR #5579, opened 2026-08-11
fix(buzz-acp): exactly-once fallback publish for agents that don't self-publish

FvanW:
## Summary

`buzz-acp` harnesses require the wrapped ACP agent to self-publish its reply
via a `buzz messages send` tool call — the harness itself never
auto-publishes (per `base_prompt.md`: "your reasoning is invisible... exists
only if you published it"). Well-behaved backends do this reliably, but a
less rigorous backend can intermittently end its turn with a full text
answer via `agent_message_chunk` updates while never emitting the publish
tool call, silently dropping the reply.

This adds an opt-in, default-off `--publish-final-if-unsent` flag
(`BUZZ_ACP_PUBLISH_FINAL_IF_UNSENT`). When armed for a given backend:

- Buffers `agent_message_chunk` text per turn (bounded at 64 KiB, matching
  `buzz_sdk::builders::build_message`'s own content-size limit; only
  buffered at all when the flag is armed, so this is a zero-cost no-op for
  backends that never set it).
- Tracks whether the turn's tool calls included an actually-successful
  `buzz messages send` (checks `rawOutput.isError`, the same success/failure
  convention this repo's own `buzz-agent` reference implementation already
  uses — not just tool-call lifecycle status) to a matching `--channel`.
- On a clean `EndTurn` (never Refusal/MaxTokens/MaxTurnRequests/Cancelled —
  those are distinct terminal outcomes this must not paper over) for a
  channel-sourced turn, if no successful publish was detected and the
  buffered text is non-blank, publishes it via the same in-process
  signing/submit path `post_failure_notice` already uses, threaded to the
  turn's own `--reply-to` anchor (now shared with `format_prompt` via one
  extracted helper instead of two hand-synced copies).

This is a best-effort fallback with duplicate suppression on a heuristic
detection signal, not a delivery guarantee — documented as such rather than
"exactly-once."

Left default-off, and currently only enabled for one backend in our own
deployment config — arming it for a backend that already self-publishes
reliably would risk a double-post, so this should stay opt-in per-backend.

### Testing

- `cargo fmt --all -- --check` clean.
- `cargo build --release -p buzz-acp` succeeds.
- `cargo test -p buzz-acp`: 736 tests pass, 0 failures, including ~40 new
  tests covering: clean-EndTurn-fires, detected-publish-suppresses (both
  success- and failure-shaped tool results), every non-EndTurn stop reason
  excluded, heartbeat-sourced turns excluded, blank-text excluded, buffer
  armed/disarmed and capped correctly, heuristic positive/adversarial/
  wrong-channel cases, and reply-anchor threading (DM/top-level/threaded/
  agent-to-agent) — the last of which also confirms the anchor-resolution
  refactor preserves `format_prompt`'s pre-existing test suite unchanged.
- Live-verified against a running relay: 3 consecutive turns each produced
  exactly one correctly-threaded reply, including one turn with a genuine
  mid-turn steer signal that still resulted in exactly one reply with no
  duplicate. The fallback path itself has not yet been observed firing live
  (the bug it targets is intermittent) — the self-publish path was
  exercised, not yet the fallback path itself in production.

### Known residual limitations (documented in code)

- The `rawOutput.isError` success signal is only guaranteed for
  `buzz-agent`-shaped tool-call payloads; other backends may report
  `status: "completed"` with no `rawOutput`, in which case this falls back
  to trusting lifecycle status alone (pre-existing behavior) rather than
  guessing.
- The self-publish detector is a token/substring scan of the tool call's
  structured `command` field, not a real shell parser — a command built at
  runtime or hidden behind a wrapper script can still be missed.

### Related issue
None found — searched open issues/PRs for related discussion, didn't find one.

### DIFF
--- a/crates/buzz-acp/src/acp.rs
+++ b/crates/buzz-acp/src/acp.rs
@@ -8,6 +8,8 @@
 //! 4. [`AcpClient::session_prompt_with_idle_timeout`] — send prompt with idle/hard deadline, return stop reason
 //! 5. [`AcpClient::session_cancel`] / [`AcpClient::cancel_with_cleanup`] — cancel in-flight turn
 
+use std::collections::HashSet;
+
 use futures_util::StreamExt;
 use tokio::io::AsyncWriteExt;
 use tokio::process::{Child, ChildStdin, ChildStdout};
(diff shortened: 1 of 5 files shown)


johncarle:
This fix is blocking real onboarding right now: 15 Hermes-backed agent profiles brought onto Buzz are generating real replies that silently never reach their channels (confirmed via harness logs — the known `PromptOutcome::Ok` gap in `crates/buzz-acp/src/lib.rs`, only failed turns get published today). This PR's opt-in `BUZZ_ACP_PUBLISH_FINAL_IF_UNSENT` approach is the correct fix (avoids double-posting for agents that self-publish) — re-verified independently against `origin/main`.

@FvanW / @block/buzz-oss-team — could this get a review? Happy to help test against the Hermes onboarding case if useful.

alm36918:
## Independent live validation with Hermes ACP

I validated the transport-owned publication direction on current `main` with a Buzz-managed Hermes ACP runtime.

### Failure reproduced before fallback

- the ACP turn completed with non-empty final assistant text;
- an explicit prompt contract was not deterministic;
- an authenticated MCP reply tool registered successfully, but Hermes still selected its ordinary terminal;
- no valid managed-runner reply was delivered through that model-selected path.

### Harness-owned publication result

A local prototype collected `agent_message_chunk` output and, after the successful channel turn, built and submitted the reply through the harness `RestClient`, using routing from the accepted inbound batch rather than model-provided identifiers.

Live outcome:

- the fallback path itself fired end to end;
- exactly one threaded kind:9 reply appeared;
- the event was signed by the managed runner identity;
- the expected channel, reply, and sender `p` tags were present;
- no event was emitted by the unrelated profile identity.

Verification on the current tree:

- `cargo test -p buzz-acp --lib`: **804 passed, 0 failed**
- `cargo clippy -p buzz-acp --all-targets --all-features -- -D warnings`: passed
- formatting and release build passed

This fills the live-validation gap noted in this PR description: the fallback publication path was directly exercised, rather than only the self-publish suppression path.

### Scope/caveat

My minimal local prototype always published final text and therefore is not proposed as a competing implementation: it lacks this PR's opt-in gate, clean-`EndTurn` restriction, bounded buffer, and successful-send duplicate suppression. Those guards are necessary before upstream adoption. I am reporting the runtime evidence here instead of opening another duplicate PR.

All community identifiers, relay details, event IDs, agent names, prompts, local paths, and credentials were excluded.

PR #5582, opened 2026-08-11
fix(buzz-acp): treat hermes-acp as a zero-arg agent

mikeholownych:
## Summary
- \`default_agent_args()\` already special-cases other zero-arg ACP adapters (\`codex-acp\`, \`claude-agent-acp\`, \`buzz-agent\`), and \`default_agent_env()\` a few lines above already special-cases \`hermes\`/\`hermes-agent\`/\`hermes-acp\` by name — but \`default_agent_args()\` was missing the same three identities.
- Without this, the harness's clap default (\`"acp"\`) is passed straight through as a positional argument to \`hermes-acp\`, which doesn't expect one.

## Test plan
- [x] Built \`buzz-acp\` in release mode and ran it against a local relay with \`BUZZ_ACP_AGENT_COMMAND=hermes-acp\`; confirmed it spawns \`hermes acp\` with no stray positional argument and authenticates successfully (NIP-42 auth succeeded, presence set to online).

### DIFF
--- a/crates/buzz-acp/src/config.rs
+++ b/crates/buzz-acp/src/config.rs
@@ -701,7 +701,9 @@ fn default_agent_args(command: &str) -> Option<Vec<String>> {
     match normalize_agent_command_identity(command).as_str() {
         "goose" => Some(vec!["acp".to_string()]),
         "codex" | "codex-acp" | "claude-agent-acp" | "claude-code-acp" | "claude-code"
-        | "claudecode" | "buzz-agent" => Some(Vec::new()),
+        | "claudecode" | "buzz-agent" | "hermes" | "hermes-agent" | "hermes-acp" => {
+            Some(Vec::new())
+        }
         _ => None,
     }
 }


ScaleLeanChris:
Field-tested this against a real stock Hermes installation and an authorized private Buzz channel.

Evidence:

- Before this patch, the Hermes preset fails during ACP initialization with `hermes: error: unrecognized arguments: acp`.
- At `743eee74c`, the same identity, relay, and channel start successfully with no appended `acp` argument.
- `buzz-acp` logs `Starting hermes-agent ACP adapter`, completes ACP v2 initialization, creates a session, receives the channel prompt, and Hermes completes the turn.
- Focused config suite: 113 passed, 0 failed.

I temporarily stopped the supervised Hermes gateway to avoid a duplicate consumer during the test and restored it afterward. This validates the narrow compatibility claim: the bundled `hermes-acp` launcher is a zero-argument ACP command.

PR #5584, opened 2026-08-11
fix(desktop): suppress fresh focus-return query herd

wesbillman:
## Summary

- gate focus-return refetches for the seven query families measured firing together
- skip focus refetch for successful data younger than five minutes while preserving normal stale times, focused polling, and live invalidation
- force recovery for non-success, missing-timestamp, and future-timestamp states despite TanStack's secondary stale gate
- cover the policy through mounted `QueryObserver` focus transitions, including exceptional recovery states

## Runtime validation

Wes tested the candidate from its isolated production-relay worktree. Channel switching felt improved; the change removed one identified focus-return request herd, though broader desktop performance investigation remains open and this is not claimed as the sole fix.

## Validation

At exact pushed HEAD `b78a469eb2532fafc5f4c85f10df098d1a3e12a3`:

- focused focus-return policy tests: 6/6
- desktop unit suite: 4,631/4,631
- desktop TypeScript
- desktop production build
- touched Biome and diff/file-size checks
- pre-push hook: branch skew, desktop check, desktop typecheck, full desktop tests, Rust tests, and desktop Tauri checks

Independent adversarial review by Princess Donut: PASS at exact HEAD.

Originating conversation: Buzz channel `ed9bd748-4cd2-4b58-8dda-af18964a4131`, thread `9538cc1973f0d65be5bf68fa2fe25e609986f51ca5bc5ec1a32b6527b07ff636`.

### DIFF
--- a/desktop/src/features/agents/hooks.ts
+++ b/desktop/src/features/agents/hooks.ts
@@ -19,6 +19,7 @@ import {
 import { updateCachedChannelMemberDisplayName } from "@/features/channels/channelMemberProfileCache";
 import { evictUsersBatchEntries } from "@/features/profile/hooks";
 import {
+  refetchOnFocusWhenOld,
   useAppFocused,
   useFocusedRefetchInterval,
 } from "@/shared/lib/useDocumentVisible";
@@ -340,7 +341,7 @@ export function useRelayAgentsQuery(options?: { enabled?: boolean }) {
     // reconcile (kinds PERSONA/TEAM/MANAGED_AGENT), never for kind:10100. So we
     // keep polling but at a relaxed cadence and pause it while backgrounded.
     refetchInterval,
-    refetchOnWindowFocus: true,
+    refetchOnWindowFocus: refetchOnFocusWhenOld,
     enabled: options?.enabled,
   });
 }
@@ -364,7 +365,7 @@ export function useManagedAgentsQuery(options?: { enabled?: boolean }) {
         ? 5_000
         : false;
     },
-    refetchOnWindowFocus: true,
+    refetchOnWindowFocus: refetchOnFocusWhenOld,
   });
 }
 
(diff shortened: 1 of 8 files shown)


wesbillman (commented):
Reviewed thoroughly and adversarially on Wes's behalf at exact head `b78a469eb2532fafc5f4c85f10df098d1a3e12a3`.

**Verdict: no material findings. This is a defensible, narrowly scoped freshness policy—not a hack.**

I traced the seven query families through TanStack's focus predicate, their ordinary stale times, focused polling, live subscriptions/reconnect invalidations, mutation cache updates, and exceptional states. The change does not globally stretch cache freshness: it suppresses only focus-triggered fetches for recent successful data. Missing, pending, error, zero-timestamp, and future-timestamp states force recovery with `"always"`, which is important because a plain `true` remains subject to TanStack's secondary stale check.

The main residual tradeoff is explicit and acceptable: after five minutes unfocused, the seven families can still catch up together. Before that threshold, their focused polls resume at their existing 5s–5m cadences, and live invalidations remain immediate. If a measured >5-minute return still hitches, scheduling/coalescing that catch-up would be the next evidence-driven change; it is not grounds to broaden this patch speculatively.

Test change callout: this adds `focusReturnPolicy.test.mjs` (no existing tests modified). The mounted `QueryObserver` tests cover the real `focusManager` transition plus missing/error/future-clock recovery and the exact age boundary. I independently reran the 6 focused tests and desktop TypeScript successfully against the clean exact head above.

CI note: the current `Desktop E2E Integration (2/2)` failure occurs during Hermit bootstrap with `curl: (60) SSL certificate problem: self-signed certificate`, before project tests execute. I do not attribute that infrastructure failure to this diff, and this review does not waive required CI.

I am leaving a comment review rather than approval because I do not have explicit authority to approve this PR.

wesbillman:
Closing at Wes’s direction: the change has not been shown to resolve the reported sluggishness, so it should not remain proposed as a solution.

PR #5585, opened 2026-08-11
fix(cli): allow safe generic file uploads

samiriasbeck:
## Summary
- align CLI upload policy with Desktop and relay generic-file support
- preserve sanitized filenames in NIP-92 imeta tags
- render non-image/video attachments as file links

## Problem
The CLI accepted only JPEG, PNG, GIF, WebP, and MP4, while Desktop and the relay already support generic files through a deny-list. As a result, documents such as DOCX and PDF could not be sent through buzz-cli.

## Safety
- blocks the same active-content and executable MIME types as Desktop and relay
- strips path components and control characters from display filenames
- keeps existing 50 MB non-video and 500 MB video limits

## Tests
- generic documents, PDF, octet-stream, and HTML policy
- blocked XHTML, SVG, and executable MIME types
- filename sanitization and imeta preservation

Draft until CI and maintainer review complete.

### DIFF
--- a/crates/buzz-cli/src/client.rs
+++ b/crates/buzz-cli/src/client.rs
@@ -34,6 +34,10 @@ pub struct BlobDescriptor {
     /// Duration in seconds for video/audio (optional).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub duration: Option<f64>,
+    /// Original filename captured client-side. The content-addressed relay
+    /// does not retain it, but generic file cards need a stable display label.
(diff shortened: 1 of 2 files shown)


samiriasbeck:
Closing this draft in favor of #4880, which already covers the same CLI/relay policy alignment, filename sanitization, imeta metadata, and generic file-card rendering more completely. We independently reproduced the DOCX/PDF upload gap and will track validation on the existing contribution instead of splitting maintainer attention.

PR #5591, opened 2026-08-11
perf(desktop): coalesce read state localStorage persistence

wesbillman:
Follow-on to #5453/#5454's localStorage work — found while investigating app-slowness reports on a real profile.

## Problem

`ReadStateManager.persistLocalState()` serialized and rewrote **all three** read-state localStorage blobs (`buzz.channel-read-state.v2`, `.publishable.v1`, `.source-created-at.v1`) synchronously on every context advance. On a real profile (1,643 contexts, ~450K chars across the three blobs) this produced ~880KB of localStorage sqlite WAL growth per 30 seconds at idle, with writes every ~5s — steady main-thread serialization + sync IPC for no user-visible benefit. Observed WAL size on the affected profile: 94–114MB.

## Fix

- Local persistence coalesced behind a **1s trailing-edge timer**: a burst of N advances produces one `writeStoredReadState` (one write per blob).
- Pending dirty state **flushes synchronously** on `pagehide`, hidden `visibilitychange`, `destroy()`, and before each relay publish — disk is current before any relay event goes out.
- Hydration still persists immediately. Publish debounce (5s), merge logic, and blob formats unchanged (`DEBOUNCE_MS` renamed to `PUBLISH_DEBOUNCE_MS` only).

## Accepted residual

A hard kill (SIGKILL/power loss — not webview teardown) inside the 1s window loses ≤1s of local read-state advances; relay max-merge bounds the effect to a message flickering back unread. On the record per review.

## Validation

- `readStateManager.test.mjs`: fake-timer/mock-storage coverage — exactly one 3-blob write per burst (zero before the timer fires), hidden-flush cancels the timer and persists, hydrate persists immediately, pre-publish flush. Suite 26/26.
- Push gate at the pushed commit: desktop check, typecheck, full desktop unit suite 4,670/4,670.
- Independent adversarial FULL REVIEW: **APPROVE** at tree `371a02cf` (commit metadata rewritten afterward for attribution; tree identical) — all six `persistLocalState` call sites traced, lifecycle/leak checks (StrictMode remount, pubkey change), no external readers of the blob keys.

### DIFF
--- a/desktop/src/features/channels/readState/readStateManager.ts
+++ b/desktop/src/features/channels/readState/readStateManager.ts
@@ -381,6 +381,7 @@ export class ReadStateManager {
     unixTimestamp: number,
     options: { publishable: boolean },
   ): void {
+    if (this.destroyed) return;
     const current = this.effectiveState.get(contextId) ?? 0;
     if (unixTimestamp <= current) {
       if (!options.publishable || this.publishableContextIds.has(contextId)) {
@@ -486,6 +487,7 @@ export class ReadStateManager {
   }
 
   private async mergeEvents(events: RelayEvent[]): Promise<void> {
+    if (this.destroyed) return;
     // Collect all own blobs (keyed by slot d-tag) to union them all.
     // NIP-RS: multiple own-slot blobs must be max-merged, not winner-takes-all.
     const ownBlobsBySlot = new Map<
@@ -495,6 +497,7 @@ export class ReadStateManager {
 
     for (const event of events) {
       const parsed = await parseReadStateEvent(event, this.pubkey);
+      if (this.destroyed) return;
       if (!parsed) continue;
 
       this.maxFetchedCreatedAt = Math.max(
@@ -531,6 +534,7 @@ export class ReadStateManager {
     // d-tag coordinate. If so, rotate our slotId to avoid clobbering.
     for (const event of events) {
       const parsed = await parseReadStateEvent(event, this.pubkey);
+      if (this.destroyed) return;
       if (!parsed || parsed.dTag !== `read-state:${this.slotId}`) continue;
       if (parsed.blob.client_id !== this.clientId) {
         this.slotId = generateHex(16);
@@ -583,14 +587,13 @@ export class ReadStateManager {
   }
 
   private async handleIncomingEvent(event: RelayEvent): Promise<void> {
-    if (event.pubkey !== this.pubkey) return;
-    if (this.destroyed) return;
+    if (this.destroyed || event.pubkey !== this.pubkey) return;
     console.debug(
       `[ReadStateManager] incoming event=${event.id.substring(0, 8)}… created_at=${event.created_at}`,
     );
 
     const parsed = await parseReadStateEvent(event, this.pubkey);
-    if (!parsed) return;
+    if (!parsed || this.destroyed) return;
 
     this.maxFetchedCreatedAt = Math.max(
       this.maxFetchedCreatedAt,
@@ -633,6 +636,7 @@ export class ReadStateManager {
   }
 
   private schedulePublish(): void {
+    if (this.destroyed) return;
     if (this.debounceTimer !== null) {
       window.clearTimeout(this.debounceTimer);
     }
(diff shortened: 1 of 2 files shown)


desktop/src/features/channels/readState/readStateManager.ts
```diff
@@ -927,10 +939,35 @@ export class ReadStateManager {
     for (const [contextId, createdAt] of stored.contextSourceCreatedAt) {
       this.contextSourceCreatedAt.set(contextId, createdAt);
     }
-    this.persistLocalState();
+    this.writeLocalState();
   }
 
   private persistLocalState(): void {
+    if (this.localPersistTimer !== null) {
+      window.clearTimeout(this.localPersistTimer);
```
wesbillman: [P1] Do not reset this timer on every advance. This makes local persistence a pure trailing-edge debounce, so a stream of read advances less than one second apart postpones the write for the entire stream. Sustained scrolling can therefore leave minutes of read positions only in memory, and a hard kill loses the whole burst—not the <=1 second accepted residual documented in the commit/PR. The relay publish timer has the same trailing-edge behavior at five seconds, so it does not bound this gap. Use a throttle/max-wait: once dirty, keep the existing timer so the latest in-memory snapshot is written no later than one second after the first unsaved advance, then open another window for later advances. Add a fake-clock test that advances a context every <1s for several seconds and proves persistence still occurs at least once per second; the current fake timer has no due times and `runAll()` cannot distinguish bounded throttling from this unbounded debounce.

wesbillman (commented):
Reviewing on Wes's behalf. GitHub will not allow this account to submit CHANGES_REQUESTED on its own PR, so this is a blocking review comment instead. The coalescing direction is sound, but the implementation does not preserve the PR's stated <=1 second data-loss bound: it is an unbounded trailing-edge debounce under sustained activity. Please switch to a bounded throttle/max-wait and add a time-aware regression test.

desktop/src/features/channels/readState/readStateManager.ts
wesbillman: Fixed in cb1b0c052. The timer is now a bounded max-wait throttle: the first dirty advance opens a single 1s window (later advances no longer reset it), the latest in-memory snapshot is written when it fires, and the next advance opens a fresh window — so persistence lands no later than 1s after the first unsaved advance under sustained activity. Added the time-aware regression test you asked for: the fake clock now tracks due times, and a new test advances a context every 200ms for 3.2s and asserts one three-blob persist per 1s window (plus verifies the final state written and that the trailing advance opens the next window). The original sub-1s burst test still proves exactly one persist. Focused suite 27/27; full desktop unit suite 4,671/4,671 at the pushed tree.

desktop/src/features/channels/readState/readStateManager.ts
```diff
@@ -927,10 +939,33 @@ export class ReadStateManager {
     for (const [contextId, createdAt] of stored.contextSourceCreatedAt) {
       this.contextSourceCreatedAt.set(contextId, createdAt);
     }
-    this.persistLocalState();
+    this.writeLocalState();
   }
 
   private persistLocalState(): void {
+    if (this.localPersistTimer !== null) return;
```
wpfleger96: [P1] Refuse to schedule persistence after destruction. `destroy()` clears the timer that exists at teardown, but an `initialize()` or publish fetch can already be in flight; when it resolves, its existing `persistLocalState()` call reaches this method and creates a new timer even though the manager is destroyed. On a React remount/pubkey-session replacement, the new manager can then write fresh state to the same keys, and one second later this stale manager overwrites it with its older snapshot (I reproduced that sequence at this head). Add a `this.destroyed` guard here and ensure awaited fetch paths do not mutate after teardown, then cover the deferred-fetch → destroy → replacement-manager case.

wpfleger96 (changes requested):
The max-wait persistence logic itself is sound, but lifecycle cleanup is incomplete: a destroyed manager can still arm a local-persist timer after an in-flight fetch resolves and later overwrite the active manager's state.

wesbillman:
Addressed the remaining lifecycle review in 0d7d5ee958. Destroyed managers now reject new local advances/persist scheduling, and async merge paths bail after awaited parsing if teardown occurred. Added a deferred-initialize-fetch regression covering destroy → replacement manager persist → stale fetch resolution; the replacement\u2019s newer local state is preserved and no stale timer is armed.

Validation on exact pushed head `0d7d5ee9582f77b5b11d2215fc1d1b2f928c83ef`:
- focused read-state suite: 28/28
- Biome on both touched files
- desktop typecheck
- pre-push gate: branch skew, desktop check/typecheck/full unit suite, Rust tests, and desktop Tauri checks all passed.

PR #5599, opened 2026-08-11
fix(desktop): remove 0.5.9+ perf regressions, speed up get_channels

wpfleger96:
Desktop input latency regressed sharply for users on v0.5.9 and worsened on latest main: multi-second stalls when clicking back into the app, slow fresh boots, intermittent lockups, and scroll/mouse degradation. Reverting to `119a84897` (pre-0.5.9) was confirmed to resolve it, isolating the regression to that range. Profiling a live production renderer plus a commit-level audit of the range found three independent, additive causes — fixed here — plus a long-standing `get_channels` cost that made every remaining refetch expensive, also addressed here.

## 1. Focus-return refetch storm (`refetchOnWindowFocus`)

#5490 wired TanStack's `focusManager` to app focus and flipped ~20 query sites to `refetchOnWindowFocus: true`. A focus return after >60s away fires them all within milliseconds — and a click into an unfocused window *is* a focus return, so the burst runs before the click is processed. That is the "click into the composer, wait 5 seconds" symptom, and it also explains why mouse input feels worse than keyboard (clicks arrive with focus transitions; typing happens while already focused). A 5-second `sample` of a live production renderer caught a single window activity-state transition consuming ~1.25s of main-thread time, dominated by `JSON.parse` in the focus listener's microtask drain.

#5535 already established the fix pattern but applied it to only two families (channels, home-feed). This PR extends the same 5-minute `staleTime` discipline to the remaining families: pulse (×5), workflows (×4), agents (×4), forum (×2), presence, user-status, custom-emoji, channel-templates, and the persona catalog. Polling cadences and push-invalidation paths are untouched — interval refetches and `invalidateQueries` both bypass `staleTime`, so live-update behavior is unchanged. Each gated family exports its focus-refetch policy as an options object that the production hook spreads into `useQuery`, and a `focusRefetchPolicy.test.mjs` drives a `QueryObserver` with that same production object — locking the policy behaviorally (fresh focus return → 0 fetches; stale → refetch) and failing if a hook's `staleTime`/`refetchOnWindowFocus` wiring drifts.

Four families deliberately keep tighter freshness, all surfaces where the 5-minute gate would suppress the only refresh path and none of which feed the app-wide storm: `repo-sync-status` keeps its fresh focus refetch (its inline comment documents the "committed in a terminal, switched back to the app" flow as intended); the workflow-runs list stale-gates at 10s because a remotely-started run has no push invalidation and its conditional 1s poll is off while the cache shows no active runs; the workflow list queries (`useChannelWorkflowsQuery` and the all-channels aggregate) stale-gate at 10s because they have no poll and no relay subscription, and mutation-driven invalidation only covers this renderer — remote workflow creates/edits/deletes surface only via focus refetch; and the managed-agent log stale-gates at one poll tick (30s) so returning to a live agent log refreshes immediately. Run approvals keep the 5-minute gate under `RUN_APPROVALS_FOCUS_STALE_TIME_MS` — their focused 10s poll already covers freshness.

## 2. Synchronous localStorage sweep on the boot/focus path

#5453's stale-cache sweep synchronously `getItem` + `JSON.parse`s every whitelisted localStorage entry on the main thread (multi-MB on seasoned profiles), scheduled with a `requestIdleCallback` timeout of 1.5s that guaranteed it landed mid-boot, and re-armed on every hidden→visible transition — stacking it onto the exact moment the focus storm fires. #5454's `trimSelfProfileCaches()` additionally scanned every localStorage key on every `writeSelfProfileCache()` call (which fires per relay self-profile delivery at boot).

Now: the first sweep waits `BOOT_SWEEP_FLOOR_MS` (30s) after startup, the scan is time-sliced across idle callbacks, and the visibility trigger is removed — boot-delayed plus hourly still covers the 14-day TTL contract. The sliced sweep re-checks staleness immediately before each removal (a key rewritten fresh mid-sweep survives), isolates per-key storage errors so one bad entry can't strand the rest of the snapshot, defers oversized values once rather than parsing them on a zero-budget slice, guarantees forward progress on timeout-fired callbacks, and cancels its scheduled slice when stopped. The profile trim keeps a lazily-initialized memoized key count so the common under-cap write is O(1); the full parse scan runs only when the count exceeds a cap, resyncs if external deletions made it stale, and a failed scan skips the trim instead of aborting the write. Sweep semantics (rules, TTLs, eviction) are unchanged, and tests cover the scheduling, slice-progress, error-isolation, defer-once, and trim short-circuit behaviors.

## 3. The macOS window was never opaque

#5478's glass appearance is correctly opt-in at the CSS layer, but the compositor cost was baked in deeper than its native `on_webview_ready` transparency call: the main window is declared `"transparent": true` in `tauri.conf.json` (added for the original glass work in #1671), which makes tao call `NSWindow.setOpaque(false)` at creation and resolve every later `set_background_color(None)` to `clearColor` — and no runtime `setOpaque(true)` path exists through tauri, while wry's runtime background setter can only force the WKWebView's `drawsBackground` off, never back on. So "restore the platform default" was unreachable: every launch, glass or not, ran with a non-opaque NSWindow, defeating WindowServer's opaque-window compositing fast path and forcing full window compositing every frame — compounded by the existing `backdrop-blur` chrome overlapping the scrolling timeline. This matches the compositor-shaped symptoms (scroll and pointer input degrading first).

The window is now created opaque (`"transparent": false`) and the NSWindow layer is never made transparent at runtime. Glass never needed a transparent window: behind-window `NSVisualEffectView` vibrancy renders inside opaque windows (this is how Finder and Notes draw vibrant sidebars); it only requires a transparent WKWebView canvas, which the `set_window_vibrancy` enable path already establishes at runtime (`macos-private-api` compiles that in independent of the window flag). Enabling glass installs the vibrancy layer and then makes only the webview canvas see-through; disabling clears the vibrancy layer — the canvas may stay non-drawing afterwards (wry's flag is one-way at runtime), which is harmless because glass-off CSS paints fully opaque above an always-opaque NSWindow. The boot-path first-frame backing writes touch only the NSWindow backing color and are therefore inert to glass state regardless of how they order against the `ThemeProvider`'s vibrancy call on a persisted-glass-on cold boot. Glass-off users (the default) get an end-to-end opaque window from boot for the first time.

## 4. `get_channels`: serial round-trips and a multi-MB payload on every refetch

The stale gates in (1) cut refetch frequency; this cuts the cost of the refetches that legitimately remain (boot, and focus returns after more than 5 minutes away — previously still a multi-second stall). `get_channels` made ~8 fully serial relay round-trips (~3.2–3.6s at 1,100+ channels), then shipped the full `ChannelInfo` list — including every channel's member pubkeys — across IPC, where the renderer's `JSON.parse` of the multi-MB payload froze the main thread (the ~1.25s stall captured in the live sample).

- **Concurrent stages**: the membership chain, the open-channel directory scan, and the hidden-DM snapshot run concurrently, as do the member-count and last-message queries that follow. The critical path drops from ~8 sequential round-trips to 2 phases. Filters, limits, pagination, and merge semantics are unchanged.
- **Not-modified short-circuit**: the command now takes a client-supplied content hash (FNV-1a 64 over the channel list, canonicalized by id and excluding `last_message_at`) and omits the channel list from the response when nothing else changed. Last-message timestamps — which change on nearly every message anywhere — ship as a small separate map that the client overlays onto its cached list with reference preservation, so React Query's structural sharing also skips downstream re-renders. On a typical refocus the renderer parses kilobytes instead of megabytes. The hash is stored in the query cache itself, tying its lifecycle to the data it describes so a community switch can never leak a stale hash.

The E2E mock bridge speaks the new payload shape — including the complete `last_messages` map the client treats as authoritative — and hash canonicalization plus overlay reference-preservation are unit-tested on both sides.

### DIFF
--- a/desktop/src-tauri/src/commands/window_vibrancy.rs
+++ b/desktop/src-tauri/src/commands/window_vibrancy.rs
@@ -35,13 +35,15 @@ pub fn set_window_vibrancy(
             .ok_or_else(|| "main window not found".to_string())?;
 
         if !enabled {
-            clear_vibrancy(&window).map_err(|e| e.to_string())?;
-            // Restore the default opaque background so WindowServer can use its
-            // opaque-window fast path. `None` resets both the NSWindow backing
-            // and the WKWebView canvas to their platform defaults.
+            // Restore opaque backing first: a failure of clear_vibrancy leaves
+            // the window with an opaque canvas and a blur layer hidden behind
+            // it, not a see-through webview. Either mixed state self-corrects
+            // on the next toggle. `None` resets both the NSWindow backing and
+            // the WKWebView canvas to their platform defaults.
             window
                 .set_background_color(None)
                 .map_err(|e| e.to_string())?;
+            clear_vibrancy(&window).map_err(|e| e.to_string())?;
             return Ok(());
         }
 
@@ -64,6 +66,9 @@ pub fn set_window_vibrancy(
         // clear is a no-op (returns `false`) when none is present.
         let _ = clear_vibrancy(&window);
 
+        // Install the blur layer first: a failure of set_background_color leaves
+        // the window opaque with vibrancy behind it, not a see-through webview.
+        // Either mixed state self-corrects on the next toggle.
         apply_vibrancy(&window, material, None, None).map_err(|e| e.to_string())?;
 
         // Make the WKWebView canvas and NSWindow backing transparent so native
--- a/desktop/src-tauri/src/commands/window_vibrancy.rs
+++ b/desktop/src-tauri/src/commands/window_vibrancy.rs
@@ -1,15 +1,29 @@
 //! Runtime macOS window vibrancy (blur-behind) toggle.
 //!
+//! **Invariant:** the main window is created opaque (`tauri.conf.json`
+//! `transparent: false`) and the NSWindow is never made transparent at runtime.
+//! Behind-window vibrancy renders correctly inside opaque windows — this is
+//! exactly how Finder and Notes render vibrant sidebars — so glass only requires
+//! runtime webview-canvas transparency, which this command sets on the enable
+//! path. When glass is disabled the webview canvas may remain non-drawing
+//! (wry's `drawsBackground` flag is one-way at runtime), but that is harmless:
+//! glass-off CSS paints the full background opaque and the always-opaque NSWindow
+//! is beneath it.
+//!
+//! Why not `transparent: true`? A creation-time transparent window causes tao to
+//! call `NSWindow.setOpaque(false)` and `setBackgroundColor(clearColor)`. The
+//! runtime `Window::set_background_color(None)` then resolves `None` to
+//! `clearColor` instead of the opaque system default — and there is no runtime
+//! `setOpaque(true)` path through tauri — leaving the compositor blending the
+//! whole window even with glass off.
+//!
 //! Vibrancy applies an `NSVisualEffectView` behind the webview so the desktop
-//! (and windows behind Buzz) blur through wherever the app's CSS is
+//! (and windows behind Buzz) blurs through wherever the WKWebView canvas is
 //! transparent. It is a native, macOS-only effect: there is no "intensity"
 //! setting at the OS level, only a set of material presets. The frontend tunes
-//! perceived intensity by changing CSS surface opacity while this command
-//! handles the native material.
-//!
-//! This is fully reversible at runtime: enabling applies the chosen material,
-//! disabling clears it. On non-macOS platforms the command is a no-op so the
-//! shared frontend can call it unconditionally.
+//! perceived intensity by adjusting CSS surface opacity while this command
+//! handles the native material. On non-macOS platforms the command is a no-op
+//! so the shared frontend can call it unconditionally.
 
 #[cfg(target_os = "macos")]
 use tauri::Manager;
@@ -35,14 +49,16 @@ pub fn set_window_vibrancy(
             .ok_or_else(|| "main window not found".to_string())?;
 
         if !enabled {
-            // Restore opaque backing first: a failure of clear_vibrancy leaves
-            // the window with an opaque canvas and a blur layer hidden behind
-            // it, not a see-through webview. Either mixed state self-corrects
-            // on the next toggle. `None` resets both the NSWindow backing and
-            // the WKWebView canvas to their platform defaults.
-            window
-                .set_background_color(None)
-                .map_err(|e| e.to_string())?;
+            // The NSWindow layer is permanently opaque, so no window-layer
+            // reset is needed here. Skipping `set_background_color(None)` at
+            // the webview layer also avoids tauri mapping `None` to opaque
+            // white, which would still force `drawsBackground=false` on the
+            // WKWebView (counterproductive). After a glass session the webview
+            // canvas may stay non-drawing — wry's `drawsBackground` flag is
+            // one-way at runtime — but that is harmless: glass-off CSS paints
+            // the full background opaque and the always-opaque NSWindow is
+            // beneath it. If `clear_vibrancy` fails, the opaque CSS already
+            // covers everything, so no see-through state is reachable.
             clear_vibrancy(&window).map_err(|e| e.to_string())?;
             return Ok(());
         }
@@ -66,15 +82,19 @@ pub fn set_window_vibrancy(
         // clear is a no-op (returns `false`) when none is present.
         let _ = clear_vibrancy(&window);
 
-        // Install the blur layer first: a failure of set_background_color leaves
-        // the window opaque with vibrancy behind it, not a see-through webview.
-        // Either mixed state self-corrects on the next toggle.
+        // Install the blur layer first: a failure of the canvas write leaves
+        // the window with vibrancy behind an opaque webview, not a see-through
+        // one. Either mixed state self-corrects on the next toggle.
         apply_vibrancy(&window, material, None, None).map_err(|e| e.to_string())?;
 
-        // Make the WKWebView canvas and NSWindow backing transparent so native
-        // vibrancy shows through. Must follow apply_vibrancy so the blur layer
-        // is present before the canvas becomes see-through.
-        window
+        // Make only the WKWebView canvas transparent so native vibrancy shows
+        // through; the NSWindow layer stays opaque by design. Targeting the
+        // webview layer directly (via `AsRef<Webview>`) avoids the
+        // `WebviewWindow::set_background_color` path, which also writes the
+        // NSWindow layer. Must follow `apply_vibrancy` so the blur layer is
+        // present before the canvas becomes see-through.
+        let webview: &tauri::Webview<_> = window.as_ref();
+        webview
             .set_background_color(Some(tauri::window::Color(0, 0, 0, 0)))
             .map_err(|e| e.to_string())?;
 
--- a/desktop/src-tauri/src/commands/window_vibrancy.rs
+++ b/desktop/src-tauri/src/commands/window_vibrancy.rs
@@ -36,6 +36,12 @@ pub fn set_window_vibrancy(
 
         if !enabled {
             clear_vibrancy(&window).map_err(|e| e.to_string())?;
+            // Restore the default opaque background so WindowServer can use its
+            // opaque-window fast path. `None` resets both the NSWindow backing
+            // and the WKWebView canvas to their platform defaults.
+            window
+                .set_background_color(None)
+                .map_err(|e| e.to_string())?;
             return Ok(());
         }
 
@@ -59,6 +65,14 @@ pub fn set_window_vibrancy(
         let _ = clear_vibrancy(&window);
 
         apply_vibrancy(&window, material, None, None).map_err(|e| e.to_string())?;
+
+        // Make the WKWebView canvas and NSWindow backing transparent so native
+        // vibrancy shows through. Must follow apply_vibrancy so the blur layer
+        // is present before the canvas becomes see-through.
+        window
+            .set_background_color(Some(tauri::window::Color(0, 0, 0, 0)))
+            .map_err(|e| e.to_string())?;
+
         Ok(())
     }
 
--- a/desktop/src/features/workflows/hooks.ts
+++ b/desktop/src/features/workflows/hooks.ts
@@ -18,10 +18,12 @@ import {
   updateWorkflow,
 } from "@/shared/api/tauriWorkflows";
 
-/** Suppresses focus refetches until workflow data is genuinely stale.
- * Channel workflows and run approvals have push-invalidation from mutations;
- * workflow runs poll when active (making the focus refetch redundant). */
-export const WORKFLOWS_FOCUS_STALE_TIME_MS = 5 * 60_000;
+/** Stale gate for workflow list queries (useChannelWorkflowsQuery, allWorkflowsQuery).
+ * These queries have no poll and no relay push subscription; invalidateWorkflowListQueries
+ * only fires for mutations performed by this renderer, so remote workflow creates/edits/deletes
+ * surface only via refetch on focus return.  10 s suppresses rapid focus-flip burst refetches
+ * while keeping remote changes visible on the next focus return. */
+export const WORKFLOW_LIST_FOCUS_STALE_TIME_MS = 10_000;
 /** Keeps focused polling for run approvals at the established 10-second cadence. */
 export const RUN_APPROVALS_REFETCH_INTERVAL_MS = 10_000;
 /** Tighter stale gate for the workflow-runs detail query.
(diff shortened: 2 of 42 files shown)


desktop/src-tauri/src/commands/window_vibrancy.rs
```diff
@@ -36,6 +36,12 @@ pub fn set_window_vibrancy(
 
         if !enabled {
             clear_vibrancy(&window).map_err(|e| e.to_string())?;
+            // Restore the default opaque background so WindowServer can use its
+            // opaque-window fast path. `None` resets both the NSWindow backing
+            // and the WKWebView canvas to their platform defaults.
+            window
+                .set_background_color(None)
```
wesbillman: [P1] `None` does not restore an opaque macOS window in this app. `tauri.conf.json` still creates the main window with `"transparent": true`; Tao records that flag, calls `NSWindow.setOpaque(false)` at creation, and its `set_background_color(None)` implementation resolves `None` to `NSColor.clearColor()` whenever that flag is true (`tao-0.35.3/src/platform_impl/macos/window.rs:543-560,908-926`). It never calls `setOpaque(true)`. Wry also configured WKWebView with `drawsBackground = false` at creation for a transparent webview, while the runtime setter only writes `underPageBackgroundColor`; it does not restore `drawsBackground` (`wry-0.55.1/src/wkwebview/mod.rs:367-384,943-987`). The same `None` call runs again 250ms after reveal in `initial_window.rs:27-30`. So the default glass-off launch remains a non-opaque NSWindow/WKWebView despite this comment, and the advertised compositor fix is not achieved. Make opacity genuinely runtime-switchable (including `NSWindow.setOpaque(true/false)` and WKWebView `drawsBackground`, or stop constructing the default window as transparent), then verify both glass states on macOS.

desktop/src/features/workflows/hooks.ts
```diff
@@ -53,7 +60,7 @@ export function useChannelWorkflowsQuery(channelId: string | null) {
     queryFn: ({ queryKey: [, resolvedChannelId] }) =>
       getChannelWorkflows(resolvedChannelId),
     enabled: channelId !== null,
-    staleTime: 30_000,
+    staleTime: WORKFLOWS_FOCUS_STALE_TIME_MS,
```
wesbillman: [P1] This five-minute stale gate suppresses the only automatic refresh for workflow lists. `useChannelWorkflowsQuery` has no interval or relay subscription, and the aggregate query in `WorkflowsView.tsx:86-109` is the same. The invalidations at `hooks.ts:39-46,116-149` cover only mutations performed by this renderer; changes from another member/device do not reach these caches. Returning to the app within five minutes now skips the focus fetch and can hide remote workflow creates/edits/deletes until the cache ages out (and an additional trigger occurs). Keep the short stale window for the non-polled list queries, add focused polling, or add real live invalidation before applying this policy to them.

desktop/src/shared/lib/localStorageSweep.ts
```diff
@@ -103,53 +118,129 @@ export function sweepStaleLocalStorage(now = Date.now()): number {
...
+
+      if (pos < pendingKeys.length) {
+        scheduleSlice(processSlice);
+        return;
+      }
+
+      // All entries processed — batch-remove stale keys.
+      for (const key of staleKeys) {
+        try {
+          storage.removeItem(key);
```
wesbillman, line 297: [P1] Revalidate before removal. The new chunked algorithm reads a value in an early slice (`:180`), records only its key as stale (`:187-188`), can spend many later callbacks finishing the snapshot, and then removes that key here without checking whether it was rewritten in the meantime. Any normal cache refresh between those points turns this into deletion of fresh data. The old synchronous scan had only a tiny same-task window; chunking deliberately widens it across callbacks. Store the observed raw value/version and remove only if the current value still matches, or re-read and re-evaluate `updatedAt` immediately before deletion. Add a test that refreshes a previously classified key between slices.

wesbillman (changes requested):
Reviewing on Wes's behalf. I found three correctness blockers in the attempted performance fixes: the macOS disable path does not actually restore opacity, workflow lists lose their only useful remote-refresh behavior, and the chunked sweep can delete a cache entry refreshed while the sweep is in progress. CI is green, but none of these lifecycle/concurrency cases is covered.

desktop/src-tauri/src/commands/window_vibrancy.rs
```diff
@@ -59,6 +65,14 @@ pub fn set_window_vibrancy(
         let _ = clear_vibrancy(&window);
 
         apply_vibrancy(&window, material, None, None).map_err(|e| e.to_string())?;
+
+        // Make the WKWebView canvas and NSWindow backing transparent so native
+        // vibrancy shows through. Must follow apply_vibrancy so the blur layer
+        // is present before the canvas becomes see-through.
+        window
+            .set_background_color(Some(tauri::window::Color(0, 0, 0, 0)))
```
wesbillman, line 98: [P1] Even after implementing a real opacity toggle, this enable can be clobbered during a glass-on cold boot. `ThemeProvider` invokes this around mount, but `initial_window::clear_initial_window_backing` independently waits until the render-ready reveal, sleeps 250ms, then calls `set_background_color(None)` (`initial_window.rs:27-30`). That later write targets the same WebviewWindow and will normally land after this transparent write, leaving startup glass opaque until another preference toggle. The old unconditional `on_webview_ready` transparency made the webview independently transparent; removing it exposes the ordering bug. Coordinate initial-backing cleanup with the glass preference/state (or apply the final glass state after cleanup) and cover/verify a persisted glass-on macOS cold boot.

wesbillman (commented):
Additional lifecycle finding from the independent Mongo pass, posted on Wes's behalf:

wesbillman (commented):
Re-review on behalf of Wes at current head `7e05d4b9c`: three original P1s remain unresolved.

1. **Glass-off still is not restored to an opaque native window/webview.** `tauri.conf.json` still constructs the window with `transparent: true`. Tao consequently calls `NSWindow.setOpaque(false)` once at construction (`tao-0.35.3/.../macos/window.rs:544-546`) and its runtime `set_background_color(None)` resolves to `clearColor()` whenever that immutable `transparent` flag is true (`:908-925`). Wry's macOS runtime background setter also unconditionally sets private `drawsBackground=false` and never restores it (`wry-0.55.1/.../wkwebview/mod.rs:943-983`). Reordering `set_background_color(None)` before `clear_vibrancy` (`window_vibrancy.rs:38-48`) improves partial-failure behavior but does not make glass-off opaque or restore WindowServer's opaque fast path. The headline compositor fix remains unimplemented.

2. **Persisted glass-on cold boot is still clobbered.** The change only adds comments. `ThemeProvider` invokes glass from its passive effect (`ThemeProvider.tsx:562-568`), while the native startup task waits for initial render, reveals, sleeps 250 ms, then calls `set_background_color(None)` (`initial_window.rs:28-33`; `lib.rs:164-183`). That later write targets the same window/webview backing and can overwrite the transparent state established by `set_window_vibrancy`. The new comment calls timing "in practice" sufficient, but does not establish ordering; in fact the explicit post-reveal sleep preserves the adverse ordering. Persisted glass-on needs deterministic sequencing or a real macOS cold-boot demonstration proving the state survives.

3. **Remote workflow-list freshness is unchanged.** The fix tightens `useWorkflowRunsQuery`, not the two workflow-list queries in the original finding. `useChannelWorkflowsQuery` and `WorkflowsView` still use a five-minute stale gate with no poll/subscription, so another member/device's create/edit/delete can remain hidden after focus return. See `desktop/src/features/workflows/hooks.ts:67-75` and `desktop/src/features/workflows/ui/WorkflowsView.tsx:86-109`.

The sweep refresh/delete race is fixed correctly: removal now re-reads and re-evaluates `updatedAt`, with a focused regression test. The new focus-policy tests still synthesize `QueryObserver` options instead of exercising the production hooks/options, so the wiring gap remains (P2).

Please keep this review blocking. The fix commits repaired one trapdoor and labeled the others.

desktop/src-tauri/src/commands/window_vibrancy.rs
wpfleger96: 🤖 good catch — this was broken exactly as described. `tauri.conf.json` created the window with `transparent: true` (added back in #1671 for glass), so tao marked the NSWindow non-opaque at creation and `set_background_color(None)` resolved to `clearColor`; there's no runtime `setOpaque(true)` path through tauri, and wry's runtime background setter can only force `drawsBackground=false`, never restore it.

I took the second option you suggested: the window is now created opaque (`transparent: false`) and the NSWindow is never made transparent at runtime (dbe7ded6d). Glass doesn't actually need a transparent NSWindow — behind-window `NSVisualEffectView` renders inside opaque windows (Finder-sidebar style); it only needs the webview canvas transparent, which the enable path already sets at runtime via wry's setter (`drawsBackground=false` + clear `underPageBackgroundColor`, compiled in via `macos-private-api` independent of the window flag). The disable path is now just `clear_vibrancy`: after a glass session the webview canvas stays non-drawing (wry's flag is one-way at runtime), but glass-off CSS paints fully opaque over an always-opaque NSWindow, so the compositor sees an opaque window either way.

One caveat: CI can't see native vibrancy, so both glass states + a glass-on cold boot still need a quick manual eyeball on a macOS branch build — I'll do that pass.

desktop/src/features/workflows/hooks.ts
wpfleger96: 🤖 yeah — the list queries were the one family where the 5-min gate removed the only refresh path, since `invalidateWorkflowListQueries` only covers this renderer's own mutations. `useChannelWorkflowsQuery` and the `WorkflowsView` aggregate now use a 10 s gate (same shape as the runs fix), so focus-flip bursts stay suppressed but remote creates/edits/deletes surface on the next focus return (9f50cbfa8). Run approvals keep the 5-min gate since their focused 10 s poll already keeps them fresh — I renamed that constant to `RUN_APPROVALS_FOCUS_STALE_TIME_MS` so it can't quietly get reused for non-polled queries again. Behavioral tests added for both gates.

desktop/src/shared/lib/localStorageSweep.ts
wpfleger96: 🤖 this one crossed with the review — it was fixed in 65e0359312: the batch-removal loop re-reads each key and re-evaluates staleness against the rule TTL immediately before `removeItem`, so an entry rewritten fresh between its slice judgment and the final step survives. The test you asked for is in that commit too: "chunked sweep: key rewritten fresh before batch removal is not removed".

desktop/src-tauri/src/commands/window_vibrancy.rs
wpfleger96: 🤖 with the window now created opaque (dbe7ded6d) this race disappears structurally instead of needing coordination: both boot-path writes (the dark backing at webview-ready and the 250 ms post-reveal clear) target only the NSWindow backing color, which is no longer load-bearing for glass — glass is the effect view + webview canvas, and only `set_window_vibrancy` touches those. Whichever order the boot clear and the ThemeProvider's enable land in, the end state is the same; `initial_window.rs` now documents this. The persisted-glass-on cold boot is part of the manual macOS check noted on the other thread.

wesbillman (commented):
Exact-head re-review on Wes's behalf at `d264cca8112c7f0ee9974b856e430dcac4b592e5`: the prior P1 blockers are fixed.

- The main window is now created with `transparent: false`, so it retains the native opaque-window compositor path.
- Glass transparency is now scoped through `Webview::set_background_color`, while startup backing writes remain scoped to `Window::set_background_color`. The delayed startup reset therefore cannot clobber the WKWebView canvas or vibrancy layer. This resolves both the glass-off opacity and persisted glass-on sequencing findings structurally.
- Both workflow-list queries now use the 10-second focus stale gate while retaining `refetchOnWindowFocus: true`, restoring timely visibility of remote create/edit/delete operations.
- The previously fixed chunked-sweep refresh/delete race remains untouched.

The remaining P2 is test quality, not a blocker: `focusRefetchPolicy.test.mjs` still supplies synthetic `QueryObserver` options instead of proving the production hooks/options are wired to those constants. A wrong or removed production option can leave these tests green.

Focused verification at this exact clean head: workflow focus-policy tests 6/6; `cargo fmt --all -- --check`; `git diff --check`. I also ran the desktop package test command, which passed 4,708/4,708, though CI remains the broad validation source. At final inspection, Unit Tests, Rust Lint, Security, both server cross-compiles, and macOS Desktop Build were green; Desktop Core, smoke E2E, relay E2E, and Windows Rust were still running.

No new actionable blocker found. A real macOS visual smoke of glass-on cold boot and glass-off-after-glass remains worthwhile because compositor pixels are not established by static source review, but the layer architecture and API paths are now correct.

wpfleger96:
🤖 addressed the remaining P2 from the re-review in 9e77ca15b: each gated family now exports its focus-refetch policy object, the production hooks spread it into their `useQuery` options, and every `focusRefetchPolicy.test.mjs` drives its `QueryObserver` with that same production object — so a hook that drops or changes its `staleTime`/`refetchOnWindowFocus` wiring now fails the test instead of leaving it green. 14 policy objects across 12 production files, all 10 test files rewired; full unit suite green.

wesbillman (commented):
Final exact-head review on Wes's behalf at `9e77ca15b69bc61e79276f184c7a611346a1f55c`: no actionable blocker found.

The sole delta since the previously cleared head extracts the focus behavior into exported policy objects and spreads those exact objects into production queries. The tests now consume those same objects, closing the prior production-wiring P2 without changing the established runtime values. I traced all 22 touched files; the refactor preserves the surrounding query options and call-site behavior.

Focused exact-head validation: all focus-policy tests passed 28/28; `git diff --check` is clean. Current CI is otherwise green, with Desktop Core still running at inspection.

Code verdict: approved to merge once required CI completes and the stale changes-requested review is cleared. GitHub does not allow the authenticated PR-owning account to submit APPROVED on its own PR.

wesbillman (approved):

PR #2483, opened 2026-07-23
fix(desktop): let Windows users enable notifications after a false denial

BradGroux:
Addresses #2445.

## What Windows users saw

Buzz could report desktop notification permission as `denied` before Windows had registered the installed app as a notification sender. The settings toggle then treated that value as final, so users could not turn notifications on even though requesting permission would repair the state.

## What changed

- `default` still requests access on every platform.
- A false `denied` state gets one recovery request only in the Windows Tauri app.
- Genuine denial remains terminal on macOS, Linux, and ordinary Windows web pages.
- Unit coverage exercises the platform/runtime policy.
- The settings-flow Playwright regression explicitly identifies itself as Windows Tauri and verifies one recovery request.

There is no retry loop, Tauri plugin fork, new dependency, or change to notification delivery.

## Verification

- `just desktop-check desktop-test desktop-typecheck desktop-build`: passed, including 3,889 JavaScript tests.
- `pnpm --dir desktop build:e2e`: passed.
- `pnpm exec playwright test profile.spec.ts --project=integration --grep "Windows retries a false denied notification permission from settings"`: 1 passed.

## What still needs a real Windows test

The automated regression uses a Windows platform identity and the real settings hook against the notification shim. It does not run inside Windows WebView2 or a packaged Tauri application.

Keep #2445 open until an affected Windows 11 installation confirms the request returns `granted`, Buzz appears under System > Notifications, and a real toast arrives.

Issue #2445, opened 2026-07-23
Windows: desktop notifications can never be enabled — "Desktop notifications are blocked. Enable them in your system settings."


opinionate:
## Summary

On Windows, enabling **Settings → Notifications → Desktop alerts** always fails with
*"Desktop notifications are blocked. Enable them in your system settings."* — even on a
fresh install where the OS has never been asked for permission. There is nothing to
enable in Windows Settings: the app never appears in the notifications app list because
it has never posted a toast.

Root cause is an upstream bug in `tauri-plugin-notification`'s injected JS shim
(reported separately: tauri-apps/plugins-workspace#3512), but Buzz's permission flow
turns it into a hard dead end, and a one-line change on Buzz's side works around it.

## Environment

- Buzz Desktop 0.4.22 (`Buzz_0.4.22_x64-setup_alpha-unsigned.exe`), also present in `main`
- Windows 11 Home 10.0.26200, WebView2 runtime (Evergreen)
- `tauri-plugin-notification` 2.3.3 (per `desktop/src-tauri/Cargo.lock`)

## Steps to reproduce

1. Install Buzz on Windows and complete onboarding.
2. Open **Settings → Notifications** and turn on **Desktop alerts**.

**Expected:** toggle turns on; a toast is shown for the next mention/DM.
**Actual:** toggle snaps back off with the "blocked" error. No OS permission prompt is
shown, and Buzz is absent from Windows Settings → System → Notifications.

## Root cause

`tauri-plugin-notification` injects an init script that replaces `window.Notification`
with a shim. At startup the shim computes its permission state
([`guest-js/init.ts`](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/notification/guest-js/init.ts#L12-L18)):

```ts
async function isPermissionGranted(): Promise<boolean> {
  if (window.Notification.permission !== 'default' || __TEMPLATE_windows__) {
    return await Promise.resolve(window.Notification.permission === 'granted')
  }
  return await invoke('plugin:notification|is_permission_granted')
}
```

On Windows `__TEMPLATE_windows__` is `true`, so this always evaluates
`window.Notification.permission === 'granted'` against the shim's own
freshly-initialized `'default'` → `false` → the bootstrap stamps the permission
**`denied`** on every launch, without ever consulting the backend (whose
`permission_state()` unconditionally returns `Granted` on Windows).

Buzz then hits this in `desktop/src/features/notifications/lib/desktop.ts`
[...]

PR #5600, opened 2026-08-11
fix(desktop): repair the Windows notification shim's false denial on read

Joxyko:
Addresses #2445. Closely related to #2483 -- see "Relationship to #2483" below; I am happy to close this in favour of that one.

## The problem

On Windows the notification plugin's injected shim decides its permission at startup like this ([`guest-js/init.ts:12-18`](https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/notification/guest-js/init.ts)):

```ts
if (window.Notification.permission !== 'default' || __TEMPLATE_windows__) {
  return await Promise.resolve(window.Notification.permission === 'granted')
}
return await invoke('plugin:notification|is_permission_granted')
```

`__TEMPLATE_windows__` is replaced with the literal `true` on Windows, so the `||` short-circuits and the shim compares its own freshly-initialised `'default'` against `'granted'` -- `false` -- and stamps **`denied`**. The line that would ask the backend is unreachable, and the backend would have said yes: `permission_state()` returns `PermissionState::Granted` unconditionally on desktop.

So on every launch of the Windows app, `window.Notification.permission` is `denied` and nothing in the OS is denying anything.

## Why the toggle is only half of it

`getDesktopNotificationPermissionState()` trusts that value, so `refreshPermission()` reports `denied` at mount, and this effect in `hooks.ts` runs before the user touches anything:

```ts
React.useEffect(() => {
  if (settings.desktopEnabled && (permission === "denied" || permission === "unsupported")) {
    setSettings((current) => ({ ...current, desktopEnabled: false }));
  }
}, [permission, settings.desktopEnabled]);
```

It writes `desktopEnabled: false` and persists it. Measured on a real install: with alerts successfully enabled and `desktopEnabled` persisted as `true`, a clean relaunch comes back with it persisted as `false`.

That means fixing only the enable path leaves a Windows user re-enabling alerts after every restart.

## What this changes

One place: `getDesktopNotificationPermissionState()` repairs the bogus value instead of reporting it.

```ts
if (
  window.Notification.permission === "denied" &&
  isTauri() &&
  isWindowsPlatform()
) {
  try {
    return await requestDesktopNotificationAccess();
  } catch {
    return window.Notification.permission;
  }
}
```

`requestPermission()` is the only call that reaches the backend, and on Windows it raises no prompt, returns `Granted`, and rewrites the shim's cached value -- so this is self-limiting: the next read short-circuits as `granted` and never requests again. There is a test for exactly that.

Repairing at the read rather than at the toggle fixes every consumer at once, including `sendDesktopNotification()`, which gates delivery on `=== "granted"` and would otherwise drop notifications at boot.

- `denied` stays terminal on macOS, on Linux, and on ordinary Windows web pages.
- A denial that survives the request is still reported as `denied`.
- A rejected request falls back to the current value instead of propagating; callers do not handle rejections today.
- Adds `isWindowsPlatform()` to `shared/lib/platform.ts`, alongside the existing `isMacPlatform()` / `isLinuxPlatform()`.
- No new dependencies. No changes to `hooks.ts` or to notification delivery.

## Verification on a real Windows 11 install

Buzz Desktop 0.5.9 release build, Windows 11 Home 10.0.26200, WebView2 151.0.4129.72, attached over CDP. Clean boot, nothing injected:

```
window.Notification.permission                -> "denied"
localStorage[...].desktopEnabled              -> false      (written by the effect above)

await window.Notification.requestPermission() -> "granted"
window.Notification.permission                -> "granted"

new window.Notification("Buzz", { body: ... })
ToastNotificationManager::History
  .GetHistory("xyz.block.buzz.app")           -> 3 -> 4     (real toast delivered)
CreateToastNotifier(...).Setting              -> Enabled
```

The OS was permissive throughout: no `ToastEnabled` override, no notification policy keys under HKCU or HKLM, and the app present under `HKCU\...\Notifications\Settings\xyz.block.buzz.app`. Full detail in [this comment](https://github.com/block/buzz/pull/2483#issuecomment-5255972424).

To be precise about scope: this confirms the mechanism the patch relies on, measured on a stock release build. The patched build itself is covered by the unit tests below, not by a packaged Windows run.

## Tests

`pnpm test` -- full desktop suite. New cases in `lib/desktop.test.mjs`:

- Windows Tauri repairs the shim's false denied state on read
- the repaired state is cached, so later reads do not request again
- a denial that survives the request is reported as denied
- denied stays terminal outside the Windows Tauri app (Windows web, Linux Tauri, Linux web)
- granted is returned untouched and never triggers a request

Plus `shared/lib/platform.test.mjs` for `isWindowsPlatform()` across `Win32` / `Win64` / `Windows` and `MacIntel` / `Darwin` / `Linux x86_64`.

## Relationship to #2483

#2483 fixes the same root cause at the toggle: it retries the request inside `setDesktopEnabled`. That makes the toggle work, which is what #2445 reports, and its `ensureDesktopNotificationPermission` helper is a nicer shape than an inline condition.

The difference is where the repair happens. Because #2483 leaves the mount-time read as `denied`, the effect quoted above still turns the setting off on every launch, so the fix does not survive a restart. Repairing at the read covers the toggle path too, since `setDesktopEnabled` calls `refreshPermission()` first and finds `granted`.

The two overlap: both add `isWindowsPlatform()` to `platform.ts`, so whichever lands second needs a trivial rebase. If maintainers prefer #2483's shape, the equivalent of this PR is to call its helper from `refreshPermission` as well as from `setDesktopEnabled` -- I am glad to close this and send that instead. I did not want to push changes into someone else's open PR uninvited.

## Once this lands

#2445 can close. I am running a per-machine launcher workaround for a small Windows team until then, which this makes unnecessary.

### DIFF
--- a/desktop/src/features/notifications/lib/desktop.test.mjs
+++ b/desktop/src/features/notifications/lib/desktop.test.mjs
@@ -23,7 +23,55 @@ class ThrowingNotification {
 
 globalThis.window = { Notification: ThrowingNotification };
 
-const { sendDesktopNotification } = await import("./desktop.ts");
+const { sendDesktopNotification, getDesktopNotificationPermissionState } =
+  await import("./desktop.ts");
+
+// Stands in for the plugin's injected shim: `permission` is a cached value and
+// only `requestPermission()` reaches the backend and rewrites it.
+function shimNotification(initialPermission, grantedPermission = "granted") {
+  class ShimNotification {
+    static permission = initialPermission;
+    static requestCount = 0;
+
+    static async requestPermission() {
+      ShimNotification.requestCount += 1;
+      ShimNotification.permission = grantedPermission;
+      return grantedPermission;
+    }
+
+    close() {}
+  }
+
+  return ShimNotification;
+}
+
+async function withEnvironment({ platform, isTauri, notification }, callback) {
+  const originalNavigator = Object.getOwnPropertyDescriptor(
+    globalThis,
+    "navigator",
+  );
+  const originalNotification = window.Notification;
+  const originalIsTauri = globalThis.isTauri;
+
+  Object.defineProperty(globalThis, "navigator", {
+    configurable: true,
+    value: { platform, userAgent: "" },
+  });
+  window.Notification = notification;
+  globalThis.isTauri = isTauri;
+
+  try {
+    return await callback();
+  } finally {
+    if (originalNavigator) {
+      Object.defineProperty(globalThis, "navigator", originalNavigator);
+    } else {
+      delete globalThis.navigator;
+    }
+    window.Notification = originalNotification;
+    globalThis.isTauri = originalIsTauri;
+  }
+}
 
 test("constructor failure is a delivery miss and does not prevent a later notification", async (t) => {
   const warnings = [];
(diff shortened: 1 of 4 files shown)


jasonbourg-design:
Posted real-installation confirmation for this over on
https://github.com/block/buzz/issues/2445#issuecomment-5380954288 — packaged Buzz 0.5.18
on Windows 11 / WebView2 Evergreen, measured live in the running app rather than
inferred.

The two measurements most relevant to this PR:

```
window.Notification.permission                        ->  "denied"
invoke('plugin:notification|is_permission_granted')   ->  true
```

and, after a window reload on a build where alerts had been successfully enabled:

```
window.Notification.permission  ->  "denied"
desktopEnabled                  ->  false     (was true before the reload)
```

So the `hooks.ts` rewrite you describe reproduces on a packaged Windows build, which is
the part #2483 alone would not cover — the enable path can be repaired and the setting
still gets wiped at next mount. That matches your argument for fixing
`getDesktopNotificationPermissionState()` on the read path.

Separately, forcing the bootstrap to fall through to `is_permission_granted` (by
neutralising the `__TEMPLATE_windows__` short-circuit in the shipped binary) yields
`granted` at boot on a cold start, after which a real agent DM and a real channel
@mention both arrive as native Windows toasts. That is consistent with the backend
having been correct the whole time and the shim being the only broken link.

Happy to run any further check on an affected Windows 11 install if it helps unblock
review.

PR #5603, opened 2026-08-11
fix(desktop): bound initial timeline retention

wesbillman:
## Summary

- let Virtua own the initial visible timeline range instead of passing every loaded row to `keepMounted`
- populate the existing bounded retention window after the virtualizer reports its first settled viewport
- cover a 10,000-row timeline to prevent an all-history initial mount regression

## Why

`useTimelineRetention` initialized its retained-key set with every loaded timeline key. Those indices were passed to Virtua's `keepMounted`, effectively defeating virtualization during initial channel positioning until `onScrollEnd` pruned the set.

On a large real channel this grew WebContent into multiple gigabytes and blocked the renderer main thread for 20+ seconds while WebKit laid out and painted the retained rows. Starting with no retained rows restores Virtua's visible-range mount; the existing reader-neighborhood and visual-tail retention is populated once the viewport is measured.

## Validation

- `node --import ./test-loader.mjs --experimental-strip-types --test src/features/messages/ui/useTimelineRetention.test.mjs`
- pre-push hook at `8e86a189de7e9a8f2cb119396c8f912ed9dacd6e`: branch-skew, desktop-check, desktop-typecheck, and all 4,671 desktop tests passed
- manual ablation against PR #5599 on the affected profile: catastrophic channel-switch stalls disappeared

## Authorship disclosure

Carl implemented and is posting this change on Wes's behalf.

### DIFF
--- a/desktop/src/features/messages/ui/useTimelineRetention.ts
+++ b/desktop/src/features/messages/ui/useTimelineRetention.ts
@@ -2,21 +2,24 @@ import * as React from "react";
 import type { VListHandle } from "virtua";
 import { nextRetainedTimelineKeys } from "./timelineRetention";
 
+const INITIAL_RETAINED_TAIL_SIZE = 100;
+
 export function useTimelineRetention(
   keys: readonly string[],
   listRef: React.RefObject<VListHandle | null>,
   isPrepend: boolean,
 ) {
-  // Let Virtua mount only its visible range on first render. Retention is
-  // populated from the measured viewport on the first scroll-end notification;
-  // seeding every key here turns `keepMounted` into an all-history mount.
+  // Retain only a bounded visual tail on the first render. The timeline opens
+  // at newest, so this gives Virtua stable rows for initial bottom positioning
+  // without turning `keepMounted` into an all-history mount.
   const [retainedKeys, setRetainedKeys] = React.useState<ReadonlySet<string>>(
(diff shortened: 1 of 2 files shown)


desktop/src/features/messages/ui/useTimelineRetention.ts
```diff
@@ -7,8 +7,11 @@ export function useTimelineRetention(
   listRef: React.RefObject<VListHandle | null>,
   isPrepend: boolean,
 ) {
+  // Let Virtua mount only its visible range on first render. Retention is
+  // populated from the measured viewport on the first scroll-end notification;
+  // seeding every key here turns `keepMounted` into an all-history mount.
   const [retainedKeys, setRetainedKeys] = React.useState<ReadonlySet<string>>(
-    () => new Set(keys),
+    () => new Set(),
```
wesbillman: Is this it!?!?! Crazy

wpfleger96 (approved):
Reviewed at `3f7d6ebe3b8062ffd3fcf91cdfc61098eb05539d`. The bounded initial tail fixes the all-history `keepMounted` regression while preserving initial bottom positioning, and the post-paint refresh correctly reuses the existing viewport/tail retention policy and prepend guard. The 10,000-row regression covers both the initial bounded state and the measured steady state. I found no blocking or non-blocking issues.

PR #3483, opened 2026-07-29
fix(scripts): derive seed PG* from DATABASE_URL

Chessing234:
## Summary
- `scripts/seed-local-community.sh` now prefers connection fields from `DATABASE_URL` when set, so a stale `PGPORT` in `.env` cannot send seeding at the host Postgres while migrations hit Buzz's container
- Document the pairing in `.env.example`

Follow-up to the #2479 / #2511 port-override discussion ([comment](https://github.com/block/buzz/pull/2511#issuecomment-5080786393)).

## Test plan
- [ ] With `DATABASE_URL=…:55432/buzz` and `PGPORT=5432` in `.env`, run `./scripts/seed-local-community.sh` and confirm it connects on `55432`
- [ ] With no `DATABASE_URL`, confirm defaults (`localhost:5432` / buzz) still work
- [ ] `bash -n scripts/seed-local-community.sh`

Made with [Cursor](https://cursor.com)

PR #5615, opened 2026-08-12
fix(dev): make local service host ports configurable

### LINKED ISSUE #2479: dev setup: allow Postgres and Redis host ports to be configured

smkthat:
## Summary

Local setup currently publishes fixed Docker host ports while application URLs
come from `.env`. Changing `DATABASE_URL` or `REDIS_URL` therefore leaves
Compose on the old port and can make migrations time out against the wrong
endpoint.

This change:

- makes every development service host port configurable through `.env`, while
  preserving the current defaults and `127.0.0.1` bindings;
- keeps container-side ports fixed and leaves service credentials unchanged;
- loads and validates service settings during setup, with a fail-fast check
  when a loopback URL disagrees with its mapped port;
- makes the relay-test launcher resolve the ports actually published by local
  Compose while pinning database, Redis, and MinIO endpoints and credentials to
  the development stack, so values from `.env` cannot redirect schema changes
  to an external service;
- always reconciles Compose before health checks so edits to `.env` take effect
  even when the existing Postgres and Redis containers are healthy;
- documents the variables and adds a regression test for default/custom
  mappings, credential isolation, URL validation, Compose reconciliation, and
  external-endpoint protection in the relay-test launcher.

### Related issue

Fixes #2479.

Related work:

- #2486 and #3854 address subsets of the Postgres/Redis port-mapping problem.
- #2511 includes local port overrides as part of a broader change.
- #3483 aligns seed database settings with `DATABASE_URL`.
- #4871 introduced loopback-only development bindings; this PR preserves that
  security boundary for every configurable mapping.

Compared with the open alternatives, this PR keeps the scope limited to local
service configuration while covering all published Compose ports and the
setup/test call sites that consume them.

### Testing

- `./scripts/test-dev-service-ports.sh`
- `bash -n scripts/dev-setup.sh scripts/start-relay-for-tests.sh scripts/lib/dev-service-env.sh scripts/test-dev-service-ports.sh`
- `shellcheck scripts/dev-setup.sh scripts/start-relay-for-tests.sh scripts/lib/dev-service-env.sh scripts/test-dev-service-ports.sh`
- `docker compose --env-file .env.example config --quiet`
- `docker compose config --quiet` with custom local Postgres, Redis, and MinIO ports
- `just setup` with those custom ports (Compose reconciliation, migrations,
  seed, dependency setup, and hooks all succeeded)
- `just test-unit`
- `just ci` (formatting, linting, builds, and all Rust, desktop, web, and mobile
  test suites, including 4,715 desktop, 2,402 Tauri, and 1,261 mobile tests)
- pre-push hook, including branch-skew validation

No UI changes; screenshots are not applicable.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,17 +7,24 @@
 # All defaults here work with `docker compose up` out of the box.
 #
 # Service ports (defaults):
-#   Postgres    → localhost:5432
-#   Redis       → localhost:6379
-#   Typesense   → localhost:8108
-#   Adminer     → localhost:8082  (DB browser UI)
+#   Postgres      → localhost:5432
+#   Redis         → localhost:6379
+#   Typesense     → localhost:8108
+#   Adminer       → localhost:8082  (DB browser UI)
+#   Keycloak      → localhost:8180  (OAuth/OIDC testing)
+#   MinIO API     → localhost:9000
+#   MinIO console → localhost:9001
+#   Prometheus    → localhost:9090
 #
-# Note: If port 8082 conflicts, change the adminer port in docker-compose.yml
+# Change host ports here; Docker Compose keeps the service ports inside the
+# dev network unchanged.
 # =============================================================================
 
 # -----------------------------------------------------------------------------
 # Database (Postgres 17)
 # -----------------------------------------------------------------------------
+# Keep DATABASE_URL aligned with PGHOST and PGPORT.
+# Compose uses PGPORT as the published host port.
 DATABASE_URL=postgres://buzz:buzz_dev@localhost:5432/buzz
 # Optional read-replica URL; unset/blank keeps all reads on the writer.
 # READ_DATABASE_URL=postgres://buzz:buzz_dev@localhost:5433/buzz
@@ -31,6 +38,8 @@ PGDATABASE=buzz
 # Redis 7
 # -----------------------------------------------------------------------------
 REDIS_URL=redis://localhost:6379
+# Host port published by Docker Compose. Keep it aligned with REDIS_URL.
+REDIS_PORT=6379
 # Max connections in the relay's shared Redis pool (default 16).
 # BUZZ_REDIS_POOL_SIZE=16
 
@@ -43,6 +52,15 @@ REDIS_URL=redis://localhost:6379
 # -----------------------------------------------------------------------------
 TYPESENSE_API_KEY=buzz_dev_key
 TYPESENSE_URL=http://localhost:8108
+
+# -----------------------------------------------------------------------------
+# Auxiliary local service ports
+# -----------------------------------------------------------------------------
+ADMINER_PORT=8082
+KEYCLOAK_PORT=8180
+MINIO_API_PORT=9000
+MINIO_CONSOLE_PORT=9001
+PROMETHEUS_PORT=9090
 
 # -----------------------------------------------------------------------------
 # Relay (WebSocket server)
@@ -85,7 +103,8 @@ RELAY_URL=ws://localhost:3000
 # -----------------------------------------------------------------------------
 # S3-Compatible Object Storage (media + Git/CAS)
 # -----------------------------------------------------------------------------
-# The local MinIO container is reachable from host processes at localhost:9000.
+# The local MinIO container is reachable from host processes at the port in
+# MINIO_API_PORT. Keep BUZZ_S3_ENDPOINT aligned if that port changes.
 # Path style keeps the bucket in the URL path and is required by this local DNS
 # setup. Use `virtual` only when the provider requires bucket-as-subdomain URLs.
 BUZZ_S3_ENDPOINT=http://localhost:9000
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -154,8 +154,12 @@ it is safe to re-run.
 
 `just setup` then starts Docker services (Postgres on `:5432`, Redis on `:6379`,
 Adminer on `:8082`, Keycloak on `:8180` for local OAuth/OIDC testing, MinIO on
-`:9000` for media storage, and Prometheus on `:9090` for metrics) and runs all
-pending database migrations.
+`:9000`/`:9001` for media storage and its console, and Prometheus on `:9090` for
+metrics) and runs all pending database migrations. These are host-port defaults;
+override `PGPORT`, `REDIS_PORT`, `ADMINER_PORT`, `KEYCLOAK_PORT`,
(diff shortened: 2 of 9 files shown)


ravarora2:
🤖 Tested locally — LGTM. ✅

Checked out the branch in an isolated worktree and ran the non-destructive checks (no containers started, DB/Desktop untouched):

- `bash -n` on all four changed scripts — clean.
- `./scripts/test-dev-service-ports.sh` — **passes** (`Dev service port checks passed`, exit 0). This covers default/custom port rendering, credential isolation, local-URL/port validation (without leaking passwords), the remote-endpoint protection in the relay-test launcher, and Compose reconciliation.
- Independent `docker compose config` render:
  - Defaults unchanged, every service still bound to `127.0.0.1`.
  - `PGPORT=15432` / `MINIO_CONSOLE_PORT=9002` move only the host port; container ports stay fixed.
  - Injecting `POSTGRES_PASSWORD=HACKED` / `BUZZ_S3_SECRET_KEY=HACKED` via env did **not** change the rendered credentials — they stay `buzz_dev`/`buzz_dev_secret`. Credential isolation holds.

Nice touch pinning the test-relay launcher's host + credentials to the local dev stack and only borrowing the port from Compose, so a developer's remote `DATABASE_URL`/`PG*` can't become an auto-approved schema target — and there's a regression test proving it.

A few non-blocking notes:
1. `validate_env` now shells out to `node` at `just setup` time — worth confirming node is guaranteed on PATH there (it is via hermit in my run).
2. Changing `PGPORT`/etc. will break a concurrently-running Desktop, since the dev Compose uses fixed container names and Desktop expects the standard ports. `TESTING.md` mentions the shared stack; a one-liner that a port override affects Desktop too could save someone a confusing debugging session.
3. Minor: a portless `BUZZ_S3_ENDPOINT` (e.g. `http://localhost`) validates against default port 80 and would fail against `MINIO_API_PORT=9000`. Fine for the common case, just noting the edge.

(I only ran the static/rendered tier; I didn't start containers on a custom port since that shares the local Docker stack.)

smkthat:
@ravarora2  Thanks for the thorough review and the LGTM. I addressed two small follow-ups from your non-blocking notes:

- `scripts/dev-setup.sh` now prepends `<repo>/bin` to `PATH` before `validate_env`. A new regression case runs `dev-setup.sh` without a system Node binary and verifies that it uses the Hermit shim before reaching the Compose step.
- `TESTING.md` now explicitly notes that changing a host-port override reconciles the shared Compose stack, so a running Desktop or relay using the previous ports will lose its service connections.

I intentionally kept a portless `BUZZ_S3_ENDPOINT=http://localhost` fail-closed: it means `:80` and should not implicitly assume the configured MinIO port.

Verified with `./scripts/test-dev-service-ports.sh`, shell syntax checks, and `just ci`.

ravarora2:
🤖 Follow-up change summary after the review blocker was fixed:

**What changed**

- Rebased this branch onto current `main`.
- Reworked the local-service URL validation so it recognizes equivalent loopback destinations instead of comparing only three literal host strings. This now covers trailing-dot `localhost`, the full IPv4 loopback range (including shorthand and integer forms), IPv6 loopback, and IPv4-mapped IPv6 loopback.
- Unspecified bind addresses such as `0.0.0.0` and `[::]` now fail closed instead of being treated as safe remote endpoints.
- The configured URL port is compared with the Compose host port for every local/loopback form, preventing setup from starting Buzz PostgreSQL on one port and migrating a different local database on another port.
- Added the expanded hostname/port matrix plus a setup-level regression test proving that a mismatch stops before either `docker compose up` or `buzz-admin migrate` is invoked.

**Verification**

- Final pushed head: `ac7abeb5bbc55ab92d52cca0c71ab9fc6b0163ee`.
- The focused port regression test passed at that head.
- Full `just ci` passed during the fix validation.
- A final full `cargo test -p buzz-acp` rerun passed: 778 library tests and 9 integration tests. The first run had one known timing-sensitive keepalive failure; the clean full rerun passed without code changes.
- `git diff --check`, formatting, clippy, DCO trailers, and clean-worktree checks passed.

The configurable-port design remains unchanged; this commit hardens only the local-address safety boundary and its regression coverage.

ravarora2 (approved):
lgtm.Let's wait for owners to have a final review

PR #5624, opened 2026-08-12
Projects v3: unify sharing, discussions, and issue ownership

thomaspblock:
## Summary

Projects v3 makes repository work shareable, discussion-aware, and easier to scan in one coherent workspace. People can copy canonical links, reopen the exact workspace tab, understand issue and pull-request context at a glance, find related channel conversations, and assign or unassign issues across Desktop and CLI.

- **Unified workspace** — top-level sections sit above repository controls in one rounded workspace, with navigation positioned close to the page heading. README and Files retain branch selection; every section has a labeled icon header, and Issues and Pull Requests expose creation from a consistent right-aligned action.
- **Repository management** — the repository selector is always available, including single-repository projects. Its integrated add flow lets project owners create a repository manually or select an existing repository without a separate toolbar button.
- **Readable work-item lists** — issue and pull-request rows use plain-language context instead of opaque metadata. Files, commits, issues, pull requests, channels, and contributors share consistent row density and right-aligned timestamps, while deterministic fallback-avatar colors keep participants distinct on light backgrounds. Inbox pull-request metadata wraps between complete phrases and truncates long channel names instead of compressing copy into narrow columns.
- **Reliable entity links** — projects, repositories, issues, pull requests, and commits have canonical `buzz://` links, preview cards, OS deep-link routing, and tab-aware navigation. Reopening the same link re-applies its destination instead of leaving the user on a locally selected tab.
- **Related conversations** — repository and work-item views surface channels discussing the current entity, including participants, channel navigation, message context, and an explicit notice when discovery reaches its 500-result cap.
- **Reversible issue ownership** — trusted assignment and unassignment events work across Desktop, Tauri, `buzz-sdk`, and `buzz issues`. Assignees appear in project views and the assigned inbox, while authorized users can remove assignments directly from the assignee row.

Assignment state is derived chronologically from labeled Nostr notes. Issue authors and repository owners may change any assignee; other users may only assign or unassign themselves. Shared golden fixtures keep entity-link grammar and validation aligned across TypeScript and Rust.

The branch also updates `webbrowser` to the patched release for RUSTSEC-2026-0257.

### Related issue

N/A.

### Testing

- [x] `just ci` — formatting, lint, typechecking, unit tests, and builds passed
- [x] Full pre-push suite — organization, branch-skew, Desktop checks, typechecking, and tests passed on the latest push
- [x] `cargo test -p buzz-cli` and focused `buzz-sdk` assignment tests passed
- [x] Focused Tauri recipient-note and 500-result search-limit tests passed
- [x] Desktop entity-link and issue-assignment unit tests passed
- [x] Playwright smoke coverage passed for assignment, repeated entity-link navigation, repository create/select flows, section headers and actions, timestamp alignment, timeline icons, sentence-style issue/PR metadata, header spacing, avatar contrast, and Inbox metadata at stacked and side-rail breakpoints
- [ ] Manual staging pass: link round-trips, Channels tab, assignment flows, and inbox routing

### Screenshots

Pull requests explain who opened the request, where it lives, and which branch it comes from; fallback avatars remain visually distinct.

Pull request list with conversational metadata

Issues use the same sentence-style hierarchy while keeping status and recency easy to scan.

Issue list with conversational metadata

The wide Inbox detail keeps author, timestamp, and origin context readable beside its metadata rail.

Pull request Inbox detail with readable metadata

[View the complete six-state Projects v3 screenshot set](https://github.com/block/buzz/pull/5624#issuecomment-5268039672) and [the compact/wide Inbox comparison](https://github.com/block/buzz/pull/5624#issuecomment-5268614585).


khorami-ahoo (approved):

khorami-ahoo (approved):

khorami-ahoo (approved):

khorami-ahoo (approved):

khorami-ahoo (approved):

thomaspblock:
🤖 Updated Projects v3 screenshots with the latest workspace layout and distinct avatar fallback colors.

### Single-box workspace
Tabs sit above repository and branch controls; contributor avatars now remain distinct on light backgrounds.

01-workspace-overview

### Issues workspace
The Issues section has consistent header, row spacing, and right-aligned dates.

02-issues-tab

### Issue detail
Issue ownership and comments use colored author avatars in the focused detail view.

03-issue-detail

### Pull request detail
Pull request context and copy-link actions use the same detail hierarchy.

04-pr-detail

### Pull request list
Each second line explains who opened the request, in which repository, and from which branch.

05-pr-list-metadata

### Issue list
Each second line reads as a sentence, with distinct author avatars for faster scanning.

06-issue-list-metadata

thomaspblock:
🤖 Final Projects v3 Inbox validation.

### Compact Inbox detail
The pull request remains actionable when the metadata rail stacks below the conversation.

01-pull-request-detail

### Wide Inbox detail
Author, time, and origin metadata wrap as complete phrases; long channel names truncate instead of collapsing into word-wide columns.

02-pull-request-detail-wide

desktop/src/features/projects/projectIssues.mjs
```diff
@@ -73,21 +83,54 @@ function statusFromEvent(issue, statusEvent) {
...
+ *
+ * Trusted signers are the issue author and repo owner (who may change anyone),
+ * plus any community member whose operation names only themselves. Same-second
+ * events use their id as a deterministic tie-breaker because relay result order
+ * is not stable.
+ */
+function assigneesForIssue(issue, issueCommentEvents) {
+  const allowedActors = allowedActorsForRoot(issue);
+  const assignees = new Set();
+  const operations = sortEvents(
```
klopez4212: **[P1] Prevent future-dated self-operations from overriding maintainer actions.** This reducer gives every operation precedence solely through the signer-controlled `created_at`, while line 114 accepts a non-owner operation when its sole `p` tag is the signer. The relay currently accepts timestamps up to 900 seconds in the future (`crates/buzz-relay/src/handlers/ingest.rs:1976-1983`). A volunteer can therefore publish a self-assignment at `now + 899`; if the owner then unassigns them at the real current time, sorting applies the owner's removal first and the volunteer's already-published assignment last, so the maintainer action has no effect. The inverse future self-unassignment defeats a subsequent owner assignment as well. The local monotonic timestamp helper is not a trust boundary because arbitrary signed kind-1 events enter this reducer. Please define conflict semantics in which a user cannot precommit a self-service operation that outranks a later owner/author decision (for example, causal/authoritative ordering not based solely on actor-controlled event time), and cover future-dated self-assign → owner unassign plus the inverse.

desktop/src/shared/ui/markdown/entityLinks.tsx
```diff
@@ -21,7 +21,17 @@ export function useOpenEntityLink(): (link: ParsedEntityLink) => void {
...
   return React.useCallback(
     (link: ParsedEntityLink) => {
+      const tab =
+        (link.type === "repo" || link.type === "project") && link.tab
+          ? link.tab
+          : undefined;
       void goProject(entityLinkProjectRouteId(link), {
+        ...(tab
+          ? {
+              entityNavigationId: crypto.randomUUID(),
```
klopez4212: **[P1] Re-trigger repeated PR and issue links too.** The request ID is only generated when a repo/project link has a tab. PR and issue links therefore navigate without `force`, and `ProjectDetailScreen` resynchronizes its local detail state only when `pullRequestId`/`issueId` changes. After the breadcrumb locally clears an open PR or issue (without clearing the route search param), opening the identical link again keeps the same prop and the effect does not run, so the detail stays closed. Generate a request key for every entity-link activation and include it when synchronizing PR/issue selection, with regression coverage for reopening the same PR and issue links—not just the repeated repo-tab case.

desktop/src/features/projects/ui/ProjectsView.tsx
```diff
@@ -467,13 +467,17 @@ export function ProjectsView() {
...
-        ? issues.filter(
-            ({ issue }) =>
-              normalizePubkey(issue.author) === normalizePubkey(currentPubkey),
-          )
-        : issues;
+      issueScope === "mine" && viewer
+        ? issues.filter(({ issue }) => normalizePubkey(issue.author) === viewer)
+        : issueScope === "assigned" && viewer
+          ? issues.filter(({ issue }) =>
+              issue.assignees.some(
```
klopez4212, line 477: **[P2] Prefetch profiles for the assignees consumed by this scope and rendered in the issue list.** `ProjectsIssuesList` now passes `issue.assignees` to `IssueAssigneeFacepile`, but the `projectPubkeys` batch assembled above includes only each issue's author, recipients, and commenters. An assignee who appears in none of those sets never has their profile loaded, so the facepile falls back to a truncated pubkey and generated avatar even when that user has profile metadata. Please add `...issue.assignees` to the issue keys in `projectPubkeys`, matching the project-detail profile collection.

klopez4212 (changes requested):
Reviewing on behalf of Kenny Lopez. I found three correctness issues that should be addressed before merge; the timestamp-ordering issue is the release blocker.

desktop/src/features/projects/ui/ProjectsView.tsx
thomaspblock: 🤖 > Please add `...issue.assignees` to the issue keys in `projectPubkeys`.

Addressed in 2aed85392: issue assignees rendered by `IssueAssigneeFacepile` are now included in the profile batch, matching project-detail behavior.

desktop/src/features/projects/projectIssues.mjs
```diff
@@ -73,21 +83,69 @@ function statusFromEvent(issue, statusEvent) {
...
+    const operation = { isAssignment, pubkeys };
+    if (allowedActors.has(signer)) {
+      authoritativeOperations.push(operation);
+    } else {
+      selfServiceOperations.push(operation);
+    }
+  }
+  for (const { isAssignment, pubkeys } of [
+    ...selfServiceOperations,
+    ...authoritativeOperations,
```
klopez4212, line 151: **[P1] Preserve legitimate self-service actions that occur after a maintainer action.** Partitioning the entire event history by authority means *every* owner/author operation is applied after *every* volunteer operation, regardless of their actual order—not just the adversarial future-dated case. For example, an owner assigns a volunteer at `created_at=200`, then the volunteer later clicks the currently offered **Unassign me** control at `created_at=1000`. This reducer applies the volunteer unassignment first and the old owner assignment last, so the UI reports success but the volunteer remains assigned after refetch. Likewise, once an owner has unassigned someone, no later self-assignment can ever take effect while that owner event remains in history. The new inverse regression test encodes this broken outcome because its inputs are indistinguishable from that normal later opt-out once both events are read. Please use causal/order semantics that stop a precommitted future self-operation from beating a later authoritative action without making all authoritative history permanently dominate all self-service history (or remove/disable self-service transitions that the state model cannot honor). Add coverage for owner assign → later volunteer unassign and owner unassign → later volunteer self-assign.

klopez4212 (changes requested):
Re-reviewing on behalf of Kenny Lopez. The repeated-link and profile-hydration fixes look correct. The assignment fix blocks the timestamp exploit, but introduces permanent authority-class precedence that breaks ordinary later self-service actions.

desktop/src/features/projects/projectIssues.mjs
thomaspblock: 🤖 > Please use causal/order semantics that stop a precommitted future self-operation from beating a later authoritative action without making all authoritative history permanently dominate all self-service history.

Addressed in 86bae900a with per-assignee causal operation heads. New self-service writes reference the assignment operation they observed; that reference must match the current head before the action applies. This preserves later “Assign me”/“Unassign me” actions while stale, malformed, and future-precommitted operations fail closed. Desktop and CLI reducers now share these semantics, with coverage for both normal owner→self transitions, both future-timestamp attacks, and stale references.

desktop/src-tauri/src/deep_link.rs
```diff
@@ -366,6 +451,19 @@ pub(crate) fn handle_deep_link_url(app: &tauri::AppHandle, url_str: &str) {
...
+            // `buzz://pr|issue?id=<eventId>&owner=<pubkey>&d=<dtag>` — the
+            // share links copied from the Projects UI. The frontend owns
+            // routing (`useEntityDeepLinks`), so the validated URL is
+            // forwarded unchanged.
+            if parse_entity_deep_link(&url).is_none() {
+                eprintln!("buzz-desktop: malformed entity deep link: {url_str}");
+                return;
+            }
+            activate_main_window(app);
+            let _ = app.emit("deep-link-entity", url_str.to_owned());
```
wpfleger96: 🤖 This emit is fire-and-forget, which can lose a valid entity link on cold start (both independent review passes converged on this):

- `lib.rs` registers `on_open_url` during app setup, but `tauri-plugin-deep-link` documents `on_open_url` for links received *while running* and says to use `get_current` for launch URLs — on Windows/Linux the plugin parses the launch CLI URL during its initialization and can emit before this callback is registered.
- Even when the Rust side does receive the launch URL, the only frontend consumer is the React effect in `useEntityDeepLinks.ts`, which mounts after this emit fires. Tauri events don't replay to late subscribers, and unlike community links there's no pending queue here (`take_pending_community_deep_link` / `acknowledge_pending_community_deep_link`).

Net effect: opening a `buzz://repo|project|pr|issue` share link while Buzz is stopped can launch/focus the app without navigating to the entity. The e2e in `entity-link-recipient-cards.spec.ts` waits for the loaded app before emitting, so it only proves the warm path.

I think the fix is to persist validated entity intents until the frontend acknowledges them — mirror the community-link pending/drain pattern (and/or consume `get_current` into that queue on startup), subscribe first then drain — plus a cold-start regression test that places the URL before frontend mount.

wpfleger96 (changes requested):
🤖 Ran two independent review passes on this at head `06a2731c68` and synthesized them. Both converged on one blocking correctness finding — cold-start entity deep-link delivery — see the inline comment on `desktop/src-tauri/src/deep_link.rs`.

Everything else traced clean across both passes: entity-link parser parity between Desktop and CLI (with the shared golden fixture in `test-fixtures/entity-links.json`), the uncaused-self → authoritative → causal-self assignment reduction and its prior/head model, the new SDK assignment builders, share-link generation/guards, discussion-channel discovery, and the deep-link plumbing outside the cold-start path. `just desktop-test` (4747/4747) and `just desktop-tauri-test` pass at this exact head; those suites don't exercise pre-mount OS launch delivery, which is what the finding is about.

thomaspblock (commented):
🤖 Retrying the required DCO status after it did not report on the latest push.

@dcoapp recheck

thomaspblock (commented):
🤖 Retrying the stalled required DCO check again.

@dcoapp recheck

thomaspblock:
Superseded by #5792 — history flattened into a single signed-off commit on latest main (tree verified identical via git merge-tree) to escape the permanently-queued required check suites on this head.

PR #5625, opened 2026-08-12
fix(buzz-relay): serve NIP-11 info document with permissive CORS

### LINKED ISSUE #5550: NIP-11 relay information document is not CORS-accessible when BUZZ_CORS_ORIGINS is set
### Summary

NIP-11 requires that the relay information document be readable cross-origin:

> Relays MUST accept CORS requests by sending `Access-Control-Allow-Origin`, `Access-Control-Allow-Headers`, and `Access-Control-Allow-Methods` headers.

When `BUZZ_CORS_ORIGINS` is set, the relay stops sending `Access-Control-Allow-Origin` for any origin outside the allowlist — including on `GET /` with `Accept: application/nostr+json` and on `GET /info`. Browser-based Nostr clients hosted anywhere else can no longer read the document.

This matters because **the official deployment guide tells operators to set that variable**: [`deploy/compose/.env.example`](https://github.com/block/buzz/blob/main/deploy/compose/.env.example#L13) ships `BUZZ_CORS_ORIGINS=https://buzz.example.com`. Following the documented VPS path therefore produces a relay that is non-conformant on this MUST.

### Reproduction

Deploy with `deploy/compose` and `BUZZ_CORS_ORIGINS=https://relay.example.com` (as `.env.example` prescribes), then:

```console
$ curl -sSI -H 'Origin: https://relay.example.com' -H 'Accept: application/nostr+json' https://relay.example.com/ | grep -i access-control
access-control-allow-origin: https://relay.example.com

$ curl -sSI -H 'Origin: https://some-web-client.example' -H 'Accept: application/nostr+json' https://relay.example.com/ | grep -i access-control
                       # (nothing — request succeeds with 200, but the browser blocks the read)
```

The response is `200` in both cases; only the CORS header differs, so this fails silently from the server's perspective.

Observed on `ghcr.io/block/buzz:0.2.1`.

### Cause

`build_cors_layer` is applied to the whole merged router, which includes the NIP-11 document route:

- [`router.rs:192`](https://github.com/block/buzz/blob/main/crates/buzz-relay/src/router.rs#L192) — `.layer(build_cors_layer(&state.config.cors_origins))` wraps every route
- [`router.rs:423-427`](https://github.com/block/buzz/blob/main/crates/buzz-relay/src/router.rs#L423) — `CorsLayer::permissive()` is returned **only** when the allowlist is empty; otherwise `AllowOrigin::list(origins)` applies to the relay-info endpoint too

So the allowlist that (correctly) protects the authenticated REST surface also narrows a document that is public by design.

### Suggested fix

Exempt the relay information document from the origin allowlist and always serve it with permissive CORS, while leaving `BUZZ_CORS_ORIGINS` governing the rest of the surface. Concretely: split the NIP-11 document route (`GET /` with `Accept: application/nostr+json`, and `GET /info`) into its own `Router` layered with `CorsLayer::permissive()` before merging, so operators get spec conformance without widening CORS on `/events`, `/query`, `/count` or the media endpoints.

The document contains no private data — capabilities, limits, and the relay's public key — so serving it to any origin does not weaken a closed relay. `auth_required`/`restricted_writes` continue to gate everything that matters.

Happy to open a PR if the approach sounds right.

### Not part of this report

Two things I checked and concluded are **not** bugs, in case they come up:

- The response `Content-Type` is `application/json` rather than `application/nostr+json`. NIP-11 constrains the request `Accept` header, not the response type, so this looks conformant.
- No `charset=utf-8` on that content type. Per RFC 8259, `application/json` defines no `charset` parameter and UTF-8 is implied; the bytes on the wire are valid UTF-8.

Sirius6907:
## Summary

When `BUZZ_CORS_ORIGINS` is configured (non-empty allowlist), the relay's NIP-11 relay information
document — `GET /` with `Accept: application/nostr+json` and `GET /info` — is served **without**
`Access-Control-Allow-Origin`, because `build_router` wrapped the *entire* merged router in the
allowlist `CorsLayer`. NIP-11 (https://github.com/nostr-protocol/nips/blob/master/11.md) requires
relays to answer CORS requests for the info document ("Relays MUST accept CORS requests by sending
Access-Control-Allow-Origin, Access-Control-Allow-Headers, and Access-Control-Allow-Methods headers"),
so any browser-based Nostr client that fetches relay metadata from an allowlisted relay gets a blocked
cross-origin read.

### Change

- Split the two NIP-11 routes (`/` via `nip11_or_ws_handler`, `/info` via `relay_info_handler`) into
  their own router layered with `CorsLayer::permissive()`.
- Apply the `BUZZ_CORS_ORIGINS` allowlist (`build_cors_layer`) to the remaining REST surface only
  (api, media, git, git-policy, admin). Metrics + trace layers stay global.
- The NIP-11 document is public by design (capabilities, limits, public key); the allowlist keeps
  governing the authenticated surface exactly as before.

### Related issue

Fixes #5550. (No duplicate PR found in search of open PRs/issues; issue is unassigned.)

### Testing

- New regression tests (in `crates/buzz-relay/src/router.rs`, `mod tests`, run via the existing
  `just admin-check` recipe's `cargo test -p buzz-relay router::tests`):
  - `nip11_document_keeps_permissive_cors_when_allowlist_is_configured` — GET / (NIP-11 Accept) and
    OPTIONS /info preflight from a **non-allowlisted** origin both return
    `Access-Control-Allow-Origin`.
  - `cors_allowlist_still_governs_the_authenticated_rest_surface` — POST /events from a
    non-allowlisted origin gets **no** CORS header; from the allowlisted origin it gets the header.
- Full `router::tests` module: **7 passed; 0 failed; finished in 0.56s** (no infra needed; DB/Redis
  pools are lazy and fail fast — `acquire_timeout` capped for the test).
- `cargo clippy -p buzz-relay --tests`: clean (exit 0, no warnings).
- `cargo fmt -p buzz-relay -- --check`: clean.
- `cargo check -p buzz-relay --tests`: clean.

### DIFF
--- a/crates/buzz-relay/src/router.rs
+++ b/crates/buzz-relay/src/router.rs
@@ -59,10 +59,21 @@ pub fn build_router(state: Arc<AppState>) -> Router {
     let admin_router = admin_enabled
         .then(|| Router::new().nest("/api/admin/v1", api::admin::router(state.clone())));
 
-    let api_router = Router::new()
-        // WebSocket + NIP-11
+    // NIP-11 relay information document — public by design. NIP-11 requires
+    // relays to answer CORS requests ("Relays MUST accept CORS requests by
+    // sending Access-Control-Allow-Origin, Access-Control-Allow-Headers, and
+    // Access-Control-Allow-Methods headers"), so the document (`GET /` with
+    // `Accept: application/nostr+json`, and `GET /info`) is always served with
+    // permissive CORS, independent of `BUZZ_CORS_ORIGINS`. The allowlist
+    // applied further down governs the authenticated REST surface only; the
+    // document contains no private data (capabilities, limits, public key).
+    let nip11_router = Router::new()
         .route("/", get(nip11_or_ws_handler))
         .route("/info", get(relay_info_handler))
+        .layer(CorsLayer::permissive())
+        .with_state(state.clone());
+
+    let api_router = Router::new()
         .route("/.well-known/nostr.json", get(api::nip05::nostr_nip05))
         // Health endpoints
         .route("/health", get(health_handler))
@@ -131,15 +142,21 @@ pub fn build_router(state: Arc<AppState>) -> Router {
         .layer(RequestBodyLimitLayer::new(1024 * 1024))
         .with_state(state.clone());
 
-    // Merge — each sub-router carries its own body limit.
-    // Metrics → Trace → CORS applied once over the combined router.
-    let mut merged = api_router
+    // Merge — each sub-router carries its own body limit. The CORS allowlist
+    // (BUZZ_CORS_ORIGINS) is applied to the non-NIP-11 surface only; the
+    // public NIP-11 document keeps permissive CORS (see `nip11_router`
+    // above), so an allowlist configured for the REST surface cannot make the
+    // relay non-conformant with the NIP-11 CORS requirement.
+    let mut rest = api_router
         .merge(media_router)
         .merge(git_router)
         .merge(git_policy_router);
     if let Some(admin_router) = admin_router {
-        merged = merged.merge(admin_router);
+        rest = rest.merge(admin_router);
     }
+    let rest = rest.layer(build_cors_layer(&state.config.cors_origins));
+
+    let mut merged = nip11_router.merge(rest);
 
     // Serve both bundles from one fallback. The admin host is checked first so
     // it can never fall through to the public web bundle.
@@ -189,7 +206,6 @@ pub fn build_router(state: Arc<AppState>) -> Router {
     merged
         .layer(middleware::from_fn(track_metrics))
         .layer(http_trace_layer())
-        .layer(build_cors_layer(&state.config.cors_origins))
 }
 
 fn http_trace_layer() -> TraceLayer<HttpMakeClassifier, fn(&Request<Body>) -> tracing::Span> {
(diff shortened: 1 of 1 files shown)


ravarora2:
🤖 Tested — LGTM.

Verified two ways:

**Unit tests** — `cargo test -p buzz-relay router::tests`: 7/7 pass, including the two new CORS regression tests.

**Live relay + curl** — booted the relay with `BUZZ_CORS_ORIGINS="https://allowed.example"` and issued real cross-origin requests:

| Request | Result |
|---|---|
| `GET /` (NIP-11 Accept) from an **unlisted** origin | `access-control-allow-origin: *` ✅ |
| `OPTIONS /info` preflight from an **unlisted** origin | ACAO + allow-methods + allow-headers all present ✅ |
| `OPTIONS /events` from an **unlisted** origin | no `Access-Control-Allow-Origin` (blocked) ✅ |
| `OPTIONS /events` from the **allowlisted** origin | `access-control-allow-origin: https://allowed.example` echoed ✅ |

That confirms the split end-to-end: the NIP-11 document is CORS-readable regardless of the allowlist, while the authenticated `/events` surface stays gated. The change is sound (per-sub-router CORS applied before the merge, no route conflict, WebSocket upgrade on `/` unaffected).

**One question:** the sibling NIP-05 endpoint `/.well-known/nostr.json` is also a public, browser-read discovery endpoint, and its handler doesn't set its own CORS header — so it's left under the `BUZZ_CORS_ORIGINS` allowlist by this PR. NIP-05 likewise expects permissive CORS (`Access-Control-Allow-Origin: *`), so the same bug plausibly affects it on allowlisted relays. Is leaving NIP-05 out of scope intentional (follow-up PR), or should `/.well-known/nostr.json` get the same permissive treatment here?

ravarora2 (commented):
lgtm, will leave it for owners to merge

wesbillman (changes requested):
The production CORS split looks sound, but the new test helper can hang the entire suite forever. `cors_test_state` retries every `Config::from_env()` error in an unbounded `loop`, despite its comment promising only “a few” retries. A persistent invalid runner environment therefore never fails with a diagnostic; it wedges the process indefinitely. Please bound the attempts or deadline and panic/return with the final `ConfigError` (or construct this test state without reading mutable process environment).

PR #5635, opened 2026-08-12
fix(desktop): deliver and retain Inbox DM messages

### LINKED ISSUE #5632: Inbox replies to managed agents can disappear before delivery
## Summary
A human reports that replying to a managed agent through **Inbox** can cause the new DM message to appear briefly in the conversation and then disappear before the agent handles it.

## Expected
An Inbox reply should persist in the DM timeline and be delivered once to the managed-agent harness (or show an explicit delivery/error state).

## Actual
The reply is shown in the agent DM, but sometimes the agent does not process it and the message later disappears from the timeline.

## Steps reported
1. Open an Inbox notification for a managed agent's message.
2. Reply from Inbox.
3. Observe the reply appear in the agent DM.
4. In the failing case, the agent never responds and the reply disappears.

## Evidence and scope
- Reported from the macOS Buzz app on 2026-08-11 around 23:30 Pacific.
- This was **not reproduced deterministically**.
- A control Inbox reply did arrive at the harness with the expected channel (`h`) and agent mention (`p`) tags, so the failure may be specific to an Inbox-reply reconciliation path or an intermittent delivery race rather than all DM delivery.
- No private message content or account identifiers are included here.

## Suggested investigation
Trace one Inbox reply event end-to-end: local optimistic insertion -> signed event submission -> relay acknowledgement/storage -> DM subscription/cache reconciliation -> harness receipt/enqueue. Preserve the event id at each stage, and ensure a failed delivery is surfaced rather than removing the message.

aemrebarut:
## Summary

- route Inbox sends through the same mutation as the channel composer, so direct-message participants receive the implicit `p` tags required for delivery
- update the shared channel-window cache for Inbox sends, preventing a locally rendered message from being replaced after newer activity
- keep a selected Inbox DM subscription live and hydrate recent DM content with ordinary NIP-10 replies
- add unit and browser regression coverage for both the hidden-agent-reply and Inbox-send paths

Fixes #5632.

## Root cause

The screenshot's highlighted Inbox send is signed event `929865d3c0…`. Its only tag is the DM channel's `h` tag: it has neither a participant `p` tag nor a thread `e` tag.

Inbox called `sendChannelMessage` directly with only the composer's explicit `@mention` pubkeys. When the user typed no explicit mention, the event was published as `h`-only. The normal channel composer instead uses `useSendMessageMutation`, whose `messageMentionPubkeys` step adds every other DM participant. Without Honey's `p` tag, the managed-agent harness was not addressed, explaining the missing working reaction, activity, and reply.

That separate Inbox send path also bypassed the shared `channelWindowStore` optimistic/success reconciliation used by the DM conversation. A locally rendered Inbox row could therefore be replaced when a later send or refresh re-projected the authoritative channel window.

There was a second projection gap: managed-agent responses sent as ordinary NIP-10 replies are intentionally absent from stream-style top-level channel windows. Inbox previously treated a DM as a flat conversation but consumed only that top-level projection, so a persisted agent reply could be visible in activity while absent from Inbox DM detail.

## Fix

Inbox now uses the shared send mutation for every supported channel. For DMs this adds the implicit participant tags and updates the same cache consumed by the conversation view.

Selected Inbox DMs also keep their subscription live, retain normal reply events in the DM window, and merge a recent direct channel-content fetch into Inbox context.

## Testing

Verified after merging current `main` on commit `268907c91c0e22c593b005e4a3fd87ae289d0ff9`:

- `just ci` — passed (formatting, lint/type/file-size gates, Rust, desktop, web, and mobile suites)
- `cd desktop && pnpm typecheck` — passed
- `cd desktop && pnpm build:e2e && pnpm exec playwright test tests/e2e/channels.spec.ts --grep 'Inbox DM send addresses the participant and survives a newer channel send'` — passed
- the browser regression asserts that an Inbox DM send carries the other participant's `p` tag, remains visible after opening the conversation, and survives a newer send


ScaleLeanChris:
Independent verification at `45c60632f`:

- Focused `useInboxThreadContext` unit test: 1 passed, 0 failed.
- `pnpm --dir desktop typecheck`: passed.
- E2E production build: passed.
- Targeted browser regression, `Inbox DM send addresses the participant and survives a newer channel send`: 1 passed, 0 failed.

The browser test verifies the important relationship: an Inbox DM send gains the other participant `p` tag through the shared send path, remains visible after opening the conversation, and survives a newer send. This is the right seam for managed-agent delivery because it removes the divergent Inbox publisher instead of adding another agent-specific exception.

